Issue #34, opened 2017-01-24
Add section on Revocation


msporny:
It is currently not clear how Verifiable Claims may be revoked by the issuer. We should add a section to the specification that clarifies the various mechanisms for revocation of claims.

Discussion on this issue started here: https://github.com/opencreds/vc-use-cases/issues/33#issuecomment-273237017

msporny:
ACTION: Manu to write section on Revocation for the Data Model spec.

msporny:
This issue is waiting on reviews for the PR.

burnburn:
Note that you will also need to add Syntax sections as well.  PR #36 only adds the model.

msporny:
I added one example for the time being, we'll have to sort out what the syntax looks like via https://github.com/opencreds/vc-data-model/issues/35.

PR #36, opened 2017-02-06
Add section on the Revocation Model.

msporny:
This PR is meant to add a section on revocation to address issue #34.

### DIFF
--- a/index.html
+++ b/index.html
@@ -391,7 +391,7 @@ WebIDL.</p>
         <h2>Revocation Model</h2>
         <p>Revocation information for the claims in the
         <a>Verifiable Claims Model</a> may be provided by adding the
-        following property:
+        following <a>property</a>:
         </p>
 
         <dl>
--- a/index.html
+++ b/index.html
@@ -390,18 +390,20 @@ WebIDL.</p>
       <section>
         <h2>Revocation Model</h2>
         <p>A <a>Verifiable Claim</a> may provide revocation information
-        by referring to a revocation list by adding the following
+        by referring to a revocation scheme by adding the following
         <a>property</a> to the <a>Verifiable Claims Model</a>:</p>
         <dl>
-          <dt><var>revocationList</var></dt>
-          <dd>The format of the revocation list will vary depending on the
-          signature scheme used, but will almost always be a URL to a
-          resource that contains zero or more revocation entries.
+          <dt><var>revocation</var></dt>
+          <dd>The value of this property MUST be a revocation scheme that
+            provides enough information to determine whether or not the
+            credential has been revoked. The revocation scheme will vary
+            depending on a variety of factors, such as whether it is
+            simple to implement or privacy-enhancing.
           </dd>
         </dl>
 
         <p class="issue" data-number="35" data-title="Define ONE concrete format for the revocationList parameter ">
-What is the format of the revocation list? Options include 1)
+What is the format of a simple revocation list? Options include 1)
 a list of strings, which are IDs, 2) a list of JSON objects each with an 'id'
 entry, 3) a data-model agnostic list of items that each have an id that is
 associated with the provided claim. If the issued verifiable claim's id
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -386,6 +386,30 @@
...
+      <section>
+        <h2>Revocation Model</h2>
+        <p>A <a>Verifiable Claim</a> may provide revocation information
+        by referring to a revocation list by adding the following
+        <a>property</a> to the <a>Verifiable Claims Model</a>:</p>
+        <dl>
+          <dt><var>revocationList</var></dt>
+          <dd>The format of the revocation list will vary depending on the
+          signature scheme used, but will almost always be a URL to a
+          resource that contains zero or more revocation entries.
```
dlongley: "the signature scheme used, but will almost always be a URL" => "the revocation scheme used. One simple example format is a URL that contains zero or more revocation entries, where the entry format is specified via the revocation scheme."

1. I don't think we should say "will almost ways" anything, we don't know.
2. The revocation scheme determines the format, not the signature scheme (directly). It may be that a signature scheme specifies a recommended revocation scheme, but it is most directly specified through the revocation scheme.

index.html
```diff
@@ -386,6 +386,30 @@
...
+          </dd>
+        </dl>
+
+        <p class="issue" data-number="35" data-title="Define ONE concrete format for the revocationList parameter ">
+What is the format of the revocation list? Options include 1)
```
dlongley: "What is the format of the revocation list" => "What is the format of a simple revocation list?"

index.html
```diff
@@ -386,6 +386,30 @@
...
+a list of strings, which are IDs, 2) a list of JSON objects each with an 'id'
+entry, 3) a data-model agnostic list of items that each have an id that is
+associated with the provided claim. If the issued verifiable claim's id
+is in the list of IDs, then its status is revoked. Note that the mechanism
+used for pseudonymous revocation will be calculated differently, which is
+why the format of the revocation list is signature scheme specific.
```
dlongley: "signature scheme specific" => "is revocation scheme specific"

I'm not convinced we should even be calling this a "revocationList", at least not at this level. I think we should specify a "revocationScheme" (bike shed) and we can define a "SimpleRevocationList" scheme which would include a "revocationList" property (at that level) that points to a list -- where the list entry format can be further refined.

index.html
gkellogg: Can we say **should**?

gkellogg: "A URL, which when dereferenced results in returns zero or more revocation entries."

index.html
gkellogg: Confusing, is the format of the value of the `revocationList` entry? Or, the format of the result of a GET against this list?

index.html
msporny: Alright, I'll update based on this feedback and make the entry more generic. I would imagine this feature would end up following what we do for "signature"... that is, you can have different revocation schemes just like you can have different signature schemes. This provides the greatest amount of flexibility. I imagine we'll also define and implement at least one revocation scheme during the VCWG's timeframe.

index.html
msporny: fixed.

index.html
msporny: I'm rewording now... will try to clarify all of this with a second revision.

index.html
```diff
@@ -386,6 +386,27 @@
           a representation of the signing date.</dd>
         </dl>
       </section>
+
+      <section>
+        <h2>Revocation Model</h2>
+        <p>A <a>Verifiable Claim</a> may provide revocation information
+        by referring to a revocation scheme by adding the following
+        <a>property</a> to the <a>Verifiable Claims Model</a>:</p>
```
burnburn: Suggest replacing all the above text with the following simplified wording to better match the existing model language:

<p>Revocation information for the claims in the <a>Verifiable Claims Model</a> may be provided by adding the following <a>property</a>:

since you already explain in the revocation property description that it must be a revocation scheme.

burnburn: (note that all styling was removed from what I typed above.  The 'Verifiable Claims Model' and 'property' should both be hyperlinks)

msporny: Done.

PR #39, opened 2017-02-28
Add section on storage providers and data mining.

msporny:
Please review. Addresses issue https://github.com/opencreds/vc-data-model/issues/10.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1267,17 +1267,17 @@ Inspector (corporation) is required to check revocation via Issuer
         <h3>Storage Providers and Data Mining</h3>
 
         <p>
-When a holder receives a claim from an issuer, it is highly likely that the
+When a holder receives a claim from an issuer, the
 claim will need to be stored somewhere (e.g. in a credential repository).
-Holders are warned that the information in a verifiable claim is often
+Holders are warned that the information in a verifiable claim may be
 sensitive in nature and highly individualized, making it a high value
-target for data mining. It is expected that there will be many services
+target for data mining. Therefore, there may be services
 that store verifiable claims for free and mine personal data and sell it
 to organizations that desire individualized profiles on people and
 organizations (i.e. if the service is free, you are the product).
         </p>
         <p>
-It is suggested that holders are aware of the terms of service for their
+It is suggested that holders be aware of the terms of service for their
 credential repository, specifically the correlation and data mining
 protections that are in place for those who store their verifiable claims
 at the service provider.
--- a/index.html
+++ b/index.html
@@ -1264,11 +1264,42 @@ Inspector (corporation) is required to check revocation via Issuer
       </section>
 
       <section>
-        <h3>Storage Location of Claims</h3>
+        <h3>Storage Providers and Data Mining</h3>
 
-        <p class="issue">
-Storage location of claims may mine data or forward data.
+        <p>
+When a holder receives a claim from an issuer, it is highly likely that the
+claim will need to be stored somewhere (e.g. in a credential repository).
+Holders are warned that the information in a verifiable claim is often
+sensitive in nature and highly individualized, making it a high value
+target for data mining. It is expected that there will be many services
+that store verifiable claims for free and mine personal data and sell it
+to organizations that desire individualized profiles on people and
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1264,11 +1264,42 @@
       </section>
 
       <section>
-        <h3>Storage Location of Claims</h3>
+        <h3>Storage Providers and Data Mining</h3>
 
-        <p class="issue">
-Storage location of claims may mine data or forward data.
+        <p>
+When a holder receives a claim from an issuer, it is highly likely that the
```
dlongley: No need to say "highly likely", just "the claim may need to be stored somewhere"

index.html
```diff
@@ -1264,11 +1264,42 @@
...
+claim will need to be stored somewhere (e.g. in a credential repository).
+Holders are warned that the information in a verifiable claim is often
```
dlongley: Generally speaking "often" can be replaced with "may" to avoid having to back up these sorts of declarations.

index.html
```diff
@@ -1264,11 +1264,42 @@
...
+sensitive in nature and highly individualized, making it a high value
+target for data mining. It is expected that there will be many services
+that store verifiable claims for free and mine personal data and sell it
+to organizations that desire individualized profiles on people and
+organizations (i.e. if the service is free, you are the product).
```
dlongley: Let's not say "it is expected" but we can say "Therefore, there may be services..."

index.html
```diff
@@ -1264,11 +1264,42 @@
...
+        </p>
+        <p>
+It is suggested that holders are aware of the terms of service for their
```
dlongley: "holders are aware" => "holders be aware"

agropper:
From a  privacy engineering perspective, we could note that a privacy sensitive claim could deserve its own one-time did so that miners would not be able to mount a correlation attack against the subject. In the many cases that require verified claims, ID verification can be done by the issuer of the claim _without_ adding the verified subject ID to the claim itself. The issuer still bears the liability of verification but the claim itself is not as valuable to miners.

index.html
msporny: Fixed.

index.html
msporny: Fixed.

index.html
msporny: Fixed.

index.html
msporny: Fixed.

msporny:
@agropper The current design direction is to not use a one-time DID at all. There is the notion of subject-less verifiable claims, which is effectively a bearer claim. The verifiable claim essentially says "There exists a person that is over the age of 16, and I have given this verifiable claim to that person, if they show up with it, I assert that they are over the age of 16". Now, clearly, that doesn't have the security characteristics that one might want to give a person access to top secret information, but it does have the security characteristics that may be good enough to meet a regulatory burden to ensure that a teenager is of a certain age before letting them into a movie theater.

agropper:
What keeps the teenager with a subject-less verifiable claim from copying or just sharing the claim with his pals?

msporny:
@agropper Nothing, which is why bearer claims aren't that great for high-risk things. We categorize single-purpose DIDs as effectively bearer claims. Keep in mind that the attack using a single-purpose DID is almost exactly the same as what you describe above. What's to keep a teenager with a single-purpose DID from copying/sharing the DID cryptokeys and claim with his pals (via a shared account)?

agropper:
@msporny A single-purpose DID could still be registered with an "oversight agency" that is separate from the issuer of the claim. The oversight agency's sole purpose is correlation around how that claim is used. In my e-prescription example, the oversight agency is the state. The issuer of the claim might be the prescriber and the inspector is the pharmacy. The subject has a right not to be tracked by anyone other than the state unless the state decides otherwise.

To answer your question about sharing the DID cryptokeys and claims, the answer is nothing except that the inspector would be looking at some other claim specific to the particular holder (e.g. the driver's license photo and license number) and filing the holder's license number along with other aspects of the transaction with the state oversight agency in order to fulfill their regulatory mandate. The same, or another pal in the shared DID account could try to do the same thing at the same or another pharmacy but they would be correlated and flagged by the oversight agent.

Issue #9, opened 2016-11-28
List all validity checks that should be performed


msporny:
We should make it clear that an inspector (e.g. corporation) is required to check revocation via the Issuer (e.g. government). In fact, there are many different types of validity checks that should be performed on data, such as the revocation status of an Issuer's signing key, the expiration date of the claim, etc.

erickorb:
+1

David-Chadwick:
This is not exactly true. The issuer may delegate the issuing of revocation information to a TTP. And if several issuers use a common TTP this is much better for privacy protection, as it makes it much more difficult for the issuer or the TTP to know which user/holder has contacted which inspector.

Change text to "is required to check revocation via the Issuer or its delegate"

Perhaps we should also recommend that revocation lists should preferably be published by TTPs that do not inform the issuer which inpsectors have contacted it, and that a TTP may merge the lists of multiple issuers.

msporny:
@David-Chadwick We're also working on Blockchain-based revocation lists for such purposes.

jonnycrunch:
By TTP, you mean Trusted Third Party.  Can you give an example of a revocation performed by a TTP?  A lot of TTP that I know of only aggregate the credentials and point the use to the source of truth, including revocation of the credential.   Typically the TTP covers themselves by giving a valid period or expiration date and upon renewal, they perform a validity check before issuing a new credential.

msporny:
> By TTP, you mean Trusted Third Party

Yes.

> Can you give an example of a revocation performed by a TTP?

Company A outsources all their credential management to Outsource B. Outsource B issues credentials under Company A's private key, and places the revocation list information in a bulk revocation list (for all customers). For example, the Department of Motor Vehicles outsources the issuance of digital drivers licenses , which are good for 7+ years. The outsourced firm maintains a revocation list for multiple DMVs, as that is their business model.

ottonomy:
I think the discussion gets a little into the weeds. Revocation will be checked in the manner that the credential tells the inspector to check for it. There will be multiple methods, but that doesn't change the task in the ticket, which is to list the checks that must be performed. Revocation is one of them (if the credential offers a method to check for revocation -- some may not).

Here are some validity checks to consider when verifying a Verifiable Claim:
* Document is valid JSON-LD.
* Required properties are present in the document.
* The issuer's `@id` may be accessed.
* The issued date is published and is in the expected range (i.e. not in the future).
[...]

PR #44, opened 2017-03-16
Add verification check instructions. Relates to #9.

ottonomy:
Adds some verification check instructions to page. Abstracts some text from the list I originally drafted in #9 about how profiles and keys are discovered to account for different possible ways these may be retrieved, accessed, and trusted.

@dlongley the data model for claim above does declare `id` is a required property, so I described it as such here as well. If the bearer token case you mentioned is to be taken into account, it'll require a change above as well.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1381,6 +1381,47 @@ Time periods should be shorter for highly dynamic information.
       </section>
     </section>
 
+    <section>
+      <h2>Verification</h2>
+
+      <p>
+This section describes a number of checks required to verify a claim.
+Some checks are essential for all verifiable claims, while some are applicable to only some claims.
+      </p>
+      <section>
+        <h3>Structural Validity</h3>
+        <p>
+          <ul>
+            <li>Document is valid JSON-LD.</li>
+            <li>Required properties are present. For example, for a Credential, <code>id</code>, <code>type</code>, and <code>claim</code> are required.</li>
+            <li>Property values match expectations described in the specification. For example, the document type for a verifiable claim must contain the class "Credential".</li>
+          </ul>
+        </p>
+      </section>
+      <section>
+        <h3>Entity Validity</h3>
+        <p>A number of checks </p>
+        <p>
+          <ul>
+            <li>The <code>issuer</code> id must match expectations. Likely, that means it is the id of a known and trusted <a>identity profile</a>.</li>
+            <li>The <a>claim id</a> must match expectations. Likely, that means it is the id of a known and trusted <a>identity profile</a> for the subject of the claim. If the entity that is subject of a claim has transmitted it to the inspector, the subject may be able to prove ownership of key identifyin properties such as email address(es) and public key(s).</li>
+            <li>The <code>issued</code> date must be in the expected range. For example, an inspector may wish to ensure that the recorded issued date of valid claims is not in the future.</li>
+            <li>The document signature is available in the form of a known signature suite.</li>
+            <li>The public key associated with the signature is available and a trustworthy link between this signing key and the issuer's <a>identify profile</a> may be established.</li>
+            <li>If revocation instructions are present, the claim must not have been revoked.</li>
+          </ul>
+        </p>
+      </section>
+      <section>
+        <h3>Fitness for Purpose</h3>
+        <p>
+          <ul>
+            <li>The <a>custom properties</a> in the <a>claim</a> are appropriate for the inspector's purpose. For example if an inspector needs to determine that a subject is older than 21 years of age, they may accept claims of specific birthdate or abstract properties such as <code>ageOver</code>.</li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1381,6 +1381,47 @@
...
+      <p>
+This section describes a number of checks required to verify a claim.
+Some checks are essential for all verifiable claims, while some are applicable to only some claims.
+      </p>
+      <section>
+        <h3>Structural Validity</h3>
+        <p>
+          <ul>
+            <li>Document is valid JSON-LD.</li>
+            <li>Required properties are present. For example, for a Credential, <code>id</code>, <code>type</code>, and <code>claim</code> are required.</li>
```
msporny: id isn't required, even for a Credential (bearer token). So, we will want to remove 'id' as a requirement.

index.html
```diff
@@ -1381,6 +1381,47 @@
...
+          </ul>
+        </p>
+      </section>
+      <section>
+        <h3>Entity Validity</h3>
+        <p>A number of checks </p>
+        <p>
+          <ul>
+            <li>The <code>issuer</code> id must match expectations. Likely, that means it is the id of a known and trusted <a>identity profile</a>.</li>
+            <li>The <a>claim id</a> must match expectations. Likely, that means it is the id of a known and trusted <a>identity profile</a> for the subject of the claim. If the entity that is subject of a claim has transmitted it to the inspector, the subject may be able to prove ownership of key identifyin properties such as email address(es) and public key(s).</li>
```
msporny: typo - 'identifyin'

index.html
```diff
@@ -1381,6 +1381,47 @@
...
+This section describes a number of checks required to verify a claim.
+Some checks are essential for all verifiable claims, while some are applicable to only some claims.
+      </p>
+      <section>
+        <h3>Structural Validity</h3>
+        <p>
+          <ul>
+            <li>Document is valid JSON-LD.</li>
+            <li>Required properties are present. For example, for a Credential, <code>id</code>, <code>type</code>, and <code>claim</code> are required.</li>
+            <li>Property values match expectations described in the specification. For example, the document type for a verifiable claim must contain the class "Credential".</li>
```
msporny: Wrap all lines to 75 characters (try to follow the format of the rest of the document). It is inevitable that we'll need to reformat the spec at some point because of inconsistencies that sneak in, but we try to more or less maintain the same format throughout the spec.

index.html
```diff
@@ -1381,6 +1381,47 @@
...
+          </ul>
+        </p>
+      </section>
+      <section>
+        <h3>Entity Validity</h3>
+        <p>A number of checks </p>
```
msporny: This looks like an incomplete sentence?

index.html
msporny: 'claim id', should probably be 'subject identifier' or 'subject of the claim'. We need to make sure this terminology exists in the terminology section (I'm working on that and will make sure we have whatever you put in here in the terminology section).

index.html
```diff
@@ -1381,6 +1381,47 @@
...
+        <p>
+          <ul>
+            <li>The <code>issuer</code> id must match expectations. Likely, that means it is the id of a known and trusted <a>identity profile</a>.</li>
+            <li>The <a>claim id</a> must match expectations. Likely, that means it is the id of a known and trusted <a>identity profile</a> for the subject of the claim. If the entity that is subject of a claim has transmitted it to the inspector, the subject may be able to prove ownership of key identifyin properties such as email address(es) and public key(s).</li>
+            <li>The <code>issued</code> date must be in the expected range. For example, an inspector may wish to ensure that the recorded issued date of valid claims is not in the future.</li>
+            <li>The document signature is available in the form of a known signature suite.</li>
+            <li>The public key associated with the signature is available and a trustworthy link between this signing key and the issuer's <a>identify profile</a> may be established.</li>
```
msporny: typo - 'identify'

We may also want to say that verifiers should check the signature according to the signature verification algorithm associated with the signature suite in use. Different signature suites will have different things that need to be checked/verified.

index.html
```diff
@@ -1381,6 +1381,47 @@
...
+            <li>If revocation instructions are present, the claim must not have been revoked.</li>
+          </ul>
+        </p>
+      </section>
+      <section>
+        <h3>Fitness for Purpose</h3>
+        <p>
+          <ul>
+            <li>The <a>custom properties</a> in the <a>claim</a> are appropriate for the inspector's purpose. For example if an inspector needs to determine that a subject is older than 21 years of age, they may accept claims of specific birthdate or abstract properties such as <code>ageOver</code>.</li>
```
msporny: "are" -> "should be"?

ottonomy:
@msporny I think I've responded to all the requested changes. I didn't disagree with any of them. Thanks!

Issue #46, opened 2017-05-30
Add section/language about presentation of claims


jandrieu:
The dominant mental model is that a singular claim is issued and later presented to an inspector. However, there has been significant discussion about how a subset of a credential may be presented as well as how claims from multiple issuers may be aggregated for presentation.  

This significantly impacts privacy around the issue of minimal disclosure and selective disclosure. 

In particular, what is the thing that is presented when it is not a single credential from a single issuer? Given a name for that we can flesh out how a controller might construct such presentations from multiple credentials. Right now, there is no language about this.

jandrieu:
Based on our conversations last meeting, https://www.w3.org/2017/06/20-vcwg-minutes.html, I think it's clear we don't have a consistent term that we use for what is presented to the inspector.

We do have the terms credentials and entity profiles.

However, we do *not* describe the workflow as 

1. Issuers issue credentials
2. Claimants create entity profiles combining one or more claims from one or more credentials
3. Claimants present entity profiles to Verifiers
4. Verifiers verify the claims are statements made by the issuer

And I think we should. The current equivalent language is
1. An Issuer issues Claims about Subjects using Identifiers.
2. Claimant presents claims (either directly or indirectly) to an Verifier.
3. A Verifier verifies that the Claims represent statements made by the original Issuer.

Even if the terms I use above aren't quite ideal, I think we need to shift our language to use language similar to this to highlight that the lifecycle of a claim is far more flexible than traditional, static credentials where you *do* have to give your entire driver's license to the bouncer to get into the bar.

jandrieu:
The poll terminology may be fine as long as we clarify in the FPWD that this is in the simplest case of a single claim. To address issues of minimal and selective disclosure we should then also expand that flow to clarify the terms credential and entity profile.

I think our conversations often stumble because we conflate credentials, claims, and entity-profiles.

I'd prefer to just have the one inclusive flow from credentials through to profiles and for us to gradually learn to be more rigorous in how we talk about it. 

For instance, I think it makes far more sense that issuers issue credentials, and that claimants present profiles. Period. The claims themselves are the individually verifiable statements within those credentials and profiles, but even when a credential has a single claim, it is clearer to discuss the issuer issuing the credential. Similarly, when a profile has a single claim, calling it a profile makes it clear the place it has in the process, who controls it, and who receives it.

I believe that clarity would improve the dialog.

David-Chadwick:
[...]

PR #49, opened 2017-06-01
Refactor some uses of "identity" to "entity".

msporny:
This is another semi-controversial PR. It attempts to align our uses of the words "identity" and "entity". We had previous had both of these terms active in the document: "Identity Profile" and "Entity Credential". This PR changes "Identity Profile" to "Entity Profile" in order to align it with "Entity Credential".

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,25 +1,24 @@
 # Verifiable Claims Task Force
 
-Contributions to this repository are intended to become part of Recommendation-track documents 
-governed by the [W3C Patent Policy](https://www.w3.org/Consortium/Patent-Policy-20040205/) and
-[Document License](https://www.w3.org/Consortium/Legal/copyright-documents). To contribute, you must 
-either participate in the relevant W3C Working Group or make a non-member patent licensing
- commitment.
+Contributions to this repository are intended to become part of Recommendation-track documents governed by the
+[W3C Patent Policy](https://www.w3.org/Consortium/Patent-Policy-20040205/) and
+[Software and Document License](https://www.w3.org/Consortium/Legal/copyright-software). To make substantive contributions to specifications, you must either participate
+in the relevant W3C Working Group or make a non-member patent licensing commitment.
 
 If you are not the sole contributor to a contribution (pull request), please identify all 
-contributors in the pull request's body or in subsequent comments.
+contributors in the pull request comment.
 
- To add a contributor (other than yourself, that's automatic), mark them one per line as follows:
+To add a contributor (other than yourself, that's automatic), mark them one per line as follows:
 
- ```
- +@github_username
- ```
+```
++@github_username
+```
 
- If you added a contributor by mistake, you can remove them in a comment with:
+If you added a contributor by mistake, you can remove them in a comment with:
 
- ```
- -@github_username
- ```
+```
+-@github_username
+```
 
- If you are making a pull request on behalf of someone else but you had no part in designing the 
- feature, you can remove yourself with the above syntax.
+If you are making a pull request on behalf of someone else but you had no part in designing the 
+feature, you can remove yourself with the above syntax.
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,2 +1,4 @@
-All documents in this Repository are licensed by contributors under the [W3C Document
-License](https://www.w3.org/Consortium/Legal/copyright-documents).
+All documents in this Repository are licensed by contributors
+under the 
+[W3C Software and Document License](https://www.w3.org/Consortium/Legal/copyright-software).
+
--- a/common.js
+++ b/common.js
@@ -1,0 +1,207 @@
+/* globals omitTerms, respecConfig, $, require */
+/* exported linkCrossReferences, restrictReferences, fixIncludes */
+
+var vcwg = {
+  // Add as the respecConfig localBiblio variable
+  // Extend or override global respec references
+  localBiblio: {
+    "REST": {
+      title: "Architectural Styles and the Design of Network-based Software Architectures",
+      date: "2000",
+      href: "http://www.ics.uci.edu/~fielding/pubs/dissertation/",
+      authors: [
+        "Fielding, Roy Thomas"
+      ],
+      publisher: "University of California, Irvine."
+    },
+    "VC-USECASES": {
+      title: "Verifiable Claims Use Cases",
+      href: "http://w3c.github.io/vc-use-cases/",
+      authors: [
+      	"Shane McCarron",
+        "Daniel Burnett",
+        "Gregg Kellogg",
+        "Brian Sletten",
+        "Manu Sporny"
+      ],
+      status: "IG-DRAFT",
+      publisher: "Web Payments Interest Group Verifiable Claims Task Force"
+    },
+    // aliases to known references
+    "HTTP-SIGNATURES": {
+      aliasOf: "http-signatures"
+    },
+    "MACAROONS": {
+      title: 'Macaroons',
+      // TODO: create spec
+      href: 'http://macaroons.io/',
+      authors: ['Arnar Birgisson', 'Joe Gibbs Politz', 'Úlfar Erlingsson',
+        'Ankur Taly', 'Michael Vrable', 'Mark Lentczner'],
+      status: 'unofficial',
+      publisher: 'Credentials Community Group'
+    },
+    'OPEN-BADGES': {
+      title: 'Open Badges',
+      href: 'https://github.com/openbadges/openbadges-specification',
+      authors: ['Brian Brennan', 'Mike Larsson', 'Chris McAvoy',
+        'Nate Otto', 'Kerri Lemoie'],
+      status:   'BA-DRAFT',
+      publisher:  'Badge Alliance Standard Working Group'
+    },
+    'RDF-NORMALIZATION': {
+      title: 'RDF Dataset Normalization',
+      href: 'http://json-ld.github.io/normalization/spec/',
+      authors: ['Dave Longley', 'Manu Sporny'],
+      status:   'CG-DRAFT',
+      publisher:  'Credentials W3C Community Group'
+    },
+  }
+};
+
+
+
+// We should be able to remove terms that are not actually
+// referenced from the common definitions
+//
+// the termlist is in a block of class "termlist", so make sure that
+// has an ID and put that ID into the termLists array so we can 
+// interrogate all of the included termlists later.
+var termNames = [] ;
+var termLists = [] ;
+var termsReferencedByTerms = [] ;
+
+function restrictReferences(utils, content) {
+    "use strict";
+    var base = document.createElement("div");
+    base.innerHTML = content;
+
+    // New new logic:
+    //
+    // 1. build a list of all term-internal references
+    // 2. When ready to process, for each reference INTO the terms, 
+    // remove any terms they reference from the termNames array too.
+    $.each(base.querySelectorAll("dfn"), function(i, item) {
+        var $t = $(item) ;
+        var titles = $t.getDfnTitles();
+        var dropit = false;
+        // do we have an omitTerms
+        if (window.hasOwnProperty("omitTerms")) {
+            // search for a match
+            $.each(omitTerms, function(j, term) {
+                if (titles.indexOf(term) !== -1) {
+                    dropit = true;
+                }
+            });
+        } 
+        // do we have an includeTerms
+        if (window.hasOwnProperty("includeTerms")) {
+            var found = false;
+            // search for a match
+            $.each(includeTerms, function(j, term) {
+                if (titles.indexOf(term) !== -1) {
+                    found = true;
+                }
+            });
+            if (!found) {
+                dropit = true;
+            }
+        }
+        if (dropit) {
+            $t.parent().next().remove();
+            $t.parent().remove();
+        } else {
+            var n = $t.makeID("dfn", titles[0]);
+            if (n) {
+                termNames[n] = $t.parent() ;
+            }
+        }
+    });
+
+    var $container = $(".termlist",base) ;
+    var containerID = $container.makeID("", "terms") ;
+    termLists.push(containerID) ;
+
+    return (base.innerHTML);
+}
+// add a handler to come in after all the definitions are resolved
+//
+// New logic: If the reference is within a 'dl' element of
+// class 'termlist', and if the target of that reference is
+// also within a 'dl' element of class 'termlist', then
+// consider it an internal reference and ignore it.
+
+require(["core/pubsubhub"], function(respecEvents) {
+    "use strict";
+    respecEvents.sub('end', function(message) {
+        if (message === 'core/link-to-dfn') {
+            // all definitions are linked; find any internal references
+            $(".termlist a.internalDFN").each(function() {
+                var $r = $(this);
+                var id = $r.attr('href');
+                var idref = id.replace(/^#/,"") ;
+                if (termNames[idref]) {
+                    // this is a reference to another term
+                    // what is the idref of THIS term?
+                    var $def = $r.closest('dd') ;
+                    if ($def.length) {
+                        var $p = $def.prev('dt').find('dfn') ;
+                        var tid = $p.attr('id') ;
+                        if (tid) {
+                            if (termsReferencedByTerms[tid]) {
+                                termsReferencedByTerms[tid].push(idref);
+                            } else {
+                                termsReferencedByTerms[tid] = [] ;
+                                termsReferencedByTerms[tid].push(idref);
+                            }
+                        }
+                    }
+                }
+            }) ;
+
+            // clearRefs is recursive.  Walk down the tree of 
+            // references to ensure that all references are resolved.
+            var clearRefs = function(theTerm) {
+                if ( termsReferencedByTerms[theTerm] ) {
+                    $.each(termsReferencedByTerms[theTerm], function(i, item) {
+                        if (termNames[item]) {
+                            delete termNames[item];
+                            clearRefs(item);
+                        }
+                    });
+                }
+                // make sure this term doesn't get removed
+                if (termNames[theTerm]) {
+                    delete termNames[theTerm];
+                }
+            };
+           
+            // now termsReferencedByTerms has ALL terms that 
+            // reference other terms, and a list of the 
+            // terms that they reference
+            $("a.internalDFN").each(function () {
+                var $item = $(this) ;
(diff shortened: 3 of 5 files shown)


jandrieu:
The refactoring of index.html is excellent. +1
I have no comments on common.js

However, the terms.html document seems far too immature for adoption. A few quotes that give me pause:

> claim
A statement made by an entity about the subject defined by an entity profile.

Why does the subject need an entity profile?  We've discussed the possibility of making statements about claims, which would imply a claim either needs or is an entity profile. This doesn't quite flow.
Perhaps: "A claim is a statement made by one entity about another. Claims are made by an issuer about a subject. The subject could also be the issuer, i.e., a claim one makes about oneself."

> creator
The entity that is associated with creating a particular resource. For example, the key that created a digital signature.

"Keys" don't create. They are inert. I would expect that the "creator" is one of:
* the individual who used a key to create a signature
* the individual who is recognized as controlling the key used to sign a claim

> credential inspector
An entity that requests a credential for processing.

I think this is reversed from its actual meaning. The inspector is the entity that receives a set of selected claims for verification. Whether or not they request them is not as material as the role they play in performing the verification.

> entity credential
A set of claims made by an entity about an entity profile referred to as the subject of the claims. An entity credential may refer to a qualification, achievement, personal quality, or other information abouth the subject such as a name, government ID, preferred payment processor, home address, or university degree that is typically used to indicate suitability.

This also appears to have the incorrect level of indirection. Claims are, generally, not made about entity profiles. They are made about subjects. A collection of claims may be an entity profile, and yes, we may make statements about them, but I think that is the edge case of claims about claims and not the formative use we all expect: claims about people, places, and things

Also "refer" lacks rigor in " An entity credential may refer to". I believe the intention is something closer to "an entity credential may represent".

What seems to be converging for me as I read this set of terms is that a credential is a set of one or more claims issued by a single issuer and a profile is a set of one or more claims issued by one or more issuers about a common subject. In other words, "profile" may be the missing term and section referenced in #46. If so, inspectors receive profiles and validate the contained claims. 

>  entity profile
A set of information that can be used to identify a particular entity. An entity may have multiple entity profiles associated with it.

The information may not be used to identify, i.e., correlate, any particular entity. Such information may be used to provide enhanced services. One might use other information for correlation/assurance that a given identifier refers to the subject in question, then use statements made referring to that identifier to customize interactions, e.g., localizing a search. The profile then is less about identifying the individual than ascribing attributes to them.

I would simplify it to say an entity profile is a "collection of claims related to the same subject. An entity may have multiple profiles and each profile may contain claims from multiple different issuers." We could add "Claims may be used to correlate subjects across services and sessions, or they may be used to tailor services for that particular subject."

There is a subtle distinction about whether or not the profile is the set of claims *that is presented* or if it could/should also refer to any collection of claims about a common subject, whether presented or not--and even whether or not the profile is ever in a single location. The latter addresses the fluidity of distributed credentials in any number of stores; the former focuses the conversation on what an inspector receives (which makes it easier to define a concrete data model).

On the whole, it's a good first stab at formalizing a bunch of this language. Some of it needs more work.

msporny:
> However, the terms.html document seems far too immature for adoption.

Keep in mind that terms.html is a source document where only SOME of the terms are pulled into the vc-data-model spec. We wouldn't be publishing terms.html in its entirety, just a subset of it (the part that actually shows up in the vc-data-model spec).

That said, I'll try to focus on the parts that we do use in the spec, noting the ones that we need to work on for later.

Also, having read through the spec /after/ making the changes, I'm not very happy with how convoluted the language has become, so will try to submit another PR to clean that up as well. The spec has readability issues at present that need to be resolved.

msporny:
>> claim
>> A statement made by an entity about the subject defined by an entity profile.
>
>Why does the subject need an entity profile? We've discussed the possibility of making statements about claims, which would imply a claim either needs or is an entity profile. This doesn't quite flow.
Perhaps: "A claim is a statement made by one entity about another. Claims are made by an issuer about a subject. The subject could also be the issuer, i.e., a claim one makes about oneself."

You're right, the subject shouldn't need an entity profile. This is a holdover from a past edit. I've removed "defined by an entity profile" in https://github.com/w3c/vc-data-model/commit/77694544b1bd3263092a43b5b95ebed448aa2b8d.

msporny:
>> creator
>> The entity that is associated with creating a particular resource. For example, the key that created a digital signature.
>
> "Keys" don't create. They are inert. I would expect that the "creator" is one of:
> * the individual who used a key to create a signature
> * the individual who is recognized as controlling the key used to sign a claim

I've removed this old definition in https://github.com/w3c/vc-data-model/commit/18733fd2d5a991294dc4a800a20748e77725a493.

msporny:
>> credential inspector
>> An entity that requests a credential for processing.
>
> I think this is reversed from its actual meaning. The inspector is the entity that receives a set of selected claims for verification. Whether or not they request them is not as material as the role they play in performing the verification.

Fixed in https://github.com/w3c/vc-data-model/commit/46dc5a05ade3298d44bf58bb3a98875c13b4ef6e. Keep in mind that we don't pull this definition into the spec yet.

msporny:
>> entity credential
>>A set of claims made by an entity about an entity profile referred to as the subject of the claims. An entity credential may refer to a qualification, achievement, personal quality, or other information abouth the subject such as a name, government ID, preferred payment processor, home address, or university degree that is typically used to indicate suitability.
>
>This also appears to have the incorrect level of indirection. Claims are, generally, not made about entity profiles. They are made about subjects. A collection of claims may be an entity profile, and yes, we may make statements about them, but I think that is the edge case of claims about claims and not the formative use we all expect: claims about people, places, and things
>
> Also "refer" lacks rigor in " An entity credential may refer to". I believe the intention is something closer to "an entity credential may represent".
>
> What seems to be converging for me as I read this set of terms is that a credential is a set of one or more claims issued by a single issuer and a profile is a set of one or more claims issued by one or more issuers about a common subject. In other words, "profile" may be the missing term and section referenced in #46. If so, inspectors receive profiles and validate the contained claims.

I simplified using your terminology in the last paragraph: https://github.com/w3c/vc-data-model/commit/358fbecf0c5742f0445c919f2aa89a9220816eef.

jandrieu:
Getting better. I think the two credential definitions could be reduced to one. I'm not sure there's a distinction other than a historical one.  Perhaps:

> A set of one or more claims made by the same <a>entity</a> about a<a>subject</a>. A credential
may refer to a qualification, achievement, personal quality, aspect of an <a>entity</a> such as a name, government ID, preferred payment processor, home address, or university degree.

Is there any distinction gained by keeping "entity credential" distinct from "credential"? Seems to me the main distinction might be if the subject is an entity or not. I was surprised to see that a "concept" could be an entity, which, FWIW, is out of sync with the ISO definition:

> The term entity should be taken in a broad sense. It represents a physical person, a moral or legal person (institution, company), an object (information, a system, a device), or a group of these individual entities. 

I think it is worth being clear that claims can be made about claims, but it seems odd to make them about general concepts, like "math" or "magenta".

msporny:
>>entity profile
>>
>>A set of information that can be used to identify a particular entity. An entity may have multiple entity profiles associated with it.
>
>The information may not be used to identify, i.e., correlate, any particular entity. Such information may be used to provide enhanced services. One might use other information for correlation/assurance that a given identifier refers to the subject in question, then use statements made referring to that identifier to customize interactions, e.g., localizing a search. The profile then is less about identifying the individual than ascribing attributes to them.
>
>I would simplify it to say an entity profile is a "collection of claims related to the same subject. An entity may have multiple profiles and each profile may contain claims from multiple different issuers." We could add "Claims may be used to correlate subjects across services and sessions, or they may be used to tailor services for that particular subject."

Simplified using your language, mostly: https://github.com/w3c/vc-data-model/commit/f4cd84590a7587faa8e9e85e099ef13286b01789

I decided to not add extra language that we didn't need yet (your last sentence)

>There is a subtle distinction about whether or not the profile is the set of claims that is presented or if it could/should also refer to any collection of claims about a common subject, whether presented or not--and even whether or not the profile is ever in a single location. The latter addresses the fluidity of distributed credentials in any number of stores; the former focuses the conversation on what an inspector receives (which makes it easier to define a concrete data model).

The profile is rarely ever in a single location. An entity has an almost unbound set of profiles about them (these profiles are dynamically created based on the query by the inspector and on the set of information that the holder wants to share). I'm not quite sure if that information changes the definition, but I don't think it does for now.

msporny:
> I think the two credential definitions could be reduced to one.

I didn't realize that there were two of them in there. I deleted the one and left 'entity credential': https://github.com/w3c/vc-data-model/commit/94d87027c0be1874329482a48f29a97243cd8598

jandrieu:
I'm not sure why we have this definition of identity in here at all. It seems redundant with entity profile. 

> identity
 A set of information that can be used to identify a particular
 entity such as a person, organization, concept, or device. An
 entity may have multiple identities associated with it.

If I recall, "identity profile", which was replaced with "entity profile" in this merge, was itself an attempt to remove "identity" as a distinct definition because it was a lightning rod for disagreement.

msporny:
> If I recall, "identity profile", which was replaced with "entity profile" in this merge, was itself an attempt to remove "identity" as a distinct definition because it was a lightning rod for disagreement.

That "identity" terminology is super old... dating back more than 2 years ago.

We don't pull the 'identity' terminology into the spec right now, but I sensed some push back on removing it, so I left it in the terminology document.

We /may/ want to say "hey, I know many of you are thinking about the word 'identity' and this is how we define it... but you know, peace, love, and respect, so think what you want about it..."... only in a less confusing manner.

In other words, we may want to draw attention to "yes, we've talked about identity... and here's the definition we came up with, but we're not claiming it's the ONLY definition and we endeavor to not use it because this spec is about stuff that's more low level than identity".

msporny:
> I think it is worth being clear that claims can be made about claims, but it seems odd to make them about general concepts, like "math" or "magenta".

Here are some possible verifiable claims about concepts and where the concept is the subject:

I hate math.
Math is the favorite subject of Martians.
My eyes are hurt by the color magenta.
Magenta has a hex value of D9017A.

jandrieu:
It seems that claims and credentials are now used almost interchangeably, losing some of the clarity that of verifiable claims. Now credentials get validated and verified... but not claims, whereas I think the burden is the other way around. 

We've also lost the distinction between claims and verifiable claims. The HTML reads to a human as "claims", but is linked to from both "verifiable claims" and "claims"
``` 
<dt><dfn data-lt="verifiable claim|claims">claim</dfn></dt>
```
> A statement made by an <a>entity</a> about a <a>subject</a>.

That's a fine definition of a claim, but we are missing the definition for a verifiable claim, e.g.,
> A claim whose authorship can be cryptographically verified

jandrieu:
> Here are some possible verifiable claims about concepts and where the concept is the subject

I know there are plenty of ways you can make a rigorous statement about a concept. I'm trying to tease out why anyone would issue a credential about "math".

This may be a flaw in our notion of credentials--which currently have the same subject. So you couldn't have a credential with the following claims:
 - Joe has a degree
 - That degree is a bachelor of science
 - That degree is in mathematics
 - Mathematics is the favorite subject of Martians.

Those claims would have three different subjects.

So, I'm guessing our definition of credential should be something like
> A set of one or more claims made by the same <a>entity</a>, usually related to a common root subject.

jandrieu:
> In other words, we may want to draw attention to "yes, we've talked about identity... and here's the definition we came up with, but we're not claiming it's the ONLY definition and we endeavor to not use it because this spec is about stuff that's more low level than identity".

But that's not the definition we came up with. We did a poll shortly after the ID2020/RWoT folks (like me) joined the conversation and we settled on "identity profile".  Because "identity" is not solely the information that can be used to identify someone. 

I'd rather not reintroduce the old term and kick that hornets nest. I don't think there is anything other than "identity provider" that depends on "identity" in the terms. I would also suggest that definition should be cleaned up to use entity profiles:

> Identity provider
 A software service that manages one or more <a>entity profiles</a> and their
 associated <a>credentials</a> on behalf of an <a>entity</a>. It typically
 handles requests to store <a>credentials</a> issued by an
 <a>issuer</a> and to retrieve <a>credentials</a> when requested by a
 <a>credential inspector</a>.

msporny:
> That's a fine definition of a claim, but we are missing the definition for a verifiable claim, e.g.,

Extended the terminology entry for 'claim' to also define 'verifiable claim': https://github.com/w3c/vc-data-model/commit/4ed42af4398e4580784931409262d746880d9bae

msporny:
> This may be a flaw in our notion of credentials--which currently have the same subject. So you couldn't have a credential with the following claims:
> 
> Joe has a degree
> That degree is a bachelor of science
> That degree is in mathematics
> Mathematics is the favorite subject of Martians.

/cc @dlongley @davidlehn ... need both of your input on this...

Technically, the data model supports a single credential making claims about multiple subjects:

```
{
  "@context": ["http://schema.org/", "https://w3id.org/identity/v1"],
  "type": ["Credential", "MultiSubjectCredential"],
  "issuer": "https://multi.example.org",
  "issued": "2017-06-05",
  "claim": [{
    "id": "did:v1:9f52df2e-7c96-42af-8dfd-1099980f8467",
    "ageOver": 18
  }, {
    "id": "did:v1:7d866a9f-c5a3-41f5-8ae1-0297b7849801",
    "ageOver": 21    
  }, {
    "id": "did:v1:c4e13c30-31dc-40df-867f-ed678d87ac54",
    "ageOver": 65   
  }]
}
```

While this may be a corner case, the data model allows for it. This means that we may have to change the definition of entity credential and entity profile from being about "a subject" to "usually about the same subject" or equivalent language.

dlongley:
> While this may be a corner case, the data model allows for it. This means that we may have to change the definition of entity credential and entity profile from being about "a subject" to "usually about the same subject" or equivalent language.

Yes, I suppose a credential could be about multiple, unconnected subjects. The key word being unconnected in this case -- as many credentials have already been used in practice that include connected subjects, such as one that would fulfill @jandrieu's example:

> Joe has a degree
> That degree is a bachelor of science
> That degree is in mathematics
> Mathematics is the favorite subject of Martians.

Which may look something like:

```json
{
  "@context": ["http://schema.org/", "https://w3id.org/identity/v1"],
  "type": ["Credential", "DegreeCredential"],
  "issuer": "https://multi.example.org",
  "issued": "2017-06-05",
  "claim": {
    "id": "did:v1:9f52df2e-7c96-42af-8dfd-1099980f8467",
    "degree": {
      "type": "Degree",
      "degreeType": "BachelorOfScience",
      "major": {
        "id": "Mathematics",
        "favoredBy": "Martians"
      }
    }
  },
  "signature": {
    "type": "LinkedDataSignature2015",
    "created": "2017-06-05T07:08:36Z",
    "signatureValue": "lQ7VZUeAA...Z5Lk="
  }
}

**Comment by jandrieu** [comment]
I think the root of my confusion was conflating our use of subject and statement with similar terms in RDF.

Reading the definition of a credential, we imply a single subject:
> A set of one or more claims made by the same <a>entity</a> about a
  <a>subject</a>.

Since the only use of subject is in the claim definition, i read that "subject" is the subject in an RDF triple.

From your example:
```
{
  "@context": ["http://schema.org/", "https://w3id.org/identity/v1"],
  "type": ["Credential", "DegreeCredential"],
  "issuer": "https://multi.example.org",
  "issued": "2017-06-05",
  "claim": {
    "id": "did:v1:9f52df2e-7c96-42af-8dfd-1099980f8467",
    "degree": {
      "type": "Degree",
      "degreeType": "BachelorOfScience",
      "major": {
        "id": "Mathematics",
        "favoredBy": "Martians"
      }
    }
  },
  "signature": {
    "type": "LinkedDataSignature2015",
    "created": "2017-06-05T07:08:36Z",
    "signatureValue": "lQ7VZUeAA...Z5Lk="
  }
}
```
I believe the claim part of this maps to the following turtle (ignoring namespaces):
```
	_:a	a	claim ;
		id	did:v1:9f52df2e-7c96-42af-8dfd-1099980f8467 ;
		degree  _:b .
	_:b	type	"Degree" ;
		degreeType BachelorOfScience ;
		major	_:c .
	_:c	id	"Mathematics" ;
		favoredBy	"Martians .
```
I look at that RDF and see three subjects in eight statements. That's how I was interpreting subjects and statements in verifiable claims, but that clearly is inconsistent with how everyone else is using those terms.

We define subject as 
> An <a>entity</a> which may have multiple <a>entity profiles</a> and
 about which <a>claims</a> may be made.

This definition was compatible with "subject" in an RDF triple so I didn't realize the distinction. Also, we do not define "statement" explicitly. So, when we then say that a claim is a statement, I thought that as like saying a claim is an RDF triple. Looking more closely at the examples, it's clear that doesn't work. "Statement" for us is not the same as an RDF statement and neither is "subject".

So I think we may not need a change in the definition per se, but clarification that a single statement as we mean it may map to multiple RDF subjects in multiple RDF triples. Unfortunately, I'm struggling to say that without relying on RDF concepts--which seems like the right idea. It almost feels like a game show where we need to get our teammates to map semantic web concepts to our data model without saying "RDF" terminology.

If we can help people avoid thinking about subjects in their RDF sense of subject/predicate/object triples, then it makes sense to have credentials be about a single subject (even as its data model maps to multiple RDF subjects in multiple triples). I'm not sure how we clarify that, however.

burnburn:
I am fine with the identity->entity changes in index.html.  I believe the terms changes are okay but may not have followed all of them.

burnburn:
The discussion about 'subject' doesn't belong in this PR and should be a new issue for dicussion.

msporny:
* If we keep identity as an entry, different definition.
* Do not do anything w/ subject/predicate/object.
* Put issue marker about multiple subjects in claims.

jandrieu:
For identity, how about:
identity -- the means for keeping track of entities across contexts. Digital identities enable tracking and customization of entity interactions across digital contexts, typically using identifiers and attributes.

My primary issue with treating identity solely as the information in a digital system is that it sets up blinders regarding privacy issues where real world identity is compromised in unanticipated ways because only the digital identity was considered when evaluating technical trade offs.

stonematt:
Should/Could we add a comment to that affect:  "... typical using identifiers and attributes, a subset of which may require a oversight (?) to respect and control access to maintain privacy"

(not great language, but do you see where I'm going?)

jandrieu:
@stonematt I like a line about privacy, but I wouldn't limit it to a subset.  The idea that only a subset of identity information needs oversight is the core flaw behind using PII as a regulatory framework.

How about adding: 
> Unintended distribution or use of identity information can compromise privacy. Collection and use of such information should follow the principles of minimal disclosure.

PR #53, opened 2017-06-06
Update head-matter to be consistent with vc-use-cases.

gkellogg:


index.html
```diff
@@ -13,10 +13,10 @@
     <script type="text/javascript" class="remove">
       var respecConfig = {
         // specification status (e.g. WD, LCWD, NOTE, etc.). If in doubt use ED.
-        specStatus: "CG-FINAL",
+        specStatus: "FPWD",
 
         // the specification's short name, as in http://www.w3.org/TR/short-name/
-        shortName: "claims-data-model",
+        shortName: "verifiable-claims-data-model",
```
msporny: We need to make sure this is vc-use-cases and vc-data-model - I thought those were the names we picked for short names?

index.html
```diff
@@ -64,31 +64,21 @@
...
-        otherLinks: [{
-          key: "Version Control",
-          data: [{
-            value: "Github Repository",
-            href: "https://github.com/opencreds/vc-data-model"
-          }, {
-            value: "Issue Tracker",
-            href: "https://github.com/opencreds/vc-data-model/issues"
-          }]
-        }],
```
msporny: Are these links auto-generated because you set 'github' above? If so, +1... if not, we should make sure folks can still get to the repo and issue tracker via links at the top of the spec. I'm assuming it's the former.

index.html
gkellogg: What I inherited for use cases was "verifiable-claims-use-cases", which is what I believe @liamquin applied for. While I can see that this may not be "short", there is some reason to want to use the same prefix for both documents. Not too late to change use cases, but probably leads to a short delay.

index.html
gkellogg: Yes, specifying "github" now generates all that; check out the [use-cases document](https://w3c.github.io/vc-use-cases/) to see. Makes sense to finally do this and intuit these values.

index.html
gkellogg: So, this depends on what @liamquin actually applied for, but I can see from the [22 May Minutes]() that it did, indeed, seem to be "vc-use-cases", at least that seems to be implied. I'll update both this PR, and the existing PR for use cases.

gkellogg:
@msporny, that should do it for your comments, I believe.

liamquin:
On Tue, 2017-06-06 at 15:01 -0700, Gregg Kellogg wrote:
> 
> What I inherited for use cases was "verifiable-claims-use-cases",
> which is what I believe @liamquin applied for. 

Yes (after brief WG discussion). I've sent a publication request using
that too, but, there's still time to change it - we could ask for
approval for something else and publish next week instead.

On the other hand "verifiable-claims" in the URL is good SEO :-) and
will help visibility. The "short name" doesn't really mean it's short,
it's just the prefix used as a filename and as a URL component.

While you're editing the status, note that we now have a
public-vc-comments mailing list for comments, as per today's WG call.

gkellogg:
I updated the mailing list to -comments in a PR, along with the short-name. Let's be consistent on the use of short names, either "vc-" or "verifiable-claims-". If we can settle this quickly and use "verifiable-claims-", I can back out the change on use-cases and we should be good to publish tomorrow.

msporny:
> @msporny, that should do it for your comments, I believe.

Thanks @gkellogg - I leave it to @liamquin, @burnburn, @stonematt and RichardVarn to figure out what shortname the group decided on and whether or not they want to delay publication if it needs to be changed. Once we pick a short name, it's very difficult to go back.

burnburn:
@gkellogg We are using verifiable-claims as the shortname.  Can you please update this PR?  Then it is ready to merge.

msporny:
The last couple of PRs created a TON of merge conflicts, so I'm going to have to apply this patch by hand...

msporny:
I applied this patch manually here: https://github.com/w3c/vc-data-model/commit/4f99d53f8c1386e88d4affb93f44b42b089d85e2

Issue #59, opened 2017-07-16
Challenges in Understanding Spectrum of Integrity/Inspection/Validation/Verification/Confirmation/Revocation


kimdhamilton:
_From @kimdhamilton on July 15, 2017 23:8_

_From @ChristopherA on July 12, 2017 0:7_

(this is a pre-draft of a post to be an issue on https://github.com/w3c/vc-data-model/issues/ or possibly added to https://github.com/w3c/vc-data-model/issues/58 once it is more thought out)

As evidenced by the failure of polling in the W3C Verifiable Claims WG last week on the name of the role of Validator vs. Inspector, and the lively discussion in the WG meeting today https://www.w3.org/2017/07/11-vcwg-minutes.html, we lack a really good model for describing the multiple actions that happen in our verifiable claim architecture, in particular when blockchain-based DIDs are being used. 

In addition, our whole industry has been terribly bad at the practical considerations as to the issues of how revocation should be designed. I personally have experienced this with SSL/TLS, which for almost a decade and a half had only lip-service support of revocation, and even now is still being challenged to deploy more advanced solutions

So I want to walk through DID:BTCR from the vantage point of a number of steps that fall into the Integrity/Inspection/Validation/Verification/Confirmation/Revocation spectrum of roles.

First, please forgive in advance the specific words I'm using below — they are used more to signify the different placeholders as opposed to a well thought out proposal as which words should be used.

----

In our [#RebootingWebOfTrust User Story](https://github.com/WebOfTrustInfo/btcr-hackathon/blob/master/RWOT-User-Story.md) **Alice**, our pseudo-anonymous programmer. daughter of immigrants, has heard that **Bob**, a refugee advocate has a need a mobile phone app. Fearing that her own extended family might be harmed if she is revealed to be helping **Bob**, she wishes to introduce herself and present credentials that she is qualified for the work, but remain pseudo-anonymous.

So first she establishes a DID and self-signed DDO. She has a professional colleague and friend **Donna** with a public persona (i.e. not anonymous) who indirectly knows **Bob** through yet another colleague (i.e. **Bob** & **Donna** share a trust network but connected by multiple-degrees of separation).

**Donna** issues a Verifiable Claim that she "knows" **Alice** and she is willing to attest to her competence in mobile development, which **Donna** gives a signed copy of back to **Alice**. Alice counter-signs this claim and adds it to her DDO (this is something unique to the very self-sovereign BTCR method, and may not apply to other methods).

[...]

PR #65, opened 2017-07-27
Restructure spec to describe basic concepts and verification

msporny:
Per my action item from last week's call, I've restructured the spec to contain "basic concepts", "advanced concepts" and "verification" sections. Laying the specification out in this way enables us to neatly answer a number of the questions that @ChristopherA raised in #59.

This restructuring has also led to the ability to simplify the core data model section, which will be done in a separate PR. For the time being, ignore the Core Data Model section, and focus on the "Basic Concepts", "Advanced Concepts", and "Verification" sections in the spec. Don't worry about typos/prose for now, this PR is just to test out the new layout for the spec to see if it resonates with the group.

You can preview this PR here:

https://htmlpreview.github.io/?https://github.com/w3c/vc-data-model/blob/msporny-verification/index.html

/cc @ChristopherA @kimdhamilton

### DIFF
--- a/index.html
+++ b/index.html
@@ -85,6 +85,12 @@
         inlineCSS: true
       };
     </script>
+    <style>
+pre .highlight {
+  font-weight: bold;
+  color: green;
+}
+</style>
   </head>
   <body>
     <section id='abstract'>
@@ -368,7 +374,7 @@ make grasping the basic concepts more difficult.
     </section>
 
     <section>
-      <h2>Data Model</h2>
+      <h2>Core Data Model</h2>
       <p>This section describes a data model for <a>entity
       profiles</a> and <a>entity credentials</a>, the latter covering
       both claims and verifiable claims, that is compatible with the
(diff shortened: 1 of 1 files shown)


dlongley:
Manu and I discussed this before he reworked it, so +1 to this approach or something very similar.

The basic idea is to introduce the core data model parts first (essentially prerequisites for understanding anything at all). Then introduce individual elements of the data model that are related to different aspects of using verifiable claims (where each action/usage is a subsection). In each of these subsections we should mention at the end of the section that the data model terms introduced therein are related to verification in such and such ways with a link off to the verification section. Then, the verification section goes through, comprehensively, all of the various things that can be verified.

ChristopherA:
+1

I like the direction this is going. I may quibble with some words and other minor items, like what to do in situations with multiple signature and/or timestamps, and there may be some revocation family of ideas that may have to be rethought, but these changes will move us in the direction of being able to address those issues.

stonematt:
i like this direction also - thanks manu.

burnburn:
Yes, this is a definite improvement.

kimdhamilton:
LGTM

msporny:
Thanks all, seeing no objections, merging now.

Issue #71, opened 2017-09-25
Perform security and privacy assessment


dlongley:
We need to produce a security and privacy assessment document by answering these questions:

https://www.w3.org/TR/security-privacy-questionnaire/

We aren't doing an browser APIs so it may not be immediately clear how to/if some of these questions can be answered.

dlongley:
An example from the Web Payments Working Group:

https://docs.google.com/document/d/1w7ginyzNg-xZUmITK4vzcGUKB4gbMOAvlkWWaRtX14k

dlongley:
cc: @David-Chadwick -- want to make sure you saw the example above.

David-Chadwick:
Thanks Dave

dlongley:
@David-Chadwick, thanks for the privacy assessment here:

https://lists.w3.org/Archives/Public/public-vc-wg/2017Oct/0003.html

We should get this into github in HTML format so others can comment or add with PRs.

David-Chadwick:
<p>3 Questions to Consider</p>

<p>3.1 Does this specification deal with personally-identifiable information? </p>

<p>Yes. The properties inside VCs are PII</p>

<p></p>

<p>Recommendation. None as the specification
makes this clear.</p>

<p></p>

<p>3.2 Does this specification deal with
high-value data?</p>

<p>Potentially yes. The specification places
no restrictions on what data can be placed in a VC, but it could be highly
sensitive personal data.</p>

<p></p>

<p>Recommendation. Protocols that carry VCs should always encrypt
the VC payload. In addition, VCs that contain highly sensitive personal data
should have that data encrypted inside the VC, so that any entity that captures
the VC is not able to see the sensitive information.</p>

<p></p>

<p>3.3 Does this specification introduce new
state for an origin that persists across browsing sessions?</p>

<p>Not of itself no, as the specification only
specifies a data structure. However, protocols that use this specification may
introduce new state.</p>

<p></p>

<p>Recommendation. None</p>

<p></p>

<p>3.4 Does this specification expose
persistent, cross-origin state to the web?</p>

<p>Yes. VCs contain a random though potentially
persistent identifier of the subject. This is passed between the issuer and the
inspector. Consequently collusion between them could identify the subject to the
inspector, even though the VC itself does not. This is because in many cases
the issuer will know the complete identity of the subject, even if the VC only
contains a small proportion of it (such as age).</p>

<p></p>

<p>Recommendation. The subject should limit
[...]

PR #73, opened 2017-10-16
Add David Chadwick's privacy analysis document

msporny:
Convert @David-Chadwick's privacy analysis to ReSpec format.

### DIFF
--- a/privacy.html
+++ b/privacy.html
@@ -123,10 +123,10 @@ or send them to
       <section>
         <h2>Does this specification deal with personally-identifiable information?</h2>
         <p>
-Yes. The properties inside Verifiable Claims are PII
+Yes. The properties inside Verifiable Claims are Personally Identifiable Information (PII).
         </p>
         <p>
-Recommendation. None as the specification makes this clear.
+Recommendation: None as the specification makes this clear.
         </p>
       </section>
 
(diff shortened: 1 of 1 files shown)


dlongley:
Addresses #71.

retog:
I'm wondering if JSON-LD's @context loading mechanism should be taken into account as this may leak some data to the server serving the context.

msporny:
@retog good point... can you raise an issue on that, please? Something along the lines of "Privacy implications of loading JSON-LD context". The first attack that I can think of is adding a "tracking context" that is unique to each credential issued, so you know which site is loading the context. Holder software (digital wallets) could warn the holder that they are using a verifiable credential that may be tracking them by detecting a non-standard context that is being used (in that, the software has never seen that particular context among the MANY holders on the system.

ChristopherA:
@manu I wonder if overuse of the signature nonce could be misused as a tracking context. Certainly something that should be added to the questions for a security review on spec.

msporny:
@ChristopherA Yes, the signature value itself can be used as a tracking context, even without a nonce. We'd need some sort of ZKP mechanism to escape that... and even then, if there are identifiers in the data that are long-lived, we fall into the same trap. Please raise another issue on this point.

dlongley:
+1 to merge -- good as a first cut.

PR #74, opened 2017-10-16
Add SAML/OIDC comparison by David Chadwick

msporny:
Add a comparison between SAML and OpenID Connect done by @David-Chadwick.

### DIFF
--- a/comparison.html
+++ b/comparison.html
@@ -1,0 +1,229 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Verificable Credentials, SAML, and OpenID Connect</title>
+    <meta http-equiv='Content-Type' content='text/html;charset=utf-8'/>
+    <!--
+      === NOTA BENE ===
+      For the three scripts below, if your spec resides on dev.w3 you can check them
+      out in the same tree and use relative links so that they'll work offline,
+     -->
+    <script src='https://www.w3.org/Tools/respec/respec-w3c-common' class='remove'></script>
+    <script src="./common.js" class="remove"></script>
+    <script type="text/javascript" class="remove">
+      var respecConfig = {
+        // specification status (e.g. WD, LCWD, NOTE, etc.). If in doubt use ED.
+        specStatus: "ED",
+
+        // the specification's short name, as in http://www.w3.org/TR/short-name/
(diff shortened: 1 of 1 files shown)


comparison.html
```diff
@@ -0,0 +1,229 @@
...
+      <h1>Comparison of Verificable Credentials with SAML and OIDC</h1>
+
+      <p>
+Existing federated identity management (FIM) systems suffer from a
+number of problems, in particular with regard to their trust model,
+economic models and the privacy of their users.
+      </p>
+
+      <p>
+The FIM trust model is wrong as it requires the Identity Provider (IdP)
```
dlongley: Let's strike "is wrong as it". It's a subjective value judgment that detracts from the point.

David-Chadwick: Accepted

comparison.html
```diff
@@ -0,0 +1,229 @@
...
+directly on <a href="https://github.com/w3c/vc-data-model/issues/">GitHub</a>,
+or send them to
+<a href="mailto:public-vc-comments@w3.org">public-vc-comments@w3.org</a>
+(<a href="mailto:public-vc-comments-request@w3.org?subject=subscribe">subscribe</a>,
+<a href="https://lists.w3.org/Archives/Public/public-vc-comments/">archives</a>).
+      </p>
+    </section>
+
+    <section>
+      <h1>Comparison of Verificable Credentials with SAML and OIDC</h1>
```
kimdhamilton: small typo: "Verificable" -> "Verifiable"

kimdhamilton: small typo: "Verificable" -> "Verifiable"

comparison.html
msporny: Fixed.

comparison.html
msporny: Fixed.

burnburn:
As a novice to SAML and OIDC I would have no idea that this was a comparison since they are referenced but not mentioned by name.  Rather, this reads to me like one of the most compact explanations of why we are doing the VC work that I have ever read.  I was surprised to find it a separate doc and not part of our spec.

The text itself is wonderful and I am willing to accept it as-is to get something in place, but I do have the two concerns above.

PR #77, opened 2017-10-22
Replace inspector-verifier terminology with Verifier.

### LINKED ISSUE #64: "Inspector-Verifier" terminology is confusing
I'm concerned about the recent decision to pick "inspector-verifier" as a glossary term for the following reasons:

* From a communication and marketing standpoint, it's awkward to say and use alongside the other terminology (issuer, holder, etc.). Having used the term to talk about the ecosystem, it sounds out of place when said out loud on a stage or when talking with customers.
* It makes it seem like there are two roles instead of one.
* It sounds like the Chairs mis-read the IRV poll results, which showed Verifier as the clearly preferred choice.

I suggest we select "Verifier" as the winner because:

* That is aligned with the results of the poll.
* It doesn't lead to confusion over number of roles.
* Its easy for readers/listeners to understand the concept based on the word.

msporny:
This PR replaces the inspector-verifier terminology with "Verifier". I tried this language out at RWoT 5 and people seemed to get it in a way that they then used the terminology correctly (not in a confused manner). I suggest we resolve this and move on, as presenting as "inspector-verifier" is creating issues (people keep asking me why the term is hyphenated... is there another type of verifier? Is there another type of inspector?). I've also noticed a few other companies, other than ours, using "verifier" in their marketing literature.

The PR also replaces "jobTitle -> Student" in the diagram with "jobTitle -> Professor", as a few people have commented that being a student is, arguably, not a job.

Closes #64.


jandrieu:
Inspector-verifier is definitely a mouthful

index.html
```diff
@@ -878,7 +878,7 @@
         <h3>Expiration</h3>
         <ul>
           <li>The <code>issued</code> date must be in the expected range.
-          For example, an inspector-verifier may wish to ensure that the recorded
+          For example, an verifier may wish to ensure that the recorded
```
dlongley: `an verifier` => `a verifier`

index.html
```diff
@@ -2051,7 +2048,7 @@
 each containing one of the following properties:
 "Staff Member", "Post Graduate Student", "Department of Computing" and
 "Department of Economics". The holder could then transfer
-the "Staff Member" and "Department of Economics" to an inspector-verifier,
+the "Staff Member" and "Department of Economics" to an verifier,
```
dlongley: `an verifier` => `a verifier`

index.html

index.html

msporny:
There are three positives, zero objections. Merging after the next telecon on 2017-10-31.

PR #78, opened 2017-10-22
Add "bearer credential" sub-section to privacy analysis section.

### LINKED ISSUE #8: Correlatability of bearer claims
We should make it clear that even bearer claims can be correlated. For example, claims that contain personally identifiable information or unique identifiers can be correlated. Bearer claims can be tracked based on usage patterns.

msporny:
Add a section on the privacy implications of the use of bearer claims.

Closes #8.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1813,7 +1813,7 @@ attribute that appears on a driver's license in addition to individual
 claims (a singular claim containing the person's birthday), and individual
 claims that are more abstract (a singular claim containing an
 <code>ageOver</code> attribute). In addition, the issuer is encouraged to
-provide secure HTTP endpoints for retrieving single-use bearer claims to
+provide secure HTTP endpoints for retrieving single-use bearer credentials to
 promote the pseudonymous usage of claims when it is safe for the issuer to
 issue such claims.
         </p>
@@ -1828,18 +1828,18 @@ that is required for the particular transaction.
       </section>
 
       <section>
-        <h3>Bearer Claims</h3>
+        <h3>Bearer Credentials</h3>
 
         <p>
-A <em>bearer claim</em> is a privacy enhancing piece of information,
+A <em>bearer credential</em> is a privacy enhancing piece of information,
 such as a concert ticket, that entitles the <a>holder</a> of the claim to a
 particular resource without divulging sensitive information about the holder.
         </p>
 
         <p>
-Verifiable Claims that are bearer claims are possible by not specifying the
-<a>subject</a> identifier in the <code>claim</code> property. For example, the
-following Verifiable Credential is a bearer claim:
+Verifiable Credentials that are bearer credentials are possible by not
+specifying the <a>subject</a> identifier in the <code>claim</code> property.
+For example, the following Verifiable Credential is a bearer credential:
         </p>
 
         <pre class="example nohighlight" title="Usage of issuer properties">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1830,9 +1830,60 @@
...
+        <p>
+<a>Holders</a> SHOULD be warned by their software if bearer claims containing
+sensitive information are issued or requested, or if there is a correlation
+risk when combining two separate bearer claims across one or more sessions.
+While it may be impossible to detect all correlation risks, there are a few
+simple ones that can be detected.
+        </p>
+
+        <p>
+<a>Verifiers</a> SHOULD not request bearer claims that can be used to correlate
```
dlongley: Perhaps "unduly correlate" is better than just "correlate". It's about information minimization.

David-Chadwick: This section would be more correct if it talked about bearer credentials rather than bearer claims. A claim that is not signed is not of much value.

msporny: Done, both suggestions applied to text.

David-Chadwick:
much better now

index.html
```diff
@@ -1828,11 +1828,62 @@
...
       <section>
-        <h3>Bearer Claims</h3>
+        <h3>Bearer Credentials</h3>
 
-        <p class="issue">
-Bearer claims containing PII or unique identifiers can be correlated.
-Bearer claims can be tracked based on usage patterns.
+        <p>
+A <em>bearer credential</em> is a privacy enhancing piece of information,
+such as a concert ticket, that entitles the <a>holder</a> of the claim to a
```
burnburn: claim -> credential

index.html
```diff
@@ -1828,11 +1828,62 @@
...
+        </p>
+
+        <pre class="example nohighlight" title="Usage of issuer properties">
+{
+  "id": "http://dmv.example.gov/credentials/temporary/28934792387492384",
+  "type": ["Credential", "ProofOfAgeCredential"],
+  "issuer": "https://dmv.example.gov/issuers/14",
+  "issued": "2017-10-22T12:23:48Z",
+  "claim": {
+    <span class="comment">// note that the 'id' property is not specified for bearer credentials</span>
```
burnburn: id property or bearer property as described above?  Needs a small bit of clarification.

index.html
```diff
@@ -1828,11 +1828,62 @@
...
+<a>Issuers</a> of bearer credentials SHOULD ensure that bearer credentials that
+are expected to provide privacy enhancing benefits 1) are single use, when
+possible, 2) do not contain personally identifying information, and 3) are not
+unduly correlatable.
+        </p>
+
+        <p>
+<a>Holders</a> SHOULD be warned by their software if bearer credentials
+containing sensitive information are issued or requested, or if there is a
+correlation risk when combining two separate bearer credentials across one
```
burnburn: two -> two or more

index.html
```diff
@@ -1828,11 +1828,62 @@
...
+or more sessions. While it may be impossible to detect all correlation risks,
+there are a few simple ones that can be detected.
```
burnburn: Either this should say "some may be detectable" or you should provide the list since it implies that a specific few (and thus nameable) ones can be detected.

index.html
```diff
@@ -1828,11 +1828,62 @@
...
+        </p>
+
+        <p>
+<a>Verifiers</a> SHOULD not request bearer credentials that can be used to
```
burnburn: SHOULD not -> SHOULD NOT

index.html

index.html
msporny: I added text above the example to clarify what the id property does.

index.html

index.html
msporny: Fixed, went with the earlier text.

index.html

PR #83, opened 2017-10-23
Add section on frequency of claim issuance.

### LINKED ISSUE #14: Correlation via frequency of claim issuance
The rate at which an issuer issues claims may be a privacy violation because the holder of those claims may use them in a particular pattern that exposes what the claims are being used for. For example, short-lived over-21 claims being used to correlate that someone went to a particular bar.

msporny:
Closes #14.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2001,9 +2001,29 @@ Granting of rights to a service via cryptographic mechanisms.
       <section>
         <h3>Frequency of Claim Issuance</h3>
 
-        <p class="issue">
-The rate at which an issuer issues claims may be a privacy violation.
+        <p>
+As Section <a href="#usage-patterns"></a> details, usage patterns can be
+correlated into certain types of behavior. Part of this correlation is
+mitigated when a <a>holder</a> uses a verifiable credential without the
+knowledge of the <a>issuer</a>. Issuers may defeat this protection by making
+their claims short lived and renewal automatic.
         </p>
+
+        <p>
+For example, an "over the age of 21" credential may be useful when gaining
+access to a bar. If an issuer makes the credential have a very short expiration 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2001,9 +2001,29 @@
       <section>
         <h3>Frequency of Claim Issuance</h3>
 
-        <p class="issue">
-The rate at which an issuer issues claims may be a privacy violation.
+        <p>
+As Section <a href="#usage-patterns"></a> details, usage patterns can be
```
dlongley: Is the use of an empty link here correct? Does ReSpec do the right thing? Is `Section` meant to be capitalized?

index.html
```diff
@@ -2001,9 +2001,29 @@
...
+mitigated when a <a>holder</a> uses a verifiable credential without the
+knowledge of the <a>issuer</a>. Issuers may defeat this protection by making
+their claims short lived and renewal automatic.
+        </p>
+
+        <p>
+For example, an "over the age of 21" credential may be useful when gaining
+access to a bar. If an issuer makes the credential have a very short expiration 
+date and an automatic renewal mechanism, then they could possibly correlate
+the holder's alcohol consumption behavior in a way that negatively impacts the
```
dlongley: Change "correlate" to "track".

Strike `alcohol consumption` and just have it say "holder's behavior".

index.html
msporny: Yes, yes, and yes.

index.html

burnburn:
Is it claims that are issued, or credentials?

msporny:
It's credentials, but we haven't made that decision in the group yet. So, claims for now, and I'll blanket update the language once we have consensus in the group.

PR #86, opened 2017-10-24
Add security section on cryptography suites and libraries.

msporny:


index.html
```diff
@@ -2026,6 +2026,34 @@
...
+systems are properly designed. Proper
+<a href="https://en.wikipedia.org/wiki/Red_team">red teaming</a>
+is also suggested to remove bias from security reviews.
+        </p>
+
+        <p>
+Cryptography suites and libraries have a shelf life and eventually fall to
+new attacks and technology advances. Any production quality system must take
+this reality into account and ensure that mechanisms exist to easily upgrade
+old or broken cryptographic suites and libraries in a proactive manner.
```
burnburn: should also mention needing process for revoking or updating credentials themselves as well in this case.

PR #89, opened 2017-10-24
Add security section on dynamic information.

### LINKED ISSUE #18: Highly dynamic information should have short expiry times
The specification should note that validity time periods should be shorter for highly dynamic information. We should also mention that shorter expiry periods may result in less privacy due to frequency of issuance/use.

msporny:
Fixes #18.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2065,9 +2065,18 @@ will be placed here.
       <section>
         <h3>Highly Dynamic Information</h3>
 
-        <p class="issue">
-Time periods should be shorter for highly dynamic information.
+        <p>
+When verifiable <a>credentials</a> are issued for highly dynamic information,
+implementers should ensure that the expiration times for the credential are
+set appropriately. Expiration periods that are longer than the timeframe
+where the credential is valid may create exploitable security vulnerabilities.
+Expiration periods that are shorter than the timeframe where the information
+expressed by the credential is valid create a burden on <a>holders</a> and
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2065,9 +2065,18 @@
...
+        <p>
+When verifiable <a>credentials</a> are issued for highly dynamic information,
+implementers should ensure that the expiration times for the credential are
+set appropriately. Expiration periods that are longer than the timeframe
+where the credential is valid may create exploitable security vulnerabilities.
+Expiration periods that are shorter than the timeframe where the information
+expressed by the credential is valid create a burden on <a>holders</a> and
+<a>verifiers</a>. It is therefore important to set validity periods for
+credentials that are appropriate to the use case and the expected lifetime
+for the information contained in the credential.
```
burnburn: Isn't this in direct contradiction to the recommendation to use short-lived credentials?

dlongley: Yes, perhaps we need more specific recommendations or a sliding scale -- or just a note that "if credentials are too short-lived the drawbacks are X and if they are too long-lived the drawbacks are Y." Followed by: "Use technology T or policy P for use cases where you can't compromise on either."

msporny: Can we raise a new issue for this? I agree with both of you, but I feel like we need to refactor/merge multiple sections (that don't exist in the spec if we don't pull in this PR) based on this discussion.

PR #91, opened 2017-10-24
Update signature-based correlation section with more bad news.

### LINKED ISSUE #87: Privacy implications of signature or nonce in correlation
Related to #75, the nonce or the signature value itself can be used as a tracking context. We'd need some sort of ZKP mechanism to escape that... and even then, if there are identifiers in the data that are long-lived, we fall into the same trap.

msporny:
Fixes #87.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1741,10 +1741,11 @@ one session or domain and the value does not change. Examples include the
         <p>
 If strong anti-correlation properties are desired,
 it is advised that signature values and metadata are regenerated
-each time using technologies like zero knowledge proofs or group signatures.
-It is also important to note that even when using anti-correlation signatures
-that information may still be contained in the credential that defeats the
-anti-correlation properties of the cryptography.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1730,16 +1730,21 @@
...
 
         <p>
 If strong anti-correlation properties are desired,
-it is strongly advised that signature values and metadata are regenerated
-each time using technologies like group signatures.
+it is advised that signature values and metadata are regenerated
+each time using technologies like zero knowledge proofs or group signatures.
+It is also important to note that even when using anti-correlation signatures
+that information may still be contained in the credential that defeats the
+anti-correlation properties of the cryptography.
```
dlongley: Another remedy would be a highly-available, commonly trusted, blinding third party that can generate "pairwise signatures". This would help bridge the gap for issuers that are not highly-available or that cannot provide certain crypto solutions for whatever reason.

PR #92, opened 2017-10-25
Remove WebIDL from spec text.

### LINKED ISSUE #4: Is the WebIDL expression necessary?
Do we need to specify the WebIDL expression of Verifiable Claims in the same document? Should it be in a separate document? It feels pretty heavy-weight in the current document.

msporny:
Closes #4.


msporny:
> burnburn weeps as all vestiges of his original text disappear from the spec.

The spirit of your text still lives on... :)

PR #94, opened 2017-11-08
Fix missing closing bracket in verifiable profile

peacekeeper:
In both the JSON and JSON-LD examples.
Also changing indentation of the "signature" block.


peacekeeper:

msporny:
@peacekeeper -- yep, you did the right think. Thank you for the fix!

PR #96, opened 2017-11-16
Initial commit of vocabulary files.

### LINKED ISSUE #33: RDFS / OWL for VC vocabulary?
Hello, I'm curious if there are plans to define an RDF schema or OWL ontology for the VC vocabulary. E.g. what would be the rdfs:domain and rdfs:range of the cred:claim property?

gkellogg:
This simply adds the vocabulary updates described in the F2F at TPAC. Ill file issues on issues raised at TPAC.

Fixes #33.

### DIFF
--- a/vocab/context.jsonld
+++ b/vocab/context.jsonld
@@ -1,0 +1,87 @@
+{
+  "@context": {
+    "id": "@id",
+    "type": "@type",
+    "cred": "https://w3id.org/credentials#",
+    "dc": "http://purl.org/dc/terms/",
+    "schema": "http://schema.org/",
+    "sec": "https://w3id.org/security#",
+    "xsd": "http://www.w3.org/2001/XMLSchema#",
+    "CryptographicKey": "sec:Key",
+    "GraphSignature2012": "sec:GraphSignature2012",
+    "Identity": "https://w3id.org/identity#Identity",
+    "LinkedDataSignature2015": "sec:LinkedDataSignature2015",
+    "created": {
+      "@id": "dc:created",
+      "@type": "xsd:dateTime"
+    },
+    "creator": {
+      "@id": "dc:creator",
+      "@type": "@id"
+    },
+    "domain": "sec:domain",
+    "expires": {
+      "@id": "sec:expiration",
+      "@type": "xsd:dateTime"
+    },
+    "nonce": "sec:nonce",
+    "normalizationAlgorithm": "sec:normalizationAlgorithm",
+    "owner": {
+      "@id": "sec:owner",
+      "@type": "@id"
+    },
+    "privateKey": {
+      "@id": "sec:privateKey",
+      "@type": "@id"
+    },
+    "privateKeyPem": "sec:privateKeyPem",
+    "publicKey": {
+      "@id": "sec:publicKey",
+      "@type": "@id"
+    },
+    "publicKeyPem": "sec:publicKeyPem",
+    "publicKeyService": {
+      "@id": "sec:publicKeyService",
+      "@type": "@id"
+    },
+    "revoked": {
+      "@id": "sec:revoked",
+      "@type": "xsd:dateTime"
+    },
+    "signature": "sec:signature",
+    "signatureAlgorithm": "sec:signatureAlgorithm",
+    "signatureValue": "sec:signatureValue",
+    "Credential": "cred:Credential",
+    "CryptographicKeyCredential": "cred:CryptographicKeyCredential",
+    "Profile": "cred:Profile",
+    "RevocationList2017": "cred:RevocationList2017",
+    "claim": {
+      "@id": "cred:claim",
+      "@container": "@graph"
+    },
+    "credential": {
+      "@id": "cred:credential",
+      "@type": "@id"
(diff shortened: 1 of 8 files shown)


vocab/credentials.jsonld
```diff
@@ -0,0 +1,251 @@
+{
```
msporny: Here's the latest context from the test suite: https://github.com/w3c/vc-test-suite/blob/gh-pages/contexts/vc-v1.jsonld

vocab/credentials.jsonld
```diff
@@ -0,0 +1,251 @@
...
+    },
+    "revoked": {
+      "@id": "sec:revoked",
+      "@type": "xsd:dateTime"
+    },
+    "signature": "sec:signature",
+    "signatureAlgorithm": "sec:signatureAlgorithm",
+    "signatureValue": "sec:signatureValue",
+    "Credential": "cred:Credential",
+    "claim": {
```
msporny: This needs to be updated in the test suite and here - "@container": "@graph", I think. @dlongley, I forget the details, but it looks like it's wrong in both the test suite and here.

vocab/credentials.jsonld
```diff
@@ -0,0 +1,251 @@
...
+    },
+    "recipient": {
+      "@id": "cred:recipient",
+      "@type": "@id"
+    },
+    "referenceId": {
+      "@id": "cred:referenceId"
+    }
+  },
+  "@graph": {
```
msporny: I'm wondering if we should split this out, where doing https://w3id.org/credentials gets this document, and https://w3id.org/credentials/v1 just gets the context file w/o the RDFs/OWL vocabulary description. I'm concerned about file size, but perhaps unnecessarily so. I know that we always cache local copies in our programs and almost never go out to the network to get known contexts.

gkellogg: I'll update the script to create both the all-inclusive vocabulary and a separate context-only file.

vocab/credentials.jsonld
gkellogg: I'll add it, but we haven't defined `@container: @graph` in the spec, yet.

gkellogg:
Updated per @msporny's suggestions. There are some things that may still require abstraction (e.g., a super-class for `RevocationList2017`.

gkellogg:
@msporny Do these changes satisfy your review? Can you approve it?

msporny:
@gkellogg LGTM. @dlongley can you review please?

vocab/credentials.jsonld
```diff
@@ -246,6 +301,18 @@
           "en": "FIXME"
         }
       }
+    ],
+    "rdfs_instances": [
+      {
+        "@id": "cred:revocation",
+        "@type": "rdf:PRoperty",
```
msporny: Typo rdf:PRoperty

vocab/context.jsonld
```diff
@@ -0,0 +1,87 @@
...
+      "@id": "dc:creator",
+      "@type": "@id"
+    },
+    "domain": "sec:domain",
+    "expires": {
+      "@id": "sec:expiration",
+      "@type": "xsd:dateTime"
+    },
+    "nonce": "sec:nonce",
+    "normalizationAlgorithm": "sec:normalizationAlgorithm",
```
dlongley: I believe this has been renamed to `canonicalizationAlgorithm` (both terms appear in the security vocabulary but we decided to move forward with `canonicalizationAlgorithm`). Should I file a separate issue for this or should we just change it here?

vocab/context.jsonld
```diff
@@ -0,0 +1,87 @@
...
+    "CryptographicKeyCredential": "cred:CryptographicKeyCredential",
+    "Profile": "cred:Profile",
+    "RevocationList2017": "cred:RevocationList2017",
+    "claim": {
+      "@id": "cred:claim",
+      "@container": "@graph"
+    },
+    "credential": {
+      "@id": "cred:credential",
+      "@type": "@id"
```
dlongley: The `credential` field should also be: `"@container": "@graph"`. Again, should I file an issue for these sorts of issues separately or should we be making the changes now in this PR?

vocab/context.jsonld
msporny: Let's make the change in this PR.

vocab/context.jsonld
msporny: Let's make the change in this PR... these are "small" changes.

vocab/credentials.jsonld
gkellogg: Got it.

vocab/context.jsonld
gkellogg: 👍

vocab/context.jsonld
gkellogg: 👍

gkellogg:
I made the three suggested changes. Note that the procedure is to update the vocab.csv file and re-run mk_vocab.rb, rather than editing generated documents by hand.

msporny:
+1 to merge, I'll request a merge at the next telecon.

PR #102, opened 2017-12-31
Fix DID syntax to include "example" method

peacekeeper:


peacekeeper:

PR #104, opened 2018-01-30
Add section on Verifiable Profiles to Basic Concepts

msporny:
Don't know why this went as long as it did undetected, but we didn't have a formal definition of what constitutes a Verifiable Profile in the spec. This is now fixed.

/cc @dlongley @burnburn @gkellogg @talltree

### DIFF
--- a/common.js
+++ b/common.js
@@ -36,6 +36,16 @@ var vcwg = {
       status: "CG-DRAFT",
       publisher: "Credentials Community Group"
     },
+    "LD-PROOFS": {
+      title: "Linked Data Proofs",
+      href: "https://w3c-ccg.github.io/ld-proofs/",
+      authors: [
+        "Manu Sporny",
+        "Dave Longley"
+      ],
+      status: "CG-DRAFT",
+      publisher: "Credentials Community Group"
+    },
     // aliases to known references
     "HTTP-SIGNATURES": {
       aliasOf: "http-signatures"
--- a/index.html
+++ b/index.html
@@ -658,6 +658,33 @@ of scope for this specification. A status scheme registry [[VC-STATUS-REGISTRY]]
 exists for implementers that would like to implement credential status checking.
         </p>
       </section>
+
+      <section>
+        <h2>Profiles</h2>
+        <p>
+Credentials MAY be composed by placing them into a data structure called a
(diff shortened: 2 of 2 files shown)


index.html
```diff
@@ -577,7 +577,7 @@ <h2>Proofs (aka Signatures)</h2>
     "created": "2017-06-18T21:19:10Z",
     "creator": "https://example.com/jdoe/keys/1",
     "nonce": "c0ae1c8e-c7e7-469f-b252-86e6a0e7387e",
-    "signatureValue": "BavEll0/I1zpYw8XNi1bgVg/sCneO4Jugez8RwDg/+
+    "proofValue": "BavEll0/I1zpYw8XNi1bgVg/sCneO4Jugez8RwDg/+
```
dlongley: I would expect this to be specific to the proof type, so `signatureValue` would still be used for a signature proof type.

index.html
```diff
@@ -845,11 +872,11 @@ <h3>Subject</h3>
...
         <ul>
-          <li>The document signature is available in the form of a known
-          signature suite.</li>
-          <li>Required signature properties are present. For example, for a
+          <li>The document proof is available in the form of a known
+          proof suite.</li>
+          <li>Required proof properties are present. For example, for a
           Linked Data Signature, <code>type</code>, <code>created</code>,
-          <code>creator></code>, and <code>signatureValue</code> are present.
+          <code>creator</code>, and <code>proofValue</code> are present.
```
dlongley: Again, wouldn't `signatureValue` be used here?

common.js
```diff
@@ -36,6 +36,16 @@ var vcwg = {
       status: "CG-DRAFT",
       publisher: "Credentials Community Group"
     },
+    "LD-PROOFS": {
+      title: "Linked Data Proofs",
+      href: "https://w3c-ccg.github.io/ld-proofs/",
```
BigBlueHat: This currently 404's. Maybe it's not published yet?

dlongley: Not published yet.

msporny:
No objections to merge after two weeks of comment period. Merged.

index.html
David-Chadwick: Need to add text to say: "The ID is the ID of the holder, who may or may not be same as a subject in one of the embedded credentials".

PR #121, opened 2018-02-27
Clarify that credentials and profiles can be multi-subject.

### LINKED ISSUE #55: Multiple subjects in a single credential
Technically, the data model supports a single credential making claims about multiple subjects:

```
{
  "@context": ["http://schema.org/", "https://w3id.org/identity/v1"],
  "type": ["Credential", "MultiSubjectCredential"],
  "issuer": "https://multi.example.org",
  "issued": "2017-06-05",
  "claim": [{
    "id": "did:v1:9f52df2e-7c96-42af-8dfd-1099980f8467",
    "ageOver": 18
  }, {
    "id": "did:v1:7d866a9f-c5a3-41f5-8ae1-0297b7849801",
    "ageOver": 21    
  }, {
    "id": "did:v1:c4e13c30-31dc-40df-867f-ed678d87ac54",
    "ageOver": 65   
  }],
  "signature": {
    "type": "LinkedDataSignature2015",
    "created": "2017-06-05T07:08:36Z",
    "signatureValue": "lQ7VZUeAA...Z5Lk="
  }
}
```

While this may be a corner case, the data model allows for it. Should we change the definition of entity credential and entity profile from being about "a subject" to "usually about the same subject" or equivalent language.

msporny:
This PR clarifies that credentials and profiles can be multi-subject. Fixes #55.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/121.html" title="Last updated on Apr 2, 2018, 9:14 PM GMT (a09e8b0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/121/ea8d840...a09e8b0.html" title="Last updated on Apr 2, 2018, 9:14 PM GMT (a09e8b0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -478,7 +478,7 @@ cards, digital proofs of age, and digital educational certificates.
       <p class="note">
 It is possible to have a <a>credential</a>, such as a marriage certificate,
 that contains multiple <a>claims</a> about different <a>subjects</a> that are
-often, but not required to be, related.
+not required to be related.
       </p>
 
     </section>
--- a/index.html
+++ b/index.html
@@ -453,14 +453,14 @@ of information.
       <h2>Credentials</h2>
 
       <p>
-A <a>credential</a> is set of one or more <a>claims</a> about a <a>subject</a>.
-It typically includes an identifier to uniquely identify the
-credential. Credential metadata may also be included to express concepts
-such as when the credential expires. A digital signature is almost always
-appended by the <a>issuer</a> of the credential to enable verifiability of
-the credential. Therefore, a <strong>verifiable credential</strong> is a set
-of claims that are tamper-resistant and whose authorship can be
-cryptographically verified.
+A <a>credential</a> is set of one or more <a>claims</a> that are often about
+the same <a>subject</a>. A <a>credential </a> typically includes an identifier
+to uniquely identify the credential. Credential metadata may also be included
+to express concepts such as when the credential expires. A digital signature
+is almost always appended by the <a>issuer</a> of the credential to enable
+verifiability of the credential. Therefore, a
+<strong>verifiable credential</strong> is a set of claims that are
+tamper-resistant and whose authorship can be cryptographically verified.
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -475,6 +475,12 @@ <h2>Credentials</h2>
 cards, digital proofs of age, and digital educational certificates.
       </p>
 
+      <p class="note">
+It is possible to have a <a>credential</a>, such as a marriage certificate,
+that contains multiple <a>claims</a> about different <a>subjects</a> that are
```
dlongley: Would they be multiple claims or a single claim with statements about different subjects? Where a "claim" is an isolated graph.

index.html
```diff
@@ -506,6 +512,12 @@ <h2>Profiles</h2>
 gaming persona, or home life persona.
       </p>
 
+      <p class="note">
+It is possible to have a <a>profile</a>, such as a business persona,
+that contains multiple <a>credentials</a> about different <a>subjects</a>
+that are often, but not required to be, related.
```
dlongley: "that are not required to be related". (strike "often").

David-Chadwick:
@msporny 
Hi Manu, how about this revised text which is more terse than yours and therefore I hope removes any ambiguity over what a credential is.

A credential is a set of one or more claims. It includes an identifier to uniquely identify the credential, as well as metadata that describes properties of the credential such as: the issuer, the expiry time etc. A verifiable credential is a set of claims and meta data that are tamper-resistant and that cryptographically prove who issued it.

Manu, note that we have already said that a claim is a statement about a subject. So we do not need to add anything more to the definition of credential. But you might want to add a clarifying note as follows

Note: A credential can contain claims about different subjects.

index.html

msporny:
@David-Chadwick accepted your modifications to the definition and added one more non-substantial thing to align it w/ the glossary.

PR #122, opened 2018-02-27
ADD a new section on Subject NE Holder.

David-Chadwick:
This is a new section that deals with all the identified cases of when the subject is not equal to the holder

***
<a href="https://pr-preview.s3.amazonaws.com/David-Chadwick/vc-data-model/pull/122.html" title="Last updated on Mar 13, 2018, 3:49 PM GMT (9eb79d7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/122/ea8d840...David-Chadwick:9eb79d7.html" title="Last updated on Mar 13, 2018, 3:49 PM GMT (9eb79d7)">Diff</a>


burnburn:
@liam please check on the IPR issue.

davidlehn:
- Remove the `.DS_Store` file from the commit.
- Shouldn't the diagram be SVG like all the others?  And named in kebab-case vs CamelCase.
- "Acts4" in diagram should be written out "Acts for"

index.html
```diff
@@ -823,6 +823,375 @@ <h2>Evidence</h2>
...
+  "claim": {
+    "id": "did:example:ebfeb1f712ebc6f1c2123456789",
+    "credential”: {
+       "id": "http://example.gov/credentials/3732",
+       "type": ["Credential", "Role"],
+       "issuer": "https://dmv.example.gov",
+       "issued": "2010-01-01",
+       "claim": {
+         "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
+         "role": "Project Manager”
```
davidlehn: Fix to normal quotes.

index.html
```diff
@@ -823,6 +823,375 @@ <h2>Evidence</h2>
...
+       "creator": "https://example.com/jdoe/keys/1",
+       "domain": "json-ld.org",
+       "nonce": "598c63d6",
+       "signatureValue": "BavEll0/I1zpYw8XNi1bgVg/sCneO4Jugez8RwDg/
++MCRVpjOboDoe4SxxKjkCOvKiCHGDvc4krqi6Z1n0UfqzxGfmatCuFibcC1wps 
+PRdW+gGsutPTLzvueMWmFhwYmfIFpbBu95t501+rSLHIEuujM/+PXr9Cky6Ed +W3JT24="
+    }
+  }
+},
+"proof": { ... } }
```
davidlehn: fix indents from "proof" above down to here.

davidlehn:
All the JSON examples should be fixed up to normalized indentation style to be easier to read.

liam:
Please don't use @Liam, it doesn't do what you think.

burnburn:
@liamquin looks like I accidentally clicked the wrong name in the list when pinging you above.  I have fixed it above.  Can you please check into the IPR issue now?

index.html
David-Chadwick: proof index fixed

index.html
David-Chadwick: @davidlehn 
Sorry I don't understand your .DS-Store comment.
I do not have a Save as SVG option in my Powerpoint, so I cannot do that. But I am happy for someone else to alter the format of the final version of the diagram for me. File name change done.
Acts4 has been changed in the updated diagram
" in Project Manager has been changed.

liam:
Still getting these mails which aren't meant for me. Please change the Liam
you're sending them to.

liamquin:
@liam sorry, you'll need to unsubscribe from the thread explicitly i think.

- otherliam :)

burnburn:
@liamquin the IPR check is now showing successful, so I no longer need you to fix anything.

David-Chadwick:
This pull request is now superceded by the new one dated 20 March. There are no textual differences.
This pull request can be deleted

davidlehn:
Moved to https://github.com/w3c/vc-data-model/pull/134.  Someone close this.

jandrieu:
I don't think these examples represent a consensus on how to represent relationships, nor do I think it is correct way to handle delegation.

In both cases, I think these suggest elements of the Profile, not the Credential. The presenter of a profile may claim a relationship to an entity in an included credential. They may also claim a delegated authority to act on behalf of the subject, presumably presenting that authority as a claim, preferably a verifiable capability (work in progress in the CCG).

In the case of delegation, there is a valid argument for Issuers to want to include in the credential the delegation which explicitly intended or restricted. For example, "this credential is only valid when presented by the subject or the subject's duly constituted heir." Such an assertion would be part of the credential, not the profile.

In both cases, the claims of different relationships and delegations require a new vocabulary, which is almost certainly out of scope for the current specification. What I propose instead is clarity around the freedom for the issuer and presenter to make whatever claims they wish, in whatever vocabulary they wish. From that foundation, we could show some examples.

Those examples would need to specify the vocabulary for those specific terms. I expect "relationship" will not pass muster as a VCWG endorsed term, but in JSON-LD we can certainly be rigorous about incorporating external vocabularies. Examples that show how to do that would be useful.

That said, it may be appropriate for the CCG (as a non-time limited entity) to host a registry of vocabularies for credentials and profiles.

burnburn:
Closing as per submitter request.

PR #131, opened 2018-03-20
Add section on Trust Model to the specification.

### LINKED ISSUE #110: Trust Model
The data model talks a lot about trust, but the text is widely dispersed throughout the document.
It would be advisable and helpful to the reader, if all the statements about trust could be brought together into a new section entitled The Trust Model. Here is some strawman text for the new section.

Section X. The Trust Model

The VC trust model is as follows. 

1.	The verifier trusts the issuer to issue the VC that it receives (although this trust could be weakened depending upon the risk assessment of the verifier). 
2.	All entities trust the identifier registry to be un-corruptible and to be a correct record of which identifiers belong to which entities.
3.	The subject trusts the issuer to issue true (i.e. not false) credentials about itself, and to revoke them quickly when requested to do so.
4.	The holder trusts the repository to store his/her VCs securely, to not release them to anyone other than the holder, and to not corrupt or lose them whilst they are in its care.

Note in particular that 
i) the issuer and the verifier do not (need to) trust the repository, and  
ii) the issuer does not (need to) trust the verifier.

msporny:
This PR adds a section on the Trust Model to the specification, which uses a modified version of @David-Chadwick's text. Fixes #110.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/131.html" title="Last updated on Apr 18, 2018, 10:15 PM GMT (95d30d1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/131/bdb0974...95d30d1.html" title="Last updated on Apr 18, 2018, 10:15 PM GMT (95d30d1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -513,6 +513,56 @@ that are often, but not required to be, related.
     </section>
 
     <section>
+      <h1>Trust Model</h1>
+
+      <p>
+The Verifiable Credentials trust model is as follows:
+      </p>
+
+      <ol>
+        <li>
+The <a>verifier</a> trusts the <a>issuer</a> to issue the Verifiable Credential
+that it receives. This trust could be weakened depending upon the risk
+assessment of the verifier.
+        </li>
+        <li>
+All entities trust the <a>identifier registry</a> to be un-corruptible and
+to be a correct record of which identifiers belong to which <a>entities</a>.
+        </li>
+        <li>
+The <a>subject</a> trusts the <a>issuer</a> to issue true (i.e. not false)
+<a>credentials</a> about the subject, and to revoke them quickly when
+appropriate.
+        </li>
+        <li>
+The <a>holder</a> trusts the <a>repository</a> to store the
+<a>credentials</a> securely, to not release them to anyone other than the
+<a>holder</a>, and to not corrupt or lose them whilst they are in its care.
+        </li>
+      </ol>
+
+      <p>
+This trust model differentiates itself from other trust models by ensuring that:
+      </p>
+
+      <ul>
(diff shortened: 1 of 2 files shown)


David-Chadwick:
Hi Manu
I dont agree to this PR because it is not simply about the Trust Model, but is also using text from the PR "Clarify that credentials and profiles can be multi-subject. #121" which already has comments  suggesting alternative less ambiguous wording that has 2 thumbs up.
So could we please edit the current PR to be only about the trust model, and then we can vote on that alone, and address the wording of credentials and claims in the other PR
Thankyou

msporny:
@David-Chadwick damn, and how annoying. This PR pulled in some text from another PR I was working on. Yes, I'll remove the bits about the credentials/profiles bit so this PR is only about the trust model.

David-Chadwick:
@msporny Once the credentials/profiles text is removed, the only other comment I have is about the IdP.

+ An identity provider, sometimes abbreviated as <em>IdP</em> is a system
that creates, maintains, and manages identity information for <a>holders</a>
while providing authentication services to <a>relying party</a> applications
within a federation or distributed network. 

In this case the holder is always the subject. Even if the credentials are bearer credentials the assumption is that they will remain with the subject, and if they are not, then they have been stolen by an attacker. Can you add this point please.

msporny:
> Can you add this point please.

Done https://github.com/w3c/vc-data-model/pull/131/commits/da8916e8a0e39fdb5984866f38d408d37b30f1e5

PR #138, opened 2018-03-22
Update vocabulary based on changes to data model document.

### LINKED ISSUE #127: Update/mint new JSON-LD context
The existing https://w3id.org/credentials/v1 context is out of date in non-backwards compatible ways at this point. There are some legacy verifiable credentials that depend on it, so we should either mint a `v2` or come up with a new name (if the group finds that desirable).

gkellogg:
Fixes #127.

There's probably some cruft left in here, so identifying extra terms, properties and classes that can be removed would be useful. Ideally, if we had a set of maintained examples we could use to run by the generated context and do some RDFS entailment checking of the result of transforming to RDF would help make sure we haven't missed something.

### DIFF
--- a/vocab/context.jsonld
+++ b/vocab/context.jsonld
@@ -1,99 +1,67 @@
 {
-  "@context": {
-    "id": "@id",
-    "type": "@type",
-    "cred": "https://w3id.org/credentials#",
-    "dc": "http://purl.org/dc/terms/",
-    "schema": "http://schema.org/",
-    "sec": "https://w3id.org/security#",
-    "xsd": "http://www.w3.org/2001/XMLSchema#",
-    "CryptographicKey": "sec:Key",
-    "GraphSignature2012": "sec:GraphSignature2012",
-    "Identity": "https://w3id.org/identity#Identity",
-    "LinkedDataSignature2015": "sec:LinkedDataSignature2015",
-    "canonicalizationAlgorithm": "sec:canonicalizationAlgorithm",
-    "created": {
-      "@id": "dc:created",
-      "@type": "xsd:dateTime"
+  "@context": [
+    {
+      "@version": 1.1
     },
-    "creator": {
-      "@id": "dc:creator",
-      "@type": "@id"
-    },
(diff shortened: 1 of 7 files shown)


gkellogg:
Also, there are various terms used as classes that are probably more example, than actual (e.g., `SimpleRevocationList2017`). Ideally, there's an abstract super-class that can be used as the _rdfs:range_ of some of these properties, although perhaps defined in another vocabulary.

vocab/context.jsonld
```diff
@@ -76,16 +82,21 @@
...
       "@id": "cred:revocation",
       "@type": "@id"
+    },
+    "termsOfUse": {
+      "@id": "cred:termsOfUse",
+      "@type": "@id"
+    },
+    "verifiableCredential": {
+      "@id": "cred:verifiableCredential",
+      "@type": "@id"
```
dlongley: Should also use `"@container": "@graph"`

vocab/context.jsonld
```diff
@@ -22,8 +22,8 @@
     },
     "domain": "sec:domain",
     "expires": {
-      "@id": "sec:expiration",
```
dlongley: Why don't we want to just reuse this?

vocab/context.jsonld
```diff
@@ -76,16 +82,21 @@
       "@id": "cred:issuer",
       "@type": "@id"
     },
-    "recipient": {
-      "@id": "cred:recipient",
+    "proof": {
```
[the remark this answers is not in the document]
gkellogg: Just a term alias, then. Is there an RDFS definition of https://w3id.org/security? When following redirects to https://web-payments.org/vocabs/security, it seems to just define classes an properties with no range or domain information.

David-Chadwick:
Hi Greg
I notice that we have both picked up on some of the same inconsistencies in the current document.
My PR #141 has some edits which are similar to yours, but others which are different. There are clearly issues to be resolved in the current document. I am not sure what is the most productive way of doing this, are you?

PR #134, opened 2018-03-20
Add new section on Subject.NE.Holder

David-Chadwick:
***
<a href="https://pr-preview.s3.amazonaws.com/David-Chadwick/vc-data-model/pull/134.html" title="Last updated on Mar 20, 2018, 10:52 PM GMT (604eb4c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/134/805722f...David-Chadwick:604eb4c.html" title="Last updated on Mar 20, 2018, 10:52 PM GMT (604eb4c)">Diff</a>

Issue #118, opened 2018-02-20
Data Model in Context of Keys, Signatures & Countersignatures


ChristopherA:
One of my challenges with the current data model focus of the WG is how it connects to the signatures of that data model. As an implementor, I find that confuses me when I consider claim + one or more signatures. 

Lets take a simple verifiable claim (from the [Verifiable Claims Playground](https://w3c-vc.github.io/playground/), which may not be current): 

```
{
  "@context": "https://schema.org/",
  "claim": {
    "id": "did:ALICE",
    "name": "Alice",
    "knows": {
      "id": "did:BOB",
      "name": "Bob"
    }
  }
}
```

When signed by one of the signing keys that is available from Alice's set of DID signing keys, it is a claim from an issuer, and has the trust level / authority that you give to that issuer. Is a claim signed any other way invalid?

When both signed by one of the signing keys that is available from Alice's set of DID signing keys AND SEPARATELY by one of the signing keys that is available from Bob's set of DID signing keys, I think this may have greater authority, as if you trust either Bob or Alice it has trust level authority of them individually, but with both signatures it has the trust authority of both.

If signed by Alice, and the entire structure (including signature) is signed by Bob, what does that mean differently than the previous?

When signed ONLY by one of the signing keys that is available from Bob's set of signing keys, but not Alice's, what is it? 

When it also signed by a third party (in addition to either Alice's or Bob's keys), that could be a witness claim.

But when it is signed by a third party (w/o Alice's or Bob's keys), what is that? An evidence claim?

When a signed claim above also has a time signature in it (a verifiable time stamp), what does that mean?

I presume if it was signed by an otherwise valid key, but not one that is designated as a signing key, then that claim should be considered invalid?

So in general, as an implementor, how do I know that the signing expresses the intent of the signer about the claim?

dlongley:
Only skimmed this briefly right now but it seems like the use of a new term in Linked Data Proofs, `proofPurpose`, would be useful here. It's a term that can be included in the signature meta data that indicates the purpose of the proof/signature and links a verifier to a predicate where they can find the appropriate application suite and key material for verification from the signer.

dlongley:
[...]

PR #141, opened 2018-03-27
Modified data model and terms to remove ambiguity

David-Chadwick:
This PR tries to remove ambiguity from the data model and terms by doing the following things:

i) makes sure that each JSON structure is clearly identified for what it is (i.e. verifiable credential or verifiable profile). A side benefit of this is that the type parameter can be simplified as it no longer needs to specify the base type.

ii) removed inconsistencies whereas sometimes a VC was called a credential and sometimes a verifiableCredential

iii) made the "id" property consistent in that it always comes first in a data type definition and always uniquely identifies that data type

iv) updated the terms to make it clear that a verifiable credential holds a set of claims, that each claim contains the properties of one subject, and that the subject can be different for each claim. Each subject has a unique subject id and each subject comprises a set of properties.

v) that each credential status scheme is clearly identified in the credential status property, and each scheme is responsible for defining the contents of the status property

vi) for verifiable profiles, added the holder id, and made it clear that the existing id is the profile id (consistent with i) above

vii) for terms of use, changed the type into a unique type ID, and made the rest of the contents be dependent upon this type. Consequently new types of terms of use can be defined with different properties to the current examples.

viii) turned evidence into a set, consistent with the existing text, with each element being identified by its unique scheme ID. The contents of the scheme properties are then determined by the scheme itself

ix) removed revocation and replaced with credential status

***
<a href="https://pr-preview.s3.amazonaws.com/David-Chadwick/vc-data-model/pull/141.html" title="Last updated on Mar 27, 2018, 8:24 AM GMT (a572b09)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/141/805722f...David-Chadwick:a572b09.html" title="Last updated on Mar 27, 2018, 8:24 AM GMT (a572b09)">Diff</a>


BigBlueHat:
@David-Chadwick would it be possible to break each of your itemized pieces into separate commits? This PR is covering many topics that while possibly interrelated may be individually contentious or problematic (to someone) in their own way. Keeping the topics as separate as possible would make it much simpler to review.

Thanks!

David-Chadwick:
Yes I will try and do that later this week or next in time for the face to face

msporny:
Hey @David-Chadwick -- yes, what @BigBlueHat said... doing PRs like this are highly problematic and almost never get merged in because they try to do too much at the same time. The other technical reason to not do it this way is that you will invalidate a ton of other PRs in the pipeline because your edits will create massive conflicts with many of the other PRs that touch the same spec text as yours.

I will still go through and comment on this PR, but I'm afraid that we'll have to close it because it's unworkable.

index.html
```diff
@@ -398,8 +398,8 @@ <h2>Core Data Model</h2>
         <h3>Claims</h3>
 
         <p>
-A <a>claim</a> is statement about a <a>subject</a>.
-A <a>subject</a> is an <a>entity</a> about which <a>claims</a> may be made.
+A <a>claim</a> is an assertion about the properties of a <a>subject</a>.
+A <a>subject</a> is an <a>entity</a> described by a claim.
```
msporny: This should be "one or more claims".

index.html
```diff
@@ -453,14 +453,12 @@ <h3>Claims</h3>
...
-A <a>credential</a> is set of one or more <a>claims</a> about a <a>subject</a>.
-It typically includes an identifier to uniquely identify the
-credential. Credential metadata may also be included to express concepts
-such as when the credential expires. A digital signature is almost always
-appended by the <a>issuer</a> of the credential to enable verifiability of
-the credential. Therefore, a <strong>verifiable credential</strong> is a set
-of claims that are tamper-resistant and whose authorship can be
-cryptographically verified.
+A <a>credential</a> is set of one or more <a>claims</a> asserted by the issuer.
+It includes an identifier to uniquely identify the credential, as well as
```
msporny: It does not always include the identifier to uniquely identify the credential.

index.html
```diff
@@ -534,17 +532,21 @@ <h2>Issuer</h2>
...
-  <span class="highlight">"issuer": "https://dmv.example.gov/issuers/14"</span>,
-  <span class="highlight">"issued": "2010-01-01T19:73:24Z"</span>,
-  "claim": {
-    "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
-    "ageOver": 21
-  },
-  "proof": { ... }
-}
-        </pre>
+"verifiableCredential": {
```
msporny: I think this confuses things. The example was expressing a verifiable credential, not a verifiable profile.

index.html
```diff
@@ -534,17 +532,21 @@ <h2>Issuer</h2>
...
+	"id": "< credential id >",
```
msporny: This is not always there.

index.html
```diff
@@ -534,17 +532,21 @@ <h2>Issuer</h2>
...
+	"type": ["ProofOfAgeCredential"],
```
msporny: If we remove "Credential", then generalized processing libraries won't be able to key off of that to process the data structure.

index.html
```diff
@@ -534,17 +532,21 @@ <h2>Issuer</h2>
...
+	"issuer": "https://dmv.example.gov/issuers/14",
+	"issued": "2010-01-01T19:73:24Z",
+	"claims": [{
```
msporny: plural form for properties is a known bad practice.... should be "claim".

index.html
```diff
@@ -534,17 +532,21 @@ <h2>Issuer</h2>
...
+		"subject": {
```
msporny: This completely breaks the data model.

index.html
```diff
@@ -670,9 +684,10 @@ <h2>Profiles</h2>
         </p>
 
         <pre class="example nohighlight" title="Basic structure of a profile">
-{
-  "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
-  "type": ["VerifiableProfile"],
+{"verifiableProfile": {
```
msporny: Highly problematic, I get what you're trying to do here, but this is problematic from a data model perspective as well as how the rest of the world (10%-20% of the top 1 million domains now) are using JSON-LD.

index.html
```diff
@@ -670,9 +684,10 @@ <h2>Profiles</h2>
...
+  "id": "< profile id >",
```
msporny: This is not guaranteed to be there, nor does it need to be.

index.html
```diff
@@ -670,9 +684,10 @@ <h2>Profiles</h2>
...
+  "holder": "< holder id >",
```
msporny: Why is the holder having to identify themselves here?

index.html
```diff
@@ -955,6 +987,9 @@ <h2>Verifiable Credential</h2>
               <li>Any empty value MUST be represented as a null value.</li>
               <li>Any other value MUST be represented as a String type.</li>
             </ul>
+           <p> Certain properties must be single valued, whilst other properties 
+ 	can be an array of values, or a set of objects. The examples and text make it clear which properties can have which types of value.
```
msporny: Hmm, except that this is not enforced by the lower data model. Don't know what to do here, I understand the desire, but getting that working would require an additional layer of JSON Schema that adds complexity and that we probably don't want to impose on developers.

index.html
```diff
@@ -1381,7 +1467,7 @@ <h2>Verifiable Profile</h2>
               either unrestricted or extended in the future.</li>
             </ul>
       <p>The following example demonstrates how to express a
-      simple <a>verifiable profile</a>.</p>
+      simple <a>verifiable profile</a>. STILL TO DO</p>
```
msporny: If you must put in a comment like "STILL TO DO", do so using <p class="issue">STILL TO DO</a> and explain what still needs to be done.

index.html
```diff
@@ -1470,15 +1556,15 @@ <h2>Expressing Verifiable Credentials in WebIDL</h2>
...
+            used as follows in JavaScript: STILL TO DO</p>
 
 <pre class="example" title="Using a simple claim">
   var claim = new Claim1();
   claim.id = "http://example.gov/credentials/3732";
   claim.ageOver = 21;
 
   var credential = new Credential();
-  credential.type = ["Credential", "ProofOfAgeCredential"];
+  credential.type = ["verifiableCredential", "ProofOfAgeCredential"];
```
msporny: Types should start with a capital letter.

terms.html
```diff
@@ -105,7 +102,7 @@
   </dd>
   <dt><dfn data-lt="verifier|verifiers|verifier's|credential verifiers|credential verifier's">verifier</dfn></dt>
   <dd>
-An <a>entity</a> that receives one or more <a>profiles</a> for
+An <a>entity</a> that receives one or more <a>verifiable profiles</a> for
```
msporny: Deutche Telekom requested that not all credentials and profiles are verifiable. This was a reasonable request. Changing the language to "verifiable profile" removes that distinction, which may just result in them formally objecting to the spec if it goes forward in that way. There was a nuance in the spec before that doesn't exist now, we lose functionality in a bad way.

David-Chadwick:
@msporny I believe it was stated at the start of the WG, that JSON-LD was not fundamental to the syntax or data model of verifiable credentials. Anyone who understands and uses plain JSON should be able to use the verifiable credentials spec and create conformant verifiable credentials that will interwork with those implementations that understand and use JSON-LD.  Unless I have misunderstood you, you now seem to be saying that JSON-LD is fundamental to understanding the data model, and to creating conformant credentials (or at least, that perfectly good and understandable JSON cannot be used for verifiable credentials because it breaks JSON-LD (in ways that people who are not familiar with JSON-LD don't understand)).

terms.html
David-Chadwick: I dont think it is a reasonable request to make of the Verifiable Credentials working group, that credentials dont need to be verifiable. This surely broadens the scope too much.

index.html
David-Chadwick: What is this lower data model you are referring to? Surely the VC data model should be self contained, or, if it refers to other standards, make it clear what other standards it relies upon. Personally I think that when the IETF JSON Schema spec is completed, we should define a schema for verifiable credentials, otherwise you are much less likely to get interworking. X.509 had a tightly defined schema, with clear extension points, but even this was too broad to guarantee interworking, which gave rise the PKIX working group being created. We should not develop a VC data model that essentially says any JSON construct can be put anywhere in a VC, as that is bound to guarantee no interworking, and in the worst case the death of VCs, or in the best case the spawning of countless other standards or profiles. (Look where we are with X.509 and PKIX with other 50 more standards and dreadful interworking - I am the co-author of several papers on this if you want the details)

index.html
David-Chadwick: Agreed

index.html
David-Chadwick: I am sure this is news to many people in the working group :-)

David-Chadwick: Which just goes to show why ambiguity needs to be removed from the specification!

index.html
David-Chadwick: But they can key off the JSON type of verifiableCredential or verifiableProfile. This also means that we have tight control over which top level types are standardised. With the current data structure anyone can define their own type (e.g. unverifiableRevocationList) and this would be perfectly acceptable and in accordance to the standard. Surely you are not advocating this?

index.html
David-Chadwick: Since it is defined as an array of values, then plural seems to be perfectly acceptable in this case

index.html
David-Chadwick: This does not break JSON nor my mental model of a verifiable credential. In fact it improves my mental model and understanding of a verifiable credential. 
I think you mean that this breaks some underlying hidden data model that is external to this specification but is not specified as being an integral part of it. This is part of the ambiguity that needs to be removed from the current specification.

index.html
David-Chadwick: Maybe it is better to cater for the top 80-90% ??

index.html
David-Chadwick: I was proposing that the ID becomes mandatory.

index.html
David-Chadwick: I think this is essential. We already know who the subject is, from the claim. But we do not know who the holder is unless the data model tells us this. If the IDs are the same then the verifier knows that the holder is asserting that subject EQ holder. If the IDs are different, then either the subject has multiple IDs, or subject NE holder. Either way it is then up to the verifier to sort out the problem. But at least the holder and data model is giving it some guidance. There have been other suggestions to insert a relationship property here, so that when the IDs are different, the holder can assert what the relationship is. I think this is a good idea as it further helps the verifier to determine who the holder is.

index.html
David-Chadwick: Sorry!!

index.html
David-Chadwick: This was a typo. "verifiableCredentail" should have been deleted entirely.

jandrieu:
There are some fundamental problems with the data model, specifically the indirection of "id". Instead of referring to the data instance of the element in which it appears (the credential or claim or profile),  id refers to the subject of that instance (the subject or holder), which means there is no definitive identifier for that specific credential, claim, or profile. Yet, there is a need for credentials about other credentials and for profiles to make claims about claims within included credentials. Note, this problem goes to the underlying data structure, not just the names or serialization. There is a relatively straightforward fix in the graph model and I would like to get consensus on that approach (perhaps dismissing it) before resolving this PR. While it might not be the best way to resolve the problem, it has put its finger on a very subtle and previously under discussed flaw.

index.html
dlongley: This is a false choice. We can have it both ways, there are many examples of it working in the past.

index.html
dlongley: We should point to use cases where an ID is not desirable. Or we could add a note about how the ID may be omitted to support certain privacy requirements.

index.html
dlongley: It breaks the consistency between the model and syntax when it is understood to be a graph. We can support both graph and tree based approaches -- so we should. There's no need to deviate.

index.html
dlongley: > We should not develop a VC data model that essentially says any JSON construct can be put anywhere in a VC...

I agree. I think this is what @msporny is alluding to when referring to the "lower data model". We have a concept of the generic data model functioning as a graph (or as a tree). We mention this graph when talking about the information related to subjects but it sounds like we should be more explicit about how the entire data model can be viewed consistently that way, which is a powerful mechanism for informing people as to how to extend it in a compliant fashion. In fact, it now seems rather glaring to me that we don't have an extensibility section in the spec at all.

terms.html
dlongley: @David-Chadwick,

> I dont think it is a reasonable request to make of the Verifiable Credentials working group, that credentials dont need to be verifiable. This surely broadens the scope too much.

I recommend taking this argument to Deutche Telekom, who has the power to formally object. If they are fine with it, that's great. If not, I don't know how much we gain by fighting over it.

dlongley:
@jandrieu,

> Instead of referring to the data instance of the element in which it appears (the credential or claim or profile), id refers to the subject of that instance (the subject or holder)...

This don't think this is true. I think the problem is in the naming of the elements. Simply change "Profile" to "Holder" and it would fall in line without any further changes. I believe we just need to be more clear about which concepts we're modeling and why. I believe that the biggest problem around "Profile" has been the name from the start -- it is clearly a loaded term that leads people to have very different ideas about it.

The existence of "Profile" came out of implementation experience. It was merely an ephemeral element that was created to allow the holder to make some statements about themselves, such as which credentials they possess. This is used in DID-Auth and presenting credentials via the Credential Handler API. So, it grew out of an implementation need -- we *knew* we needed this kind of concept (despite it being poorly named) in order to use Verifiable Credentials to authenticate on websites, regardless of other protocol considerations (which are out of scope here).

Therefore, my view of "Profile" has always been "some particular statements about the holder". Given that there is an open world assumption, and anyone can say anything about anything whenever they want, the name "Profile" was chosen to mean "This is just one set of things I'm electing to say about myself." In fact, a "Profile" in current implementations, is just conjured it up, on demand, in order to share some assertions about myself (and credentials I possess proving them) with a website in order to authenticate. After this exchange I'll discard the Profile and never think about it again.

Given this, there does not need to be a special identifier for some particular set of statements made, however, if there is, then it is for the graph of information in which those statements are captured or it is for some other data element that establishes a relationship with such a graph. It is not something internal to that same graph. I wasn't saying anything special about the graph of information I was presenting, I was just presenting the statements in the graph itself. Perhaps the recipient of a Profile would create such an identifier, assign it to the graph presented, and store it internally. I don't really care. But... maybe we should care. And I think that's the flip side to this that you and some others are suggesting.

So, it seems to me, we need to tease out where the differences are in what exactly we think "Profiles" are. Based on what I've been hearing from you and some others, there's a desire for Profiles to be surfaced as more enduring data elements (with their own long-lasting identifiers). I think if that approach better fits the use cases that's fine. But that would be a change in what they are and how they've been implemented. It would not be a simple reconciling of what "id" means inside of them. That is, the meaning of "id" is *already presently consistent* with what a "Profile" is today, at least as understood to some (e.g. me).

Again, I'm happy to change that meaning, if it makes sense -- and, yes, it would require adding another property such as "holder" in order to keep our use of "id" consistent. This "holder" (or other name) property would point at the graph of statements made about the holder.

index.html
David-Chadwick: @dlongley I think you are confusing two things: the ontology and vocabulary for properties of claims which will be defined by communities of users; and the top level data structures that are being standardised. So I agree that communities will define their own vocabularies of claim properties, but they should not define their own top level data structures. Currently we have just two, VCs and VPs, and these should be fixed until we see a requirement for some other type of data structure. Then that can be standardised as well. The current data model does not place any restrictions on the top level data types/structures. This loophole should be closed.

index.html
David-Chadwick: @dlongley I am glad that you agree. But having the "type": construct as the topmost data type, without any restrictions on what types can be placed there, is essentially an unconstrained structure that allows any data type to be defined. Pulling out the two types (VC and VP) into JSON objects, constrains our data model to only support VCs and VPs (for now).
"In fact, it now seems rather glaring to me that we don't have an extensibility section in the spec at all." I would actually say that the converse is currently true, that we do not constrain what can be placed in most places in the data model.

David-Chadwick:
@dlongley I am happy to change profile to holder in my next update if that is the concensus of the group.

jandrieu:
@dlongley we have the same problem with claims within a credential. There is no way to say in a profile that claim X in credential Y is no longer accurate, e.g., if the address is no longer current but it's part of the credential.  Similarly there is no way for someone to witness a particular profile without asserting the underlying assertions. I.e., no way to say "I affirm Joe presented profile X on such and such a date."  These missing ids make it hard to make claims about claims and claims about profiles. We do have usable ids for credentials, but not claims or profiles.

jandrieu:
Btw, I think for profiles, this disappears if we look at profiles as a credential about other credentials, since credentials have a clear id.

index.html
msporny: >  The current data model does not place any restrictions on the top level data types/structures. This loophole should be closed.

This is the core of the miscommunication and the disconnect. You are advocating something that no other W3C Linked Data specification has done. This is a discussion that has happened time and time again over the past 15 years at W3C and unfortunately, I think we're going to have to repeat that discussion to come to consensus on this item. @David-Chadwick, please raise an issue for this and we can move the discussion there.

index.html
msporny: @David-Chadwick I noticed that my markup was overwritten, fixed it, see example above again to see what I was talking about.

terms.html
msporny: @AxelNennker and Jörg Heuer at Deutche Telekom.

msporny: btw, I also agree that there is no reason to insist that there be a proof on the credential.

msporny:
Hey guys, so this PR and set of issues are taking us completely off of the rails. There's way too much to respond to in here and we're getting to the point that we need resolve multiple things in a high bandwidth setting. I'd suggest that we do this at the face-to-face, but only Joe and I will be there. So, it's going to have to be a set of calls and we need to get to the bottom of this quickly as it has the possibility to drastically disrupt the current implementations, deployments, test suite and the WGs timeline.

@burnburn @stonematt -- this is a request to dedicate telecon time (or a task force) to get to the bottom of this and get us all aligned on these questions about the data model.

msporny:
@David-Chadwick said:
>  I have misunderstood you, you now seem to be saying that JSON-LD is fundamental to understanding the data model

No, that's not what I'm saying. What I am saying is that JSON-LD is fundamental to ***extensibility***.

> perfectly good and understandable JSON cannot be used for verifiable credentials because it breaks JSON-LD

I'm saying that the JSON  that you are proposing breaks extensibility using JSON-LD and confuses the subject identifiers in a way that will lead to interoperability failure. There is a delicate balance here that we're trying to strike and what you are proposing, while valid JSON, creates a variety of problems wrt. extensibility and JSON-LD. That is, the specification designers need to think about both worlds so that the developers don't need to.

dlongley:
> That is, the specification designers need to think about both worlds [(JSON and JSON-LD)] so that the developers don't need to.

Yes, this.

terms.html
David-Chadwick: A verifier must always be able to trust the credentials it receives. This is written into the trust model. 

Trust in the credential can be established in two ways:
a) it can be done by using a trusted channel between the issuer and the verifier, and transferring an unsigned credential along a trusted channel, or 
b) it can use an untrusted channel and a credential signed by the issuer. 

I do not see how the VC data model, which requires the VC to be passed from the trusted issuer to an untrusted user, can then establish a trusted channel between the verifier and the untrusted user. Therefore it is essential that for the VC data model, all credentials have proofs and are verifiable.

jandrieu:
I second Manu's request for a focus on resolving these issues. We can discuss at IIW and the face-to-face, but that will be a limited subset of those who have spoken up on the matter.

jandrieu:
@dlongley wrote:
>So, it seems to me, we need to tease out where the differences are in what exactly we think "Profiles" are. Based on what I've been hearing from you and some others, there's a desire for Profiles to be surfaced as more enduring data elements (with their own long-lasting identifiers). I think if that approach better fits the use cases that's fine. But that would be a change in what they are and how they've been implemented. It would not be a simple reconciling of what "id" means inside of them. That is, the meaning of "id" is already presently consistent with what a "Profile" is today, at least as understood to some (e.g. me).

Agreed. Empirically we know there is no such thing as a short lived data element. Without intending snark, the idea that data lives in temporal or spatial isolation is what is at the root of privacy and identity problems globally.

IMO, any data element constituted according to a VCWG specification should have a clear identifiers so that further claims and credentials about those elements can be constructed. I'm good with SHOULD instead of MUST, although I'd prefer the latter.

My second argument is that the only semantic distinction between a profile and a verifiable credential that includes other verifiable credentials is that the profile INCLUDES the innate assertion that the current presenter is the "holder". As such, why isn't it just a special case verifiable credential where that additional semantic is explicit?

This brings us to the third problem. Which I have taken to a separate issue: #148.

index.html
dlongley: One way to resolve this is to just change instances like this to use `SHOULD` language and leave it at that:

`It SHOULD include an identifier to uniquely identify the credential...`

dlongley:
@jandrieu,

> Agreed. Empirically we know there is no such thing as a short lived data element. Without intending snark, the idea that data lives in temporal or spatial isolation is what is at the root of privacy and identity problems globally.

:heart: 

> IMO, any data element constituted according to a VCWG specification should have a clear identifiers so that further claims and credentials about those elements can be constructed. I'm good with SHOULD instead of MUST, although I'd prefer the latter.

I think if we just use SHOULD language for most if not all of the identifiers then that should be good enough to resolve a number of issues.

> My second argument is that the only semantic distinction between a profile and a verifiable credential that includes other verifiable credentials is that the profile INCLUDES the innate assertion that the current presenter is the "holder". As such, why isn't it just a special case verifiable credential where that additional semantic is explicit?

Perhaps we should make it a Verifiable Credential of a specific type rather than having it be a separate element unto itself. This would enable implementations to look for or request that type of credential in particular when performing certain authentication operations.

A "Profile" could be thought of as a type of Verifiable Credential that endorses other credentials. At least, this is how it is presently used in implementations. You present `P` which includes claims about yourself, either self-asserted or made by others through the use of other verifiable credentials. This is essentially an endorsement of what others have said about you (or others, to cover holder != subject cases). So perhaps a "Profile" is really just an "EndorsementCredential".

I remember you took some issue with the use of the term "endorsement" but I can't recall why -- so perhaps there's a different term that would be more amenable here. But I do think the concept can work.

This same type of credential can be used for other use cases where you want to assert that you are in support of what others have said, regardless of whom it is about. Binding yourself to "whom it is about"  may have to be asserted through another mechanism, but that's for #134.

jandrieu:
@dlongley Cool. I think we are in agreement about potentially profiles as special case VCs.

FWIW, my issue with endorsement was that it was ambiguous. This ends up in the issues about the semantics of signatures. #118

index.html
David-Chadwick: @dlongley. Can you please explain why the JSON structure I have proposed is not a graph. All it has done is added an extra node called 'claims'

I did not say it didn't represent a graph. I said it breaks the existing consistency between the model and the syntax from the graph perspective -- and what we've done so far. If it is to be consistent with how everything else works, then it doesn't represent a graph we would prefer. So to accept this is to choose between consistency (and the oddity that results from it here) or inconsistency (the "breaking the data model" that @msporny speaks of).

To elaborate, the current graph model and syntax (JSON-LD) works like this:

```js
{
  "id": "<identifier of an RDF subject>",
  "<term referring to an RDF property>": {
    "id": "<identifier of another RDF subject in the object position>",
    /* so on */
  }
}
```

So something like this:

```js
{
  "id": "identifierA",
  "name": "Alice",
  "knows": {
    "id": "identifierB",
    "name": "Bob"
  }
}
```

Yields the triples:

```
<identifierA> <name> "Alice" .
<identifierB> <name> "Bob" .
<identifierA> <knows> <identifierB> .
```

This is inline with what's in the data model when talking about simple statements forming a graph.

When "id" is omitted, you can think of a graph node as merely existential, unlabeled, or "blank". These get assigned "blank node identifiers" (of the form `_:something`) in triple form. So, if we had this:

```js
{
  "id": "identifierA",
  "name": "Alice",
  "knows": {
    "name": "Bob"
  }
}
```

These are the triples:

```
<identifierA> <name> "Alice" .
_:b0 <name> "Bob" .
<identifierA> <knows> _:b0 .
```

This means Alice knows someone with the name Bob. We don't have an identifier for Bob, though.

Furthermore, we can embed graphs within graphs by defining the properties/terms we use as having values that occur in their own subgraphs. This is important for isolating groups of statements and what is done with "claim" presently. It allows each value of "claim" to be interpreted as a separate graph full of whatever triples are expressed therein.

So -- anyway, the problem with what you suggested is not with "claims" but with "subject". When you introduce "subject", if we are to be consistent with what we've done so far, then you create an entity that has no identifier that has a property "subject". I don't know what this relationship means or what this extra unidentified thing is. In other words, we have this:

A credential contains a claim which is a subgraph of statements. In that graph exists an unidentified thing that has a property "subject" that relates this unidentified thing (somehow) to another thing that we do identify.

What is this bizarre unidentified thing? Why does it exist? From a graph perspective, it has no purpose -- it appears to be added for reasons unrelated to consistency of the graph perspective or because of a misunderstanding of it. Why not eliminate it and just talk about the things you want to claim like usual?

Alternatively, we could define "subject" to have no real meaning in the graph data model -- causing it to be some kind of "pass through" nesting that only appears because we failed to model things properly. Either we make this explicit so graph-based processors can understand this blight, or we break consistency. Neither is a good choice, particularly when we're inventing something new, not trying to adapt something we're stuck with. It would be better not to break with how everything else works, by introducing a special case that causes additional cognitive load for those who understand and want to use the graph perspective. In fact, from a pure JSON perspective, I find the additional nesting of "subject" unappealing as well.

I think we can come up with something that doesn't cause this problem. Other standards groups have been quite successful at doing data modeling that works from both the JSON-LD graph perspective and the plain JSON tree perspective.

msporny:
PR superseded by smaller PRs that achieve what is mergeable. Closing.

PR #145, opened 2018-03-30
Fixed Credential Status Scheme

David-Chadwick:
This is item v) from PR#141

***
<a href="https://pr-preview.s3.amazonaws.com/David-Chadwick/vc-data-model/pull/145.html" title="Last updated on Mar 30, 2018, 10:38 AM GMT (c11e7d6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/145/805722f...David-Chadwick:c11e7d6.html" title="Last updated on Mar 30, 2018, 10:38 AM GMT (c11e7d6)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -628,7 +628,8 @@ will cease to be valid.
           <dt><var>credentialStatus</var></dt>
           <dd>The value of this property MUST be a status scheme that
             provides enough information to determine the current status of the
-            credential (e.g. suspended, revoked, etc.). The status scheme will
+            credential (e.g. suspended, revoked, etc.). The contents of the status scheme
+	 property are determined by the definition of the scheme. This will
             vary depending on a variety of factors, such as whether it is
             simple to implement or privacy-enhancing.
           </dd>
@@ -645,7 +646,7 @@ will cease to be valid.
     "ageOver": 21
   },
   <span class="highlight">"credentialStatus": {
-    "id": "https://dmv.example.gov/status/24,
+    "schemeID": "https://dmv.example.gov/status/24,
     "type": "CredentialStatusList2017"
   }</span>,
   "proof": { ... }


index.html
```diff
@@ -645,7 +646,7 @@ <h2>Status</h2>
     "ageOver": 21
   },
   <span class="highlight">"credentialStatus": {
-    "id": "https://dmv.example.gov/status/24,
+    "schemeID": "https://dmv.example.gov/status/24,
```
msporny: This breaks the standard JSON-LD data model. We have to keep "id".

David-Chadwick: The problem is an inconsistency between the text and the JSON example. The text refers to a status scheme but the original JSON does not contain any mention of a status scheme. Closer alignment of the text and the JSON is needed so that the reader can better understand how the JSON is an example of the text.

burnburn:
@David-Chadwick would it be okay to make the change to 'id' for JSON-LD reasons so we can merge and then you can open a new issue or PR regarding the attribute name?

David-Chadwick:
Yes OK. However the text would need to change to be consistent with the JSON. Can I suggest that it changes from this:

"The value of this property MUST be a status scheme that provides enough information to determine the current status of the credential (e.g. suspended, revoked, etc.). The status scheme will vary depending on a variety of factors, such as whether it is simple to implement or privacy-enhancing. "

to this

"The value of this property MUST be a credential status scheme that provides enough information to determine the current status of the credential (e.g. suspended, revoked, etc.). The credential status scheme is identified by the type and id properties, and will vary depending on a variety of factors, such as whether it is simple to implement or privacy-enhancing."

msporny:
> the type and id properties

Just be clear that `id` may not be required in all cases, but `type` is. That is `id` is a SHOULD, but `type` is a MUST.

msporny:
Superceded by PR #176. Closing without merging.

PR #152, opened 2018-04-03
Add section for Disputes. Fixes #111.

### LINKED ISSUE #111: Dispute Resolution
Currently there is no text about dispute resolution for example when the subject disagrees with a claim made by an issuer.

5. Dispute Resolution
The situation may arise when a subject or holder determines that the claim information held by an issuer is incorrect, and consequently, that all VCs issued by the issuer to the holder contain incorrect information about the subject. This is a tricky situation, since the issuer is the owner of the information that it issues, but the subject often knows better than the issuer whether the information is correct or not. Before changing any of the held information, the issuer needs to be convinced that it is incorrect, that the requester is the true subject (or his/her agent and not an imposter) and that the subject is not trying to either insert false information or remove true information. 

Consequently it is recommended that all issuers publish their procedures by which subjects may
i)	obtain the complete set of information held by the issuer about the subject
ii)	determine which subset of this information may be issued as VCs to holders (and as a corollary, which must not be inserted into VCs)
iii)	obtain one or more VCs
iv)	ask for an issued VC to be revoked
v)	ask for information held by the issuer to be updated
vi)	resolve disputes arising from either the information held by the issuer about the subject, or the issuing and revoking of VCs by the issuer.

Legislation such as the EU GDPR [http://ec.europa.eu/justice/data-protection/reform/files/regulation_oj_en.pdf] provide the subject with the right to request a copy of the information held by the issuer, the right to correct erroneous information, and the right to lodge a complaint with a supervisory authority if any dispute with the issuer cannot be resolved amicably.

msporny:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/152.html" title="Last updated on Apr 16, 2018, 8:48 PM GMT (a2a8f1a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/152/23bd58b...a2a8f1a.html" title="Last updated on Apr 16, 2018, 8:48 PM GMT (a2a8f1a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -827,7 +827,7 @@ An <a>entity</a> may dispute a <a>credential</a> issued by another entity.
 The mechanism for doing this is the same as issuing a regular credential
 except that the <a>subject</a> identifier for the <a>claims</a> are those
 of the disputed credential. For example, if a disputed credential with an
-identifier of <code>https://example.org/credentials/123</code> contains
+identifier of <code>https://example.org/credentials/245</code> contains
 disputed statements, an entity may issue the following credential in a
 public venue to make it known that the credential is disputed:
         </p>
--- a/index.html
+++ b/index.html
@@ -820,6 +820,46 @@ very simple scheme for evidence as a part of this specification.
 
       </section>
 
+      <section>
+        <h2>Disputes</h2>
+        <p>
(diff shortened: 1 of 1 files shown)


msporny:
Good catch, @David-Chadwick ... fixed.

PR #155, opened 2018-04-03
Add proofPurpose requirement to spec. Fixes #103.

### LINKED ISSUE #103: Property name for bunding credentials in a profile?
We've had a discussion item come up in the Credentials CG where we're trying to determine the best name for the set of ways a Decentralized ID may issue verifiable credentials (Issuing suites), vs. the container name for verifiable credentials (here are my credentials). Currently, the options are:

* issue (downside is similarity to `issuer` property)
* credential (downside is this could be confused with application/purpose)
* issuing (a bit strange - term is a verb)
* issuingSuite (verbose, but may be a good pattern... e.g. authenticationSuite, encryptionSuite, issuingSuite, etc.)

So, maybe it would be best for nouns to hold stuff that you'd associate with "things you own" and "...Suite" would be used for "ways to perform actions". Fundamentally, we need to harmonize these terms in the VC Data Model and the DID specification.

### LINKED ISSUE #118: Data Model in Context of Keys, Signatures & Countersignatures

msporny:
Fixes #118.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/155.html" title="Last updated on Apr 7, 2018, 6:24 PM GMT (7dc8330)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/155/23bd58b...7dc8330.html" title="Last updated on Apr 7, 2018, 6:24 PM GMT (7dc8330)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -871,28 +871,50 @@ applicable to only some credentials.
       </section>
       <section>
         <h3>Signatures / Proofs</h3>
+
+        <p>
+The cryptographic mechanism used to prove that the information in a
+<a>verifiable credential</a> or a <a>verifiable profile</a> has not been
+tampered with is called a proof. There are many types of cryptographic proofs
+including but not limited to, digital signatures, zero knowledge proofs,
+proof of work, and proof of stake. In general, when verifying proofs,
+implementations MUST ensure that:
+        </p>
+
         <ul>
-          <li>The document proof is available in the form of a known
-          proof suite.</li>
-          <li>Required proof properties are present. For example, for a
-          Linked Data Signature, <code>type</code>, <code>created</code>,
-          <code>creator</code>, <code>proofPurpose</code> and
-          <code>signatureValue</code> are present.
-          </li>
-          <li>The public key associated with the signature is available
-          and a trustworthy link between this signing key and the
-          issuer's <a>verifiable profile</a> may be established. The key must
-          not be revoked or expired.</li>
-          <li>The cryptographic signature is valid.</li>
-          <li>If a <code>proofPurpose</code> exists, it MUST be a valid value
-          per the cryptographic suite. For example, for
-          most digital signature mechanisms the value of
-          <code>proofPurpose</code> MUST be
-          <code>verifiableCredentialIssuance</code> and the public key material
-          used to create the signature MUST be available via the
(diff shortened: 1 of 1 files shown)


nage:
+1 As per the discussion, making this a checked field provided by the signer for signature suites that require it, does appear to help signature schemes that don't differentiate at a more fundamental level.

msporny:
Clarify that proofPurpose is optional, then pull in the PR.

msporny:
Fixes #118.

msporny:
TODO: Clarify that proofPurpose is optional, then pull in the PR.

Issue #158, opened 2018-04-08
Add "criticalProperty" field


msporny:
We may want to add a "criticalProperties" field to the specification to match the security properties provided by x509 Critical extensions flag: http://www.securesenses.net/2013/05/x509-certificates-critical-vs-non.html

It would specify terms that MUST be understood and processed by a verifier. The field would prevent the usage of credentials that are only partially understandable by a verifier.

The drawback of this approach is that some issuers may mark fields as critical that are not for a particular verifier. Ignoring this field is always up to the verifier.

Here's an example:

```
{
  "id": "http://dmv.example.gov/credentials/3732",
  "type": ["Credential", "ProofOfAgeCredential"],
  // the property below is the suggested addition to the data model
  "criticalProperty": [
    "ageOver", "termsOfUse", "assigner", "assignee", 
    "prohibition", "target", "action"
  ],
  "issuer": "https://dmv.example.gov/issuers/14",
  "issued": "2010-01-01T19:73:24Z",
  "claim": {
    "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
    "ageOver": 21
  },
  "termsOfUse": [{
    "type": "Policy",
    "uid": "http://example.com/policies/credential/4",
    "profile": "http://example.com/profiles/credential",
    "prohibition": [{
      "assigner": "https://dmv.example.gov/issuers/14",
      "assignee": "AllVerifiers",
      "target": "http://dmv.example.gov/credentials/3732",
      "action": ["Archival"]
    }]
  },
  "proof": { ... }
}
```

David-Chadwick:
It is true that any verifier can ignore any field in an X.509 certificate or a verifiable credential, and accept it as valid, even if it has been revoked, or contains unknown properties. There is a significant difference however between obeying the fields in a certificate/credential, and ignoring them. In the former case the issuer takes the liability, in the latter case the verifier takes the liability.

dlongley:
I don't think we need "critical" flags at all, I don't think they are actually useful.

An issuer that does not want liability will mark all properties critical. An issuer that wants their VCs to be accepted (i.e. interoperability) will only mark fields critical based on the consensus of at least a majority verifiers (maybe more). In short, whether or not the flag is relied upon determines whether or not the flag exists. This defeats the purpose of the flag.

[...]

PR #159, opened 2018-04-08
Add a section on extensibility to the specification.

msporny:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/159.html" title="Last updated on Apr 18, 2018, 9:53 PM GMT (9722e47)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/159/80dadfb...9722e47.html" title="Last updated on Apr 18, 2018, 9:53 PM GMT (9722e47)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -694,6 +694,134 @@ Linked Data Proofs [[LD-PROOFS]] specification.
       <h1>Advanced Concepts</h1>
 
       <section>
+        <h2>Extensibility</h2>
+
+        <p>
+One of the goals of the Verifiable Credentials Data Model is to enable
+permissionless innovation. This requires that the data model is extensible
+in a number of different ways:
+        </p>
+
+        <ul>
+          <li>
+The requirement to model complex multi-entity relationships is provided
+through the use of a graph-based data model.
+          </li>
+          <li>
+The requirement to be able to extend the machine-readable vocabularies
+used to describe information in the data model without the use of a
+centralized system for doing so is accomplished via the use
+of [[LINKED-DATA]].
+          </li>
+          <li>
+The requirement to support multiple types of cryptographic proof formats is
+accomplished via the use of [[LINKED-DATA-PROOFS]], [[LINKED-DATA-SIGNATURES]],
+and a variety of signature suites.
+          </li>
+          <li>
+The requirement to provide all of the extensibility mechanisms outlined above
+in a data format that is popular among software developers and web page authors
+is enabled via the use of [[!JSON-LD]].
+          </li>
+        </ul>
+
+        <p>
+This approach to data modelling is often called an "open world assumption",
+meaning that any entity can say anything about any other entity. This approach
+often feels in conflict with building simple and predictable software systems.
+Balacing extensibility with program correctness is always more challenging
+with an open world assumption than it is with closed software systems.
+        </p>
+
+        <p>
+The rest of this section describes how both extensibility and program
+correctness are achieved through a series of examples.
+        </p>
+
+        <p>
+Let us assume that we start with the following <a>verifiable credential</a>:
+        </p>
+
+        <pre class="example nohighlight" title="A simple credential">
+{
+  "@context": "https://w3id.org/credentials/v1",
+  "id": "http://example.com/credentials/4643",
+  "type": ["Credential"],
+  "issuer": "https://example.com/issuers/14",
+  "issued": "2018-02-24T05:28:04Z",
+  "claim": {
+    "id": "did:example:abcdef1234567",
+    "name": "Jane Doe"
+  },
+  "proof": { ... }
+}
+        </pre>
+
+        <p>
+The credential above simply states that the entity associated with
+<code>did:example:abcdef1234567</code> has a <code>name</code> with a
+value of <code>Jane Doe</code>. Let's assume that a developer wanted to
+extend the <a>verifiable credential</a> to store two additional pieces of
+information: an internal corporate reference number, and Jane's favorite
+food.
+        </p>
+
+        <p>
+The first thing that a developer would do is create a JSON-LD Context
+containing the two new terms:
+        </p>
+
+        <pre class="example nohighlight" title="An example JSON-LD Context">
+{
+  "@context": {
+    "referenceNumber": "https://example.com/vocab#referenceNumber",
+    "favoriteFood": "https://example.com/vocab#favoriteFood"
+  }
+}
+        </pre>
+
+        <p>
+Now that the JSON-LD Context has been created, the developer must publish it
+somewhere that is accessible to <a>verifiers</a> that will be processing the
+<a>verifiable credential</a>. For this example, let us assume that the
+JSON-LD Context above is published at the following URL:
+<code>https://example.com/contexts/mycontext.jsonld</code>. At this point,
+extending the first example in this section is a simple matter of including the
+context above and adding the new properties to the <a>verifiable credential</a>.
+        </p>
+
+        <pre class="example nohighlight"
+          title="A verifiable credential with a custom extension">
+{
+  "@context": <span class="highlight">[
+    </span>"https://w3id.org/credentials/v1",
+    <span class="highlight">"https://example.com/contexts/mycontext.jsonld"
+  ]</span>,
+  "id": "http://example.com/credentials/4643",
+  "type": ["Credential"],
+  "issuer": "https://example.com/issuers/14",
+  "issued": "2018-02-24T05:28:04Z",
+  <span class="highlight">"referenceNumber": 83294847,</span>
+  "claim": {
+    "id": "did:example:abcdef1234567",
+    "name": "Jane Doe",
+    <span class="highlight">"favoriteFood": "Papaya"</span>
+  },
+  "proof": { ... }
+}
+        </pre>
+
+        <p>
+The examples so far have shown that it is easy to extend the Verifiable
+Credentials Data Model in a permissionless and decentralized way. The
+mechanism shown also ensures that verifiable credentials that were created
+in this way provide a mechanism to prevent namespace conflicts and
+semantic ambiguity.
+        </p>
+
+      </section>
+
+      <section>
         <h2>Terms of Use</h2>
         <p>Terms of use can be utilized by an <a>issuer</a>, <a>subject</a> or a
         <a>holder</a> to express limitations on the use of information
--- a/index.html
+++ b/index.html
@@ -819,6 +819,17 @@ in this way provide a mechanism to prevent namespace conflicts and
 semantic ambiguity.
         </p>
 
+        <p>
+An extensibility model that is this dynamic does increase implementation burden.
+Software written for such a system will have to determine if accepting
+<a>verifiable credentials</a> with extensions is acceptable based on the
+risk profile of the application. Some applications may choose to accept but
+ignore extensions, others may choose to only accept certain extensions, while
+highly secure environments may require that no extensions are allowed. These
+decisions are up to the developers of these applications and are specifically
+not the domain of this specification.
+        </p>
+
       </section>
 
       <section>
--- a/index.html
+++ b/index.html
@@ -830,6 +830,29 @@ decisions are up to the developers of these applications and are specifically
 not the domain of this specification.
         </p>
 
+        <p>
+Implementations MUST produce an error when an extension JSON-LD Context
+overrides the expanded URL for a term specified in the base JSON-LD Context
+(<code>https://w3id.org/credentials/v1</code>). To avoid the possibility of
+accidentally overriding terms, developers are urged to scope their extensions.
+For example, the following extension scopes the new <code>favoriteFood</code>
+term so that it may only be used within the <code>claim</code> property:
+        </p>
+
+        <pre class="example nohighlight" title="Scoping terms in a JSON-LD Context">
+{
+  "@context": {
+    "referenceNumber": "https://example.com/vocab#referenceNumber",
+    <span class="highlight">"claim": {
+      "@id": "https://w3id.org/credentials#claim",
+      "@context": {
+        "favoriteFood": "https://example.com/vocab#favoriteFood",
+      }
+    }</span>
+  }
+}
+        </pre>
+
       </section>
 
       <section>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -701,6 +701,145 @@ <h2>Profiles</h2>
...
+in a data format that is popular among software developers and web page authors
+is enabled via the use of [[!JSON-LD]].
+          </li>
+        </ul>
+
+        <p>
+This approach to data modelling is often called an "open world assumption",
+meaning that any entity can say anything about any other entity. This approach
+often feels in conflict with building simple and predictable software systems.
+Balacing extensibility with program correctness is always more challenging
```
waywardmonkeys: Balancing.

jandrieu:
@David-Chadwick wrote:
> What is within scope should be telling people how VCs can be extended in a controlled manner that does not undo the base model.

That's what this PR does. Literally. It explains, with a concrete example, how someone can extend the JSON properties without undoing the base model. That said, much thanks to Dave & Manu for putting something concrete up that we can discuss.

Perhaps a sentence clarifying that any such additional contexts MUST avoid redefining any existing terms? That would make sense to me. I'm sure JSON-LD has a technique for that, but it should be clarified.

What I think is missing from this PR is language that within a claimset of a credential or a profile, the issuer or holder, respectively, are free to say ANYTHING. The VC spec itself enables complete freedom in the semantics of the claims asserted.

The example in the PR includes extensibility in both the "envelope" and in the claims. It hadn't occurred to me that JSON-LD is the solution for both, but in hindsight, it should have been obvious.

I don't know if JSON-LD allows scoping of contexts, but I would prefer an example of vocabularies within claims that are separated from the terms of the credential properties. In fact, because we can and should expect external vocabularies, I'm hoping this is a simple matter. Issuers and Holders will definitely be pulling in vocabularies that overlap with ours.

In short, I think 
1. We need to point out that contexts pulled in to extend the envelope MUST NOT have name collisions with the VC context.
2. We need an example that illustrates how to pull in claim-specific contexts that are limited in scope and thereby unable to infect the envelope.
3. To what I see as David's security concern, we should have a standard way for Issuers & Holders to specify that given extended terms have specific security-relevant properties, e.g., optional, mandatory, must be understood, etc.
4. To David's point about offline use, it may be a good idea that for non-standard extensions **to the envelope**, the context MUST be included in-line. That addresses, I think, your core concern while still leveraging the JSON-LD approach for extensibility. It also has the added benefit of incentivizing issuers with such requirements to promote their terminology at the standards level if the added burden of inlining is too great.

All of that said, I don't think we have a non-JSON-LD option for extensibility on the table, especially not one that concisely handles both extensions to the envelope and the inevitable domain-specific vocabularies in claims. If there is an option worth considering, speak up, because a non-extensible solution, or a solution that is extensible but collidable (like pure JSON), is not going to be sufficient for the use cases under development.

David-Chadwick:
The text I wrote concerning DDOS attacks when you have to fetch information online to validate a VC seems to have been lost (although it is touched on in Joe's point 4 above). Does anyone know where the text has gone to, since it provides the rationale for the above?
Concerning extensibility, I think the standard should do one of two things, Either:
a) mandate, with a MUST statement, that a verifier must discard any VC containing any field it does not understand, OR
b) for each parameter in the envelope that is not specified in the base standard, mandate with a MUST statement, that it must say whether it is mandatory or optional for the verifier to understand it, and if it is mandatory, the verifier must discard the VC if it does not understand it.
It should then be relatively easy to text for extensibility in the test suite, since must statements always have to be tested for

msporny:
> We need to point out that contexts pulled in to extend the envelope MUST NOT have name collisions with the VC context.

Done. https://github.com/w3c/vc-data-model/pull/159/commits/8da83cd1639d27ed55fe97cd581677e8b0295a6e

> We need an example that illustrates how to pull in claim-specific contexts that are limited in scope and thereby unable to infect the envelope.

Done. https://github.com/w3c/vc-data-model/pull/159/commits/8da83cd1639d27ed55fe97cd581677e8b0295a6e

> To what I see as David's security concern, we should have a standard way for Issuers & Holders to specify that given extended terms have specific security-relevant properties, e.g., optional, mandatory, must be understood, etc.

This is captured in https://github.com/w3c/vc-data-model/issues/158 and will be dealt with in a separate PR.

> To David's point about offline use, it may be a good idea that for non-standard extensions to the envelope, the context MUST be included in-line. That addresses, I think, your core concern while still leveraging the JSON-LD approach for extensibility. It also has the added benefit of incentivizing issuers with such requirements to promote their terminology at the standards level if the added burden of inlining is too great.

I hesitate to put "MUST" as there are good reasons to not always inline (highly cache-able contexts, content-addressed contexts, etc.). As a compromise, I've put language into the spec that makes it clear that extension JSON-LD Contexts should be made highly available: https://github.com/w3c/vc-data-model/pull/159/commits/9722e47dddcd16045ce288da90ddcadbc25731e7

msporny:
@David-Chadwick wrote:
> Concerning extensibility, I think the standard should do one of two things, Either:
> a) mandate, with a MUST statement, that a verifier must discard any VC containing any field it does not understand, OR
> b) for each parameter in the envelope that is not specified in the base standard, mandate with a MUST statement, that it must say whether it is mandatory or optional for the verifier to understand it, and if it is mandatory, the verifier must discard the VC if it does not understand it.

This is being tracked in #158.

msporny:
I believe I've addressed all of the issues pertinent to this PR. Pulling it in. I expect a separate PR against #158 to address @David-Chadwick's remaining concerns.

David-Chadwick:
Hi Manu

On 18/04/2018 22:58, Manu Sporny wrote:
> I believe I've addressed all of the issues pertinent to this PR.

I do not believe that this is the case. Specifically it prejudges the
outcome of issue #158 with the text

"An extensibility model that is this dynamic does increase
implementation burden.
Software written for such a system will have to determine if accepting
<a>verifiable credentials</a> with extensions is acceptable based on the
risk profile of the application. Some applications may choose to accept
but ignore extensions, others may choose to only accept certain
extensions, while highly secure environments may require that no
extensions are allowed. These decisions are up to the developers of
these applications and are specifically not the domain of this
specification."

The above has already determined how extensions can be handled and
negates the need for any 'must understand' field.

Furthermore the following text is a nonsense, given that the above text
allows implementors to ignore extensions.

"Implementations MUST produce an error when an extension JSON-LD Context
overrides the expanded URL for a term specified in the base JSON-LD
Context (<code>https://w3id.org/credentials/v1</code>)."

If an implementation ignores extensions then it cannot know if an
unknown extension overrides a base term that it does understand.

I would therefore like to request that this PR be removed until it can
be fully agreed what text should be inserted into the standard

Kind regards

David

David-Chadwick:
Hi Manu

Concerning the unavailability of context information, your text

"Implementations that cannot fetch a context will produce an error."

should have said

"Implementations that cannot fetch a context MUST produce an error."

so that it can be tested for. (I have no idea what "will" means wrt
conformance testing)

P.s. Yet another reason why this PR should be removed until the full
text can be agreed upon

kind regards

David

msporny:
@David-Chadwick wrote:
>> I believe I've addressed all of the issues pertinent to this PR.
> I do not believe that this is the case. Specifically it prejudges the outcome of issue #158 with the text

This PR has to do with adding a section on extensibility to the specification, not in specifying critical properties. That needs to be a separate PR (otherwise, we're going to tie the existence of a section that explains extensibility to a specific mechanism on how we specify critical properties for extensibility.

> The above has already determined how extensions can be handled and negates the need for any 'must understand' field.

No, it doesn't negate the need for a 'must understand' field. Note that "must understand" is still advisory, implementations can choose to ignore them, that's what the text is saying. Yes, we can write spec text saying you "must understand", but implementations can still choose to ignore that warning at their peril.

How would you change the wording to address this? The intent was certainly not to ignore the need for a critical properties field, although I note that issue is being challenged at present and may take a while to sort out. 

> Furthermore the following text is a nonsense, given that the above text allows implementors to ignore extensions.
>> "Implementations MUST produce an error when an extension JSON-LD Context overrides the expanded URL for a term specified in the base JSON-LD Context (<code>https://w3id.org/credentials/v1</code>)."

This is something that @jandrieu asked for and I believe it makes technical sense.

> If an implementation ignores extensions then it cannot know if an unknown extension overrides a base term that it does understand.

An implementation MUST NOT ignore an extension at a JSON-LD syntax level, but it may ignore it at a semantic level. The text above is about syntactic checking, not semantic checking. You MUST do the former, you may not need to do the latter (unless there is a critical properties field, but even then, implementations can choose to ignore it regardless of what the spec says).

> "Implementations that cannot fetch a context will produce an error."
> should have said
> "Implementations that cannot fetch a context MUST produce an error."

Avoiding the use of MUST was a deliberate editorial decision. If we put a MUST there, then we are re-stating conformance criteria for JSON-LD processors in a specification that is not the JSON-LD specification. This is almost always a bad idea. A JSON-LD processor that cannot fetch a context will throw an error per the spec, and every implementation of JSON-LD that I know of does this. We could start putting extra conformance criteria for JSON-LD processors in the VC spec, but now we're creating a subclass of JSON-LD processors with new normative requirements, and I expect that to not go over well with the JSON-LD implementers and community group.

> I would therefore like to request that this PR be removed until it can be fully agreed what text should be inserted into the standard.

To be clear, you're asking that the Extensibility section be removed from the specification because you feel that issue #158 is intricately linked to the Extensibility section? I suggest that it is not and the issues can be deal with separately. Can you suggest some concrete text changes that would resolve your concerns?

David-Chadwick:
@msporny I think proposing alternative text at this point will not be productive because we have very different world views of what should be allowed and what should not be. If I interpret your view correctly, it is that anything is allowed in a VC providing it is JSON-LD, even if it is frowned upon in the VC standard, because implementors can do anything they want to.

My view is that the contents of a verifiable credential should be very specific, with specific rules for what is allowed and what is not allowed. If an implementation does not follow the standard then it will fail the conformance testing. Of course, implementors can do anything they want to, as you say, but the standard should make it very clear that this is a non-conformant implementation. This is the key difference between us: whether an implementation conforms to the VC standard or not.

So to put a concrete example of these different world views.
You say "An implementation ... may ignore [an extension] at a semantic level."
I say "An implementation MUST NOT ignore an extension that is marked critical and MUST reject the VC if it does not understand the extension"

We have so many problems with non-conformant X.509 implementations accepting certificates that must be rejected, and ignoring fields that should not be ignored etc. Sure, it makes the system more usable, because users can continue to access web sites with bad/faulty certificates or by using bad client software, but it makes the security of the user (the relying party) non-existent in the worst case.  We dont want to make the security of verifiers non-existent. VCs are meant to be a security tool, so the standard should make it very clear what makes a VC secure and what does not. This does not stop implementors building insecure and non-conformant systems, but they cannot say, 'well we followed the standard, so it must be OK'. Verifiers that want to build a secure site will choose to use only standards conforming implementations. Verifiers that dont really care about risk or security can use any implementation that will accept any VC providing it is standard JSON LD syntax.

msporny:
> I think proposing alternative text at this point will not be productive because we have very different world views of what should be allowed and what should not be.

I don't think they're as different as you think they are. We are talking about two different things, but I think you still think we're talking about a single thing.

> If I interpret your view correctly, it is that anything is allowed in a VC providing it is JSON-LD, even if it is frowned upon in the VC standard, because implementors can do anything they want to.

From a syntax perspective, yes. From a semantics perspective, not necessarily.

> An implementation MUST NOT ignore an extension that is marked critical and MUST reject the VC if it does not understand the extension.

So, if we put this in the spec, you would be fine w/ the PR? I'm fine w/ putting this in the spec for now. 

Keep in mind that the statement would come out of the discussion on https://github.com/w3c/vc-data-model/issues/158, which is why I keep saying that that specific text has more to do with PR #158 than it does #159.

David-Chadwick:
@msporny Do you think that extensibility and criticality are separate issues? This comment of yours  "This PR has to do with adding a section on extensibility to the specification, not in specifying critical properties." indicates that you do. I think they are the same issue for the following rationale.
1. The standard defines the common basic terms that every implementation must understand.
2. The standard defines how the basic VC structure and content can be extended and also specifies how implementations must deal with these extensions.
3. Clearly a basic implementation will not understand any extension but this is not necessarily a fatal flaw since the standard will say how unimportant extensions can be safely ignored whilst other ones (the critical ones) must not be ignored. 
4. Thus every implementation, even the most basic one, knows whether it can legitimately accept an extended VC (one with no critical extensions) or must not accept it (one with critical extensions).
Thus criticality and extensibility are two sides of the same coin.

dlongley:
@David-Chadwick,

> "This PR has to do with adding a section on extensibility to the specification, not in specifying critical properties." indicates that you do.

I don't think this means that "critical properties" has nothing to do with extensibility, but rather the *PRs* are separable. Pulling in a PR to make progress does not mean that another PR won't update the same section to further fill it out. It means we get to break things down into smaller pieces and keep moving forward. There can be many different PRs for the same section of the spec. I think what @msporny is going for here is making progress and trying to keep conversations focused on more bite sized chunks because they are easier to deal with.

There *will be* an extensibility section in this spec because we are chartered to do exactly that. If we don't provide extensibility, we have failed. Getting that section started doesn't mean it's finished -- we can add to it and modify it as needed.

From my perspective, what was added in this PR is the core of how we've been saying extensibility will work since the inception of this group, but it wasn't written down explicitly in the spec (until now). In other words, it is "the current state" just transformed into spec text. Whether or not we add "critical properties" to the extensibility section is presently being debated. I believe that this is how they are "separate issues".

David-Chadwick:
On the issue of whether a verifier MUST or SHOULD reject an extended VC with properties the verifier does not understand, I have the following text to suggest

"If a verifier is validating a VC as part of an automatic process to access a protected resource then the verifier MUST reject any VC containing any property that it does not understand. If however a verifier is validating a VC as part of process that involves human or non-automatic VC validation, or does not give rise to any risk that the verifier is willing to accept, then a verifier MAY accept a VC containing properties it does not understand."

David-Chadwick:
OOPS, it should have said "or does not give rise to any risk that the verifier is NOT willing to accept" or perhaps "or only gives rise to a risk that the verifier is willing to accept" is better

PR #144, opened 2018-03-30
Modified the Terms

David-Chadwick:
This is item iv) from PR#141

PR #167, opened 2018-04-30
Modified the terms

David-Chadwick:


David-Chadwick:
This replaces #144 . It contains the same changes and has removed conflicts with the latest spec.

msporny:
LGTM, merging as the PR seems to reflect the consensus of the group.

PR #173, opened 2018-05-08
Add security section on token binding. Fixes #148.

### LINKED ISSUE #148: Need section on binding of profile to current user
We have no language specifying the binding of a presenter to a profile in a specific session. 

There is consensus that the holder is not necessarily the subject. They may have complicated relationships to subjects in the included credentials, but there is some notion that presenting a profile MEANS that the presenter is claiming some relationship to the credentials. 

Unfortunately, we have no language explaining that this is necessary for valid use of a profile.

There is no discussion about how a verifier can tell if the profile has been hijacked by a man-in-the-middle attack. Yes, the profile is signed. By someone claiming the assertions in the profile, e.g., "I am over 21" or "Maria Hernandez is my mother, here is my birth certificate. I am Manuel Hernandez." So we can accept that the controller of the keys is asserting those claims. But how do we know the controller of the keys is the end user of the current session, aka the Presenter?

This is why I have always felt that proving holder==subject is an unsolved problem. If it is our intention that the profile MUST be bound to a session identifier and that such binding MEANS the current end-user asserts the claims in the profile, we need to be explicit about that.

Specifying how such binding occurs is beyond the scope of the current specification. However, specifying whether or not such binding is necessary is.

msporny:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/173.html" title="Last updated on May 15, 2018, 9:47 PM GMT (58ac051)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/173/f3f8e01...58ac051.html" title="Last updated on May 15, 2018, 9:47 PM GMT (58ac051)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2361,7 +2361,9 @@ of a <a>verifiable presentation</a> and are not the target of a
 <a href="https://en.wikipedia.org/wiki/Man-in-the-middle_attack">
 man in the middle attack</a>. Any protocol that utilizes the
 Verifiable Credentials Data Model and requires protection against these
-sorts of attacks needs to perform some sort of token binding that ties
+sorts of attacks needs to perform some sort of token binding, such
+as using <a href="https://tools.ietf.org/html/draft-ietf-tokbind-protocol-18">
+The Token Binding Protocol v1.0</a>, that ties
 the request for a <a>verifiable presentation</a> with the response. Any
 protocol that does not perform token binding is susceptible to
 man in the middle attacks.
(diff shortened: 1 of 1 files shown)


msporny:
> perhaps a "for example" with a link to one or more of the IETF token binding drafts

Fixed in https://github.com/w3c/vc-data-model/pull/173/commits/58ac0510d8e3f2df0e92a5ca0a4694d6a3609334.

msporny:
Blocking on @jandrieu since this was his issue. Don't want to merge w/o him.

msporny:
Thanks @jandrieu, merged.

Issue #174, opened 2018-05-10
Various questions related to the use of the "type" property


vieillevigne:
I am trying to figure out the normative use of the property, “type”. From the draft of May 8 I observe the following:
 
- Every object at the top level (except for @context and verifiable profiles) has a “type” property
- Every subobject, except for subobjects which are the values of “claim” properties, have “type”properties.
- Apart from its occurrence in sections 7.2, 8.1 and 8.2 ‘type’ only appears in examples. The sections just mentioned also contain the only apparently normative occurrences of ‘type’.
- The identifier “Credential” appears on the right-hand side of a “type” property only in ‘[..]’.
- With two exceptions “Credential” always appears with another identifier containing ‘Credential’.
- The only occurrences of a multiplicity of types embraced by ‘[..]’ (presumably denoting lists not sets) lead with “Credential”.
- The type to the right of “Credential” always seems to be a specialization (subtype ?) of “Credential”.
- “Credential”, “DocumentVerfification” and “VerfiableProfile” occur in singleton lists.

My questions follow:
 
1. Except for the fact that the object values associated with “claim” properties do not themselves have “type”  properties, I’d really like to be able to say that “every object and subobject MUST have a “type”  property”. Why don’t, for example, the apparent claim objects have “type” properties?
2. I’d like to be able to say that lists of types denote increasing specialization of type. Why then are there singleton occurrences of types in lists – especially when there are lots of singleton occurrences of types not in lists? What use can I make of a type specifically called out in a specialization hierarchy (if that is what a list intends)? In the case of types not used in lists, how do I know that they are not parts of specialization hierarchies elsewhere? Indeed, how do I know that “Credential” does not appear in a specialization hierarchy? Under what circumstance does one use a list for a singleton?
3. I am particularly puzzled by the assertion at section 5.2 that “A credential MAY be made verifiable by adding the following property …”. This suggests that a credential may be made verifiable in some other way than providing a proof. It also suggests that a credential need not be verifiable at all. Moreover, the example which immediately follows does have a “proof” property and also has a “ProofOfAgeCredential” type. Why is the latter needed in the presence of the former? What are sufficient conditions to be a verifiable credential? What are necessary conditions to be a verifiable credential? Is it intended that there be credentials which are not verifiable?
[...]

PR #177, opened 2018-05-19
Add section on types. Related to #174.

msporny:
This PR adds a new section to the spec that tightens up the language and requirements around type information.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/177.html" title="Last updated on May 31, 2018, 2:03 PM GMT (8469ab0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/177/daed52d...8469ab0.html" title="Last updated on May 31, 2018, 2:03 PM GMT (8469ab0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -523,8 +523,13 @@ The Verifiable Credentials trust model is as follows:
 
       <ol>
         <li>
-The <a>verifier</a> trusts the <a>issuer</a> to issue the Verifiable Credential
-that it receives. This trust could be weakened depending upon the risk
+The <a>verifier</a> trusts the <a>issuer</a> to issue the <a>credential</a>
+that it receives. In order to establish this trust, a credential MUST
+either 1) include a <a href="#proofs-aka-signatures">proof</a> that establishes
+that the issuer generated the credential (it is a <a>verifiable credential</a>),
+or 2) have been transmitted in a way that clearly establishes that the
+issuer generated the credential and that the credential has not been tampered
+with in transit. This trust could be weakened depending upon the risk
 assessment of the verifier.
         </li>
         <li>
--- a/index.html
+++ b/index.html
@@ -721,8 +721,8 @@ information associated with the <var>claim</var> property became valid.
       <section>
         <h2>Proofs (aka Signatures)</h2>
         <p>
-A <a>credential</a> or a <a>presentation </a> MAY be made verifiable by
-adding the following <a>property</a>:
+In order for a <a>credential</a> or <a>presentation</a> to be made
+verifiable, the following property MUST be present:
         </p>
         <dl>
           <dt><var>proof</var></dt>
--- a/index.html
+++ b/index.html
@@ -566,6 +566,103 @@ such that market competition and customer choice is increased.
 
     <section>
       <h1>Basic Concepts</h1>
+
+      <section>
+        <h2>Types</h2>
+        <p>
+Type information may be expressed via the <a>type</a> property:
+        </p>
+        <dl>
+          <dt><dfn>type</dfn></dt>
+          <dd>
+The value of this property MUST ultimately be one or more URIs.
+If more than one URI is provided, the array MUST be interpreted as a set of
+values where order is not significant. Note that syntax such as JSON-LD terms
+MAY be used to ease developer usage. It is RECOMMENDED that dereferencing the
+URI results in a document containing machine-readable information about the
+type.
+          </dd>
+        </dl>
+
+        <pre class="example nohighlight" title="Usage of the type property">
+{
+  "id": "http://dmv.example.gov/credentials/3732",
+  <span class="highlight">"type": ["VerifiableCredential", "ProofOfAgeCredential"]</span>,
+  "claim": {
+    "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
+    "ageOver": 21
+  },
+  "proof": { ... }
+}
+        </pre>
+
+        <p>
+With respect to this specification, the following <a>type</a>s MUST be
+specified on the following objects:
+        </p>
+
+        <table class="simple">
+          <thead>
+            <tr>
+              <th>Concept</th>
+              <th>Type</th>
+            </tr>
+          </thead>
+
+          <tbody>
+            <tr>
+              <td>
+A <a href="#credentials">credential</a> containing a
+<a href="#proofs-aka-signatures">proof</a>
+              </td>
+              <td>
+<tt>"type": ["<code>VerifiableCredential</code>"]</tt>
+              </td>
+            </tr>
+
+            <tr>
+              <td>
+A <a href="#credentials">credential</a> NOT containing a
+<a href="#proofs-aka-signatures">proof</a>
+              </td>
+              <td>
+<tt>"type": ["<code>UnverifiableCredential</code>"]</tt>
+              </td>
+            </tr>
+
+            <tr>
+              <td>
+A <a href="#presentations">presentation</a> containing a
+<a href="#proofs-aka-signatures">proof</a>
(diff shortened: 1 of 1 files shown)


msporny:
Ping'ing @vieillevigne so he knows that this PR exists and is in process.

David-Chadwick:
In section 5.1 where you specify the various types that must be used, you should add the other types that are used in this specification, namely
When  it is credential status, a type of "CredentialStatus"
when it is a terms of use, a type of "TermsOfUse"
when it is evidence, a type of "Evidence"
and any other types that we may define

msporny:
@David-Chadwick wrote: 
> you should add the other types that are used in this specification

Yes, good point, I'll do that.

David-Chadwick:
I dont agree with the changes you added for the other types such as Terms of Use. They should have the specific type as stated in my comment above and not an example which you have used, as this can be any string. Otherwise if examples suffice, there is no need to define the type 'VerifiableCredential' as a simple example of 'Under21Credential' will suffice.

index.html
```diff
@@ -566,6 +566,130 @@ <h1>Trust Model</h1>
...
+
+      <section>
+        <h2>Types</h2>
+        <p>
+Type information may be expressed via the <a>type</a> property:
+        </p>
+        <dl>
+          <dt><dfn>type</dfn></dt>
+          <dd>
+The value of this property MUST ultimately be one or more URIs.
```
davidlehn: The "ultimately" language sounds odd as used here.  I'm guessing this is to say `type` is a URI after JSON-LD processing?  Maybe reword the below "Note" sentence and move it to be second sentence.  Something like "Note that JSON-LD terms that represent URIs MAY be used to ease developer usage."  I'm not sure that's quite right but might make this paragraph less confusing to people not familiar with how JSON-LD works.

msporny:
@David-Chadwick wrote:
> I dont agree with the changes you added for the other types such as Terms of Use. 

Those types are not defined in the VC spec and if we were to specify those types in the VC spec, there would be no logic one could write for a processor. For example, let's say we say that all `termsOfUse` property values should have `"type": "TermsOfUse"`... what code would we write to use that class? I assert that there is no code and thus the property is not useful.

> They should have the specific type as stated in my comment above and not an example which you have used, as this can be any string. Otherwise if examples suffice, there is no need to define the type 'VerifiableCredential' as a simple example of 'Under21Credential' will suffice.

No, the reason you have Verifiable Credential is because you can check to see if the `proof` field exists and contains a value that you understand as a verifier. So, you can write code against `VerifiableCredential`. You can't write code against "CredentialStatus", "TermsOfUse", or "Evidence" because we don't specify any required processing rules that are associated with those types.

index.html
msporny: Agreed, and fixed in https://github.com/w3c/vc-data-model/pull/177/commits/89a757368ab89847c5c61489b68a032c4ca5867e.

msporny:
Waiting on @David-Chadwick's response to pull in this PR.

msporny:
Implementing actions as a result of today's VCWG meeting: 

https://www.w3.org/2018/05/29-vcwg-minutes.html#ActionSummary

@David-Chadwick and @dlongley ... I think I've made the appropriate changes here https://github.com/w3c/vc-data-model/pull/177/commits/e0e094c5e330c68b672e49eaa84adf11a9477e95 and here https://github.com/w3c/vc-data-model/pull/177/commits/324edb732d1e99911af5bc34fafd9e21d76f68e0.

Please check to make sure I got everything.

index.html
```diff
@@ -605,8 +721,8 @@ <h2>Issuer</h2>
       <section>
         <h2>Proofs (aka Signatures)</h2>
         <p>
-A <a>credential</a> MAY be made verifiable by adding the following
-<a>property</a>:
+A <a>credential</a> or a <a>presentation </a> MAY be made verifiable by
+adding the following <a>property</a>:
```
dlongley: I recommend changing to:

"In order for a credential or presentation to be made verifiable, the following property MUST be added:"

index.html
```diff
@@ -566,6 +571,117 @@ <h1>Trust Model</h1>
...
+
+        <p class="issue" data-number="182">
+Whether or not <code>CredentialStatus</code>, <code>TermsOfUse</code>,
+or <code>Evidence</code> provide useful type information in addition to
+more specific types is currently a topic of active discussion.
+        </p>
+
+        <p>
+The type system for the Verifiable Credentials Data Model is the same as the
+one for [[JSON-LD]] and is covered in depth in that specification. All
```
David-Chadwick: I don't find this statement acceptable for three reasons:
1. It has already been said that it is not necessary to know JSON-LD to work with VCs, but this statement is making it mandatory to fully understand JSON-LD in order to understand the use of the "type" property in VCs.
2. Having taken a quick look at the JSON-LD spec I could not find the "type" property described anywhere, although "@type" occurred a copious number of times. Are they the same or different?
3. The text for "@type" says specifically it is non-normative, viz

5.4 Specifying the Type
This section is non-normative.

The type of a particular node can be specified using the @type keyword. In Linked Data, types are uniquely identified with an IRI.

My conclusion is that the VC spec needs to definitively say what the 'type' property is and how the 'type' property is used

index.html
```diff
@@ -566,6 +571,117 @@ <h1>Trust Model</h1>
...
+          <tbody>
+            <tr>
+              <td>
+A <a href="#credentials">credential</a> object
+              </td>
+              <td>
+<tt>"type": ["<code>VerifiableCredential</code>"]</tt>
+              </td>
+            </tr>
+
```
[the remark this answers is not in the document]
msporny: Fixed in 1bd1fd5eb634ea0082956e19268b3c33420b0ace.

I couldn't add it to VerifiablePresentation because there are no other types at present that are contemplated.

index.html
msporny: > It has already been said that it is not necessary to know JSON-LD to work with VCs, but this statement is making it mandatory to fully understand JSON-LD in order to understand the use of the "type" property in VCs.

You don't need to have a full understanding of JSON-LD to use `type` in the same way as we don't require that developers have a full understanding of cryptography to use Verifiable Credentials. Most developers cut/paste code, which is made evident through the proper use of JSON-LD via schema.org on hundreds of millions of Web domains.

It's been suggested that we need to talk about `type` in this specification, which we borrow from JSON-LD and is defined in the JSON-LD specification (again, used on hundreds of millions of sites, so we do have deployment experience with it -- people seem to be able to use the feature correctly). We either duplicate the text in the JSON-LD specification in this specification (which is a bad specification writing practice), or we refer to it (which is considered the best practice currently). I suggest we refer to it. If we don't do that, we'll have to re-invent the type system just for Verifiable Credentials (which would effectively end up being JSON-LD's type system). We should re-use specs where we can.

To this end, I have referred to the exact section that needs to be read (both the non-normative and normative sections) in 8469ab0c3054c386c624c586db20ab71a9e96b04.

> Having taken a quick look at the JSON-LD spec I could not find the "type" property described anywhere, although "@type" occurred a copious number of times. Are they the same or different?

Clarified in 8469ab0c3054c386c624c586db20ab71a9e96b04.

> The text for "@type" says specifically it is non-normative, viz

Clarified in 8469ab0c3054c386c624c586db20ab71a9e96b04.

msporny:
Merging as all comments have been responded to w/ changes/edits and there don't seem to be any remaining issues (that are not already being tracked in the issue tracker).

index.html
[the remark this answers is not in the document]
David-Chadwick: Yes fine, thanks

Issue #178, opened 2018-05-19
Credentials Must always be Verifiable


David-Chadwick:
A verifier must always be able to trust the credentials it receives. This is written into the trust model.

Trust in the credential can be established in two ways:
a) it can be done by using a trusted channel between the issuer and the verifier, and transferring an unsigned/not cryptographically protected credential along a trusted channel, or
b) it can use an untrusted channel and a credential digitally signed/proofed by the issuer.

I do not see how the VC data model, which requires the VC to be passed from the trusted issuer to an untrusted user, can then establish a trusted channel between the verifier and the untrusted user. Therefore it is essential for the VC data model, that all credentials have proofs and are verifiable. Otherwise you have to introduce the concept of trusted users, which currently we do not have.

msporny:
> Therefore it is essential for the VC data model, that all credentials have proofs and are verifiable. 

-1, there are use cases where you may just want to package up data for processing w/o a proof and store it somewhere public. For example, an IoT device w/o a public/private keypair may want to publish information w/o proofs. If we require proofs, we'd disallow this use case.

> Otherwise you have to introduce the concept of trusted users, which currently we do not have.

Trust can be established via a protocol. The specification is about data model.  We shouldn't unnecessarily limit what can be done with the data model without a very good reason for doing so. What's the reason for limiting the data model in the way that you describe?

> This is written into the trust model.

Sounds like the trust model needs to change.

David-Chadwick:
@msporny "There are use cases where you may just want to package up data for processing w/o a proof and store it somewhere public."
I don't accept your use case as currently stated. If information is published publicly without any proof, then anyone can publish the information and it is impossible to verify its provenance. So it should be out of scope of the VC work. 

One way of doing what you require is to have a trusted repository that verifies the information and the IoT submitter without using public keys. But in this case the repository becomes the issuer of a verifiable credential and it conforms to our data model.

David-Chadwick:
This issue is related to 'unverifiable credentials' in #175 and #179.

David-Chadwick:
@msporny . It does not matter whether the receiver authenticates the user/holder/sender or not, as this does not affect the trust model, nor whether the VC can be trusted. If the user is untrusted, then authenticating the user and knowing that I am connected to an untrusted user, will not help me to accept an unsigned and therefore untrusted VC from this user. I need to see a signature from the issuer to know that I can trust the VC. Therefore VCs always have to be verifiable.
[...]

PR #179, opened 2018-05-20
Responses to Allen Brown's Review Comments

msporny:
This PR is ready to be reviewed and is in response to 104 comments on the spec that Allen Brown made as a result of his review of the specification.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/179.html" title="Last updated on May 31, 2018, 2:10 PM GMT (acbf32c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/179/1843947...acbf32c.html" title="Last updated on May 31, 2018, 2:10 PM GMT (acbf32c)">Diff</a>


terms.html
```diff
@@ -92,13 +93,13 @@
   </dd>
   <dt><dfn data-lt="issuers|issuer's">issuer</dfn></dt>
   <dd>
-An <a>entity</a> that asserts <a>claims</a> about <a>subject</a>, 
-creates a <a>verifiable credential</a> from these claims,
+A roles where an <a>entity</a> asserts <a>claims</a> about one or more
```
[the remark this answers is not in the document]
dlongley: None of these role definitions seem to be ... definitive. Rather, they seem to describe one thing that is happening when the role is being played by an entity (i.e. "a role where such and such is going on" ... but what else? anything? is "such and such" about the role or does it "such and such" just always happen to coincide with when the role is being performed?).

I think this could be improved by changing:

"A role where an entity asserts claims..."

To:

"A role an entity may perform by asserting claims..."

This makes it clearer that when an entity is "asserting claims...", it is performing the role of "issuer".

terms.html
```diff
@@ -72,7 +73,7 @@
   </dd>
   <dt><dfn data-lt="identifier registries">identifier registry</dfn></dt>
   <dd>
-Registries typically mediate the creation and verification of <a>subject</a>
+A roles where a system mediates the creation and verification of <a>subject</a>
```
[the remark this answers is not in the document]
msporny: Fixed.

terms.html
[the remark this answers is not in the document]
msporny: Changed all the definitions.

msporny:
This PR is ready to go, please review @stonematt @dlongley @David-Chadwick. I tried to only make editorial changes, although I can see why some might see some of the changes as not editorial (even though I think I can argue that they are).

index.html
```diff
@@ -196,20 +198,23 @@ <h3>What is a Verifiable Credential?</h3>
...
-distances are also possible through the Internet, making them more convenient
-when needing to establish trust over long distances.
+A verifiable credential is capable of representing all of the same information
+that a physical credential is intended to represent. The addition of
+digital technologies, such as digital signatures, also make verifiable
+credentials more tamperproof and therefore more trustworthy
+than their physical counterparts. Exchanging verifiable credentials rapidly,
+and where all parties are not physically co-located is made possible through
+the Internet, making them more convenient when needing to establish trust
+at a distance.
```
dlongley: Simplify last sentence ("Exchanging verifiable...") to:

"Verifiable credentials can also be rapidly exchanged through the Internet, making them more convenient than their physical counterparts when needing to establish trust at a distance."

index.html
```diff
@@ -566,8 +587,134 @@ <h1>Trust Model</h1>
...
+        <h2>Types</h2>
+
+        <p>
+Type information may be expressed via the <a>type</a> property:
+        </p>
+        <dl>
+          <dt><dfn>type</dfn></dt>
+          <dd>
+The value of this property MUST be one or more URIs.
+If more than one URI is provided, the array MUST be interpreted as a set of
```
dlongley: I recommend changing "If more than one URI ..." to

"If more than one URI is provided, the URIs must be understood be an unordered set. If a syntax, such as JSON-LD, uses an array to represent the set of URIs, the order of the values is not significant. Note that syntax..."

index.html
```diff
@@ -566,8 +587,134 @@ <h1>Trust Model</h1>
...
+              </td>
+            </tr>
+
+            <tr>
+              <td>
+A <a href="#credentials">credential</a> object NOT containing a
+<a href="#proofs-aka-signatures">proof</a>
+              </td>
+              <td>
+<tt>"type": ["<code>UnverifiableCredential</code>"]</tt>
```
dlongley: I'm not a fan of this name but I don't have an alternative. This name really only exists because we had to adopt "VerifiableCredential" rather than "Credential". Otherwise a "Verifiable Credential" would just be a "Credential" with a "proof" on it -- with no need to distinguish its type in this way. You can make a "Credential" verifiable by simply attaching a proof to it. As it stands now, to make an "UnverifiableCredential" verifiable, you'd have to *remove* its type, add a type of "VerifiableCredential" and add a proof.

This seems odd to me but I have nothing actionable to say that I think would be accepted. Likely unacceptable options are:

1. Rename "VerifiableCredential" to "Credential" and just use that type as I mentioned above.
2. Rename "UnverifiableCredential" to "Credential".
3. Remove type of "UnverifiableCredential" and leave it untyped.
4. Change "UnverifiableCredential" to "Assertion" and introduce further vernacular/confusion.

Option 4 would be to say that a simple "Assertion" stops being a mere assertion and becomes a "VerifiableCredential" once you attach proof to it. However, that, I believe this may further confuse the meaning of "proof". "Proof" does not prove that some statements are true, but rather the authenticity of the issuer.

That means that what we're really talking about here is more like "Gossip" or "Hearsay" or "Allegation". I *heard* that so and so said X about Y. I don't have any proof that they did.

index.html
```diff
@@ -566,8 +587,134 @@ <h1>Trust Model</h1>
...
+              </td>
+            </tr>
+
+            <tr>
+              <td>
+A <a href="#presentations">presentation</a> object NOT containing a
+<a href="#proofs-aka-signatures">proof</a>
+              </td>
+              <td>
+<tt>"type": ["<code>UnverifiablePresentation</code>"]</tt>
```
dlongley: Again, I'm not a fan of this type. How can we make such a claim that it is unverifiable? All we're saying here is that there's no proof attached. Perhaps "Presentation" is better? That doesn't seem to have the same problems that "Credential" has above.

index.html
```diff
@@ -1013,21 +1165,31 @@ <h2>Terms of Use</h2>
...
 
-
       <section>
         <h2>Evidence</h2>
-        <p>Evidence information for the credential in the
-        Verifiable Credentials Model may be provided by adding the
-        following <a>property</a>:
+
+        <p>
+The <em>evidence</em> property is used by an <a>issuer</a> to describe the
```
dlongley: "describe" => "represent"?

index.html
```diff
@@ -1013,21 +1165,31 @@ <h2>Terms of Use</h2>
...
+set of evidence that was used when determining whether or not to issue
```
dlongley: "when determining" or "to determine"?

index.html
```diff
@@ -1059,24 +1220,35 @@ <h2>Evidence</h2>
...
         </pre>
 
+        <p class="note">
+For the avoidance of doubt, the <em>evidence</em> <a>property</a> provides
+information that is different from and complementary to the
+<a href="#proofs-aka-signatures">proof</a> property. The <em>evidence</em>
+property is used to express supporting information, such as documentary
+evidence, related to the integrity of the credential. In contrast, the
+<em>proof</em> property is used to express machine-verifiable mathematical
+proofs related to the integrity of the credential.
```
dlongley: Not just the integrity of the credential but it is also used to prove the authenticity of the credential's issuer.

index.html
David-Chadwick: I do not support introducing 'unverifiable credential'. As pointed out in #178, credentials MUST always be verifiable since users are not trusted (regardless of whether they have been authenticated or not.)

index.html
David-Chadwick: Ditto above for this.

index.html
msporny: Fixed in bc39f5560c280d971c8deb91d291d8822d92fab5.

index.html
msporny: Fixed in PR #177.

index.html
msporny: Fixed in 782accd34b749f362518646d8d392fdd5eee1b23.

index.html
msporny: Fixed in dc3b3b19cc8ad888d81bedcb932130ffe38825d6.

index.html
msporny: Fixed in dc3b3b19cc8ad888d81bedcb932130ffe38825d6.

index.html
msporny: Fixed in dc3b3b19cc8ad888d81bedcb932130ffe38825d6.

msporny:
Made all of the changes requested by @dlongley and @David-Chadwick... please re-review so I can pull the PR in.

David-Chadwick:
It is rather difficult to comment on individual edits in this large PR,
but the one that I am currently not happy with is the one entitled "Fix
language around type property." The text is still too vague in my
opinion, and does not answer my question in #182, viz, what is the
actual purpose of the type property.

index.html
```diff
@@ -1059,24 +1213,35 @@ <h2>Evidence</h2>
...
         </pre>
 
+        <p class="note">
+For the avoidance of doubt, the <em>evidence</em> <a>property</a> provides
+information that is different from and complementary to the
+<a href="#proofs-aka-signatures">proof</a> property. The <em>evidence</em>
+property is used to express supporting information, such as documentary
+evidence, related to the integrity of the credential. In contrast, the
+<em>proof</em> property is used to express machine-verifiable mathematical
+proofs related to the authenticity of the isuer and integrity of the credential.
```
dlongley: typo "isuer" => "issuer"

index.html
```diff
@@ -1114,51 +1279,60 @@ <h1>Verification</h1>
...
-          <li>Recent metadata about the <code>issuer</code> which was published
-          by the issuer MUST be available.</li>
+          <li>The value associated with the <code>issuer</code>
+          property MUST identify an <a>issuer</a> that is known to and trusted
+          by the <a>verifier</a>.</li>
+          <li>Pertinent metadata about the <code>issuer</code> MUST be
+          available to the <a>verifier</a>. For example, an issuer may publish
+          information that contains the public keys they use to digitally
+          sign <a>verifiable credentials</a> that they have issued. This
+          metadata is pertinent when checking the signatures on the
```
dlongley: "signatures" => "proofs"

index.html
```diff
@@ -1214,459 +1400,168 @@ <h3>Revocation</h3>
...
+          <li>Any policy information expressed by the <a>issuer</a>
+          in the <a>verifiable credential</a> SHOULD be enforced. For example,
+          the issuer may limit the use of the claims to
+          specific <a>verifiers</a>, certain <a>holder</a> age ranges,
+          or during certain dates.</li>
+          <li>Any policy information expressed by the <a>holder</a>
+          in the <a>verifiable presentation</a> SHOULD be enforced. For example,
+          the holder may limit the use of the claims to
+          specific <a>verifiers</a> or for specific purposes
+          (such as authentication, but not data mining).</li>
```
dlongley: I can't remember if we decided to reword these "SHOULD" and "MUST" statements because they weren't testable or not. Just making a note here in case this was unintentional and needs to be addressed.

David-Chadwick: They should be MUSTs, since they must be enforced by all standard's conforming verifiers. The issuer or holder place policy information in the VC or VP precisely to tell the verifier what it must do.
Note that this does not stop a verifier's implementation ignoring all policy fields and doing what it wants. But in this case it is non-conformant to the standard, and therefore can do what it likes. But the standard does not condone this nor support it.

index.html
msporny: Fixed in 94904807a4f23cc7922d44de9f9c6719eddb2b75.

index.html
msporny: Fixed in 94904807a4f23cc7922d44de9f9c6719eddb2b75.

index.html
msporny: Fixed in 2110618f0e7454bc6357f46942872dc316425c3a.

David-Chadwick:
@msporny . Ok pull in this PR. The concept I am still unhappy with is the squishy nature of the type property. Specifically the current text says " All credentials SHOULD specify additional, more narrow types (e.g. ProofOfAgeCredential) such that software systems can use the additional information to more easily process the credential. " This leaves far too much wiggle room.
In my opinion the above SHOULD should be a MUST, so that the verifier definitively knows what contents to expect in the VC. Otherwise we are back to the long debated issue of what does a verifier do with properties it does not understand or know how to process. Is the VC acceptable or not?
This same comment applies to all the objects with types that are embedded in VCs. The types should be there to tell the verifier exactly what it is receiving.

msporny:
@David-Chadwick wrote:
> In my opinion the above SHOULD should be a MUST, so that the verifier definitively knows what contents to expect in the VC. Otherwise we are back to the long debated issue of what does a verifier do with properties it does not understand or know how to process. Is the VC acceptable or not?
> This same comment applies to all the objects with types that are embedded in VCs. The types should be there to tell the verifier exactly what it is receiving.

Fixed in https://github.com/w3c/vc-data-model/commit/1843947670b930b8f0f71f8c7fb54688b5d1e16e#diff-eacf331f0ffc35d4b482f1d15a887d3bR677.

msporny:
Merging as I've responded to all comments and remaining issues are tracked elsewhere.

PR #185, opened 2018-06-12
Minor changes to disputed credential data model. Fixes #180.

### LINKED ISSUE #180: Disputed credential section is confusing
https://github.com/w3c/vc-data-model/blob/gh-pages/index.html#L1064-L1089

A type of "DisputedCredential" is a statement about the credential doing the disputing, so this name seems wrong. Also, it may be that the IDs appear in the wrong place in the example.

msporny:
See if this addresses both of your concerns @lovesh and @dlongley.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/185.html" title="Last updated on Jun 13, 2018, 12:03 AM GMT (a8804e6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/185/ba43ef7...a8804e6.html" title="Last updated on Jun 13, 2018, 12:03 AM GMT (a8804e6)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1243,23 +1243,29 @@ credential.
         <h2>Disputes</h2>
 
         <p>
-The time may come when an <a>entity</a> wants dispute a <a>credential</a>
+The time may come when an <a>entity</a> wants to dispute a <a>credential</a>
 issued by another entity. The mechanism for doing this is the same as issuing
 a regular credential except that the <a>subject</a> identifier for the
 <a>claims</a> are those of the disputed credential. For example, if a disputed
-credential with an identifier of <code>https://example.org/credentials/245</code>
+credential with an identifier of
+<code>http://con-artist.example.com/credentials/3732</code>
 contains disputed statements, an entity would issue the following credential
 in a public venue to make it known that the credential is disputed:
         </p>
 
-        <pre class="example nohighlight" title="Expressing a disputed credential">
+        <pre class="example nohighlight"
+          title="Expressing a disputed credential">
 {
   "id": "http://example.com/credentials/245",
-  "type": ["VerifiableCredential", "DisputedCredential"],
+  "type": ["VerifiableCredential", "DisputeCredential"],
   <span class="highlight">"claim": {
     "id": "http://con-artist.example.com/credentials/3732",
     "currentStatus": "Disputed",
-    "statusReason": "Credential contains disputed statements"
+    "statusReason": "Credential contains disputed statements",
+    "disputedClaim": {
+      "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
+      "ageOver": 21
+    }
   }</span>,
   "issuer": "https://example.com/people#me",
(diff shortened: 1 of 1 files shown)


msporny:
Ping @lovesh and @dlongley ... I'm going to merge this and move on in 3 days if I don't hear from either of you.

David-Chadwick:
There is another issue with this section, namely who is entitled to issue a "DisputeCredential". We must consider the case of an attacker issuing a false dispute credential. If we regard a "DisputeCredential" as being similar to a revoked credential, then if anyone is allowed to issue a DisputeCredential we have the opportunity for a DOS attack, or at least  the opportunity to give the verifier sufficient cause for concern to temporarily suspend access to its resources for the credential owner. I would therefore like to see a statement along the lines of:

Only the subject of a verifiable credential, or its authorised agent, is entitled to issue a "DisputeCredential". A "DisputeCredential" issued by anyone other than the subject, or its authorised agent, SHOULD be disregarded by the verifier, unless the verifier has some out of band means of ascertaining the truth of the dispute.

msporny:
Made your suggested change in b392070ec79516bd530c00a89b8202dc8c63488e, @David-Chadwick.

jandrieu:
@msporny I hate to say this, but this adjustment confuses me. The example leaves out the information needed to understand facts of the dispute.

Who is disputing the claim? Is it **https://example.com/people#me**? Or is the issuer simply reporting the fact of the dispute. If so, how do they know there is a dispute? Unless  **https://example.com/people#me** is directly disputing the credential (which I think is @David-Chadwick's intention), it seems like there should be more data involved. 

This may be as trivial as changing the "isDisputed" predicate, which is passive voice to the equivalent RDF statement 
```
https://example.com/people#me disputes http://con-artist.example.com/credentials/3732 . 
```

Or 
```
http://con-artist.example.com/credentials/3732 isDisputedBy https://example.com/people#me . 
```

Stylistically, I prefer the former's active voice over the latter's passive, but the "isDisputedBy" would be closer to the JSON-LD as suggested.

David-Chadwick:
It would also help if the opening sentence was changed from

The time may come when an entity wants to dispute a credential issued by another entity.

to

The time may come when a subject wants to dispute a credential.

Note that we do not need 'by another entity' since a self issued VC can always be revoked by the issuer.

Also can I suggest we change

an entity would issue the following credential in a public venue to make it known that the credential is disputed:

to 

the subject (or its authorised agent) of the disputed credential would issue the following credential in a public venue to make it known that the credential is disputed:

Then of course the example also needs updating as suggested by @msporny

jandrieu:
One minor nit:
> the subject (or its authorised agent) of the disputed credential would issue the following credential in a public venue to make it known that the credential is disputed:
Might be better as
> the subject (or its authorised agent) of the disputed credential would issue the following credential, for distribution to those they wish to notify of the dispute.

Issue #108, opened 2018-02-06
Typo in Example 6


David-Chadwick:
The Example 6  title and description are inconsistent
The title says

Usage of termsOfUse property by an Issuer

but the text below states
In the example above, the holder is prohibiting a verifier from using the information provided to correlate the holder.

David-Chadwick:
#170 fixes this

msporny:
Resolved in https://github.com/w3c/vc-data-model/commit/f96e446802928bb7d80516d5ef23da8415c56406 ... closing issue.

Issue #164, opened 2018-04-24
Issuers may express a Time to Live on the credential


stonematt:
@David-Chadwick your comment in issue #135 about having to be online to verify the referenced credential is a larger challenge in general.  VC should be verifiable w/out going to the source in general.  This protects the VC from becoming useless if the issuer site goes offline.  If Issuers care to limit how long a credential can be trusted without live verification from the source, the Issuer must include something like a Time To Live in the credential to indicate how long it's valid before the data is considered stale. --- TTL is different than the expiration of the credential itself.  This has better privacy protection than a revocation list, and is less exposed to DDOS attacks.

Consider: 

- Andy's driver's license expires in 2 years and has a VC that shows the expiration date in 2020.
- When the state cuts the VC, they put at 60day TTL on the credential, with a "refresh URL"  
- When the verifier receives the VC, as long as the signatures validate and it's within 60d, the credential is considered valid
- If the verifier receives a VC that was created more than 60d ago, the verifier should ignore it and ask for a refresh.
- Later, in our protocol, the Holder should have a way to "refresh" the TTL and re-share, also the ability to "refresh" could be a capability that's granted by the Holder to the Verifier.  
-- Consider a Professional (Holder) giving a license (VC) to an Employer (Verifier) and granting them the permission to verify license status on demand, as long as the professional continues employment.

David-Chadwick:
@stonematt  I agree that the issue of remote retrieval is a much larger problem in general than fetching a referenced credential. It covers all new information that extends the base standard for VCs and is defined somewhere else online. Your idea for a TTL property is a good one, and whilst it does not address the general issue of remote retrieval, it does partially cover the revocation/staleness issue. However, what to do after the TTL expires is a separate issue, and can be solved in a variety of ways such as: reissuing the credential (short TTL and non-revocable), periodically publishing a CRL at least every TTL, credential status stapling (cf OCSP stapling). So I think we could add a TTL property independent of the credential status method.

msporny:
Time to Live is probably a bad name for this feature... I think what we're really going for is something that already exists in HTTP - the Cache Control header:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

Let's not re-invent the wheel and just re-use the header name and the acceptable values of the field as they seem to map cleanly to the use case presented above.

[...]

PR #188, opened 2018-06-19
Terms of use

David-Chadwick:
Rebased original PR#170 and addressed additional comments made to it


stonematt:
Fixes #108

stonematt:
@David-Chadwick  both of these examples are prohibitions.  Can we include an example that grants usage.  I'm thinking this might be a course to solve the time to live issue #164 

The issuer could issue this VC for use until dateTime:x at which time the verifier must update from the renewal service?

David-Chadwick:
@stonematt I think you might find that all terms of use are actually prohibitions, as they are there to limit the use of the VC in some respect. So  useUntil is a prohibition of the VC being available forever.

David-Chadwick:
@msporny @dlongley . What is holding up your review?

dlongley:
> What is holding up your review?

Vacation! :smile:

PR #189, opened 2018-06-24
Attempt to fix #182 by providing type expectations.

### LINKED ISSUE #182: Are CredentialStatus, TermsOfUse, and Evidence useful types?
There are PRs pending that establish `CredentialStatus`, `TermsOfUse`, and `Evidence` as types in the data model. It is argued that these types 1) do not result in any code being written that utilizes them, and 2) are already encoded in the `credentialStatus`, `termsOfUse`, and `evidence` properties.

msporny:
Attempt to address #182 by providing type expectations for encapsulating objects and associated objects.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/189.html" title="Last updated on Jul 8, 2018, 6:01 PM GMT (1188ba5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/189/ba43ef7...1188ba5.html" title="Last updated on Jul 8, 2018, 6:01 PM GMT (1188ba5)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -695,15 +695,20 @@ the additional information to more easily process the credential.
         </p>
 
         <p>
-Encapsulating objects, such as <a>credential</a>, MAY convey the types
-of associated objects so that the <a>verifier</a> can quickly determine
-the contents of the associated object based on the type of the encapsulating
-object. For example, a <a>credential</a> with the additional <code>type</code>
-of <code>ProofOfAgeCredential</code> would signal to the <a>verifier</a> that
-the object associated with the <a>claim</a> property will contain the
-identifier for the <a>subject</a> in the <code>id</code> property and the age
-assertion in the <code>ageOver</code> property. This enables implementers to
-rely on the <code>type</code> field for verification purposes.
+When processing encapsulated objects in this specification, (e.g. objects
+associated with the <code>claim</code> property or deeply nested therein),
+a software system SHOULD use type information specified in encapsulating
+objects higher in the hierarchy. For the avoidance of doubt, an
+encapsulating object such as <a>credential</a>, SHOULD convey the types of
+associated objects so that the <a>verifier</a> can quickly determine the
+contents of the associated object based on the type of the encapsulating object.
+To provide a concrete example, a <a>credential</a> with the additional
+<code>type</code> of <code>ProofOfAgeCredential</code> would signal to the
+<a>verifier</a> that the object associated with the <a>claim</a> property will
+contain the identifier for the <a>subject</a> in the <code>id</code> property
+and the age assertion in the <code>ageOver</code> property. This enables
+implementers to rely on values associated with the <code>type</code> property
+for verification purposes.
         </p>
 
         <p class="note">
--- a/index.html
+++ b/index.html
@@ -704,12 +704,6 @@ assertion in the <code>ageOver</code> property. This enables implementers to
 rely on the <code>type</code> field for verification purposes.
         </p>
 
-        <p class="issue" data-number="182">
-Whether or not <code>CredentialStatus</code>, <code>TermsOfUse</code>,
-or <code>Evidence</code> provide useful type information in addition to
-more specific types is currently a topic of active discussion.
-        </p>
-
         <p class="note">
 The type system for the Verifiable Credentials Data Model is the same as the
 one for [[JSON-LD]] and is detailed in
--- a/index.html
+++ b/index.html
@@ -693,10 +693,15 @@ the additional information to more easily process the credential.
         </p>
 
         <p>
-Objects expressed in a <a>credential</a> MUST provide type information so
-that credential processing software can determine the sort of properties
-to expect in the object unless the type is not known or not including the
-type would not result in ambiguity when processing the credential.
+Encapsulating objects, such as <a>credential</a>, MAY convey the types
+of associated objects so that the <a>verifier</a> can quickly determine
+the contents of the associated object based on the type of the encapsulating
+object. For example, a <a>credential</a> with the additional <code>type</code>
+of <code>ProofOfAgeCredential</code> would signal to the <a>verifier</a> that
+the object associated with the <a>claim</a> property will contain the
+identifier for the <a>subject</a> in the <code>id</code> property and the age
+assertion in the <code>ageOver</code> property. This enables implementers to
+rely on the <code>type</code> field for verification purposes.
         </p>
 
         <p class="issue" data-number="182">
--- a/index.html
+++ b/index.html
@@ -652,7 +652,9 @@ a more specific credential type (e.g. <code>ProofOfAgeCredential</code>)
 A <a href="#presentations">presentation</a> object
               </td>
               <td>
-<tt>"type": ["<code>VerifiablePresentation</code>"]</tt>
+<tt>"type": ["<code>VerifiablePresentation</code>"]</tt> and optionally a
+more specific presentation type
+(e.g. <code>CredentialManagerPresentation</code>).
               </td>
             </tr>
 
--- a/index.html
+++ b/index.html
@@ -661,8 +661,7 @@ A <a href="#presentations">presentation</a> object
 A <a href="#status">credentialStatus</a> object
               </td>
               <td>
-<tt>"type": ["<code>CredentialStatus</code>"]</tt> and
-a valid credential status type (e.g. <code>CredentialStatusList2017</code>)
+A valid credential status type (e.g. <code>CredentialStatusList2017</code>)
               </td>
             </tr>
 
@@ -671,8 +670,7 @@ a valid credential status type (e.g. <code>CredentialStatusList2017</code>)
 A <a href="#terms-of-use">termsOfUse</a> object
               </td>
               <td>
-<tt>"type": ["<code>TermsOfUse</code>"]</tt> and
-a valid terms of use type (e.g. <code>OdrlPolicy2017</code>)
+A valid terms of use type (e.g. <code>OdrlPolicy2017</code>)
               </td>
             </tr>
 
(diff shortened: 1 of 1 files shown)


David-Chadwick:
Unfortunately this text is still not sufficient. The offending text is "unless the type is not known or not including the type would not result in ambiguity when processing the credential. " Firstly I cannot think of an example where omitting the type would not result in ambiguity. Secondly I cannot think why the type of the VC would not be known by the issuer. 
Furthermore the extensibility section is currently erroneous as this only allows for extensibility if JSON-LD is used. It does not show types being added when new properties are added to a VC. See issue #190.

msporny:
> Firstly I cannot think of an example where omitting the type would not result in ambiguity. 

Here's one:

```
{
  type: ['VerifiableCredential', 'ProofOfAgeCredential'],
  ...
  claim: {
    'id': 'did:example:1234',
    'ageOver': 21
  }
}
```

The object under 'claim' does not have a type, but the purpose (and use) of the credential is clear.

msporny:
> Secondly I cannot think why the type of the VC would not be known by the issuer.

Measurement scenarios would result in the type not being known by the issuer. For example, an IoT device that is measuring temperature, that is given an identifier, would not know if the temperature is associated with a location, person, ocean, lake, river, produce, etc. It can make a claim about the measurement without knowing the type of entity that it is measuring.

msporny:
> Furthermore the extensibility section is currently erroneous as this only allows for extensibility if JSON-LD is used. It does not show types being added when new properties are added to a VC. See issue #190.

That discussion is being tracked by a separate issue (and is a separate issue). This PR is not mean to address #190.

David-Chadwick:
@msporny " For example, an IoT device that is measuring temperature, that is given an identifier, would not know if the temperature is associated with a location, person, ocean, lake, river, produce, etc." This is a spurious example. The VC would be a "Temperature" VC, and the claim would contain the ID of the entity and the property would be temperature of the entity. The VC does not need to say what type of entity this is. Another VC that links the ID to the type of entity would be created by the issuer that knows what type of entity it is.

msporny:
> The VC would be a "Temperature" VC, and the claim would contain the ID of the entity and the property would be temperature of the entity. The VC does not need to say what type of entity this is. Another VC that links the ID to the type of entity would be created by the issuer that knows what type of entity it is.

If this is true, then we're saying exactly the same thing. I think I know where the disconnect is now. You are saying "The Verifiable Credential object should always have two types: VerifiableCredential and then a more specific type for the exact type of verifiable credential". If you agree to that, then this PR accomplishes that.

msporny:
> You do not appear to want to have strongly typed VCs, and want to give issuers the flexibility to put anything in a VC without indicating to the verifier what is in the VC, other than by the verifier having to parse through it and understanding each property once they encounter it. 

That's definitely not my position (nor anyone elses AFAICT).

msporny:
> I want the type values to tell the verifier what they can expect to find in a VC

To be very pedantic, you mean "what they can expect to find in the top-level of a VC and a good chunk of the information they can expect to find in the claims object and its sub objects".

This is an open world data model, we can't define everything that may appear, only some of the things that may appear. There are many use cases where things may be included by reference or included by a different system.

> and what will not be in the VC, before having to parse every property.

We can't specify what will not be in a VC without some incredibly heavy implementation requirements that will most likely result in the failure of the specification. It's not very useful to take an open world data model and then restrict what can be placed in it to only things defined in machine-readable specifications.

msporny:
> The VC would be a "Temperature" VC, and the claim would contain the ID of the entity and the property would be temperature of the entity. The VC does not need to say what type of entity this is. Another VC that links the ID to the type of entity would be created by the issuer that knows what type of entity it is.

If this is true, then we (and the spec text) are saying exactly the same thing. 

I think I know where the disconnect is now. 

You are saying "The Verifiable Credential object should always have two types: VerifiableCredential and then a more specific type for the exact type of verifiable credential". If you agree to that, then this PR (in fact the one before it) accomplishes that.

Look at the table here: https://w3c.github.io/vc-data-model/#types note that it says that "A credential object" ... MUST contain ""type": ["VerifiableCredential"] and a more specific credential type (e.g. ProofOfAgeCredential)". Why does that not address where the type information should be?

In addition to that, I thought you were requesting that *every object*, no matter how deeply nested, MUST contain a *type*... and that is what this PR does. The only issue is that both you and I agree that the object *associated with* "claim" doesn't need to have type because the "TemperatureCredential" gives you enough context to make the VC unambiguous.

Given the above, does this PR resolve #182?

David-Chadwick:
If you read #182  you will see that I have proposed text for how it should be resolved. So can you please address #182 in #182 and comment upon my proposed text.
Many thanks

msporny:
> So can you please address #182 in #182 and comment upon my proposed text.

Done and the PR has been updated to use your approach (with a few editorial changes that tighten up the concept). 

Do these changes resolve #182?

David-Chadwick:
@msporny . Thankyou. This is much better. 

However, the note on extensibility viz: "While application developers and document authors do not need to understand the specifics of JSON-LD's type system, implementers of this specification that want to support extensibility in an interoperable fashion do." is still contentious. 

I think there might be some confusion about what interoperable means. We have at least 3 different scenarios:
1. The issuer issues a VC in pure JSON-LD that a verifier who only understands pure JSON cannot process. But it can be extended in ways that JSON-LD verifiers can understand
2. The issuer issues a VC in pure JSON that a verifier who only understands pure JSON-LD cannot process. But it can be extended in ways that pure JSON verifiers can understand
3. The issuer issues a VC in a hybrid JSON-LD that a verifier who only understands pure JSON can process. But it cannot be extended in ways that a pure JSON verifier can understand
4. Any other variant?
Which are we trying to address?

msporny:
> I think there might be some confusion about what interoperable means

I'm using the concept of "interoperability" that is defined in the W3C Process document:  https://www.w3.org/2018/Process-20180201/#implementation-experience

and discussed here: https://www.w3.org/blog/2008/05/open-standards-interoperability/

msporny:
> However, the note on extensibility viz: "While application developers and document authors do not need to understand the specifics of JSON-LD's type system, implementers of this specification that want to support extensibility in an interoperable fashion do." is still contentious.

That text was not introduced in this PR, merely reformatted. That text was introduced here:

https://github.com/w3c/vc-data-model/commit/1843947670b930b8f0f71f8c7fb54688b5d1e16e#diff-eacf331f0ffc35d4b482f1d15a887d3bR691

If you have concerns over that text, please raise it in a new PR so we can discuss it separately.

msporny:
Just to reiterate, Issue #182 asked the question: "Are CredentialStatus, TermsOfUse, and Evidence useful types?"... which is answered by this PR: No, they are not useful... but more specific types are and they MUST be provided in implementations of this specification.

Can we merge this particular PR @David-Chadwick and discuss the further questions you raised above in a new PR?

David-Chadwick:
Let me try to clear what the issue is from a modelling perspective.  Before we had the discussion on types we had an ambiguous situation, where some objects contained their types, other objects did not contain any type information, and some encapsulating objects contained the type(s) of an encapsulated object. This was very confusing to me when I tried to work out what the precise rules for type information were (other than, anything goes!). I think we have now agreed that the rule is:

Either each object MUST contain its type(s) in the type property, or an encapsulating object MAY contain the values of the type property on behalf of the encapsulated object. The inheritance  of the type property from an encapsulated  object to the encapsulating object MAY be interatively repeated up to the outer object i.e. verifiable credential or verifiable presentation.

Is this the rule or not? Currently the text does not quite say this. It implies that type inheritance must be done from a claim object. The text also says "Type information may be expressed via the type property. This is hardly mandating it.

So please can we agree on how type information is specified in the data model and then state unambiguously and succinctly what the rule is.

David-Chadwick:
@msporny re interoperability. I understand what W3C means by this. So in fact, it would be perfectly feasible for the data model document to specify two different and non-interoperable transfer encodings for the same data model, and to have 4 implementations, 2 of which implement each transfer encoding and interoperate perfectly well with the other. This would be sufficient for the data model document to progress towards a standard, would it not?

msporny:
> This would be sufficient for the data model document to progress towards a standard, would it not?

Yes, although I think both of us would agree that that is not the ideal scenario.

msporny:
> Either each object MUST contain its type(s) in the type property, or an encapsulating object MAY contain the values of the type property on behalf of the encapsulated object. The inheritance of the type property from an encapsulated object to the encapsulating object MAY be interatively repeated up to the outer object i.e. verifiable credential or verifiable presentation.

I have added additional text (with editorial changes) to what you propose above into the document here:

https://github.com/w3c/vc-data-model/pull/189/commits/f8dc869387814db02a04a98b6edd44b3efb611b7

Does this resolve your concerns? Can we merge this PR now?

David-Chadwick:
@msporny Your text does not convey exactly the same semantics as my text. My text is specific and generic and applies to the types of all embedded objects. Your text is skewed towards credentials and claims, and leaves unanswered questions e.g. if a TermsOfUse type is specified in the type of the credential, should it also occur in the TermsOfUse object as well?

burnburn:
@David-Chadwick In the interest of moving forward, the chairs would like to suggest that we merge the current PR that addresses the text that exists in the spec and modify it as appropriate if/when we complete the TermsOfUse discussion and its impact on the spec.

msporny:
> In the interest of moving forward, the chairs would like to suggest that we merge the current PR that addresses the text that exists in the spec and modify it as appropriate if/when we complete the TermsOfUse discussion and its impact on the spec.

Acknowledged, merging in the next edit cycle.

msporny:
> @msporny Your text does not convey exactly the same semantics as my text. My text is specific and generic and applies to the types of all embedded objects. Your text is skewed towards credentials and claims, and leaves unanswered questions e.g. if a TermsOfUse type is specified in the type of the credential, should it also occur in the TermsOfUse object as well?

Acknowledged, let me try and address that in the next edit cycle and submit a separate PR for that.

msporny:
@David-Chadwick I made one last attempt to address your concerns by strengthening and generalizing the language here: https://github.com/w3c/vc-data-model/pull/189/commits/1188ba51f1815427b1ce78f4e280b044b863a5e3

Merging... if you still don't like it, raise a new issue.

PR #191, opened 2018-06-26
Fix #149 by changing the purpose of the id property.

### LINKED ISSUE #149: Need to restrict top level data types to those specified in the standard
The current data model does not place any restrictions on the top level data types/structures. The type parameter can have any value. Consequently anyone can define their own type (e.g. unverifiableDocument). This loophole should be closed. It is proposed to close it by introducing two standard data types: Verifiable Credential and Verifiable Profile (or just one, depending upon whether profiles are discontinued or not). This would change the current JSON structure to 
```
"verifiableCredential": {
     "id": "< credential id >",
     "type": ["ProofOfAgeCredential"]
```

msporny:
Fix #149 by changing the purpose of the id property. Specifically, the presentation id was the DID before, and now it is the id of the presentation.

### DIFF
--- a/index.html
+++ b/index.html
@@ -890,7 +890,7 @@ is provided below:
 
         <pre class="example nohighlight" title="Basic structure of a presentation">
 {
-  "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
+  "id": "urn:uuid:3978344f-8596-4c3a-a978-8fcaba3903c5",
   "type": ["VerifiablePresentation"],
   <span class="highlight">"verifiableCredential": [{ ... }],
   "proof": [{ ... }]</span>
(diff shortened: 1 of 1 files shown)


msporny:
I think this addresses a long-standing issue that @jandrieu raised at the 2018 VCWG F2F... that presentations are their own object.

David-Chadwick:
+1

index.html
```diff
@@ -890,7 +890,7 @@ <h2>Presentations</h2>
 
         <pre class="example nohighlight" title="Basic structure of a presentation">
 {
-  "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
+  "id": "urn:uuid:3978344f-8596-4c3a-a978-8fcaba3903c5",
```
dlongley: I guess in the case of DID Auth via the Credential Handler API you would now either look for the DID in a `PublicKeyCredential` within `verifiableCredential` ... or would just look in the proof and use the owner of the key that provided the signature? Just making sure that this still works for use cases for which we have already written initial implementations.

msporny: Yes, you could use either approach. You have to look in a different location and reason about the data differently, but the original use case is still supported AFAICT.

PR #196, opened 2018-07-08
Add processing restrictions to enable semantic interop. Fixes #183.

### LINKED ISSUE #183: Value of Type Property
The Type property must be a URI according to the data model specification, viz : "The value of this property MUST be one or more URIs."

Implementors of VCs do not need to know JSON-LD, and indeed there is a section that says how to encode VCs in pure JSON.

However, all the examples of the <code>type</code> property in the data model spec use strings such as "VerifiableCredential" as examples of type values. Presumably some JSON-LD context turns these strings into URIs. However, these strings will not be interpreted as URIs by implementations that use pure JSON, as they are unaware of JSON-LD contexts. 

So am I correct in saying that a JSON implementation will reject these VCs, and therefore there cannot be interworking between JSON-LD and JSON implementations?

msporny:
Add a section on Semantic Interoperability to ensure that JSON-based processors don't have to process extensions as JSON-LD in order to be conformant processors.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/196.html" title="Last updated on Jul 10, 2018, 9:36 PM GMT (0edbe7d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/196/b41e408...0edbe7d.html" title="Last updated on Jul 10, 2018, 9:36 PM GMT (0edbe7d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -642,6 +642,7 @@ the type.
 
         <pre class="example nohighlight" title="Usage of the type property">
 {
+  "@context": "https://w3id.org/credentials/v1",
   "id": "http://dmv.example.gov/credentials/3732",
   <span class="highlight">"type": ["VerifiableCredential", "ProofOfAgeCredential"]</span>,
   "claim": {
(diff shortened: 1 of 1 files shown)


msporny:
Seeing no objections, I'm going to merge this PR by the end of today.

index.html
```diff
@@ -1156,15 +1161,46 @@ <h2>Extensibility</h2>
...
+when a JSON-LD Context redefines any term in the
+<a href="https://www.w3.org/TR/json-ld/#dfn-active-context">active context</a>.
+            </li>
+          </ul>
+
+          <p>
+To avoid the possibility of accidentally overriding terms, developers are urged
+to scope their extensions. For example, the following extension scopes the
+new <code>favoriteFood</code> term so that it may only be used within the
+<code>claim</code> property:
```
dlongley: This may count as a "redefinition" of the `claim` term. I think we should still merge this PR as it gives us a starting point for tweaking these rules -- but we should take note of this.

PR #198, opened 2018-07-09
Subjectnotequal holder

David-Chadwick:
I have re-based PR #169 to the latest changes in github, and removed the relationship object. I do believe that it has now addressed all the issues that have been raised.

***
<a href="https://pr-preview.s3.amazonaws.com/David-Chadwick/vc-data-model/pull/198.html" title="Last updated on Aug 6, 2018, 5:15 PM GMT (e7c8d81)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/198/ba43ef7...David-Chadwick:e7c8d81.html" title="Last updated on Aug 6, 2018, 5:15 PM GMT (e7c8d81)">Diff</a>


index.html
```diff
@@ -1238,6 +1238,369 @@ <h2>Evidence</h2>
...
+
+{
+	"id": "http://dmv.example.gov/credentials/3732",
+	"type": ["Credential", "ProofOfAgeCredential"],
+	"issuer": "https://dmv.example.gov/issuers/14",
+	"issued": "2010-01-01",
+	"claim": {
+		"id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
+		"name": "Mr John Doe",
+		"address": "10 Some Street, Anytown, ThisLocal, Country X",
```
msporny: I suggest using schema.org's address property: https://schema.org/address

index.html
```diff
@@ -1238,6 +1238,369 @@ <h2>Evidence</h2>
...
+an aspect of the identity of the subject, and which together unambiguously identify
+the subject. In this case, it does not matter who the holder is, since the verifier
+can unambiguously identify the subject of the verifiable credential.
+   </p>
+
+        <pre class="example nohighlight" title="An example of a credential 
+that uniquely identifies a subject">
+
+{
+	"id": "http://dmv.example.gov/credentials/3732",
```
msporny: Missing context, I suggest:

```"@context": ["https://w3id.org/credentials/v1", "https://schema.org/"]```

File: index.html [hunk 3]
```diff
@@ -1238,6 +1238,369 @@ <h2>Evidence</h2>
...
+	"type": ["Credential", "ProofOfAgeCredential"],
+	"issuer": "https://dmv.example.gov/issuers/14",
+	"issued": "2010-01-01",
+	"claim": {
+		"id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
+		"name": "Mr John Doe",
+		"address": "10 Some Street, Anytown, ThisLocal, Country X",
+		"dateOfBirth": "2001-01-01",
+		"gender": "male",
```
msporny: Suggest https://schema.org/GenderType is used... specifically `"Male"`.

index.html
```diff
@@ -1238,6 +1238,369 @@ <h2>Evidence</h2>
...
+		"placeOfBirth": "Anytown, ThisLocal, Country X",
```
msporny: Try to use real data and https://schema.org/address

index.html
```diff
@@ -1238,6 +1238,369 @@ <h2>Evidence</h2>
...
+		"nationality": "X",
```
msporny: Use real data here... GB, FR, etc.

index.html
```diff
@@ -1238,6 +1238,369 @@ <h2>Evidence</h2>
...
+
+    <p>
+This process of recursion could continue without restriction. Therefore 
+a new property “delegatable”, is added to the data model.
+
+     </p>
+
+   <section>
+
+   <h3> Delegatable Property </h3>
```
msporny: This is stomping on what we expect OCAPs to be used for... don't know how I feel about this other than it makes me uneasy. Can you at least put a <p class="issue"> beside it stating that the WG has not determined if this is the proper approach yet? I don't want this to keep this PR from getting into the spec.

index.html
```diff
@@ -1238,6 +1238,369 @@ <h2>Evidence</h2>
...
+        </pre>
+
+       </section>
+       
+       <section>
+
+     <h3> Delegation through Recursion </h3>
+
+    <p>
+With recursion, the subject issues a credential to the delegated holder in which:
```
msporny: This creates confused deputy attacks, doesn't it? I'm struggling to understand what the use case is here other than ones we know we don't want "Here, use my access badge"?

index.html
```diff
@@ -1238,6 +1238,369 @@ <h2>Evidence</h2>
...
+
+{
+	"id": "1234",
+	"type": ["Credential", "EmbeddedDelegatedCredential"],
+	"issuer": "did:example:ebfeb1f712ebc6f1c276e12ec21",
+	"issued": "2010-01-06T00:00:00Z ",
+	"expires": "2010-01-08T00:00:00Z ",
+	"claim": {
+		"id": "did:example:ebfeb1f712ebc6f1c2123456789",
+		"credential": {
```
msporny: This property is now "verifiableCredential"

index.html
```diff
@@ -1238,6 +1238,369 @@ <h2>Evidence</h2>
...
+			"id": "http://example.gov/credentials/3732",
+			"type": ["Credential", "Role"],
```
msporny: Should probably be RoleCredential... even though this is getting into OCAP territory... this feels like the wrong way to model this... we need a different use case here.

index.html
```diff
@@ -1238,6 +1238,369 @@ <h2>Evidence</h2>
...
+			"issuer": "https://dmv.example.gov",
+			"issued": "2010-01-01",
+			"claim": {
+				"id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
+				"role": "Project Manager"
```
msporny: Why `role` and not `type`? ... and wouldn't you want to also associate their role w/ an organization and department? Again, this is why we shouldn't be expressing this sort of access control stuff as a VC... it's too easy to mess up.

index.html
```diff
@@ -1238,6 +1238,369 @@ <h2>Evidence</h2>
...
+			},
+			"revocation": {
+				"id": "http://example.gov/revocations/738",
+				"type": "SimpleRevocationList2017"
+			},
+			"proof": {....}
+		},
+		{
+			"id": "https://some.dns.com/VC/123456789",
```
msporny: W3C specs are required to use example.com and example.org in examples. dns.com is a real website.

index.html
```diff
@@ -1238,6 +1238,369 @@ <h2>Evidence</h2>
...
+  </section>
+
+
+
+   
+   <section>
+      <h2>Holder Acts on Behalf of Subject</h2>
+
+      <p>
+When the holder wants to act on behalf of the subject, the issuer MAY indicate this by
```
msporny: Yes, this is good.

index.html
```diff
@@ -1238,6 +1238,369 @@ <h2>Evidence</h2>
...
+either including the relationship of the holder in the subject's <a>claim</a>, 
+or by issuing a separate relationship credential to the holder. 
+Either way, the relationship describes the relationship between the holder and the subject of the claim. </p>
+
+<p>It is outside
+the scope of this standard how the issuer verifies the subject, the holder and the
+relationship between them.
+
+<p>
```
msporny: You need a `class="issue"` on the `p` tag.

index.html
```diff
@@ -1238,6 +1238,369 @@ <h2>Evidence</h2>
...
+
+{
+	"id": "http://dmv.example.gov/credentials/3732",
+	"type": ["Credential", "ProofOfAgeCredential"],
+	"issuer": "https://dmv.example.gov/issuers/14",
+	"issued": "2010-01-01T19:73:24Z",
+	"claim": {
+		"id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
+		"ageUnder": 16,
+		"parent": {
```
msporny: Yes, this is the correct way to express a relationship between two entities.

index.html
```diff
@@ -1238,6 +1238,369 @@ <h2>Evidence</h2>
...
+a parent so that the parent can hold their child's verifiable credentials">
+
+{
+	"id": "http://dmv.example.gov/credentials/3732",
+	"type": ["Credential", "RelationshipCredential"],
+	"issuer": "https://dmv.example.gov/issuers/14",
+	"issued": "2010-01-01T19:73:24Z",
+	"claim": {
+		"id": "did:example:ebfeb1c276e12ec211f712ebc6f",
+		"child": {
```
msporny: +1, this is good.

index.html
```diff
@@ -1238,6 +1238,369 @@ <h2>Evidence</h2>
...
+		"created": "2017-06-17T10:03:48Z",
+		"creator": "https://dmv.example.gov/issuers/14/keys/234",
+		"nonce": "d61c4599-0cc2-4479-9efc-c63add3a43b2",
+		"signatureValue": "pY9...Cky6Ed = "
+	}
+}
+
+</pre>
+     
+<p> It can easily be seen from the above example how an authoritative issuer could indicate that a person is the owner of a pet dog </p>
```
msporny: While it may be self-explanatory, I feel like this may be the central use case: guardianship (of people, pets, and things).

The delegation bit seems like the wrong way to do things...

index.html
```diff
@@ -1238,6 +1238,369 @@ <h2>Evidence</h2>
...
+
+    </section>
+
+  <section>
+      <h2>Issuer Authorises Holder</h2>
+
+      <p>
+When the issuer wishes to authorise a holder to possess a credential that
```
msporny: I don't understand the use case here. Why does the law enforcement officer have this credential in the first place? It's not explained in the text.

index.html
```diff
@@ -1238,6 +1238,369 @@ <h2>Evidence</h2>
         </p>
 
       </section>
+<section>
+
+      <h1>Subject Not Equal To Holder</h1>
```
msporny: We need a less computer-science-y way to describe these scenarios.

index.html
```diff
@@ -1238,6 +1238,369 @@ <h2>Evidence</h2>
...
+        </pre>
+
+
+   </section>
+
+    <section>
+       <h2>Subject Delegates to Holder</h2>
+
+     <p>
+This is supported in the data model in two different ways: through the use of either recursion or verifiable presentations. Both methods involve the subject issuing a new verifiable credential to the delegated holder.
```
msporny: Why doesn't the subject just issue an ocap to the holder for a very specific grant of authority. Delegating a credential is dangerous because you're not also delegating what that delegation can be used for. For example, if I delegate a driver's license to you... does that mean you can use it to open a bank account online? or open a line of credit? Or get into a bar? How are the limitations for that grant put into place? This is why we have OCAPs.

index.html
David-Chadwick: We have had long discussions about OCAPs vs VCs and it is clear that there are separate use cases and roles for both of these.
Role delegation in organisations is very common - so common in fact that most organisations could not function without the ability to delegate roles. This is why the example is the delegation of a role.
RBAC is also very common in organisations, and role delegation is needed in this.

index.html
David-Chadwick: Because the property that is being delegated is the role.
Yes you could also associate organisation if needed. But typically the organisation will be the initial issuer who will issue roles to all its employees. And the organisation will be the root of trust for the verifier. So once an employee gets a role VC (issued by the organisation) and is allowed to delegate it to another employee, we do not really need the organisation name to be repeated in the delegated VC.

index.html
David-Chadwick: OOPS. I have omitted the figure that goes at the start of the section. I will need to add it.
(But if you look at the figure in the previous PR, which shows all the cases where Subject NE Holder, you will see that Law Enforcement is one of the examples where a holder may be given a VC by an issuer without the subject being aware of this, or not requesting it.)

index.html
David-Chadwick: People do not delegate driving licenses, or age credentials. These will clearly be marked as not delegatable (in fact this should be the default. The issuer should clearly mark a VC as delegatable if it wants it to be)

msporny: I'd argue that people don't delegate roles either. They grant permission to operate on their behalf, given their authority as a particular role in an organization. I don't say "here, have my key card so you can do anything I can" (ambient authority)... I say "here, have my key card so you can do X, and only X". That's an OCAP.

index.html
msporny: > Role delegation in organisations is very common

I'd argue that role delegation in organizations, while common, is executed in very different ways... some of them, dangerously. I'd also argue that we need to be crystal clear about what "role delegation" means... because in may cases, it means "I'm granting permission for person X to do action Y for time period Z." ... but in reality, what usually happens is "I'm enabling person X to be me without restriction".

... and that's what it looks like you're modeling here, which is why I'm objecting to the data model enabling that scenario. It's the text book definition of ambient authority - https://en.wikipedia.org/wiki/Ambient_authority

index.html
msporny: I understand and now am more strongly against the proposal. This will lead to ambient authority attacks and we must not suggest that as a standard operating practice in the core specification: https://en.wikipedia.org/wiki/Ambient_authority

msporny: I suggest that we do not support this use case in this version of the VC spec, noting that we may address it via OCAPs or in the next version of the specification and noting that delegation is very easy to get wrong and that extensions should be wary of any sort of delegation mechanism.

index.html
David-Chadwick: I guess we just have to disagree over this point. Role delegation is very common in RBAC systems, and VCs are a new way of handing out roles to employees. (Our existing software, PERMIS (www.openpermis.org) uses X.509 ACs to hold roles. It would be nice to move to VCs).

index.html
David-Chadwick: Look at this from the perspective of the verifier. The verifier is given a VC (for subject A) by holder B. The verifier needs to know (at least):
a) why has B got hold of this VC?
b) how is B related to A?
c) does B have A's permission to handle A's VC?
d) who to trust?
The answer to d) is the issuer. This is part of our trust model. Now if the issuer inserts into the VC that B is the holder of the VC and is related to A in the following way, then the verifier has all the above questions answered for him. This is why it is beneficial for some of the use cases for the issuer to insert information about the holder into the VC.

David-Chadwick: Alternatively, if B is not related to A, then the verifier needs to know (at least)
a) why has B got hold of this VC?
b) how is B related to the issuer?
c) does B have the issuer's permission to handle A's VC?
d) who to trust.
Again, the verifier trusts the isssuer, and the issuer is saying that it gave the VC to the holder for it to present to verifiers, and that the holder is of the following type (which indicates the relationship between the issuer and the holder). So all the verifier's questions are answered.

@msporny. If the VC does not indicate who the holder is, and the verifier has to rely on the presentation (signed by the holder) to verify this, then I assert that this is not possible....
unless either:
a) the verifier modifies its trust policy (e.g. I trust X to hold this set of VCs) or
b) the holder presents some other credential proving that they have the right to hold the VCs of strangers.
I think what you are proposing unnecessarily complicates the work of either or both the verifier and the holder (and possibly the issuer as well).

index.html
David-Chadwick: No I do not believe that delegation of an organisational role from one person to another person creates a confused deputy, as the delegate inherits the permissions of the delegator, which is what is intended.
Could you please explain why you think the delegate might be a confused deputy?

index.html
David-Chadwick: Two points
1. By not mentioning delegation in the spec does not stop it from happening, but rather, allows unrestricted delegation to happen. This is what we learnt from v1 X.509, where anyone with an X.509 PKC could issue another PKC and thereby become a de-facto CA. We had to insert the basic constraints extension into v3 X.509 to stop this from happening. The Delegatable property that I am suggesting is equivalent to the X.509 basic constraints extension.
2. Role delegation does not introduce vulnerabilities of either the confused deputy or ambient authority types because the delegation of all privileges assigned to the role is intended. To quote Wikipedia  confused deputy is when "the associated permission changes unintentionally".

David-Chadwick:
Concerning #203

The Subject NE Holder text on "Holder Acts on Behalf of Subject" goes some way to preventing the issue of an unreliable verifier forwarding the VC when it should not, because it addresses the case when the holder is authorised by the issuer to be the holder. Clearly the verifier is not the original holder, nor is it the subject, so VC forwarding is made more difficult in this case.

index.html
cwebber: It does create a confused deputy problem because you're authorizing based on identity, and if you're authorizing based on identity you're holding onto a larger set of credentials about "who you are" rather than "what you're doing".  It's not that the inheriting of permissions is *unintended* in the confused deputy... a confused deputy is usually behaving as designed, and that is part of the point of the confused deputy problem.  I encourage reading the [original confused deputy example](http://www.cap-lore.com/CapTheory/ConfusedDeputy.html).  The problem is that the program holding that authority cannot distinguish the *purpose* for which it should be using its wide range of authority.  This is why an identity-centric approach to authority will always lead to confused deputies... not because they are misprogrammed, but because the problem is inherent in that design.

index.html
cwebber: > d) who to trust?
> The answer to d) is the issuer. This is part of our trust model. 

Is it?  I don't understand this statement.  Anyone can say anything about anyone... VCs just give us a way to prove who's saying it.  Just because Bob issues a claim of foo about Carol in this VC doesn't mean that I trust that Bob's claim is true... I just know that Bob said it.  Deciding who I trust is a different step, internal to the program using the VC tooling, not part of the VC workflow itself, right?

David-Chadwick: Please read Section 4, point 1
"The verifier trusts the issuer to issue the credential that it receives. "
If the verifier does not trust the issuer it will most likely discard the VC.

index.html
David-Chadwick: @cwebber I beg to disagree with you. If I intend someone to take over my role, say because I am on holiday or ill for a period of time, then I do want them to inherit ALL my privileges. The purpose is clearly to inherit all my authority, otherwise they could not take on my role in its entirety. Most organisations today operated in this way. They cannot function without delegation.

msporny: @David-Chadwick, I think there are three statements you're making here that we need to separate:

1. Organizations need to delegate authority.

I think everyone agrees on that.

2. The way organizations delegate authority are Verifiable Credentials.

We do not have consensus on that and I really doubt we'll get there because this creates confused deputy and ambient authority issues. We have at least @cwebber, @dlongley, and myself in agreement on that (having put a tremendous amount of thought into this).

3. You want to be able to grant ALL of your privileges to an entity.

This is the definition of confused deputy AND ambient authority and the academic and corporate literature on why this is a terrible security design paradigm goes back to the 1970s.

At this point, I suggest that we wipe the section you've authored on delegation, put an issue marker in there pointing to an issue on delegation when subject != holder, and sort the solution out in that issue.

I don't want to delay getting this PR merged due to this one issue, which doesn't seem like it's going to be resolved any time soon.

dlongley: I want to clarify -- the ability to granting all privileges to another entity is not the confused deputy problem/ambient authority, rather, a system where there is only a binary choice of delegating all privileges (*for all targets/verifiers*) or no delegation at all leads straight to it.

People should be able to delegate authority. The system that enables it should allow for granularity from zero delegation to full (and everything in between). Furthermore, authority should be *targeted*.

The VC model is about identity not authority. It is also designed, primarily, to address use cases where a VC can be issued by one party and verified *by many different parties*. There is not a single target. That is not an appropriate design for decentralized authorization and delegation. But you can build something appropriate on top of it -- that uses identity as a starting place to instantiate rights/privileges. This is what OCAP-LD is for.

David-Chadwick: @dlongley. Thankyou for your clarification of confused deputy, which is very helpful to this discussion. I will go further than your clarification however, and say that confused deputy is not in essence to do with delegating authority, nor whether there is a binary choice in the delegation, but rather, it is when one party asks another party (the deputy) to perform an action that the deputy is authorised to perform but the requestor is not. In RBAC, the process of role delegation is separate from the process of requesting an action, so providing the process for requesting an action ensures that the requestor has full authority for the requested action, then the confused deputy problem cannot happen. Furthermore hierarchical RBAC allows for fine grained delegation of authority by delegating a subordinate role to the one possessed, so your issue of binary choice is not relevant here. Since the text on delegation is only concerned with delegation and not with requesting actions, then I do not see that the confused deputy problem is relevant to the proposed text.

David-Chadwick:
The latest version brings the text (and diagram) in line with the VC WG minutes of 31 July 2018

index.html
```diff
@@ -1238,6 +1238,318 @@ <h2>Evidence</h2>
...
+	"type": ["Credential", "IdentityCredential"],
+	"issuer": "https://dmv.example.gov/issuers/14",
+	"issued": "2010-01-01",
+	"claim": {
+		"id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
+		"name": "Mr John Doe",
+		"address": {
+                   "streetAddress": "10 Rue de Chose",
+                   "postalCode": "98052",
+                   "addressLocality": "Paris, France"
```
msporny: This isn't quite correct, but we can fix it after we pull the PR in as it's editorial.

index.html
```diff
@@ -1238,6 +1238,318 @@ <h2>Evidence</h2>
...
+  "id": "http://dmv.example.gov/credentials/3732",
+  "type": ["VerifiableCredential", "ProofOfAgeCredential"],
+  "issuer": "https://dmv.example.gov/issuers/14",
+  "issued": "2010-01-01T19:73:24Z",
+  "claim": {
+    "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
+    "ageOver": 21
+  },
+  <span class="highlight">"termsOfUse": [{
+    "type": "SubjectOnly",
```
msporny: I'm fine w/ pulling this PR in, even though I disagree with this way of modelling "Subject Only". I think we need to clean up the data model for "presenter must be subject" terms of use... or "presenter may differ from subject" terms of use. There are security attack vectors that we need to think through wrt. what the default should be wrt. allowing the presenter to not be the subject.

index.html
```diff
@@ -1238,6 +1238,318 @@ <h2>Evidence</h2>
...
+  <section>
+   <h3> Passing on a Verifiable Credential </h3>
+
+   <p>
+
+When the subject passes a verifiable credential to a
+holder, the subject SHOULD issue a new verifiable credential to the holder in which: 
+the issuer is the subject, 
+the subject is the holder to whom the verifiable credential is being passed, 
+and the claim contains the properties that are being passed on. In addition, the holder creates a verifiable presentation that contains these two
```
msporny: "the claim contains the properties that are being passed on" creates an enormous amount of complexity when processing credentials that are handed off... We should probably make it an all or nothing thing for v1.0.

Again, fine with pulling this in as long as we can change this later.

index.html
```diff
@@ -1238,6 +1238,318 @@ <h2>Evidence</h2>
...
+			"proof": {....}
+		},
+		{
+			"id": "https://example.com/VC/123456789",
+			"type": ["Credential", "PrescriptionCredential"],
+			"issuer": "did:example:ebfeb1f712ebc6f1c276e12ec21",
+			"issued": "2010-01-03",
+			"claim": {
+				"id": "did:example:76e12ec21ebhyu1f712ebc6f1z2",
+				"prescription": {....}
```
msporny: This is the wrong way to model this... again, happy to pull the PR in as long as we can clean this up after we pull the PR in.

index.html
```diff
@@ -1238,6 +1238,318 @@ <h2>Evidence</h2>
...
+    }]
+  }</span>,
+  "proof": { ... }
+}
+        </pre>
+
+       </section>
+       
+  <section>
+   <h3> Passing on a Verifiable Credential </h3>
```
msporny: Upon further review, we need to remove this section or not support it... stating "I'm letting person X use attributes Y and Z of my credential" is the wrong way to solve the "delegation" problem.

msporny:
Pulling this PR in and making the changes based on the conversation in the WG last week.

PR #199, opened 2018-07-10
Add section on caching credentials and refreshService. Fixes #164. Fixes #139.

### LINKED ISSUE #139: How are verifiable credentials updated
Observing a scenario where a person's driver's license is required to be updated.. in the traditional non-digital model the card holder would go to the local government office and request a new driver's licence with the new address. In many cases the old license is destroyed either by the person.. or requested by the office and discarded before the new license is given out.. 

How does this scenario translate into the VC world?

Is it possible to overwrite an existing VC? or do both old and new VCs co-exist and are differentiated by their version field.. and the identity owner has the option to use the appropriate one depending on the situation?

The issuer can invalidate the old VC by invalidating the signature.. however the can the issuer request the identity owner  to delete the old VC upon receiving the new one? perhaps through ODRL?

### LINKED ISSUE #164: Issuers may express a Time to Live on the credential

msporny:
This is an attempt to propose how we do credential caching and credential refreshing.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/199.html" title="Last updated on Jul 17, 2018, 2:29 PM GMT (fa18583)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/199/db4abd5...fa18583.html" title="Last updated on Jul 17, 2018, 2:29 PM GMT (fa18583)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1244,9 +1244,7 @@ the Cache-Control header in HTTP [[RFC7234]] Section
           <dt><var>cacheRefreshService</var></dt>
           <dd>
 The value of this property MUST be a URL. The URL MUST be constructed such that
-an HTTP GET with a query parameter of <code>id</code> and the value of the
-credential `id` property appended will result in an updated credential or an
-error.
+an HTTP GET on it will result in an updated credential or an error.
           </dd>
         </dl>
 
@@ -1258,7 +1256,7 @@ error.
   "issuer": "https://example.com/issuers/14",
   "issuanceDate": "2018-02-24T05:28:04Z",
   <span class="highlight">"cacheControl": "public, must-revalidate, max-age=5184000"</span>,
-  <span class="highlight">"cacheRefreshService": "https://example.org/credentials"</span>,
+  <span class="highlight">"cacheRefreshService": "https://example.org/credentials?id=3fgJ3k"</span>,
   "claim": {
     "id": "did:example:abcdef1234567",
     "name": "Jane Doe"
@@ -1268,7 +1266,7 @@ error.
         </pre>
 
         <p>
-It is strongly RECOMMENDED that the `cacheRefreshService` URL authenticates
+It is RECOMMENDED that the `cacheRefreshService` URL authenticates
 the caller and SHOULD only allow the holder of the original credential to
 retrieve a refreshed credential in order to mitigate privacy violations.
         </p>
--- a/index.html
+++ b/index.html
@@ -1279,6 +1279,12 @@ their operational needs.
         </p>
 
         <p>
+Any relative time value (e.g. <em>delta-seconds</em>), such as values associated
+with <code>max-age</code>, specified in the <code>cacheControl</code> field
+MUST be specified as being relative to the <code>issuanceDate</code>.
+        </p>
+
+        <p>
 If used, HTTP Cache-Control Header values MUST be at least as restrictive
 (i.e. cause the cached value to invalidate) as <code>cacheControl</code> values
 specified in an HTTP message (e.g. <a>presentation</a>). For example, the value
--- a/index.html
+++ b/index.html
@@ -1218,6 +1218,70 @@ new <code>favoriteFood</code> term so that it may only be used within the
 
       </section>
 
+
+      <section>
+        <h2>Caching</h2>
+
+        <p>
+<a>Issuers</a> may wish to signal information related to caching a
+<a>credential</a> to <a>holders</a> and <a>verifiers</a>. Some issuers may
+assert that they do not want a credential to be cached, while others may
+specify certain caching rules and refresh time periods that are acceptable.
+        </p>
+
+        <p>
+Caching information for the credential in the Verifiable Credentials Data
+Model is provided by adding the following <a>properties</a>:
+        </p>
+
+        <dl>
+          <dt><var>cacheControl</var></dt>
+          <dd>
+The value of this property MUST be one or more of the acceptable values for
+the Cache-Control header in HTTP [[RFC7234]] Section
+<a href="https://tools.ietf.org/html/rfc7234#section-5.2">5.2. Cache-Control</a>.
+          </dd>
+          <dt><var>cacheRefreshService</var></dt>
+          <dd>
+The value of this property MUST be a URL. The URL MUST be constructed such that
+an HTTP GET with a query parameter of <code>id</code> and the value of the
+credential `id` property appended will result in an updated credential or an
+error.
+          </dd>
+        </dl>
+
+        <pre class="example nohighlight" title="A simple credential with caching information">
+{
+  "@context": "https://w3id.org/credentials/v1",
+  "id": "http://example.com/credentials/4643",
+  "type": ["VerifiableCredential"],
+  "issuer": "https://example.com/issuers/14",
+  "issuanceDate": "2018-02-24T05:28:04Z",
+  <span class="highlight">"cacheControl": "public, must-revalidate, max-age=5184000"</span>,
+  <span class="highlight">"cacheRefreshService": "https://example.org/credentials"</span>,
+  "claim": {
+    "id": "did:example:abcdef1234567",
+    "name": "Jane Doe"
+  },
+  "proof": { ... }
+}
+        </pre>
+
+        <p>
+It is strongly RECOMMENDED that the `cacheRefreshService` URL authenticates
+the caller and SHOULD only allow the holder of the original credential to
+retrieve a refreshed credential in order to mitigate privacy violations.
+        </p>
+
+        <p>
+If a <code>cacheControl</code> value is not present, then a <a>verifier</a>
+MAY cache the data up to the <code>expirationDate</code>, while ensuring to
+process the credential <a href="#status">status</a> at an interval that meets
+their operational needs.
+        </p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1104,6 +1104,70 @@ <h2>Extensibility</h2>
...
+The value of this property MUST be one or more of the acceptable values for
+the Cache-Control header in HTTP [[RFC7234]] Section
+<a href="https://tools.ietf.org/html/rfc7234#section-5.2">5.2. Cache-Control</a>.
+          </dd>
+          <dt><var>cacheRefreshService</var></dt>
+          <dd>
+The value of this property MUST be a URL. The URL MUST be constructed such that
+an HTTP GET with a query parameter of <code>id</code> and the value of the
+credential `id` property appended will result in an updated credential or an
+error.
```
davidlehn: I think the spec should avoid arbitrary URL construction rules like this.  Another option is to use a [URI Template](https://tools.ietf.org/html/rfc6570) with an input of an `id`.  Perhaps an even better option is to just avoid such issues and let the credential specify the full refresh URL in whatever format is appropriate in a `cacheRefreshUrl` property or similar.

msporny: Ok, agreed. Changing it now.

index.html
```diff
@@ -1104,6 +1104,68 @@ <h2>Extensibility</h2>
...
+        </p>
+
+        <dl>
+          <dt><var>cacheControl</var></dt>
+          <dd>
+The value of this property MUST be one or more of the acceptable values for
+the Cache-Control header in HTTP [[RFC7234]] Section
+<a href="https://tools.ietf.org/html/rfc7234#section-5.2">5.2. Cache-Control</a>.
+          </dd>
+          <dt><var>cacheRefreshService</var></dt>
```
msporny: If we just change this to credentialRefreshService, then we could address #139 as well.

index.html
```diff
@@ -1188,6 +1188,68 @@ <h2>Extensibility</h2>
...
+    "id": "did:example:abcdef1234567",
+    "name": "Jane Doe"
+  },
+  "proof": { ... }
+}
+        </pre>
+
+        <p>
+It is RECOMMENDED that the `refreshService` URL authenticates the caller and 
+SHOULD only allow the holder of the original credential to retrieve a
```
davidlehn: Trailing whitespace on above two lines.

davidlehn:
Some `cacheControl` property issues:
- What are the guidelines for handling HTTP `Cache-Control` values along with credential `cacheControl` values?  What should clients do if they conflict?  Caches themselves will only look at the HTTP header so servers might want to send that.
- The section about `expirationDate` should probably be like the RFC Expires Section 5.3 and say `expirationDate` is ignored if `max-age` is present.  As written, any `cacheControl` header would cause `expirationDate` to be ignored.
- Various `Cache-Control` fields have `delta-relative` time values.  I assume they are relative to the `issuanceDate`?  That should be specified.
- `cacheControl` content potential changing on a `refreshService` call will require new proofs to be generated.  Are there issues with that?
- The text suggest only holder should use the `refreshService`.  It seems inefficient to be transferring around and storing that data if only a holder would use it.
- Does the `refreshService` URL leak info?  Probably no more than a HTTP id or issuer property?

David-Chadwick:
I agree that have a TTL or cache time is a good idea and helpful to the verifier.

But I do not agree with adding a Cache Refresh Service URL. Remember that the holder has already got a copy of the VC. The first time around he did not need a Cache Refresh URL, so there is no reason for the holder to need it for subsequent retrieves. If you want the verifier to be able to automatically refresh a VC then I suggest that this is an additional feature that has not had sufficient discussion and should not be in the first version of the standard. This sounds much more like an authorisation issue and should be the subject of another token that the holder gives to the verifier. By inserting the Refresh Service URL into the VC you are providing this to all verifiers, which is not something the holder may want.

Furthermore the wording is too vague. What are "one or more of the acceptable values". Which values are acceptable and which are not. I suggest that such vague and ambiguous wording has no place in a standard.

msporny:
@davidlehn wrote:
> What are the guidelines for handling HTTP Cache-Control values along with credential cacheControl values? 

I've added specific text here: a83ec030f1862b7dbf8bd1d58e55269a2c671827

> What should clients do if they conflict? Caches themselves will only look at the HTTP header so servers might want to send that.

Also fixed here: a83ec030f1862b7dbf8bd1d58e55269a2c671827

> The section about expirationDate should probably be like the RFC Expires Section 5.3 and say expirationDate is ignored if max-age is present. 

expirationDate is always checked and has nothing to do w/ caching. If a credential is past the expirationDate, it's expired.

> As written, any cacheControl header would cause expirationDate to be ignored.

I don't understand where you're getting that from?

> Various Cache-Control fields have delta-relative time values. I assume they are relative to the issuanceDate? That should be specified.

Fixed in: ef54e99f950ff778323c055e104590a910070c21

> cacheControl content potential changing on a refreshService call will require new proofs to be generated. Are there issues with that?

I don't think there are. If there are, we need a concrete example.

> The text suggest only holder should use the refreshService. It seems inefficient to be transferring around and storing that data if only a holder would use it.

If we have any other party than the holder (or an agent of a holder) do the refresh, there are negative privacy implications. So, at present, only the holder can use the refreshService.

> Does the refreshService URL leak info? Probably no more than a HTTP id or issuer property?

It does leak info, but no more than the issuer property if the URL is designed correctly (such as just being the @id for the issued credential... or a URL w/ a random string on the end). It also enables correlation, but again, no more than the digital signature on the credential itself.

msporny:
> But I do not agree with adding a Cache Refresh Service URL. Remember that the holder has already got a copy of the VC. The first time around he did not need a Cache Refresh URL, so there is no reason for the holder to need it for subsequent retrieves. If you want the verifier to be able to automatically refresh a VC then I suggest that this is an additional feature that has not had sufficient discussion and should not be in the first version of the standard. 

The first time they received the credential was a manual process and they may not know how to navigate back to that location (for example, getting a coupon can happen through a variety of different channels/sites). Some of those coupon issuers may want to provide a new coupon to people that have an expired coupon in their wallet in order to continue to drive the customer to buy their product. Asking the customer if they want to refresh their $1 coupon and then asking them to navigate back to the site that they got the coupon on, which they probably don't remember, would create friction such that the issuer of the coupon would lose their potential customer.

This is a real use case from the Web Commerce Interest Group and a real PoC/Pilot that is in the works.

> This sounds much more like an authorisation issue and should be the subject of another token that the holder gives to the verifier. By inserting the Refresh Service URL into the VC you are providing this to all verifiers, which is not something the holder may want.

-1 to this information being for the Verifier, as that can lead to privacy violations. This information is meant for the Holder's software (e.g. digital wallet) to use when it detects that a credential has expired.

> Furthermore the wording is too vague. What are "one or more of the acceptable values". Which values are acceptable and which are not. I suggest that such vague and ambiguous wording has no place in a standard.

The values are defined explicitly in HTTP RFC7234 Section <a href="https://tools.ietf.org/html/rfc7234#section-5.2">5.2. Cache-Control</a>... what language would you prefer?

msporny:
> I feel like we're jumping too far ahead of the problem and that's introducing complexities that may not need to spend time solving. Before going any further down the caching path I'd prefer that we very clearly delineate the use cases and requirements.

Agreed, the more I work on this PR, the more complex it's becoming. Implementing software to parse HTTP Cache-Control is non-trivial and I'm certain that implementers won't do a thorough job of it.

I do find value in the `refreshService` URL, which is something the Holder's software can use to auto-refresh credentials that have expired.

If we do support cache control moving forward, I think I'd rather just introduce a "cacheUntil" property that is specified in the presentation by the Holder (not the issuer)... but even then, I'm struggling to understand how this is that much of an improvement over `expirationDate` and `status`. That is, you cache until `expirationDate` and you always have to check `status`.

msporny:
I suggest that we close this PR and create a new one for `refreshService` URL.

msporny:
This PR is replaced by #210, which focuses on adding a refreshService feature.

PR #205, opened 2018-07-21
Update context and vocabulary.

### LINKED ISSUE #202: Align JSON-LD Context with latest Verifiable Credentials specification
The current Verifiable Credentials JSON-LD Context is horribly out of date and needs to be updated.

gkellogg:
This is what I gleaned from the data model document. Note that `Policy` has no good definition.

Fixes #202.


vocab/context.jsonld
```diff
@@ -1,94 +1,68 @@
...
-    "signatureValue": "sec:signatureValue",
-    "termsOfUse": {
-      "@id": "cred:termsOfUse",
-      "@type": "@id"
-    },
-    "verifiableCredential": {
-      "@id": "cred:verifiableCredential",
-      "@type": "@id",
-      "@container": "@graph"
+    "https://w3id.org/security/v1",
```
msporny: Should be https://w3id.org/security/v2

vocab/context.jsonld
```diff
@@ -1,94 +1,68 @@
...
+      "issuanceDate": {
+        "@id": "cred:issuanceDate",
+        "@type": "xsd:dateTime"
+      },
+      "issuer": {
+        "@id": "cred:issuer",
+        "@type": "@id"
+      },
+      "proof": {
+        "@id": "cred:proof"
```
dlongley: Proof is in the security vocab.

gkellogg:
It should just leave it out, then, as the security vocabulary is included.

Issue #19, opened 2016-11-28
Add renewalService URL to standard data model


msporny:
We may want to provide a verifiable claim refresh/renewal service to a verifiable claim such that it may be automatically renewed upon revocation or expiration. This could help smooth out  the experience around "expired credential" use cases such that expired credentials with a renewal service URL could still be considered for use without the individual explicitly having to renew the credential.

David-Chadwick:
I think this is primarily an implementation issue rather than a protocol one. In our implementation, the client knows which attributes the user is able to assert, and whether there is a current credential available or not. But the user is never pestered to renew his credentials. They are automatically renewed when the attribute needs to be asserted and the existing credential has expired. There is no point in renewing a credential prior to it being used. It may never be used before it expires.

msporny:
> they are automatically renewed when the attribute needs to be asserted and the existing credential has expired. 

Yes, that's exactly what this issue is driving at. We need a way to represent this renewal URL in the data model. We will also need to define a protocol for doing so (eventually).

stonematt:
What are the next steps on this issue?    This should have clarity early in our process.

msporny:
@stonematt next steps are that we need a concrete proposal as a PR. So, someone needs to volunteer to do that. @dlongley might be the best person to do that PR.

David-Chadwick:
I still do not see the need for this feature. Given that the user already has the VC that will need renewing sometime in the future, then the user's client already knows how to get the VC (otherwise the user would not have it). So why does it need to be told something it already knows?

stonematt:
seems related to #139 we'll try to address them together!

msporny:
There is a PR now that adds this feature: https://github.com/w3c/vc-data-model/pull/199

dlongley:
Closed via #210.

PR #210, opened 2018-07-31
Add refreshService feature.

msporny:
It is often useful for systems to enable the manual or automatic refresh of a verifiable credential that has expired. This PR enables an issuer to include a link to a refresh service.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/210.html" title="Last updated on Oct 2, 2018, 3:49 PM GMT (46fa745)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/210/143664b...46fa745.html" title="Last updated on Oct 2, 2018, 3:49 PM GMT (46fa745)">Diff</a>


David-Chadwick:
My feeling is that this property does not belong in a VC. It is a protocol issue how the holder
a) got the VC in the first instance
b) gets a subsequent VC later on
This data should not be in the VC as it is essentially public information and the issuer may not want this information to be public.

stonematt:
> What if the information is provided in a presentation from the issuer to the holder? That way, the holder can keep the information private and it doesn't go in the credential itself?

If we decouple the refresh information from the VC by putting it in the profile, how will the verifier know that "this" refresh link belongs to "that" specific VC?  If the refresh service is VC specific, it will have to be have that sort of link.  If the holder has the option to create a new presentation of an arbitrary collection of VCs in their repository, this seems like a detail that will be forgotten, or mistakenly included.  (Can a holder grab a refresh detail from Presentation X and put it in Presentation Y?).

msporny:
> If we decouple the refresh information from the VC by putting it in the profile, how will the verifier know that "this" refresh link belongs to "that" specific VC? 

You can use the identifier for the VC:

```
{
  "@context": "https://w3id.org/credentials/v1",
  "verifiableCredential": [{ 
    "id": "https://example.com/credentials/123",
    ...
  }],
  "refreshService: [{
    "type": "ManualRefreshService2018",
    "id": "https://example.com/refresh/289347028934",
    "verifiableCredential": "https://example.com/credentials/123",
    ...
  }]
  "proof": [{ ... }]
}
```

stonematt:
is the refresh service that same as renewal service in issue #19?  There is a valid use case that requires the refresh service be exposed directly to the verifier.  Consider a license that's in the public domain, such as a state medical license in the US.  Anyone can verify a physician's medical license without the physician being party to the verification.  Specialty Board Certification is similar.  

When a physician gets privileges at hospital, it's predicated on having and maintaining a valid medical license and often current board certification.  In today's world, the state or the board has an API or process to lookup current status based on the credential ID. In the VC world, I would expect that URL to be included in the VC or VP, and the Doctor would provide it to the hospital when they apply for privileges.  After that point, the hospital can check currency on demand, if it's revoked, or updated w/ new expiration attributes, they can update their records and proceed accordingly.

The process today is manual, labor intensive, and error prone.  In some cases, the physician today actually has to have the university reprint a copy of their diploma and send it by registered mail directly to the licensing board.  This is slowly changing but VCs could vastly simplify and accelerate this.

Ongoing verifications shouldn't ALWAYS require the holder/subject.

stonematt:
@msporny I added a paragraph that's captures my thinking about this detail. I don't think the language is exactly final and expect you'll have feedback. (see commit above).

msporny:
@stonematt -- I thought the group decided that we're not going to pull this PR in and I was supposed to close it?

stonematt:
I think the idea of a refreshService pointer as an optional attribute at Presentation and Verifiable Credential in addition to the option of the issuer embedding it in the core claim gives the data model a great deal of flexibility for the Issuer and Holder to expose this critical piece of information to a verifier, as appropriate.

The group is on the fence about the value and/or implication of this attribute, although at least one implementer plans to implement this as a data element right away, and the market will expect it.  

Moving forward, we can:
1) Remove the language all together and hope various implementers happen to add this capability the same way
2) Move this idea to the "Advanced Concepts" section as an example extension of the data model so future implementers have a hint as to how we might head in a subsequent version of the spec.

I vote for option 2.

David-Chadwick:
Can I suggest a change of wording as follows:
From

The issuer may choose to include the refresh service as an element inside the verifiable credential or the verifiable presentation, depending upon market and privacy needs. Including the refresh reference in the presentation requires the holder to refresh the credential details before creating a presentation to share with a verifier, while including the refresh reference in the verifiable credential exposes the refresh details to the verifier enabling future updates by the verifier. 

To

The issuer may include the refresh service as an element inside the verifiable credential if it intended for the verifier or for the verifier and the holder, or inside the verifiable presentation if it is intended for the holder only.

Including the refresh reference in the presentation enables the holder to refresh the credential details before creating a presentation to share with a verifier, while including the refresh reference in the verifiable credential enables the both the holder and the verifier to perform future updates of the credential.

stonematt:
@David-Chadwick I tweaked your language a little (please review):

The issuer may include the refresh service as an element inside the verifiable credential if it is intended for either the verifier or the holder, or inside the verifiable presentation if it is intended for the holder only.

Including the refresh reference in the presentation enables the holder to refresh the credential details before creating a presentation to share with a verifier, while including the refresh reference in the verifiable credential enables the both the holder and the verifier to perform future updates of the credential.

stonematt:
Content is in the Advanced Concepts section and is optional. Hopefully this meets the needs of the group and we can move forward.

David-Chadwick:
for either the verifier or the holder ->
for either the verifier or the holder or both,

PR #214, opened 2018-08-14
Replace identifier registry with Verifiable data registry. Clarify …

lovesh:
…that the verifiable data repositrory does not store subject identifiers or credentials but issuer identifiers, keys and some other data

Signed-off-by: lovesh harchandani <lovesh.bond@gmail.com>

***
<a href="https://pr-preview.s3.amazonaws.com/sovrin-foundation/vc-data-model/pull/214.html" title="Last updated on Dec 11, 2018, 5:19 AM GMT (4cdb01b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/214/6198a5e...sovrin-foundation:4cdb01b.html" title="Last updated on Dec 11, 2018, 5:19 AM GMT (4cdb01b)">Diff</a>


lovesh:
@msporny We did not intend **verifiable data repository** to contain credentials. It is meant to hold DIDs, public keys, schema, revocation registries and similar non-holder private data. We find the term **identifier registry** to be narrow

msporny:
> We did not intend verifiable data repository to contain credentials. It is meant to hold DIDs, public keys, schema, revocation registries and similar public data. We find the term identifier registry to be narrow

Ahhh got it, thanks; that's definitely not clear from the PR.

Perhaps we need to rename it to "cryptographic material registries", which includes supporting material that are used to cryptographically verify credentials (e.g., DID Ledgers, public key repositories, cryptographic schemas, and revocation lists)? Thoughts?

index.html
```diff
@@ -253,11 +253,12 @@ <h3>Ecosystem Overview</h3>
 <a>verifiable credentials</a> for processing.
 Examples of verifiers include employers, security personnel, and websites.
         </dd>
-        <dt><a>identifier registry</a></dt>
+        <dt><a>Verifiable data repository</a></dt>
```
msporny: This should be changed to something like "cryptographic material registry"

index.html
```diff
@@ -253,11 +253,12 @@ <h3>Ecosystem Overview</h3>
...
         <dd>
-A role a system may perform by mediating the creation and verification of
-<a>subject</a> identifiers. Examples of identifier registries include
-corporate employee databases, government ID databases, and distributed ledgers.
+Mediates the creation and verification of <a>issuer</a> identifiers, keys and
```
msporny: The registry is not only for issuers, it's for subjects in general of which issuers are a subtype.

index.html
```diff
@@ -579,19 +580,15 @@ <h1>Trust Model</h1>
...
+to be a correct record of which data is owned (and can be updated) by which
+<a>entities</a>.
         </li>
         <li>
 The <a>subject</a> trusts the <a>issuer</a> to issue true (i.e., not false)
 <a>credentials</a> about the subject, and to revoke them quickly when
 appropriate.
         </li>
-        <li>
-The <a>holder</a> trusts the <a>repository</a> to store the
```
msporny: This shouldn't be removed. This is talking about the holder's 'credential repository'... not the cryptographic material registry.

index.html
lovesh: I think schema is not a cryptographic material, neither are identifiers like a DID, do you agree? Can we call it **public data regsitry** so that it is clear that it contains no holder specific data

index.html
lovesh: Thanks, added back

index.html
lovesh: Fixed

index.html
msporny: Public data registry is too vague. Schema is cryptographic material -- it's used as input to cryptographic functions to perform calculations. Specifically, schemas enable byte streams to be created in a deterministic fashion, so yes, I would classify schemas as cryptographic material for the purposes of the VC Data Model.

index.html
```diff
@@ -602,8 +603,9 @@ <h1>Trust Model</h1>
 risk assessment of the verifier.
         </li>
         <li>
-All entities trust the <a>identifier registry</a> to be incorruptible and
-to be a correct record of which identifiers belong to which <a>entities</a>.
+All entities trust the <a>Public data repository</a> to be un-corruptible and
```
dlongley: typo `un-corruptible` => `incorruptible` ... or maybe just use `tamper-evident` instead like we do elsewhere?

index.html
```diff
@@ -254,11 +254,12 @@ <h3>Ecosystem Overview</h3>
 <a>verifiable presentations</a> for processing.
 Examples of verifiers include employers, security personnel, and websites.
         </dd>
-        <dt><a>identifier registry</a></dt>
+        <dt><a>Public data repository</a></dt>
```
dlongley: `Public` => `public` (throughout)

index.html
lovesh: Done

index.html
lovesh: I am re-inclining to Verifiable Data Repository. Verifiable communicates the intent, cryptographic seems like a blanket term. As an example, seed to a PRF is a cryptographic data but not something we want to put in the repository

brentzundel:
Can this PR move forward? @msporny  has requested it be discussed by the group before it can be merged: 
 > I can't just process this as editorial without running it by the group.

What is the next step here?

David-Chadwick:
One minor comment I have is what is the difference between repository and registry?
Of the top of my head, I would say that a repository is less restrictive than a registry in that the former might allow duplication (e.g. two different entities to store the same ID), whereas the latter would not as it is registering them.
I would therefore prefer the term verifiable data registry, as @lovesh said it would be in his comment above. But the current PR text uses repository instead.

jandrieu:
We use repositories for where data is *stored*, such as identity hubs or wallets.

Maybe there is a better choice than registries for recording the unique namespacing of identifiers, but repositories definitely conflates registries with storage locations.

msporny:
The result of the discussion at W3C TPAC 2018 was that we pull the PR in after no more than 7 days of bike shedding on the word "schema".

index.html
```diff
@@ -257,10 +257,10 @@ <h3>Ecosystem Overview</h3>
 <a>verifiable presentation</a> that proves the holder possesses the required verifiable credentials with certain characteristics.
 Examples of verifiers include employers, security personnel, and websites.
         </dd>
-        <dt><a>identifier registry</a></dt>
+        <dt><a>Verifiable data registry</a></dt>
```
msporny: Concrete suggestion: Change to "Cryptographic Data Registry" 

Reasoning: The data isn't always verifiable (e.g. federated or DHT-style systems). In almost every case, the registry does contain data that is used in cryptographic processes of one kind or another.

index.html
```diff
@@ -257,10 +257,10 @@ <h3>Ecosystem Overview</h3>
...
         <dd>
 A role a system may perform by mediating the creation and verification of <a>issuer</a> identifiers,
-keys and other relevant data required to use verifiable credentials. Some configurations may require correlatable
+keys and other relevant data like schemas, revocation registries, etc. required to use verifiable credentials. Some configurations may require correlatable
```
msporny: Concrete suggestion: Change "schemas" to "data validation schemas"

Reasoning: The term "schema" is too broad and has led to confusion in the group.

terms.html
```diff
@@ -83,7 +83,7 @@
 distribution or use of identity information can compromise privacy. Collection
 and use of such information should follow the principles of minimal disclosure.
   </dd>
-  <dt><dfn data-lt="identifier registries">identifier registry</dfn></dt>
+  <dt><dfn data-lt="verifiable data registries">Verifiable data registry</dfn></dt>
```
msporny: This definition will need to be expanded based on the expansion of scope above.

David-Chadwick:
LGTM

terms.html
stonematt: @brentzundel could you take at improving this definition?

brentzundel: done

index.html
brentzundel: unless there is significant consensus for "Cryptographic Data Registry," or a lot of opposition to "Verifiable data registry" I would prefer to keep "Verifiable data registry" for now.

msporny: This would mean that a "DID Ledger" is a type of "Verifiable Data Registry"... which conflates the way "Verifiable" is used in the VC data model spec, and the general usage of the word. It doesn't feel quite right to me because some DID Ledgers, like Veres One, store data (like service descriptions, that aren't "Verifiable" in the "Verifiable Credentials" sense of the word, so it's a strange fit. 

I think the bigger issue is that a federated/mirrored corporate system would also be a "Verifiable Data Registry", but some of that information is not verifiable (i.e., it is only valid because the corporation says it's valid)... but the fact that all of these systems hold cryptographic data in one form or another is the common denominator here.

Keep in mind that this started out as an "Identifier Registry" because the block had a specific purpose. We are now expanding that purpose and have to be careful to not expand too far, or pick the wrong word.

We need more input on this, perhaps on a call (because I don't think the people that may have something to say about this are paying attention to code commit comments), perhaps via a run off poll because we're bike shedding at this point and it'll go on forever w/o some finality to the name change. We had to bike shed this name early on and did an IRV poll to get finality that time. We should do this again so we can get this PR sorted quickly and into the spec.

ken-ebert: Verifiable Data Registry is inaccurate. The credential itself contains the verifiable data.

I think this registry potentially contains a broad variety of data that is used for verification, such as service descriptions, schemas, credential definitions, DIDs, revocation information, etc. 

How does the name Verification Material Registry sound? I think it reflects the scope and intent for which it will be used.

msporny: "Verification Material Registry" is probably the most accurate thing so far, so I'd be fine w/ that ... and now we're squarely in "we need to ask the WG" territory. I'll try to draft up an Instant Runoff Vote for this... We have "Verifiable Data Registry", "Cryptographic Data Registry", "Cryptographic Material Registry", and "Verification Material Registry"... any other contenders that folks want to put forward?

brentzundel: Verification material registry is fine with me as well.

dlongley: I'm a little concerned that we're perhaps losing the original main intent for the registry which was to be an identifier registry. I understand that there are other registry needs as well (e.g. Sovrin credential schemas) -- and that many of them are related to verification. Perhaps we can just call this system component "Registries" and then talk about what different types of registries might be out there?

What's important, IMO, is that they are seen as independent components in the ecosystem, i.e. they are not tightly coupled to the issuer, verifier, or holder, allowing for agility and greater decentralization.

msporny:
There is now a document that is attempting to collect options for an Instant Run-off Vote: 

https://docs.google.com/document/d/1dCPGKiW9uqVd1X_Ye6BCFt4qzRRGjPNfTuYEQKIZk6k/edit#

I hope that everyone is happy with themselves, this is why we can't have nice things. :stuck_out_tongue_winking_eye:

msporny:
I tried to resolve the merge conflicts here, but failed miserably. Lots of merges and edits. Let me know if you get this fixed @brentzundel ... if not, a new PR would probably be easier to merge in.

dlongley:
@brentzundel -- I took care of it, we're good here.

PR #220, opened 2018-08-16
Distinguish credentials from presentations

dhh1128:
Signed-off-by: Daniel Hardman <daniel.hardman@gmail.com>

This embodies item #4 from https://docs.google.com/document/d/10e6lcsX0kiXkWX4_79hD1fb4p_AbFGsRm90eJJKFayI/edit -- chiefly, defining presentation and clarifying its relationship to a credential.

***
<a href="https://pr-preview.s3.amazonaws.com/dhh1128/vc-data-model/pull/220.html" title="Last updated on Aug 21, 2018, 5:08 PM GMT (e52e043)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/220/838e1f4...dhh1128:e52e043.html" title="Last updated on Aug 21, 2018, 5:08 PM GMT (e52e043)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -980,51 +980,11 @@ exists for implementers that would like to implement credential status checking.
         <h2>Presentations</h2>
 
         <p>
-Credentials MAY be composed by placing them into a data structure called a
-<a>presentation</a>. A <a>verifiable presentation</a> is a presentation that contains
-<a>verifiable credentials</a> and one or more proofs that are appropriate
-for the presentation. The basic structure of a <a>verifiable presentation</a>
-is provided below:
+Credentials MAY be used to make a <a>verifiable presentation</a>. The presentation
+may be of a credential itself, or of a profile, or of data derived from credentials
+in a format that is cryptographically verifiable.
         </p>
 
-        <pre class="example nohighlight" title="Basic structure of a presentation">
-{
-  "@context": "https://w3id.org/credentials/v1",
-  "id": "urn:uuid:3978344f-8596-4c3a-a978-8fcaba3903c5",
-  "type": ["VerifiablePresentation"],
-  <span class="highlight">"verifiableCredential": [{ ... }],
-  "proof": [{ ... }]</span>
-}
-        </pre>
-
-        <p class="issue" data-number=133>
-The group is currently discussing whether the "id" field is appropriately
-defined for Presentations.
-        </p>
-
-        <p class="issue" data-number=120>
-The group is currently discussing whether the "id" field is appropriately
-defined across all objects.
-        </p>
-
-        <p class="issue" data-number=106>
-The group is currently discussing how the data model expresses when the
-subject is not the holder of the credential/presentation.
-        </p>
-
-        <p class="issue" data-number=105>
-The group is currently discussing holders, subjects, and identifier
-control.
-        </p>
-
-        <p>
-The contents of the <code>verifiableCredential</code> property are
-<a>verifiable credentials</a> as described by this specification. The contents
-of the <code>proof</code> property are proofs as described by the
-Linked Data Proofs [[LD-PROOFS]] specification. The <code>id</code> property
-is optional and MAY be used to provide a unique identifier for the presentation.
-The value associated with the <code>id</code> property MUST be a URI.
-        </p>
       </section>
     </section>
 
@@ -1248,10 +1208,9 @@ new <code>favoriteFood</code> term so that it may only be used within the
         <p>Terms of use can be utilized by an <a>issuer</a>, <a>subject</a> or a
         <a>holder</a> to express limitations on the use of information
         expressed by the Verifiable Credentials Data Model. The <a>issuer</a> places
-        the terms of use inside the <a>credential</a> before it is converted into a
-        <a>verifiable credential</a>. The <a>holder</a> places the terms of use inside
-        the <a>presentation</a> before it is converted into a
-        <a>verifiable presentation</a>.</p>
+        their terms of use inside the <a>credential</a> before it is converted into a
+        <a>verifiable credential</a>. The <a>holder</a> places issuer terms
+        of use, if applicable, plus holder terms of use, inside a <a>presentation</a>.</p>
 
        <p class="note"> it is for further study how a subject who is not a
        holder places terms of
@@ -1282,6 +1241,7 @@ new <code>favoriteFood</code> term so that it may only be used within the
 The group is currently exploring a variety of ways of expressing the terms of
 use associated with a Verifiable Credential, namely, the
 <a href="http://w3c.github.io/poe/model/">Open Digital Rights Language</a>.
+A related initiative is the one at customercommons.org.
         </p>
 
         <pre class="example nohighlight" title="Usage of termsOfUse property by an Issuer">
--- a/index.html
+++ b/index.html
@@ -1522,14 +1522,16 @@ should treat this entire section as at-risk and currently under debate.
       <p>
 Normally verifiable credentials will be presented to verifiers by the
 subject. However in some cases, the subject may need to pass the whole or
-part of a verifiable credential to a holder. The data model allows for both
-cases as follows.
+part of a verifiable credential to another holder. The data model allows for
+both cases as follows.
       </p>
 
       <p>
 If only the subject is allowed to present the verifiable credential to a
 verifier, the issuer MUST insert the "subject only" Terms of Use into the
-verifiable credential, as defined below.
+verifiable credential, as defined below. <span class="issue">
+This feature is at risk and is likely to be removed due to lack of consensus.
+</span>
       </p>
 
       <p>
@@ -1538,12 +1540,21 @@ verifiable credential to a holder, for example, a patient may pass a
 prescription verifiable credential to a relative for the relative to present
 it to a pharmacist for dispensing, then the subject MUST issue a verifiable
 credential to the holder containing the claim properties that are being
-passed on, as described below.
+passed on, as described below. <span class="issue">
+This feature is at risk and is likely to be removed and replaced with
+another solution, possibly created by another Working Group, that focuses on
+delegation of authority instead of delegation of attributes.
+</span>
       </p>
 
    <section>
 
    <h3> 'Subject Only' Terms of Use </h3>
+
+      <p class="issue" data-number=204>
+The group is currently debating the best security model for delegation. Readers
+should treat this entire section as at-risk and currently under debate.
+      <p>
 
   <p>
 The Subject Only Terms of Use states that a verifiable credential MUST only be presented to a verifier by the subject. If a verifier is presented with a verifiable credential containing the Subject Only Terms of Use, by anyone other than the subject, it MUST refuse to accept it.</p>
@@ -1571,6 +1582,11 @@ The Subject Only Terms of Use states that a verifiable credential MUST only be p
   <section>
    <h3> Passing on a Verifiable Credential </h3>
 
+      <p class="issue" data-number=204>
+The group is currently debating the best security model for delegation. Readers
+should treat this entire section as at-risk and currently under debate.
+      <p>
+
    <p>
 
 When the subject passes a verifiable credential to a
@@ -1580,8 +1596,6 @@ the subject is the holder to whom the verifiable credential is being passed,
 and the claim contains the properties that are being passed on. In addition, the holder creates a verifiable presentation that contains these two
 verifiable credentials.
 </p>
-
-
 
         <pre class="example nohighlight" title="An example of a holder presenting verifiable
 credential properties that have been passed to it by the subject">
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -605,7 +626,7 @@ <h1>Trust Model</h1>
 repository</a>, and
         </li>
         <li>
-the <a>issuer</a> does not need to trust the <a>verifier</a>.
+the <a>issuer</a> does not need to trust (or even know) the <a>verifier</a>.
```
msporny: Try to avoid the parenthetical... maybe "does not need to trust or know"...

terms.html
```diff
@@ -96,6 +97,17 @@
 A role an <a>entity</a> may perform by asserting <a>claims</a> about one or more
 <a>subjects</a>, creating a <a>verifiable credential</a> from these claims,
 and transmitting the verifiable credential to a <a>holder</a>.
+  </dd>
+  <dt><dfn data-lt="presentation">presentation</dfn></dt>
+  <dd>
+Data derived from one or more credentials, fitted to the needs of sharing
+with a particular verifier, and encoded in such a way that the verifier may
+trust authorship of the data after a process of cryptographic verification. A
+credential itself may be presented, in which case the credential becomes a
```
msporny: The following three sentences are problematic. Can we strike them entirely? Over-specifying things will make it more difficult for us to support broader use cases in the future.

index.html
```diff
@@ -980,51 +980,11 @@ <h2>Status</h2>
...
-<a>presentation</a>. A <a>verifiable presentation</a> is a presentation that contains
-<a>verifiable credentials</a> and one or more proofs that are appropriate
-for the presentation. The basic structure of a <a>verifiable presentation</a>
-is provided below:
+Credentials MAY be used to make a <a>verifiable presentation</a>. The presentation
+may be of a credential itself, or of a profile, or of data derived from credentials
+in a format that is cryptographically verifiable.
         </p>
 
-        <pre class="example nohighlight" title="Basic structure of a presentation">
```
msporny: I'm guessing that the deletion of the majority of the content in this section was an error?

index.html
```diff
@@ -1248,10 +1208,9 @@ <h2>Terms of Use</h2>
         <p>Terms of use can be utilized by an <a>issuer</a>, <a>subject</a> or a
         <a>holder</a> to express limitations on the use of information
         expressed by the Verifiable Credentials Data Model. The <a>issuer</a> places
-        the terms of use inside the <a>credential</a> before it is converted into a
-        <a>verifiable credential</a>. The <a>holder</a> places the terms of use inside
-        the <a>presentation</a> before it is converted into a
-        <a>verifiable presentation</a>.</p>
+        their terms of use inside the <a>credential</a> before it is converted into a
+        <a>verifiable credential</a>. The <a>holder</a> places issuer terms
```
msporny: This is not how the data model is designed. The presentation is something created by the holder. The holder can put  credentials (which may or may not contain **issuer termsOfUse**) into the presentation in the `verifiableCredential` property. The holder may also place **holder termsOfUse** into the presentation.

index.html
```diff
@@ -1282,6 +1241,7 @@ <h2>Terms of Use</h2>
 The group is currently exploring a variety of ways of expressing the terms of
 use associated with a Verifiable Credential, namely, the
 <a href="http://w3c.github.io/poe/model/">Open Digital Rights Language</a>.
+A related initiative is the one at customercommons.org.
```
msporny: Change customercommons.org to &lt;a href="https:// customercommons.org">Customer Commons&lt;/a>

index.html
```diff
@@ -980,51 +980,11 @@ <h2>Status</h2>
...
-{
-  "@context": "https://w3id.org/credentials/v1",
-  "id": "urn:uuid:3978344f-8596-4c3a-a978-8fcaba3903c5",
-  "type": ["VerifiablePresentation"],
-  <span class="highlight">"verifiableCredential": [{ ... }],
-  "proof": [{ ... }]</span>
-}
-        </pre>
-
-        <p class="issue" data-number=133>
```
msporny: We can't remove issues from the spec that are not resolved.

index.html
```diff
@@ -980,51 +980,11 @@ <h2>Status</h2>
...
-subject is not the holder of the credential/presentation.
-        </p>
-
-        <p class="issue" data-number=105>
-The group is currently discussing holders, subjects, and identifier
-control.
-        </p>
-
-        <p>
-The contents of the <code>verifiableCredential</code> property are
```
msporny: We can't remove spec text that has previously achieved consensus unless we go back to the group and request that the text be removed... and I doubt that will happen for this paragraph.

index.html
```diff
@@ -373,17 +376,22 @@ <h3>Use Cases and Requirements</h3>
...
         <li>
-<a>Holders</a> that share <a>verifiable credentials</a> are not required to
+<a>Holders</a> that share <a>verifiable presentations</a> are not required to
 reveal the identity of the <a>verifier</a> to <a>issuers</a>.
         </li>
         <li>
-A <a>verifiable credential</a> is expressed in one or more standard,
-machine-readable data formats which can also be extended with minimal
-coordination.
+<a>Holders</a> in <a>presentations</a> can either disclose the attributes or satisfy <a>predicates</a>
```
msporny: predicates are linked to here, but not defined. If we are going to introduce that new term, we'll have to define it in terms.html

msporny: Thinking about it a bit more, do we need to introduce the "predicate" term? It's just another attribute isn't it? Can't we just provide an example of an attribute that is a predicate w/o increasing the cognitive burden of the reader to have to learn a new term?

index.html
```diff
@@ -458,6 +466,12 @@ <h3>Claims</h3>
         </figcaption>
       </figure>
 
+      <p>
+If a presentation supports predicates, a claim about birthdate (not true or false, but tied to
+a specific date value) can become the basis of proof that at an arbitrary point in time, the
+age of a subject did or will lie within an arbitrary interval.
```
msporny: I misread "lie" to mean "will not tell the truth" because my mind was in "verifiable credentials land" where some things have a true/false nature. Can we change the language to "the age of the subject has or will be within the specified interval." ... don't even know if that does it. It's hard to understand what's being said here. 

Maybe: "Predicates may be used in presentations to increase privacy by specifying things in a more coarse grained manner, such as the subject being over the age of 65 without exposing their exact birthday."

That feels like it's easier to read... but again, I question the need to introduce the term "predicate" to the document's vocabulary.

index.html
```diff
@@ -959,51 +980,11 @@ <h2>Status</h2>
         <h2>Presentations</h2>
 
         <p>
-Credentials MAY be composed by placing them into a data structure called a
-<a>presentation</a>. A <a>verifiable presentation</a> is a presentation that contains
-<a>verifiable credentials</a> and one or more proofs that are appropriate
-for the presentation. The basic structure of a <a>verifiable presentation</a>
-is provided below:
+Credentials MAY be used to make a <a>verifiable presentation</a>. The presentation
```
msporny: The specification makes a distinction between presentations (which are not verifiable) and verifiable presentations (which are verifiable). The modified language goes back and forth between these definitions, which muddies the waters. What about this:

> Credentials MAY be used to create presentations, which are composed of information from one or more credentials. A verifiable presentation is a presentation that is composed of information from one or more verifiable credentials and is thus cryptographically verifiable. The information in a verifiable presentation may be directly or indirectly derived from one or more verifiable credentials. The basic structure of a verifiable presentation is provided below:

terms.html
```diff
@@ -15,20 +15,21 @@
 A set of one or more <a>claims</a> made by the issuer.
 A <dfn data-lt="verifiable credentials">verifiable credential</dfn>
 is a credential that is tamper-resistant and whose authorship can be
-cryptographically verified.
+cryptographically verified. Credentials can be used to build
+<a>presentations</a>, which can also be cryptographically verified.
```
msporny: We need to be careful with the language here. A credential and a presentation are not cryptographically verifiable. A **verifiable credential** and a **verifiable presentation** are cryptographically verifiable.

terms.html
```diff
@@ -15,20 +15,21 @@
...
-well-formed.
+The process that demonstrates the information in a <a>credential</a>
+or <a>presentation</a> is well-formed.
   </dd>
-  <dt><dfn data-lt="credential verification">verification</dfn></dt>
+  <dt><dfn data-lt="verification">verification</dfn></dt>
   <dd>
 The process that cryptographically demonstrates the authenticity of a
-<a>credential</a>.
+<a>credential</a> or a <a>presentation</a>.
```
msporny: The original text is wrong... we should say "verifiable credential" and "verifiable presentation"

terms.html
dhh1128: Can you give me the first few words and the last few words of the sentences you're talking about? I'm not sure what "the following 3 sentences" means because of the way github is showing me the map between comment and lines...

index.html
```diff
@@ -238,8 +238,8 @@ <h3>Ecosystem Overview</h3>
         <dt><a>holder</a></dt>
         <dd>
 A role an <a>entity</a> may perform by possessing one or more
-<a>verifiable credentials</a>. Examples of holders include students,
-employees, and customers.
+<a>verifiable credentials</a> and generationg <a>presentations</a> from them
```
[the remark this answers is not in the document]
dhh1128: Yes, I believe I did this.

index.html
```diff
@@ -373,17 +376,22 @@ <h3>Use Cases and Requirements</h3>
...
+        </li>
+        <li>
+<a>Verifiable credentials</a> and <a>verifiable presentations</a> are expressed in one or more standard,
+machine-readable data formats which can also be extended with minimal coordination.
         </li>
         <li>
-<a>Verifiable credentials</a> are independently issued, stored,
-and verified.
+Issuers independently issue <a>verifiable credentials</a>, and holders store them.
+Holders independently offer <a>verifiable presentations</a>, and verifiers
```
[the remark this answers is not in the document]
dhh1128: Will try to reword on Monday.

index.html
```diff
@@ -458,6 +466,12 @@ <h3>Claims</h3>
         </figcaption>
       </figure>
 
+      <p>
+If a presentation supports predicates, a claim about birthdate (not true or false, but tied to
```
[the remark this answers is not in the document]
dhh1128: Will refine on Monday.

index.html
```diff
@@ -531,8 +545,15 @@ <h2>Presentations</h2>
       </p>
 
       <p>
-A <a>verifiable presentation</a> is a collection of one or more
-<a>verifiable credentials</a> that are often about the same <a>subject</a> that
+A <a>verifiable presentation</a> is data from one or more <a>verifiable
```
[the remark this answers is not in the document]
dhh1128: Absolutely. Will improve on Monday.

index.html
```diff
@@ -584,7 +605,7 @@ <h1>Trust Model</h1>
 to be a correct record of which identifiers belong to which <a>entities</a>.
         </li>
         <li>
-The <a>subject</a> trusts the <a>issuer</a> to issue true (i.e. not false)
+The <a>subject</a> trusts the <a>issuer</a> to issue true
```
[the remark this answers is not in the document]
dhh1128: We're in the section about how trust works. We don't trust a credential because the issuer "asserts certain types of information about the subject" -- we trust the credential because we think it's likely that the issuer's assertions are actually true. The verification of a credential does not, of course, test the truth value of assertions--only the authorship of them--but as a description of how trust works, merely stating that we trust a party to make assertions is not going far enough, I think.

index.html
dhh1128: Will revise on Monday.

index.html
dhh1128: No, it was not a mistake. I was proposing that we omit any example of a presentation, because the example shown is SO different from the one that ZKPs have to use. I could submit a PR with a second example of a presentation (this one ZKP-style), but with little in common between the two examples, it's not clear to me what value there would be in pairing them.

Perhaps this needs to be the subject of a different PR.

index.html
dhh1128: Okay. I will look for a way to revise. Where do subject termsOfUse go?

index.html
dhh1128: will revise

index.html
dhh1128: this was a mistake on my part. Will revise.

index.html
dhh1128: Yes, I see how I muddied the distinction. I will revise, using your language as a guide.

terms.html
dhh1128: Will adjust.

terms.html
dhh1128: will adjust

index.html
```diff
@@ -531,8 +545,15 @@ <h2>Presentations</h2>
       </p>
 
       <p>
-A <a>verifiable presentation</a> is a collection of one or more
-<a>verifiable credentials</a> that are often about the same <a>subject</a> that
+A <a>verifiable presentation</a> is data from one or more <a>verifiable
+credentials</a>, packaged in such a way that the authorship of the data
+is verifiable. If credentials are directly presented, they become a
+presentation. Profiles are presentations. Data formats derived from
```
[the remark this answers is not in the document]
dhh1128: will adjust

index.html
dhh1128: For now, I am removing the idea of a holder adding issuer terms of use to a presentation. However, I have logged #224 to track a conceptual mismatch that needs to be resolved.

index.html
dhh1128: The way I've been using it, a "predicate" (I've now renamed to "derived predicate" in latest incarnation of PR) is NOT an attribute in a credential. It is an attribute in a presentation that is derived from an attribute in a credential.

index.html
David-Chadwick: @dhh1128 I think you have misunderstood the current model. The holder is not adding issuer terms of use to a presentation, but rather his/her own terms of use to the presentation. The issuer terms of use are visible in the credential.
Now I do understand your problem with anonymous credentials, in that the VC is opaque. But I would argue that the way to resolve this is not to alter the structure of the VP, but rather, when the holder asserts what the opaque VC contains, that the issuer terms of use are also asserted as well. I suspect this is a protocol issue rather than a data model issue. I realise that a holder may wish to hide certain issuer terms of use, but again, this is a problem that anonymous credentials has to resolve (in its protocol), rather than in the VC data model.

dhh1128:
@msporny I have removed the example of a ZKP-style presentation.

index.html
```diff
@@ -1027,62 +1027,9 @@ <h2>Presentations</h2>
...
-                        <i>// meant to be parsed by a library supporting "z-mix"</i>
-                        ...
-                        ....
-                    }
-                }
-              ]
-            </span>
-            }
-            </pre>
+          We are currently working on an example of a ZKP-style verifiable presentation that contains
```
msporny: I'm pulling in this PR, but for future reference. You can do this:

&lt;p class="issue">Text explaining a work in progress item that will be highlighted appropriately in the spec&lt;/p>

msporny:
Good PR, aligned w/ VCWG consensus as of the call today. Merged.

PR #227, opened 2018-08-29
Disputed credentials

David-Chadwick:
Added clarifying text and a second example.

***
<a href="https://pr-preview.s3.amazonaws.com/David-Chadwick/vc-data-model/pull/227.html" title="Last updated on Oct 18, 2018, 9:05 AM GMT (37a366c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/227/ba43ef7...David-Chadwick:37a366c.html" title="Last updated on Oct 18, 2018, 9:05 AM GMT (37a366c)">Diff</a>


index.html
```diff
@@ -1243,21 +1243,48 @@ <h2>Evidence</h2>
...
-credential with an identifier of <code>https://example.org/credentials/245</code>
-contains disputed statements, an entity would issue the following credential
-in a public venue to make it known that the credential is disputed:
-        </p>
+The time may come when an <a>entity</a> wants to dispute a <a>credential</a>
+issued by the issuer. There at least two different cases to consider:
+a subject disputes a claim made by the issuer, e.g. the address property is out
+of date, or an entity disputes a (false) claim made by the issuer about a different subject, e.g. an imposter is claiming the entity's social security number. 
+Only the subject of a verifiable credential, or its authorised
+agent, is entitled to issue a "DisputeCredential". A "DisputeCredential" issued by anyone other than the subject, or its authorised agent, SHOULD be disregarded by the verifier, unless the verifier has some out of band means of ascertaining the truth of
```
dlongley: Perhaps instead of saying who is entitled to issue a "DisputeCredential" we should say that verifiers should only accept those issued by the subject? We say this in the next sentence, so perhaps just drop the one about entitlement?

Also, w3c specs use American English spelling so "authorised" => "authorized".

index.html
```diff
@@ -1243,21 +1243,48 @@ <h2>Evidence</h2>
...
+the dispute. This is to prevent denial of service attacks whereby an attacker
+falsely disputes a true claim. 
+
+The mechanism for issuing a "DisputeCredential"
+is the same as issuing a regular credential except that the <a>subject</a> identifier
+for the <a>claim</a> in the "DisputeCredential" is the identifier of the disputed credential.  For example, if a credential with an identifier of
+<code>https://example.org/credentials/245</code>
+is disputed, an entity may issue one of the following credentials. In the former case
+the subject might present this to the verifier along with the disputed credential. In
```
dlongley: comma: `In the former case,`

index.html
```diff
@@ -1243,21 +1243,48 @@ <h2>Evidence</h2>
...
+the later case, the entity might publish the "DisputeCredential"
```
dlongley: `later` => `latter`

index.html
```diff
@@ -1243,21 +1243,48 @@ <h2>Evidence</h2>
...
+
+        <pre class="example nohighlight" title="Subject disputes a credential">
+{
+  "id": "http://example.com/credentials/123",
+  "type": ["VerifiableCredential", "DisputeCredential"],
+  <span class="highlight">"claim": {
+    "id": "http://example.com/credentials/245",
+    "currentStatus": "Disputed",
+    "statusReason": "Address is out of date",
+    "address": "10 Some Street, Anytown, ThisLocal, Country X"
```
dlongley: We should probably use something like schema.org for these sorts of examples. Also, the address should be that of the subject... not that of the credential as is suggested here. I'm also not clear on whether what's being modeled here is the disputed address or the proper address. Maybe we just want to strike it and keep these dispute credentials simple for now?

index.html
msporny: Agreed. I think we get in trouble when talking about "authorized agent" because we then have to explain what one is. Also, yes... W3C specs are in American English (for better or worse).

index.html
msporny: Agreed, strike it, keep dispute credentials simple for now. Only include currentStatus and statusReason. Strike `address`.

index.html
```diff
@@ -1243,21 +1243,48 @@ <h2>Evidence</h2>
...
+  }</span>,
+  "issuer": "https://example.com/people#me",
+  "issued": "2017-12-05T14:27:42Z",
+  "proof": { ... }
+}
+        </pre>
+
 
-        <pre class="example nohighlight" title="Expressing a disputed credential">
+        <pre class="example nohighlight" title="Another entity disputes a credential">
```
msporny: I don't understand what is meant by "Another".

msporny:
Ping @David-Chadwick -- @gannan reminded me to do this as it had once again fallen off of my radar.

index.html
David-Chadwick: I have removed authorized agent as requested.

index.html
David-Chadwick: Done

index.html
David-Chadwick: done

index.html
David-Chadwick: Address has been deleted.

index.html
David-Chadwick: This is to cover the case where an imposter has managed to get my SS number in his VC. In this case, I am not the subject of the VC, as the imposter is the subject. So I am 'Another entity' who is publishing the dispute in a public place. We cannot replace 'another entity' by 'subject' or 'real subject' because it might be a DOS attack. So in reality, we have no idea who this other entity is, except by some OOB unspecified means.

dlongley:
Closing in favor of #259, which rebases this one. This one was based on a version of the spec that was around 5 months old.

David-Chadwick:
Thanks Dave

PR #230, opened 2018-09-13
Consistently use the VerifiableCredential type

cwebber:
Additionally, the context at https://w3id.org/credentials/v1 says Credential instead of VerifiableCredential as well, as I discovered today while writing some VC-related code.

***
<a href="https://pr-preview.s3.amazonaws.com/cwebber/vc-data-model/pull/230.html" title="Last updated on Sep 13, 2018, 7:44 PM GMT (886ee4a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/230/a75b347...cwebber:886ee4a.html" title="Last updated on Sep 13, 2018, 7:44 PM GMT (886ee4a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1513,7 +1513,7 @@ that uniquely identifies a subject">
 {
   "@context": ["https://w3id.org/credentials/v1", "https://schema.org/"]
   "id": "http://dmv.example.gov/credentials/332",
-  "type": ["Credential", "IdentityCredential"],
+  "type": ["VerifiableCredential", "IdentityCredential"],
   "issuer": "https://dmv.example.gov/issuers/4",
   "issued": "2017-02-24",
   "claim": {
@@ -1636,7 +1636,7 @@ credential properties that have been passed to it by the subject">
   "type": ["VerifiablePresentation"],
   "credential": [{
       "id": "http://example.gov/credentials/3732",
-      "type": ["Credential", "PrescriptionCredential"],
+      "type": ["VerifiableCredential", "PrescriptionCredential"],
       "issuer": "https://dmv.example.gov",
       "issued": "2010-01-01",
       "claim": {
@@ -1651,7 +1651,7 @@ credential properties that have been passed to it by the subject">
     },
     {
       "id": "https://example.com/VC/123456789",
-      "type": ["Credential", "PrescriptionCredential"],
+      "type": ["VerifiableCredential", "PrescriptionCredential"],
       "issuer": "did:example:ebfeb1f712ebc6f1c276e12ec21",
       "issued": "2010-01-03",
       "claim": {
@@ -1724,7 +1724,7 @@ standardised by the working group.
       <pre class="example nohighlight" title="An example of the relationship
 property in a child's credential">{
   "id": "http://dmv.example.gov/credentials/3732",
-  "type": ["Credential", "ProofOfAgeCredential"],
+  "type": ["VerifiableCredential", "ProofOfAgeCredential"],
   "issuer": "https://dmv.example.gov/issuers/14",
   "issued": "2010-01-01T19:73:24Z",
   "claim": {
@@ -1748,7 +1748,7 @@ credential if it is provided by the child or the parent.
       <pre class="example nohighlight" title="An example of a relationship
 credential issued to a parent">{
   "id": "http://dmv.example.gov/credentials/3732",
-  "type": ["Credential", "RelationshipCredential"],
+  "type": ["VerifiableCredential", "RelationshipCredential"],
   "issuer": "https://dmv.example.gov/issuers/14",
   "issued": "2010-01-01T19:73:24Z",
   "claim": {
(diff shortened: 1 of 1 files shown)


gkellogg:
Perhaps there's no mechanism to pull from https://w3c.github.io/vc-data-model/vocab/credentials.jsonld? I see that context defines VerifiableCredential.

cwebber:
I didn't realize that we had two different types for `Credential` and `VerifiableCredential`, I thought it was an accidental inconsistency.  Based on conversation on the call today, the reason we had both was that the former for was when a proof wasn't attached.  But it does sound like from the call that in general people did feel that the spec shouldn't mention both.

Probably two paths forward:

* Switch all uses of `VerifiableCredential` to `Credential`, and a `Credential` becomes verifiable when it has a proof that checks out.  I like this better myself personally, but it sounds like some people have issues with the term `Credential` being ambiguous or something.
* Switch all uses of `Credential` to `VerifiableCredential`, as is currently done in this PR and ensure that all of them have `proof`s (is this complicated by the `proof` term not yet making it into a standard of recommendation status?).  At that point, we make `Credential` somehow optionally available but mostly sweep it off this spec (maybe we can mention it, but just not use it in any examples).

Thoughts?

David-Chadwick:
There are two different places that VCs occur:

1. In standalone VCs, as in most of the examples. In this case the outer type is not specified. The JSON starts with
```
{
  "@context": "https://w3id.org/credentials/v1",
   "id": "http://dmv.example.gov/credentials/3732",
  "type": ["VerifiableCredential", "ProofOfAgeCredential"],
.....
```

2. When it is encapsulated in a presentation. In this case the JSON is
```
"verifiableCredential": [{ ... }],
```

I think @cwebber is only addressing 1. above. In my opinion a verifiable credential without a proof is not a credential, but is simply a claim or a statement, and the standard does not provide any way of verifying whether the claim is true, or whether anyone can prove who made the statement. My opinion is that the standard should say nothing about such constructs. They should be out of scope. The standard should only talk about verifiable credentials (as the title of the data model document says).

Concerning 2. it should be pretty obvious what the encapsulating type should be, because if the JSON construct is a verifiable credential then it will already have the proof attached, but if it does not have a proof attached, then in this case it is a claim or statement and the encapsulating type can state this. However I propose that such proof-less constructs should out of scope of the current data model.

cwebber:
So the conversation on the call was roughly as follows:

- We can't do just `Credential` because that will be a political problem in the standards process, regardless of whether it's a good idea or not.  So `VerifiableCredential` will remain.
- We are making a strong distinction between "credentials" and "verifiable credentials" in the spec, so `Credential` will remain.

That leaves us with two options, which it didn't seem like achieved consensus on in the call:

1. `VerifiableCredential` always requires a proof, and any example we show with a `VerifiableCredential` should have a proof attached, and we just won't show proofs attached to any `Credential` objects in the examples to keep things clear.  I think this is what @msporny and I think is the right direction.
2. Alternately, I think @jandrieu believes that there may still be times where we might want to use `VerifiableCredential` without a `proof` attached.  I'll admit I don't understand this though; it seems like since we agreed that we'll keep `Credential` that those are the use cases where we'd use `Credential`... otherwise why the distinction?

David-Chadwick:
I agree with option 1. ie. always use the type VerifiableCredential in the examples.

But whether it has a proof attached or not depends upon the schema of a VerifiableCredential, and I do not know where or how this is defined. In particular whether some properties are mandatory or optional in VCs. For example, a VerifiableCredential may or may not contain a TermsOfUse property as it is optional. But the type is still VerifiableCredential in both cases. The same goes for expiry time and a whole bunch of properties.

So I would like to know how or where the schema of a verifiable credential is defined

cwebber:
@David-Chadwick We could define a schema that we recommend applications test against for VerifiableCredential for the properties we know and expect, but we have to leave it open world for the properties we don't know about or understand.  I don't think that aside from the `proof` field, which maybe is mandatory for `VerifiableCredential`, that conversations about the rest of the schema probably applies to this PR though.

David-Chadwick:
In which case I will raise an issue about the more general issue of mandatory and optional properties

msporny:
Where are we wrt. this PR? I think we all agreed to use "VerifiableCredential" everywhere in the spec, but still have "Credential" in the JSON-LD Context?

jandrieu:
I think there is some serious confusion here.

"VerifiableCredential", IMO, should be used in all JSON-LD examples, NOT "everywhere in the spec". Replacing "credential" everywhere in the spec with "VerifiableCredential" would, IMO, make it unreadable, confusing, and inaccessible to most readers.

I believe there is consensus on this first part: All JSON-LD examples should use VerifiableCredential as the value of the "type" property at the top level of any Verifiable Credential.

I don't think you mean to suggest that the *rest* of the spec also gets Credential replaced with VerifiableCredential, but maybe you do.

Could you clarify?

msporny:
> I don't think you mean to suggest that the rest of the spec also gets Credential replaced with VerifiableCredential, but maybe you do.

I am not suggesting that. Just an update to all of the code examples.

msporny:
> I believe we are better off if VCs--as a data format--can support both proofed and unproofed claims, allowing the protocols, storage, ingestion, and representation to use a common underlying schema regardless of the need for a signature in the use case.

@jandrieu -- this feels like we need an issue to track this. There are a few nuances in what you're saying where I don't think we have had those nuanced discussions yet. The best way to raise the issue may be to suggest exactly what text should be updated/removed from the spec.

As for the rest of this PR, I think we're good to go, so merging.

PR #232, opened 2018-09-14
Remove Pearson reference and Varn

stonematt:
Remove Pearson reference from co-chair Matt Stone and delist Richard Varn as co-chair


burnburn:
If you fix the spelling of my last name (Burnett) then I am fine with this and you can merge.

PR #254, opened 2018-10-28
Add examples JSON-LD Context file. Fix #245.

### LINKED ISSUE #245: Should terms like ageOver be in the context?
(See also #231.)

We currently have examples using such terms as `ageOver` which are not currently in the context.  We could argue either way whether they they should or should not be:

* Maybe the credentials context should be minimal enough just include the base functionality.  In that case, we should define another context for common terms, or at least include some `example.org` type imaginary context in such examples in the spec to make clear that these are not defined terms.
* Alternately maybe we know everyone is going to use these a lot, so it's worth including some of the most common ones in the spec.

(While working on the test suite I converted the suite over to the latest context and discovered these terms were not in there, so currently I have "dumped them" back in, even though this does not reflect the current context, so I'll need to update that once we make a decision.)

msporny:
Fix #245.

### DIFF
--- a/vocab/examples.jsonld
+++ b/vocab/examples.jsonld
@@ -1,0 +1,13 @@
+{
+  "@context": [
+    {
+      "@version": 1.1
+    },
+    "https://w3id.org/security/v2",
+    {
+      "ex": "https://example.org/examples#",
+      "name": "ex:name",
+      "ageOver": "ex:ageOver"
+    }
+  ]
+}


BigBlueHat:
Related to issue #245 (apparently GitHub doesn't "sniff" the title for issue numbers).

msporny:
Hold on review by @brentzundel until Nov 1st.

msporny:
> It’s safe to include `@version` in the second context rather than having it come first.

I just did it there to make it clear that VCs should be using JSON-LD 1.1.

Issue #207, opened 2018-07-24
Rename "claim" to "subject"


msporny:
@David-Chadwick would like to change "claim" to "subject", viz:

```js
{
  "id": "http://dmv.example.gov/credentials/3732",
  "type": ["Credential", "ProofOfAgeCredential"],
  "issuer": "https://dmv.example.gov/issuers/14",
  "issued": "2010-01-01T19:73:24Z",
  "claim": {
    "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
    "ageOver": 21
  },
```

TO:

```js
{
  "id": "http://dmv.example.gov/credentials/3732",
  "type": ["Credential", "ProofOfAgeCredential"],
  "issuer": "https://dmv.example.gov/issuers/14",
  "issued": "2010-01-01T19:73:24Z",
  "subject": {
    "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
    "ageOver": 21
  },
```

Related to #120

stonematt:
I have a few thoughts on this suggestion.

1. This change might reduce confusion between "credential" and "claim" - which is a positive
2. This change could confuse the term "subject" as it relates to "subject != holder" - I don't think this is the same "subject" and subject is already too overloaded as a term in the wide world.
3. Finally it's very late in the process to make such a substantial change in fundamental terminology in our group

As a result, it I would vote against this request, unless someone can make a compelling argument for it.

David-Chadwick:
I think the term subject is very clear in the context of VCs. The subject is the subject of the credential. This is exactly the same as the use of subject in X.509 as well. The solution to this is to ensure that the definitions section clearly defines subject, and clearly defines holder.

Concerning the timing of this issue, I would like to make two points.
1. It has been an issue for many months, so it not new. At the time it was discussed several months ago, @dlongley said he was sympathetic to this change. But other issues (e.g. the PR on subject NE holder) took precedence for the group, so it got parked until I raised it again.
2. We should be concerned about the long term use of VCs and their use becoming ubiquitous. Consequently our objective should be for the data model to be understood by the widest group of users. Anything we can do now to make the data model more understandable to the average person should be welcomed with open arms, before it goes for review by other W3C groups such as the security group.

msporny:
Since we're bikeshedding the name, other choices include:

- assertion
- attestation
- affirmation
- declaration

[...]

PR #255, opened 2018-10-28
Add section on syntactic sugar. See #207 and #268.

msporny:
Addresses https://github.com/w3c/vc-data-model/issues/268 and https://github.com/w3c/vc-data-model/issues/207.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/255.html" title="Last updated on Dec 4, 2018, 9:17 PM GMT (9ed1551)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/255/bdce6b9...9ed1551.html" title="Last updated on Dec 4, 2018, 9:17 PM GMT (9ed1551)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2487,12 +2487,12 @@ stronger type guarantees for use cases that require them.
           </li>
           <li>
 If processed using a JSON-LD processor, the <code>verifiableCredential</code>
-and <code>claim</code> properties are treated as "graph containers", which
-are mechanisms that are used to isolate what one entity claims from
-another entity. This ensures proper cryptographic separation between the data
-graph provided by each <a>issuer</a> and the one provided by the
-<a>presenter</a> to ensure the provenance of the information for each graph
-is preserved.
+and <code>proof</code> properties are treated as "graph containers", which
+are mechanisms that are used to isolate sets of data asserted by different
+parties entities. This ensures, for example, proper cryptographic separation
+between the data graph provided by each <a>issuer</a> and the one provided by
+the <a>presenter</a> to ensure the provenance of the information for each
+graph is preserved.
           </li>
         </ul>
 
--- a/index.html
+++ b/index.html
@@ -2459,6 +2459,46 @@ built on top of this specification.
       </section>
 
       <section>
+        <h2>Syntactic Sugar</h2>
+
+        <p>
+In general, the data model and syntaxes described in this document are
+designed around a "cargo cult programming" mentality, where developers copy and
+paste examples to add verifiable credentials into their software systems.
+The design goal of this approach is to provide a low barrier to entry while
+still ensuring global interoperability between a heterogenous set of software
+systems. This section describes some of these approaches, which will most
+likely go unnoticed by most developers, but whose details will be of
+interest to implementers. The most noteworthy syntactic sugars used throughout
+the ecosystem are as follows:
+        </p>
+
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1980,6 +1980,46 @@ <h2>Disputes</h2>
...
+stronger type guarantees for use cases that require them.
+          </li>
+          <li>
+If processed using a JSON-LD processor, the <code>verifiableCredential</code>
+and <code>claim</code> properties are treated as "graph containers", which
+are mechanisms that are used to isolate what one entity claims from
+another entity. This ensures proper cryptographic separation between the data
+graph provided by each <a>issuer</a> and the one provided by the
+<a>presenter</a> to ensure the provenance of the information for each graph
+is preserved.
```
dlongley: We might want to say more here about how a "claim" is really a separate graph of statements made by the issuer via a credential.

David-Chadwick: This is really the wrong way to go. I do not believe we should have claims in the RDF. see #207

dlongley: @David-Chadwick, I'm currently exploring if it's possible to drop the `@graph` container (to at least see if it's a viable option) by checking to see what kind of JSON-LD frame would properly separate the claim information from the proof information.

msporny:
> "are designed such that developers may easily copy and paste examples to add ..."

Fixed.

msporny:
> * It adds another JSON-LD concept (graph containers) thereby increasing the comprehension distance between a JSON user and the specification. We should not need this concept in VCs.

You need to be able to make statements about statements -- that's what graph containers enable. If you can't do this, the information model becomes non-deterministic and the whole ecosystem falls down. Graph containers (or a concept like it) are not optional.

Keep in mind that developers don't need to know any of this. They just slot the data in the right place. I added this PR because a number of people wanted to know the details... and now that the details have been added, it sounds like you're saying: The details are not necessary.

> * Since a VC only contains one cryptographic proof, regardless of the number of claims, then the issue of graph separation is irrelevant. (The same applies for a presentation, which only has one signature.)

VCs don't always contain just one cryptographic proof. For example, VCs that are issued by a group (like a board of directors on a particular vote) may contain N signatures.

> * It is not code implementors I am particularly concerned about as the SDK implementors are professionals who should know what they are doing. Application implementors should use SDKs with well defined APIs so this is not a particular concern either. It is the millions of users who will define their own VCs that are of concern. They need to be able to easily understand how the JSON of a VC is composed. What are the (simple) rules for this, and how can I define new VCs and new properties of existing VCs so that they conform to the (largely hidden) rules of JSON-LD. In other words, we should have a simple set of rules (a subset of JSON-LD rules) for how to create correct JSON for VCs, which conform to JSON-LD. This should necessarily remove many of the rules that JSON-LD supports (such as graphs). In other words, all VCs will be valid JSON and JSON-LD, but not all valid JSON-LD will create valid VCs.

While I agree with a subset of what you say above... I need concrete text that I can apply to this PR to make progress.

index.html
msporny: This was done, thus resolving @David-Chadwick's concern.

msporny:
I believe I have addressed all of @David-Chadwick's concrete concerns. Merging. If a mistake was made, an issue or PR to fix would be appreciated.

PR #257, opened 2018-10-28
Add section on Authorization. Fix #204.

### LINKED ISSUE #204: Delegated authorization & VC distribution
PR #198 has some bits about delegation for the sake of authorization.  AFAICT there are actually two different "delegation" discussions here, and we should address them seperately:

1. Delegation in terms of authorization, where a VC is being used for an authorization mechanism, and party A wants to allow party B to act on their behalf.
2. "Delegation" in terms of Holder A wants to share a VC with Holder B, but the issuer maybe did or didn't want them to do that.  I suggest we call this "distribution" instead of "delegation".

-----

So, as for 1, I think that VC should not be addressing authorization *at all*.  Not everyone in the group agrees that identity-based authorization is safe, and VCs are all about identity.  We should keep VCs to being a mechanism where we know that party A said something about party B.  Whether we trust A, whether that means that B should be able to do things... those are things which are out of scope of VCs.

Now, I know that @David-Chadwick has pointed out that even if we don't want people to use VCs for authorization, we can't stop them from doing it and they probably will to build something RBAC like, and honestly I think that's true and even fine.  But we don't need to add any structure for that inside of the VC model spec.  The good news is we're using an extensible model, if you want to add an extension for specifying delegation of *authorization*, you can do that in a separate spec.

-----

As for 2, I think it's important to recognize that a Holder is just... anyone holding a VC!  You don't need to authorize someone to be a Holder, they become a Holder by holding onto that VC.  And mathematically, there's no way to prevent someone from copying and sharing a VC with another entity.  But it makes perfect sense to *express intent* that you would not like a VC to be copied and shared.  If I have a medical condition and I share it with you, it's perfectly valid for me to say, "don't tell anyone else about this, it's for your eyes only".  From a computing perspective, I can't stop you from doing so, though from a social and legal perspective, I may be able to punish you for doing so if I find out (and a well behaving receiving holder may be able to notice, "oops, I wasn't supposed to see this... I'll be nice and burn it and throw it in the trash", but there is no guarantee structurally that such a receiver will be well behaved).

I think that's what terms of service is for IMO.  We can make a special property for it if we like, but I think it'll still be a special case ToS.

msporny:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/257.html" title="Last updated on Oct 28, 2018, 2:29 PM GMT (92ea371)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/257/04e1533...92ea371.html" title="Last updated on Oct 28, 2018, 2:29 PM GMT (92ea371)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2047,14 +2047,18 @@ as the specific vocabulary terms for disputed claims.
         <h2>Authorization</h2>
 
         <p>
-It is arguable that <a>verifiable credentials</a> or
-<a>verifiable presentations</a> should be used as authorization mechanisms
-that a <a>holder</a> could use to access various
-systems. This specification does not contemplate such a usage of verifiable
-credentials and MUST NOT be considered an authorization framework on its own.
-The Working Group did consider authorization use cases
-during the creation of this specification and is pursuing that work as an
-architectural layer that is built on top of this specification.
+<a>Verifiable credentials</a> are intended as a means of reliably identifying
+subjects. Whilst it is recognized that Role Based Access Controls (RBAC) and
+Attribute Based Access Controls (ABAC) rely on this identification as a means
+of authorizing subjects to access resources, this specification
+MUST NOT be used for authorization purposes without an accompanying
+authorization framework.
+        </p>
+
+        <p>
+The Working Group did consider authorization use cases during the creation of
+this specification and is pursuing that work as an architectural layer that is
+built on top of this specification.
         </p>
 
       </section>
--- a/index.html
+++ b/index.html
@@ -2044,6 +2044,22 @@ as the specific vocabulary terms for disputed claims.
       </section>
 
       <section>
+        <h2>Authorization</h2>
+
+        <p>
+It is arguable that <a>verifiable credentials</a> or
+<a>verifiable presentations</a> should be used as authorization mechanisms
+that a <a>holder</a> could use to access various
+systems. This specification does not contemplate such a usage of verifiable
+credentials and MUST NOT be considered an authorization framework on its own.
+The Working Group did consider authorization use cases
+during the creation of this specification and is pursuing that work as an
+architectural layer that is built on top of this specification.
+        </p>
+
+      </section>
+
+      <section>
         <h2>Conformance</h2>
 
         <p>


David-Chadwick:
Sorry but I still have problems with this wording, specifically "This specification does not contemplate such  a usage (i.e. an authorization mechanism) of verifiable credentials".  If you read the Use Cases document, you will that the use case invariably use VCs for authorisation purposes. Quote "From educational records to payment account access, the next generation of web applications will authorize entities to perform actions based on rich sets of credentials issued by trusted parties." So how we can assert that we do not contemplate that VCs will be used for authorisation when our use case document says they will be? This really is a Kafkaesque world.

So lets be honest. We do envisage that VCs will be used for authorisation, but we recognise that additional controls will be needed in order to provide a fault proof authorisation framework built on top of the data model. If we are not to provide a full set of controls in V1 of the specification, then we need to say that this is what we have not done. (Rather than pretending that VCs will not be used for what we know they will be used for)

index.html
```diff
@@ -1980,6 +1980,22 @@ <h2>Disputes</h2>
...
       </section>
 
+      <section>
+        <h2>Authorization</h2>
+
+        <p>
+It is arguable that <a>verifiable credentials</a> or
+<a>verifiable presentations</a> should be used as authorization mechanisms
+that a <a>holder</a> could use to access various
+systems. This specification does not contemplate such a usage of verifiable
```
brentzundel:
```suggestion
systems. This specification MUST NOT be considered an authorization framework on its own.
```

index.html
```diff
@@ -1980,6 +1980,22 @@ <h2>Disputes</h2>
...
+credentials and MUST NOT be considered an authorization framework on its own.
```
brentzundel:
```suggestion
```

index.html
David-Chadwick: How about

Verifiable credentials are intended as a means of reliably identifying subjects. Whilst it is recognized that role based access controls (RBAC) and attribute based access controls (ABAC) rely on this identification as a means of authorizing subjects to access resources, this ?recommendation|specification? should not be used for authorization purposes without an accompanying authorization framework. This is because complex authorization issues such as delegation of authority are not addressed in this recommendation.

msporny: +1, I like this.

msporny: @brentzundel only issue with your MUST NOT is that it's not testable and we try to not put RFC language on non-testable things. That said, my point is a nitpick and I'd like us to take a stronger position on it... so we could just put it in and remove the language if anyone complains.

msporny:
Applied both of @David-Chadwick and @brentzundel's suggestions.

Issue #206, opened 2018-07-23
Default namespace URL


burnburn:
From @ianbjacobs  on 20 July 2018:

> I have one observation: section 9.2 (JSON-LD) defines a default namespace URL which is https://w3id.org/credentials/v1 
> 
> Would it be possible to use a W3C namespace instead following this policy?
>  https://www.w3.org/2005/07/13-nsuri

msporny:
> Would it be possible to use a W3C namespace instead following this policy?

The URL for the credentials JSON-LD context has a number of requirements:

1. It's easy for developers to remember.
2. It can be dereferenced, and the file may be requested hundreds of millions to billions of times a day.
3. The hosting provider for the URL finds the second requirement acceptable.

When discussing this with W3C Staff before (I think we talked w/ @sandhawke the most), it was the third item that was difficult to resolve. Note that we expect most uses of this URL to aggressively cache the document and not go out to the network (most implementations ship with a static file instead of going out to the network)... but there is always the chance that a big implementation won't do that and thus hammer the URL.

I would expect the W3C VCWG to welcome a JSON-LD context at any of the following URLs:

https://www.w3.org/ns/credentials/v1
https://w3.org/2018/credentials/v1

@ianbjacobs, do you think either of the two URLs above are workable given the requirements listed above?

ianbjacobs:
Hi @Msporny,

Thank you for enumerating requirements. I would add a fourth (if acceptable): 
 
 * W3C should host the resource.

We have had experience with widely referenced resources (namely, DTDs) and managed them reasonably well despite rogue software that should have cached but didn't.

I recommend that the Working Group write a proposal for the Team Contact to bring to the management team, since I am hearing this is a question of resources and maintenance.

I hope that helps,

Ian

msporny:
> I recommend that the Working Group write a proposal for the Team Contact to bring to the management team

Thanks @ianbjacobs, that gives us a concrete next step.

@stonematt and @burnburn, can we put this on an upcoming VCWG agenda?

msporny:
ping -- Chairs of VCWG:
> @stonematt and @burnburn, can we put this on an upcoming VCWG agenda?

burnburn:
From TPAC 2018:
- [ ] Reference proposed URLs (https://w3.org/2018/credentials/v1 and https://w3.org/2018/credentials) in the specification
- [ ] Request proposed URLs above from W3C web team
[...]

PR #258, opened 2018-10-28
Finalize the context URL for the spec to address #206.

msporny:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/258.html" title="Last updated on Oct 28, 2018, 2:29 PM GMT (79c29b4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/258/04e1533...79c29b4.html" title="Last updated on Oct 28, 2018, 2:29 PM GMT (79c29b4)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -744,7 +744,7 @@ information about the identifier.
 
         <pre class="example nohighlight" title="Usage of the 'id' property">
 {
-  "@context": "https://w3id.org/credentials/v1",
+  "@context": "https://w3.org/2018/credentials/v1",
   <span class="highlight">"id": "http://dmv.example.gov/credentials/3732"</span>,
   "type": ["VerifiableCredential", "ProofOfAgeCredential"],
   "claim": {
@@ -1381,7 +1381,7 @@ new <code>favoriteFood</code> term so that it may only be used within the
   "@context": {
     "referenceNumber": "https://example.com/vocab#referenceNumber",
     <span class="highlight">"claim": {
-      "@id": "https://w3id.org/credentials#claim",
+      "@id": "https://w3.org/2018/credentials#claim",
       "@context": {
         "favoriteFood": "https://example.com/vocab#favoriteFood"
       }
@@ -2266,7 +2266,7 @@ enters the Candidate Recommendation stage.
     "aud": "www.example.com",
     "sub": "did:example:ebfeb1f712ebc6f1c276e12ec21",
     "verifiableCredential": {
-      "@context": "https://w3id.org/security/v1",
+      "@context": "https://w3.org/2018/security/v1",
       "id": "http://example.gov/credentials/3732",
       "type": ["VerifiableCredential", "ProofOfAgeCredential"],
       "issuer": "https://dmv.example.gov",
--- a/index.html
+++ b/index.html
@@ -678,7 +678,7 @@ document containing machine-readable information about the context.
 {
   <span class="highlight">"@context": [
     "https://w3.org/2018/credentials/v1",
-    ...
+    "https://example.com/examples/v1"
   ]</span>,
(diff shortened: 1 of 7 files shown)


gkellogg:
Do we need to make corresponding changes to the namespace documents, which include the context file?

msporny:
> Do we need to make corresponding changes to the namespace documents, which include the context file?

Yes, we do. :) -- this PR needs to be updated to do that.

msporny:
Hold on review by @brentzundel

index.html
```diff
@@ -653,6 +651,70 @@ <h1>Basic Concepts</h1>
...
+people communicate; both people must use the same language and the words
+that each of them use must mean the same thing to the other party. This
+specification uses the <code>@context</code> property to express the context
+of the conversation.
+        </p>
+
+        <dl>
+          <dt><dfn>@context</dfn></dt>
+          <dd>
+The value of this property MUST be one or more URIs where first URI is the
```
brentzundel: The meaning of a verifiable credential could change if its context changes. I'm concerned that providing only a URI (which by its nature is a mutable reference), we are not requiring any sort of guarantee that the context pointed to by the URI won't change over time.

David-Chadwick: This is also my concern, especially since it is proposed to move Issuer Terms of Use to the @context from the VC itself. The advantage of this move is that the holder of a ZPK credentials cannot hide the terms of use, which he could if it were an assertion in the VC. The disadvantage is that the verifier has to rely on the contents of URLs being immutable. Can we propose some crypto to guarantee that if the contents change, the verifier can detect this.

gkellogg:
This is an issue the JSON-LD WG expects to tackle, although is not solved just yet. See w3c/json-ld-syntax#9 and the [TPAC F2F Minutes](https://www.w3.org/2018/json-ld-wg/Meetings/Minutes/2018/2018-10-26-json-ld#section7-2).

index.html
msporny: We can't solve this issue in this PR. As @gkellogg mentioned, it's a separate broader Web/JSON-LD ecosystem issue. There are options, but we are unlikely to settle on a standard before the VCWG is done. I know that @dmitriz is going to document all the options over in the JSON-LD WG's issue repository.

msporny:
Review received by @brentzundel, only issue raised is one we can't deal with in this PR (it's a broader ecosystem issue). Made changes requested by @gkellogg. Merging.

vocab/credentials.html
```diff
@@ -10,7 +10,7 @@
     specStatus:  "base",
     shortName:   "vc-vocab",
     publishDate: "2018-07-25",
-    thisVersion: "http://w3id.org/credentials",
+    thisVersion: "http://w3.org/2018/credentials",
```
davidlehn: Should this be https?

vocab/template.html
```diff
@@ -10,7 +10,7 @@
     specStatus:  "base",
     shortName:   "vc-vocab",
     publishDate: "<%=ont["dc:date"]%>",
-    thisVersion: "http://w3id.org/credentials",
+    thisVersion: "http://w3.org/2018/credentials",
```
davidlehn: https?

davidlehn:
How relevant is the namespace advice from https://www.w3.org/2005/07/13-nsuri?  That would imply URIs should be `http://www.w3.org/2018/ssss` vs `https://w3.org/2018/ssss`.

msporny:
> How relevant is the namespace advice from https://www.w3.org/2005/07/13-nsuri? That would imply URIs should be http://www.w3.org/2018/ssss vs https://w3.org/2018/ssss.

I'm hoping not very relevant. :P

PR #261, opened 2018-10-29
Use IPLD as a JSON syntax for claim

jonnycrunch:
I'd like y'all to have a serious look at IPLD for VC (as well as the DID document).  He is my draft document from the recent Rebooting Web of Trust arguing my point.

https://github.com/WebOfTrustInfo/rwot7/blob/master/draft-documents/ipld_did_documents.md

While I do like JSON-LD, having cryptographic resolution of the payload is important in this application. 

Only change here that is necessary is to reserve the key "/" for a CID and resolution over IPLD. 

```
"@context" : 
    { "/" : "zdpuAmoZixxJjvosviGeYcqduzDhSwGV2bL6ZTTXo1hbEJHfq" }, 
   ... 
 "creator" : 
    { "/" : "zdpuAvqt1YTeAdcyyncBDdJLVgKyZNbHm17rBEJNqFzVbg24B" }
```

This also binds the entire claim as a self-describing content addressed hash.  Much more secure to DNS poisoning attack. 

I don't like to argument that we have settled on JSON-LD as the only solution for a VC.  

Also, see my PR for the DID-spec: 

https://github.com/w3c-ccg/did-spec/pull/110#issuecomment-431356177

***
<a href="https://pr-preview.s3.amazonaws.com/jonnycrunch/vc-data-model/pull/261.html" title="Last updated on Mar 5, 2019, 1:19 PM UTC (a3d1090)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/261/04e1533...jonnycrunch:a3d1090.html" title="Last updated on Mar 5, 2019, 1:19 PM UTC (a3d1090)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2180,11 +2180,11 @@ content indentifier (CID) for content addressing resolved using
  <p>The following example demonstrates how one could
           express this data model using IPLD.</p>
 
-  <pre class="example" title="A verifiable claim with @context and previous linked data using IPLD">
+  <pre class="example" title="A verifiable claim with @context and creator of the claim as linked data using IPLD">
 { "@context" : 
     { "/" : "zdpuAmoZixxJjvosviGeYcqduzDhSwGV2bL6ZTTXo1hbEJHfq" }, 
     ... 
-    "previous" : 
+    "creator" : 
     { "/" : "zdpuAvqt1YTeAdcyyncBDdJLVgKyZNbHm17rBEJNqFzVbg24B" }
     ... 
 }
--- a/index.html
+++ b/index.html
@@ -2148,6 +2148,51 @@ JSON-LD document, then a <code>@context</code> property with a value of
 
       </section>
 
+  <section>
+        <h2>IPLD</h2>
+        
+        <p>
+Content addressing through hashes has become a widely-used means of connecting 
+data in distributed systems. <a href="https://ipld.io">IPLD</a> is a way of representing 
+hash-linked data to be used in content-addressed data retrieval systems such as 
+<a href="https://ipfs.io">IPFS</a>. Other content that can be resolved using IPLD include 
+blockchains such as Bitcoin, Ethereum, ZCash and Git repositories. IPLD enables 
+creation of decentralized data-structures that are universally addressable facilitating 
+resolving content accross different protocols. It achieves this through an interoperable 
+data model that represents various protocol formats. IPLD relies on self-describing 
+<a href="https://github.com/ipld/cid">Content Identifiers (CIDs)]</a> for content addressing. 
+CIDs are a self-describing, flexible, and interoperable way of expressing cryptographic hashes. 
+It uses several multiformats to achieve flexible self-description, namely 
+<a href="https://github.com/multiformats/multihash"> multihash </a> for hashes, 
+<a href="https://github.com/multiformats/multicodec"> multicodec </a> for data content types, and 
+<a href="https://github.com/multiformats/multibase">multibase </a> to represent the base encoding 
+of the CID itself. This interoperability makes IPLD a valuable structure for the DID document 
+that can be used across a variety of DID methods or distributed ledgers and 
+ensures cryptographic validity of the DID document.  Since both IPLD and JSON-LD 
+are 100% compatible with JSON, the large number of JSON parsers and libraries are already available. 
+      </p>
+      <p>
+The key symbol <code>"/"</code> (foward slash) SHOULD be reserved to represent a self describing 
+content indentifier (CID) for content addressing resolved using 
+<a href="https://github.com/ipld/specs">Interplanetary Linked Data (IPLD)</a>.  
+      </p>
+
+ <p>The following example demonstrates how one could
+          express this data model using IPLD.</p>
+
+  <pre class="example" title="A verifiable claim with @context and previous linked data using IPLD">
+{ "@context" : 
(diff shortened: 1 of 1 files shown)


burnburn:
Hi @jonnycrunch have you joined the Working Group?  I don't see you in the participants list (https://www.w3.org/2000/09/dbwg/details?group=98922&public=1&order=org).  You can check here (https://www.w3.org/2004/01/pp-impl/98922/join) to see if your org is a member.

BigBlueHat:
> While I do like JSON-LD, having cryptographic resolution of the payload is important in this application.

@jonnycrunch fwiw, solving for these scenarios is something we're exploring more broadly in the JSON-LD WG. We have a broad issue for [content addressable `@context` files](https://github.com/w3c/json-ld-syntax/issues/9). That might be a better layer to explore this at--since it would effect the "document loading" options available to JSON-LD syntax processors. See also the [Remote Document and Context Retrieval](https://w3c.github.io/json-ld-api/#remote-document-and-context-retrieval) section of the JSON-LD API.

Short version, though, is that the exact JSON structure above won't work because you'd be defining a "term" of `/`--which isn't your intent. A URI-based structure for IPLD is probably a more generally usable thing that would avoid collision with existing identifier and term spaces.

But let's continue the chat over at https://github.com/w3c/json-ld-syntax/issues/9 😄

jonnycrunch:
@burnburn I was an invited expert to the group in the past before it became a working group.  

@BigBlueHat BigBlueHat I think the syntax should apply for VC/DID syntax more broadly first and take on the argument as an enhancement to JSON-LD `@context` in particular later.  

 @msporny  the reason I don't like the URI format i.e. `ipfs://<hash>` is that it allows for cycles in the graph and therefore NOT a true DAG, which opens it up for attack.   Allowing `{"/" : "<cid"}` syntax for VC/DID truly binds the meaning of the data.   As I suggested in the PR,  the "/" `SHOULD` be reserved keyword for IPLD syntax for cryptographic validation.  The proof of its success would be in its adoption and implementation.

msporny:
> @burnburn I was an invited expert to the group in the past before it became a working group.

@jonnycrunch -- I think that @burnburn is pointing out the fact that you're attempting to contribute something that isn't covered under a W3C Intellectual Property Release... and we need that sorted before we can accept substantive changes from you. So, either 1) your company needs to join W3C, or 2) the Chairs need to add you as an Invited Expert to the Working Group. Both processes take multiple weeks to perform and are there to protect the Web from patent trolls (which you are not, but the W3C Membership takes this sort of thing very seriously and does not make exceptions).

> @BigBlueHat BigBlueHat I think the syntax should apply for VC/DID syntax more broadly first and take on the argument as an enhancement to JSON-LD `@context` in particular later.

The VCWG's Charter does not allow us to make the sort of changes that you are proposing to how the JSON-LD Context would be interpreted. Only the JSON-LD WG is cleared to make those sorts of changes, of which @BigBlueHat is the co-Chair of that group. I'd take him up on his offer to help out.

> @msporny the reason I don't like the URI format i.e. `ipfs://<hash>` is that it allows for cycles in the graph and therefore NOT a true DAG, which opens it up for attack. 

I don't understand how using a different format for the identifier opens us up to attack? What's the attack? 

> Allowing `{"/" : "<cid"}` syntax for VC/DID truly binds the meaning of the data. As I suggested in the PR, the "/" `SHOULD` be reserved keyword for IPLD syntax for cryptographic validation. 

I don't understand how doing this is any different than doing ipfs:<cid> - it feels like if one of them is susceptible to attack, the other approach would be as well.

> The proof of its success would be in its adoption and implementation.

Standards typically happen in the opposite way... you prove a successful approach first via demonstrating adoption/implementation and THEN you standardize on that. Standardizing on something that's theoretical first often leads to failed standards (and is why there are so many failed "standards" these days).

msporny:
@jonnycrunch how would you like us to proceed on this PR? There are a number of people above that are attempting to engage w/ you on alternatives. I don't think this PR is going to make it through, can we close it?

jonnycrunch:
I am requesting to reopen this PR to consider IPLD as a valid JSON type.

joeltg:
> the reason I don't like the URI format i.e. ipfs://<hash> is that it allows for cycles in the graph and therefore NOT a true DAG, which opens it up for attack.

Can you elaborate on this? How do `ipfs://` or `dweb:/` permit cycles? What is the attack? The IPFS folks have a nice [writeup](https://github.com/ipfs/in-web-browsers/blob/master/ADDRESSING.md) about recommended IPFS URIs for different contexts, and it seems like that's a much better conceptual fit for use in JSON-LD.

jonnycrunch:
so this isn't a format for JSON-LD but a JSON format that is on par with JSON-LD.   in this case, I am using Interplanetary Linked Data ( IPLD ) https://ipld.io not ipfs.  IPFS uses IPLD under the covers.   IPLD is a serialized dag CBOR (among others ).  While ipfs:// is on track for ietf schema name, ipld:// is not.  IPLD and JSON-LD are both a type of JSON.   This PR is for IPLD to be considered under the JSON heading, NOT as a valid JSON-LD.  IPLD is like JSON-LD but with cryptographic hash linking to the content via the "/" path syntax.

burnburn:
@jonnycrunch this is a new feature that we can't consider until the IPR issues are worked out, and feature freeze was last November anywy.  Leaving open but marking as deferred so it can be considered for a future version.

PR #263, opened 2018-11-05
Clarified URIs may be references to verifiable data registry

brentzundel:

### DIFF
--- a/index.html
+++ b/index.html
@@ -668,6 +668,7 @@ of the conversation.
           <dd>
 The value of this property MUST be one or more URIs where first URI is the
 value <code>https://w3.org/2018/credentials/v1</code>.
+The required URIs MAY be a reference to an object in a verifiable data registry, rather than a DNS-rooted link.
 If more than one URI is provided, the URIs MUST be interpreted as an
 ordered set. It is RECOMMENDED that dereferencing the URI results in a
 document containing machine-readable information about the context.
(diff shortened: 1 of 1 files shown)


ken-ebert:
+1 
By adding language to enable URIs to reference a verifiable data registry, the data model allows the Sovrin use case to occur.

davidlehn:
Saying the properties are URIs should be enough to indicate the values can be DNS or non-DNS scheme based.  Why is this more specific new language needed?

BigBlueHat:
@ken-ebert that's been the case from the beginning as URI's have always been in place in this spec--and in any JSON-LD-based data format fwiw.

Identifiers are not Locators, and Locators don't necessarily use DNS in their locating, so I'm concerned this particular content would be more confusing, not less.

URI is sufficiently defined in [RFC3986](https://tools.ietf.org/html/rfc3986), so perhaps what's needed here is some `<abbr>` tags around URI and a link to RFC3986 in the reference section.

brentzundel:
I don't have a problem closing this PR if PR-265 is merged, as the examples added there clarify the issue sufficiently for me.

msporny:
> I don't have a problem closing this PR if PR-265 is merged, as the examples added there clarify the issue sufficiently for me.

Waiting on @brentzundel to answer questions on #265 so we can merge that one. 

I suggest we close this PR as a URI does not mean DNS-rooted link so it's not necessary to point that out... I don't think we need to hold off on #265 (as I assume it'll be merged in some form or another that addresses what this PR was meant to address).

@brentzundel can we close this PR?

brentzundel:
That is fine.

Issue #266, opened 2018-11-06
Provide support for JWT data format


awoie:
As per section 1.3 Use Cases and Requirements:
> Verifiable credentials and verifiable presentations are expressed in one **or more** standard, machine-readable data formats which can also be extended with minimal coordination.

As per section 7.1.1 JSON Web Token:
> This section will be moved into its own specification before this document enters the Candidate Recommendation stage.

This indicates that the JWT section is going to be disregarded. While JSON and JSON-LD are useful options in many cases, there would be a lack of supporting the well-known and mature JWT industry standard. 

As per RFC 7519 JSON Web Token (JWT):
> JSON Web Token (JWT) is a compact claims representation format

Consequently, the specification does not support other formats than JSON and JSON-LD at the moment. 

The W3C VC specification should support JWT to ensure backwards-compatibility with existing solutions and easier adoption of the specification due to JWT's wide distribution, support by many frameworks and maturity of existing libraries.

A suitable mapping of the W3C VC data model onto JWT should reuses standard JWT claims as much as possible to allow seamless integration. In my mind, this could be considered as "cr-blocker" and "discuss".

burnburn:
PR merged, so I suggest closing.  Clearly other PRs (such as #366) are working on updates, but that's fine.

Please re-open if necessary.

Fak3:
> Clearly other PRs (such as #366) are working on updates, but that's fine.

@burnburn there is no such pr

burnburn:
Weird.  I don’t know what that was supposed to  be referencing, but the
point was that edits can still continue. We no longer need an issue to tell
us to add JWT support now that initial text is in the spec.

PR #267, opened 2018-11-06
Addressing issue #266 by providing a mapping of the data model for JWT

awoie:
This is an initial PR to address issue #266 by providing a mapping of the data model for JWT.

***
<a href="https://pr-preview.s3.amazonaws.com/awoie/vc-data-model/pull/267.html" title="Last updated on Dec 15, 2018, 8:24 PM UTC (7358723)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/267/a7ec4d4...awoie:7358723.html" title="Last updated on Dec 15, 2018, 8:24 PM UTC (7358723)">Diff</a>


awoie:
I'm a member of the W3C VC WG. Could someone please fix this?

Thanks,
Oliver

awoie:
I'm a member of W3C VC WG and I've connected my W3C account to my Github handle. Could some please provide some help on how to fix this?

Thanks,
Oliver

index.html
```diff
@@ -2187,7 +2191,7 @@ <h2>Syntaxes</h2>
 by example using the JSON syntax. This section formalizes how the data model
 (described in Sections <a href="#core-data-model"></a>,
 <a href="#basic-concepts"></a>, and <a href="#advanced-concepts"></a>
-) is realized in JSON and JSON-LD. Although syntactic mappings are only provided for these two syntaxes,
+) is realized in JSON, JSON-LD and JWT. Although syntactic mappings are only provided for these three syntaxes,
```
msporny: Editorial: "JSON-LD, and" -- we do oxford comma 'round these parts. :)

index.html
```diff
@@ -2322,6 +2258,169 @@ <h2>JSON-LD</h2>
...
+
+            <p>
+                The verifiable credential data model as described in Section <a href="#core-data-model"></a> can be
+                encoded as a JSON Web Token (JWT) [[!RFC7519]] by doing the following:
+            </p>
+
+            <ul>
+                <li>The signature of the JWT is produced by the issuer of the verifiable credential.</li>
+                <li>Because JWS implicitly provides a verifiable proof of the contained verifiable credentials, the "proof" is no longer needed. Please note that
+                    the "jti" claim represents a nonce value.</li>
```
msporny: There are two ways that we could go with this: Add the JWT keywords (jti, iat, etc.) to the JSON-LD Credentials vocab... or leave them out, and move the JSON-LD Context down one level.

The question this raises is: Is the top level just an encapsulating container? Or is it a VC? If it's  just a container, where does evidence and termsOfUse go? I expect we'll end up with the top level being a VC, which then begs the question of whether or not JSON-LD processing is required or not during verification to ensure the semantic mappings between the sending and receiving application are checked.

index.html
```diff
@@ -2322,6 +2258,169 @@ <h2>JSON-LD</h2>
...
+                <li>The IANA JWT claims registry should be enforced whenever possible:</li>
```
msporny: Does this mean that the VC Core Data Model properties should be added to the IANA JWT claims registry? What happens with `evidence` and `termsOfUse`, for example?

index.html
```diff
@@ -2322,6 +2258,169 @@ <h2>JSON-LD</h2>
...
+                <ul>
+                    <li>The "expirationDate" becomes the JWT "exp" claim (epoch time).</li>
+                    <li>The "id" of the verifiable credential becomes the JWT "jti" claim.</li>
+                    <li>The "id" contained in the "claim" object becomes the JWT "sub".</li>
+                    <li>If "created" and "creator" are required, these claims could be added to the JWT payload.</li>
+                    <li>The "issuer" becomes the JWT "iss" claim.</li>
+                    <li>The "issuanceDate" becomes the JWT "iat" claim.</li>
```
msporny: Where is the nonce and domain specified? If you don't do that, the message can be replayed. This is confusing layers of abstraction... the VC Data Model spec isn't supposed to talk about the crypto... I don't think the group is chartered to do that. We may need to find non-normative weasel wording to make these sorts of assertions since certain W3C Member companies requested that we NOT make statements like the above.

index.html
```diff
@@ -2322,6 +2258,169 @@ <h2>JSON-LD</h2>
...
+            <pre class="example nohighlight" title="Presentation JWT payload">
+{
+  "@context": [
+    "https://w3.org/2018/credentials/v1",
+    "https://example.com/examples/v1"
+  ],
+  "jti": "urn:uuid:3978344f-8596-4c3a-a978-8fcaba3903c5",
+  "iss": "did:example:ebfeb1f712ebc6f1c276e12ec21",
+  "aud": "https://dmv.example.gov/issuers/14",
+  "nonce": "343§$FSFDa-",
```
msporny: What prevents replay across multiple domains?

index.html
```diff
@@ -2322,6 +2258,169 @@ <h2>JSON-LD</h2>
...
+yeGhhVzBpT25zaVlXZGxUM1psY2lJNk1qRjlmUS5UUXltWms2Z0IyekRhUmVFNlVUYTlLQnA2MVRlX21
+0bzl3V1JDZDFDcWNkQWpOMFNZMTItVWhPbHN4em4tWGhaRE50R2RUd3pZX1ZVUklPR2JiQkROZ25HYlR
+zdHJ3bjgtU29oX3FpZzJsc3k3ZlM1ckF6QmdtZGFmSkxiYUtTQ0VsSGJ5RElsMVZ4NFFmYzk0LVdOVWJ
+pQ1RTaEp3dENJN204MmxMWjltSmdmIl19.EjTFSFneWu2duLhy8DE1to943Shl_u70sSy8BSLBLlr10K
+Q1dzyb-L8P14luAd9Jk-8bNkqI5XuqJiFF2mgZuEwYugh98U0I5d7g2jlOqeXWHMAYaKEELfUsiNn5Bj
+AqEp58JEN7e-pSWYHCX49BsQmQumDsnDFBDzIHoTQe3no"]
+}
+          </pre>
+
+            <p class="issue">
```
msporny: I'm supportive of pulling this PR in as long as:

1. There are multiple implementers - who are the multiple implementers for this PR?
2. We are very clear about the limitations of VCs encoded as JWTs - signatures are tied to a very specific serialization where whitespace destroys the signature, signatures are tied to JSON, base64 encoding makes developers have to use twice as much storage in the simples case, set and chained signatures are problematic, no support for other types of mathematical proof formats, if JSON-LD processing is not done there is no guarantee that the semantics of the message is well formed, etc. We need a very large warning section here so that developers know what they're getting themselves into when using JWTs.

Digital Bazaar's position is that combining VCs with JWTs are a footgun, we won't be implementing it, but we also don't want to stand in the way of other companies that still want to go down that path in spite of all of the warnings.

index.html
```diff
@@ -2322,6 +2258,169 @@ <h2>JSON-LD</h2>
...
+            <pre class="example nohighlight" title="Verifiable credential JWT header">
+{
+  "alg": "RS256",
+  "typ": "JWT"
+}
+          </pre>
+
+            <pre class="example nohighlight" title="Verifiable credential JWT payload">
+{
+  "@context": [
```
msporny: What is to prevent the @context from using different values such that JSON-LD processors interpret the message differently?

index.html
```diff
@@ -2322,6 +2258,169 @@ <h2>JSON-LD</h2>
...
+    "https://w3.org/2018/credentials/v1",
+    "https://example.com/examples/v1"
+  ],
+  "sub": "did:example:ebfeb1f712ebc6f1c276e12ec21",
+  "jti": "http://dmv.example.gov/credentials/3732",
+  "iss": "https://dmv.example.gov/issuers/14",
```
msporny: How is the issuer key bound to the message?

index.html
dlongley: I think `aud` could be used to do that, but here it appears to be an issuer URL. But perhaps this VP is actually being sent to the same issuer that the VC was issued from? If so, it may confuse the reader.

msporny:
>> I'm supportive of pulling this PR in as long as:
>> 1. There are multiple implementers - who are the multiple implementers for this PR?
> 
> @msporny , that has not been a required bar to date for any feature, and this came in before feature freeze. That said, if there are insufficient implementations when we reach our final CR review date we will of course need to pull it out at that time, as with any of our features that are insufficiently implemented.

Yep, sorry, that's what I meant.

So, I suggest we:

1. Mark this as a feature at risk so we can get it into the spec.
2. Outline when it's safe to use the feature and when you may want to choose a different approach.

index.html
awoie: Yes, it is not the best example. The intention was to use `aud` for that. I will update the example.

index.html
msporny: Yes, that should work, I think.

index.html
```diff
@@ -2322,6 +2258,169 @@ <h2>JSON-LD</h2>
...
+    "https://example.com/examples/v1"
+  ],
+  "jti": "urn:uuid:3978344f-8596-4c3a-a978-8fcaba3903c5",
+  "iss": "did:example:ebfeb1f712ebc6f1c276e12ec21",
+  "aud": "https://dmv.example.gov/issuers/14",
+  "nonce": "343§$FSFDa-",
+  "iat": "1541493724",
+  "exp": "1573029723",
+  "type": ["VerifiablePresentation"],
+  "verifiableCredential": ["
```
msporny: This should probably change to `verifiableCredentialJwt` as the value is encoded.

index.html
```diff
@@ -2322,6 +2258,169 @@ <h2>JSON-LD</h2>
...
+          </pre>
+
+            <pre class="example nohighlight" title="Verifiable credential JWT payload">
+{
+  "@context": [
+    "https://w3.org/2018/credentials/v1",
+    "https://example.com/examples/v1"
+  ],
+  "sub": "did:example:ebfeb1f712ebc6f1c276e12ec21",
+  "jti": "http://dmv.example.gov/credentials/3732",
```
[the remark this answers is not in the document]
msporny: Just looked at rfc7519, `sub` and `jti` are optional, so they don't in and of themselves ensure correlation. There are no ZKP-style signature mechanisms for JWTs, so I'd be interested in hearing how they support the various types of non-correlation identified in the security/privacy section.

index.html
awoie: True, `nonce` has also to be included. Domain would be the `iss` claim. I am now aware of the group charter. It would be good to have a non-normative example similar to the JSON LD representation where the `nonce` is part of the `proof` attribute.

index.html
awoie: This is true. We might end up removing the `@context` from the JWT data model. Let's discuss that during our call.

index.html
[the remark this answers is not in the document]
awoie: One approach will be to add the `proof` to the JWT if something else than digital signatures is required. In this case, a ZKP representation could have no JWS but include the `proof` attribute.

msporny:
Given all of the discussion, I've tried to capture the sets of features JWTs and LD-Proofs enable: https://docs.google.com/spreadsheets/d/1WJ1pOH6zmZacuWgoAVfIHLI9Q1EPXbScY44w2gqv9vo/edit

index.html
```diff
@@ -2322,6 +2258,169 @@ <h2>JSON-LD</h2>
...
+
+            <ul>
+                <li>The signature of the JWT is typically produced by the holder of the enclosed verifiable credentials.</li>
+                <li>Because JWS implicitly provides a verifiable proof of the contained verifiable credentials, the "proof" is no longer needed.</li>
+                <li>The presentation keeps the JWS of the verifiable credentials intact by including the base64 encoded verifiable credentials in the "verifiableCredentials" array.</li>
+                <li>The IANA JWT claims registry should be enforced whenever possible:</li>
+                <ul>
+                    <li>The "id" of the presentation becomes the JWT "jti" claim.</li>
+                    <li>A "nonce" claim MUST be included to mitigate the risk of replay attacks.</li>
+                    <li>A "aud" claim may be included and contains the subject of the verifier. The "aud" claim MUST be one or more URIs.</li>
```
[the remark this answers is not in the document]
awoie: My understanding was to not make any statements on security. Same applies to nonce. I will update the PR accordingly.

index.html
awoie: I added the `nonce`to the VP and VC examples. Please note, the `jti`serves as a nonce in the VC already. In general, we don't talk about specific risk mitigation strategies and everybody is free to implement their own means using what the data model allows.

index.html
awoie: I updated the examples.

index.html
awoie: I added the changes we discussed during the our JWT call. I hope, this will resolve the discussion.

index.html
awoie: In my mind, key discovery is out of scope of this specification. We are talking about a data model and not about protocols. People could implement that either by DID and DID Document resolution based on the `iss` or even based JWK. These would be part of a different spec.

index.html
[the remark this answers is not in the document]
awoie: The PR does not support native JWT integration of ZKP, or complex proofs but however, we can include a JSON/JSON-LD verifiable credential/presentation, we would be able to make use of the `proof`.

index.html
awoie: I updated the the JWT data model to reflect the outcome of our JWT discussion. The changes should address this issue.

index.html
```diff
@@ -2322,6 +2258,169 @@ <h2>JSON-LD</h2>
...
+              }
+            }
+            </pre>
+
+            <p class="issue">
+                Add a short description explaining the example above. Also mention that @context might be omitted.
+            </p>
+
+            <p class="issue">
+                Add an example of how a base64 JWT encoded verifiable credential looks like.
```
dlongley: `how` => `what`

index.html
```diff
@@ -2322,6 +2258,169 @@ <h2>JSON-LD</h2>
...
+              "iat": "1541493724",
+              "exp": "1573029723",
+              "nonce": "343§$FSFDa-",
+              "vcp": {
+                "@context": [
+                  "https://w3.org/2018/credentials/v1",
+                  "https://example.com/examples/v1"
+                ],
+                "type": ["VerifiablePresentation"],
+                "vc_jwt": ["..."]
```
dlongley: `vc_jwt` should be `verifiableCredential`.

index.html
```diff
@@ -2322,6 +2258,169 @@ <h2>JSON-LD</h2>
...
+
+            <pre class="example nohighlight" title="JWT payload of a JWT based verifiable presentation (non-normative)">
+            {
+              "iss": "did:example:ebfeb1f712ebc6f1c276e12ec21",
+              "jti": "urn:uuid:3978344f-8596-4c3a-a978-8fcaba3903c5",
+              "aud": "did:example:4a57546973436f6f6c4a4a57573",
+              "iat": "1541493724",
+              "exp": "1573029723",
+              "nonce": "343§$FSFDa-",
+              "vcp": {
```
dlongley: Should `vcp` be `vp` to match `vc`?

index.html
```diff
@@ -2322,6 +2258,169 @@ <h2>JSON-LD</h2>
...
+                        Key discovery is out of scope of this specification. For example, the <code>kid</code> can
+                        refer to a key in a DID Document, or can be the identifier of a key inside a JWKS.</li>
+                    <li><code>typ</code> MUST be set to JWT.</li>
+                </ul>
+            </p>
+            <p>
+                For backwards compatibility with JWT processors the following JWT registered claim names MUST be used
+                instead or in addition of their respective verifiable credential counterparts:
+                <ul>
+                    <li><code>exp</code> SHOULD represent expirationDate.</li>
```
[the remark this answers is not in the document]
pelle: I agree with MUST, but if applicable

index.html
```diff
@@ -2322,6 +2258,169 @@ <h2>JSON-LD</h2>
...
+                Add an example of how a base64 JWT encoded verifiable presentation looks like.
+            </p>
+
+            <p class="issue">
+                A number of the concerns have been raised around security,
+                composability, reusability, and extensibility with respect
+                to the use of JWTs for Verifiable Credentials. These concerns
+                will be documented in time in at least the Verfiable Claims Model
+                and Security Considerations section of this document.
+            </p>
```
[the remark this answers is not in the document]
pelle: @msporny I recommend having that in a separate document an not in the spec.

index.html
```diff
@@ -2322,6 +2258,169 @@ <h2>JSON-LD</h2>
...
+                can be used to represent more complex proofs e.g., if the creator is different from the issuer, or a
+                <code>proof</code> which is not based on digital signatures such as proof of work. The issuer MAY
+                include both, a JWS and a <code>proof</code> property. For backwards compatibility reasons, the issuer
+                MUST use JWS to represent proofs based on a digital signature.
+            </p>
+            <p>
+                In addition, the following JOSE headers MUST be present in the JWS:
+                <ul>
+                    <li><code>alg</code> MUST be used for RSA and ECDSA based digital signatures.</li>
+                    <li><code>kid</code> MUST be used if there are multiple keys associated with the issuer of the JWT.
```
pelle: I don't think this is needed. It's an option but adds extra verbosity. Our implementation checks the DID document and tests each public key. This works fine in almost all use cases and removes duplication.

msporny:
@awoie, can you please respond to @pelle so we can get this merged?

index.html
```diff
@@ -2322,6 +2258,169 @@ <h2>JSON-LD</h2>
...
+              "iat": "1541493724",
+              "exp": "1573029723",
+              "nonce": "660!6345FSer",
+              "vc": {
+                "@context": [
+                  "https://w3.org/2018/credentials/v1",
+                  "https://example.com/examples/v1"
+                ],
+                "type": ["VerifiableCredential", "ProofOfAgeCredential"],
+                "credential_subject": {
```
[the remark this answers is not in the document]
awoie: I know that camelCase is the preferred way of naming attributes with JSON-LD. But I was wondering if it makes sense to define a context for snake_case, or even have shorter attribute names? During W3C, guys from the JWT community were a bit concerned about the size of the credential.

index.html
awoie: agree

index.html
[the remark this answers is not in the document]
awoie: agree

index.html
```diff
@@ -2322,6 +2258,169 @@ <h2>JSON-LD</h2>
...
+            <p class="issue">
+                Add a short description explaining the example above.
+            </p>
+
+            <pre class="example nohighlight" title="JWT payload of a JWT-based verifiable credential using JWS as a proof (non-normative)">
+            {
+              "sub": "did:example:ebfeb1f712ebc6f1c276e12ec21",
+              "jti": "http://dmv.example.gov/credentials/3732",
+              "iss": "did:example:abfe13f712120431c276e12ecab",
+              "iat": "1541493724",
```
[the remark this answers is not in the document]
awoie: I agree. Applies to `iat` and `exp`.

index.html
```diff
@@ -2322,6 +2258,169 @@ <h2>JSON-LD</h2>
...
+                MUST use JWS to represent proofs based on a digital signature.
+            </p>
+            <p>
+                In addition, the following JOSE headers MUST be present in the JWS:
+                <ul>
+                    <li><code>alg</code> MUST be used for RSA and ECDSA based digital signatures.</li>
+                    <li><code>kid</code> MUST be used if there are multiple keys associated with the issuer of the JWT.
+                        Key discovery is out of scope of this specification. For example, the <code>kid</code> can
+                        refer to a key in a DID Document, or can be the identifier of a key inside a JWKS.</li>
+                    <li><code>typ</code> MUST be set to JWT.</li>
```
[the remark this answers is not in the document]
awoie: agree

index.html
```diff
@@ -2322,6 +2258,169 @@ <h2>JSON-LD</h2>
 
       </section>
 
+        <section>
+            <h2>JSON Web Token</h2>
+            <p>
+                JSON Web Token (JWT) [[!RFC7519]] is still a widely used means to express claims to be transferred
```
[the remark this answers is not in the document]
awoie: agree

index.html
```diff
@@ -2322,6 +2258,169 @@ <h2>JSON-LD</h2>
...
+              "exp": "1573029723",
+              "nonce": "660!6345FSer",
+              "vc": {
+                "@context": [
+                  "https://w3.org/2018/credentials/v1",
+                  "https://example.com/examples/v1"
+                ],
+                "type": ["VerifiableCredential", "ProofOfAgeCredential"],
+                "credential_subject": {
+                  "age_over": 21
```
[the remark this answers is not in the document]
awoie: See my answer above.

index.html
awoie: agree

index.html
awoie: - Agree with MUST. 
- I also would prefer to see this in a separate document. 
- `kid` MAY be used.

awoie:
Answered the questions and updated the PR. Also addressed merge conflicts.

Thanks,
Oliver

index.html
```diff
@@ -2787,9 +2787,10 @@ <h3>JWT and JWS Considerations</h3>
       "https://w3.org/2018/credentials/v1",
       "https://example.com/examples/v1"
     ],
-    "type": ["VerifiableCredential", "ProofOfAgeCredential"],
-    "credentialSubject": {
-      "ageOver": 21
+    "type": ["VerifiableCredential", "UniversityDegreeCredential"],
+    "degree": {
```
davidlehn: This VC should have a "credentialSubject" property that contains the "degree" property and an "id" that matches "sub" above.

index.html
pelle: 👍

Issue #240, opened 2018-10-08
Inconsistency in presentation


David-Chadwick:
In section 2 'presentation' is defined as follows

Data derived from one or more credentials, issued by one or more issuers, that is shared with a particular entity.

then 'verifier' is defined as follows

A role an entity may perform by receiving one or more verifiable presentations for processing.

These two definitions are inconsistent because presentation implies that any type of entity can receive a presentation, whilst verifier says that an entity is a verifier if it receives a presentation.

SOLUTION

change the presentation definition to state

Data derived from one or more verifiable credentials, issued by one or more issuers, that is shared with a verifier.

Note the above is consistent with the definition of verifiable credential.

jandrieu:
Hmmm... I like this. It is a modest correction that I believe improves readability and clarity.

However, our definitions are still a little off.

Technically the verifier is the one who verifies the integrity of the credential, which may or may not be the entity who receives the presentation from the presenter/holder. One could imagine needing to send a presentation through an intermediary for any number of reasons. The important thing is not the receipt of the presentation from the creator of that presentation, but rather the evaluation of a presentation for authorial integrity (determining, to the best of our ability, whether or not the utterances embedded in the presentation were made by the alleged issuers and do those issuers still stand by them). There are a lot of situations where the relying party with outsource the verification check, where the verifier actually doesn't rely on the utterance in the claim at all; they merely pass on the results of their evaluation to the actual party using the claims.

We have similar conflicts with holder/presenter. The distinguishing characteristic of the holder is that they take one or more credentials--perhaps ones they created, but they could be issued *to* them, given to them by an ally, or literally found online, published publicly--and create a presentation that is, by definition, a self-asserted relating of those credentials to themselves. That is, the holder is acting as an issuer when creating a presentation with the unique characteristic that the verifier is not expected to accept those assertions based on authority, but rather based on the evaluation of the presented information. (Whereas issuers' claims *are* typically evaluated based on their authority, e.g., the DMV). 

[...]

PR #272, opened 2018-11-11
Update presentation definition

stonematt:
Update definition of presentation to call out presenting to a verifier specifically. Fixes issue #240


David-Chadwick:
LGTM

PR #279, opened 2018-11-15
Add note on transitive trust to Trust Model section. Fix #252.

### LINKED ISSUE #252: Review & Clarify Trust Model Section re: No Transitive Trust & Delegated Keys
Related regards to portions conversation at TPAC about "Terms of Use", issues of failures of CA-like transitive trust models and delegated key usage was brought up.

We should review the the Trust Model section and possibly clarify that no transitive trust is implied by the data model (validators only choose to trust an issuer or not, no CA-like delegation), and that it is responsibility of signatures suites to define any limitations on usage re-usage of key material.

msporny:
This attempts to talk briefly about why we didn't adopt a CA-based trust model (but don't preclude one from being used once the implementer is aware of the downsides of such a system).


msporny:
Review request ping @ChristopherA.

ChristopherA:
Text is good. I might recommend this 2005 link: https://www.cs.cornell.edu/people/egs/papers/dnssurvey.pdf

msporny:
> I might recommend this 2005 link: https://www.cs.cornell.edu/people/egs/papers/dnssurvey.pdf

Done.

David-Chadwick:
Whilst making changes to the trust model, no 3 needs changing to 
3. The subject and the verifier trust the issuer....
This was pointed out during our discussions leading up to TPAC


Actually I don't think the DNS survey paper is very good from the perspective of understanding the problem domain, because it does not describe what the underlying cause it. It only describes the symptoms. If you look at Figure 1 there is no explanation of why these dependencies exist. There is no legend so we do not know the meanings of the names in oblong boxes. We do not know which are primary name servers, which are secondary servers or why for example, cornell.edu should contain cayuga.cs.rochester.edu. Without this explanation we have no idea how to resolve the problem. Russ Housley's paper on the other hand is more up to date and more relevant.

msporny:
> Russ Housley's paper on the other hand is more up to date and more relevant.

Is that the IETF paper that I cited? Or some other paper?

David-Chadwick:
Yes it is the one you cited

index.html
```diff
@@ -667,6 +667,20 @@ <h1>Trust Model</h1>
 <a>relying party</a>, a more flexible and dynamic trust model is created
 such that market competition and customer choice is increased.
       </p>
+
+      <p class="note">
+Experts may recognize that the data model detailed in this specification
```
davidlehn: Should avoid qualifying readers as "experts" or not.  Perhaps remove that first part and make it a statement like "The data model detailed in this specification..."

David-Chadwick:
Did the following change make it into your PR or not

3. The subject and the verifier trust the issuer....

This was pointed out during our discussions leading up to TPAC

Issue #241, opened 2018-10-08
clarification of extensibility


David-Chadwick:
The current text in section 6.1 says:

This approach to data modelling is often called an "open world assumption," meaning that any entity can say anything about any other entity. This approach often feels in conflict with building simple and predictable software systems. Balancing extensibility with program correctness is always more challenging with an open world assumption than it is with closed software systems. 

In order to be in tune with the rest of the data model document, and preceding text in the document, I suggest we change this text to read:

This approach to data modelling is often called an "open world assumption," meaning that any entity can make any claim about any other entity. This approach often feels in conflict with building simple and predictable software systems. Balancing extensibility with program correctness is always more challenging with an open world assumption than it is with closed software systems. This is why section 5.1 makes strong statements about the type property and mandates that the types of credentials, presentations, and encapsulated objects MUST be specified or associated with additional, more narrow types.

burnburn:
From TPAC 2018:
- [ ] Needs to go into Implementation Guidance NOTE

PR #280, opened 2018-11-15
Add section on Data Schemas. Related to #241.

msporny:
This is an attempt to address #241 and some of the ZKP requirements in the specification.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/280.html" title="Last updated on Dec 4, 2018, 3:40 PM GMT (c0f0b67)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/280/99b0dc9...c0f0b67.html" title="Last updated on Dec 4, 2018, 3:40 PM GMT (c0f0b67)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1435,7 +1435,7 @@ verifiable credential to an alternative representation format such as a
 binary format used in a zero-knowledge proof.
         </p>
 
- <p>
+        <p>
 The expression of a data schema may be performed via the following
 <a>property</a>:
         </p>
@@ -1447,7 +1447,7 @@ The expression of a data schema may be performed via the following
             determine if the provided data is conformant to the provided schema.
             Each <code>credentialSchema</code> comprises at least its
             <code>type</code> (e.g. <code>JsonSchemaValidator2018</code>),
-            and a <code>id</code> property that MUST be a URI identifying
+            and an <code>id</code> property that MUST be a URI identifying
             the schema file. The precise contents of each
             data schema is determined by the specific type definition.
           </dd>
@@ -1478,8 +1478,13 @@ The expression of a data schema may be performed via the following
         <p>
 In the example above, the <a>issuer</a> is specifying a
 <code>credentialSchema</code> that points to a JSON Schema file that can be
-used by <a>verifier</a> to determine if the <a>verifiable credential</a> is
-well-formed.
+used by a <a>verifier</a> to determine if the <a>verifiable credential</a> is
+well formed.
+        </p>
+
+        <p class="issue">
+The group is still exploring how to utilize this feature for ZKPs and will
+most likely adjust this section based on implementer feedback.
         </p>
 
         <p>
--- a/index.html
+++ b/index.html
@@ -1422,6 +1422,104 @@ new <code>favoriteFood</code> term so that it may only be used within the
       </section>
 
       <section>
+        <h2>Data Schemas</h2>
+
+        <p>
+Data schemas are often useful when enforcing a particular structure on a given
+collection of data. There are at least two types of data schemas that this
+specification contemplates. The first are data validation schemas, which are
+used to verify that the structure and contents of a
+<a>verifiable credential</a> are conformant to a published schema. The second
+are data encoding schemas, which are used to map the contents of a
+verifiable credential to an alternative representation format such as a
+binary format used in a zero-knowledge proof.
+        </p>
+
+ <p>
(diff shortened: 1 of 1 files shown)


David-Chadwick:
LPGTMM (looks pretty good to me Manu)

msporny:
> LPGTMM

*lol* :laughing:

index.html
```diff
@@ -1421,6 +1421,104 @@ <h3>Semantic Interoperability</h3>
...
+        </p>
+
+        <dl>
+          <dt><var>credentialSchema</var></dt>
+          <dd>The value of this property MUST be one or more data schemas
+            that provide <a>verifiers</a> with enough information to
+            determine if the provided data is conformant to the provided schema.
+            Each <code>credentialSchema</code> comprises at least its
+            <code>type</code> (e.g. <code>JsonSchemaValidator2018</code>),
+            and a <code>id</code> property that MUST be a URI identifying
```
dlongley: typo => `a <code>id` => `an <code>id`

index.html
```diff
@@ -1421,6 +1421,104 @@ <h3>Semantic Interoperability</h3>
...
+    "type": "JsonSchemaValidator2018"
+  }</span>,
+  "proof": { ... }
+}
+        </pre>
+
+        <p>
+In the example above, the <a>issuer</a> is specifying a
+<code>credentialSchema</code> that points to a JSON Schema file that can be
+used by <a>verifier</a> to determine if the <a>verifiable credential</a> is
```
dlongley: typo: `by <a>verifier` => `by a <a>verifier`

index.html
```diff
@@ -1421,6 +1421,104 @@ <h3>Semantic Interoperability</h3>
...
+well-formed.
```
dlongley: No hyphens needed for compound adjectives when they follow a noun, only when they precede one. Yay English.

index.html
```diff
@@ -1421,6 +1421,104 @@ <h3>Semantic Interoperability</h3>
...
+  "proof": { ... }
+}
+        </pre>
+
+        <p>
+In the example above, the <a>issuer</a> is specifying a
+<code>credentialSchema</code> that points to a zero-knowledge packed binary
+data format that is capable of transforming the input data into a format
+that can then be used by a <a>verifier</a> to determine if the
+proof provided with the <a>verifiable credential</a> is valid.
```
dlongley: This section might need a little tweaking to make it clear if we're looking at the verifiable credential that's directly from the issuer or not.

If we are, then the text needs to be tweaked to avoid sounding like the `proof` here is being presented when, in fact, a whole new credential must be derived and presented. I imagine, in this case, that the holder/presenter would need to include the `credentialSchema` in a derived credential in the presentation such that the verifier could retrieve it from a registry and confirm its validity... and then use it to encode the data presented in the derived credential so that the ZKP-based proof (on the derived credential) could be verified.

If we're not, we shouldn't have things like the credential ID in there and so on.

@brentzundel ... could you take a look at this language here and see if it needs tweaking and propose something that works?

brentzundel: It may need tweaking.
The notion of a data schema seems very similar to the objects that @ken-ebert and I have been working on, which we will be discussing next Monday. 
I propose we wait to merge this PR until that conversation, after which I hope we will all be on the same page.

index.html
```diff
@@ -1421,6 +1421,104 @@ <h3>Semantic Interoperability</h3>
...
+collection of data. There are at least two types of data schemas that this
+specification contemplates. The first are data validation schemas, which are
+used to verify that the structure and contents of a
+<a>verifiable credential</a> are conformant to a published schema. The second
+are data encoding schemas, which are used to map the contents of a
+verifiable credential to an alternative representation format such as a
+binary format used in a zero-knowledge proof.
+        </p>
+
+ <p>
```
davidlehn: Indent (for consistency)

index.html
msporny: Ok, on hold until after the ZKP call.

index.html

index.html

index.html

index.html
msporny: (╯°□°）╯︵ HSI⅂⅁ᴎƎ

index.html
msporny: I'm going to add an issue, merge this, and then modify it based on the end result of the ZKP discussion.

PR #283, opened 2018-11-16
Remove delegation

David-Chadwick:
***
<a href="https://pr-preview.s3.amazonaws.com/David-Chadwick/vc-data-model/pull/283.html" title="Last updated on Dec 18, 2018, 2:08 PM UTC (37d9ffa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/283/b75cec4...David-Chadwick:37d9ffa.html" title="Last updated on Dec 18, 2018, 2:08 PM UTC (37d9ffa)">Diff</a>


index.html
```diff
@@ -1777,150 +1818,30 @@ <h2>Credential Uniquely Identifies Subject</h2>
...
-
-
+<p>
+The data model allows for this, by the subject issuing a new verifiable
+credential and giving it to the new holder, so that the holder can present
+both verifiable credentials to the verifier. However, the content of this
+second verifiable credential is likely to be application specific, and therefore
+this specification does not standardise the contents of this second verifiable
+credential. Nevertheless a non-normative example is given in Appendix A.2
+      </p>
```
brentzundel: I do like this paragraph.

index.html
```diff
@@ -1733,18 +1733,59 @@ <h2>Subject is the Holder</h2>
...
+
+   <section>
+
+   <h3> 'Subject Only' Terms of Use </h3>
+
+  <p>
+The Subject Only Terms of Use states that a verifiable credential MUST only be
+presented to a verifier by the subject. If a verifier is presented with a
+verifiable credential containing the Subject Only Terms of Use, by anyone other
+    than the subject, it MUST refuse to accept it.
```
brentzundel: With selective disclosure, there is no way to ensure that a verifier ever sees the Subject Only Terms of Use.

dlongley: Please see the question here about this issue: https://github.com/w3c/vc-data-model/pull/282#issuecomment-440430172

msporny:
Merge conflicts are preventing the merge of this PR.

index.html
dlongley: Post ZKP discussion: The selective disclosure schemes being implemented do indicate that a ToU is present in a credential definition, allowing for a verifier to request it (and thereby require it). The presenter can also prove if their particular credential happens to not possess a ToU should it be requested but not present.

David-Chadwick: Great. So I will update the PR to remove conflicts with the latest version of the spec

David-Chadwick:
I believe I have now resolved the conflicts, and I have also replaced the subjectOnly ToU with subjecOnly property.

@msporny. I already posses two physical plastic cards that are marked subjectOnly by the issuer. So I am quite sure that this property will be a common use case for VCs.

msporny:
> @msporny. I already posses two physical plastic cards that are marked subjectOnly by the issuer. So I am quite sure that this property will be a common use case for VCs.

It's not the use case that I'm concerned about. It's the data modeling aspect of it, which I haven't been able to think about deeply. At present, almost all VCs our company is issuing are "subjectOnly" credentials, which are implicit and which the verifier determines whether or not to allow anyone except for the subject to use them.

David-Chadwick:
@msporny. I think you are introducing red herrings in your discussion above. 
First you do not hobble any system with our conformance clauses because we have already agreed that any verifier can accept anything at its own liability. So in this case the verifier would switch off any software checking that might be in the validation code (and be non-conformant to whichever protocol spec it was using). So breaking conformance as you suggest is not a data model issue, it is a usage and protocol issue which is out of the scope of our standard.

In one of my plastic cards, the local council is the issuer of my free bus pass, and it has EXPLICITLY stipulated Subject Only on it, to all the verifiers (buses) in the country. i.e. only I am entitled to free travel on buses with this VC. So 
a) it does not matter who the verifier is
b) it does not matter what the usage is  
b) it is not implicit it is explicit
A standard way of indicating this in the data model will be a common requirement.

David-Chadwick:
@msporny I don't know how the index.text file got there and I don't know how to delete it. Could you help please

msporny:
> A standard way of indicating this in the data model will be a common requirement.

Yes, agreed. I don't think the current way it's being expressed in the data model is the right way to do it... but don't have an alternate suggestion at this point (I will in time, but haven't been able to think about it deeply). In other words, not suggesting how to do it is better than suggesting the wrong way to do it.

msporny:
> @msporny I don't know how the index.text file got there and I don't know how to delete it. Could you help please

Fixed.

PR #291, opened 2018-11-23
Mode of operation

David-Chadwick:
A new section outlining the 8 steps in the VC ecosystem

***
<a href="https://pr-preview.s3.amazonaws.com/David-Chadwick/vc-data-model/pull/291.html" title="Last updated on Nov 26, 2018, 6:26 PM GMT (1fcc196)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/291/a318451...David-Chadwick:1fcc196.html" title="Last updated on Nov 26, 2018, 6:26 PM GMT (1fcc196)">Diff</a>


index.html
```diff
@@ -1497,6 +1497,92 @@ <h2>Refreshing</h2>
...
       </section>
 
+      <section>
+       <h2> Mode of Operation</h2>
+
+<em>This section is non-normative.</em>
+
+<p>
+Section <a href="#Ecosystem Overview"></a> provided an overview of the verifiable 
+credential eco-system. A more detailed look at the way the eco-system is envisaged
```
msporny: No dashes in "ecosystem".

index.html
```diff
@@ -1497,6 +1497,92 @@ <h2>Refreshing</h2>
 
       </section>
 
+      <section>
+       <h2> Mode of Operation</h2>
+
+<em>This section is non-normative.</em>
+
+<p>
+Section <a href="#Ecosystem Overview"></a> provided an overview of the verifiable
```
msporny: "#Ecosystem Overview" -> "#ecosystem-overview"

index.html
```diff
@@ -1497,6 +1497,92 @@ <h2>Refreshing</h2>
...
+<em>This section is non-normative.</em>
+
+<p>
+Section <a href="#Ecosystem Overview"></a> provided an overview of the verifiable 
+credential eco-system. A more detailed look at the way the eco-system is envisaged 
+to operate is as follows.
+</p>
+
+      <p>
+Note that the following ordering of steps is not fixed and each step may be repeated
```
msporny: "each step" -> "some steps"

index.html
```diff
@@ -1497,6 +1497,92 @@ <h2>Refreshing</h2>
...
+   <li>
+Step 4. The final holder presents its verifiable credentials to the verifier in 
+a verifiable presentation, requesting a supported action.
+  </li>
+   <li>
+Step 5. The verifier verifies the authenticity of the verifiable presentation
+and verifiable credentials.
+  </li>
+   <li>
+Step 6. The verifier verifies the right of the holder to posses the verifiable
```
msporny: "right" seems to be a bit strong here, but don't know what to change it to.

index.html
```diff
@@ -1497,6 +1497,92 @@ <h2>Refreshing</h2>
...
+Step 5. The verifier verifies the authenticity of the verifiable presentation
+and verifiable credentials.
+  </li>
+   <li>
+Step 6. The verifier verifies the right of the holder to posses the verifiable 
+credentials.
+  </li>
+   <li>
+Step 7. The verifier determines whether to accept the verifiable credentials for 
+the requested action, taking into account its policy, the holder and the contents
```
msporny: Add oxford comma after "the holder"

index.html
```diff
@@ -1497,6 +1497,92 @@ <h2>Refreshing</h2>
...
+specifications do specify how they are transferred between the various entities, then:
+  </p>
+
+  <ul>
+  <li>
+This verifiable credentials specification is directly applicable to steps 
+2, 3, 4, 5 and 6.
+  </li>
+  <li>
+This verifiable credentials specification is not sufficient on its own for
```
msporny: We will want to call out that those steps are typically defined by authorization frameworks.

index.html
```diff
@@ -1497,6 +1497,92 @@ <h2>Refreshing</h2>
...
+2, 3, 4, 5 and 6.
+  </li>
+  <li>
+This verifiable credentials specification is not sufficient on its own for 
+steps 1, 7, and 8.
+  </li>
+  </ul>
+ 
+<p>
+In particular, Section <a href="#Terms of Use"></a> and Section
```
msporny: "#Terms of Use" -> "#terms-of-use"

index.html
```diff
@@ -1497,6 +1497,92 @@ <h2>Refreshing</h2>
...
+  </li>
+  <li>
+This verifiable credentials specification is not sufficient on its own for 
+steps 1, 7, and 8.
+  </li>
+  </ul>
+ 
+<p>
+In particular, Section <a href="#Terms of Use"></a> and Section 
+<a href="#Subject-Holder Relationships"></a> specify how a verifier can
```
msporny: "#Subject-Holder Relationships" -> "#subject-holder-relationships"

index.html
```diff
@@ -1497,6 +1497,92 @@ <h2>Refreshing</h2>
...
+  </li>
+  </ul>
+ 
+<p>
+In particular, Section <a href="#Terms of Use"></a> and Section 
+<a href="#Subject-Holder Relationships"></a> specify how a verifier can 
+determine whether:
+</p>
+<p>
+- the holder is the subject of a verifiable credential,
```
msporny: These should be a bulleted list `<ul><li>...</li></ul>`

David-Chadwick: Done

index.html
David-Chadwick: Done

index.html
David-Chadwick: Done

index.html
David-Chadwick: Changed to
This verifiable credentials specification is not sufficient on its own for 
steps 1, 7, and 8 as an authorisation framework will be needed as well.

index.html
David-Chadwick: Done

index.html
David-Chadwick: changed it to
The verifier verifies that the holder may posses the verifiable 
credentials.

index.html
David-Chadwick: Done

index.html
David-Chadwick: done

index.html
David-Chadwick: done

index.html
dlongley: Perhaps it should say "Is authorized to possess"? "May possess" doesn't seem clear. Perhaps even "is authorized to present" would be better. Even that doesn't seem quite right -- it's not up to the verifier to make that determination. The verifier simply controls how it reacts to the use of the credential. So really, maybe this should be something like "The verifier verifies that the presentation of the credentials by the holder is acceptable for its intended purpose."

msporny: >  "The verifier verifies that the presentation of the credentials by the holder is acceptable for its intended purpose."

Yes, this.

David-Chadwick: "its intended purpose" is ambiguous. Is this the purpose of the verifier or the purpose of the issuer, or the purpose of the subject etc. Furthermore, the next step (7) checks what you are proposing. All step 6 is checking is if step 3 took place or not, and if it did, if it conforms to the standard.
How about "The verifier verifies that the holder may posses the verifiable credentials in conformance to the standard."
Rationale: We are testing whether the verifier conforms to the standard or not, not whether the verifier wishes to accept a VC or not. A verifier can accept all VCs and all VPs, even ones that have expired and that have been revoked.

burnburn:
@David-Chadwick in issue #252 you made the statement that "the DM model is 100% responsible for step 3".  How is that the case?

David-Chadwick:
@burnburn . If you read section 1.2, Ecosystem Overview you will see that it describes how VCs are issued, passed to the holder and from there to the verifier. The ecosystem forms part of the data model, it is the context of use, otherwise it would not be described informally in the standard. This is why I said it was a DM issue (perhaps more correctly I should have said it is an ecosystem issue, and the ecosystem is supported by, or is part of, the overall DM - for example, this is why a VC contains an issuer field).
Now, unless the DM states categorically that VCs cannot be passed between parties (and our use cases say just the opposite, that they are passed between parties) then if follows that the passing of VCs between holders is part of the ecosystem and is sanctioned by it.

Now the verifier has no knowledge about the actual issuing process, as it took place independently of, and before, the holder passed the VC to the verifier. The verifier does not know if the holder was the original holder or a subsequent holder, unless there are some properties in the DM that tell the verifier this. Furthermore, if the holder is a subsequent holder, is he/she a thief or was the VC passed on with the full knowledge and agreement of the original holder. Again the DM can inform the verifier about this, independently of the protocol that is used i.e. this is not a protocol issue, but an ecosystem issue, and the DM supports the ecosystem.

Issue #293, opened 2018-11-27
Remove language about "accepting" credentials


jandrieu:
In the recent PR #282, in response to issue #276, the language for Terms of Use described "accepting" a VC.

This term of art, "accepting" is ill-defined, confusing, and IMO, out of scope since authorization is out of scope.

The specification should state what VCs can and cannot guarantee. That is, what "verification" means for a verifiable credential. What we should state clearly is how verification is to occur and what verification means in terms of guarantees: cryptographic math and processes for currency (aka revocation).

Any other directive that presupposes what verifiers or issuers MUST do is out of scope, especially the current language for ToUs.

Terms of Use have no computational resolution. They are only a statement of policy as expressed by the issuer. Whether or not a Verifier accepts or rejects the ToU is technically independent of whether or not the Verifier accepts the credential. One is a technical step. The other a legal one.

It is fundamentally an unsolved problem to represent ToUs such that they are programmatically **and** legally enforceable. The actual resolution of any ToU **necessarily** depends on legal interpretation: first by the issuer (and their attorneys), then by the holder (and their attorneys), then by verifier (and their attorneys), then, ultimately, by any courts involved. As much as some members of this community would like to replace the legal system with smart contracts, this type of algorithmic capture, presentation, evaluation, acceptance, and enforcement of a legal agreement is completely out of scope.

What we **can** say is how to verify a credential. 

What we should NOT say is that the issuer has some magical ability to control what verifiers do. They do not. They cannot. That's the whole point of separating the issuer from the verifier.

Issuers can say anything they want. 

Verifiers can prove that what is presented is, in fact, a statement by the issuer, to the extent of cyptographic and procedural guarantees.

What issuers cannot do, and what we should not imply or state outright, is control the use or distribution of a verifiable credential once created and given to its initial holder. They do not get to tell verifiers what they MUST do, MAY do, or CAN'T do. Issuers don't get to decide whether or not verifiers accept a given credential for any specific purpose. They just get to decide whether or not to make the claims in that credential in a verifiable way. Once they do, the *use* of that credential is no longer their business. Period. Full stop.

To suggest that verifiable credentials somehow magically give issuers the ability to control whether or not verifiers accept or reject a VCs is a misstatement at best, an intentionally misleading hyperbole at worst, and will absolutely lead to false expectations, bad implementations, and issuers participating in systems that are fundamentally broken because VCs don't have that particular magic.

[...]

PR #298, opened 2018-11-28
Replaced language about requirements for "accepting" a verifiable credential

jandrieu:
This addresses issue #293.

I attempted to retain the clarifications about the goals of ToUs and the recipient's liability should they choose to ignore those terms. However in mentioning that liability, I pointed out the legal limits of assuming how and to whom such liability accrues. Perhaps we should excise that completely, but it was the best way I could find to describe the possible means of enforcement of a ToU, given there is no technical enforcement supported by VCs.

To wit, the previous language asserted the verifier would be responsible if a holder violated the terms of use under which a credential is issued. However, the verifier receives the information covered by the ToU **before** they are able to review, much less agree to, the asserted terms. (There is no protocol to enforce that VCs are only given to those who have agreed to such ToUs). In such a case, I don't believe the ToU will be found legally binding on the verifier. I did add a note about seeking reliable counsel, but maybe the whole section should just be removed.

I also attempted to couch the language about "accepting" in the subject != holder discussions such that the original intention remains. The actual examples do not, IMO, address the use cases, but they work as illustrations that "may" be appropriate for some use cases.  I initially started rewriting these to address my concerns, but that quickly got way too complicated to do properly in the timeframe we have.

FWIW, I'd start with having the original holder create a new credential with claims that explicitly say the new holder has been given the original credential for a specific purpose rather than including the original claims and implying the delegation without formally stating it. This gets into territory where VCs are decidedly challenging around the edge cases. In general, I am not a fan of quirky, non-explicit semantics of these use cases. If the thing is delegated, say it in claims rather than overload the semantics of the credentials themselves, which is simply that the issuer made an utterance. 

Similarly, I tried to shift the parent/child examples to be clear that the assertions of relationship are information for consideration when deciding whether or not to rely on a VC. I believe this retains the intention without implying that all verifiers MUST accept the DMV's assertion of a relationship as a means of appropriate delegation for their particular use. I think it also avoids the ambiguity between "verify" and "accept". Whether or not the DMV's assertion is suitable for the delegation is subject to the business rules of the verifier. For some cases, parentage doesn't imply any rights to delegation. For instance, money stored in a trust is often completely out of reach of the parent, even if the DMV is accepted as authoritative for the relationship asserted. As such, it's inappropriate to assert that a given arrangement of multiple credentials MUST be accepted by any given verifier for any given use.

I, for one, would not view the DMV as an authoritative source for a parenting relationship (nor for weight, height, or even current address). I attempted to frame these from the perspective that verifiers reliance on any VC needs to take into account several factors beyond verification, which is what I believe @David-Chadwick's means by "accept".

Anyway, I hope this fits the bill. I made a sincere effort to meet the use cases @David-Chadwick wants addressed while fixing my concerns about untestable conformance and dependence on solving demonstrably unsolved, out-of-scope problems.

Comments and suggestions welcome.

***
<a href="https://pr-preview.s3.amazonaws.com/jandrieu/vc-data-model/pull/298.html" title="Last updated on Dec 24, 2018, 4:50 PM UTC (a6d0c1e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/298/09fb38d...jandrieu:a6d0c1e.html" title="Last updated on Dec 24, 2018, 4:50 PM UTC (a6d0c1e)">Diff</a>


index.html
```diff
@@ -1866,12 +1851,15 @@ <h3> 'Subject Only' Terms of Use </h3>
...
+The Subject Only Terms of Use states that a verifiable credential is only to be presented by the subject. If a verifier is presented with a verifiable presentation containing a verifiable credential with a Subject Only Terms of Use, by anyone other
+    than the subject, it relies on that presentation on its own responsibility. Similarly, if a holder other than the subject presents such a presentation, they do so on their own responsibility.
      </p>
 
+	   <p class="note">
+There is an open question about whether or not the issuer can bind the verifier to an agreement embodied in a Terms of Use. This is especially problematic because the verifier is unable to view or commit to the terms before receiving the information covered by those terms. Necessarily, they receive the information before they can agree to the terms. In contrast, the holder is well aware of the Terms of Use prior to creating the presentation and therefore may be more likely to be held liable. However, this specification is not legal advice. Anyone relying on the legal foundation of a Terms of Use policy should seek the advice of competent counsel in the relevant jurisdictions.
+      </p>
```
David-Chadwick: Needs CR/LFs adding otherwise it is unreadable.

As a general point, verification software can be configured to know before hand which ToUs it supports and which it does not support. So it is perfectly feasible for a verifier's implementation to either accept or reject a VC after parsing it, checking its contents, and determining if it is acceptable or not. Your note does not adequately address the point that the verifier needs to read a VC and its ToU before deciding whether to accept them or not. Your note implies that the mere act of reading and parsing a VC, and then rejecting it, is somehow committing to the ToUs.

index.html
```diff
@@ -2572,22 +2555,23 @@ <h3>Fitness for Purpose</h3>
           <li>The custom properties in the <a>credential</a> are
           appropriate for the <a>verifier</a>'s purpose. For example if a
           verifier needs to determine that a <a>subject</a> is older than 21
-          years of age, they may accept claims of specific
+          years of age, they may relyin on claims of specific
```
David-Chadwick: relyin -> rely

index.html
```diff
@@ -1532,33 +1532,23 @@ <h2>Refreshing</h2>
...
           <dt><var>termsOfUse</var></dt>
-          <dd>The value of this property MUST be one or more terms of use that tell the 
-verifier what actions it either MUST perform (i.e. an Obligation) or MUST NOT perform
-(i.e. a Prohibition) or MAY perform (i.e. a Permission) if it is to accept the 
-verifiable credential or verifiable presentation. 
-          If the verifier is not willing to accept these terms of use then it MUST reject
-          the verifiable credential or verifiable presentation.
-          Each <code>termsOfUse</code> comprises its type,
+          <dd>The value of this property MAY specify one or more terms of use policies under which
+the creator issued the credential or presentation. If the recipient (holder or verifier) is not willing to adhere to the specified terms of use then they do so on their own responsibility and may incur legal liability if they violate the stated Terms of Use.
```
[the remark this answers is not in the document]
jandrieu: Ok. It wasn't clear to me that the language allowed for the property's absence.

How about
```suggestion
If present, the value of this property MUST specify one or more terms of use policies under which
the creator issued the credential or presentation. If the recipient (holder or verifier) is not willing to adhere to the specified terms of use then they do so on their own responsibility and may incur legal liability if they violate the stated Terms of Use.
```

index.html
jandrieu: Thanks for the point about linefeeds. I'll make an update that corrects for that.

Yes. You could whitelist ToUs. But that isn't in the specification and I doubt we'd have consensus on requiring that method.

But I think you missed my point because whitelisting doesn't address the ordering of agreement (or not) to the ToU after receiving the information.

Any information given to me before I agree to a ToU will NOT be covered by that ToU if I choose to reject the ToU. What you're arguing for is a contract of adhesion https://legaldictionary.net/adhesion-contract/, aka, a shrink-wrap license https://www.wilmerhale.com/en/insights/publications/the-origin-of-click-wrap-software-shrink-wrap-agreements-march-22-2000. Historically, the boundary on the bindability of shrink-wrap licenses required an explicit additional step after the presentation of the license before the reveal of the licensed information. VCs don't have that. They especially don't have any way to record acceptance or not.

Those articles highlight the problems both for courts and for organizations dealing with shrink-wrap licenses. To wit:

> It is important to note, however, that there will be situations in which Internet companies may be well-advised not to use click-wrap agreements. For example, click-wrap agreements are more suitable when end-users are expected to be individuals, rather than organizations. Potential problems arise when a company, rather than an individual, enters into a click-wrap agreement. In this situation, the online merchant must be careful to ensure that the individual clicking to accept has power and authority to accept on behalf of the company. 

Your proposed requirement would suggest that the standard itself binds the recipient of a VC to certain actions (non-acceptance of the VC) in the case they don't understand or don't like a ToU. In effect, your language would create a potential legal liability for recipients of a VC should they reject the ToU. 

But... what constitutes acceptance and non-acceptance? Does forwarding the VC to legal counsel for review of the ToU constitute acceptance? What about ingesting it into a database, flagged for review but not used in active services? Does not acceptance require immediate purging? What about log files used for security and capacity analysis? Is it in appropriate to retain a VC to try to understand why it triggered non-acceptance?

What are the business and legal tests for compliance with any given ToU? Is the simple phrase "Subject must equal Holder" supposed to encapsulate a formal legal framework recognizable by a court of law? What is the burden of proof that the Subject *is* the Holder? And who has it? How would a verifier meet that standard? Is username/password acceptable or does it require the mythical DID-Auth? If the DID keys are stolen, does that leave the verifier not liable for the breach? Can the verifier simply rely on the good faith assertion of the holder? The verifier doesn't have a legal relationship with the issuer, so why are they on the hook when a holder impersonates someone else?

What are the remedies for violations and who has grounds for action? Is it only the issuer who can seek damages? Or can the verifier seek damages from the holder for misrepresentation? Can the issuer actually go after a verifier--with whom they have no relationship--or is their only remedy to sue the holder masquerading as the subject? What about the subject? Can they go after the issuer or the holder when a VC ToU is violated? Should the issuer have required a more rigorous proof? Did the verifier fail to meet the requirements and caused harm to the subject? What about the state? Does the state have grounds to pursue "improper acceptance" of a VC? Under what laws? False advertising, like current ToUs? Or privacy? This gets jurisdictional immediately. Is the verifier bound by the definitions of "accept" in the country in which they operate? Or does the location of the issuer, subject, or holder also create jurisdictional exposure?

In short, this is a minefield of legal liability. One which, IMO, makes such a requirement a complete non-starter.

Like a terms of use on a website, the ToU in a VC is a communication of policy. That's all. What recipients do in response to that policy is their business. It is NOT the business of the software to enforce.

Even courts acknowledge that the initial load of a web page doesn't constitute agreement to the sites ToU--because it is not possible to know what ToU, if any, is applicable before viewing that page. Continued use of a website, AFTER the opportunity to review the ToU can be interpreted as a binding click-wrap license... but VCs don't even have that. 

Could you imagine if web browsers refused to render web pages that had Terms of Use the user didn't understand or want? It would break the web. 

We simply can't put these kinds of requirements in the spec.

index.html
```diff
@@ -1866,12 +1851,15 @@ <h3> 'Subject Only' Terms of Use </h3>
       </p>
 
   <p>
-The Subject Only Terms of Use states that a verifiable credential MUST only be
-presented to a verifier by the subject. If a verifier is presented with a
-verifiable credential containing the Subject Only Terms of Use, by anyone other
-    than the subject, it MUST refuse to accept it.
+The Subject Only Terms of Use states that a verifiable credential is only to be presented by the subject. If a verifier is presented with a verifiable presentation containing a verifiable credential with a Subject Only Terms of Use, by anyone other
```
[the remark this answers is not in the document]
stonematt: This particular case seems similar to the verification rules (section 8) related to revocation where the DM states in section 8.8, Revocation: "If revocation instructions are present, the credential must not have been revoked."  In this case, "If the ToU type is Subject only, the credential must have been presented by the Subject".  The implication here is that if the opposite is true, the it should fail to "verify" -- the idea and language of "acceptable" is a Fitness for Purpose argument and it's appropriate.

index.html
stonematt: The whole language of "accept" or "reject" is irrelevant.  The VC/VP is either verifiable or not.  In this case, the Presentation wouldn't verify with a code of "NotPresentedBySubject".  The Verifier has discretion over disposal of the payload. Section 8.9 Fitness for Purpose describes the obligation, discretion, and risks for the verifier in an appropriately "hand wavy" way. What the Verifier does with the presented artifact following verification isn't material to the Data model.

For ToU that are machine readable and can be part of the verification scheme, verification success/failure is appropriate.  For ToU that are are machine readable and aren't purely verification, then we're in the realm of protocol and judgement, so the DataModel can't express an opinion.  For ToU that aren't machine readable, the DM only needs to provide a method for transmission and the market will have to sort it out...

jandrieu: @stonematt It sounds like you want acceptance of a ToU to be a part of verification:
> For ToU that are machine readable and can be part of the verification scheme, verification success/failure is appropriate.

My contention is that the specification doesn't provide any means of assurance about adherence to a ToU. We also  don't have anything in the data model that enables recording acceptance of a ToU (a Verifiable Presentation Receipt for example) . That might be a way to bootstrap a data-model centric way of testable commitment to a ToU, but it hasn't been a part of our conversation to date and certainly would be out of scope for 1.0.0. (A future specification for claims requests, credential reciepts, and presentation receipts seems like a good set of future work.)

If verification requires acceptance of the ToU, then we are making the verification process dependent on the issuer and/or holder, depending on the ToU, without a data model that has any way to represent acceptance must less the parties to the ToU. In contrast, if verification is purely about the cryptographic proof and procedural check (for revocation/status), then VCs can be verified independent of the issuer, holder, or verifier, as long as the key resolution and status resolution systems are running.

If verification requires agreement to a ToU, then the ToU needs to specify, and verification needs to programmatically prove, the counter parties to the ToU and their acceptance. In the Subject Must Be Holder ToU, are those terms binding on the Holder or the Verifier? How do we know if either party accepts the ToU? All the legal questions I raised above now become part of verification, turning it into, IMO, an intractable, unsolved problem.

As long as acceptance is not formally part of verification, then ToUs at either stage can state anything... which is fair. There will be use cases we won't imagine in our wildest inspirations and flexible ToUs are going to have to be part of that. 

But if acceptance is part of verification, then I ask, humbly, which fields in the data model are used to check if the correct counterparty (or parties) has accepted the ToU?

stonematt: Acceptance is not formally part of verification.  My primary point was to say the language of "accept" and "reject" are out of bounds.  That concept requires the judgement of the parties involved.  Acceptance is a Fit-for-Purpose decision that requires more context that the VP alone.  A Verifier may reject a VP b/c they don't trust the issuer, or the credential isn't appropriate.  In either cases, the VP verified.  

ToU are policy statements that stakeholders must follow or bear the risk of not following (see verification step 8.9). Some may be expressed in a machine readable way, others won't be. Our narrow scope is to provide something verifiable. At this stage of the process, perhaps no ToU can be "verifiable" in an automated way because we haven't adopted such a technology, and are thus advisory.  In the event the market adopts such a technology, parsing a ToU will result in one of 3 outcomes. 

1. ToU may explicitly pass verification (verifier can read it, reason over it, and determine whether it's fit for purpose and choose to comply), 
2. explicitly fail verification (one of the above is explicitly false), 
3. result in an "unknown" status  (based on ToU content,  details included in the VP and perhaps some future protocol) 

The verifier ultimately decides what to do with the artifact that was presented, regardless of the Acceptance decision. Also, I agree that the compelling idea of receipts (of any kind) definitely don't fit in v1.0 of this spec.

Can you suggest a path forward (quickly)? Otherwise, this will be published as an at-risk feature and likely fall out of the spec before publication.

jandrieu: I'll update ASAP based on the feedback so far. 

Sadly, I'm still a bit confused about your position. It seems like you're trying to leave the door open for language that verification depends on the ToU, but that you aren't recommending any spec-text that says that at this time.

If you look at Section 8.9 in the PR, I replaced "MUST be enforced" with "SHOULD be respected" to separate conformant, testable verification (which I see as the "MUST" requirements) from business rules, policies, and their enforcement. 

If you're ok with that language, I think we're good.

index.html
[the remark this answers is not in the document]
jandrieu: Hmmmm...

What is being verified by evaluating the ToU? 

The semantics of VC verification, IMO, mean verifying that the utterance is authentic (crypto check) and still valid (revocation check).

Evaluating the ToU is something done once you know the underlying credential is authentic. After all, if it isn't authentic, it may not be the actual current terms from the alleged issuer. 

Seems to me there is a big difference between "Is this statement authentic and current?" and "Can we use this statement legally for our purposes?"

The latter is about the use of the VC, not its authenticity.

In short, isn't verification about authenticity?

jandrieu:
Ok. I've incorporated @David-Chadwick's suggestions. 

How does this look?

index.html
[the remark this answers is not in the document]
David-Chadwick: Yes fine

index.html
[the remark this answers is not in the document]
David-Chadwick: @jandrieu Authentic and current is different from simply authentic. An expired VC is still authentic. So if you want to add "current" to the authenticity test, then you can also add "presented by the subject" to the authenticity test. This is nothing to do with legality. It is about testing properties of a VC, such as: not revoked, not expired, not presented by the subject, and anything else of generic use and value that we may wish to standardise.

msporny:
There are merge conflicts preventing a merge of this branch, please fix them.

msporny:
@jandrieu, I heard you say that both you and @David-Chadwick agree on this PR, which is great. I also saw a few requests to change text that you didn't respond to from @David-Chadwick ... so, I just want to make sure there is agreement to pull in this PR before I do it. Also, merge conflicts need to be fixed.

jandrieu:
LOL. There weren't conflict before you merged those other PRs. I'll get a fix in ASAP.

FWIW, I responded to all of David's comments in the latest edit to the PR, but can't speak to whether or not he or Matt feel their comments are addressed. David at least approved the PR.

msporny:
> LOL. There weren't conflict before you merged those other PRs. I'll get a fix in ASAP.

The end is near... the axe is dropping... contributors are bound to lose some fingers. THIS IS W3C! :)

index.html
```diff
@@ -1851,12 +1857,24 @@ <h3> 'Subject Only' Terms of Use </h3>
...
 
   <p>
-The Subject Only Terms of Use states that a verifiable credential is only to be presented by the subject. If a verifier is presented with a verifiable presentation containing a verifiable credential with a Subject Only Terms of Use, by anyone other
-    than the subject, it relies on that presentation on its own responsibility. Similarly, if a holder other than the subject presents such a presentation, they do so on their own responsibility.
+The Subject Only Terms of Use states that a verifiable credential is only to be 
+presented by the subject. If a verifier is presented with a verifiable presentation
+containing a verifiable credential with a Subject Only Terms of Use, by anyone other
+than the subject, it relies on that presentation on its own responsibility. 
+Similarly, if a holder other than the subject presents such a presentation, they do 
+so on their own responsibility.
```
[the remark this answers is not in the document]
jandrieu: Well... we don't have a notion of "valid". We have a notion of "verified". The point of my language in this section is to distinguish the liability of use (which the verifier must accept in any case) from the cryptographic and procedural checks of verification.

index.html
```diff
@@ -1851,12 +1857,24 @@ <h3> 'Subject Only' Terms of Use </h3>
       </p>
 
   <p>
-The Subject Only Terms of Use states that a verifiable credential is only to be presented by the subject. If a verifier is presented with a verifiable presentation containing a verifiable credential with a Subject Only Terms of Use, by anyone other
-    than the subject, it relies on that presentation on its own responsibility. Similarly, if a holder other than the subject presents such a presentation, they do so on their own responsibility.
+The Subject Only Terms of Use states that a verifiable credential is only to be 
+presented by the subject. If a verifier is presented with a verifiable presentation
+containing a verifiable credential with a Subject Only Terms of Use, by anyone other
+than the subject, it relies on that presentation on its own responsibility.
```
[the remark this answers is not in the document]
jandrieu: Same comment as the prior section.

jandrieu:
@David-Chadwick, unfortunately, your two remaining suggestions don't fit with the intention of this PR.

index.html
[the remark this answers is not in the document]
David-Chadwick: If we do not have a notion of valid as you suggest then please tell me why the definition of expiry date states:

expirationDate
    The value of the expirationDate property MUST be a string value of an [ISO8601] combined date and time string representing the date and time the credential ceases to be valid. 

There is no difference between a VC being presented after its expirationDate (if one is present) and a VC being presented by someone other than subject (if ToU SubjectOnly is present). So we need to treat both properties in exactly the same way for consistency purposes.

index.html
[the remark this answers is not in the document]
David-Chadwick: Ditto :-)

David-Chadwick:
@jandrieu . I understand your intention, and we both agree that a verifier can do anything it wants to with any VC that it receives, legal or otherwise. I think we also both agree that deterministic automatic  tests can be built for the issues that we are discussing (if not then they are out of scope of the current discussion and therefore we should agree).
The point of contention is whether a verifier is standards conformant or not by accepting a VC that clearly states that it is not valid (according to the issuer's properties inserted into the VC) in terms of the time, place and way that it is presented. (Examples include time of issuance, expiry time, Subject only ToU).
I think your position is that the verifier is standards conformant in accepting these VCs, though may not be legally conformant.
My position is that the verifier is not standards conformant, and legality does not come into the discussion (since we are writing a technical standard).

jandrieu:
Actually, I don't believe we can test the behavior of the verifier, which is independent of the verifiability of the credential.

My contention is that the action of the verifier is outside conformance. We don't verify the verifier, we verify the software.

David-Chadwick:
We both agree that what we are testing is the verifier's software. We can test that the verifier's software will do the following (at least)
- reject a VC with a faulty proof
- reject a VC that has expired
- reject a VC that has been revoked
- reject a VC presented by a holder NE subject with a ToU of Subject Only
.... there are lots more, as the test suite shows

index.html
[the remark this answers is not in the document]
David-Chadwick: There is clearly a notion of VC validity in the current specification. Here is more evidence. The note in section 8.5 (Proofs) states

For example, a Linked Data Signature created property establishes a date and time before which the credential SHOULD NOT be considered valid.

dlongley:
@David-Chadwick,

I don't want to further complicate things, but, JFYI, we need to be careful with language like:

```
reject a VC that has expired
reject a VC that has been revoked
```

As, for example, a historical verification check should still pass under the right circumstances even though a VC has since expired or been revoked. So any language used should take this sort of thing into consideration.

jandrieu:
I think we still have language issues about accept/reject. To me, whether or not a verifier accepts or rejects a VC is a much bigger issue than the addition of the ToU. It includes whether or not the issuer is known and respected for the claims asserted. It includes the legal evaluation of liability associated with relying on a given verifiable credential, from any and all sources. For example, whether or not the ABC (in California, the Alcoholic Beverage Control who manages adherence to liquor licenses) recognizes the VC in question as legitimate diligence for a VC who is an alcohol vendor. Such a determination requires evaluating which jurisdiction applies, which may be dependent on the current location of the individual--or perhaps on the shipping address. These downstream considerations are out of scope. 

What the spec should deal with, IMO, is whether or not a verifiable credential or presentation can be "verified". 

Verification, IMO, cannot depend on the verifier (and its acceptance of a given ToU) nor on the holder (and whatever identity assurance may or may not be applied).

A verifiable credential is verifiable entirely on the proof and the revocation check. Nothing more, nothing less.

To @dlongley's point, it's perfectly viable for a VC to pass verification and be expired. That will likely affect what the verifier does with the VC, but if the proof is good and the revocation check passes, then the VC is "verified". Full stop.

Downstream, what the verifier does with a VC (accept/reject) is their business and not part of verification.

David-Chadwick:
The issuer issues a VC with or without constraints. In the former case we have no disagreements. The VC can be used by anyone for anything. In the latter case the issuer is stating when the VC is valid and when it is not. It is the issuer's right to do this. There are an almost infinite number of constraints that could be placed on a VC by an issuer, but we are only standardising the most commonly useful ones, and we have an extension point (the ToU) to allow the issuer to add non-standardised constraints in a standard conforming way. Some of the constraints can be easy to test in code, others cannot be. Most of @jandrieu's discussion concerns the latter. I am more concerned with the former than the latter (and I mostly agree with @jandrieu discussion of the latter, that they are legal rather than implementation issues). So lets close that discussion now please. 

Lets concentrate on constraints inserted by the issuer that can be tested for in code. This is important because the issuer is stating when the VC is valid. I am not sure that @jandrieu agrees on this point. So here is a direct question to @jandrieu. Do you believe that an issuer can place constraints on a VC to say when the VC is valid and when it is not?

Next question. If the answer is Yes, then should a verifier's code respect this, or ignore it?

jandrieu:
@David-Chadwick wrote:
> So here is a direct question to @jandrieu. Do you believe that an issuer can place constraints on a VC to say when the VC is valid and when it is not?

An issuer can say anything they like, so they can state any constraints they desire, whether in the ToU or in the claims payload directly.

However, whether or not a claim or a credential or presentation is "valid", whether it should be "accepted" or "rejected" is dependent on more factors than are contained in the data-model and therefore are outside the scope of the specification.

The specification should say what it means to verify a verifiable credential and a verifiable presentation.

Full stop. 

Anything beyond verification is, IMO, at best a non-normative illustration of what one *might* do with given credentials in a given use case, and those illustrations may or may not apply to any particular use of the technology. At worst, it will confuse and mislead people into believing VCs and VPs are capable of things they aren't.

David-Chadwick:
@jandrieu. There is a flaw in your logic "However, whether or not a claim or a credential or presentation is "valid", ... is dependent on more factors than are contained in the data-model and therefore are outside the scope of the specification." This is not quite correct.

If an issuer says a VC is invalid, then no-one and nothing outside of the data model can make it valid. However, if an issuer says a VC is valid, then a verifier can still treat it as invalid using factors that are inside or outside the data model.

Since I am dealing with the former case of when the issuer says a VC is invalid, then the verification stage should cover this.

dlongley:
@David-Chadwick,

> If an issuer says a VC is invalid, then no-one and nothing outside of the data model can make it valid. However, if an issuer says a VC is valid, then a verifier can still treat it as invalid using factors that are inside or outside the data model.

A VC is a tool that an issuer uses to assert that something is true. They are able to say that they only know something is true for a little while (expiration date) and they can say that what they said previously was true actually isn't (revocation).

I don't know what it means for them to say that what they said isn't true if the person presenting the information isn't the one the information is about. That seems to defy logic and is out of bounds for a mere assertion of truth. So I don't think the analogies to those things you've presented are working for me here.

Rather, ToU is getting into expressing acceptance policy rather than whether or not what was said is true. It's essentially the issuer saying that they won't stand behind what they said if, for instance, the holder is not the subject; it is about liability. I do think that's fundamentally different from everything else we've expressed so far and that this is at the heart of what @jandrieu is arguing.

Perhaps the common ground to be reached is not to say that the VC is "invalid" when the ToU are not complied with, but rather to say that the issuer has no liability in its use -- and this can be said in prose? I don't think there's anything we can enforce at a data model level here.

David-Chadwick:
@dlongley The issuer says "this is a ticket to concert A that was sold to X and can only be used by X on Date Y". This is a statement of several facts that are true. If the ticket is presented to concert B it is invalid, if it is presented to concert A on Date Z it is invalid, and if it is presented to concert A on Date Y by W it is invalid. However the statement remains true all the time i.e. it is authentic regardless of when or where it was presented. So we are not talking about authenticity, we are talking about validity.

You appear to be equating truth to validity, whereas I am equating truth to authenticity. So perhaps we should not talk about truth, due to its ambiguity, but rather about authenticity and validity. A statement can be authentic and valid, authentic and invalid, and unauthentic and invalid. It cannot realistically be unauthentic and valid, as this is a believed lie.

The point we are debating is the status of an authentic and invalid VC, which is determined purely by the data model. This is no more a ToU issue than the expiry time. (By this I mean that we could encode the validity time as a ToU, and we could encode SubjectOnly as a simple property like expirationDate).

David-Chadwick:
To try to summarise our discussions about authenticity and validity I have drawn the following 2x2 matrix for the case where a VC is proved to be authentic to the satisfaction of the verifier, but the issuer has used the data model to indicate when the VC is valid and when it is invalid
```
                        Verifier determines
                       Valid        Invalid
                   ------------------------
Issuer Says  Valid |  OK        |   OK    |
in DM              |---------------------
           Invalid | Impossible |    OK   |
                   ----------------------
 ```
In the BLH corner, because the verifier's software has no choice in determining the outcome of the validation process, I am saying that the DM document should state categorically that this type of VC  MUST not be accepted/used/trusted/validated by the verifier's DM conformant software.

Consequently I would expect our test suite to have tests to validate the processing of this type of VC.

**Comment by jandrieu** [comment]
"Validity" is out of scope.

"Verified" is in scope.

These are not "Validatable Credentials"

They are "Verifiable Credentials".

The specification says how you verify a verifiable credential.

It's up to the relying party to decide if any given credential--verified or not--is valid for their particular use case.

The DM doesn't say its valid. It says whether or not it is verified. The matrix above doesn't apply.

Further, the only way the issuer can say something is no longer verifiable is revocation. There is no way, within the data-model, for an issuer to say that a VC is not verifiable because of the a particular ToU, because of a particular Holder, or because of a particular Verifier. The issuer either makes a credential (with a good proof) or they don't. They can then either revoke that credential or they don't.

They don't get to decide what happens with that credential after the fact--EXCEPT for revocation triggering a subsequent verification failure. 

Clearly we're spinning our wheels here. You want the spec to define what is "valid". I, and others, have said validity--the full means by which a receiving party decides whether to accept or reject a credential for their particualr use--is out of scope. 

"Verification" is what we need to define rigorously. 

Anything about validity is non-normative illustration, at best.

**Comment by TallTed** [comment]
**Most** of this thread is simply and entirely **out of scope.**

**Compliance with a data model** simply means **using the data model.**

It has no connection to what is *done* with the data that is expressed in that model.

The model should make any intended protocols *possible*, but the model cannot make any assertions, demands, requirements about what those protocols might be, do, say, etc.

See the [Provenance Data Model](https://www.w3.org/TR/prov-dm/) and the [Exit Criteria used by that WG](https://www.w3.org/2011/prov/wiki/ProvCRExitCriteria) for a successful example.

**Comment by jandrieu** [comment]
@msporny I like the focus on "verification" and "use", especially because @David-Chadwick wont be the last person to want to use VCs as he does. Lots of people are going to be confused with some of the current language and it would probably serve us well to have David's issues well documented as approaches for how one might use VCs (in a way that retains clarity about what verification means).

**Comment by David-Chadwick** [comment]
@msporny If we use the IEEE definition of verification "The evaluation of whether or not a product, service, or system complies with a regulation, requirement, specification, or imposed condition. It is often an internal process." then this agrees with your 4 stages and we should not use the term validation in our specification (which is defined by IEEE as "The assurance that a product, service, or system meets the needs of the customer and other identified stakeholders. It often involves acceptance and suitability with external customers". This is often what @jandrieu talks about.)

An example of an "imposed condition" from the IEEE definition would be a constraint on a VC imposed by the issuer, in conformance to our specification. Discussions above have indicated that only constraints that are independent of usage should be considered during verification. Thus compliance with usage-independent constraints is part of the verification process. Thus the validity time of a VC (issuance date and expiration date) are imposed constraints on a VC, are independent of its use, and are thus part of the verification process. So is a 'subject only' constraint. As I noted above we can simply encode this as 

"subjectOnly": "True",

and removes it from the ToU definition and discussion, as it is independent of the use. The definition of this property could be:
"The subjectOnly property states that the verifiable credential can only be encapsulated into a verifiable presentation created by the subject."

@TallTed. I believe the above agrees with your statement "Compliance with a data model simply means using the data model." 

@jandrieu "It's up to the relying party to decide if any given credential--verified or not--is valid for their particular use case."
I have a problem with the above statement, particularly a VC that has not been verified. I find it difficult to imagine any use case in which an unverified VC is useful, and I would think this is outside the scope of a verifiable credential specification. Please note that the 2x2 matrix I drew has NOTHING to do with usage, and according to the IEEE definitions, is part of the verification process. The bottom left hand square of the this matrix is saying that in NO use cases can such a VC be verified.

**Comment by David-Chadwick** [comment]
Below are a set of suggestions to move forward with this issue.
1. Try to remove the term valid from our specification (e.g. as in issuance Date)
2. Introduce the IEEE definition for verification and say that our specification is constrained to verifying VCs independent of their usage
3. Remove Subject Only ToU and replace with Subject Only property
4. Ensure we have text to indicate that enforcing usage is outside of the scope and is a legal as well as technical issue (@jandrieu's text does a good job of this already).

**Comment by David-Chadwick** [comment]
In #337 I have updated our definition of  verification to comply with the IEEE definition, and also added the validation definition, along with a Note to say that the spec is concerned with verification. Then
I have found each occurrence of valid/validation and replaced with verify/ication where appropriate.

Note that I have replaced Subject Only Tou with Subject Only property in #283 since this one dealt with the Subject NE Holder text.

**Comment by msporny** [comment]
Is this PR ready to go? If so, need merge conflicts resolved.

**Comment by jandrieu** [comment]
Apologies, I thought I got this pushed a couple days ago.

@David-Chadwick and @burnburn is this good to go? I realize we'll need more PRs to clean up the rest of the doc for the new conformant language, but I'd like to get this merged in if possible.

File: index.html [hunk 7]
```diff
@@ -1433,20 +1433,16 @@ <h3>Extensibility</h3>
...
+extensions. These decisions are up to the developers of these
 applications and are specifically not the domain of this specification.
         </p>
 
-        <p>
-Applications that do not understand the semantic meaning of all properties
-while processing a <a>verifiable credential</a> or a
-<a>verifiable presentation</a> MUST produce an error.
-        </p>
+
```
David-Chadwick: Software that does not understand the semantic meaning of all the properties in a <a>verifiable credential</a> or a
<a>verifiable presentation</a> MUST fail to verify the data structure.

index.html
```diff
@@ -1793,8 +1789,17 @@ <h3>Terms of Use</h3>
...
+<a>holder</a> places terms of use on their <a>verifiable credentials</a>. One
+way could be for the <a>subject</a> to request the <a>issuer</a> to place the
+terms of use inside the issued <a>verifiable credentials</a>. Another way
+could be for the <a>subject</a> to delegate a <a>verifiable credential</a> to
+a <a>holder</a> and placing terms of use restrictions on the delegated
+<a>verifiable credential</a>.
+        </p>
+        <p>
+Terms of use can be expressed using the following <a>property</a>:
         </p>
```
David-Chadwick: Terms of use MAY be expressed......

index.html
```diff
@@ -3042,26 +2948,35 @@ <h3>Revocation</h3>
...
+corporate headquarters of the franchise.</li>
+Policy information expressed by the <a>issuer</a>
+in the <a>verifiable credential</a> SHOULD be respected by holders 
+  and verifiers unless they accept the liability of ignoring the policy.
+</p>
+          <p>
+        <p>
+The <a>issuer</a> is trusted by the <a>verifier</a> to make the claims at
+hand. For example, a franchised fast food restaurant location will trust
+discount coupon claims made by the corporate headquarters of the franchise.
```
David-Chadwick: This looks like duplication of the paragraph above

index.html
```diff
@@ -3612,11 +3527,12 @@ <h3>Usage Patterns</h3>
 person.
           </li>
           <li>
-When the same subject identifier of a credential refers to the same subject 
-across presentations or verifiers, then loss of privacy is possible. Even 
-when different credentials are presented, if the subject identifier is the 
-same, verifiers (and those with access to verifier logs) could infer that
-the holder of the credential is the same person.
+The same <a>subject</a> identifier of a <a>credential</a> refers to the same
```
David-Chadwick: You have missed out "When" from the beginning of this sentence.

index.html
jandrieu: Your proposed language makes verification dependent on the software rather than on conformance. See https://github.com/w3c/vc-data-model/issues/317

index.html
jandrieu: Thanks. Fixed.

index.html
jandrieu: Yep. Thanks.

index.html
jandrieu: Fixed.

index.html
David-Chadwick: Accepted. You are good to go

jandrieu:
@burnburn I believe @David-Chadwick signed off, although I'm having a difficult time "resolving" all the issues. I thought I had, but the status still shows David wants changes.

@David-Chadwick are we good to go?

@burnburn Can you chime in here?

index.html
David-Chadwick: This is now resolved

PR #321, opened 2018-12-18
Add "id" to JWT example.

davidlehn:
Match up with the JWT "sub" property.

***
<a href="https://pr-preview.s3.amazonaws.com/davidlehn/vc-data-model/pull/321.html" title="Last updated on Dec 18, 2018, 12:32 AM UTC (e764364)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/321/8c132b4...davidlehn:e764364.html" title="Last updated on Dec 18, 2018, 12:32 AM UTC (e764364)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2766,6 +2766,7 @@ Add a short description explaining the example above.
     ],
     "type": ["VerifiableCredential", "UniversityDegreeCredential"],
     "credentialSubject": {
+        "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
         "degree": {
           "type": "BachelorDegree",
           "name": "Bachelor of Science in Mechanical Engineering"


index.html
```diff
@@ -2766,6 +2766,7 @@ <h3>JWT and JWS Considerations</h3>
     ],
     "type": ["VerifiableCredential", "UniversityDegreeCredential"],
     "credentialSubject": {
+        "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
```
msporny: I'm pretty sure @awoie left this off on purpose. The JWT folks don't want to duplicate information. That said, we probably need to add a toJWT and fromJWT process for how the data model needs to change from when it's expressed as a JWT to when it gets used by a VC library.

awoie: As @msporny mentioned, the goal was to also reduce duplication:

From the current spec:

> For backward compatibility with JWT processors the following JWT registered claim names MUST be used **instead of or in addition** to their respective verifiable credential counterparts:
> ...
> - `sub` MUST represent the `id` property contained in the credential subject.

The `id` property could be omitted.

msporny: Right, so that was on purpose.

With my editorial/implementer/interop hats on -- this is bad for interop. We need to define a "conversion to JWT" and "conversion from JWT" process so that the appropriate fields are stripped/replaced so that Verifiable Credential libraries don't need to understand the strange things that are introduced/stripped via JWT. Fundamentally, the verifiable credential libraries shouldn't need to care about the crypto layer of the stack.

msporny:
Closing this PR, raising an issue to address this concern.

msporny:
Raised https://github.com/w3c/vc-data-model/issues/336 ... closing this PR.

PR #377, opened 2019-01-17
added zkp to explainer, also

brentzundel:
cleaned up other explainer language to match the data model.


ken-ebert:
I worked with Brent to make sure the zkp parts were good.

PR #378, opened 2019-01-18
Explainer input with respect to JWT

awoie:
reordered tricky design decisions; provided input for JWT

### DIFF
--- a/VCDMExplainer.md
+++ b/VCDMExplainer.md
@@ -114,7 +114,7 @@ Here is an example of usage of the ID property in a credential, using the DID sc
 
 ### JSON vs JSON-LD
 
-### Support for JWT
+Support for JWT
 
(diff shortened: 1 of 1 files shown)


VCDMExplainer.md
```diff
@@ -111,9 +111,21 @@ Here is an example of usage of the ID property in a credential, using the DID sc
 ```
 
 ## Tricky Design Choices
-* JOSE vs JWT - see [Issue 93](https://github.com/w3c/vc-data-model/issues/93)
-* JSON vs JSON-LD
-* Selective disclosure as a requirement see [Issue 224](https://github.com/w3c/vc-data-model/issues/224) for example
+
+### JSON vs JSON-LD
```
dlongley: I think we should remove this heading and just leave "Support for JWT" below.

awoie: Ok, I am fine with that. Updated the PR.

PR #384, opened 2019-01-20
Add section on Syntax and Proof Format Trade-Offs.

msporny:
This PR adds a section on Proof Format Trade-Offs since it is a question that comes up often when discussing the different proof formats in the spec.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/384.html" title="Last updated on Feb 23, 2019, 10:29 PM UTC (64bd3f0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/384/5f09c57...64bd3f0.html" title="Last updated on Feb 23, 2019, 10:29 PM UTC (64bd3f0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4428,7 +4428,8 @@ whitespace does not invalidate signature.
 
         <tr>
           <td>
-<a href="#pf16">PF16.</a> Support experimental signature systems.
+<a href="#pf16">PF16.</a> Designed to easily support experimental signature 
+systems.
           </td>
           <td class="missing">✖</td>
           <td class="missing">✖</td>
@@ -4662,7 +4663,7 @@ shortcoming of JWTs make it challenging to, for example, express signed data in
 web pages that search crawlers index.
       </dd>
       <dt id="pf16">
-PF16: Support experimental signature systems.
+PF16: Designed to easily support experimental signature systems.
       </dt>
       <dd>
 The LD-Proof format is naturally extensible, not requiring the format to be
--- a/index.html
+++ b/index.html
@@ -4261,24 +4261,20 @@ All or any combination of the above.
   </section>
 
   <section class="appendix">
-    <h2>Proof Format Trade-Offs</h2>
+    <h2>Syntax and Proof Format Trade-Offs</h2>
 
     <p>
 The Verifiable Credentials Data Model is designed to be compatible with a
 variety of existing and emerging syntaxes and digital proof formats. Each
-format has benefits and drawbacks. The following table is intended to summarize
-a number of these native trade-offs.
+approach has benefits and drawbacks. The following table is intended to
+summarize a number of these native trade-offs.
     </p>
 
     <p>
-The table below compares three proof format ecosystems; JSON+JWTs,
-JSON-LD+JWTs, and JSON-LD+LD-Proofs. At the time of publication of this
-specification, no organization is planning to use JSON+JWTs. The existence
-of JSON+JWTs in the table below is provided based on review comments on the
-specification for readers that were curious about the benefits that JSON-LD
-provided over JSON. Readers should also be aware that Zero-Knowledge Proofs
-are currently proposed as a sub-type of LD-Proofs and thus fall into the
-final colum below.
+The table below compares three syntax and proof format ecosystems; JSON+JWTs,
+JSON-LD+JWTs, and JSON-LD+LD-Proofs. Readers should be aware that
+Zero-Knowledge Proofs are currently proposed as a sub-type of LD-Proofs and
+thus fall into the final column below.
     </p>
 
     <table class="simple">
--- a/index.html
+++ b/index.html
@@ -4334,8 +4334,8 @@ in semantics or structure.
         </tr>
         <tr>
           <td>
-<a href="#pf6">PF6.</a> A way to associate datatypes with values such as
-dates and times.
+<a href="#pf6">PF6.</a> A way to associate arbitrary datatypes, such as dates
+and times, with arbitrary property values.
           </td>
           <td class="missing">✖</td>
           <td class="supported">✓</td>
@@ -4535,12 +4535,13 @@ or Japanese, in which a text string is expressed via the use of language tags.
 JSON does not provide such a feature.
       </dd>
       <dt id="pf6">
-PF6: A way to associate datatypes with values such as dates and times.
+PF6: A way to associate arbitrary datatypes, such as dates
+and times, with arbitrary property values.
       </dt>
       <dd>
-JSON-LD enables a developer to specify the data type, such as Date,
-unsigned integer, or Temperature, in which a string is expressed as a native
-feature of the language. JSON does not provide such a feature.
+JSON-LD enables a developer to specify the data type of a property value,
+such as Date, unsigned integer, or Temperature by specifying it in the
+JSON-LD Context. JSON does not provide such a feature.
       </dd>
       <dt id="pf7">
 PF7: A facility to express one or more directed graphs, such as a social
--- a/index.html
+++ b/index.html
@@ -4391,8 +4391,8 @@ double for every embedding.
 
         <tr>
           <td>
-<a href="#pf13">PF13.</a> Proof format supports arbitrary proofs such as
-Proof of Work, Proof of Elapsed Time, and Proof of Stake.
+<a href="#pf13">PF13.</a> Proof format supports arbitrary proofs such as Proof
+of Work, Timestamp Proofs, and Proof of Stake.
           </td>
           <td class="missing">✖</td>
           <td class="missing">✖</td>
@@ -4586,58 +4586,116 @@ PF11: Nesting signed data does not cause data size to double for every
 embedding.
       </dt>
       <dd>
-
+When a JWT is encapsulated by another JWT, the entire payload must be base-64
+encoded in the initial JWT, and then base-64 encoded again in the encapsulating
+JWT. This is often necessary when a cryptographic signature is required on a
+document that contains a cryptographic signature, such as when a Notary
+signs a document that has been signed by someone else seeking the Notary's
+services. LD-Proofs do not require base-64 encoding the signed portion of a
+document and instead rely on a canonicalization process that is just as
+secure, and that only requires the cryptographic signature to be encoded
+instead of the entire payload.
       </dd>
       <dt id="pf12">
 PF12: Proof format supports Zero-Knowledge Proofs.
       </dt>
       <dd>
-
+The LD-Proof format is capable of modifying the algorithm that generates
+the hash or hashes that are cryptographically signed. This cryptographic
+agility enables digital signature systems, such as Zero-Knowledge Proofs,
+to be layered on top of LD-Proofs instead of an entirely new digital signature
+container format to be created. JWTs are designed such that an entirely new
+digital signature container format will be required to support Zero-Knowledge
+Proofs.
       </dd>
       <dt id="pf13">
-PF13: Proof format supports arbitrary proofs such as Proof of Work, Proof of
-Elapsed Time, and Proof of Stake.
+PF13: Proof format supports arbitrary proofs such as Proof of Work, Timestamp
+Proofs, and Proof of Stake.
       </dt>
       <dd>
-
+The LD-Proof format was designed with a broader range of proof types in mind
+and supports cryptographic proofs beyond simple cryptographic signatures.
+These proof types are in common usage in systems such as decentralized ledgers
+and provide additional guarantees to
+<a>verifiable credentials</a>, such as the ability to prove that a particular
+claim was made at a particular time or that a certain amount of energy was
+expended to generate a particular credential. The JWT format does not support
+arbitrary proof formats.
       </dd>
       <dt id="pf14">
-PF14: Proofs can be expressed unmodified in other data syntaxes such as YAML,
-N-Quads, and CBOR.
+PF14: Proofs can be expressed unmodified in other data syntaxes such as XML,
+YAML, N-Quads, and CBOR.
       </dt>
       <dd>
-
+The LD-Proof format utilizes a canonicalization algorithm to generate a
+cryptographic hash that is used as an input to the cryptographic proof
+algorithm. This enables the bytes generated as the cryptographic proof to be
+compact and expressible in a variety of other syntaxes such as XML,
+YAML, N-Quads, and CBOR. Since JWTs require the use of JSON to be generated,
+they are inextricably tied to the JSON syntax.
       </dd>
       <dt id="pf15">
 PF15: Changing key-value ordering, or introducing whitespace does not
 invalidate signature.
       </dt>
       <dd>
-
+Since LD-Proofs utilize a canonicalization algorithm, the introduction of
+whitespace that does not change the meaning of the information being expressed
+has no effect on the final cryptographic hash over the information. This means
+that simple changes in whitespace formatting, such as those changes made when
+writing data to a schema-less database and then retrieving the same information
+from the same database do not cause the digital signature to fail. JWTs
+encode the payload using the base-64 format which is not resistant to
+whitespace formatting that has no effect on the information expressed. This
+shortcoming of JWTs make it challenging to, for example, express signed data in
+web pages that search crawlers index.
       </dd>
       <dt id="pf16">
 PF16: Support experimental signature systems.
       </dt>
       <dd>
-
+The LD-Proof format is naturally extensible, not requiring the format to be
+extended in a formal international standards working group in order to
+prevent namespace collisions. The JWT format requires entries in a centralized
+registry and does not support experimentation as easily as the LD-Proof format
+does. LD-Proof format extension is done through the decentralized publication
+of cryptographic suites that are guaranteed to not conflict with other LD-Proof
+extensions. This approach enables developers to easily experiment with new
+cryptographic signature mechanisms that support selective disclosure,
+zero-knowledge proofs, and post-quantum algorithms.
       </dd>
       <dt id="pf17">
 PF17: Supports signature chaining.
       </dt>
       <dd>
-
+A signature chain is an ordered set of signatures over a data payload. Use
+cases, such as cryptographic signatures applied to a notarized document,
+typically require a signature by the signing party and then an additional one
+by a notary to be made after the original signing party has made their
+signature. Linked Data Proofs, including Linked Data Signatures, natively
+support chains of signatures. JWTs only enable a single signature over a
+single payload.
       </dd>
       <dt id="pf18">
 PF18: Does not require pre-processing or post-processing.
       </dt>
       <dd>
-
+In order to encode a <a>verifiable credential</a> or a 
+<a>verifiable presentation</a> in a JWT, an extra set of steps
+are required to convert the data to and from the JWT format. No such extra
+converstion step are required for <a>verifiable credentials</a> and
+<a>verifiable presentations</a> protected by LD-Proofs.
       </dd>
       <dt id="pf19">
 PF19: Canonicalization requires more than base-64 encoding.
       </dt>
       <dd>
-
+The JWT format utilizes a simple base-64 encoding format to generate the
+cryptographic hash of the data. The encoding format for LD-Proofs require 
+a more complex canonicalization algorithm to generate the cryptographic 
+hash. The benefits of the JWT approach are simplicity at the cost of 
+encoding flexibility. The benefits of the LD-Proof approach are flexibility at 
+the cost of implementation complexity.
       </dd>
     </dl>
 
--- a/index.html
+++ b/index.html
@@ -109,6 +109,12 @@ pre .comment {
   font-weight: bold;
   text-shadow: -1px 0 black, 0 1px black, 1px 0 black, 0 -1px black;
 }
+.supported {
+  background-color: #93c47d;
+}
+.missing {
+  background-color: #e06666;
+}
     </style>
   </head>
   <body>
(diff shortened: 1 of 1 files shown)


pelle:
@msporny @David-Chadwick I do not think this belongs to the spec. It's an invitation to never ending bikeshedding around what is essentially technical taste.

This would make a great series of blog posts by different people. But not in this document itself.

awoie:
@msporny I also believe we are opening pandora's box that starts never ending discussions. The table addresses some points that are not even related to the spec itself, and does not address some of the drawbacks that JSON-LD has. And as far as I can remember, the group has already agreed to not include a table like that.

msporny:
> It could be a part of a VC primer document, or at most, as an informative annex of the main specification.

Without my editor hat on and with my Digital Bazaar hat on, I would be fine w/ moving it to the appendix.

> the group has already agreed not to include a table like that.

That's not my memory. I know that @pelle mentioned he didn't want to see it in the document. @David-Chadwick seems that he could go either way. You are opposed. But we haven't had feedback from others (as requested during the call last week). If we did agree to not include it, could you cite where the group made this decision?

With my editor's hat on... we have had feedback from people that are confused about when to pick one over the other, so this text attempts to address those review concerns. I'm happy to put in additional drawbacks for JSON-LD as long as they're statements of fact (there is already one in there). That said, the section is JWTs and LD Proofs, not about JSON-LD as everyone seems to be using that format.

msporny:
> It's an invitation to never ending bikeshedding around what is essentially technical taste.

With my editorial hat on, I tried to just make plain statements of fact instead of introducing technical taste. Happy to remove statements that are just about technical taste. The table is meant to be a "You can do X with Y today." or "X is specifically designed to do Y". So implementers can look down the table and if they need X, then they know they need to use Y.

OR13:
I'm in favor of adding to the appendix, confusion over JWT and Linked Data Proofs is especially prevalent.

brentzundel:
I'm not opposed to having this information in an appendix.

index.html
```diff
@@ -3127,6 +3133,208 @@ <h5>JWT Decoding</h5>
...
+Proof format supports Zero-Knowledge Proofs.
+              </td>
+              <td class="missing">✖</td>
+              <td class="missing">✖</td>
+              <td class="supported">✓</td>
+            </tr>
+
+            <tr>
+              <td>
+Proof format supports arbitrary proofs such as Proof of Work, Proof of
```
pelle: ???? Why is this relevant

index.html
```diff
@@ -3127,6 +3133,208 @@ <h5>JWT Decoding</h5>
...
+              <td class="missing">✖</td>
+              <td class="missing">✖</td>
+              <td class="supported">✓</td>
+            </tr>
+
+            <tr>
+              <td>
+Support experimental signature systems.
+              </td>
+              <td class="missing">✖</td>
```
pelle: Of course this is possible in JWT as well

index.html
```diff
@@ -3127,6 +3133,208 @@ <h5>JWT Decoding</h5>
...
+              <td class="supported">✓</td>
+            </tr>
+
+            <tr>
+              <td>
+Does not require pre-processing or post-processing.
+              </td>
+              <td class="missing">✖</td>
+              <td class="missing">✖</td>
+              <td class="supported">✓</td>
```
pelle: I disagree strongly with this. With JSON-LD to get a safe view of what is actually signed, the data must be post processed. Linked data needs to be checked, pulled in and verified or you don't know what you're signing.

index.html
dlongley: I think the purpose of the table is for you to do exactly what you're doing here -- find those things that are relevant to you and make a choice on that basis.

dlongley: That being said, for this particular line, perhaps a better example would be Proof of Publication or Proof of Existence.

msporny:
> I apologize up front, but I’m being intentionally annoying in this review. This PR is written by the author of JSON-LD, who is exhibiting difficulties not being biased here.

I have attempted to be as unbiased as possible and submitted this as a PR specifically to counteract any bias in the language. The driving motivator was the confusion that reviewers have had around picking one format over another.

> I am not personally super interested in having such a debate as I don’t think it’s of much use, but if this proof of format trade-off is required lets agree what the scope of this is.

There seemed to be consensus on the last call to move this proposed section from where it is right now and include this section as an appendix.

> JSON-LD is not a proof format it’s a linked data format built on JSON. JWT’s consist of JSON’s which can just as easily be JSON-LDs.

It seems like you are looking at the diffs and not actually reading the rendered specification text. The specification text does make the distinction you are requesting. It differentiates between "JSON+JWT", "JSON-LD+JWT", and "JSON-LD+LDP". Here is a rendered version so it's more clear:

https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/384.html#proof-format-trade-offs

> As such JSON supports all JSON-LD features.

No, it does not, specifically JSON does not support:

* a universal identifier mechanism for JSON objects via the use of IRIs,
* a way to disambiguate keys shared among different JSON documents by mapping them to IRIs via a context,
* a mechanism in which a value in a JSON object may refer to a resource on a different site on the Web,
* the ability to annotate strings with their language,
* a way to associate datatypes with values such as dates and times,
* and a facility to express one or more directed graphs, such as a set of parties contributing to a verifiable credential, in a single document.

It is true you can express all these things in a JSON document as long as you have some standardized interpretation mechanism to interpret the syntax as the features above. That standardized interpretation mechanism is JSON-LD.

> He managed to find one benefit of JWT over JSON-LD while listing 20 or so benefits of JSON-LD over JWT. Which in reality is like comparing letters with envelopes.

There are only 10 advantages of JSON-LD+LDP over JSON-LD+JWT listed in the table.

> The majority of the points in this PR have nothing to do with Proof formats, but with the underlying data format. JWT is just a simple way of signing and encoding JSON.

The reason for this is that a few reviewers have asked about the difference between pure JSON+JWT and JSON-LD+JWT, and as a result, the table attempts to answer that question as well.

> JSON-LD has a separate [Linked Data Signatures 1.0](https://w3c-dvcg.github.io/ld-signatures/), which is what we’re actually supposed to be talking about here.

I believe the table is attempting to talk about those differences. Please take another look and let me know if you believe that to be true:

https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/384.html#proof-format-trade-offs

> A list of pro’s and cons was created at the last IIW and were posted in an email to the mailing list, together with many years of debate on this subject. As an editor of this draft, it would be a good idea to look those up and add it to this list and remove all the irrelevant ones.

Can you provide a link to that comparison so I can include it in the table?

> * Lack of good library support. Very few implementations exist.

This is something that fluctuates throughout time. JWTs have been out in the market much longer than LDPs. There was a time where there were very few JWT implementations and people used that as an argument to not develop JWTs because "very few implementations exist".

> * Very complicated to verify a signature for Joe Developer with little or no JSON-LD experience. It is not clear what is signed

Everything is signed in an LDP-protected document that goes through signature verification.

> * Http Linked Data format makes it even more difficult to know what is signed as the linked data is not immutable. (IPLD does not have this problem and would make a very good format for signing)

What is signed is the document you get, everything else is that is in the document isn't secured. The exact same thing is true for JWTs.

That is one of the reasons we are all working on the Hashlink spec in the CCG: https://tools.ietf.org/html/draft-sporny-hashlink-02

> * Too much flexibility. JWT also has this problem, but LD-Signatures goes even further. This is generally considered an anti-pattern in security unless you’re only targeting extremely experienced engineers

Flexibility in what way?

> * Canonicalization is hard. I believe this is one of the reasons there are so few implementations in the wild

It is, but so are digital signatures. Developers use libraries to get the hard work done.

> * Anything based on JSON-LD such as JSON LD Signatures is extraordinarily verbose, which makes it difficult to use in limited size data formats such as QR codes and push notification payloads

You seem to be arguing against JSON-LD, which you said above was perfectly fine for VCs and DID Documents?

> * Unless you perform another encoding step, JSON-LD data can not easily be used as an authentication token

You probably shouldn't be using JSON-LD as an authentication token. You should use a JWT or some other compact data format for that.

I'm out of time for today, but I'll go through and modify the PR based on your line-by-line input below. Thank you for the review comments, I'll update the PR to address some of your concerns.

index.html
```diff
@@ -4201,6 +4207,458 @@ <h3>Device Theft and Impersonation</h3>
...
+          <td class="missing">✖</td>
+          <td class="missing">✖</td>
+          <td class="supported">✓</td>
+        </tr>
+
+        <tr>
+          <td>
+<a href="#pf16">PF16.</a> Support experimental signature systems.
+          </td>
+          <td class="missing">✖</td>
```
pelle: This is factually incorrect. You can support any experimental signature system. Many people including uPort and Sovryn's implementations use experimental signature systems.

index.html
```diff
@@ -4201,6 +4207,458 @@ <h3>Device Theft and Impersonation</h3>
...
+          <td class="missing">✖</td>
+          <td class="supported">✓</td>
+        </tr>
+
+        <tr>
+          <td>
+<a href="#pf12">PF12.</a> Proof format supports Zero-Knowledge Proofs.
+          </td>
+          <td class="missing">✖</td>
```
pelle: As I wrote in my previous comment this is likely possible using JWT as well

index.html
```diff
@@ -4201,6 +4207,458 @@ <h3>Device Theft and Impersonation</h3>
...
+<a href="#pf6">PF6.</a> A way to associate datatypes with values such as
+dates and times.
+          </td>
+          <td class="missing">✖</td>
+          <td class="supported">✓</td>
+          <td class="supported">✓</td>
+        </tr>
+        <tr>
+          <td>
+<a href="#pf7">PF7.</a> A facility to express one or more directed graphs,
```
pelle: How is this relevant to VC and proof format tradeoffs?

index.html
```diff
@@ -4201,6 +4207,458 @@ <h3>Device Theft and Impersonation</h3>
...
+          <td>
+<a href="#pf5">PF5.</a> The ability to annotate strings with their language.
+          </td>
+          <td class="missing">✖</td>
+          <td class="supported">✓</td>
+          <td class="supported">✓</td>
+        </tr>
+        <tr>
+          <td>
+<a href="#pf6">PF6.</a> A way to associate datatypes with values such as
```
pelle: JWT since it uses JSON also does this. There are standard fields within JWT that express timestamps as integers. It depends on the application like so much else.

index.html
```diff
@@ -4201,6 +4207,458 @@ <h3>Device Theft and Impersonation</h3>
...
+<a href="#pf3">PF3.</a> A way to disambiguate keys shared among different
+JSON documents by mapping them to IRIs via a context.
+          </td>
+          <td class="missing">✖</td>
+          <td class="supported">✓</td>
+          <td class="supported">✓</td>
+        </tr>
+        <tr>
+          <td>
+<a href="#pf4">PF4.</a> A mechanism to refer to data in an external document,
```
pelle: How is this relevant to VC and proof format tradeoffs? JSON can easily refer to data outside the document. When you start. What is described here is a major security anti-pattern that has been used in XMLSignature in the past.

index.html
```diff
@@ -4201,6 +4207,458 @@ <h3>Device Theft and Impersonation</h3>
...
+          <td class="missing">✖</td>
+          <td class="supported">✓</td>
+          <td class="supported">✓</td>
+        </tr>
+        <tr>
+          <td>
+<a href="#pf3">PF3.</a> A way to disambiguate keys shared among different
+JSON documents by mapping them to IRIs via a context.
+          </td>
+          <td class="missing">✖</td>
```
pelle: When do you need this within a VC? This is a feature of DID documents I believe.

index.html
```diff
@@ -4201,6 +4207,458 @@ <h3>Device Theft and Impersonation</h3>
...
+          <th style="text-align: center;">JSON<br>+<br>JWTs</th>
+          <th style="text-align: center;">JSON&#8209;LD<br>+<br>JWTs</th>
+          <th style="text-align: center;">JSON&#8209;LD<br>+<br>LD&#8209;Proofs</th>
+        </tr>
+      </thead>
+
+      <tbody>
+        <tr>
+          <td>
+<a href="#pf1">PF1.</a> Support for open world data modelling.
```
pelle: How is this relevant to VC and proof format tradeoffs? JSON-LD supports one exact view of "Open World" modelling. JWT's live at a lower layer. So I don't understand the reasoning for this.

index.html
```diff
@@ -4201,6 +4207,458 @@ <h3>Device Theft and Impersonation</h3>
...
+          </td>
+          <td class="missing">✖</td>
+          <td class="supported">✓</td>
+          <td class="supported">✓</td>
+        </tr>
+        <tr>
+          <td>
+<a href="#pf2">PF2.</a> Universal identifier mechanism for JSON objects via
```
pelle: Interesting in a comparison between JSON and JSON-LD. But not in a comparison between proof formats.

index.html
```diff
@@ -4201,6 +4207,458 @@ <h3>Device Theft and Impersonation</h3>
...
+The Verifiable Credentials Data Model is designed to be compatible with a
+variety of existing and emerging syntaxes and digital proof formats. Each
+format has benefits and drawbacks. The following table is intended to summarize
+a number of these native trade-offs.
+    </p>
+
+    <p>
+The table below compares three proof format ecosystems; JSON+JWTs,
+JSON-LD+JWTs, and JSON-LD+LD-Proofs. At the time of publication of this
+specification, no organization is planning to use JSON+JWTs. The existence
```
pelle: This is incorrect. Our base layer at uPort is JSON+JWT's with JSON-LD an option for implementers.

David-Chadwick:
@pelle 
> As is right now, the insistence on pushing people to LD-Proofs is going to stall the widespread adoption of W3C VCs as people are having difficulties building it into their applications.

We have just added new text to the normative part of the specification that says that no proof method is standardised. So this might help you. The only thing that is mandated is that a proof property exists, and that it contains a type property to say what type of proof it is. So you can easily add an X.509 signature type, or an External signature type, or any other type that you think would allow folks to easily sign the verifiable credential using today's crypto libraries. Just name the type and define how it works. I believe that a register of proof types will also be created so that you can register it there. Then the marketplace will determine which proof type they like best.

awoie:
@David-Chadwick @msporny I tried to raise the question again during today's W3C VCWG call. Haven't we discussed that the tradeoffs section should be use case specific rather than listing a table? This was either early this year or before Christmas last year. I couldn't find the minutes. I still don't think it makes sense to compare the pros and cons of these technologies in an unfiltered way without referencing the VC data model.

index.html
msporny: Ok, thanks. Fixed in 0691a7e6f3adb294fa914e22c2fde56504bcf72d.

index.html
msporny: @pelle wrote: 
> Our base layer at uPort is JSON+JWT

Given your comment above, I was under the mistaken impression that no implementer had chosen to make JSON+JWT their base layer. Now that you've corrected that misunderstanding, I have renamed the section "Syntax and Proof Format Trade-Offs" for two reasons 1) reviewers wanted to know why they would choose JSON vs. JSON-LD, and 2) reviewers want to know why they would choose JWTs vs LD-Proofs.

The first 9 features discuss the differences between uPort's base layer, and the JSON-LD option you provide for developers.

> JSON-LD supports one exact view of "Open World" modelling.

The definition that we use is located in the PF1 elaboration section below. I have tried to use a consensus position shared by not just the JSON-LD community, but the RDF, Linked Data, and Credentials CG communities wrt. "open world data modelling". If you feel there are changes that could be made there that would garner broader consensus, please suggest concrete text and we can run it by the WG.

index.html
msporny: @pelle wrote: 
> Our base layer at uPort is JSON+JWT

Given your comment above, I was under the mistaken impression that no implementer had chosen to make JSON+JWT their base layer. Now that you've corrected that misunderstanding, I have renamed the section "Syntax and Proof Format Trade-Offs" for two reasons 1) reviewers wanted to know why they would choose JSON vs. JSON-LD, and 2) reviewers want to know why they would choose JWTs vs LD-Proofs.

index.html
msporny: Ah, I don't mean "keys" as in the "cryptographic key" sense. I mean "keys" as in Javascript Object properties -- key-value pairs -- sense. You are right, the language is confusing. The JSON spec calls these "names", which is also confusing. I've tried to apply a fix by using "properties", as in "property-value" statements in 43ff3769938b324059b9f9681c878a2c81d78a7d.

msporny: Also, I hope those reading this particular comment thread appreciate the irony that both @msporny and @pelle were miscommunicating specifically because English doesn't provide a `@context` feature either.

index.html
msporny: JSON doesn't differentiate between a string and a URL, which means that in order to interpret a pure JSON value as a URL requires you to add application logic. JSON-LD has this feature built in. It is true that you don't want to merge external information w/ something that is digitally signed without being very careful about it... but that's what the Hashlink specification is designed to protect against: https://tools.ietf.org/html/draft-sporny-hashlink-02. 

Hashlink is a bit beyond what this statement is saying though... which is just that JSON-LD has a mechanism to cleanly identify a URL, and then merge the data at that URL in w/ the local graph (and there exist several mechanisms... IPFS, Hashlinks, Magnet URIs, ni:/// URIs, that prevent the sort of disaster that happened w/ XMLSignature in the past).

index.html
msporny: Yes, JWTs have that, but only for certain fields... and the datatype is hard coded. JSON-LD enables you to state that "in this context, interpret the value of `foo` as a Date" without hardcoding that into a specification. It's dynamic and machine readable. JSON has no such feature.

I've attempted to clarify the language and the meaning in 89a0645b9e291ff027359d94dfd060f5175a551d.

index.html
msporny: I have tried to explain this in the explanation section via this text:

> JSON-LD's abstract data model supports the expression of information
as a directed graph of labeled nodes and edges, which enables an open world
data model to be supported. JSON's abstract data model only supports the
expression of information as a tree of unlabeled nodes and edges, which
restricts the types of relationships and structures that can be natively
expressed in the language.

index.html
msporny: The explanation section states why it's not possible without a pretty big rewrite for JWT (and is most likely why ZKPs are not being implemented on top of JWTs and why the current ZKP implementation is being built on top of LD-Proofs instead):

> The LD-Proof format is capable of modifying the algorithm that generates
the hash or hashes that are cryptographically signed. This cryptographic
agility enables digital signature systems, such as Zero-Knowledge Proofs,
to be layered on top of LD-Proofs instead of an entirely new digital signature
container format to be created. JWTs are designed such that an entirely new
digital signature container format will be required to support Zero-Knowledge
Proofs.

msporny: Do you have text that could clarify the point you're trying to make and balance the text above at the same time?

index.html
```diff
@@ -4201,6 +4207,458 @@ <h3>Device Theft and Impersonation</h3>
...
+syntaxes such as YAML, N-Quads, and CBOR.
+          </td>
+          <td class="missing">✖</td>
+          <td class="missing">✖</td>
+          <td class="supported">✓</td>
+        </tr>
+
+        <tr>
+          <td>
+<a href="#pf15">PF15.</a> Changing key-value ordering, or introducing
```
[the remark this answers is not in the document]
msporny: > There are no use cases where this would be a problem where JWT is used.

Here is one:

You get a JWT in and want to store the JSON in a JSON database. You first check the signature and then, in order to take advantage of the indexing properties in the JSON database, you base-64 decode the data and store it in the database. You are forced to store the JWT header and signature separately.

Time passes, you operate on the JSON data in the database, and then one day, you need to ship the whole header+payload+signature back out to the wire (like when a verifier requests a Verifiable Credential). You pull the JSON object out of the database, base-64 encode it, and send it off. The signature fails because your database rearranges some property-value pair, or injects whitespace in a different way than the entity that generated the signature in the first place.

This is a problem because it means you have to store all of the data twice. Once in base-64 encoded form to preserve the exact serialization that was signed, and once in JSON format, so the database can index the data properly.

index.html
msporny: The explanation elaborates on what this one liner is getting at:

> The LD-Proof format is naturally extensible, not requiring the format to be
extended in a formal international standards working group in order to
prevent namespace collisions. The JWT format requires entries in a centralized
registry and does not support experimentation as easily as the LD-Proof format
does. LD-Proof format extension is done through the decentralized publication
of cryptographic suites that are guaranteed to not conflict with other LD-Proof
extensions. This approach enables developers to easily experiment with new
cryptographic signature mechanisms that support selective disclosure,
zero-knowledge proofs, and post-quantum algorithms.

Is there concrete specification text that you can suggest that would summarize that paragraph?

msporny: I have tried to make the statement more aligned with the paragraph text in 12c109d39c56eccc9b5a6e8a79df85228a094451.

index.html
```diff
@@ -3127,6 +3133,208 @@ <h5>JWT Decoding</h5>
...
+              <th style="text-align: center;">JSON<br>+<br>JWTs</th>
+              <th style="text-align: center;">JSON&#8209;LD<br>+<br>JWTs</th>
+              <th style="text-align: center;">JSON&#8209;LD<br>+<br>LD&#8209;Proofs</th>
+            </tr>
+          </thead>
+
+          <tbody>
+            <tr>
+              <td>
+Support for open world data modelling.
```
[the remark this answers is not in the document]
msporny: Based on your input, I have added explanatory sections to each feature. There is now a section to describe what "open world data modelling" means here: https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/384.html#pf1

index.html
```diff
@@ -3127,6 +3133,208 @@ <h5>JWT Decoding</h5>
...
+              </td>
+              <td class="missing">✖</td>
```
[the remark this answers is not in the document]
msporny: Yes, that is true, but readers were having a hard time understanding what was and was not possible and were asking for a comparison to be done. This PR provides that comparison.

index.html
```diff
@@ -3127,6 +3133,208 @@ <h5>JWT Decoding</h5>
...
+              <td class="supported">✓</td>
+              <td class="supported">✓</td>
+            </tr>
+            <tr>
+              <td>
+Universal identifier mechanism for JSON objects via the use of URLs.
+              </td>
+              <td class="missing">✖</td>
```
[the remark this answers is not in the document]
msporny: It is missing in JSON. Perhaps you haven't seen the fully rendered table? https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/384.html

index.html
```diff
@@ -3127,6 +3133,208 @@ <h5>JWT Decoding</h5>
...
+              <td class="supported">✓</td>
+              <td class="supported">✓</td>
+            </tr>
+            <tr>
+              <td>
+A way to disambiguate keys shared among different JSON documents by mapping
+them to IRIs via a context.
+              </td>
+              <td class="missing">✖</td>
```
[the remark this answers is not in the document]
msporny: The comparison isn't just signature systems, but the combination of language plus signature systems (based on your input that uPort is doing JSON+JWTs, with an option to do JSON-LD+JWTs).

index.html
```diff
@@ -3127,6 +3133,208 @@ <h5>JWT Decoding</h5>
...
+Universal identifier mechanism for JSON objects via the use of URLs.
+              </td>
+              <td class="missing">✖</td>
+              <td class="supported">✓</td>
+              <td class="supported">✓</td>
+            </tr>
+            <tr>
+              <td>
+A way to disambiguate keys shared among different JSON documents by mapping
+them to IRIs via a context.
```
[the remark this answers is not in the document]
msporny: Based on your input, I have provided descriptions for every feature. The description for the particular feature you highlight above can now be found here: https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/384.html#pf3

index.html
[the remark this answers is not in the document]
David-Chadwick: I dont see that this is a big problem at all. VCs are small in size, so storing the signed VC "as is" makes perfect sense to me. If X.509/ISO presentation layer had recommended this then the distinguished encoding rules would not have been needed.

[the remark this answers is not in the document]
David-Chadwick: p.s. this is just another example of the CPU vs storage trade off. You prefer using CPU cycles to rebuild the signed VC, others might prefer to store the signed VC.

index.html
```diff
@@ -3127,6 +3133,208 @@ <h5>JWT Decoding</h5>
...
+Embeddable in HTML such that search crawlers will index the machine-readable
+content.
+              </td>
+              <td class="missing">✖</td>
+              <td class="missing">✖</td>
+              <td class="supported">✓</td>
+            </tr>
+            <tr>
+              <td>
+Data on the wire is easy to debug.
```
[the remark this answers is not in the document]
msporny: While that's debatable, I've updated the description to be more accurate:

> Data on the wire is easy to debug and serialize to database systems.

... and added the following explanation that is linked to:

> When developers are debugging software systems, it is beneficial for them to be
able to see the data that they are operating on using common debugging tools.
Similarly, it is useful to be able to serialize data from the network to a
database and then from the database back out to the network using a minimal
number of pre and post processing steps. LD-Proofs enable developers to use
common JSON tooling without having to convert the format into a different
format or structure. JWTs base-64 encode payload information, resulting in
complicated pre and post processing steps to convert the data into JSON data
while not destroying the digital signature. Similarly, schema-less databases,
which are typically used to index JSON data, cannot index information
that is expressed in an opaque base-64 encoded wrapper.

index.html
```diff
@@ -3127,6 +3133,208 @@ <h5>JWT Decoding</h5>
...
+Nesting signed data does not cause data size to double for every embedding.
+              </td>
+              <td class="missing">✖</td>
+              <td class="missing">✖</td>
+              <td class="supported">✓</td>
+            </tr>
+
+            <tr>
+              <td>
+Proof format supports Zero-Knowledge Proofs.
```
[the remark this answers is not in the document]
msporny: It is needed right now and it is not easy given some design limitations of JWTs. An explanation is provided for another comment you made on the same topic, which I'm copying to this response:

> The LD-Proof format is capable of modifying the algorithm that generates
the hash or hashes that are cryptographically signed. This cryptographic
agility enables digital signature systems, such as Zero-Knowledge Proofs,
to be layered on top of LD-Proofs instead of an entirely new digital signature
container format to be created. JWTs are designed such that an entirely new
digital signature container format will be required to support Zero-Knowledge
Proofs.

index.html
msporny: Based on your input, the language has been changed to this:

> PF16: Designed to easily support experimental signature systems.

with the following explanation:

> The LD-Proof format is naturally extensible, not requiring the format to be
extended in a formal international standards working group in order to
prevent namespace collisions. The JWT format requires entries in a centralized
registry in order to avoid collisions and does not support experimentation as easily as the LD-Proof format does. LD-Proof format extension is done through the decentralized publication
of cryptographic suites that are guaranteed to not conflict with other LD-Proof
extensions. This approach enables developers to easily experiment with new
cryptographic signature mechanisms that support selective disclosure,
zero-knowledge proofs, and post-quantum algorithms.

index.html
msporny: Hmm, the JSON-LD Proof/Signature libraries throw errors by default if anything is not signed in the data. Once you've verified a JSON-LD VC that was protected by LD-Proofs/Signatures, you know that everything in the data structure was protected. So, I don't understand what you're getting at here. Could you elaborate, please?

msporny:
> This whole section is still not very clear on it's purpose. It is meant as a proof format trade off but keeps slipping into a JSON-LD versions plain JSON comparison.

Based on your feedback, I have tried to clarify the purpose of the section in the introduction. I have also added explanations for every single feature that is being compared in order to communicate what each feature does with more clarify.

> In particular the following sections have nothing to do with Proof Formats:
> 
> * PF1, PF2, PF3, PF4, PF5, PF6, PF7

Yes, and based on your feedback (and other reader feedback) the text around the entire section has been corrected as a result. We had multiple reviewers request a comparison between all of the options, and since uPort is using JSON+JWT as their base layer, and because readers asked why one would use JSON-LD vs. JSON, that comparison is included.

> Besides that, there are several factual errors that I've pointed out in my comments.

I have attempted to address all factual errors you raised and added 19 paragraphs of elaboration on each feature listed in the table.

> Finally I still think it is really problematic to have this section here in the first place. It boils down to [bike shedding](http://bikeshed.com) which we are very good at in the technical community.

This section was requested by multiple reviewers who were confused by the options available to them.

> I would really like the LD-Proofs group work on creating better library support, documentation etc. Address the problems many people in the community has raised. As is right now, the insistence on pushing people to LD-Proofs is going to stall the widespread adoption of W3C VCs as people are having difficulties building it into their applications.

The LD-Proofs group seems to be working as hard as they can to advance implementations. Pull requests are welcome.

msporny:
> @David-Chadwick @msporny @stonematt I tried to raise the question again during today's W3C VCWG call. Haven't we discussed that the tradeoffs section should be use case specific rather than listing this table?

I vaguely having a conversation some time ago to avoid talking about proof formats in the specification. That was before the JWT section was added. Once that section was added, readers started asking for why they would pick JSON vs JSON-LD vs JWTs. So, this PR is responding to those (new) review comments. The latest discussion in the WG led to placing this information in a non-normative appendix, which there was support for doing during this call:

https://www.w3.org/2019/02/05-vcwg-minutes.html#item05

index.html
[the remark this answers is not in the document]
msporny: > VCs are small in size, so storing the signed VC "as is" makes perfect sense to me.

Some are small, some (like educational transcripts and healthcare records) are 100KB+ in size. 

> this is just another example of the CPU vs storage trade off

Note that the example I provided (database indexing) made it impossible to index the base-64 encoded information. So, it wasn't an either or trade-off, you were forced to do both things w/ JWTs (no choice) where you only need to do one thing w/ LD-Proofs (choice).

msporny:
@pelle wrote:
> I would really like the LD-Proofs group work on creating better library support, documentation

Let's focus on uPort for a second. This sounds like you're saying that if there was library support for uPort, that you would switch from JWTs to LD-Proofs? If so, we may be willing to find/spend money to build that library support out for your team. So, let me ask a direct question -- if we did that, would uPort switch to using LD-Proofs (along with supporting JWTs). If not, where is the line, at what point are you convinced that there is "enough" to support LD-Proofs?

index.html
rxgrant: DID method BTCR uses a proof format that is not technically a signature scheme.

rxgrant:
> I would really like the LD-Proofs group work on creating better library support, documentation etc.

Digital Contract Design is committing developer resources to the Java, C, and C++ ports.  We are developing software that incorporates the Javascript library in our binaries in order to compare implementations in our test suites.  This code is not yet ready for the public.

OR13:
Transmute is also developing Libraries to support LD-Proofs in JavaScript, with a focus on support for the DID Spec, which has a number of context attributes which align well with LD-Proofs.

jonnycrunch:
just adding myself to this issue, thanks @pelle for the plug for IPLD as an approach.  See my paper on this topic including linking signatures built in IPLD:  https://github.com/WebOfTrustInfo/rwot7-toronto/blob/master/final-documents/ipld-did.md

stonematt:
Barcelona F2F decision:  https://www.w3.org/2019/03/04-vcwg-minutes.html
Reformat Appendix section.  Rather than rendering a table of comparison, proponents will submit content describing each relevant proof/syntax combination.

- [ ] ACTION: oliver (@awoie) volunteers to write JSON+JWT and JSON-LD+JWT sections
- [ ] ACTION: manu (@msporny) volunteers to write JSON-LD proofs section

msporny:
Migrated this PR to the implementation guide: https://github.com/w3c/vc-imp-guide/commit/4009e552c8edc7cc4d5985432c242eb3795497d8

Issue #201, opened 2018-07-12
Typos


CipherQueen:
Attached please find a nine-page document with typos, spelling errors, and nits.

[Verifiable Credentials Data Model notes, V2.docx](https://github.com/w3c/vc-data-model/files/2188576/Verifiable.Credentials.Data.Model.notes.V2.docx)

stonematt:
There are a couple items in the attached document that aren't addressed in pr #219. I'll follow up on those and open new issues if necessary.

dlongley:
@stonematt, any updates here? Can we close this out?

rhiaro:
**The following issues are fixed in https://github.com/w3c/vc-data-model/pull/389**

> Section: 1. Introduction
> * Change “machine-verifiable” to “machine verifiable”

✔

> Section 9.9 Fitness for Purpose
> * Suggestion: “**it** may accept”

✔ (The exact sentence has now changed but the principle of the comment holds.)

> Section 9.11.5
> * Current (title of the section): “Long Lived Identifier-based Correlation”
> * Suggestion: “Long-Lived Identifier-Based Correlation”

✔

> Section 9.11.10 Validity Checks
> * Current: “a variety of business process specific checks”
> * Suggestion: “a variety of business-process-specific checks”

✔

> Section 9.11.12 Aggregation of Credentials
> * Current: “Two pieces of information about the same subject almost always reveals more”
> * Suggestion: “Two pieces of information about the same subject almost always reveal more”

✔ Rephrased to:

> **Holding** two pieces of information about the same <a>subject</a> almost always 
> reveals more about the <a>subject</a> than just a single piece of information, 
> even when **the information is** delivered through different channels.

**The following can't be fixed:**

> Section: Status of This Document
> * Add a comma after “replaced”

We can't modify the SotD, it's automatically generated W3C boilerplate.

**The following have not yet been addressed, but I can do so with some feedback.**

> Section: 1. Introduction
> Current: “Verifiable Credentials can be revoked by the issuer.”
> Question: Are holders notified when an issuer revokes the holder’s credentials?

I can't find text in the spec that answers that question. Should it be answered here? (Or it's out of scope?)

> Section 1.1 What is a..
> Current: “The addition of digital technologies, such as digital signatures, also make verifiable credentials more tamperproof and therefore more trustworthy than their physical counterparts.” 
> [...] Just because something is more tamperproof does not mean it is more trustworthy. [...]
[...]

PR #389, opened 2019-01-22
Fixes remaining editorial issues from #201

rhiaro:


dontcallmedom:

Issue #317, opened 2018-12-15
Only DOCUMENT conformance matters


burnburn:
TL;DR We need to ensure that only DOCUMENT conformance statements exist in the specification, rewriting or removing existing statements if necessary.

This is a specification for a data model and a few syntactic realizations of that data model.
We are not defining an API or a protocol, and thus there are no formal requirements on processors other than to state implementation of features of the data model / syntax as described.
It is appropriate to define what a 'conforming document' is.  Any statements about what a 'conforming processor' must do beyond 'support a conforming document as described in this specification' are suspect.

Here are several examples of W3C specifications that define a document format, as we do, without an execution semantics (except for JSON-LD itself, for which various transformation processes are defined).  For each, first take a look at the Implementation Report to see what ours might look like.  Then you can look at the spec itself for the conformance language they adopted.

DCAT: IR (http://www.w3.org/2011/gld/wiki/DCAT_Implementations), Spec conformance (https://www.w3.org/TR/vocab-dcat/#conformance)
EMMA: IR (https://www.w3.org/TR/emma/#s5), Spec conformance (https://www.w3.org/TR/emma/#s5)
JSON-LD: IR (N/A), Spec conformance (https://www.w3.org/TR/2014/REC-json-ld-20140116/#conformance)

Note in the above examples that requirements are generally around support of the syntactic features of the language.  Although some semantics can stray into the descriptions, they are typically co-occurrence or dependency statements, similar in principle to subject-verb agreement in English.

@stonematt and I believe this is the source of much of the scope confusion that has occurred in our discussions over the last few months, and in particular the concerns that Ted has (appropriately) raised.  As chairs, we want to see all of the conformance language in the specification (MUST, MAY, SHOULD, etc.) be checked and, if needed, rewritten in the form of requirements on the specification itself and not on processors.  If it is not possible to write it that way, it CANNOT be a normative requirement in our specification, although it could be informative text or moved to the Implementation Guide.

When evaluating whether the requirement is on the document itself, it may be helpful to try to phrase the requirement in the form "It is a property of a verifiable credential that XXXX", where XXXX is not allowed to refer to an Issuer, Holder, or Verifier (in short, a processor).  If this is possible, then the requirement is most likely a data model/syntax requirement and can be written as such.
You can also think of document conformance statements as "things that are true no matter when checked", while processor conformance statements are "things whose truth depends on when they are checked".

Examples of appropriate conformance statements (not necessarily correct, but of the right form)

[...]

PR #397, opened 2019-01-26
Clean up conformance section

burnburn:
This PR modifies the conformance section to only talk about document conformance as suggested in issue #317.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/397.html" title="Last updated on Jan 26, 2019, 5:32 PM UTC (987ca82)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/397/ce656da...987ca82.html" title="Last updated on Jan 26, 2019, 5:32 PM UTC (987ca82)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -710,21 +710,12 @@ A concrete expression of the data model in this specification is a
 this specification regarding syntax. That is, the content in the
 <a href="#basic-concepts">Basic Concepts</a>,
 <a href="#advanced-concepts">Advanced Concepts</a>, and
-<a href="#syntaxes">Syntaxes</a> sections of this document. For convenience,
-normative statements for <em>conforming documents</em> are
-often phrased as statements on the syntax used in properties and their
-associated values in the document (for example, "MUST be a URI", or
-"MUST be a string value of an ISO8601 combined date and time string").
+<a href="#syntaxes">Syntaxes</a> sections of this document.
       </p>
-
+      
       <p>
-A <em>conforming processor</em> is a software or hardware-based implementation
-of the normative statements in this specification regarding the expected
-contents of property-value pairs (for example, the content in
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -710,21 +710,12 @@ <h2>Trust Model</h2>
...
+      
       <p>
-A <em>conforming processor</em> is a software or hardware-based implementation
-of the normative statements in this specification regarding the expected
-contents of property-value pairs (for example, the content in
-<a>Verification</a>). For convenience, normative statements for
-<em>conforming processors</em> are often phrased as behavioral statements
-regarding the contents of property-value pairs (for example, "MUST NOT be
-revoked", or "MUST be in the expected range").
+This document makes no normative statements with regard to conformance of
```
msporny: Hmm, I think this is wrong. We need to make a statement about "conforming document processors". I believe you have identified "conformant behavior for roles" in the system, which the spec shouldn't say anything about.

I'm going to pull the PR in, but modify it to effectively state what I think you're going for (and thus it will be an editorial change).

msporny:
Cleaned this up after merging, please check: https://github.com/w3c/vc-data-model/commit/355ede3e21051f6cffdbe0af89d283d508b6bd4f

Note that while doing so, I found a huge gap in the spec. Specifically, we talk about Verification normatively and that section (for the most part) focuses on document conformance, but it presumes the existence of a conforming processor. So, I tried to add that concept to the specification and tied it directly to document conformance. This may mean that we have to remove some statements in the Verification section to make sure it only talks about document conformance.

burnburn:
Manu, you have made a change I disagree with.  Since the PR has already been merged by you, there is no way to "reopen" it, thus orphaning my comment here.  I will create a new issue for this.

Issue #391, opened 2019-01-22
Vocabulary location decisions


msporny:
We need to make two decisions before going into CR:

1. Where does the serviceEndpoint vocabulary term go? Which vocabulary?
2. Where does the security vocabulary go?

msporny:
@rhiaro fixed "Where does the serviceEndpoint vocabulary term go?" in https://github.com/w3c/vc-data-model/commit/0313cc4c8b116b0b330b7374651107d457dd0963.

To address the next issue, I recommend that the Chairs and staff contact work with W3C to acquire the following vocabulary URL at W3C:

https://www.w3.org/2018/security

The JSON-LD Context will go in:

https://www.w3.org/2018/security/v1
https://www.w3.org/2018/security/v2

We'll copy the https://w3id.org/security/v1 and https://w3id.org/security/v2 contexts over.

Reassigning to @burnburn @stonematt and @ashimura.

brentzundel:
Decision from F2F in Barcelona: The chairs will take the actions proposed by @msporny in his comment.
Specifically:
- [ ] contact work with W3C to acquire the following vocabulary URL at W3C: https://www.w3.org/2018/security
- [ ] Put the JSON-LD security contexts in 
    - https://www.w3.org/2018/security/v1
    - https://www.w3.org/2018/security/v2

burnburn:
Requested @ashimura  to obtain the URL.

burnburn:
@ashimura says W3C webmaster not happy with VCWG claiming the "security" context and that users will be confused by what it means. Chair decision to go with Kaz' suggestion of https://www.w3.org/2018/credentials/security for now.  Can always request another top-level vocab later.  Waiting for @ashimura .

msporny:
@ashimura, please map the following URLs:

https://www.w3.org/2018/credentials/v1
https://www.w3.org/2018/credentials/security/v1
https://www.w3.org/2018/credentials/security/v2
https://www.w3.org/2018/credentials/examples/v1

to the following locations:

https://w3c.github.io/vc-data-model/contexts/credentials/v1
https://w3c.github.io/vc-data-model/contexts/credentials/security/v1
https://w3c.github.io/vc-data-model/contexts/credentials/security/v2
https://w3c.github.io/vc-data-model/contexts/credentials/examples/v1

burnburn:
Done.

PR #399, opened 2019-01-28
Remove serviceEndpoint property, towards #391, fixes #380

### LINKED ISSUE #380: Suggest changing `refreshService`'s `serviceEndpoint` to `id` to match other objects
Effectively, [example 15](https://w3c.github.io/vc-data-model/#ex-15-usage-of-the-refreshservice-property-by-an-issuer) would change from:

```json
  "refreshService": {
    "type": "ManualRefreshService2018",
    "serviceEndpoint": "https://example.edu/refresh/283947"
  },
```
to
```json
  "refreshService": {
    "id": "https://example.edu/refresh/283947",
    "type": "ManualRefreshService2018"
  },
```
which would mirror the formats seen for [`credentialStatus`](https://w3c.github.io/vc-data-model/#ex-7-usage-of-the-status-property) and [`credentialSchema`](https://w3c.github.io/vc-data-model/#ex-14-usage-of-the-credentialschema-property-to-perform-zero-knowledge-validation) (among possibly others).

rhiaro:


BigBlueHat:
@msporny described a need for an `alternateEndpoint` property--which would solve for a need for fallback endpoints.

They looked like this when he proposed them today on the call:
```json
"serviceEndpoint": {
  "id": "https://example.com/refresh-service/123",
  "type": "ManualRefreshService2018",
  "alternateEndpoint": ["https://blue.com/refresh/567",  "https://red.com/refresh/567"]
}
```

I'm 👍 on this approach as it maps nicely to similar patterns used in other JSON & JSON-LD formats.

Will that be a separate PR? or will it be added to this one?

BigBlueHat:
This is related to #380

rhiaro:
@BigBlueHat I believe this will be handled in the DID spec rather than the VC Data Model, so nothing will be added to the example this PR

PR #401, opened 2019-01-30
Update VCDMExplainer.md

David-Chadwick:

### DIFF
--- a/VCDMExplainer.md
+++ b/VCDMExplainer.md
@@ -67,6 +67,11 @@ A visual depiction of the verifiable credentials ecosystem is shown below.
   <img src="diagrams/ecosystem.svg" width="100%" height="400">
 </a>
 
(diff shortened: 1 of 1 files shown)


VCDMExplainer.md
```diff
@@ -67,6 +67,11 @@ A visual depiction of the verifiable credentials ecosystem is shown below.
   <img src="diagrams/ecosystem.svg" width="100%" height="400">
 </a>
 
+It is worth pointing out why the data model has separate roles of holder and subject. It is recognised that in most cases the holder
```
davidlehn: - This doesn't need to say "it is worth pointing out" since the text wouldn't be here otherwise.
- Please adjust line lengths to match the rest of the file.

davidlehn:
Moved to #402 I think?  Should this be closed?

TzviyaSiegman:
@David-Chadwick If this issue is replaced by #402 , please close

David-Chadwick:
replaced by #402

PR #402, opened 2019-01-30
Update VCDMExplainer.md

David-Chadwick:

### DIFF
--- a/VCDMExplainer.md
+++ b/VCDMExplainer.md
@@ -67,6 +67,14 @@ A visual depiction of the verifiable credentials ecosystem is shown below.
   <img src="diagrams/ecosystem.svg" width="100%" height="400">
 </a>
 
+It should be noted that the data model has separate roles of holder 
+and subject. Whilst it is recognised that in many cases the holder
+of a verifiable credential will be the subject, in certain cases it
+will not be. For example, a parent (the holder) may hold the verifiable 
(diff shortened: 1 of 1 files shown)


VCDMExplainer.md
```diff
@@ -67,6 +67,14 @@ A visual depiction of the verifiable credentials ecosystem is shown below.
...
 </a>
 
+It should be noted that the data model has separate roles of holder 
+and subject. Whilst it is recognised that in many cases the holder
+of a verifiable credential will be the subject, in certain cases it
+will not be. For example, a parent (the holder) may hold the verifiable 
+credentials of a child (the subject), or a pet lover (the holder) may hold 
+the verifiable credentials of its pet (the subject). The advanced concepts
+section of the data model document deals with these special cases.
+
```
brentzundel: I have no problem with the role of subject being added to the explainer.
In my view, it would fit better along with the other definitions of ecosystem roles, rather than as a separate paragraph.

brentzundel:
Is this PR intended as a replacement of #401? If so, we should close that one.

VCDMExplainer.md
TzviyaSiegman: I agree with @brentzundel

David-Chadwick: No problem we can add subject to the set of roles. I had assumed because you had not done that originally, that you did not want to do this.

PR #404, opened 2019-01-30
Update VCDMExplainer.md

David-Chadwick:


brentzundel:
I'm fine with merging these changes, but will want to do an editorial pass once it is merged before considering the explainer ready to publish.

TzviyaSiegman:
thanks @David-Chadwick. I am merging with the assumption that @brentzundel will continue edting.

PR #407, opened 2019-01-31
Update features at risk section

burnburn:
Clarify Features at Risk with current state of affairs.


burnburn:
I think with Matt's modified commits this is ready to go.  Merging.

PR #412, opened 2019-02-02
Convert "SubjectOnly" property to termsOfUse use case.

msporny:
This PR suggests that the "SubjectOnly" use case can be handled by the termsOfUse property and suggests making that change. Need buy-in from @David-Chadwick that this approach would meet his needs.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/412.html" title="Last updated on Feb 10, 2019, 8:29 PM UTC (f97e05f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/412/8610f8f...f97e05f.html" title="Last updated on Feb 10, 2019, 8:29 PM UTC (f97e05f)">Diff</a>


David-Chadwick:
We have a had a long and difficult discussion about this topic already. The reason it was moved to a separate property is that the conformance statement for ToU is not correct for SubjectOnly, since SubjectOnly is equivalent to IssuanceDate and ExpirationDate in terms of conformance.

From a modelling and ontological perspective, all constraints on a VC can be regarded as Terms of Use. So SubjectOnly, IssuanceDate and ExpirationDate are all subclasses of the most generic ToU. The difference between these and some other ToUs, is that they are verifier and protocol independent (as was pointed out in the previous long discussion that we had), which some others are not.

So I would object strongly to moving SubjectOnly to ToU, unless all constraints, including IssuanceDate and ExpirationDate are moved as well.

brentzundel:
I don't see issuance date or expiration date as dependent on terms of use, but meta-data about the credential that could be used in any number of ways.

David-Chadwick:
@brentzundel Ditto subject only

David-Chadwick:
@manu. If we move SubjectOnly to ToU then we need to copy across the conformance clause as well since the verifier cannot independently determine whether the VC is valid or not, just as for IssuanceDate and ExpirationDate

brentzundel:
@David-Chadwick I think I disagree. Where expirationDate and issueDate are simply facts, the subjectOnly property is specifically intended to limit the use to which a credential may be used. Am I incorrect in thinking that if subjectOnly is true, the issuer intends that any holder who is not the subject should not use the credential? That sounds exactly like a term of use to me.

David-Chadwick:
@brentzundel . We definitely disagree, but this is possibly because we are talking about three separate issues: facts, usage and conformance.
1. Facts. SubjectOnly is a statement of fact like Issuance Date (and ExpirationDate). All are independent of the intended use. With all 3 constraints, anybody can present the VC to any verifier for any purpose at any time. However, for SubjectOnly, if anyone other than the subject presents the VC it is invalid, for IssuanceDate if anyone presents it before the IssuanceDate it is invalid. And the spec states this (for all three).
2. Usage. ExpirationDate like SubjectOnly are both constraints on use (so they could both be stated as Terms of Use, except for their different conformance statements). Both properties say the VC is invalid under a certain condition (i.e. after this date, or if not presented by the subject).
3. Conformance. TermsOfUse according to our current spec are somewhat different to the previous three since it does not talk about valid or invalid, viz: "If the recipient (a holder or verifier) is not willing to adhere to the specified terms of use, then they do so on their own responsibility and might incur legal liability if they violate the stated terms of use." If it for this last reason that SubjectOnly was made a property like IssuanceDate and ExpirationDate.

brentzundel:
One thing that keeps me from agreeing is that so much of what may or may
not be "valid" about a credential depends more on the verifier and their
use case than it depends on the issuers opinion on how the credential ought
to be used.
E.g. A US passport is only valid for travel after it has been issued,
before it is expired, and only by the subject. A US Passport is valid for
proof of citizenship after it has been issued, only by the subject, but
regardless of the expiration date. The "border security verifier" has a
different use case from the "citizenship for employment verifier."
A type of credential that is issued before its issue date might be a
coupon. And while the issuer of the coupon (the store) might intend that
the credential only be used after the issue date has passed, a competitor
who matches prices may not care about the issue date at all.
These are examples of issue date and expiration date being used other than
as possibly intended by the issuer, but in perfectly valid ways. This
indicates to me that issue date and expiration date have a role outside of
terms of use which might constrain them.
What I'm not seeing is a use case where subjectOnly could be considered
anything other than a term of use. Perhaps you could help me better
understand.

dlongley:
@David-Chadwick,

> 2\. ExpirationDate like SubjectOnly are both constraints on use (so they could both be stated as Terms of Use, except for their different conformance statements).

I disagree about ExpirationDate. "MustNotUseIfExpired" would be an equivalent, but that's not the same thing as ExpirationDate. "The issuer says X is true until this date" is not the same thing as "The issuer says you cannot use this VC beyond this date".

A verifier can, for example, use an expired VC as one piece of evidence about someone -- free and clear of any "terms of use" from the issuer (because there are none in that statement alone).

David-Chadwick:
@brentzundel > One thing that keeps me from agreeing is that so much of what may or may not be "valid" about a credential depends more on the verifier and their use case than it depends on the issuers opinion on how the credential ought to be used.

This is precisely where you are wrong. The issuer issues the VC and the issuer says when the data structure is valid or not valid, independent of any use of the VC or who the verifier is. It is the issuer's prerogative to do this. They are after all the issuer. 

Whether a verifier wishes to accept an invalid VC or not is outside the scope of the specification. But a verifier cannot make an invalid VC valid. It can accept and process an invalid VC, but it cannot make it valid.

David-Chadwick:
@dlongley I agree with you. This is why ExpirationDate, IssuanceDate and SubjectOnly are properties of the VC and are not tied into any terms of use. (Note this is why I said "except for their different conformance statements")

dlongley:
@David-Chadwick,

> I agree with you. This is why ExpirationDate, IssuanceDate and SubjectOnly are properties of the VC and are not tied into any terms of use. (Note this is why I said "except for their different conformance statements")

Both ExpirationDate and IssuanceDate can be understood independently from usage. I'm struggling to figure out how the same can be said of `SubjectOnly`. Isn't the meaning of the property "This VC can only be presented by the subject"? Such a definition is intrinsically bound to use; it is a directive.

dlongley:
VCs are a form of speech. You can trust that something was said by someone else (i.e. the issuer) through the use of cryptography.

IssuanceDate and ExpirationDate are both constraints on what an issuer has said. The first means "The issuer said X is true on this date" and the second means "The issuer says X is true until at least this date." Neither of these makes any statement about what the issuer says before the IssuanceDate or after the ExpirationDate (whether X be true *or* false). A recipient of a VC is welcome to use this sort of information to make whatever decisions they please.

My understanding is that SubjectOnly is a constraint on how someone may present what the issuer has said. It is not a constraint on what the issuer has said. The issuer certainly cannot be saying "I said X is true if Y tells it to you (otherwise I didn't say it)." This would cut against the core function of VCs, which is that you can trust that the issuer said X is true based on the cryptographic proofs on the VC, not on the channels through which that information travels. The main purpose of VCs is to eliminate dependence on those channels, instead allowing a verifier to rely upon cryptography.

An issuer can, of course, say that they would prefer that their speech not be used if it was not presented by a certain party (or for any other conceivable reason). But this is "Terms of Use".

msporny:
Ok, I think I've made the changes necessary to create normative tests around this use case in this commit:

https://github.com/w3c/vc-data-model/pull/412/commits/f97e05fc3f5ade1ccb7df139e01e3340a34279d6

@dlongley, could you please review. I'm trying to not put the ToS stuff in the core credentials context as it creates a number of new terms. I'm also trying to implement the compromise w/ @David-Chadwick where we can have normative tests in the test suite and "one way" to do "SubjectOnly" via Terms Of Use. The downside is that if we do that test, we have to do them all, which we haven't done to date. Again, this is a compromise, so trying to see if both sides of the debate find this path forward acceptable enough to not block the PR.

msporny:
I've now added Terms of Use tests to the test suite, one of which is to support SubjectOnly: https://github.com/w3c/vc-test-suite/commit/24054050c380ee8f5ab2ca2a699b3ba094a774b2

msporny:
At this point, I think all of the changes have been made that we discussed on the call last week. At present, this PR:

* Enables the SubjectOnly use case via termsOfUse.
* Adds a Terms of Use JSON-LD Context to enable testing of the use case (and other Terms of Use use cases such as prohibiting archival and 3rd party correlation).
* Adds test stubs to the test suite that enable these features in Terms of Use to be tested optionally. That is, there is one way to achieve them that the test suite supports, but implementing these aspects of terms of use are optional so that if we don't get enough implementations, it doesn't endanger the feature from being removed from the specification. I can explain more about this on the upcoming call.

I need @dlongley and @David-Chadwick to sign off on this PR and we can pull it into the spec.

David-Chadwick:
I am not happy with this change for two reasons
1. Specifying it using Terms of use is much more complex than a simple property, and does not carry the correct conformance statement.
2. If ToU was to be used, there are many different ways of specifying a subjectOnly ToU. You seem to have chosen a way that is more complex than it needs to be, and is much more complex than the original subjectOnly ToU.

Your only argument for replacing the SubjectOnly property is that no-one will implement it. But I find it hard to understand why anyone would prefer to implement the complex ToU structure that you have defined rather than the simple property that is currently defined, given that a substantial proportion of today's existing plastic card VCs contain a subject only statement (and therefore it is a popular and common requirement). Therefore I would let the implementors decide and leave the SubjectOnly property in the CR.

msporny:
> 1. Specifying it using Terms of use is much more complex than a simple property, and does not carry the correct conformance statement.

It is true that it is more complex, and it is true that it doesn't carry an equivalent conformance statement, but I've tried to increase the chances of this feature surviving CR due to lack of implementer support in its present state. The benefits gained have been outlined in the comment thread above.

> 2. If ToU was to be used, there are many different ways of specifying a subjectOnly ToU. You seem to have chosen a way that is more complex than it needs to be, and is much more complex than the original subjectOnly ToU.

It is more complex, but enables more use cases, for example: What happens if you want more than just the subject to be able to use it like the subject AND a guardian, or a subject AND a legal agent, or a subject AND an officer of the law. subjectOnly is a very specific ToU use case and is not useful for the previous use cases (among many others).

In addition, I don't think the group has had enough debate over this very specific term and the danger is that we put a feature in the specification that is too specific or misused.

> Your only argument for replacing the SubjectOnly property is that no-one will implement it. 

Yes, and the downside there is that we don't suggest anything in the specification. There will be no suggestion on how to do this.

> But I find it hard to understand why anyone would prefer to implement the complex ToU structure that you have defined rather than the simple property that is currently defined

Hopefully the above explains why -- subjectOnly is a very specific sub-use case of ToU, which we already have a way of expressing. @dlongley has noted why issuanceDate and expirationDate above are different from subjectOnly and are not ToU statements here: https://github.com/w3c/vc-data-model/pull/412#issuecomment-462070365

> given that a substantial proportion of today's existing plastic card VCs contain a subject only statement (and therefore it is a popular and common requirement). 

I just checked every single plastic and paper card in my wallet, thinking that I must have missed this language somewhere, but not a single one of my cards says anything approaching "subject only". Many of my bank cards say that the use of the card is pursuant to the cardholder agreement, and I opened one of my card holder agreements and I couldn't find any language that said that only I can use the card. It just listed what I'm liable for if the card is used with my consent. Perhaps SubjectOnly is a European thing?

> Therefore I would let the implementors decide and leave the SubjectOnly property in the CR.

Ooof, I really would hope we don't take that route, but if you insist, we will have to. So, options at this point are:

1. We compromise in some way that I'm not seeing, possibly changing the text "subjectOnly" to something more general like `"presentationAllowedBy": [array of entity ids that are allowed to present]`, which we would most likely not implement because it's too restrictive to a specific use case. ODRL already covers this use case and others.
2. The Chairs call for at least two implementers that would implement whatever is proposed, and if there are no commitments or interest in implementing by two implementers, we strike the feature before going into CR.
3. We go into CR marking it as a feature at risk and see what happens. If it doesn't get implemented by at least two implementers, then there is zero text that talks about subject only in the spec.

Thoughts, @David-Chadwick?

David-Chadwick:
@msporny "not a single one of my 12 cards says anything approaching "subject only"". 
Then I suggest you have a very narrow set of cards.
In the previous thread about "subject only" I gave examples of my cards that contain the equivalent of subject only. (Sorry I cannot remember the issue number). But here is a subset again: My
Bus pass says "the person named on the front of this pass...is entitled to the travel..."
Cathederal pass says "Not transferable"
Visa debit card says "The authorised signatory may only use this card.."
So certainly in the UK it is a common use case. And none of them say the subject plus some other person (which is what you propose for ToU)

So wrt to your 3 suggestions, I would say ask the chairs to do 2), for the two proposals (SubjectOnly property and SO ToU) and if two implementors can commit to implementing one of them, then this solution is better than no solution. If two implementors do not commit to either proposal, then I would go with 3) (the SO property) since I think it has much more chance of being implemented by people who are not part of the WG, then the SO ToU proposal has. (If I get funding I will implement the SO property).

Finally, lets not forget Occam's Razor. I think it has lessons for us here :-)

ken-ebert:
I agree with dlongley's analysis of "IssuanceDate" and "ExpirationDate" being statements regarding the credential and "SubjectOnly" being a Terms of Use object. If we are going to have a Subject only property, I think it belongs in the Terms of Use. The conformance language should be adjusted to match.

msporny:
@ken-ebert said:
> If we are going to have a Subject only property, I think it belongs in the Terms of Use.

Does that mean Sovrin (or Evernym, @brentzundel ?) would implement SubjectOnly using the Terms of Use property? If so, we have two implementers and a clear direction to go in for this PR. If not, we're still at a stand still for this PR. 

Note that "supporting" for the purposes of the test suite means "does not remove the SubjectOnly Terms of Use markup from the credential". It does not mean "take the value and does something with it".

brentzundel:
After much consideration, we have found a potential use case that might
make use of a terms of use property. If we pursue this use case, we would
probably support a "subjectOnly" property within terms of use.
We cannot foresee a use case that would require "subjectOnly" outside of
terms of use.

msporny:
@brentzundel said:
> If we pursue this use case, we would probably support a "subjectOnly" property within terms of use.

Great! @David-Chadwick, does this work for you? You have two implementers that are committing to support the feature outlined in this PR. Do you want harder commitments? Or do you want me to go ahead and pull this PR in?

ken-ebert:
I agree that if we are going to have "subjectOnly", it belongs in terms of use and not as a top-level property. However, any terms of use without cryptographic assurances are dangerous and give false assurance to credential issuers that a credential will only be used in a specific way. That is why we (Sovrin) are not planning to utilize them in their present form in any production system. I can't speak for Brent and Evernym.

I am hopeful that we will be able to strengthen terms of use in the future so that they can be cryptographically enforced.

brentzundel:
Just to state clearly, it is not definite that we will pursue the use case we are exploring. What @ken-ebert said is exactly right, we still need to work out cryptographic guarantees in order to safely use terms of use.

David-Chadwick:
@msporny  I am still not persuaded that your case is strong enough to replace the existing text for 2 reasons
1) the ToU you specify is far more complex than it needs to be
2) the conformance statement is not correct.

David-Chadwick:
I suggest that we change the Appendix "subjectOnly" property to the "notTransferrable" property. Rationale. This equates to what is already in use today in plastic cards. This property can then exist alongside any other complex terms of use properties that we want to define. The revised text would read:

The notTransferrable property states that the holder of the verifiable credential MUST at all times be the subject. A verifiable credential held by anyone other than the subject is invalid. 

Furthermore, a verifiable credential MUST only be encapsulated into a verifiable presentation whose proof was issued by the credentialSubject. A verifiable presentation that contains a verifiable credential containing the notTransferrable property, whose proof creator is not the credentialSubject, is invalid.

ken-ebert:
In ZKP Verifiable Credentials, credentials are not transferrable by nature
of the cryptography. All credentials, ZKP or otherwise, should by default
be non-transferable unless the issuer explicitly does something to enable
transfer at original credential creation time.

David-Chadwick:
@ken-ebert It is good to know that ZKP VCs already have the property that I am saying all VCs should optionally have. Thus we need a simple way of allowing non-ZKP VCs to have the same property that ZKP VCs have. I am suggesting that the not-transferrable property is it.

msporny:
Closing this PR, there is no consensus to merge it. It will be replaced with a separate PR that will rename subjectOnly to non-transferable.

PR #415, opened 2019-02-03
Update non-normative requirements

stonematt:
refine the original requirement list to reflect the intersection of the use case document and the data model specification.

section remains non-normative.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/415.html" title="Last updated on Feb 19, 2019, 3:45 PM UTC (583b7ae)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/415/ac6c5c9...583b7ae.html" title="Last updated on Feb 19, 2019, 3:45 PM UTC (583b7ae)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -349,26 +349,26 @@ As a result of documenting and analyzing the use cases document, a number of
 desirable ecosystem characteristics were identified for this specification,
 specifically:
         </p>
-
+<!-- requirement list start -->
         <ol>
           <li>
 <a>Verifiable Credentials</a> represent statements made by an <a>Issuer</a> in
 a tamper evident and privacy respecting manner.
           </li>
           <li>
-<a>Holders</a> assemble collections of claims from different issuers into a
+<a>Holders</a> assemble collections of credentials from different issuers into a
 single artifact, a <a>Verifiable Presentation</a>.
           </li>
           <li>
-<a>Issuers</a>  can issue <a>Verifiable Credentials</a> about any <a>subject</a> .
+<a>Issuers</a> can issue <a>Verifiable Credentials</a> about any <a>subject</a> .
           </li>
           <li>
 Acting as <a>Issuer</a>, <a>Holder</a>, or <a>Verifier</a> requires neither registration
 nor approval by any authority.
           </li>
           <li>
-<a>Verifiable Presentations</a>  allow any <a>Verifier</a> to verify the authenticity
-of claims from any Issuer.
+<a>Verifiable Presentations</a> allow any <a>Verifier</a> to verify the authenticity
+of credentials from any <a>Issuer</a>.
           </li>
           <li>
 <a>Holders</a> can receive <a>Verifiable Credentials</a> from anyone.
@@ -383,13 +383,13 @@ affecting their verifiability and without the <a>Issuer</a> knowing anything abo
 where they are stored or when they are accessed.
           </li>
           <li>
-<a>Holders</a> can present <a>Verifiable Presentations</a>  to any <a>Verifier</a>
+<a>Holders</a> can present <a>Verifiable Presentations</a> to any <a>Verifier</a>
 without affecting authenticity of the claims and without revealing that action
 to the Issuer.
           </li>
           <li>
-A <a>Verifier</a> can verify <a>Verifiable Presentations</a>  from any <a>Holder</a>,
-containing proofs of claims from any Issuer.
+A <a>Verifier</a> can verify <a>Verifiable Presentations</a> from any <a>Holder</a>,
+containing proofs of claims from any <a>Issuer</a>.
           </li>
           <li>
 Verification can't depend on direct interactions between <a>Issuers</a> and <a>Verifiers</a> . 
@@ -404,11 +404,11 @@ conformant software to support that feature.
           </li>
           <li>
 <a>Issuers</a>  can issue <a>Verifiable Credentials</a> which support
-selectivedisclosure.
+selective disclosure.
           </li>
           <li>
 If a <a>Verifiable Credential</a> supports selective disclosure, then <a>Holders</a>
-present proofs of claims without revealing the entire <a>Verifiable Credential</a>.
+may present proofs of claims without revealing the entire <a>Verifiable Credential</a>.
           </li>
           <li>
 <a>Issuers</a>  can issue <a>Verifiable Credentials</a> that are revocable.
@@ -418,12 +418,12 @@ present proofs of claims without revealing the entire <a>Verifiable Credential</
 serializable in one or more interoperable machine-readable data formats.
           </li>
           <li>
-The data-model and serialization must be extendable with minimal coordination.
+The datamodel and serialization must be extendable with minimal coordination.
           </li>
           <li>
 Revocation can't reveal to the <a>Issuer</a> any identifying information about
 the <a>Subject</a>, the <a>Holder</a>, the specific <a>Verifiable Credential</a>,
-nor the Verifier.
+nor the <a>Verifier</a>.
           </li>
           <li>
 <a>Issuers</a>  can disclose the reason for revocation.
--- a/index.html
+++ b/index.html
@@ -350,66 +350,93 @@ desirable ecosystem characteristics were identified for this specification,
 specifically:
         </p>
 
-        <ul>
+        <ol>
           <li>
-<a>Holders</a> receive and store <a>verifiable credentials</a> from
-<a>issuers</a>. <a>Holders</a> might interact with an <a>issuer</a> through an
-agent that the <a>issuer</a> does not need to trust.
+<a>Verifiable Credentials</a> represent statements made by an <a>Issuer</a> in
+a tamper evident and privacy respecting manner.
           </li>
           <li>
-<a>Holders</a> are positioned between <a>issuers</a> and <a>verifiers</a>
-and use <a>verifiable credentials</a> to make <a>verifiable presentations</a>
-to <a>verifiers</a>.
+<a>Holders</a> assemble collections of claims from different issuers into a
+single artifact, a <a>Verifiable Presentation</a>.
           </li>
           <li>
-<a>Holders</a> might interact with <a>verifiers</a> through an agent that
-<a>verifiers</a> do not need to trust; <a>verifiers</a> only need to trust
-<a>issuers</a>.
+<a>Issuers</a>  can issue <a>Verifiable Credentials</a> about any <a>subject</a> .
           </li>
           <li>
-<a>Verifiable credentials</a> are associated with <a>subjects</a>, not
-specific services; <a>holders</a> decide how to aggregate and manage
-<a>verifiable credentials</a> and the <a>verifiable presentations</a>
-associated with them.
+Acting as Issuer, <a>Holder</a>, or <a>Verifier</a> requires neither registration
+nor approval by any authority.
           </li>
           <li>
-<a>Subjects</a> can easily acquire and control their identifiers.
+<a>Verifiable Presentations</a>  allow any <a>Verifier</a> to verify the authenticity
+of claims from any Issuer.
           </li>
           <li>
-<a>Holders</a> control which <a>verifiable credentials</a> to use and when.
+<a>Holders</a> can receive <a>Verifiable Credentials</a> from anyone.
           </li>
           <li>
-<a>Holders</a> can freely choose and change the agents they employ to help
-them manage their <a>verifiable credentials</a> and generate and share
-<a>verifiable presentations</a>.
+<a>Holders</a> can interact with any <a>Issuer</a> and any <a>Verifier</a> through
+any User Agent.
           </li>
           <li>
-<a>Holders</a> can share <a>verifiable presentations</a> that can be verified
-without revealing the identity of the <a>verifier</a> to the <a>issuer</a>.
+<a>Holders</a> can store <a>Verifiable Credentials</a> in any location, without
+affecting their verifiability and without the <a>Issuer</a> knowing anything about
+where they are stored or when they are accessed.
           </li>
           <li>
-<a>Holders</a> in <a>presentations</a> can either disclose the attributes or
-satisfy <a>derived predicates</a> requested by the <a>verifier</a>.
-<a>Derived predicates</a> are boolean conditions like greater than, less than,
-equal to, is in set, and so on.
+<a>Holders</a> can present <a>Verifiable Presentations</a>  to any <a>Verifier</a>
+without affecting authenticity of the claims and without revealing that action
+to the Issuer.
           </li>
           <li>
-<a>Verifiable credentials</a> and <a>verifiable presentations</a> are
-expressed in one or more standard, machine-readable data formats, which can
-also be extended with minimal coordination.
+A <a>Verifier</a> can verify <a>Verifiable Presentations</a>  from any <a>Holder</a>,
+containing proofs of claims from any Issuer.
           </li>
           <li>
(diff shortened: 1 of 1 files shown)


stonematt:
submitted by @stonemat,t @jandrieu, @TzviyaSiegman, and @gannan08

index.html
```diff
@@ -350,66 +350,93 @@ <h3>Use Cases and Requirements</h3>
...
           </li>
-        </ul>
+          <li>
+The specification must provide a means for <a>Issuers</a>  to issue
+<a>Verifiable Credentials</a> that support selective disclosure, without requiring all
+conformant software to support that feature.
+          </li>
+          <li>
+<a>Issuers</a>  can issue <a>Verifiable Credentials</a> which support
+selectivedisclosure.
```
msporny: typo

index.html
```diff
@@ -350,66 +350,93 @@ <h3>Use Cases and Requirements</h3>
...
+          </li>
+          <li>
+If a <a>Verifiable Credential</a> supports selective disclosure, then <a>Holders</a>
```
msporny: This is not accurate, see the Selective Disclosure paper: "As with the strategy of data minimization, they may issue multiple, related, granular sub-credentials." -- you can support selective disclosure by atomizing and reconstructing credentials as well.

index.html
```diff
@@ -350,66 +350,93 @@ <h3>Use Cases and Requirements</h3>
...
+          </li>
+          <li>
+<a>Verifiable Credentials</a> and <a>Verifiable Presentations</a>  must be
+serializable in one or more interoperable machine-readable data formats.
+          </li>
+          <li>
+The data-model and serialization must be extendable with minimal coordination.
+          </li>
+          <li>
+Revocation can't reveal to the <a>Issuer</a> any identifying information about
```
msporny: -1, as this would imply that revocation can only be done in Zero Knowledge, we don't want to limit the specification to that.

msporny: Hmm, come to think of it, I don't understand what this statement is getting at.

index.html
```diff
@@ -350,66 +350,93 @@ <h3>Use Cases and Requirements</h3>
 specifically:
         </p>
 
-        <ul>
+        <ol>
```
msporny: Why is this a numbered list?

index.html
```diff
@@ -350,66 +350,93 @@ <h3>Use Cases and Requirements</h3>
...
           <li>
-<a>Holders</a> receive and store <a>verifiable credentials</a> from
-<a>issuers</a>. <a>Holders</a> might interact with an <a>issuer</a> through an
-agent that the <a>issuer</a> does not need to trust.
+<a>Verifiable Credentials</a> represent statements made by an <a>Issuer</a> in
```
msporny: Have you mapped all of the old requirements to the new requirements for the ecosystem? I can't easily tell if we're losing capabilities w/ the new list that we were asserting in the older list. The reason I'm concerned about this is that it took us 6+ months to come up with the old list, so I want to make sure we're not losing capability expectations.

index.html
```diff
@@ -350,66 +350,93 @@ <h3>Use Cases and Requirements</h3>
...
+a tamper evident and privacy respecting manner.
           </li>
           <li>
-<a>Holders</a> are positioned between <a>issuers</a> and <a>verifiers</a>
-and use <a>verifiable credentials</a> to make <a>verifiable presentations</a>
-to <a>verifiers</a>.
+<a>Holders</a> assemble collections of claims from different issuers into a
```
msporny: They assemble collections of credentials, not claims.

index.html
```diff
@@ -350,66 +350,93 @@ <h3>Use Cases and Requirements</h3>
...
-<a>Holders</a> can freely choose and change the agents they employ to help
-them manage their <a>verifiable credentials</a> and generate and share
-<a>verifiable presentations</a>.
+<a>Holders</a> can interact with any <a>Issuer</a> and any <a>Verifier</a> through
+any User Agent.
           </li>
           <li>
-<a>Holders</a> can share <a>verifiable presentations</a> that can be verified
-without revealing the identity of the <a>verifier</a> to the <a>issuer</a>.
+<a>Holders</a> can store <a>Verifiable Credentials</a> in any location, without
```
msporny: True in general, but problematic in practice. For example, if the issuer also provides the revocation list via their systems, and tracks access to the list, then they could tell when the credential is used.

index.html
```diff
@@ -350,66 +350,93 @@ <h3>Use Cases and Requirements</h3>
...
+containing proofs of claims from any Issuer.
           </li>
           <li>
-<a>Issuers</a> issue <a>verifiable credentials</a>, and <a>holders</a> store
-them. <a>Holders</a> offer <a>verifiable presentations</a>, and
-<a>verifiers</a> verify them. Issuing of <a>credentials</a>, storage of
-<a>credentials</a>, generation of <a>presentations</a>, sharing of
-<a>presentations</a>, and verification of <a>presentations</a> are
-each independent processes.
+Verification can't depend on direct interactions between <a>Issuers</a> and <a>Verifiers</a> .
```
msporny: Not always true, because you could have a verifier do a ping-back to get evidence.

index.html
```diff
@@ -350,66 +350,93 @@ <h3>Use Cases and Requirements</h3>
...
-them. <a>Holders</a> offer <a>verifiable presentations</a>, and
-<a>verifiers</a> verify them. Issuing of <a>credentials</a>, storage of
-<a>credentials</a>, generation of <a>presentations</a>, sharing of
-<a>presentations</a>, and verification of <a>presentations</a> are
-each independent processes.
+Verification can't depend on direct interactions between <a>Issuers</a> and <a>Verifiers</a> . 
           </li>
           <li>
-<a>Verifiable credentials</a> can be revoked by the <a>issuer</a>.
+Verification can't reveal the identity of the <a>Verifier</a> to any Issuer.
```
msporny: Again, not always true.

index.html
```diff
@@ -350,66 +350,93 @@ <h3>Use Cases and Requirements</h3>
...
+          </li>
+          <li>
+<a>Issuers</a>  can issue <a>Verifiable Credentials</a> that are revocable.
+          </li>
+          <li>
+<a>Verifiable Credentials</a> and <a>Verifiable Presentations</a>  must be
+serializable in one or more interoperable machine-readable data formats.
+          </li>
+          <li>
+The data-model and serialization must be extendable with minimal coordination.
```
msporny: no hyphen in data model.

index.html
msporny: "can't" should be "SHOULD NOT" or "MUST NOT" or something non normative. Don't know what you're going for here.

index.html
```diff
@@ -350,66 +350,93 @@ <h3>Use Cases and Requirements</h3>
 specifically:
         </p>
 
-        <ul>
+        <ol>
           <li>
-<a>Holders</a> receive and store <a>verifiable credentials</a> from
-<a>issuers</a>. <a>Holders</a> might interact with an <a>issuer</a> through an
```
msporny: I don't see this "does not need to trust" surface in the new list?

index.html
```diff
@@ -350,66 +350,93 @@ <h3>Use Cases and Requirements</h3>
...
           <li>
-<a>Holders</a> are positioned between <a>issuers</a> and <a>verifiers</a>
-and use <a>verifiable credentials</a> to make <a>verifiable presentations</a>
-to <a>verifiers</a>.
+<a>Holders</a> assemble collections of claims from different issuers into a
+single artifact, a <a>Verifiable Presentation</a>.
           </li>
           <li>
-<a>Holders</a> might interact with <a>verifiers</a> through an agent that
-<a>verifiers</a> do not need to trust; <a>verifiers</a> only need to trust
```
msporny: again, the trust between issuer/verifier only needs to exist. No trust between issuer/holder needs to happen. I may have missed where this point is made on the right.

index.html
```diff
@@ -350,66 +350,93 @@ <h3>Use Cases and Requirements</h3>
...
           <li>
-<a>Verifiable credentials</a> are associated with <a>subjects</a>, not
-specific services; <a>holders</a> decide how to aggregate and manage
-<a>verifiable credentials</a> and the <a>verifiable presentations</a>
-associated with them.
+Acting as <a>Issuer</a>, <a>Holder</a>, or <a>Verifier</a> requires neither registration
+nor approval by any authority.
           </li>
           <li>
-<a>Subjects</a> can easily acquire and control their identifiers.
```
msporny: I don't see this point on the right.

index.html
```diff
@@ -350,66 +350,93 @@ <h3>Use Cases and Requirements</h3>
...
           </li>
           <li>
-<a>Holders</a> can share <a>verifiable presentations</a> that can be verified
-without revealing the identity of the <a>verifier</a> to the <a>issuer</a>.
+<a>Holders</a> can store <a>Verifiable Credentials</a> in any location, without
+affecting their verifiability and without the <a>Issuer</a> knowing anything about
+where they are stored or when they are accessed.
           </li>
           <li>
-<a>Holders</a> in <a>presentations</a> can either disclose the attributes or
```
msporny: I don't see the derived predicates stuff in the new list?

index.html
```diff
@@ -350,66 +350,93 @@ <h3>Use Cases and Requirements</h3>
...
+A <a>Verifier</a> can verify <a>Verifiable Presentations</a>  from any <a>Holder</a>,
+containing proofs of claims from any Issuer.
           </li>
           <li>
-<a>Issuers</a> issue <a>verifiable credentials</a>, and <a>holders</a> store
-them. <a>Holders</a> offer <a>verifiable presentations</a>, and
-<a>verifiers</a> verify them. Issuing of <a>credentials</a>, storage of
-<a>credentials</a>, generation of <a>presentations</a>, sharing of
-<a>presentations</a>, and verification of <a>presentations</a> are
-each independent processes.
```
msporny: Don't see the point about independent processes on the right?

index.html
stonematt: To enable referencing a concept in discussions, etc. subjective decision, but made it easier when we were editing.

index.html
stonematt: We were careful and methodical, starting from the original list. I think the biggest change was dropping the specific term “derived predicate” to introduce selective disclosure language because “derived predicate” read like an implementation detail of a specific ZKP rather than a market requirement. (You called that out below). As expected, you have several comments that we’ll need to consider to make sure we’re all aligned.

msporny: Ok, makes me feel better about the rework that it was done in a methodical fashion, I'll most likely defer to each of you unless I see something glaring.

wrt. "derived predicate" -- agree that the language in the spec needs to change, it sticks out as a sore thumb. If we keep it, I suggest changing it to something like "derived attribute" or something of that nature.

index.html
stonematt: "may present proofs" ok?

index.html
stonematt: the implication is that the Issuer shouldn't necessarily glean insight to the use of a VC or the verifier community as a result of revocation.  It should be just as anonymous and privacy respecting as the original issuance....

index.html
stonematt: leaving as <ol> unless you strongly protest.

index.html
brentzundel: I strongly feel that if a verifier needs more information, they should be seeking it from the holder rather than the issuer.

index.html
brentzundel: perhaps this should say shouldn't instead of can't

index.html
brentzundel: This sentence is fine: If a (single) Verifiable credential supports selective disclosure, then Holders may present proofs of claims without revealing the entire Verifiable Credential.
For greater accuracy, another sentence could be added to mention the ability to support data minimization using multiple granular credentials.

index.html
brentzundel: I agree with @stonematt here.

index.html
```diff
@@ -349,67 +349,94 @@ <h3>Use Cases and Requirements</h3>
...
           <li>
-<a>Holders</a> can share <a>verifiable presentations</a> that can be verified
-without revealing the identity of the <a>verifier</a> to the <a>issuer</a>.
+<a>Holders</a> can store <a>Verifiable Credentials</a> in any location, without
+affecting their verifiability and without the <a>Issuer</a> knowing anything about
+where they are stored or when they are accessed.
           </li>
           <li>
-<a>Holders</a> in <a>presentations</a> can either disclose the attributes or
-satisfy <a>derived predicates</a> requested by the <a>verifier</a>.
```
brentzundel: I'd like derived predicates to remain in the list. 
It is an important feature that not only may a presentation not contain the entire credential (selective disclosure of claims), but also only reveal a predicate based on the original value rather than the value itself. I think "derived predicate" is the most accurate description of this (it is a predicate derived from a claim).

index.html
brentzundel: I think this is fine, "stored and accessed" without the issuer knowing is different than "used" without the issuer knowing.

index.html
msporny: There is some evidence that a verifier will not be able to get from a holder, such as the current status of a background check on the individual, or the current status of a medical debarment, if any.

index.html
msporny: +1 for the addition of "single", and the further sentence @brentzundel has outlined.

index.html
```diff
@@ -404,26 +404,31 @@ <h3>Use Cases and Requirements</h3>
...
-<a>Issuers</a>  can issue <a>Verifiable Credentials</a> that are revocable.
+<a>Issuers</a> can issue <a>Verifiable Credentials</a> that are revocable.
           </li>
           <li>
 <a>Verifiable Credentials</a> and <a>Verifiable Presentations</a>  must be
 serializable in one or more interoperable machine-readable data formats.
           </li>
           <li>
-The data-model and serialization must be extendable with minimal coordination.
+The datamodel and serialization must be extendable with minimal coordination.
```
msporny: "data model" -- two words. :)

index.html
msporny: "Issuer shouldn't necessarily glean insight to the use of a VC or the verifier community as a result of revocation" -- I don't understand this sentence... and I don't understand how it is related to "Revocation can't reveal to the <a>Issuer</a> any identifying information about the <a>Subject</a>, the <a>Holder</a>, the specific <a>Verifiable Credential</a>, nor the Verifier."

Are you saying, "*When a Verifiable Credential's status is checked, for example to determine if the credential has been revoked*, the process should not reveal to the <a>Issuer</a> any identifying information about the <a>Subject</a>, the <a>Holder</a>, the specific <a>Verifiable Credential</a>, nor the Verifier."

The language you're using is confusing because it makes it seem like you could be saying "the act of revocation" vs. "the act of checking to see if a credential has been revoked". I think you mean the latter, but the language you're using isn't clear about that.

Also, it depends on the status checking process and ZKP is not the only mechanism implementers are using to ensure privacy preservation. Herd privacy and blockchain-based revocation lists are two other approaches. In all 3 approaches, you can determine who the verifier is if you have control over the website/gateway/or blockchain node, for example. So, I think softening the wording is appropriate. We all know what we're going for, but the details matter and make, for example, preventing the verifier from being identified in all cases not possible.

index.html
brentzundel: That makes sense. Perhaps if the statement was softened and used "shouldn't" instead of "cant" it would be better.

index.html
brentzundel: Perhaps "Checking revocation status shouldn't reveal to the issuer" etc.

PR #416, opened 2019-02-03
Update revocation to reference credentialStatus

stonematt:
Update revocation language to be more explicit about what might be revoked and where the instructions would be expressed.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/416.html" title="Last updated on Feb 3, 2019, 8:43 PM UTC (d01d94d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/416/ac6c5c9...d01d94d.html" title="Last updated on Feb 3, 2019, 8:43 PM UTC (d01d94d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3010,7 +3010,7 @@ date of a <a>verifiable credential</a> is not in the past.
         <h3>Status Revocation</h3>
 
         <p>
-If the <code>credentialStatus</code> propery is available the status of the 
+If the <code>credentialStatus</code> property is available, the status of the 
 claim MUST be in good standing according to the <code>credentialStatus</code> 
 <a>type</a> definition for the <a>credential</a>. For example, a <a>verifier</a>
 can ensure that the status hasn't been withdrawn for cause by the <a>issuer</a>.
--- a/index.html
+++ b/index.html
@@ -3012,7 +3012,8 @@ date of a <a>verifiable credential</a> is not in the past.
         <p>
 If the <code>credentialStatus</code> propery is available the status of the 
 claim MUST be in good standing according to the <code>credentialStatus</code> 
-<a>type</a> definition for the <a>credential</a>.
+<a>type</a> definition for the <a>credential</a>. For example, a <a>verifier</a>
+can ensure that the status hasn't been withdrawn for cause by the <a>issuer</a>.
         </p>
       </section>
 
--- a/index.html
+++ b/index.html
@@ -3011,8 +3011,8 @@ date of a <a>verifiable credential</a> is not in the past.
 
         <p>
 If the <code>credentialStatus</code> propery is available the status of the 
-claim MUST be in good standing according to the credential status scheme of the
-<code>credentialStatus</code> <a>type</a> defintion for the <a>credential</a>.
+claim MUST be in good standing according to the <code>credentialStatus</code> 
+<a>type</a> definition for the <a>credential</a>.
         </p>
       </section>
 
--- a/index.html
+++ b/index.html
@@ -3010,8 +3010,8 @@ date of a <a>verifiable credential</a> is not in the past.
         <h3>Revocation</h3>
 
         <p>
-If revocation instructions are present, the <a>credential</a> MUST NOT
-have been revoked.
+If a proof is present, the keys used to sign the document must not have been 
+revoked or otherwised deemed compromised and untrustworthy.
         </p>
       </section>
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3010,8 +3010,8 @@ <h3>Expiration</h3>
         <h3>Revocation</h3>
 
         <p>
-If revocation instructions are present, the <a>credential</a> MUST NOT
-have been revoked.
+If a proof is present, the keys used to sign the document must not have been 
+revoked or otherwised deemed compromised and untrustworthy.
```
stonematt: ignore this commit.  I refined the text in the wrong direction.

index.html
```diff
@@ -3007,11 +3007,13 @@ <h3>Expiration</h3>
       </section>
 
       <section>
-        <h3>Revocation</h3>
+        <h3>Status Revocation</h3>
```
msporny: I wouldn't rename the section, as it will probably confuse readers. What exactly is "status revocation"? I know what revocation is... but status revocation sounds like I'm revoking the status of the credential. Suggest moving it back to just "Revocation".

index.html
```diff
@@ -3007,11 +3007,13 @@ <h3>Expiration</h3>
...
 
         <p>
-If revocation instructions are present, the <a>credential</a> MUST NOT
```
msporny: I agree that this is terrible language and should be fixed.

index.html
```diff
@@ -3007,11 +3007,13 @@ <h3>Expiration</h3>
...
-have been revoked.
+If the <code>credentialStatus</code> propery is available the status of the
```
msporny: Editorial nits:

* "propery" is misspelled
* Missing comma between "available, the"

index.html
```diff
@@ -3007,11 +3007,13 @@ <h3>Expiration</h3>
...
-        <h3>Revocation</h3>
+        <h3>Status Revocation</h3>
 
         <p>
-If revocation instructions are present, the <a>credential</a> MUST NOT
-have been revoked.
+If the <code>credentialStatus</code> propery is available the status of the 
+claim MUST be in good standing according to the <code>credentialStatus</code> 
+<a>type</a> definition for the <a>credential</a>. For example, a <a>verifier</a>
+can ensure that the status hasn't been withdrawn for cause by the <a>issuer</a>.
```
msporny: Suggest rewriting to "ensure that the credential has not been suspended or revoked by the issuer". I think we should be more direct about what this property is about since it's caused confusion by being too generic about it before.

index.html
stonematt: fixed

index.html
stonematt: I didn't want to over-specify status and get in the way of definitions that might be introduced and accepted in the "credentialStatus type" registry - it's so open world right now, we have to be careful.  I'm not sure how I would make it better.

index.html
stonematt: ... since we're invoking the `credentialStatus` property, I thought "Status Revocation" would be ok.  It's not a strongly held opinion... :)

brentzundel: But the value of the credentialStatus property describes the status of the credential, i.e. whether or not it had been revoked. 
My understanding is that it is the credential that may be revoked, not the status of the credential.

stonematt: @brentzundel we're getting a little meta here....  The fact is the `credentialStatus` property is where we tell Verifiers to look for this information -- there isn't a `revocation` service...  As an issuer, if I issue and invalidate a credential, I would indicate that in the credentialStatus service in our specification (assuming I'm not doing something cool like you do w/ the distributed ledger, where I can just "disconnect" it :) )

index.html
stonematt: @msporny any suggestions here?  Is it wrong, confusing, or "ok, but could be better" - I'd like to merge and move on if it's "ok, but could be better".

index.html
msporny: I don't think we're getting meta. Both @brentzundel and I are reading the title rename in the same way and we find it confusing. For example, why didn't you rename it to "Status Suspension" or "Status Retraction". Why did you pick "Revocation" as the title instead of those other two things, both of which are perfectly reasonable things to associate with the `credentialStatus` property.

Does that help explain why @brentzundel and I find the proposed renaming confusing?

index.html
msporny: Ok, but could be better. :)

index.html
msporny: I'm going to merge this, and undo the title change... in an attempt to move things forward. Feel free to submit another PR if you feel the title change is necessary.

Issue #406, opened 2019-01-31
Authorisation


David-Chadwick:
The current text in section 7.10 states "this specification MUST NOT be used for authorization purposes without an accompanying authorization framework." This is not a statement that can be enforced, nor can you mandate that implementations will not use the VC data model document for authorisation applications. Therefore we should change the text to 
"It is recommended that this specification SHOULD NOT be used for authorization purposes without an accompanying authorization framework."

jandrieu:
Agreed with the point of the correction although my concern is that "authorization framework" is ill defined. Shifting to "SHOULD" helps, but my first take was to replace "accompanying" to "appropriate" to support a bit more flexibility in the definition of the term.

stonematt:
we'll discuss this in the PR as we refine spec text

stonematt:
closed w/ PR

PR #417, opened 2019-02-03
Update authorization language

stonematt:
fixes issue #406

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/417.html" title="Last updated on Feb 10, 2019, 3:40 PM UTC (fbfaf0c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/417/ac6c5c9...fbfaf0c.html" title="Last updated on Feb 10, 2019, 3:40 PM UTC (fbfaf0c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2324,10 +2324,10 @@ as the specific vocabulary terms for disputed claims.
         <p>
 <a>Verifiable credentials</a> are intended as a means of reliably identifying
 <a>subjects</a>. While it is recognized that Role Based Access Controls (RBAC) and
-Attribute Based Access Controls (ABAC) rely on this identification as a means
-of authorizing <a>subjects</a> to access resources, this specification
-SHOULD NOT be used for authorization purposes without an appropriate
-authorization framework.
+Attribute Based Access Controls (ABAC) rely on this identification as a means of
+authorizing <a>subjects</a> to access resources, this specification does not
+provide a complete solution for RBAC or ABAC - authorization is not an appropriate
+for this specification.
         </p>
 
         <p>
--- a/index.html
+++ b/index.html
@@ -2327,7 +2327,7 @@ as the specific vocabulary terms for disputed claims.
 Attribute Based Access Controls (ABAC) rely on this identification as a means of
 authorizing <a>subjects</a> to access resources, this specification does not
 provide a complete solution for RBAC or ABAC - authorization is not an appropriate
-for this specification.
+use for this specification.
         </p>
 
         <p>
--- a/index.html
+++ b/index.html
@@ -2326,7 +2326,7 @@ as the specific vocabulary terms for disputed claims.
 <a>subjects</a>. While it is recognized that Role Based Access Controls (RBAC) and
 Attribute Based Access Controls (ABAC) rely on this identification as a means
 of authorizing <a>subjects</a> to access resources, this specification
-MUST NOT be used for authorization purposes without an accompanying
+SHOULD NOT be used for authorization purposes without an appropriate
 authorization framework.
         </p>
 
--- a/index.html
+++ b/index.html
@@ -2327,7 +2327,7 @@ as the specific vocabulary terms for disputed claims.
 Attribute Based Access Controls (ABAC) rely on this identification as a means of
 authorizing <a>subjects</a> to access resources, this specification does not
 provide a complete solution for RBAC or ABAC - authorization is not an appropriate
-use for this specification.
+use for this specification without an accompanying authorization framework.
         </p>
 
         <p>


index.html
```diff
@@ -2326,7 +2326,7 @@ <h3>Authorization</h3>
 <a>subjects</a>. While it is recognized that Role Based Access Controls (RBAC) and
 Attribute Based Access Controls (ABAC) rely on this identification as a means
 of authorizing <a>subjects</a> to access resources, this specification
-MUST NOT be used for authorization purposes without an accompanying
+SHOULD NOT be used for authorization purposes without an appropriate
```
msporny: I think both "MUST NOT" and "SHOULD NOT" are wrong. We can't enforce those statements and having MUSTs and SHOULDs that are not enforceable via software is impossible to test. Suggest "this specification is not appropriate when used for authorization purposes", or something to that effect. Basically, we want to say -- don't use this specification directly for authorization, you need to have considered how this specification fits into a larger authorization framework.

index.html
```diff
@@ -2324,10 +2324,10 @@ <h3>Authorization</h3>
...
 <a>Verifiable credentials</a> are intended as a means of reliably identifying
 <a>subjects</a>. While it is recognized that Role Based Access Controls (RBAC) and
-Attribute Based Access Controls (ABAC) rely on this identification as a means
-of authorizing <a>subjects</a> to access resources, this specification
-MUST NOT be used for authorization purposes without an accompanying
-authorization framework.
+Attribute Based Access Controls (ABAC) rely on this identification as a means of
+authorizing <a>subjects</a> to access resources, this specification does not
+provide a complete solution for RBAC or ABAC - authorization is not an appropriate
+use for this specification.
```
brentzundel: I think this would be better if it ended with "without an accompanying authorization framework."

index.html
David-Chadwick: Our document should not be schizophrenic about authorisation. Firstly it says a VC is  "a standard way to express credentials on the Web" (defn of credential: evidence of authority, status, rights, entitlement to privileges, or the like,...) then it says "These credentials provide benefits to us when used in the physical world". Now if neither of these statements are about authorisation I am not sure what is.

So if later on we want to say as @brentzundel suggests "authorization is not an appropriate use for this specification" then we are surely going to confuse our readers.

So let's be up front and honest. This spec is about authorisation, but on its own it is not sufficient as the Mode of Operation section clearly states.

index.html
David-Chadwick: @brentzundel the latter phrase is essential is my opinion

index.html
jandrieu: Credentials are a standard way for issuers to make a verifiable statement about someone. That need not be any sort of authorization at all. When Equifax makes a statement about my credit rating, that's not an authorization. Period.

Please stop bringing this up. We all realize that people will use VCs for authorization. Most of us also realize that this is a bad idea without further design considerations, including threat modeling, privacy design, and risks of recentralization. ALL of which would be a part of a proper authorization framework. Which VCs are not. Full stop.

David-Chadwick: @jandrieu Why would anyone gratuitously collect Equifax credit ratings? I dont think they would. They would only get them if they needed them for something. The only times I am aware of them being collected about me is when I have applied for something like a mobile phone contract. So they have in fact been part of the authorisation process (and indeed I was once refused a mobile phone contract because of the (what turned out to be erroneous) Equifax credit report.)

We are not disagreeing that an authz framework is beneficial to supplement the VC specification, but we should not make hard and fast statements such as that proposed by @brentzundel  ("authorization is not an appropriate use for this specification") since this contradicts many other statements in the specification.

You say "this spec is about authorisation". But to get there you seem to equate "doing something for a purpose" with "authorization" (e.g. one doesn't collect Equifax reports gratuitously, they do it with some goal in mind). I don't think that's how most people think or should think about authorization.

URLs help us do things. Why else would we use them? (This seems to be begging the question, really). However, the URL spec is not about authorization. It is quite clear, however, that you can *use* URLs to help you within a larger authorization framework.

I would not disagree with an argument that "speech is generally for some purpose". VCs are a form of speech -- and therefore they are for some purpose. But I think it is a strawman to argue that anyone here thinks VCs are for no purpose (i.e. for merely gratuitous collection). With VCs, "some purpose" is quite open ended from the issuer's perspective. If it were not, then we'd be talking about a closed system where the issuer and the verifier are essentially the same party. In fact, this open-ended nature is precisely *why* they are a bad tool (on their own) for authorization.

I may issue a VC saying that I know Brent and hand it to him. I have no idea what he might do with that information, but I hope it would help him. Saying that this is about "authorization" is likely to only cause confusion. The only way that this is about authorization is in that I've given Brent something he can use to prove I said something. I've authorized him to show others my claim -- a claim that I can revoke at any time. But, to be honest, the semantics of that particular VC are perhaps irrevocable -- one can't "unknow" Brent ... he's just that special. Anyway, the authorization here goes no further.

So the extent to which the VC spec is "about authorization" is only in that VCs enable party A to present something to party B that party C has said whilst preserving trusted authorship. This has nothing to do with party B's authorization policy (or what they decide to do with the information) -- the spec is entirely silent on that point. In this sense (which I believe to be the common sense), the spec is *not* about authorization.

index.html
stonematt: done -- ready to merge?

index.html
stonematt: this authorization language has changed quite a bit since your comments.  Please do a new review.

PR #418, opened 2019-02-03
Update Evidence language

stonematt:
fixes issue #405

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/418.html" title="Last updated on Feb 10, 2019, 4:17 PM UTC (71c440e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/418/ac6c5c9...71c440e.html" title="Last updated on Feb 10, 2019, 4:17 PM UTC (71c440e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2044,7 +2044,16 @@ credentials are supported by the specification.
     "evidenceDocument": "DriversLicense",
     "subjectPresence": "Physical",
     "documentPresence": "Physical"
-  }]</span>,
+  },{
+    "id": "https://example.edu/evidence/f2aeec97-fc0d-42bf-8ca7-0548192dxyzab",
+    "type": ["SupportingActivity"],
+    "verifier": "https://example.edu/issuers/14",
+    "evidenceDocument": "Fluid Dynamics Focus",
+    "subjectPresence": "Digital",
+    "documentPresence": "Digital"
+  }
+  ]
+</span>,
   "proof": { ... }
 }
         </pre>
--- a/index.html
+++ b/index.html
@@ -1981,7 +1981,7 @@ the presentation.
 Evidence is used by an <a>issuer</a> to inform the <a>verifier</a> of the basis
 for the <a>verifiable credential</a>. This could be used by the <a>verifier</a>
 to establish the confidence with which it relies on the claims in the
-<a>verifiable credential</a>. For example, an issuer might check physical 
+<a>verifiable credential</a>. For example, an <a>issuer</a> might check physical 
 documentation provided by the <a>subject</a> or might perform a set of background
 checks before issuing the credential. In certain scenarios, this information is
 useful to the <a>verifier</a> when determining the risk associated with relying
--- a/index.html
+++ b/index.html
@@ -2051,9 +2051,7 @@ credentials are supported by the specification.
     "evidenceDocument": "Fluid Dynamics Focus",
     "subjectPresence": "Digital",
     "documentPresence": "Digital"
-  }
-  ]
-</span>,
+  }]</span>,
   "proof": { ... }
 }
         </pre>
--- a/index.html
+++ b/index.html
@@ -1978,12 +1978,14 @@ the presentation.
         <h3>Evidence</h3>
 
         <p>
-Evidence is used by an <a>issuer</a> to determine whether or not to issue
-a <a>credential</a>. For example, an issuer might check physical documentation
-provided by the <a>subject</a> or might perform a set of background checks
-before issuing the credential. In certain scenarios, this information is useful
-to the <a>verifier</a> when determining the risk associated with relying on a
-given credential.
+Evidence is used by an <a>issuer</a> to inform the <a>verifier</a> of the basis
+for the <a>verifiable credential</a>. This could be used by the <a>verifier</a>
+to establish the confidence with which it relies on the claims in the
+<a>verifiable credential</a>. For example, an issuer might check physical 
+documentation provided by the <a>subject</a> or might perform a set of background
+checks before issuing the credential. In certain scenarios, this information is
+useful to the <a>verifier</a> when determining the risk associated with relying
+on a given credential.
         </p>
 
         <p>
--- a/index.html
+++ b/index.html
@@ -1978,14 +1978,16 @@ the presentation.
         <h3>Evidence</h3>
 
         <p>
-Evidence is used by an <a>issuer</a> to inform the <a>verifier</a> of the basis
-for the <a>verifiable credential</a>. This could be used by the <a>verifier</a>
-to establish the confidence with which it relies on the claims in the
-<a>verifiable credential</a>. For example, an <a>issuer</a> might check physical 
-documentation provided by the <a>subject</a> or might perform a set of background
-checks before issuing the credential. In certain scenarios, this information is
-useful to the <a>verifier</a> when determining the risk associated with relying
-on a given credential.
+Evidence may be included by an <a>issuer</a> to provide the <a>verifier</a>
+additional supporting information in a <a>verifiable credential</a>. This could
+be used by the <a>verifier</a> to establish the confidence with which it relies
+on the claims in the <a>verifiable credential</a>. 
+        </p>
+        <p>
+For example, an <a>issuer</a> might check physical documentation provided by the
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1978,12 +1978,14 @@ <h3>Terms of Use</h3>
...
-Evidence is used by an <a>issuer</a> to determine whether or not to issue
-a <a>credential</a>. For example, an issuer might check physical documentation
-provided by the <a>subject</a> or might perform a set of background checks
-before issuing the credential. In certain scenarios, this information is useful
-to the <a>verifier</a> when determining the risk associated with relying on a
-given credential.
+Evidence is used by an <a>issuer</a> to inform the <a>verifier</a> of the basis
+for the <a>verifiable credential</a>. This could be used by the <a>verifier</a>
+to establish the confidence with which it relies on the claims in the
+<a>verifiable credential</a>. For example, an issuer might check physical
```
msporny: Missing &lt;a>&lt;/a> around issuer.

index.html
```diff
@@ -1978,12 +1978,14 @@ <h3>Terms of Use</h3>
         <h3>Evidence</h3>
 
         <p>
-Evidence is used by an <a>issuer</a> to determine whether or not to issue
-a <a>credential</a>. For example, an issuer might check physical documentation
-provided by the <a>subject</a> or might perform a set of background checks
-before issuing the credential. In certain scenarios, this information is useful
-to the <a>verifier</a> when determining the risk associated with relying on a
-given credential.
+Evidence is used by an <a>issuer</a> to inform the <a>verifier</a> of the basis
```
msporny: elaborate more on what "basis" means. Be more concrete.

index.html
```diff
@@ -1978,12 +1978,14 @@ <h3>Terms of Use</h3>
...
+for the <a>verifiable credential</a>. This could be used by the <a>verifier</a>
+to establish the confidence with which it relies on the claims in the
+<a>verifiable credential</a>. For example, an issuer might check physical 
+documentation provided by the <a>subject</a> or might perform a set of background
+checks before issuing the credential. In certain scenarios, this information is
+useful to the <a>verifier</a> when determining the risk associated with relying
+on a given credential.
```
msporny: The language is an improvement. I'd like us to go further and mention Level of Assurance processes and links to physical documentation in the evidence field. The description we have at present is vague and we should provide more examples of the types of things you can put in the evidence field. I'm happy to provide concrete extra examples (downloadable links, LoA)... just thinking out loud, but hopefully this PR can add the language and I'll add the examples if we can get this PR in.

msporny:
> This language is better, but I still do not believe an 'evidence' section makes any sense.
> If a verifier trusts an issuer, they will have faith in the attributes signed by that issuer.

Not always. We have a number of use cases where different government agencies across jurisdictions want to see the same paperwork that was submitted when the credential was issued. We also have other regulatory requirements wrt. Level of Assurance as the same credential can be issued with different levels of assurance, and the verifier wants to know with what level of assurance a particular credential was issued.

> If a verifier does not trust an issuer, why would they trust the 'evidence' section?

It doesn't have to do with trust. You can trust an issuer, but still not know what process they used to issue a credential. There are thousands of banks all over the world that do KYC and there are different levels of assurance that they apply to different types of accounts. A name credential for a personal checking account doesn't go through the same level of vetting as a name credential for a business checking account that has millions in it.

> Putting 'evidence' in a credential seems nonsensical to me and I've yet to encounter a use case that convinces me it is necessary.

Do you find the two use cases above convincing? (Our customers need `evidence` to ensure thta they do their part in regulatory compliance).

index.html
David-Chadwick: @brentzundel Evidence is not there to engender trust in the issuer. Trust in the issuer is a necessary pre-cursor to relying on the Evidence. Evidence is there so that the verifier can determine the level of trust to place in the attributes of the subject. The more evidence the subject provided to the issuer, the higher the trust the verifier can have in the subject's claims (obviously up to the maximum trust that the verifier has in the issuer). So the trust in a subject is always LE the trust in the issuer.

stonematt: +1 @David-Chadwick 

@brentzundel in the professional licensing and credentialing space, the hiring body (hospital, for example) will look at the body of work the subject/holder has done as they exercise judgement around fitness of the applicant.  Having an option to put additional information in a credential from the issuer solves the challenge that holders lie and commit fraud in the job hunt all the time.  Evidence allows the verifier to see a more complete picture of the applicant, it has a clear chain of custody and trust.  Selective disclosure is not always the objective of the ecosystem.  We also want a trustworthy and standard way to share details that add nuance to the credential.

index.html
stonematt: +1 see updates.

index.html
stonematt: see next commit.

David-Chadwick: on which it issued

index.html
```diff
@@ -1978,12 +1978,14 @@ <h3>Terms of Use</h3>
...
 
         <p>
-Evidence is used by an <a>issuer</a> to determine whether or not to issue
-a <a>credential</a>. For example, an issuer might check physical documentation
-provided by the <a>subject</a> or might perform a set of background checks
-before issuing the credential. In certain scenarios, this information is useful
-to the <a>verifier</a> when determining the risk associated with relying on a
-given credential.
+Evidence is used by an <a>issuer</a> to inform the <a>verifier</a> of the basis
+for the <a>verifiable credential</a>. This could be used by the <a>verifier</a>
```
David-Chadwick: the verifiable......

stonematt:
@msporny I think the new language on evidence is ready for you to review again.

index.html
David-Chadwick: Evidence is used by an <a>issuer</a> to inform the <a>verifier</a> of the basis
for issuing the <a>verifiable credential</a>. This could be used by the <a>verifier</a>

brentzundel:
@jandrieu the I9 seems to me more like a presentation than a credential. 

@David-Chadwick I don't see the claims in an issued credential as coming from the subject, but rather from the issuer. Perhaps I've misunderstood you. 

@msporny, you mention a verifier needing to know the the process an issuer has used to issue a credential. This is intriguing to me, but it is not clear precisely how the evidence section provides that. 

@stonematt I don't see how your example of a hiring body needing evidence from a prover requires an evidence section, more than it needs a presentation of multiple credentials. Selective disclosure has nothing to do with it. 

From the breadth of responses to my confusion it feels like everyone is interpreting the evidence section differently. 
If each implementation is uses evidence differently, is that okay?

stonematt:
@brentzundel I agree, a community could decide to use multiple credentials. That may be a perfectly reasonable solution for sharing a collection of activities or work product about a subject.  

Evidence is a particular kind of attribute of a credential that meets a particular need.  In professional credentialing, the challenge for a licensing body is expressing the value of the data that a hierarchy addresses. The License is the credential that enables someone to get a job -- it is required to legally practice a craft (think MD for Doctor, Broker for Realtors, etc). Whether the professional took exam A or course B may inform an areas of specialty to the Verifier, but they May NOT be used as discreet credentials in the marketplace.  The issuer would share those details as elements supporting the license, but never as stand alone facts.

Evidence, as an attribute of the Credential, provides insight to the verifier into the issuers decision making process, rather than giving an unordered bag of data to the verifier and expecting them to construct a proper model and hoping they reason over it appropriately.

Finally, because we expect VC to serve a universe of credentials, and evidence is simply "supporting data" about a credential, I'm not at all surprised or concerned that each community/implementation has a different use of the concept.

I will volunteer to look at the example in this section and add an evidence blob that's more along these lines - in addition to the document verification blob.

index.html
```diff
@@ -2040,6 +2044,13 @@ <h3>Evidence</h3>
     "evidenceDocument": "DriversLicense",
     "subjectPresence": "Physical",
     "documentPresence": "Physical"
+  },{
+    "id": "https://example.edu/evidence/f2aeec97-fc0d-42bf-8ca7-0548192dxyzab",
```
msporny: Hmm, don't quite understand this. I'm going to merge anyway and try to rework the examples by putting in a "LoA" example, and a physical document example.

PR #420, opened 2019-02-05
Lifecycle Additions

gannan08:
Continuation of https://github.com/w3c/vc-data-model/pull/313

***
<a href="https://pr-preview.s3.amazonaws.com/digitalbazaar/vc-data-model/pull/420.html" title="Last updated on Feb 11, 2019, 5:03 PM UTC (abffc0b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/420/95d7a68...digitalbazaar:abffc0b.html" title="Last updated on Feb 11, 2019, 5:03 PM UTC (abffc0b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -629,7 +629,95 @@ that draws on multiple <a>credentials</a> about different <a>subjects</a>
 that are often, but not required to be, related.
         </p>
 
-      </section>
+    </section>
+    <section>
+      <h3>Lifecycle</h3>
+
+      <em>This section is non-normative.</em>
+
+      <p>
+        After outlining the core data model concepts, it is now possible to
+        reach the culmination of this section, the Lifecycle. Claims allow one
+        to make statements about a subject. Credentials provide the composition
+        mechanism for tamper-evident claims that provide cryptographic
+        guarantees on the provenance of those claims. Presentations enable the
+        ability for holders to use one or more credentials to satisfy requests
+        from a verifier and proof of control of those credentials.
+      </p>
+
+      <p>
+        The lifecycle of credentials will take the same general path: issuance
+        of claims to a credential subject and then the presentation to a
+        verifer. In <a href="#ex-1-example-verifiable-credential">Example 1</a> we can envision Pat receiving an
+        "AlumniCredential" from "Example University".
+      </p>
+
+      <pre class="example nohighlight" title="Example Verifiable Credential">
+{
+  "@context": [
+    "https://w3.org/2018/credentials/v1",
+    "https://example.com/examples/v1"
+  ],
+  "id": "http://example.edu/credentials/1872",
+  "type": ["VerifiableCredential", "AlumniCredential"],
+  "issuer": "https://example.edu/issuers/565049",,
+  "issuanceDate": "2010-01-01T19:73:24Z",
+  "credentialSubject": {
+    "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
+    "alumniOf": "Example University"
+  },
+  "proof": {
+    "type": "RsaSignature2018",
+    "created": "2017-06-18T21:19:10Z",
+    "creator": "https://example.com/jdoe/keys/1",
+    "nonce": "c0ae1c8e-c7e7-469f-b252-86e6a0e7387e",
+    "signatureValue": "BavEll0/I1zpYw8XNi1bgVg/sCneO4Jugez8RwDg/+
+      MCRVpjOboDoe4SxxKjkCOvKiCHGDvc4krqi6Z1n0UfqzxGfmatCuFibcC1wps
+      PRdW+gGsutPTLzvueMWmFhwYmfIFpbBu95t501+rSLHIEuujM/+PXr9Cky6Ed
+      +W3JT24="
+  }
+}
+      </pre>
+
+      <p>
+        In the next chapter of Pat's "AlumniCredential" lifecycle, is
+        presenting the credential for a discount. The verifier states that any
+        alumni of "Example University" will receive the best discount in the
+        world, 100% of. Pat will generate a Verifiable Presentation to achieve
+        this discount, this can be found in <a href="#ex-2-example-verifiable-presentation">Example 2</a>.
+      </p>
+      <pre class="example nohighlight" title="Example Verifiable Presentation">
+{
+  "@context": [
+    "https://w3.org/2018/credentials/v1",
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -629,7 +629,101 @@ <h3>Presentations</h3>
...
 
-      </section>
+    </section>
+    <section>
+      <h3>Lifecycle</h3>
+
+      <em>This section is non-normative.</em>
+
+      <p>
+This section seeks to illuminate the lifecycle of a credential. Credentials
```
brentzundel: I'd recommend changing this to "This section illustrates the lifecycle of a credential."

index.html
```diff
@@ -629,7 +629,101 @@ <h3>Presentations</h3>
...
+proof of control of those credentials. Credentials, when in motion, form a 
+general flow. This flow provides the bedrock for the lifecycle of a 
+credential, it highlights the issuance of credentials to their verification.
+      </p>
+
+      <p>
+The lifecycle of credentials will take a common path: issuance
+of claims to a credential subject, storage of credentials into credential
+repository engineered to store verifiable credentials, and then the
+presentation to a verifer. In <a href="#ex-1-example-verifiable-credential">
```
brentzundel: I would like this better if there was mention of composing credentials into a presentation, because credentials are not always simply presented.
Perhaps: "and then composition of credentials into a presentation to a verifier."

index.html
```diff
@@ -629,7 +629,101 @@ <h3>Presentations</h3>
...
+      +W3JT24="
+  }
+}
+      </pre>
+
+      <p>
+In the final chapter of this lifecycle, Pat's "AlumniCredential" will be 
+presented for a discount. The verifier, a ticket sales counter, states that any
+alumni of "Example University" will receive a 72% discount on football season 
+tickets. Pat uses thier mobile device to select the Verifiable Credential
```
brentzundel: "Pat uses his mobile device . . ."

index.html
```diff
@@ -629,7 +629,101 @@ <h3>Presentations</h3>
...
+  }
+}
+      </pre>
+
+      <p>
+In the final chapter of this lifecycle, Pat's "AlumniCredential" will be 
+presented for a discount. The verifier, a ticket sales counter, states that any
+alumni of "Example University" will receive a 72% discount on football season 
+tickets. Pat uses thier mobile device to select the Verifiable Credential 
+received earlier to send to the verifier in the form of a Verifiable
```
brentzundel: "received earlier to compose into a Verifiable Presentation to send to the verifier in order to receive the discount."

index.html
```diff
@@ -629,7 +629,101 @@ <h3>Presentations</h3>
...
+      "creator": "https://example.com/jdoe/keys/1",
+      "nonce": "c0ae1c8e-c7e7-469f-b252-86e6a0e7387e",
+      "signatureValue": "BavEll0/I1zpYw8XNi1bgVg/sCneO4Jugez8RwDg/+
+      MCRVpjOboDoe4SxxKjkCOvKiCHGDvc4krqi6Z1n0UfqzxGfmatCuFibcC1wps
+      PRdW+gGsutPTLzvueMWmFhwYmfIFpbBu95t501+rSLHIEuujM/+PXr9Cky6Ed
+      +W3JT24="
+    }
+  }],
+  "proof": { ... }
+}
```
brentzundel: As this is the only (is there another one?) example of a complete presentation in the data model, could we also have an example of a presentation done with selective disclosure?

index.html
```diff
@@ -629,7 +629,101 @@ <h3>Presentations</h3>
...
+    </section>
+    <section>
+      <h3>Lifecycle</h3>
+
+      <em>This section is non-normative.</em>
+
+      <p>
+This section seeks to illuminate the lifecycle of a credential. Credentials 
+allow one to make statements about a subject. Moreso, it provides a succint 
+composition mechanism for tamper-evident claims that provide cryptographic
```
msporny: Overly flowery language, no need to embellish. This section is also re-hashing concepts introduced elsewhere in the document, I think? Unless, this is a leading section, can't tell from the line numbers.

index.html
```diff
@@ -629,7 +629,101 @@ <h3>Presentations</h3>
...
+guarantees on the provenance of those claims. Presentations enable subjects to 
+use one or more credentials to satisfy requests from a verifier and assert 
+proof of control of those credentials. Credentials, when in motion, form a 
+general flow. This flow provides the bedrock for the lifecycle of a 
+credential, it highlights the issuance of credentials to their verification.
+      </p>
+
+      <p>
+The lifecycle of credentials will take a common path: issuance
+of claims to a credential subject, storage of credentials into credential
```
msporny: Simplify to: "issuance of one or more credentials, storage of credentials, and then the composition and presentation of credentials to a verifier.

index.html
```diff
@@ -629,7 +629,101 @@ <h3>Presentations</h3>
...
+repository engineered to store verifiable credentials, and then the
+presentation to a verifer. In <a href="#ex-1-example-verifiable-credential">
+Example 1</a> Pat receives an "AlumniCredential" from "Example University" that 
+will be stored in Pat's digital wallet on his mobile device, a type of
```
msporny: The reason we use "Pat" as a name is to avoid picking a gender for the actor. You've made Pat a man in the example :), please undo that.

index.html
msporny: "thier" typo

msporny: Pat is neither a man or a woman... Pat is without gender on purpose. :)

index.html
msporny: I'm concerned that while the language is accurate, it's a bit difficult to track.

index.html
msporny: We may want to compare/contrast between a credential that has certain attributes selectively disclosed vs. one that has not. Since these examples take up a lot of space, I suggest putting it in the appendix like we did for the JWT / LDP compare-contrast.

gannan08: @brentzundel I see no reason why we couldn't have another example. My thinking is to get this PR in and then make another one that would add to this section.

@msporny Are you suggesting to add a selective disclosure example to this section and then complete the compare-contrast in the appendix? If so, I agree with that line of thinking.

brentzundel: Sounds good to me.

index.html
ken-ebert: "thier" should be "their"

PR #445, opened 2019-03-11
Verification section updates...

grantnoble:
...to clarify holder vs subject. For the most part I believe this section is ok, however I'm suggesting some changes to tighten up what we are talking about. Specifically, verifying verifiable presentations (which contain verifiable credentials) and in a few instances, clarify that it's the holder making the presentation on behalf of a subject.

***
<a href="https://pr-preview.s3.amazonaws.com/grantnoble/vc-data-model/pull/445.html" title="Last updated on Mar 11, 2019, 11:31 PM UTC (eba78bf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/445/7edea6b...grantnoble:eba78bf.html" title="Last updated on Mar 11, 2019, 11:31 PM UTC (eba78bf)">Diff</a>


index.html
```diff
@@ -3178,30 +3180,33 @@ <h3>Issuer</h3>
...
         <p>
-The value associated with the <code>id</code> <a>property</a> for each
-<code>credentialSubject</code> MUST identify a <a>subject</a> to the
-<a>verifier</a>. For example, if a <a>subject</a> is identified and the
-<a>verifier</a> has public key metadata used for authentication purposes that
-is related to the <a>subject</a>, the <a>verifier</a> MAY be able to
-authenticate the <a>subject</a> using a signature generated by the
-<a>subject</a> contained in the <a>verifiable presentation</a>.
+In the <a>verifiable credentials</a> <a>presented</a> by a <a>holder</a> on
+behalf of a <a>subject</a>, the value associated with the <code>id</code>
```
David-Chadwick: This is wrong on two counts:
a) a VC can contain multiple credentialSubjects, not just one
b) the VCs may not be presented on behalf of the subject, but may be presented on behalf of the holder him/herself

index.html
```diff
@@ -3293,22 +3298,22 @@ <h3>Status</h3>
 
         <p>
 If the <code>credentialStatus</code> property is available, the status of the
-claim MUST be in good standing according to the <code>credentialStatus</code>
-<a>type</a> definition for the <a>credential</a>. For example, a <a>verifier</a>
-can ensure that the status has not been withdrawn for cause by the
-<a>issuer</a>.
+a <a>verifiable credential</a> MUST be in good standing according to the
+<code>credentialStatus</code> <a>type</a> definition for the
+<a>verifiable credential</a>. For example, a <a>verifier</a> can ensure that the
```
brentzundel: Whether a 'revoked' or 'expired' credential is still valid is primarily determined by the verifier's use case, which may not follow the 'good standing' according to the credentialStatus. Having MUST here is inappropriate.
I recognize this isn't something you've introduced, but it is an error in the verification section that needs fixing.

index.html
grantnoble: Removed "on behalf of the subject".
I believe that the word "each" in "...the value associated with the id property for each credentialSubject MUST identify a subject..." covers the case where a VC contains multiple credentialSubjects.

index.html
grantnoble: Changed the MUST to a MAY and that the VC be evaluated according to the credentialStatus type definition AND the verifier's own status evauation criteria.

PR #454, opened 2019-03-17
tweak i18n considerations

chaals:
Use the more relevant direction example string. Note that raw UTF-8 text often isn't sufficient for proper text presentation.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/454.html" title="Last updated on Mar 19, 2019, 2:41 PM UTC (81e49c3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/454/b945f06...81e49c3.html" title="Last updated on Mar 19, 2019, 2:41 PM UTC (81e49c3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4230,7 +4230,8 @@ There are a number of internationalization considerations implementers should be
 aware of when publishing data described in this specification. As with any web
 standards or protocols implementation, ignoring internationalization makes
 it difficult for data to be produced and consumed across a disparate set
-of languages and societies.
+of languages and societies, which would limiting the applicability of the specification
+and significantly diminish its value as a standard.
     </p>
 
     <p>
@@ -4242,11 +4243,9 @@ account when utilizing this data model.
       <h3>Language Declarations</h3>
 
       <p>
-When expressing human-readable text content in the data model, it is often
-useful to specify the language of the text to software agents that are
-responsible for rendering the information to an individual. The expressibility
-of language information is often a function of the syntax utilized. Options
-for encoding language information in are described below:
+When expressing human-readable text content in the data model, it is important
+for presentation, accessibility, and further processing to specify the language
+of the text.
       </p>
 
       <ul>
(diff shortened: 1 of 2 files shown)


chaals:
working on updating this now.

index.html
```diff
@@ -4230,7 +4230,8 @@ <h2>Internationalization Considerations</h2>
 aware of when publishing data described in this specification. As with any web
 standards or protocols implementation, ignoring internationalization makes
 it difficult for data to be produced and consumed across a disparate set
-of languages and societies.
+of languages and societies, which would limiting the applicability of the specification
```
r12a: limiting -> limit

index.html
```diff
@@ -4268,50 +4267,53 @@ <h3>Language Declarations</h3>
...
+    </section>
+    
+    <section class="informative">
+      <h3>Ruby Text</h3>
+  <p>
+In some situations it is common to use
+<a href="https://en.wikipedia.org/wiki/Ruby_character">Ruby characters</a>)
+to ensure that text information is clear and as useful as possible.
+JSON-LD's support for HTML data allows text to be encoded with ruby, supporting
+these cases.
```
r12a: Actually, my understanding is that what would be represented by ruby in running text is often represented in databases of string (such as people's names, addresses, etc.) by a separate pronunciation field within a record, rather than by using ruby markup.  I don't have any hard evidence of this or the extent to which ruby is or isn't used, but i did hear it from a Japanese person recently in the context of web publishing metadata.

r12a:
Again, i see these edits as editorial in nature.  On the assumption that we could revisit them if needed at a later date, i'm happy with the changes made, but would prefer it if you could consider the comments i just made, 10 mins ago.

burnburn:
Thank you Richard.  I would accept revisiting them at a later date.  Meanwhile, I believe @chaals is considering your most recent comments.

index.html
chaals: fixed.

index.html
chaals: Yes, I know of other examples where this is done - and I don't know how widespread it is either although I presume it is much more common that using HTML as the format to manage that kind of information.
I'll think about it, and see what the group and editor's think, but as I believe the change would be non-substantive, I will look at it in a week or so...

Issue #458, opened 2019-03-18
Refreshing questions


ken-ebert:
In section 5.5 Refreshing, there are two topics I have trouble understanding:

1.      "The issuer can include the refresh service as an element inside the verifiable credential if it is intended for either the verifier or the holder (or both), or inside the verifiable presentation if it is intended for the holder only."

It is my understanding that the holder is the one who generates a verifiable presentation, therefore an issuer would not be able to include a refresh service inside a verifiable presentation, only the holder would be able to do so.

2.      "Including the refresh reference inside the verifiable presentation enables the holder to refresh the credential details before creating a presentation to share with a verifier, while including the refresh reference inside the verifiable credential enables both the holder and the verifier to perform future updates of the credential."

Since the issuer is the one who creates and signs the credential, is it more correct to say "enables the holder to request that the issuer refresh the credential details..."

If a verifier requests an update to the credential, how does the verifier receive the updated credential, since issuers only know the identity of the holder and don't know the identity of a verifier. Or how is the holder notified to produce a new presentation based on the updated credential requested by a verifier?

David-Chadwick:
The issuer may generate a VP in order to send the VC to the holder, so would include refresh info in this VP that is specifically for the holder. Better wording might be

 The issuer can include the refresh service as an element inside the verifiable credential if it is intended for either the verifier or the holder (or both), or inside the verifiable presentation if it is intended for the holder only. In the latter case, this enables the holder to refresh the verifiable credential before creating a verifiable presentation to share with a verifier. In the former case, including the refresh reference inside the verifiable credential enables both the holder and the verifier to perform future updates of the credential.

David-Chadwick:
This has now been addressed in PR#501

David-Chadwick:
Revised text is now in PR#544

msporny:
PR #544 has been merged. Are there further non-normative changes you would like to make to the specification text @ken-ebert?

msporny:
Ping @ken-ebert -- are you happy with the changes made in PR #544 as a result of the issue you raised?

stonematt:
From VCWG call on April 30, 2019: 
RESOLUTION: Issue #458 is addressed by PR #544, which makes non-normative changes noting when it is and is not appropriate to use the refresh service. The PR has been approved by multiple parties and has been merged. Issue #458 should be closed.

PR #501, opened 2019-03-27
Clarification to Refresh Service to address #458

David-Chadwick:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/501.html" title="Last updated on Mar 27, 2019, 5:26 PM UTC (295c545)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/501/1ad7698...295c545.html" title="Last updated on Mar 27, 2019, 5:26 PM UTC (295c545)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2173,7 +2173,7 @@ The <a>issuer</a> can include the refresh service as an element inside the
 for the <a>holder</a> only. In the latter case, this enables the <a>holder</a> to refresh
 the <a>verifiable credential</a> before creating a <a>verifiable presentation</a> to share 
 with a <a>verifier</a>. In the former case, including the refresh service inside the
-<a>verifiable credential</a> enables both the <a>holder</a> and the <>verifier</a> to
+<a>verifiable credential</a> enables both the <a>holder</a> and the <a>verifier</a> to
 perform future updates of the credential.
         </p>
         <p>
--- a/index.html
+++ b/index.html
@@ -2167,18 +2167,14 @@ It is useful for systems to enable the manual or automatic refresh of a
 specification enables an <a>issuer</a> to include a link to a refresh service.
         </p>
         <p>
-The <a>issuer</a> can include the refresh service as an element inside the
-<a>verifiable credential</a> if it is intended for either the <a>verifier</a>
-or the <a>holder</a> (or both), or inside the <a>verifiable presentation</a>
-if it is intended for the <a>holder</a> only.
-        </p>
-        <p>
-Including the refresh reference inside the <a>verifiable presentation</a>
-enables the <a>holder</a> to refresh the <a>credential</a> details before
-creating a <a>presentation</a> to share with a <a>verifier</a>, while
-including the refresh reference inside the <a>verifiable credential</a>
-enables both the <a>holder</a> and the <a>verifier</a> to perform future
-updates of the <a>credential</a>.
+The <a>issuer</a> can include the refresh service as an element inside the 
+<a>verifiable credential</a> if it is intended for either the <a>verifier</a> or the
+<a>holder</a> (or both), or inside the <a>verifiable presentation</a> if it is intended
+for the <a>holder</a> only. In the latter case, this enables the <a>holder</a> to refresh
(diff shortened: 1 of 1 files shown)


ken-ebert:
Thank you for clarifying the language. It makes it more apparent what is going on. However this clarity raises several concerns that were not obvious to me before:

"The <a>issuer</a> can include the refresh service as an element inside the 
<a>verifiable credential</a> if it is intended for either the <a>verifier</a> or the
<a>holder</a> (or both), or inside the <a>verifiable presentation</a> if it is intended
for the <a>holder</a> only."

In the case of an issuer including the refresh service inside a credential to enable a holder to refresh the credential, I can understand the use case and possible ways this could be securely implemented.

In the case of an issuer including the refresh service inside a credential to enable a verifier to refresh the credential, I believe this has extremely negative consequences:
1. It takes the holder out of the transfer of the refreshed credential between the issuer and the verifier. This removes the holder's ability to control their data. When and for how long can the verifier refresh the credential? How does the holder revoke this capability?
2. Cryptographically how does the holder sign the presentation for the new credential, since the verifier is directly accessing the refresh service of the issuer to refresh the credential?
3. How does an issuer know that the holder has authorized the refresh of the credential?

In the case of 
"or inside the <a>verifiable presentation</a> if it is intended for the <a>holder</a> only."
I don't believe that an issuer can place anything inside a verifiable presentation. Only holders can place data in a presentation.
I also don't see any value to a refresh service in a presentation for a holder's use. The holder is the creator of presentations. Once created, presentations are passed to verifiers. If the presentation is unacceptable to the verifier, the verifier could request a new presentation and request that a holder refresh the credential before submitting a new presentation.

I have serious reservations about a refresh service offered to a verifier that bypasses the holder from both a cryptographic assurance standpoint and from the basic model of issuer/holder and holder/verifier pairwise relationships.

It is possible that I am still not understanding the refresh service correctly.

David-Chadwick:
Hi Ken

First let me say that I disagreed with adding the Refresh Service to the standard. I don't think it is needed and as you point out an issuer is never going to send a VC to a verifier without prior authorisation. And if prior authorisation has been given, then the verifier would presumably already know where to get the VC so not need to be told inside the VC. I will let Manu argue for the value of this property, as I am not going to defend it.

However, with respect to VPs, these can be sent from the Issuer to the Holder when the VC is first issued. So in this case the refresh service can be in the VP. It is not only the Holder that can create a VP. If the data model implies this then it needs correcting.

msporny:
> It is possible that I am still not understanding the refresh service correctly.

This PR has generated far more questions than it's closing and it would take me a very long time to get all three of us on the same page through the issue tracker. I suggest we setup a time to discuss that doesn't take WG call time to get some closure on this PR.

index.html
```diff
@@ -2167,18 +2167,14 @@ <h3>Refreshing</h3>
...
-including the refresh reference inside the <a>verifiable credential</a>
-enables both the <a>holder</a> and the <a>verifier</a> to perform future
-updates of the <a>credential</a>.
+The <a>issuer</a> can include the refresh service as an element inside the 
+<a>verifiable credential</a> if it is intended for either the <a>verifier</a> or the
+<a>holder</a> (or both), or inside the <a>verifiable presentation</a> if it is intended
+for the <a>holder</a> only. In the latter case, this enables the <a>holder</a> to refresh
+the <a>verifiable credential</a> before creating a <a>verifiable presentation</a> to share 
+with a <a>verifier</a>. In the former case, including the refresh service inside the
+<a>verifiable credential</a> enables both the <a>holder</a> and the <>verifier</a> to
```
msporny: Syntax error... `<>` should be `<a>`.

David-Chadwick: Done!

msporny:
We discussed this on a call today and are going to attempt to produce non-normative text to resolve the various concerns. There will be a new PR, which we will note below.

TallTed:
Closing a circle -- is #544 the new PR?

brentzundel:
> Closing a circle -- is #544 the new PR?

yes

Issue #463, opened 2019-03-19
Not every entity can perform a role in the ecosystem.


RieksJ:
The definition of "entity" says: "A thing with distinct and independent existence, such as a person, organization, concept, or device. An entity can perform one or more roles in the ecosystem." 

However, a concept, while being one of the examples of a thing, cannot perform any of the roles in the ecosystem. I suggest to leave the last sentence out and just stick with the definition.

talltree:
+1.

jandrieu:
The goal of that sentence is to make it clear that those entity performing a given role, such as an Issuer, may also perform the other roles in the system. So I'm hesitant to cut out the last sentence.

That said, I agree. The definition could be improved. Personally, thinking of concepts as entities makes no sense to me, as "independent existence" doesn't seem to be met for ideas, which by their nature are dependent on the sentient being thinking the concept, or at best, the representation of the idea in words or bits.  

This prompts me to re-invite you to suggest a work item for the CCG about terminology. I'm keen to have a more effective solution to approaching these questions. The bikeshedding approach is historically a mess.

RieksJ:
I prefer to distinguish between standards documents (that serve the purpose of different people doing or knowing stuff in the same way) and documents that serve educational/awareness purposes, that can be more elaborate. I hate having to go through educational texts when I'm just looking for what is commonly agreed upon. So I suggest to strike the last sentence, because this is a standard. If we need to educate people about VCs, we should do so in (a) separate document(s).

The 2nd paragraph of @jandrieu touches the very basics of semantics and knowledge, which some people refer to as 'philosophical', and which I am convinced we should understand really well if we want to do a good job on VCs. And that is the notion than things only exist because there is a sentient being that has knowledge of that existence. In other words: 'existence' is subjective. For example, I only exist because there are sentient beings (such as you) that have knowledge of my existence. I am not saying that you need to agree on this (it took me a while to accept it), but accepting this premise has helped us a lot in our thinking, in particular about the semantics of 'subject', 'identifier', 'subject-identifier' and the like. 

What's CCG?

jandrieu:
Ok. That's a great argument, @RieksJ, I withdraw my opposition.

[...]

PR #504, opened 2019-03-27
Update entity definition (Issue #463)

David-Chadwick:


terms.html
```diff
@@ -55,7 +55,7 @@
   <dt><dfn data-lt="entities|entity's">entity</dfn></dt>
   <dd>
 A thing with distinct and independent existence, such as a person,
-organization, concept, or device. An entity can perform one or more roles in
+organization, or device that performs one or more roles in
```
msporny: This change blows up a number of sentences in the specification. For example:

A <a>subject</a> is an <a>entity</a> about which <a>claims</a> can be made.

So, the change of the definition creates a cascading set of changes to the specification that are not going to be able to reconcile. :(

msporny: Okay, maybe I was freaking out prematurely. I'd rather not make this change, but it looks like we have consensus in the issue among multiple people that regularly disagree about terminology, so that's something I don't want to get in the way of. :)

Merging, I'll make other non-normative changes to align other spec text to this new definition.

msporny: Yeah, we used to say "An entity may perform"... and it was changed to "can" because people were concerned about the normative-ness of the statement, which subtly changed what the definition was saying. Anyway, merging. Editorial, I'll do the re-analysis of how the definition is used in the entire spec and make sure there are no normative changes as a result.

Issue #482, opened 2019-03-27
Introduction is misleading


nadalin:
The introduction is misleading as this talks about signing .. "he addition of technologies, such as digital signatures, makes verifiable credentials more tamper-evident and more trustworthy than their physical counterparts." this specification does not implement signatures.

brentzundel:
I believe this data model document does not specify which signatures must be used with verifiable credentials because the choice of signature type depends on the protocol within which verifiable credentials will be used.
Per the working group charter, this group may not specify protocol.
This does not mean that a place for signatures (the proof property) cannot be specified in the data model. The specification does not implement signatures, but it does specify where signatures may be used.

nadalin:
it does unless you have a signature that actually works with the data model chosen, a solution exists for JWT as there are JWS/E/K specifications.

brentzundel:
Is it your contention that unless a signature scheme is part of a published specification, that it cannot be considered as a means of verifying a credential?

nadalin:
I'm not set on a signature as a proof mechanism so I would say unless there is a agreed upon proof mechanism, which there is not today, in the case of a JWT/CWT a JWS or JWE could be used as a proof.

brentzundel:
I agree that in the case of JWT/CWT, a JWS or JWE could certainly be used as a proof. 
To my understanding, there is nothing in the data model that prohibits this.

msporny:
> The introduction is misleading as this talks about signing

The editors are having a hard time translating this into a concrete specification text change. Could you please provide some concrete text change that would address the issue you raised so that the WG has something concrete to discuss?

stonematt:
WG resolution: https://www.w3.org/2019/04/02-vcwg-minutes.html#resolution04
RESOLUTION: Add more clarifying text to the specification noting that while the specification does not standardize on signature format, the WG is aware of at least three mechanisms (JWTs, LD-Proofs, and ZKPs) that are capable of protecting the contents of the data model at the time of publication that are being used in deployments of the technology and expects those mechanisms to mature independently as well as new mechanisms to become standardized.

msporny:
Clarifying text has been merged in PR #531 to address the issue raised by @nadalin. The resolution from the WG has been applied. More than 7 days has passed since the resolution passed. Closing the issue per the WG's approved CR issue processing procedure.

PR #531, opened 2019-04-07
Clarify that multiple proof mechanisms are supported.

msporny:
Related to issue #482.

### DIFF
--- a/index.html
+++ b/index.html
@@ -568,12 +568,12 @@ publication, the Working Group included members that had implemented using
 at least three proof mechanisms. These implementations utilized
 JSON Web Tokens, Linked Data Signatures, and Camenisch-Lysyanskaya
 Zero-Knowledge Proofs. It is expected that each of these mechanisms will mature
-independently and additional mechanisms will also become standardized. The
-goal of this specification is to provide a data model that can be protected by a
-variety of current and future digital proof approaches. Conformance with
-respect to digital proofs in this specification is focused on the clear
-identification of each digital proof mechanism and not the details of any
-particular proof mechanism.
+independently and additional mechanisms will also become standardized. One
+of the goals of this specification is to provide a data model that can be
+protected by a variety of current and future digital proof approaches.
+Conformance with respect to digital proofs in this specification is focused
+on the clear identification of each digital proof mechanism and not the details
+of any particular proof mechanism.
         </p>
       </section>
 
--- a/index.html
+++ b/index.html
@@ -559,6 +559,22 @@ conformance of roles in the ecosystem, such as <a>issuers</a>, <a>holders</a>,
 or <a>verifiers</a>, because the conformance of ecosystem roles are highly
 application, use case, and market vertical specific.
         </p>
+
+        <p>
+While digital proofs, a subset of which are digital signatures, are required
+to ensure the protection of a <a>verifiable credential</a>, this specification
+does not standarize on any single digital signature format. At the time of
+publication, the Working Group included members that had implemented using
+at least three proof mechanisms. These implementations utilized
+JSON Web Tokens, Linked Data Signatures, and Camenisch-Lysyanskaya
+Zero-Knowledge Proofs. It is expected that each of these mechanisms will mature
+independently and additional mechanisms will also become standardized. The
+goal of this specification is to provide a data model that can be protected by a
+variety of current and future digital proof approaches. Conformance with
+respect to digital proofs in this specification is focused on the clear
+identification of each digital proof mechanism and not the details of any
+particular proof mechanism.
+        </p>
       </section>
 
     </section>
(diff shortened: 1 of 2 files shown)


msporny:
Tagging @jricher for a review of this PR.

index.html
```diff
@@ -559,6 +559,22 @@ <h3>Use Cases and Requirements</h3>
 or <a>verifiers</a>, because the conformance of ecosystem roles are highly
 application, use case, and market vertical specific.
         </p>
+
+        <p>
+While digital proofs, a subset of which are digital signatures, are required
+to ensure the protection of a <a>verifiable credential</a>, this specification
+does not standarize on any single digital signature format. At the time of
+publication, the Working Group included members that had implemented using
```
chaals: "...Working Group members had implemented verifiable credenitals using at least three..." ?

index.html
```diff
@@ -559,6 +559,22 @@ <h3>Use Cases and Requirements</h3>
...
+at least three proof mechanisms. These implementations utilized
```
chaals: "...proof Mechanisms: JSON Web Tokens [ref]; Linked Data Signatures [ref]..." ?

index.html
```diff
@@ -559,6 +559,22 @@ <h3>Use Cases and Requirements</h3>
...
+JSON Web Tokens, Linked Data Signatures, and Camenisch-Lysyanskaya
+Zero-Knowledge Proofs. It is expected that each of these mechanisms will mature
```
chaals: s/It is expected that each of these/The group expects these/ ?

index.html
```diff
@@ -559,6 +559,22 @@ <h3>Use Cases and Requirements</h3>
...
+independently and additional mechanisms will also become standardized. One
+of the goals of this specification is to provide a data model that can be
+protected by a variety of current and future digital proof approaches.
+Conformance with respect to digital proofs in this specification is focused
+on the clear identification of each digital proof mechanism and not the details
+of any particular proof mechanism.
```
chaals: "Conformance to this specification does not depend on the details of a particular proof mechanism; it requires clearly identifying the mechanism a verifiable credential uses."?

index.html
msporny: Fixed in 276007376a0082128439f109c32779c50f740656.

index.html
msporny: Fixed in 276007376a0082128439f109c32779c50f740656.

index.html
msporny: Fixed in 276007376a0082128439f109c32779c50f740656.

index.html
msporny: Fixed in 276007376a0082128439f109c32779c50f740656.

msporny:
Multiple independent reviews and approvals are in. Holding this merge on @jricher.

index.html
```diff
@@ -3025,7 +3041,7 @@ <h4>Relation to the Verifiable Credentials Data Model</h4>
 use of specific JWT-registered <a>claim</a> names and specific JWS-registered
 header parameter names to allow systems based on JWT to comply with this
 specification. If these specific <a>claim</a> names and header parameters are
-present, their respective counterpart in the standard 
+present, their respective counterpart in the standard
```
jricher: This looks like a spurious change

index.html
```diff
@@ -559,6 +559,22 @@ <h3>Use Cases and Requirements</h3>
 or <a>verifiers</a>, because the conformance of ecosystem roles are highly
 application, use case, and market vertical specific.
         </p>
+
+        <p>
+While digital proofs, a subset of which are digital signatures, are required
```
jricher: This could be worded to also explain that processing of the proof may be dependent on the syntax, such as using the JWS of a JWT for proofing a key holder. Some of the text of #539 tries to address this.

index.html
```diff
@@ -559,6 +559,22 @@ <h3>Use Cases and Requirements</h3>
...
+does not standarize on any single digital signature format. At the time of
+publication, Working Group members had implemented
+<a>verifiable credentials</a> using at least three proof mechanisms:
+JSON Web Tokens [[JWT]], Linked Data Signatures [[?LD-SIGNATURES]], and
+Camenisch-Lysyanskaya Zero-Knowledge Proofs [[?CL-SIGNATURES]]. The group
+expects some of these mechanisms, as well as new ones, to mature independently
+and become standardized in time. One of the goals of this specification is to
+provide a data model that can be protected by a variety of current and future
+digital proof mechanisms. Conformance to this specification does not
+depend on the details of a particular proof mechanism; it requires clearly
```
jricher: This makes it sound like the proof doesn't matter, where in practice the proofing mechanism for the credential is what makes it trustable within a real system. This important aspect of VC's might get lost by someone reading this paragraph. I would recommend adding text that having and validating a proof are an essential part of processing a VC but that the details of validating or applying the proof/signature are outside the scope of this specification.

index.html
msporny: Fixed in 56a7b6b382369b68f345ec8c854ed64c12b91f3f.

index.html
msporny: Fixed in 56a7b6b382369b68f345ec8c854ed64c12b91f3f.

index.html
msporny: Some of the folks that provide PRs don't have "strip trailing whitespace" turned on in their text editor. Rather than being super pedantic about it, I tend to just accept those PRs, knowing that my editor will eventually strip that trailing whitespace. It is unfortunate that doing so makes it look like there are spurious changes, but I have been unable to train everyone that produce PRs in proper trailing whitespace hygiene :).

msporny:
Reviewed by 3 independent parties, all suggestions applied to spec text, non-normative, merging.

nadalin:
I would agree with Justin's comments, what is he is not enough information

PR #533, opened 2019-04-08
added clarifying note to credentialSchema section

brentzundel:
This PR is intended to address issue #474 

### DIFF
--- a/index.html
+++ b/index.html
@@ -2092,6 +2092,13 @@ specific type definition.
           </dd>
         </dl>
 
+        <p class="note">
+The <code>credentialSchema</code> provides an opportunity to annotate type
+definitions or lock them to specific versions of the vocabulary. Authors of VCs
+can include a static version of their vocabulary via credentialSchema that is
+locked to some content integrity protection mechanism.
+        </p>
+
         <pre class="example nohighlight" title="Usage of the credentialSchema property to perform JSON schema validation">
 {
   "@context": [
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2092,6 +2092,13 @@ <h3>Data Schemas</h3>
           </dd>
         </dl>
 
+        <p class="note">
+The <code>credentialSchema</code> provides an opportunity to annotate type
+definitions or lock them to specific versions of the vocabulary. Authors of VCs
+can include a static version of their vocabulary via credentialSchema that is
+locked to some content integrity protection mechanism.
```
msporny: You may also want to make a point about doing syntactic checking on the credential as well using stuff like JSON Schema. Just reminding folks that this property has at least two uses.

msporny: In any case, approve of change whether or not you take the above into account.

brentzundel: Made some changes  in [cd62b44](https://github.com/w3c/vc-data-model/pull/533/commits/cd62b44daf6b6767d8eee86280c948096668de61)

msporny:
I forget where we got to here. I think we agreed that we can add lots of non-normative text to this section, and @David-Chadwick said he'd do it... but are we:

1. Merging this PR and waiting for an additional one from @David-Chadwick, or
2. Closing this PR and waiting for @David-Chadwick to write a PR?

David-Chadwick:
@msporny As I recall, you said you would write this text since you are more intimately knowledgeable about JSON-LD and its nuances. I said I would write the refresh text (which I did)

Issue #478, opened 2019-03-26
protocol statement in id section


brentzundel:
In [section 4.2](https://w3c.github.io/vc-data-model/CR/2019-03-26/#identifiers), there is a protocol statement that incorrectly uses normative language that cannot be enforced.

The text reads:
>  Identifiers that others are expected to use when expressing statements about a specific thing MUST be expressed using the id property.

**This does not match the pattern of all other normative statements in the spec, and therefore I argue that this is actually an editorial change.**

To follow the pattern of all other statements of this type, the text should read
> This specification defines the id property for identifiers. This allows for the expression of statements about specific things in the verifiable credential.

David-Chadwick:
Brent, whilst I don't object to your rewording, why is this a protocol statement? Surely this is a modelling statement. It is saying that if an object has an identifier that others use when referring to this object, then it must be the ID property.

Note that we have two different types of normative statements about properties:
1. Properties that may be present, but if present MUST have a specific format, and
2. Properties that MUST be present, and MUST have a specific format.

The ID property is rather unusual in that both of the above apply to it e.g. a VC MUST have an ID property, whereas a CredentialSubject may have an ID property. 

What is missing from the spec, is a statement that says "A verifiable credential MUST have an ID property". I think the sentence you have identified was a proxy for this missing sentence!

Another related point is that the NOTE implies that a VC can omit the ID property, and I do not think this is intentional. So the NOTE is dangerous as it implies that no IDs are mandatory (even ones that are stated to be).

In conclusion, I think this entire section needs some re-writing and clarification.

brentzundel:
1. According to the spec, an id property is entirely optional.
2. The normative statement here is describing how an id property must be **used**, which is protocol and impossible to test as a data model.

ken-ebert:
As I remember from previous discussions, the intent was that if the optional identifiers were present, then they MUST be contained in the "id" and not in some other attribute.

I think this language might capture that thought better:
"Identifiers that others are expected to use when expressing statements about a specific thing, if present, MUST be expressed using the id property."

I agree with Brent that it might be very difficult to test whether the issuer used another attribute for an identifier instead of "id", such as:
"myid" // relatively easy to spot
or
"my-noncompliant-identity" // relatively easy to spot
or
"qxz" // should this have been "id"?

Although I think this is very difficult to test, I don't think this is a protocol issue. I think it is a question related to the basic filling out of the data model.

brentzundel:
[...]

PR #443, opened 2019-03-07
Conformance language changes/additions.

msporny:
WIP - DO NOT MERGE YET.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/443.html" title="Last updated on Mar 14, 2019, 7:35 PM UTC (59ff3ba)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/443/441726a...59ff3ba.html" title="Last updated on Mar 14, 2019, 7:35 PM UTC (59ff3ba)">Diff</a>

PR #534, opened 2019-04-08
minor edit of text in Identifiers section.

brentzundel:
This PR is intended to address issue #478 

### DIFF
--- a/index.html
+++ b/index.html
@@ -1052,8 +1052,9 @@ others can express statements about the same thing. This specification defines
 the <code>id</code> <a>property</a> for identifiers. The <code>id</code>
 <a>property</a> is intended to unambiguously refer to an object such as a person,
 product, or organization. This allows for the expression of statements about
-specific things in the verifiable credential. Objects that others can refer to,
-MAY contain the <code>id</code> <a>property</a>.
+specific things in the verifiable credential. If present, identifiers that
+others are expected to use when expressing statements about a specific thing
+MUST be expressed using the <code>id</code> <a>property</a>.
         </p>
 
         <p class="note">
--- a/index.html
+++ b/index.html
@@ -2869,7 +2869,7 @@ CBOR, that is capable of expressing the data model.
         <h3>JSON</h3>
 
         <p>
-The data model as described in Section <a href="#core-data-model"></a> can be
+The data model, as described in Section <a href="#core-data-model"></a>, can be
 encoded in Javascript Object Notation (JSON) [[!RFC8259]] by mapping property
 values to JSON types as follows:
         </p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1037,9 +1037,9 @@ <h3>Identifiers</h3>
         <p>
 When expressing statements about a specific thing, such as a person, product,
 or organization, it is often useful to use some kind of identifier so that
-others can express statements about the same thing. Identifiers that others
-are expected to use when expressing statements about a specific thing MUST be
-expressed using the <code>id</code> <a>property</a>.
+others can express statements about the same thing. This specification defines
+the <code>id</code> <a>property</a> for identifiers. This allows for the
+expression of statements about specific things in the verifiable credential..
```
```suggestion
expression of statements about specific things in the verifiable credential.
```

index.html
```diff
@@ -1037,9 +1037,9 @@ <h3>Identifiers</h3>
         <p>
 When expressing statements about a specific thing, such as a person, product,
 or organization, it is often useful to use some kind of identifier so that
-others can express statements about the same thing. Identifiers that others
-are expected to use when expressing statements about a specific thing MUST be
-expressed using the <code>id</code> <a>property</a>.
+others can express statements about the same thing. This specification defines
+the <code>id</code> <a>property</a> for identifiers. This allows for the
```
msporny: This is a normative change as you removed the MUST language. Clearly, other identifiers can be used and associated with other properties. We should probably change it such that the normative requirement is "about a specific thing that has associated properties". We never intended to say you MUST use `id`... rather, if you think you're going to make statements about the same subject as someone else, you MUST use `id`.

What about this:

The <code>id</code> <a>property</a> MUST be used when statements about a particular subject are expected to be expressed by more than one entity.

brentzundel: My contention is that this "normative" statement is unenforceable in a data model spec and therefore a mistake in need of the suggested editorial change.

brentzundel: This is the one inappropriately "normative" statement left in the spec (the rest were edited out in PR #443, but this one was missed). 
This one should be changed for the same reason the rest of them were: we have no protocol to to enforce how the `id` property must be used, so the spec shouldn't say how it must be used.

David-Chadwick: We have been around this discussion several times before. I think we have agreed that there are two types of conformance statement about properties:

1.    Properties that MAY be present, but if present MUST have a specific format, and
2.    Properties that MUST be present, and MUST have a specific format.

The contentious issue appears to be about properties of type 1. How do we stop someone defining a property that is semantically the same as  a type 1 property but has a different name and different format? I thought that in Barcelona we agreed that we cannot stop people from doing this. In fact this issue also applies to properties of type 2 as well. For example, we cannot stop someone defining a DateOfIssuance property with a format that is not RFC 3999. 

So for the id property, all we can do is:

a. define what its semantics and name are
b. say it MAY be present in objects
c. say that it MUST be a URL.

This leads to the following much simplified text:

"The id property is defined to unambiguously refer to an object such as a person, product, or organization. This is so that others can express statements about the same thing. Objects that others can refer to, MAY contain the id property. "

brentzundel: please see changes in [4dcd022](https://github.com/w3c/vc-data-model/pull/534/commits/4dcd0221699298fa6e9ae1ab75888438468645e9)

msporny:
I think we have a test that already checks this normative requirement. Let me try to find it.

index.html
```diff
@@ -1037,9 +1037,12 @@ <h3>Identifiers</h3>
         <p>
 When expressing statements about a specific thing, such as a person, product,
 or organization, it is often useful to use some kind of identifier so that
-others can express statements about the same thing. Identifiers that others
```
ken-ebert: I think a simpler change that reflects the thinking of the group would be, "If present, identifiers that others ..."

brentzundel:
the goal of this PR was to address what seemed to be an untestable "normative" statement.

Upon discovering that there is a test for this normative statement, I have modified the PR so that it simply adds clarifying non-normative text to the section.

index.html
```diff
@@ -1048,9 +1048,13 @@ <h3>Identifiers</h3>
...
-are expected to use when expressing statements about a specific thing MUST be
-expressed using the <code>id</code> <a>property</a>.
+others can express statements about the same thing. This specification defines
+the <code>id</code> <a>property</a> for identifiers. The <code>id</code>
+<a>property</a> is intended to unambiguously refer to an object such as a person,
+product, or organization. This allows for the expression of statements about
+specific things in the verifiable credential. If present, identifiers that
+others are expected to use when expressing statements about a specific thing
+MUST be expressed using the <code>id</code> <a>property</a>.
```
TallTed: I do not think this phrasing -- 

```
If present, identifiers that 
others are expected to use when expressing statements about a specific thing 
MUST be expressed using the <code>id</code> <a>property</a>.
```

-- expresses what was intended.  Flipping the word order around without changing what the sentence is saying --

```
If identifiers (that others are expected to use when expressing statements 
about a specific thing) are present, they MUST be expressed using the 
<code>id</code> <a>property</a>.
```

I think the intended statement is about `id` (not about identifiers intended for reuse) and ought to include something about cardinality.  Something more like: 

```
If the <code>id</code> <a>property</a> is present, it MUST express an 
identifier that others are expected to use when expressing statements 
about specific thing identified by that identifier; the <code>id</code> 
<a>property</a> MUST NOT have more than one value; and the value of the 
<code>id</code> <a>property</a> MUST be a URI.
```

Issue #491, opened 2019-03-27
5.3.1 Semantic Interoperability


nadalin:
No need to process the @Context unless dealing with JSON-LD, should be made optional

nadalin:
JWT/CWT are deployed widely in the web, most every RP uses these, so I don't think this should be at risk, what makes this at risk is that these are not first class objects that is they are always within a JSON-LD container, so if that restriction is removed then I think you would see JWT/CWT be widely used instead of JSON-LD. Microsoft is using JWT/CWT

brentzundel:
It is my understanding that processing the **@context** is optional, even though the property is required to be present.

nadalin:
It says "It is RECOMMENDED that dereferencing the URIs results in a document containing machine-readable information about the context. " it also says that the @context value pair MUST be there, no reason for this other than to force JSON-LD IMHO

brentzundel:
There are comment in #483 that I believe are pertinent to this issue as well.

David-Chadwick:
@nadalin "The use of an IANA registry for the various JWT claims would serve the purpose that the @context does unless I'm missing something here". What you are missing is that VCs are built on an open model, meaning that anyone is free to extend VCs in any way that they want to – “permissionless innovation”, without the need for a centrally managed registry.

nadalin:
@David-Chadwick There is no requirement in the spec to be decentralized that I can find, there are already dependencies on DNS with JSON-LD and dependencies on registries in this specification so IANA is just fine since the algorithm identifiers are defined in IANA

dmitrizagidulin:
@nadalin Nobody is forcing JSON-LD processing on anyone. Yes, `@context` is required, for wider interop.

> JWT/CWT are deployed widely in the web

The use of `@context` on the web is at least one order of magnitude (if not more) greater than the use of JWT. It is literally used by billions of web pages, by _every_ major search engine, including by Microsoft's Bing.

The 'RECOMMENDED that dereferencing the URIs results in a document containing machine-readable information' is just developer courtesy and common sense, but it is still not forcing anything. Your implementation is still free to break interop and not have a derefernceable URI.

jandrieu:
@context is mandatory, as per the discussion and consensus at the Barcelona face-to-face meeting.

@nadalin, including the @context means interoperability for JWT and JSON-LD. It doesn't mean you must use JSON-LD, just that you are conformant to a spec that supports interoperability.

chaals:
[...]

Issue #483, opened 2019-03-27
Section 4.1 Context


nadalin:
There is no reason to have the @Context this should removed or at least made optional. @Context is not needed when suing JWT or even CWTs as claims

nadalin:
There is no reason to have the JSON-LD Context mandatory, this could signify that JSON-LD was being used and not JWT/CWT but not required when having a JWT/CWT Claim

brentzundel:
There is nothing (to my knowledge) in the data model that prohibits the use of JWT or CWT. 
The requirement of a **@context** property does not limit an implementer to JSON-LD, nor is there a requirement that any @context be parsed. The property must simply be present.
My understanding of why this is so, is to increase potential interoperability between JSON-LD and other JSON types. I believe that having the @context property present does not break a standard JSON parser, but not having the property could break a JSON-LD parser, and there are several implementers who are planning to use JSON-LD.
It is entirely possible that I have misrepresented the facts here due to a misunderstanding on my part. I look to @msporny or @dlongley  or others to correct me if I am mistaken.

David-Chadwick:
I think this issue should be linked to #474 and clarifying text added to the spec for both properties.

nadalin:
We have implemented w/o the use of context, there is absolutely no need for the @Context in most cases, I don't understand the requirement here, this should be optional nit mandatory, you don't need this if using JWT yet it is required.

chaals:
@nadalin:

> I don't understand the requirement here

That's OK - strictly speaking it is not necessary to understand.

That said, as @brentzundel said above. If you have the @context token  then JSON-LD processors (there are several) can do more useful reasoning over the credential, e.g. for managing them, or extensibility. Meanwhile, it has no more practical impact on simplistic tools that don't use it than the rest of the syntax requirements imposed by using JSON.

nadalin:
@chaals ok, so don't force to be mandatory @context as it's not always needed, as it complicates processing when really not needed in all cases

chaals:
Any time interoperability with those who are doing JSON-LD processing is needed - i.e. for a valid credential that you share outside a walled garden, `@context` is necessary. The same as any time interop with JSON processors is needed, getting the commas and braces right is necessary.

**I** don't understand how a token that you recognise but don't process complicates processing. This is a really common pattern to support interoperability.

chaals:
@nadalin 
> There is abosolltly no need for the @context when using a JWT

[...]

PR #535, opened 2019-04-08
added clarifying note to @context section

brentzundel:
This PR address some points of issues #483 and #491

### DIFF
--- a/index.html
+++ b/index.html
@@ -982,7 +982,13 @@ express context information. It is RECOMMENDED that dereferencing the
 the context.
           </dd>
         </dl>
-
+        <p class="note">
+Though this specification requires that a <code>@context</code> <a>property</a>
+be present, there is no requirement that the value of the <code>@context</code>
+property be processed. This is to support compatibility with certain JSON
+processing libraries such as those that may be used when the
+<a>verifiable credential</a> is encoded as a JWT.
+        </p>
         <pre class="example nohighlight" title="Usage of the @context property">
 {
   <span class="highlight">"@context": [
--- a/index.html
+++ b/index.html
@@ -985,13 +985,14 @@ the context.
         <p class="note">
 Though this specification requires that a <code>@context</code> <a>property</a>
 be present, there is no requirement that the value of the <code>@context</code>
-property be processed using JSON-LD. This is to support compatibility with
-certain JSON processing libraries such as those that may be used when the
+property be processed using JSON-LD. This is to support processing using plain
+JSON libraries such as those that may be used when the
 <a>verifiable credential</a> is encoded as a JWT.
-For JSON-LD aware processors to properly function, the <code>@context</code>
-needs to be processed. For JSON processors, it is sufficient to ensure that the
-order of the values in <code>@context</code> is what is expected for the
-particular application, but no JSON-LD processing of those values is required.
+JSON-LD aware processors can process the <code>@context</code> using full
+JSON-LD processing as expected. For other processors, the only processing
+necessary is to ensure that the order of the values in <code>@context</code>
+is what is expected for the particular application, but no JSON-LD processing
+of those values is required.
         </p>
         <pre class="example nohighlight" title="Usage of the @context property">
 {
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -982,7 +982,13 @@ <h3>Contexts</h3>
 the context.
           </dd>
         </dl>
-
+        <p class="note">
+Though this specification requires that a <code>@context</code> <a>property</a>
+be present, there is no requirement that the value of the <code>@context</code>
+property be processed. This is to support compatibility with certain JSON
```
msporny: Note quite correct.

For JSON-LD aware processors, you have to process the `@context`.

For JSON processors, you have to make sure the order of the values in `@context` is what you expect them to be for your particular application (but you don't have to use the values to do JSON-LD based processing). So, JSON-based processors have to look for something like this:

```
  "@context": ["https://www.w3.org/2018/credentials/v1", EXPECTED_CONTEXT_2, EXPECTED_CONTEXT_3]
```

msporny: I was (also) going to add some text to the Syntactic Sugar section of the spec to clarify this point (possibly with an example of what a JSON-based processor should do): https://w3c.github.io/vc-data-model/#syntactic-sugar

brentzundel: I've made some changes in [b6088de](https://github.com/w3c/vc-data-model/pull/535/commits/b6088dea301b11deba1b3ddac4434074e691a4b3)
(I'm not feeling qualified to make changes to the syntactic sugar section, but would welcome a PR to my PR)

David-Chadwick: Looks OK now

index.html
```diff
@@ -982,7 +982,17 @@ <h3>Contexts</h3>
 the context.
           </dd>
         </dl>
-
+        <p class="note">
+Though this specification requires that a <code>@context</code> <a>property</a>
+be present, there is no requirement that the value of the <code>@context</code>
+property be processed using JSON-LD. This is to support compatibility with
+certain JSON processing libraries such as those that may be used when the
```
jricher: I dislike how this is phrased, the use of "certain" sounds weasely, like some JSON libraries are insufficient but others are good. Truth is, any plain JSON library isn't going to care about JSON-LD, but it can do a syntax check for specific values in a specific field. Suggest backing this off a bit:

```suggestion
This is to support processing using plain JSON libraries such as those...```

File: index.html [hunk 3]
```diff
@@ -982,7 +982,17 @@ <h3>Contexts</h3>
...
+<a>verifiable credential</a> is encoded as a JWT.
+For JSON-LD aware processors to properly function, the <code>@context</code>
+needs to be processed. For JSON processors, it is sufficient to ensure that the
```
jricher: What worries me about this is that if I get sent a JSON-LD document every time, why wouldn't I just always do the (simpler) JSON processing?

dlongley: The only time you need to do any more than that simple processing is if you want to translate the data to use some other `@context` or into some other structure or format. It may be better to say "JSON-LD aware processors can process the `@context` using a full JSON-LD processor as expected. For other processors, the only processing necessary is to ensure that the...".

index.html
brentzundel: changes made in [665e076](https://github.com/w3c/vc-data-model/pull/535/commits/665e07665cfb3bdfa4c5b625257e49bc709109fa)

index.html
brentzundel: changes made in [665e076](https://github.com/w3c/vc-data-model/pull/535/commits/665e07665cfb3bdfa4c5b625257e49bc709109fa)

msporny:
Merging, non-normative, editorial. Multiple independent review approvals and the author has implemented all requested changes to the spec text.

index.html
msporny: > The only time you need to do any more than that simple processing is if you want to translate the data to use some other @context or into some other structure or format.

Also, if you're using LD Proofs / LD Signatures to generate a signature (or verify one), then you will have to do C14N, which will require you to do JSON-LD processing. So, if you live in JWT land and don't want to use JSON-LD, you don't have to (modulo the simple `@context` array string comparison). If you want what JSON-LD provides, or want LD Proofs/Signatures, then you can do extra processing to get those benefits.

Issue #541, opened 2019-04-09
Decoding rules for jti and sub are mixed up


awoie:
The current spec mixes up JWT decoding rules for `jti` and `sub`. `jti` must be added to the VC (top level) and `sub` to `credentialSubject`.

awoie:
PR was created: https://github.com/w3c/vc-data-model/pull/542

brentzundel:
I believe this is what you are saying, please correct me if I have misunderstood:

- `jti` is used as an identifier for the JWT, so it maps more appropriately to the `id` property of the credential itself.
- `sub` is used to identify the subject of the JWT, so it maps more appropriately to the `id` property of the `credentialSubject` object.
- Right now the spec has this backwards and your PR corrects this.

awoie:
Exactly

stonematt:
Resolved on VCWG call 4/16/2019.
RESOLVED: Issue #541 is addressed by PR #542, which makes a non-substantive change that fixes a bug in the specification related to JWT jti/sub values being swapped. This is a non-substantive change because there were two conflicting normative statements in the specification, the examples were clearly the intent, and any reasonable implementer would have implemented the feature correctly based on the existing JWT encoding rules in the specification. The PR has
9:58 AM been approved by multiple parties as a non-substantive change and should be merged. Issue #541 should be closed after the PR is merged.

Plan to close in 7 days

PR #542, opened 2019-04-09
fixed issue 541 - decoding rules for jti and sub are mixed up

awoie:
fixed issue #541 - decoding rules for jti and sub are mixed up

### DIFF
--- a/index.html
+++ b/index.html
@@ -3204,12 +3204,12 @@ of the <code>expirationDate</code> <a>property</a> of
 of the <code>issuanceDate</code> <a>property</a> of the new JSON object.
                   </li>
                   <li>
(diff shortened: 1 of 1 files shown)


msporny:
Resolved on VCWG call 4/16/2019.

RESOLVED: Issue #541 is addressed by PR #542, which makes a non-substantive change that fixes a bug in the specification related to JWT jti/sub values being swapped. This is a non-substantive change because there were two conflicting normative statements in the specification, the examples were clearly the intent, and any reasonable implementer would have implemented the feature correctly based on the existing JWT encoding rules in the specification. The PR has
9:58 AM been approved by multiple parties as a non-substantive change and should be merged. Issue #541 should be closed after the PR is merged.

PR #544, opened 2019-04-10
Update refresh service (Issue #458)

David-Chadwick:


David-Chadwick:
Talking to a doctor friend he has said that this feature (refresh in the VC) is going to be abused by issuers to violate a user's privacy. IdPs such as google are likely to put this property into every vc they issue so that verifiers will call them back, then they can track the user. Even though we give guidelines as to when it should and should not be used these are not normative and therefore can be ignored

dlongley:
@David-Chadwick,

> IdPs such as google are likely to put this property into every vc they issue so that verifiers will call them back, then they can track the user.

Abuses like this can't be stopped by a standard. Issuers can always just make non-standard properties that provide the same function. Standards are not law.

If we call out a standard place for refresh services, it may reduce the implementation burden for digital wallets that want to provide privacy guidance. It means that digital wallet implementers can write code to look at `refreshService` and give more specific guidance to users about potential privacy and tracking risks even when they don't understand the refresh service type. Without this, digital wallet implementers must either give only general advice "We don't have any idea what this property is; it could mean anything, be careful" and/or keep updating their implementations for every new type of non-standard refresh service property that becomes popular.

David-Chadwick:
@dlongley "Abuses like this can't be stopped by a standard.". Correct. But that does not mean that the standard should condone or support it, or even provide a mechanism for it. For the latter reason it should be removed.

Not understanding properties in a VC is something that all wallet and verifier software providers are going to have to live with, so this is not an argument for having a standard refresh service. All non-understood extensions are potentially dangerous and therefore a conservative implementation will reject all VCs with undefined (from its perspective) properties.

index.html
```diff
@@ -2194,23 +2194,26 @@ <h3>Data Schemas</h3>
...
-enables the <a>holder</a> to refresh the <a>credential</a> details before
-creating a <a>presentation</a> to share with a <a>verifier</a>, while
-including the refresh reference inside the <a>verifiable credential</a>
-enables both the <a>holder</a> and the <a>verifier</a> to perform future
-updates of the <a>credential</a>.
+The refresh service is only expected to be used when either the credential 
+has expired, or when the issuer does not publish credential status information. 
+It is ill advised for issuers to put the <code>refreshService</code> in a 
+<a>verifiable credential</a> that does not contain public information, 
+or whose refresh service is not protected in some way.
```
ken-ebert: I think that putting lines 2214-2216 and the following sentence in a warning Note would help implementers be more aware of the risks associated with refresh services: "Placing a refresh service in a verifiable credential in such a way that it is available to a verifier could remove control and consent from the holder and essentially allow a verifiable credential about the subject to be issued directly to the verifier, bypassing the holder."

David-Chadwick: I have had another thought. If implementors only support the Refresh Service for VPs and not for VCs, then this feature would be at risk and would be removed along with all other at risk features that do not have 2 implementations.

David-Chadwick: If we extend the implementation spreadsheet by having two columns for the refreshService (one for refreshService in VPs, the other for refreshService in VCs), instead of the current single column, then if 2 implementations do not support it in VCs, we could remove the feature from the final version of the standard, without needed to go through another CR now.

msporny: This is not the discussion we had during the call last week. You are suggesting something we covered, which is what @dlongley is stating above. Specifically not standardizing this makes it not possible to achieve the use case around updating public VCs provided through a digital wallet. It also doesn't prevent what you are concerned about, and makes it more likely that it'll be done w/o any way to detect it wrt. wallets warning their users that the VC may be correlatable. -1 to the suggested change. We'll have to discuss this more on another call.

David-Chadwick: @msporny I think you are confusing two things,
i) the text of this PR, which is exactly in line with the discussion we had, so can be merged, and
ii) my suggestion about altering the implementation spreadsheet, which does not effect the text of this PR, but is certainly a discussion we can have on another call.

msporny: Ah, ok, I thought you were arguing against your own PR.

PR #546, opened 2019-04-14
Clarify why processing @context is useful for JSON-based processors.

msporny:
Related to issue #491.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/546.html" title="Last updated on Apr 24, 2019, 2:40 AM UTC (0221fbb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/546/8fcd004...0221fbb.html" title="Last updated on Apr 24, 2019, 2:40 AM UTC (0221fbb)">Diff</a>

PR #548, opened 2019-04-14
Explain why @context provides interop

chaals:
An editorial alternative to proposal #546, related to issue #491

### DIFF
--- a/index.html
+++ b/index.html
@@ -2072,20 +2072,18 @@ with implementations, or enabling aggressive caching of contexts.
           <h4>Semantic Interoperability</h4>
 
           <p>
-This specification endeavors to enable both the JSON and JSON-LD syntaxes to
-be semantically compatible with one another without the JSON implementations
-needing to process the documents as JSON-LD. To achieve this, the
-specification imposes the following additional restrictions on both syntaxes:
+This specification ensures that "plain" JSON and JSON-LD syntaxes are semantically
+compatible without requiring JSON implementations to use a JSON-LD processor.
+To achieve this, the specification imposes the following additional restrictions,
+regarding the use of <code>@context</code>:
           </p>
 
           <ul>
             <li>
-JSON-based processors MUST process the <code>@context</code> <a>property</a>,
-ensuring the expected values exist in the expected order for the
-<a>credential</a> type being processed. It is advised that the expected order
-of values of the <code>@context</code> <a>property</a> should be defined by
-at least a human-readable extension specification and preferably by a
-machine-readable specification as well.
+JSON-based processors MUST process the <code>@context</code> <a>property</a>
+to the extent of ensuring the expected values exist in the expected order for
+the <a>credential</a> type being processed.
+              
             </li>
             <li>
 JSON-LD-based processors MUST produce an error when a JSON-LD context
@@ -2097,6 +2095,19 @@ Authors that are interested in this feature should read about the
 <code>@protected</code> feature in the JSON-LD 1.1 specification.
             </li>
           </ul>
+          <p>
+A human-readable document describing the expected order of values for the
+<code>@context</code> <a>property</a> SHOULD be available. A machine-readable
+description (i.e. a normal JSON-LD "context" document) SHOULD be available
+at the URL specified in the <code>@context</code> <a>property</a>.
+          </p>
+          <p>
+These requirements guarantee semantic interoperability between JSON and JSON-LD
+for terms defined by the <code>@context</code> mechanism. While JSON-LD processors
+will use the specific mechanism provided and can verify that all terms are correctly
+specified, JSON-based processors implicitly accept the same set of terms
+without testing that they are correct.
+          </p>
 
         </section>
       </section>


index.html
```diff
@@ -2097,6 +2095,19 @@ <h4>Semantic Interoperability</h4>
 <code>@protected</code> feature in the JSON-LD 1.1 specification.
             </li>
           </ul>
+          <p>
+A human-readable document describing the expected order of values for the
+<code>@context</code> <a>property</a> SHOULD be available. A machine-readable
+description (i.e. a normal JSON-LD "context" document) SHOULD be available
```
msporny: A JSON-LD Context isn't a SHOULD, it's a MUST, as described elsewhere in the specification (so we shouldn't restate it here).

index.html
```diff
@@ -2097,6 +2095,19 @@ <h4>Semantic Interoperability</h4>
 <code>@protected</code> feature in the JSON-LD 1.1 specification.
             </li>
           </ul>
+          <p>
+A human-readable document describing the expected order of values for the
+<code>@context</code> <a>property</a> SHOULD be available. A machine-readable
```
msporny: This is not testable, and thus should not be normative... thus the lowercase 'should' in the previous text.

index.html
```diff
@@ -2097,6 +2095,19 @@ <h4>Semantic Interoperability</h4>
...
+description (i.e. a normal JSON-LD "context" document) SHOULD be available
+at the URL specified in the <code>@context</code> <a>property</a>.
+          </p>
+          <p>
+These requirements guarantee semantic interoperability between JSON and JSON-LD
+for terms defined by the <code>@context</code> mechanism. While JSON-LD processors
```
msporny: I really like this last sentence and will be stealing it and integrating it into the PR I provided in an attempt to merge the best language in both PRs.

index.html
```diff
@@ -2072,20 +2072,18 @@ <h3>Extensibility</h3>
           <h4>Semantic Interoperability</h4>
 
           <p>
-This specification endeavors to enable both the JSON and JSON-LD syntaxes to
-be semantically compatible with one another without the JSON implementations
-needing to process the documents as JSON-LD. To achieve this, the
-specification imposes the following additional restrictions on both syntaxes:
+This specification ensures that "plain" JSON and JSON-LD syntaxes are semantically
```
msporny: I like this language better than the language in my PR, I'll move it over to my PR in an attempt to align the language.

index.html
```diff
@@ -2072,20 +2072,18 @@ <h3>Extensibility</h3>
...
 
           <ul>
             <li>
-JSON-based processors MUST process the <code>@context</code> <a>property</a>,
-ensuring the expected values exist in the expected order for the
-<a>credential</a> type being processed. It is advised that the expected order
-of values of the <code>@context</code> <a>property</a> should be defined by
-at least a human-readable extension specification and preferably by a
-machine-readable specification as well.
+JSON-based processors MUST process the <code>@context</code> <a>property</a>
```
msporny: Agree with the simplification here and moving the text to a paragraph below.

index.html
```diff
@@ -2072,20 +2072,18 @@ <h3>Extensibility</h3>
...
+to the extent of ensuring the expected values exist in the expected order for
+the <a>credential</a> type being processed.
+
```
David-Chadwick: Why is the order of values important?

index.html
David-Chadwick: I have just been designing the parser for our JSON implementation, and it seems to me that the only reason that I need to process the @context is to ensure that the terms (aliases that are used for URIs) in the body of the VC are pointed to by the @context values i.e. that no terms are present in the VC that are not pointed to by the @context values. Of course in our implementation a human will have read the contents of the @context web pages that are pointed to and will have configured the alias terms into our JSON parser so that it will be happy to encounter either the full URIs or the short form aliases. 

It might be useful to tell JSON implementors that they do not need to define new @context values themselves, since they can use full URIs for all their new VC terms if they want to.

index.html
```diff
@@ -2097,6 +2095,19 @@ <h4>Semantic Interoperability</h4>
...
+description (i.e. a normal JSON-LD "context" document) SHOULD be available
+at the URL specified in the <code>@context</code> <a>property</a>.
+          </p>
+          <p>
+These requirements guarantee semantic interoperability between JSON and JSON-LD
+for terms defined by the <code>@context</code> mechanism. While JSON-LD processors
+will use the specific mechanism provided and can verify that all terms are correctly
+specified, JSON-based processors implicitly accept the same set of terms
+without testing that they are correct.
+          </p>
```
David-Chadwick: If you expect JSON processors to check that the set of @context values are in the correct order, then why don't you expect JSON processors to check that all the terms in the VC are also defined by the @context? Only a human administrator can configure the JSON parser to check both of these. So why one without the other? Or have I misunderstood something?

index.html
msporny: Applied in https://github.com/w3c/vc-data-model/pull/546/commits/0221fbb2eb2833505848649e65980e38110d0f55.

index.html
msporny: I question whether it would be good to do this as this is not a common practice for HTTP APIs that use JSON (or JSON documents in general).

index.html
msporny: JSON-LD typically works in a last-defined-wins model. There is a feature in JSON-LD 1.1, called `@protected` that works (more or less) in a first defined wins mode. So, order is important in both cases and if we enforce order, you don't need to use full blown JSON-LD processing if you just want to do a JSON-based processor.

index.html
msporny: Fixed in https://github.com/w3c/vc-data-model/pull/546/commits/0221fbb2eb2833505848649e65980e38110d0f55.

index.html
msporny: Fixed in https://github.com/w3c/vc-data-model/pull/546/commits/0221fbb2eb2833505848649e65980e38110d0f55 ... but using your text.

index.html
msporny: Stolen in https://github.com/w3c/vc-data-model/pull/546/commits/0221fbb2eb2833505848649e65980e38110d0f55 :)

index.html
msporny: > then why don't you expect JSON processors to check that all the terms in the VC are also defined by the @context?

Frankly, because JSON developers have said that they don't want to do that. Experience w/ schema.org demonstrates that most will happily just copy-paste examples and work from those.

In an ideal world, everyone would just point a JSON-LD processor at the document and hit a button... but there is non-trivial subset of developers that don't want the "burden" of running a JSON-LD processor on their content and would rather risk semantic non-interop than pull an extra library into their software stack. In any case, we're trying to make that subset of JSON implementers happy... the ones that don't want to be bothered by semantics (but we still have to make sure that we provide enough tooling for them to interop when it becomes important to them.

PR #539, opened 2019-04-08
Clarify primacy of data model vs. syntax

jricher:

PR #550, opened 2019-04-15
Update conformance language

jricher:
This adds language to the definition of a conforming document. Even though this adds normative language, I fully believe this does not change the normative intent of the document as decided by the working group and therefore is not a substantive change as per W3C guidelines.

### DIFF
--- a/index.html
+++ b/index.html
@@ -562,7 +562,8 @@ that complies with the normative statements in this specification.
 Specifically, all relevant normative statements in Sections
 <a href="#basic-concepts"></a>, <a href="#advanced-concepts"></a>, and
 <a href="#syntaxes"></a> of this document MUST be enforced. A serialization
-format for the conforming document MUST be deterministic and bi-directional. 
+format for the conforming document MUST be deterministic, bi-directional, 
+and lossless as described in <a href="#syntaxes"></a>. 
 The conforming document MAY be transmitted or stored in any such 
 serialization format.
         </p>
--- a/index.html
+++ b/index.html
@@ -561,7 +561,10 @@ A <em>conforming document</em> is any concrete expression of the data model
 that complies with the normative statements in this specification.
 Specifically, all relevant normative statements in Sections
 <a href="#basic-concepts"></a>, <a href="#advanced-concepts"></a>, and
-<a href="#syntaxes"></a> of this document MUST be enforced.
+<a href="#syntaxes"></a> of this document MUST be enforced. A serialization
+format for the conforming doucment MUST be deterministic and bi-directional. 
+The conforming document MAY be transmitted or stored in any such 
+serialization format.
         </p>
 
         <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -544,7 +544,10 @@ <h3>Use Cases and Requirements</h3>
 that complies with the normative statements in this specification.
 Specifically, all relevant normative statements in Sections
 <a href="#basic-concepts"></a>, <a href="#advanced-concepts"></a>, and
-<a href="#syntaxes"></a> of this document MUST be enforced.
+<a href="#syntaxes"></a> of this document MUST be enforced. A serialization
+format for the conforming doucment MUST be deterministic and bi-directional.
```
brentzundel: document

jricher: Fixed, thank you!

index.html
```diff
@@ -544,7 +544,10 @@ <h3>Use Cases and Requirements</h3>
 that complies with the normative statements in this specification.
 Specifically, all relevant normative statements in Sections
 <a href="#basic-concepts"></a>, <a href="#advanced-concepts"></a>, and
-<a href="#syntaxes"></a> of this document MUST be enforced.
+<a href="#syntaxes"></a> of this document MUST be enforced. A serialization
+format for the conforming document MUST be deterministic and bi-directional.
```
msporny: What does being "bi-directional" mean? I know what you mean, but implementers may not. You may also mean "MUST be transformable to and from the data model in a deterministic and bi-directional fashion". Can we point to where you discuss this, such as the Syntax section?

jricher: This is defined further by the explanatory text included in #539

jricher: The associated non-normative text from #539 could be stated here instead as normative requirements: 

> The serialization
and de-serialization process has to be deterministic, bi-directional, and lossless.
Any serialization of a <a>verifiable credential</a> or <a>verifiable presentation</a> needs to
be transformable to the generic data model defined in this document in a deterministic
process such that the resulting verifiable credential can be processed
in an interoperable fashion. The serialized form also needs to be able
to be generated from the data model without loss of data or content.

msporny: Let me restate: You should refer the reader to that section.

I wouldn't try to transform all of that text into a normative statement and put it in the conformance section as 1) while the non-normative language is fine, it's fuzzy and partially untestable as a normative statement, and 2) it enables arguments stating that there are new normative requirements (instead of just some new non-normative clarification text). Transforming to normative statements would raise more than a few eyebrows and puts the group into a position to defend something we wouldn't have to defend if you were to just refer to the non-normative language.

So, just do something like this:

"...deterministic and bi-directional (see Section <a href="#syntaxes"></a>)"

jricher: Done.

dontcallmedom:

msporny:
Multiple independent positive reviews, non-substantive, merging.

Issue #551, opened 2019-04-15
Appendix examples not complete


David-Chadwick:
The examples in Appendix B (Subject-Holder Relationships) do not contain an @context property, which is mandatory. Furthermore. "https://www.w3.org/2018/credentials/examples/v1" does not contain aliases for all of the vocabulary used in Appendix B.

David-Chadwick:
This has been fixed in PR #563

burnburn:
Resolution on the VCWG 7 May 2019 call:
Issue #551 is addressed by PR #563 which makes non-normative changes to fix the missing @context values in a number of the examples in the appendix. Issue #551 should be closed.

PR #563, opened 2019-04-21
Update Appendix

David-Chadwick:
This fixes issue #551


TallTed:
This sets @context to *two* URIs, but the test suite (as linked in another recent thread) checks that @context is set to *one* URI.  Also, one of these two URIs is not one of the "only three" that currently exist.

Issue #363, opened 2019-01-03
[Mode of Operation] Is it necessary to specify only the *final* holder ...


mwherman2000:
In https://w3c.github.io/vc-data-model/#mode-of-operation, it states
> The final holder presents its verifiable credentials to the verifier in a verifiable presentation, requesting a supported action.

Is it necessary to specify that only the *final* holder a present its VCs to the verifier?  Why can't any holder in the chain present the VCs to a verifier vs. just the final holder?

David-Chadwick:
On 03/01/2019 15:55, Michael Herman (Toronto) wrote:
> In https://w3c.github.io/vc-data-model/#mode-of-operation, it states
> 
>     The final holder presents its verifiable credentials to the verifier
>     in a verifiable presentation, requesting a supported action.
> 
> Is it necessary to specify that only the /final/ holder a present its
> VCs to the verifier? Why can't any holder in the chain present the VCs
> to a verifier vs. just the final holder?

Of course any holder MAY present the VC, but in this case only the final
holder DID present the VC. So how to differentiate between the holder
that did and the holder that didnt?

David

msporny:
I believe @David-Chadwick has answered your question @mwherman2000 and I don't expect we need to add more text to the spec to make that distinction. Closing, please re-open if you disagree (and suggest concrete spec text so we have something to consider/refine).

brentzundel:
I believe this has been addressed via the solution for issue #577 and should be closed

PR #577, opened 2019-04-24
"final holder" may not be "final"

TallTed:
revisiting #363 - I think this deserves changing. "Final holder" suggests that the VC cannot be passed to another holder after this one -- but that's not so. As any step in the lifecycle can recur, and the order of those steps may vary, "final" has no place in it. (I also think this is editorial, but others may have differing opinion.)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/577.html" title="Last updated on Apr 24, 2019, 9:54 PM UTC (b1f6df9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/577/6207410...b1f6df9.html" title="Last updated on Apr 24, 2019, 9:54 PM UTC (b1f6df9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1820,24 +1820,24 @@ The <a>verifier</a> defines its policies for accepting
 <a>verifiable credentials</a> from <a>holders</a> for its supported actions.
           </li>
           <li>
-The <a>issuer</a> issues a <a>verifiable credential</a> to the <a>holder</a>.
+The <a>issuer</a> issues some <a>verifiable credential(s)</a> to the <a>holder</a>.
           </li>
           <li>
-The <a>holder</a> might pass on its <a>verifiable credentials</a> to another
+The <a>holder</a> might pass on its <a>verifiable credential(s)</a> to another
 <a>holder</a>.
           </li>
           <li>
-The final <a>holder</a> presents its <a>verifiable credentials</a> to the
+The <a>holder</a> presents its <a>verifiable credential(s)</a> to the
 <a>verifier</a> in a <a>verifiable presentation</a>, requesting a supported
 action.
           </li>
           <li>
 The <a>verifier</a> <a>verifies</a> the authenticity of the
-<a>verifiable presentation</a> and <a>verifiable credentials</a>.
+<a>verifiable presentation</a> and <a>verifiable credential(s)</a>.
           </li>
           <li>
 The <a>verifier</a> <a>verifies</a> that the <a>holder</a> possesses the
-<a>verifiable credentials</a>.
+<a>verifiable credential(s)</a>.
           </li>
           <li>
 The <a>verifier</a> decides whether to accept the <a>verifiable credentials</a>
--- a/index.html
+++ b/index.html
@@ -1816,37 +1816,39 @@ how the ecosystem is envisaged to operate, as follows:
 
         <ol>
           <li>
-The <a>verifier</a> defines its policies for accepting
+A <a>verifier</a> defines its own policies for accepting
 <a>verifiable credentials</a> from <a>holders</a> for its supported actions.
           </li>
           <li>
-The <a>issuer</a> issues some <a>verifiable credential(s)</a> to the <a>holder</a>.
+An <a>issuer</a> issues one or more <a>verifiable credentials</a> to 
+a <a>holder</a>.
           </li>
           <li>
-The <a>holder</a> might pass on its <a>verifiable credential(s)</a> to another
-<a>holder</a>.
+A <a>holder</a> might pass on one or more of its <a>verifiable credentials</a> 
+to another <a>holder</a>.
           </li>
           <li>
-The <a>holder</a> presents its <a>verifiable credential(s)</a> to the
-<a>verifier</a> in a <a>verifiable presentation</a>, requesting a supported
+A <a>holder</a> presents one or more of its <a>verifiable credential(s)</a> 
+to a <a>verifier</a> in a <a>verifiable presentation</a>, requesting a supported
 action.
           </li>
           <li>
-The <a>verifier</a> <a>verifies</a> the authenticity of the
-<a>verifiable presentation</a> and <a>verifiable credential(s)</a>.
+A <a>verifier</a> <a>verifies</a> the authenticity of a <a>verifiable presentation</a> 
+and the <a>verifiable credential(s)</a> contained therein.
           </li>
           <li>
-The <a>verifier</a> <a>verifies</a> that the <a>holder</a> possesses the
-<a>verifiable credential(s)</a>.
+A <a>verifier</a> <a>verifies</a> that the <a>holder</a> that made the 
+<a>verifiable presentation</a> possesses the <a>verifiable credential(s)</a>
+contained therein.
           </li>
           <li>
-The <a>verifier</a> decides whether to accept the <a>verifiable credentials</a>
-for the requested action, taking into account its policy, the <a>holder</a>,
-and the contents of the <a>verifiable credentials</a>.
+A <a>verifier</a> decides whether to accept the <a>verifiable credentials</a> 
+for the requested action, taking into account its policy, the presenting 
+<a>holder</a>, and the contents of the <a>verifiable credentials</a>.
           </li>
           <li>
-The <a>verifier</a> either performs the requested action or rejects the
-request.
+A <a>verifier</a> either performs a requested action or rejects the request
+for that action.
           </li>
         </ol>
 


TallTed:
It could be argued that each iteration of `The` (`The holder`, `The issuer`, `The verifier`) should also be changed to `A` ... which I think would remove @David-Chadwick's earlier concern about differentiating between "the holder
that did and the holder that didn't ...[present the VC]", as it's no longer "the" holder.

index.html
```diff
@@ -1820,24 +1820,24 @@ <h3>Lifecycle Details</h3>
 <a>verifiable credentials</a> from <a>holders</a> for its supported actions.
           </li>
           <li>
-The <a>issuer</a> issues a <a>verifiable credential</a> to the <a>holder</a>.
+The <a>issuer</a> issues some <a>verifiable credential(s)</a> to the <a>holder</a>.
```
David-Chadwick: The <a>issuer</a> issues one or more <a>verifiable credentials</a> to the <a>holder</a>.

index.html
```diff
@@ -1820,24 +1820,24 @@ <h3>Lifecycle Details</h3>
...
           </li>
           <li>
-The <a>holder</a> might pass on its <a>verifiable credentials</a> to another
+The <a>holder</a> might pass on its <a>verifiable credential(s)</a> to another
```
David-Chadwick: The <a>holder</a> might pass on one or more of its <a>verifiable credentials</a> to another

index.html
```diff
@@ -1820,24 +1820,24 @@ <h3>Lifecycle Details</h3>
...
 <a>holder</a>.
           </li>
           <li>
-The final <a>holder</a> presents its <a>verifiable credentials</a> to the
+The <a>holder</a> presents its <a>verifiable credential(s)</a> to the
```
David-Chadwick: A <a>holder</a> presents one or more of its <a>verifiable credentials</a> to the

Note. Now that we have more than one holder we can no longer refer to "the" holder as it is ambiguous. Final holder was not ambiguous.

TallTed: "Final holder" was not ambiguous, but it implied a termination of holder-to-holder transfers was required for a holder to present a VC to a verifier, which requirement does not exist.  As I noted in my earlier comment, "The" seems not to be appropriate for any of these bullets.

TallTed:
As I see it, relative to a single VC, there are a number of actions which may be taken by its single Issuer, any number of Holders (who may receive the VC from the Issuer or from a previous Holder), and any number of Verifiers (who may receive the VC from any Holder).

Some of those actions are dependent on specific previous actions (e.g., until a VC is Issued, nothing else can be done; until it is Presented, it cannot be Verified); some are not.

Most (possibly all) of those actions may be performed by multiple actors in the Holder and Verifier roles.

Many may be repeated by the same actor.  More may happen multiple times, as they are performed by multiple actors.

I am thinking that these actions might be better presented in a graphical form which indicates all of the above, than in purely textual form as they currently exist.

If the presentation does stay in textual form (and possibly also if it is reworked as an image), it seems to me that each action's description should include which action(s) are necessary (immediate?) prerequisites, and which action(s) may occur (immediately?) after it.

index.html
```diff
@@ -1816,37 +1816,39 @@ <h3>Lifecycle Details</h3>
...
-The <a>holder</a> might pass on its <a>verifiable credentials</a> to another
-<a>holder</a>.
+A <a>holder</a> might pass on one or more of its <a>verifiable credentials</a> 
+to another <a>holder</a>.
           </li>
           <li>
-The final <a>holder</a> presents its <a>verifiable credentials</a> to the
-<a>verifier</a> in a <a>verifiable presentation</a>, requesting a supported
+A <a>holder</a> presents one or more of its <a>verifiable credential(s)</a> 
+to a <a>verifier</a> in a <a>verifiable presentation</a>, requesting a supported
```
David-Chadwick: This has to say optionally since VPs are optional not mandatory. Therefore change to

A <a>holder</a> presents one or more of its <a>verifiable credential(s)</a> 
to a <a>verifier</a>, optionally in a <a>verifiable presentation</a>, requesting a supported

index.html
```diff
@@ -1816,37 +1816,39 @@ <h3>Lifecycle Details</h3>
...
 action.
           </li>
           <li>
-The <a>verifier</a> <a>verifies</a> the authenticity of the
-<a>verifiable presentation</a> and <a>verifiable credentials</a>.
+A <a>verifier</a> <a>verifies</a> the authenticity of a <a>verifiable presentation</a> 
+and the <a>verifiable credential(s)</a> contained therein.
           </li>
```
David-Chadwick: The previous step had a holder passing to a verifier. So this step should be talking about the same holder and the same verifier. It makes no sense to talk about any holder or any verifier since they may not have communicated with each other. You are already talking about THE verifiable credentials and not any old VCs, so you are inconsistent. Furthermore the VP is optional so may not be there to be verified. Therefore this text should read


The <a>verifier</a> <a>verifies</a> the authenticity of the presented <a>verifiable credentials</a> 
and the <a>verifiable presentation</a> containing them, if present.

index.html
```diff
@@ -1816,37 +1816,39 @@ <h3>Lifecycle Details</h3>
...
-<a>verifiable credentials</a>.
+A <a>verifier</a> <a>verifies</a> that the <a>holder</a> that made the 
+<a>verifiable presentation</a> possesses the <a>verifiable credential(s)</a>
+contained therein.
           </li>
           <li>
-The <a>verifier</a> decides whether to accept the <a>verifiable credentials</a>
-for the requested action, taking into account its policy, the <a>holder</a>,
-and the contents of the <a>verifiable credentials</a>.
+A <a>verifier</a> decides whether to accept the <a>verifiable credentials</a>
```
David-Chadwick: Similarly this should be The verifier for the reasons stated above. (Note you are already talking about The verifiable credentials :-)

index.html
```diff
@@ -1816,37 +1816,39 @@ <h3>Lifecycle Details</h3>
...
-and the contents of the <a>verifiable credentials</a>.
+A <a>verifier</a> decides whether to accept the <a>verifiable credentials</a> 
+for the requested action, taking into account its policy, the presenting 
+<a>holder</a>, and the contents of the <a>verifiable credentials</a>.
           </li>
           <li>
-The <a>verifier</a> either performs the requested action or rejects the
-request.
+A <a>verifier</a> either performs a requested action or rejects the request
+for that action.
```
David-Chadwick: This is rejected for the same reason as above. The original text was correct in using definite rather than indefinate articles.

TallTed:
@David-Chadwick - I refer you to the paragraph which follows the list of actions:

> The order of the steps above is not fixed, and some steps might be taken more than once. Such step-recurrence may be immediate or at any later point in the sequence as performed.

In other words, there isn't really a "previous step" nor a "next step" even though this is (currently) an ordered/numbered list.  Perhaps you might read [my other comment](https://github.com/w3c/vc-data-model/pull/577#issuecomment-486670445) and address thereby the basis of this PR, instead of focusing on details which depend on and descend from that basis?

Following on my previous -- This list of actions appears to me to be generic, and thus to apply to any/all VCs.  If you want to present the lifespan of a single, example VC, I think that should get distinct treatment -- perhaps in an Appendix -- in which the indefinite articles I'm suggesting for this generic list might become definite, for that specific example (say, one in which the VC is passed from Issuer to HolderA to VerifierA which permits RequestedActionA *and* from HolderA to HolderB to VerifierB which denies RequestedActionB)...

David-Chadwick:
@TallTed Saying that the order of the steps is not fixed is actually incorrect. The order of some of the steps is fixed, and some of the others is not fixed precisely but rather is approximately fixed. Some of the steps can be repeated and others not. So, you send a request and get a response. This order is fixed. You cannot get a response to a request that you have not yet issued (unless you are a time traveller). Also you must get a VC from an issuer before you can send it to the verifier. This order is fixed. So you are reading too much into the sentence you quoted and I suggest you change that sentence rather than other sentences that apply to a fixed order.

TallTed:
I believe there is *one* action which cannot take place more than once per VC, that being Issuance.  (Actually, I think there are *two*, with the other being Revocation, which I have just realized is not currently found in this lifecycle!)

I believe there are several actions which depend on one or more of the others occurring first.  Most obviously, all other actions on a VC depend on an Issuer having issued that VC -- and *that* should be #1, rather than the current "Verifier creates Policies" (which may not belong in this lifecycle at all, as such policies are not at all tied to a VC, per se).

Bottom line, this non-normative section hasn't received the attention it deserves, in part, I think, because it *is* non-normative.

I am increasingly convinced that the simple ordered-list is the wrong presentation.  Some sort of process flow diagram is making more sense to me, with clear indication of which actions may be followed by which others, including (possible) loopbacks.  I am probably not the best to create such a thing, but would be happy to work with someone who has experience with the relevant tool(s).

msporny:
We need more opinions on this PR in order to move it forward. I can weigh in, but would prefer that others do as well. We'll bring it up in the call tomorrow.

msporny:
This PR is replaced by PR #624, which is ready to merge. Closing this PR in favor of PR #624.

PR #588, opened 2019-05-03
clarify language around @context

TallTed:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/588.html" title="Last updated on May 6, 2019, 2:46 PM UTC (63bc986)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/588/ad2ec52...63bc986.html" title="Last updated on May 6, 2019, 2:46 PM UTC (63bc986)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1025,8 +1025,8 @@ related to specific <a>verifiable credentials</a> or
 The value of the <code>@context</code> <a>property</a> MUST be an ordered set
 where the first item is a <a>URI</a> with the value
 <code>https://www.w3.org/2018/credentials/v1</code>. Subsequent items in the
-array MUST be composed of any combination of <a>URIs</a> or objects that
-express context information. It is RECOMMENDED that each <a>URI</a> in the
+array MUST be composed of a combination of <a>URIs</a> and/or objects that
+expresses context information. It is RECOMMENDED that each <a>URI</a> in the
 <code>@context</code> be one which, if dereferenced, results in a document
 containing machine-readable information about the <code>@context</code>.
           </dd>
--- a/index.html
+++ b/index.html
@@ -1025,8 +1025,8 @@ related to specific <a>verifiable credentials</a> or
 The value of the <code>@context</code> <a>property</a> MUST be an ordered set
 where the first item is a <a>URI</a> with the value
 <code>https://www.w3.org/2018/credentials/v1</code>. Subsequent items in the
-array MUST be composed of a combination of <a>URIs</a> and/or objects that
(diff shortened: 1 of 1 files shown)


TallTed:
Rereading, I wonder if I misread this previously, and rather than --
```
Subsequent items in the
array MUST be composed of a combination of <a>URIs</a> and/or objects that
expresses context information.
```
-- the change should be to --
```
Subsequent items in the
array MUST express context information and be composed of any combination 
of <a>URIs</a> and/or objects.
```

TallTed:
OK.  The most important (and simpler, so best to come first) MUST is that the subsequent items must be expressions of `@context` info. Second is the composition, which I think is clearest with both `any` and `and/or`.
```
Subsequent items in the
array MUST express context information and be composed of any combination 
of <a>URIs</a> and/or objects.
```

David-Chadwick:
The main problem with the @context definition is that it does not clearly state what the @context is for. For example, if the entire VC compromises only of URIs for all its terms that require them, and has no short form aliases, then the @context is not needed, is it? There is no vocabulary to define. So I would like the definition to be changed to the following:
```
When two software systems need to exchange data, they must use terminology (and a protocol)
that both systems understand. As an analogy, consider how two people communicate. Both people
must use the same language and the words they use must mean the same thing to each other. 
This is often referred to as "the context in which a conversation happens". Verifiable credentials 
have many terms (cf words) that must be URIs. However URIs are long and not very user friendly to 
use. Short form user friendly aliases are more helpful. This specification uses the @context 
property to define the short form aliases for the URIs that specific verifiable credentials or 
verifiable presentations require. 
```

TallTed:
@David-Chadwick - I think that becomes a distinct PR. I have some tweaks to your wording, which I hope you'll take there --

`When two software systems need to exchange data, they need to use terminology that both systems understand. (They also need to use a shared protocol; one or more protocols for exchanging Verified Credentials will be defined by future work.) As an analogy, consider how two people communicate. Both people must use the same language, and the words they use must mean the same thing to each other. This may be referred to as "the context of a conversation".`

`Verifiable credentials and verifiable presentations have many attributes and values which are identified by URIs. However, those URIs can be long and not very human-friendly. In such cases, short-form human-friendly aliases can be more helpful. This specification uses the @context property to map such short-form aliases to the URIs required by specific verifiable credentials or verifiable presentations.`

David-Chadwick:
Done in PR  #595

PR #595, opened 2019-05-06
Update @context definition

David-Chadwick:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/595.html" title="Last updated on May 8, 2019, 10:23 AM UTC (6d335c5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/595/bb5ef42...6d335c5.html" title="Last updated on May 8, 2019, 10:23 AM UTC (6d335c5)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1005,8 +1005,7 @@ document.
 
         <p>
 When two software systems need to exchange data, they need to use terminology that both
-systems understand. (They also need to use a shared protocol; one or more protocols for
-exchanging <a>verified credentials</a> will be defined by future work.)
+systems understand. 
 As an analogy, consider how two people communicate. Both people must use the same
 language and the words they use must mean the same thing to each other. This may be
 referred to as "the context of a conversation". 
@@ -1019,6 +1018,15 @@ aliases can be more helpful. This specification uses
 the <code>@context</code> <a>property</a> to map such short-form aliases to the URIs
 required by specific <a>verifiable credentials</a> and <a>verifiable presentations</a>.
        </p>
+  <p class="note">
+In JSON-LD, the <code>@context</code> <a>property</a> can also be used
+to communicate various other details (datatype information, language
+information, transformation rules, etc.) which are beyond the needs of
+this specification, but may be useful in the future or to related work.
+Interested readers may wish to review
+<a href="https://www.w3.org/TR/json-ld11/#the-context">Section 3.1: The Context</a>
+of the [[JSON-LD]] specification.
+        </p>
         <p>
 <a>Verifiable credentials</a> and <a>verifiable presentations</a> MUST include a
 <code>@context</code> <a>property</a>.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1004,16 +1004,21 @@ <h2>Basic Concepts</h2>
...
-people communicate. Both people must use the same language and the words they
-use must mean the same thing to each other. This is often referred to as
-"the context in which a conversation happens". This specification defines the
-use of the <code>@context</code> <a>property</a> when communicating information
-related to specific <a>verifiable credentials</a> or
-<a>verifiable presentations</a>.
-        </p>
-
+When two software systems need to exchange data, they need to use terminology that both
+systems understand. (They also need to use a shared protocol; one or more protocols for
```
brentzundel: recommend removing the parenthetical comment.

msporny: Agreed, it'll ruin the surprise. :P 

It also adds more than necessary to get the point across.

index.html
```diff
@@ -1004,16 +1004,21 @@ <h2>Basic Concepts</h2>
...
+As an analogy, consider how two people communicate. Both people must use the same
+language and the words they use must mean the same thing to each other. This may be
+referred to as "the context of a conversation". 
+        </p>
+        <p>
+<a>Verifiable credentials</a> and <a>verifiable presentations</a> have many
+attributes and values which are identified by <a>URIs</a>. However, those <a>URIs</a>
+can be long and not very human-friendly. In such cases, short-form human-friendly
+aliases can be more helpful. This specification uses
+the <code>@context</code> <a>property</a> to map such short-form aliases to the URIs
```
msporny: This is not the only thing the @context does... it also specifies datatype information, language information, transformation rules, etc. So, if we're going to mention URI terms, we should also state that the @context does more and if the reader is interested, they should read more about it in the JSON-LD specification: https://www.w3.org/TR/json-ld11/#the-context

David-Chadwick: I don't want the reader to have to read more. I would like the reader to know from this specification what the purpose of @context is *for this spec*, and if they then want to know more details, they can read the JSON-LD spec. Since you know all that @context does for VCs then can you please add the necessary extra sentence(s). Thanks.

index.html
David-Chadwick: Happy to do that. The only reason that the protocol part is there is because protocol was there in the original text. But it is not needed in the revised text in my opinion.

index.html
TallTed: just a quick thought --
```
        <p class="note">
In JSON-LD, the <code>@context</code> <a>property</a> can also be used
to communicate various other details (datatype information, language
information, transformation rules, etc.) which are beyond the needs of
this specification, but may be useful in the future or to related work.
Interested readers may wish to review
<a href="https://www.w3.org/TR/json-ld11/#the-context">Section 3.1: The Context</a>
of the [[JSON-LD]] specification.
        </p>
```

David-Chadwick: Many thanks. I have incorporated the Note

Issue #545, opened 2019-04-13
Disputes


David-Chadwick:
There are no conformance statements in the Disputes section. If we want there to be a standard disputes mechanism, then we should have conformance statements that say the following:
i) if a subject wants to dispute a VC he/she MAY issue a disputes VC
ii) the credentialSubject of a disputes VC MUST contain the "currentStatus" property with the value set to "Disputed"
iii) the id of the credentialSubject MUST be set to the id of the disputed VC
iv) the credentialSubject of a disputes VC MAY contain the "statusReason" and "disputedClaim" properties

msporny:
I expect that the group clearly meant this to be a non-normative section (and I could swear I wrote normative text for this at some point... perhaps that got removed in a series of edits?). The examples show what properties we intended to be used (at the very least, type `DisputeCredential`).

`statusReason` and `disputedClaim` were meant to be examples that we'd verify during implementation.

We'll check to see what the expectation from the WG was for this section. Specifically, confirm that it was supposed to be normative, and then confirm the intended normative text, which would align with the non-normative text in the section.

msporny:
Looking at this a bit more deeply, we did warn implementers that the feature might change during the CR period, so that provides more leeway than normal to modify the feature. I think we can do the following two things that @David-Chadwick mentioned w/o issue:

> i) if a subject wants to dispute a VC he/she MAY issue a disputes VC

> ii) the credentialSubject of a disputes VC MUST contain the "currentStatus" property with the value set to "Disputed"

> iii) the id of the credentialSubject MUST be set to the id of the disputed VC

This one is a bit more difficult:

> iv) the credentialSubject of a disputes VC MAY contain the "statusReason" and "disputedClaim" properties

Saying it may contain `statusReason` is fine.

The "disputedClaim" property name is problematic (for the same reason that `claim` was problematic). We could do something like `disputedSubject`, but that feels weird. We could re-use `credentialSubject`, but again... feels a bit strange. Need someone else to weigh in on this last item. If we can't get to consensus wrt. what to rename it to, then I suggest we don't put any normative statements in about it and instead write about it in the implementation guide or in the VC Extensions Registry.

David-Chadwick:
I suggest we use disputedProperty since this is what is being disputed

stonematt:
From VCWG call on April 30, 2019: 
[...]

PR #599, opened 2019-05-07
Move Disputes section to Implementation Guidance.

msporny:
Related to #545.

The Disputes section now exists in the implementation guidance here: https://w3c.github.io/vc-imp-guide/#disputes

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/599.html" title="Last updated on May 7, 2019, 4:06 AM UTC (8046ce0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/599/5577f77...8046ce0.html" title="Last updated on May 7, 2019, 4:06 AM UTC (8046ce0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1025,7 +1025,7 @@ related to specific <a>verifiable credentials</a> or
 The value of the <code>@context</code> <a>property</a> MUST be an ordered set
 where the first item is a <a>URI</a> with the value
 <code>https://www.w3.org/2018/credentials/v1</code>. Subsequent items in the
-array MUST express context information and be composed of any combination 
+array MUST express context information and be composed of any combination
 of <a>URIs</a> and/or objects. It is RECOMMENDED that each <a>URI</a> in the
 <code>@context</code> be one which, if dereferenced, results in a document
 containing machine-readable information about the <code>@context</code>.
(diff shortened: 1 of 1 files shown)


David-Chadwick:
I dont agree with this, since the implementation guide has no formal requirements placed on implementations (unless it also becomes a Recommendation). I originally raised this issue because the conformance statements about Disputes were missing by oversight. Instead of fixing this, you are proposing to move the section entirely. I presume that this is so that another CR does not need to be issued. We have had a number of these issues that should have required a new CR, each issue being small on its own, but when taken together eventually these straws do break the camels back. Is it now time that we agreed that a new CR is needed because we have a number of changes that are required?

jricher:
I'm not sure on this move either. Disputed claims aren't really an implementation detail, they're core to the nature of verification of the claim.

TallTed:
A VC issued to dispute a second VC is just a VC.  (And I think this applies equally to a VC issued to rescind or revoke or cancel a VC.)

A VC ***Protocol*** may involve checking whether such a dispute or revocation exists, whether conflicting VCs exist, etc., via whatever means (which we have not discussed in any detail), but just as everything else about what is done with a VC (vs the shape of its content), I think such a check cannot be properly considered to be a part of the data ***model.***

Frankly, even putting this into an implementation guide doesn't make much sense to me at this point, because implementing guidance for a data model is not much more than "when describing this class of entity, use these attributes, with these ranges of values." 

Thinking about it, "Suggested Usage Patterns" (or even "Anticipated Usage Patterns") might be a better name for that doc....

David-Chadwick:
@TallTed Whilst we agree that a dispute checking protocol is out of scope, nevertheless the data model does need to have properties to support the issuing of dispute type credentials. This is what the current text is there to support. Unfortunately it failed to place any conformance statements on these properties, which was purely an oversight collectively by the group.
Can we avoid a new CR by placing non-mandatory conformance statements in the current CR (ie. MAY or SHOULD statements). Its less than ideal I know, but better than removing the text entirely.

msporny:
@David-Chadwick wrote:
> Is it now time that we agreed that a new CR is needed because we have a number of changes that are required?

As mentioned on the call last week, we are out of time and cannot go through another CR without a charter extension (which would be a very difficult process and the WG has already agreed is not something it wants to do).

@David-Chadwick and @jricher, this PR implements the following RESOLUTION from the WG:

From VCWG call on April 30, 2019:
RESOLUTION: Issue #545 should be addressed by moving the majority of at risk Disputes section to the Implementation Guide and referring to that section from the Data Model specification, and making the Disputes section in the VC Data Model specification non-normative. Issue #545 should be closed after the PR is merged.

I'm going to leave this hanging for now, but you two need to be clear if you're objecting to a WG resolution and are requesting that the WG re-review this resolution... and if so, what is the new information being presented that would trigger a re-review.

msporny:
This PR is replaced by PR #629, closing.

Issue #612, opened 2019-05-09
Proofs Note


nadalin:
"No single proof mechanism is currently standardized or recommended for verifiable credentials." This is incorrect as JWS is standardized in IETF and widely used. Suggest rewording this NOTE

TallTed:
The operative elements in your quoted sentence are "No **single** ... **for verifiable credentials**."  In other words, we're not picking a winner.

That said, a viable rewording of this sentence might be `As discussed in Section <a href="#conformance"></a>, there are multiple viable proof mechanisms, and this specification does not standardize nor recommend any single proof mechanism for use with <a>verifiable credentials</a>.`  What do you think?

nadalin:
+1

burnburn:
WG resolution on 28 May 2019 call:

RESOLUTION: Issue #612 is addressed via PR #614, which rewords the text of concern in a non-normative way. Issue #612 should be closed after 7 days from when the PR was merged.

nadalin:
So no real interop can take place based upon PR #614

burnburn:
Your interop concern is being tracked by issue #632.  Closing.

PR #614, opened 2019-05-09
rephrasing of proof mechanism note

TallTed:
addresses #612

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/614.html" title="Last updated on May 14, 2019, 2:23 PM UTC (0f54ac4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/614/d2a8a5f...0f54ac4.html" title="Last updated on May 14, 2019, 2:23 PM UTC (0f54ac4)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1570,8 +1570,9 @@ the signing date. The example below uses RSA digital signatures.
         </pre>
 
         <p class="note">
-No single proof mechanism is currently standardized or recommended for
-<a>verifiable credentials</a>.
+As discussed in Section <a href="#conformance"></a>, there are multiple viable 
+proof mechanisms, and this specification does not standardize nor recommend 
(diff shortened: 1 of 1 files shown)


msporny:
Conflict. Also, this text is not the only place it exists in the spec, IIRC. You might want to make a pass to make sure you're getting all the places where we state this.

TallTed:
@msporny - Conflict resolved. I had borrowed wording from the only other section I found that discussed this, so I think (hope) all such are now addressed.

nadalin:
@msporny This should really be split so readers understand what are already developed standards  and what are still being worked on and subject to change and analysis as this PR merges it all into one bucket. So this does not solve my concern

PR #620, opened 2019-05-11
Update definition of issuance date

David-Chadwick:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/620.html" title="Last updated on May 13, 2019, 6:16 PM UTC (0ae3195)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/620/5577f77...0ae3195.html" title="Last updated on May 13, 2019, 6:16 PM UTC (0ae3195)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1467,7 +1467,7 @@ The value of the <code>issuer</code> property may also be a JWK, e.g.,
 
         <p>
 This specification defines the following <a>property</a> for expressing the
-date and time when the <a>credential</a> was issued:
+date and time when the <a>credential</a> becomes valid:
         </p>
 
         <dl>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1476,9 +1476,9 @@ <h3>Issuance Date</h3>
 A <a>credential</a> MUST have an <code>issuanceDate</code> <a>property</a>. The
 value of the <code>issuanceDate</code> <a>property</a> MUST be a string value of
 an [[!RFC3339]] combined date and time string representing the date and time
-the <a>credential</a> was issued. Note that this date represents the earliest
+the <a>credential</a> becomes valid. Note that this date represents the earliest
 date when the information associated with the <code>credentialSubject</code>
```
brentzundel:
```suggestion
```

index.html
```diff
@@ -1476,9 +1476,9 @@ <h3>Issuance Date</h3>
 A <a>credential</a> MUST have an <code>issuanceDate</code> <a>property</a>. The
 value of the <code>issuanceDate</code> <a>property</a> MUST be a string value of
 an [[!RFC3339]] combined date and time string representing the date and time
-the <a>credential</a> was issued. Note that this date represents the earliest
+the <a>credential</a> becomes valid. Note that this date represents the earliest
```
brentzundel: Perhaps:
> the <a>credential</a> becomes valid, which could be a date and time in the future.

Then we don't need the redundant:
> Note that this represents . . .

David-Chadwick: @brentzundel. The original text talked about both the VC and the credential subject. So I have kept this for now, unless people think it is unnecessary.

David-Chadwick: @brentzundel . Take a look at the new Diff and see if you like it now

brentzundel: I like it better. The different sentence constructions help to better clarify that they are talking about two different things. 
I wonder if the sentence starting with "Note" should actually be a note, set apart as such in the text.

Issue #622, opened 2019-05-13
Example context needs updating


dmitrizagidulin:
From @yancyribbens's [original issue at vc-test-suite](https://github.com/w3c/vc-test-suite/issues/19):

The context: https://www.w3.org/2018/credentials/examples/v1 is invalid JSON. If you paste the context into https://jsonlint.com/ the following parse error occurs:

```
Error: Parse error on line 24:
...			"schema:alumniOf",			"@type": rdf: H
-----------------------^
Expecting ':', got ','
```

Moving it here, since this is where the `examples/v1` context is hosted. (A valid examples context can be found at https://github.com/digitalbazaar/vc-js/blob/master/lib/contexts/vc-examples-v1.js , if that's helpful)

dmitrizagidulin:
Fix pending at PR #623.

msporny:
PR #623 is merged, we should close this issue.

burnburn:
WG resolution on 28 May 2019 call:

RESOLUTION: Issue #622 is resolved via PR #623, which fixes a bug in the JSON-LD context. Issue #622 should be closed immediately.

PR #623, opened 2019-05-13
Fix invalid context entries, add required fields.

dmitrizagidulin:
Fixes issue https://github.com/w3c/vc-data-model/issues/622 and https://github.com/w3c/vc-test-suite/issues/19.

(Also adds some other missing fields required to pass the vc-test-suite).

### DIFF
--- a/contexts/credentials/examples/v1
+++ b/contexts/credentials/examples/v1
@@ -20,11 +20,15 @@
     "UniversityDegreeCredential": "ex:UniversityDegreeCredential",
     "ZkpExampleSchema2018": "ex:ZkpExampleSchema2018",
 
-    "alumniOf": {"schema:alumniOf", "@type": rdf:HTML},
+    "alumniOf": {"@id": "schema:alumniOf", "@type": "rdf:HTML"},
     "child": {"@id": "ex:child", "@type": "@id"},
     "degree": "ex:degree",
-    "name": {"schema:name", "@type": rdf:HTML},
+    "name": {"@id": "schema:name", "@type": "rdf:HTML"},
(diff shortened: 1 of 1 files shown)


contexts/credentials/examples/v1
```diff
@@ -20,11 +20,15 @@
     "UniversityDegreeCredential": "ex:UniversityDegreeCredential",
     "ZkpExampleSchema2018": "ex:ZkpExampleSchema2018",
 
-    "alumniOf": {"schema:alumniOf", "@type": rdf:HTML},
+    "alumniOf": {"@id": "schema:alumniOf", "@type": "rdf:HTML"},
```
msporny: Based on a discussion with @chaals today, this is going to be changing to a scoped context that supports language expressions. I don't have the details yet, have to talk w/ @dlongley about it, but it will most likely be a scoped context that aliases @value, @language, and @dir to `value`, `language`, and `dir` respectively (and presumes that @dir is going to be a special JSON-LD 1.1 keyword.

contexts/credentials/examples/v1
```diff
@@ -20,11 +20,15 @@
...
     "child": {"@id": "ex:child", "@type": "@id"},
     "degree": "ex:degree",
-    "name": {"schema:name", "@type": rdf:HTML},
+    "name": {"@id": "schema:name", "@type": "rdf:HTML"},
```
msporny: Same as above.

contexts/credentials/examples/v1
dmitrizagidulin: @msporny Got it, sounds good.
Although - can we merge this change now (so that at least the context doesn't error out), and do the language/dir change separately?

Issue #604, opened 2019-05-08
@Context


nadalin:
"Though this specification requires that a @context property be present, there is no requirement that the value of the @context property be processed using JSON-LD. This is to support processing using plain JSON libraries such as those that may be used when the verifiable credential is encoded as a JWT. JSON-LD aware processors can process the @context using full JSON-LD processing as expected. For other processors, the only processing necessary is to ensure that the order of the values in @context is what is expected for the particular application, but no JSON-LD processing of those values is required."

This is odd wording, as you are saying that "processing of the @Context is required" no matter if you are using JSON or JSON-LD, which is not stated in the @Context section at all, just in the note

brentzundel:
I'm not sure what issue you are bringing up with this comment. What you have said is true, but is not clear to me what you feel ought to be changed.

nadalin:
@brentzundel Issue as that can't even begin to parse what is needed to be processed when and by what

chaals:
> the only processing necessary is to ensure that the order of the values in @context is what is expected for the particular application

This appears to be the key requirement for everyone.

> JSON-LD aware processors can process the @context using full JSON-LD processing as expected

That seems pretty clear.

The rest seems to be rationale.

Would it be clearer if we broke the paragraph into pieces? Or have I misinterpreted it too?

TallTed:
A rephrasing is definitely worthwhile here. Something along the lines of the following might replace the entire quote in the OP here.  

> Plain JSON libraries or processors (such as those that may be used when the verifiable credential is encoded as a JWT) MUST ensure that the order of the values in the **@context** property is what is expected for the particular application. Libraries or processors that support JSON-LD SHOULD process the **@context** property using full JSON-LD processing.

brentzundel:
Would it be accurate to clarify it even further?

All libraries or processors MUST ensure that the order of the values
in the *@context
<https://github.com/context>* property is what is expected for the
particular application. Libraries or processors that support JSON-LD SHOULD
process the *@context <https://github.com/context>* property using full
JSON-LD processing.

msporny:
Adopt suggestions from @brentzundel to address this issue.

msporny:
Non-substantive suggestions by @brentzundel have been merged. Suggest closing this issue since it's been addressed by PR #630.

burnburn:
WG resolution on 28 May 2019 call:

[...]

PR #630, opened 2019-05-14
clarified @context processing

brentzundel:
addresses issue #604 

***
<a href="https://pr-preview.s3.amazonaws.com/brentzundel/vc-data-model/pull/630.html" title="Last updated on May 15, 2019, 9:44 PM UTC (3ae0681)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/630/66e81c7...brentzundel:3ae0681.html" title="Last updated on May 15, 2019, 9:44 PM UTC (3ae0681)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1065,11 +1065,10 @@ be present, there is no requirement that the value of the <code>@context</code>
 property be processed using JSON-LD. This is to support processing using plain
 JSON libraries such as those that may be used when the
 <a>verifiable credential</a> is encoded as a JWT.
-JSON-LD aware processors can process the <code>@context</code> using full
-JSON-LD processing as expected. For other processors, the only processing
-necessary is to ensure that the order of the values in <code>@context</code>
-is what is expected for the particular application, but no JSON-LD processing
-of those values is required.
+All libraries or processors MUST ensure that the order of the values in the
+<code>@context</code> <a>property</a> is what is expected for the particular
+application. Libraries or processors that support JSON-LD SHOULD
+process the <code>@context</code> <a>property</a> using full JSON-LD processing.
         </p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1065,11 +1065,10 @@ <h3>Contexts</h3>
...
 <a>verifiable credential</a> is encoded as a JWT.
-JSON-LD aware processors can process the <code>@context</code> using full
-JSON-LD processing as expected. For other processors, the only processing
-necessary is to ensure that the order of the values in <code>@context</code>
-is what is expected for the particular application, but no JSON-LD processing
-of those values is required.
+All libraries or processors MUST ensure that the order of the values in the
+<code>@context</code> <a>property</a> is what is expected for the particular
+application. Libraries or processors that support JSON-LD SHOULD
+process the <code>@context</code> <a>property</a> using full JSON-LD processing.
```
```suggestion
process the <code>@context</code> <a>property</a> using full JSON-LD processing as needed.
```

I suggest we add `as needed` at the end there because you don't *need* to do any JSON-LD processing if your application already knows the contexts used in the VC -- even if your library/processor supports JSON-LD. You'd only do JSON-LD processing, for example, to convert from an unknown context to a known one. In other words, library/processors that support JSON-LD will have this additional capability and they can use it as expected when necessary. I think the original text captured that part a bit better (while it was lacking on addressing the ordering issue).

Perhaps we want to say:

```
All libraries or processors MUST ensure that the order of the values in the
<code>@context</code> <a>property</a> is what is expected for the particular
application. Libraries or processors that support JSON-LD can process
the <code>@context</code> <a>property</a> using full JSON-LD processing
as expected.
```

nadalin:
Looks good

Issue #608, opened 2019-05-08
Section 4.3


nadalin:
"The value of the type property MUST be, or map to (via interpretation of the @context property), one or more URIs. If more than one URI is provided, the URIs MUST be interpreted as an unordered set. Syntactic conveniences, such as JSON-LD terms, SHOULD be used to ease developer usage. It is RECOMMENDED that each URI in the type be one which, if dereferenced, results in a document containing machine-readable information about the type."

It is unclear as written since JSON-LD is NOT required why a SHOULD is used. This information would be better served in a non-normative note

nadalin:
"Verifiable credentials and verifiable presentations MUST have a type property. " is stated then it goes on to list more … "With respect to this specification, the following table lists examples of objects that MUST have a type specified. " I would suggest removing the 1st sentence ""Verifiable credentials and verifiable presentations MUST have a type property" and just having 1 place and list them all

David-Chadwick:
All the examples in the table have duplicate statements about the type property being mandatory. For VCs and VPs the duplicate statement is in section 4.3. For the other properties the duplicate is in the appropriate section. If you want to remove anything it should be the table, which is only providing examples. It is not the definitive list.

nadalin:
@David-Chadwick The issue is they are not all inclusive, there is a mismatch is what the Statement provides and what the table provides

TallTed:
@nadalin - 

To the initial comment in this issue, would that be sufficiently addressed by rewording --

>Syntactic conveniences, such as JSON-LD terms, SHOULD be used to ease developer usage.

-- to the following (which says exactly the same thing) --

>Syntactic conveniences SHOULD be used to ease developer usage; such conveniences may include JSON-LD terms.

----

I wonder if your second concern here could be addressed by adding entries for verifiable credentials and verifiable presentations (and any others that are missing, if there are any such) to the table (i.e., making it a complete list of all objects (rather than "examples of objects") in the spec that require type specification, and thus a summary presentation of details laid out in the other sections), rather than removing the initial sentence?

msporny:
PR #631 has been merged. Suggest we close this issue since the issue submitter's concerns have been addressed with PR #631.

burnburn:
WG resolution on 28 May 2019 call: 

RESOLUTION: Issue #608 is addressed via PR #631 which adds non-normative clarification around the SHOULD of concern to the reviewer. Issue #608 will be closed after a 7 day wait period.

PR #631, opened 2019-05-15
clarify when type property is required

TallTed:
to address #608

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/631.html" title="Last updated on May 16, 2019, 3:25 PM UTC (9bfd788)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/631/66e81c7...9bfd788.html" title="Last updated on May 16, 2019, 3:25 PM UTC (9bfd788)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1276,7 +1276,7 @@ objects that MUST have a <a>type</a> specified.
 <a>verifiable credential</a>&nbsp;object <br />(a subclass of <a href="#credentials">Credential</a>&nbsp;object)
               </td>
               <td>
-<code>VerifiableCredential</code> and a more specific
+<code>VerifiableCredential</code> and, optionally, a more specific
 <a>verifiable credential</a> <a>type</a>. For example,<br>
 <code>"type": ["VerifiableCredential", "UniversityDegreeCredential"]</code>
               </td>
--- a/index.html
+++ b/index.html
@@ -1219,7 +1219,10 @@ expression of type information.
 
         <p>
 <a>Verifiable credentials</a> and <a>verifiable presentations</a> MUST have a
-<code>type</code> <a>property</a>.
+<code>type</code> <a>property</a>; in other words, any <a>credential</a> or 
+<a>presentation</a> that lacks a <code>type</code> <a>property</a> <em>is not 
+verifiable<em>, so is not a <a>verifiable credential</a> nor a 
+<a>verifiable presentation</a>.
         </p>
 
         <dl>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1268,6 +1271,17 @@ <h3>Types</h3>
...
 
           <tbody>
+            <tr>
+              <td>
+<a>verifiable credential</a>&nbsp;object <br />(a subclass of <a href="#credentials">Credential</a>&nbsp;object)
+              </td>
+              <td>
+<code>VerifiableCredential</code> and a more specific
+<a>verifiable credential</a> <a>type</a>. For example,<br>
+<code>"type": ["VerifiableCredential", "UniversityDegreeCredential"]</code>
```
dlongley: We don't have a hard requirement to have more than one type for a credential or a presentation; the only MUST is that there is *at least* `VerifiableCredential` on a VC and *at least* `VerifiablePresentation` on a VP. The above change might make people think you must have two types on each. I've suggested `optionally` in both cases to address this.

index.html
```diff
@@ -1268,6 +1271,17 @@ <h3>Types</h3>
           </thead>
 
           <tbody>
+            <tr>
+              <td>
+<a>verifiable credential</a>&nbsp;object <br />(a subclass of <a href="#credentials">Credential</a>&nbsp;object)
+              </td>
+              <td>
+<code>VerifiableCredential</code> and a more specific
```
```suggestion
<code>VerifiableCredential</code> and, optionally, a more specific
```

index.html
```diff
@@ -1279,6 +1293,17 @@ <h3>Types</h3>
               </td>
             </tr>
 
+            <tr>
+              <td>
+<a>verifiable presentation</a>&nbsp;object <br />(a subclass of <a href="#presentations">Presentation</a>&nbsp;object)
+              </td>
+              <td>
+<code>VerifiablePresentation</code> and a more specific
```
```suggestion
<code>VerifiablePresentation</code> and, optionally, a more specific
```

index.html
dlongley: Well, apparently our spec *does* say this is required at present. I feel like that is a mistake, but one that can be lived with. It likely creates an undue burden on VPs, though on VCs it's sensible. I need to check to see how long that's been in the spec or if we added as an attempt to clarify and made a mistake.

dlongley: Looks like this change brought in that requirement on VPs and encapsulated objects:

https://github.com/w3c/vc-data-model/commit/30f8bfc338ff3c9449e41a78a8c09133dbe79751

But it left open the door to omitting specifying this information with this clause (emphasis mine):

> All credentials, presentations, and encapsulated objects MUST specify, ***or be associated with***, additional more narrow types...

I'm not sure how to interpret "or be associated with", but we should clarify. One interpretation is that VCs and VPs should contain enough sufficiently typed information to allow for processing. I don't think the intent was to ever require narrow types for every VP and encapsulated object and it seems that clause allows for wiggle room there. Prior to this change there was a requirement for a more narrow type for credentials, again, which is sensible, but even that, I don't think, should be strictly necessary.

TallTed: I believe I agree... Merging your tweaks.

David-Chadwick: @dlongley "We don't have a hard requirement to have more than one type for a credential or a presentation"

I believe we do for a verifiable credential at least. The type parameter is there to tell the JSON parser what type of VC this is so that it knows what properties to expect. The credentialSchema property further says which properties are mandatory and which are optional.

The phrase "or be associated with" was there so that encapsulated objects did not have to have their types in the objects themselves, but they could be in the encapsulating object so that the parser could sooner determine if it can process this type of VC or not.

We had several long discussions about whether VCs should be strongly typed or not and if you look through the minutes you will see that strongly typed was agreed.

dlongley: @David-Chadwick, as stated above, I'm less concerned about this for VCs and it's generally good practice to include a more narrow type. This is because there's an expectation that extensions will be created to provide the necessary terms to express whatever is being claimed about credential subject(s).

However, VPs are quite usable without such extensions. I think requiring an extra type for VPs is a problem that would just ultimately result in the creation of an additional, generally useless (semantically irrelevant) type that people tack onto their VPs just to make them conformant.

Issue #436, opened 2019-03-04
Many natural language text fields (such as "statusReason") have no language or direction support


aphillips:
A number of natural language text fields appear in the document without support for language or base direction. For discussion of why this is important and some guidelines, see [String-Meta](https://w3c.github.io/string-meta). One example is:

6.2 Disputes
https://w3c.github.io/vc-data-model/#disputes

>     "statusReason": "Credential contains disputed statements",

The property `statusReason` can contain natural language text (not merely an identifier or value string, as in most other examples). Correct rendering in some languages depends on additional metadata.

This is [I18N comment 580](https://github.com/w3c/i18n-activity/issues/580)

aphillips:
(please add the `i18n-comment` label)

brentzundel:
Decision from F2F Barcelona:
The support for this depends on the underlying syntax standards.
- [ ] @msporny will submit a PR that adds language explaining how JSON-LD 1.1 addresses this.
- [ ] @chaals will submit a PR that recommends the approach an issuer who uses JSON should take to address this issue, along with an example.

burnburn:
Hey @msporny @chaals I thought both of you said that Manu had submitted the second PR, but I can't find it.  Can you please link it here?

msporny:
> Hey @msporny @chaals I thought both of you said that Manu had submitted the second PR, but I can't find it. Can you please link it here?

Not yet, working on it now.

msporny:
I added an i18n considerations section in https://github.com/w3c/vc-data-model/commit/1cadfb11f2c9e7ed09dcba7cde3fe2c9d1b41f77. It provides some guidance wrt. how to handle i18n correctly based on work done in the W3C Publishing WG and W3C JSON-LD 1.1 WG.

Closing issue, please re-open if you disagree with the solution.

chaals:
The current approach assumes that processors can handle HTML - it makes sense to test this in CR as implementation information so I am not re-opening.

An alternative would be to make strings into arrays or objects,  taking some of the @language approach from JSON-LD 1.1 (which could be readily extended to handle base direction with something like `@textDirection` ...)

For cases where a field is essentially bilingual, this might be a better approach. Instead of requiring HTML support it requires specific ability to handle the structure. On the other hand, this would enable plain JSON to actually deal with language information, in a way that is compatible with JSON-LD (by requiring processors to implement whatever subset of the JSON-LD mechanism is chosen)...

aphillips:
[...]

PR #641, opened 2019-05-25
Rewrite Internationalization section.

msporny:
Attempt to address all remaining i18n considerations after talking w/ i18n folks, JSON-LD folks, and various people with strong opinions. This PR suggests a mechanism that works for JSON, JSON-LD, RDF, and HTML syntaxes/approaches while addressing all known i18n concerns raised to date.

Related to issue #436.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/641.html" title="Last updated on Jun 30, 2019, 6:53 PM UTC (ea43b25)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/641/446c421...ea43b25.html" title="Last updated on Jun 30, 2019, 6:53 PM UTC (ea43b25)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4737,7 +4737,7 @@ this topic.
 This section outlines general internationalization considerations to take into
 account when utilizing this data model and is intended to highlight specific
 parts of the <em>Strings on the Web: Language and Direction Metadata</em>
-document [[STRING-META]] that implementers might be interested in perusing.
+document [[STRING-META]] that implementers might be interested in reading.
     </p>
 
     <section class="informative">
@@ -4767,28 +4767,34 @@ specific base direction:
 
       <p>
 An example utilizing the design pattern above is shown below to express the
-human-readable text "revoked" in the English language without specifying a
-text direction:
+title of a book in the English language without specifying a text direction:
       </p>
 
       <pre class="example nohighlight" title="Expressing natural language text as English">
-"statusReason": {
-  "value": "<span class="highlight">Revoked</span>",
+"title": {
+  "value": "<span class="highlight">HTML and CSS: Designing and Creating Websites</span>",
   "lang": "<code>en</code>"
 }
       </pre>
 
       <p>
-The next example uses a term in Arabic with a base direction of right-to-left:
+The next example uses a similar title expressed in the Arabic language with a
+base direction of right-to-left:
       </p>
 
       <pre class="example nohighlight" title="Arabic text with a base direction of right-to-left">
-"statusReason": {
-  "value": "<span class="highlight">إلغاء</span>",
+"title": {
+  "value": "<span class="highlight">HTML و CSS: تصميم و إنشاء مواقع الويب</span>",
   "lang": "<code>ar</code>"
   "dir": "<code>rtl</code>"
 }
       </pre>
+
+      <p class="note">
+The text above would most likley be rendered incorrectly as left to right
+without the explicit expression of language and direction as many systems will
+use the first character of a text string to determine text direction.
+      <p>
 
       <p>
 Utilization of the design pattern above assumes that the JSON-LD
@@ -4801,9 +4807,9 @@ doing so is shown below:
       </p>
 
       <pre class="example nohighlight" title="Specifying scoped aliasing for language information">
-"statusReason": {
+"title": {
   <span class="highlight">"@context": {"value": "@value", "lang": "@language", "dir": "@direction"}</span>,
-  "@id": "https://www.w3.org/2018/credentials/examples#statusReason"
+  "@id": "https://www.w3.org/2018/credentials/examples#title"
 }
       </pre>
 
--- a/index.html
+++ b/index.html
@@ -4778,7 +4778,7 @@ text direction:
       </pre>
 
       <p>
-The next example expresses the human-readable text "revoked" in the Arabic
+The next example expresses the English text "revoked" in the Arabic
 language with a base direction of right-to-left:
       </p>
 
--- a/index.html
+++ b/index.html
@@ -4746,8 +4746,8 @@ parts of the Strings on the Web: Language and Direction Metadata document
 Data publishers are strongly encouraged to read the section on
 <a href="https://w3c.github.io/string-meta/#cross-syntax-expression">
 Cross-Syntax Expression</a> in order to ensure that the expression of
-language and base direction information is possible across multiple expression
-syntaxes such as [[JSON-LD]], [[JSON]], and CBOR [[?RFC7049]].
+language and/or base direction information is possible across multiple
+expression syntaxes such as [[JSON-LD]], [[JSON]], and CBOR [[?RFC7049]].
       </p>
 
       <p>
@@ -4813,10 +4813,10 @@ doing so is shown below:
       <h3>Complex Language Markup</h3>
 
       <p>
-When multiple languages, base directions, and annotations are used in a single
-natural language string, more complex mechanisms are typically required.
+When multiple languages, base directions, and/or annotations are used in a
+single natural language string, more complex mechanisms are typically required.
 It is possible to utilize markup languages, such as HTML, to encode
-text with multiple language and base directions. It is also
+text with multiple language and/or base directions. It is also
 possible to use the <code>rdf:HTML</code> datatype to encode such values
 accurately in JSON-LD.
       </p>
@@ -4824,17 +4824,17 @@ accurately in JSON-LD.
       <p>
 Implementers are strongly discouraged from encoding information as HTML
 because doing so 1) requires some version of an HTML processor, which
-increases the burden of processing language and base direction information,
+increases the burden of processing language and/or base direction information,
 and 2) increases the security attack surface when utilizing this data model
 as blindly processing HTML could result in executing a <code>script</code>
 tag that an attacker injected at some point during the data production process.
       </p>
 
       <p>
-Implementers considering the use of HTML to encode complex language and
+Implementers considering the use of HTML to encode complex language and/or
 base direction information might consider deconstructing the data into a
 format that does not require complex markup, such as an array of elements
-that can use a simpler form of language and base direction expression.
+that can use a simpler form of language and/or base direction expression.
 If such an approach is not possible, implementers are advised to analyze
 how an attacker would use the data to mount injection attacks against a
 consumer of the data and deploy mitigations for the identified attacks.
--- a/index.html
+++ b/index.html
@@ -4760,7 +4760,7 @@ specific base direction:
       <pre class="example nohighlight" title="Design pattern for natural language strings">
 "<a>property</a>": {
   "value": "<span class="highlight">The string value</span>",
-  "lang: "<code>LANGUAGE</code>"
+  "lang": "<code>LANGUAGE</code>"
   "dir": "<code>DIRECTION</code>"
 }
       </pre>
@@ -4774,7 +4774,7 @@ text direction:
       <pre class="example nohighlight" title="Expressing natural language text as English">
 "statusReason": {
   "value": "<span class="highlight">Revoked</span>",
-  "lang: "<code>en</code>"
+  "lang": "<code>en</code>"
 }
       </pre>
 
@@ -4786,7 +4786,7 @@ language with a base direction of right-to-left:
       <pre class="example nohighlight" title="Arabic text with a base direction of right-to-left">
 "statusReason": {
   "value": "<span class="highlight">إلغاء</span>",
-  "lang: "<code>ar</code>"
+  "lang": "<code>ar</code>"
   "dir": "<code>rtl</code>"
 }
       </pre>
--- a/index.html
+++ b/index.html
@@ -4773,7 +4773,7 @@ text direction:
       <pre class="example nohighlight" title="Expressing natural language text as English">
 "statusReason": {
   "value": "<span class="highlight">Revoked</span>",
-  "language: "<code>en</code>"
+  "lang: "<code>en</code>"
 }
       </pre>
 
@@ -4810,80 +4810,35 @@ doing so is shown below:
     </section>
 
     <section class="informative">
-       <h3>Complex Language Markup</h3>
+      <h3>Complex Language Markup</h3>
 
-       <p>
+      <p>
 When multiple languages, base directions, and annotations are used in a single
 natural language string, more complex mechanisms are typically required.
-[[!JSON-LD]] supports the expression of language information using the
-<code>rdf:HTML</code> string literal type, as described in [[!JSON-LD]],
-<a href="https://www.w3.org/TR/json-ld11/#typed-values">Section 4.2.1: Typed Values</a>.
-This mechanism enables developers to use the internationalization features
-present in HTML to express a variety of language and base direction information.
-For example, consider the following key-value pair that expresses two languages
-in the same natural language text string:
-
-       <pre class="example nohighlight" title="Expressing multiple languages in a single natural language text string">
-"nameHtml": "&lt;span <span class="highlight">lang="en"</span>>Kazumi&lt;/span> (&lt;span <span class="highlight">lang="jp"</span>>和美&lt;/span>)"
-       </pre>
-
-       <p>
-If the JSON-LD Context expressed that any value associated with the
-<code>nameHtml</code> property as being of type <code>rdf:HTML</code>, then
-software agents rendering the property can deterministically identify the
-various languages in use. JSON and CBOR implementations following the
-<code>@context</code> ordering requirements in this specification will also be
-able to benefit from this feature.
-       </p>
-
-       <p>
-This feature is also useful when mixing languages and base direction in a
-single natural language string:
-       </p>
-
-       <pre class="example nohighlight" title="Expressing multiple text directions in single natural language text string">
-"descriptionHtml": "&lt;span <span class="highlight">lang="en" dir="ltr"</span>>The country names include
-                      &lt;span <span class="highlight">lang="ar" dir="rtl"</span>>مصر&lt;/span> and
-                      &lt;span <span class="highlight">lang="he" dir="rtl"</span>>ישראל&lt;/span>.
-                    &lt;/span>"
-       </pre>
-
-       <p>
-If the JSON-LD Context states that any value associated with the
-<code>nameHtml</code> property is of type <code>rdf:HTML</code>, then software
-agents have sufficient information to deterministically identify the base
-direction of the language, even if the direction changes within the same string.
+It is possible to utilize markup languages, such as HTML, to encode
+text with multiple language and base directions. It is also
+possible to use the <code>rdf:HTML</code> datatype to encode such values
+accurately in JSON-LD.
       </p>
 
       <p>
-In some situations it is common to use
-<a href="https://en.wikipedia.org/wiki/Ruby_character">Ruby characters</a>
-to ensure text information, such as how to pronounce a particular glyph,
-is clear and as useful as possible.
-      </p>
-
-      <pre class="example nohighlight" title="Expressing Ruby markup in a natural language text string">
-"descriptionHtml": "これは<span class="highlight">&lt;ruby>&lt;rb>紙芝居&lt;/rb>&lt;rt>かみしばい&lt;/rt>&lt;/ruby></span>です。"
-      </pre>
-
-      <p>
-The markup above would result in the following rendered string in a
-software system that is capable of rendering Ruby markup:
-これは<ruby><rb>紙芝居</rb><rt>かみしばい</rt></ruby>です。
+Implementers are strongly discouraged from encoding information as HTML
+because doing so 1) requires some version of an HTML processor, which
+increases the burden of processing language and base direction information,
+and 2) increases the security attack surface when utilizing this data model
+as blindly processing HTML could result in executing a <code>script</code>
+tag that an attacker injected at some point during the data production process.
       </p>
 
       <p>
-An example of a JSON-LD Context snippet that sets a datatype of
-<code>rdf:HTML</code> is shown below:
+Implementers considering the use of HTML to encode complex language and
+base direction information might consider deconstructing the data into a
+format that does not require complex markup, such as an array of elements
+that can use a simpler form of language and base direction expression.
+If such an approach is not possible, implementers are advised to analyze
+how an attacker would use the data to mount injection attacks against a
+consumer of the data and deploy mitigations for the identified attacks.
       </p>
-
-      <pre class="example nohighlight" title="Specifying the rdf:HTML datatype">
-"descriptionHtml": {
-  "@id": "https://www.w3.org/2018/credentials/examples#descriptionHtml"
-  <span class="highlight">"@type": "rdf:HTML",</span>,
-}
-      </pre>
-
     </section>
 
   </section>
--- a/index.html
+++ b/index.html
@@ -4674,7 +4674,7 @@ aware of when publishing data described in this specification. As with any web
 standards or protocols implementation, ignoring internationalization makes it
 difficult for data to be produced and consumed across a disparate set of
 languages and societies, which would limit the applicability of the
-specification and significantly diminish its value as a standard.
+specification and significantly diminish its value as a global standard.
     </p>
 
     <p>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -4683,82 +4683,134 @@ <h2>Internationalization Considerations</h2>
...
     <section class="informative">
-      <h3>Language Declarations</h3>
+      <h3>Language and Direction</h3>
 
       <p>
-When expressing human-readable text content in the data model, it is important
-for presentation, accessibility, and further processing to specify the language
-of the text, as follows:
+Data publishers are strongly encouraged to use the design pattern described
+in this section when expressing simple internationalized strings in
```
aphillips: > ...simple internationalized strings...

I'm not sure what these are. I'd probably say either just "strings" or I would clarify that we mean human language strings, which we normally call "natural language strings".

index.html
```diff
@@ -4683,82 +4683,134 @@ <h2>Internationalization Considerations</h2>
...
 
     <section class="informative">
-      <h3>Language Declarations</h3>
+      <h3>Language and Direction</h3>
 
       <p>
-When expressing human-readable text content in the data model, it is important
-for presentation, accessibility, and further processing to specify the language
-of the text, as follows:
+Data publishers are strongly encouraged to use the design pattern described
```
aphillips: It's hard to get excited for non-normative language such as "strongly encouraged". Is that like *SHOULD*?

index.html
```diff
@@ -4683,82 +4683,134 @@ <h2>Internationalization Considerations</h2>
...
+in this section when expressing simple internationalized strings in
+JSON, JSON-LD, or similar data formats to ensure the broadest interoperability
+across syntaxes. The design pattern is a result of extensive collaboration
```
aphillips: I think this might be the wrong place for this. I18N has a document [String-Meta](https://www.w3.org/TR/string-meta) whose job it is to do this. I would be much happier to have you point at our document and we place whatever text we all (as a community or group thereof) agree to there. That document includes illustrations and use cases and other reference material not appropriate to VC's document and of general use by others in the Linked Data realm.

index.html
```diff
@@ -4683,82 +4683,134 @@ <h2>Internationalization Considerations</h2>
...
-        <li>
-[[!JSON-LD]] also supports the expression of language information using the
-<code>rdf:HTML</code> string literal type, as described in [[!JSON-LD]] 1.0,
+      <p>
+The general design pattern is to use the following markup template when
+expressing a text string that is tagged with a language and, optionally, a
+specific text direction:
+      </p>
+
+      <pre class="example nohighlight" title="Design pattern for internationalized text strings">
```
aphillips: s/internationalized text strings/natural language text strings/g ?

index.html
```diff
@@ -4683,82 +4683,134 @@ <h2>Internationalization Considerations</h2>
...
+      <pre class="example nohighlight" title="Internationalized English text">
+"statusReason": {
+  "value": "<span class="highlight">Revoked</span>",
+  "language: "<code>en</code>"
+}
+      </pre>
+
+      <p>
+The next example expresses the human-readable text "revoked" in the Arabic
+language using a right-to-left text layout:
```
aphillips: > using a right-to-left text layout

The word "layout" is wrong here. This is the "base direction", so perhaps:

> The next example expresses the natural language text "revoked" in the Arabic language with a base direction of right-to-left:

index.html
```diff
@@ -4683,82 +4683,134 @@ <h2>Internationalization Considerations</h2>
...
+}
+      </pre>
+
+      <p>
+An example utilizing the design pattern above is shown below to express the
+human-readable text "revoked" in the English language without specifying a
+text direction:
+      </p>
+
+      <pre class="example nohighlight" title="Internationalized English text">
```
aphillips: One reason I don't care for using the adjective "internationalized" is that it implies that the other strings are somehow "not internationalized" or that there is a use case for "not internationalizing" your application. Non-language-bearing strings are still internationalized: they support Unicode, for example.

index.html
```diff
@@ -4683,82 +4683,134 @@ <h2>Internationalization Considerations</h2>
...
+
+      <p>
+The next example expresses the human-readable text "revoked" in the Arabic
+language using a right-to-left text layout:
+      </p>
+
+      <pre class="example nohighlight" title="Internationalized Arabic text with a right-to-left layout">
+"statusReason": {
+  "value": "<span class="highlight">إلغاء</span>",
+  "language: "<code>ar</code>-x-d-<code>rtl</code>"
```
aphillips: It's a good idea to have a few more examples of language tags so that people understand that the more complex tags are allows. Some examples:

```
撤销  (zh-Hans-CN) [Simplified Chinese for China]
रद्द (hi-IN-x-d-ltr) [Hindi]
```

index.html
```diff
@@ -4683,82 +4683,134 @@ <h2>Internationalization Considerations</h2>
...
+      <p>
+An example utilizing the design pattern above is shown below to express the
+human-readable text "revoked" in the English language without specifying a
+text direction:
+      </p>
+
+      <pre class="example nohighlight" title="Internationalized English text">
+"statusReason": {
+  "value": "<span class="highlight">Revoked</span>",
+  "language: "<code>en</code>"
```
aphillips: Should the JSON-LD context mechanism and `@language` be used here instead? If we can just add fields, why not just add a direction metadata field (the `Localizable` dictionary in String-Meta, for example)??

index.html
```diff
@@ -4683,82 +4683,134 @@ <h2>Internationalization Considerations</h2>
...
+
+      <pre class="example nohighlight" title="Internationalized Arabic text with a right-to-left layout">
+"statusReason": {
+  "value": "<span class="highlight">إلغاء</span>",
+  "language: "<code>ar</code>-x-d-<code>rtl</code>"
+}
+      </pre>
+
+      <p class="note">
+Utilization of the design pattern above assumes that the
```
aphillips: Why not show `@language` as the example?

index.html
```diff
@@ -4683,82 +4683,134 @@ <h2>Internationalization Considerations</h2>
...
+<code>language</code> and <code>value</code>, respectively.
+      </p>
+
+    </section>
+
+    <section class="informative">
+      <h3>Multiple Languages</h3>
+
+      <p>
+When multiple languages are utilized in the same value, a more complex
```
aphillips: Still not keen on this subsection. Yes, it is sometimes necessary to provide substring language metadata. This is rarely something that can be done with normal string data, but if markup information exists, one benefit of it is allowing the HTML to be transmitted with the language (and direction) markup intact.

However, this text suggests that it `is required`, which is too strong in my opinion. I would downgrade this subsection to a note, citing the fact that if you use `rdf:HTML` you can transmit strings that have embedded language and direction markup for a richer text experience.

Note that many consumers of these strings just rip the HTML out so they can make the strings into plain text again (which is one argument for metadata).

index.html
```diff
@@ -4683,82 +4683,134 @@ <h2>Internationalization Considerations</h2>
...
-This enables text direction to be directly determined in simple cases,
-but the mechanism is insufficient for general use.
-      </p>
+      <h3>Varied Text Directions</h3>
 
       <p>
-[[!JSON-LD]] supports the expression of text layout information using the
-<code>rdf:HTML</code> string literal type, as described in [[!JSON-LD]] 1.0,
-<a href="https://www.w3.org/TR/json-ld11/#typed-values">Section 4.2.1: Typed Values</a>.
+When multiple text directions are utilized in the same value, a more
```
aphillips: This section is wrong in several ways.

First, multiple text directions may not require anything additional. The whole point of the Unicode Bidi Algorithm is that, except for certain edge cases, if you have the base direction set correctly, you can determine the display of mixed direction text.

Second, while HTML markup is preferred in a markup context, Unicode provides bidi control characters to handle the aforementioned corner cases, so `is required` is too strong.

index.html
```diff
@@ -4683,82 +4683,134 @@ <h2>Internationalization Considerations</h2>
...
+in this specification will also be able to benefit from this feature.
       </p>
 
     </section>
 
     <section class="informative">
-      <h3>Ruby Text</h3>
+      <h3>Pronunciation</h3>
       <p>
 In some situations it is common to use
```
aphillips: I think I commented on this before: this seems like a stretch that you'd be better off ignoring (leaving out altogether). In the same way that you wouldn't put text here saying "In some situations it is common to use markup to indicate emphasis and you might need the `<b>` and `<strong>` tags". I love me some Ruby markup and implementers who want to exchange data with embedded ruby markup are totally welcome to do so. But it is necessary to call it out specially?

index.html
```diff
@@ -4767,7 +4819,9 @@ <h3>Further Information</h3>
       <p>
 The W3C Internationalisation document [[string-meta]] gives more information
```
aphillips: I'd prefer an early citation to a late one?

index.html
```diff
@@ -4674,7 +4674,7 @@ <h2>Internationalization Considerations</h2>
 standards or protocols implementation, ignoring internationalization makes it
 difficult for data to be produced and consumed across a disparate set of
 languages and societies, which would limit the applicability of the
-specification and significantly diminish its value as a standard.
+specification and significantly diminish its value as a global standard.
```
aphillips: It's a personal nit, but I tend to omit qualifying adjectives like "global" and "internationalized" (you'll see another comment later to this effect) unless they have some additional meaning because it implies that there is an alternative.

index.html
```diff
@@ -4683,82 +4683,134 @@ <h2>Internationalization Considerations</h2>
...
+      <p>
+The general design pattern is to use the following markup template when
+expressing a text string that is tagged with a language and, optionally, a
+specific text direction:
+      </p>
+
+      <pre class="example nohighlight" title="Design pattern for internationalized text strings">
+"<a>property</a>": {
+  "value": "<span class="highlight">The string value</span>",
+  "language: "<code>LANGUAGE</code>-x-d-<code>DIRECTION</code>"
```
aphillips: Does this section also need to show and address the [language map](https://www.w3.org/TR/json-ld/#dfn-language-map) feature of JSON-LD?

index.html
msporny: Unfortunately, we're in the W3C Candidate Recommendation phase and don't have enough time in our charter to go through another CR phase (our charter would expire), so we're language lawyering to provide strong guidance w/o triggering another CR. It's the best we can do right now. I expect the next revision of the specification will convert parts of this section into normative language.

index.html
msporny: > I would be much happier to have you point at our document and we place whatever text we all (as a community or group thereof) agree to there

Happy to do this, let's get the language at least you are happy with in the specification, and then move it to String-Meta when it's something the WG has signed off on.

msporny: Also, we're under a time crunch, so the discussion can't take more than 1-2 weeks. I do appreciate how frustrating that is. We have WG charter pressures that we need to meet. I have no doubt we'll figure it out.

index.html
msporny: Fixed in b2f3fdb952a9ca4a96411170bb5d465334cdcad3.

index.html
msporny: Moved reference to STRING-META to the front matter, so people see it first in b2f3fdb952a9ca4a96411170bb5d465334cdcad3.

index.html
msporny: Fixed in b2f3fdb952a9ca4a96411170bb5d465334cdcad3.

index.html
msporny: Fixed in b2f3fdb952a9ca4a96411170bb5d465334cdcad3.

index.html
msporny: Rewrote the section, merged it into a "complex markup" section, removed the "is required" language. Fixed in b2f3fdb952a9ca4a96411170bb5d465334cdcad3.

index.html
msporny: I merged the Ruby bits into the "Complex Markup" section to give people an idea on when to use `rdf:HTML`, and when to use the other mechanism. I still don't mention BiDi specifically, but it feels like something that happens automatically when a developer copy-pastes text or types it in... and thus, we don't need to mention it?

index.html
msporny: Done. We mention it at the top now: b2f3fdb952a9ca4a96411170bb5d465334cdcad3

index.html
msporny: Fixed in b2f3fdb952a9ca4a96411170bb5d465334cdcad3.

index.html
msporny: We don't have enough field experience with language maps being used for Verifiable Credentials and thus I feel that it's a bit premature to provide any guidance there. We are doing some work with governments and Verifiable Credentials that cross international boundaries so will be experimenting with this over the next couple of years to see what the best mode of operation is (an array of langtag'ed content, or a language map).

One potential downside here is that the `-x-dir-` mechanism might throw a wrench into using language maps... because now you have to look for `obj.title["ar-x-dir-rtl"]` as well as `obj.title["ar-x-dir-rtl"]`. In any case, don't think we'll solve that one any time soon...

index.html
aphillips: I think this is a false urgency. It's easier than before to move back through CR than in older versions of the process and artificial dates (including charter expirations) should not gate doing the right thing. Charters have been extended for short periods of time in the past. This is particularly the case here, where the normative language is in what is effectively "advice to authors" and doesn't impact implementations.

index.html
aphillips: I don't buy the time crunch argument. The best way out of this for VC is to avoid dragging in a lot of things you don't have to solve and let us solve it in the right place (on the Linked Data level).

index.html
aphillips: I don't understand the "potential downside"?

The `-x-dir` mechanism comes at the end of the language tag, so BCP47 Lookup matching (or other matching such as ES-402 best fit) can still easily be applied. In most cases, we want any extensions to be ignored.

index.html
msporny: We have already received an administrative charter extension. We have done the math and quite literally cannot go through another CR w/o expiring our charter. The WG has already had this discussion. We can talk more about this on Wednesday if you'd like.

index.html
msporny: This is not just about Linked Data... it's about expression in JSON, CBOR, and a variety of other syntaxes. It is also about being able to canonicalize and digitally sign the data.

index.html
msporny: Can discuss more on Wed call. The downside here is that plain 'ol JSON developers will have a larger number of variations to consider if the `-x-dir-` or `-d-` mechanism is used.

msporny:
@aphillips wrote:
> b2f3fdb looks much improved. Much depends on the outcome of our discussions, of course.

You might also take a look at the more human readable version here:

https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/641.html#internationalization-considerations

index.html
```diff
@@ -4677,98 +4722,165 @@ <h2>Internationalization Considerations</h2>
...
+}
+      </pre>
 
       <p>
-When expressing human-readable text content in the data model, it is important
-to be able to explicitly encode the text direction. JSON and its derived formats
-use the UTF-8 encoding of [[Unicode]] for serializing and transmitting text.
-This enables text direction to be directly determined in simple cases,
-but the mechanism is insufficient for general use.
+The next example expresses the human-readable text "revoked" in the Arabic
```
TallTed: `human-readable text "revoked"`
-> `human-readable English text "revoked"`

index.html
```diff
@@ -4677,98 +4722,165 @@ <h2>Internationalization Considerations</h2>
...
-"nameHtml": "&lt;span dir="rtl" lang="ar">HTML و CSS: تصميم و إنشاء مواقع الويب&lt;/span>"
-      </pre>
+       <pre class="example nohighlight" title="Expressing multiple languages in a single natural language text string">
+"nameHtml": "&lt;span <span class="highlight">lang="en"</span>>Kazumi&lt;/span> (&lt;span <span class="highlight">lang="jp"</span>>和美&lt;/span>)"
+       </pre>
 
-      <p>
+       <p>
+If the JSON-LD Context expressed that any value associated with the
+<code>nameHtml</code> property as being of type <code>rdf:HTML</code>, then
```
TallTed: `as being of type`
-> `is of type`

index.html
```diff
@@ -4677,98 +4722,165 @@ <h2>Internationalization Considerations</h2>
...
+software agents rendering the property can deterministically identify the
+various languages in use. JSON implementations following the
+<code>@context</code> ordering requirements in this specification will also be
+able to benefit from this feature.
+       </p>
+
+       <p>
+This feature is also useful when mixing languages and base direction in a
```
TallTed: `languages and base direction in`
-> `languages and/or base directions in`

index.html
```diff
@@ -4677,98 +4722,165 @@ <h2>Internationalization Considerations</h2>
 specification and significantly diminish its value as a standard.
     </p>
 
+    <p>
+Implementers are strongly advised to read the Requirements for
```
r12a: Note that the current title of this document is "Strings on the Web: Language and Direction Metadata".

index.html
brentzundel: `human-readable text "revoked"`
-> `English text "revoked"`

TallTed: fine by me

index.html
msporny: We covered this in the call, and it doesn't matter, we're trying to do the right thing and are working toward that solution with the i18n community. Resolving this conversation as it's not useful wrt. the latest PR.

index.html
msporny: We covered this in the call and I think it's clear now... we're trying to do the right thing and are working toward that solution with the i18n community. Resolving this conversation as it's not useful wrt. the latest PR.

index.html
msporny: After consulting w/ @dlongley, he noted several attacks that can be pulled off by the use of HTML + Verifiable Credentials that are scary enough to rewrite that section noting the attacks. We're discouraging the use of HTML to express language and base direction unless absolutely necessary. Please re-read the section given the new, new, new complete rewrite.

index.html
msporny: Based on the calls w/ the i18n, JSON-LD, and VC communities, it looks like we're dropping x-dir- and -d- in favor of doing the right long term fix.

index.html
msporny: Rewrote the section, which dropped this text entirely.

index.html
msporny: Fixed in e544fe83ea9600947ae5f2a45925878abd6e8972.

index.html
msporny: Fixed in 87d5b6a75db500a1663418544eb8abd470716322

index.html

msporny:
Alright folks, after about 4 weeks of work and 5 complete rewrites, I think we're getting close to fixing all the i18n issues in the Verifiable Credentials Data Model specification! :P 

A number of us from the VCWG, JSON-LD WG, I18N WG, IETF langtag community, and random friends on the Internet got together and worked through a variety of the concerns here and elsewhere. The result was a new PR against the string-meta spec:

https://github.com/w3c/string-meta/pull/35

A complete rewrite of this PR (again), human-readable version here:

https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/641.html#internationalization-considerations

Please review and see if you agree with the latest, noting that the thing that is really time critical is the Verifiable Credentials spec (this PR). We have time to fine tune the string-meta spec (and of course, fix JSON-LD, RDF, etc.).

/cc @aphillips @r12a @dlongley @burnburn @chaals @iherman @gkellogg and @pchampin.

index.html
```diff
@@ -4669,105 +4714,130 @@ <h3>Data First Approaches</h3>
...
       <p>
-[[!JSON-LD]] supports the expression of text layout information using the
-<code>rdf:HTML</code> string literal type, as described in [[!JSON-LD]] 1.0,
-<a href="https://www.w3.org/TR/json-ld11/#typed-values">Section 4.2.1: Typed Values</a>.
-This mechanism enables developers to use the internationalization features
-present in HTML to express language information. For example, consider the
-following key-value pair:
-
-      <pre>
-"nameHtml": "&lt;span dir="rtl" lang="ar">HTML و CSS: تصميم و إنشاء مواقع الويب&lt;/span>"
```
chaals: This text is a more meaningful example, because if you don't have the base direction it comes out wrong.

index.html
```diff
@@ -4669,105 +4714,130 @@ <h3>Data First Approaches</h3>
...
+The next example expresses the English text "revoked" in the Arabic
```
chaals: instead of 
> expresses the English text "revoked" in the Arabic

how about something like 

> uses a term in Arabic

index.html
```diff
@@ -4669,105 +4714,130 @@ <h3>Data First Approaches</h3>
...
+  "dir": "<code>rtl</code>"
+}
       </pre>
 
       <p>
-If the JSON-LD Context states that any value associated with the
-<code>nameHtml</code> property is of type <code>rdf:HTML</code>, then software
-agents have sufficient information to deterministically identify the text
-direction of the language.
+Utilization of the design pattern above assumes that the JSON-LD
```
chaals: This paragraph is confusing especially if you're not automatically thinking about JSON-LD. I'll have a think about how to make it clearer...

chaals:
I also agree that discouraging HTML content is generally the right thing to do... (which makes me think we should be asking the HTML and Security communities to provide some guidance around how to sanitise HTML input and why...)

index.html
```diff
@@ -4669,105 +4714,130 @@ <h3>Data First Approaches</h3>
...
-but the mechanism is insufficient for general use.
+An example utilizing the design pattern above is shown below to express the
+human-readable text "revoked" in the English language without specifying a
+text direction:
       </p>
 
+      <pre class="example nohighlight" title="Expressing natural language text as English">
+"statusReason": {
+  "value": "<span class="highlight">Revoked</span>",
+  "lang: "<code>en</code>"
```
```suggestion
  "lang": "<code>en</code>"
```

index.html
```diff
@@ -4677,98 +4722,165 @@ <h2>Internationalization Considerations</h2>
...
-When expressing human-readable text content in the data model, it is important
-to be able to explicitly encode the text direction. JSON and its derived formats
-use the UTF-8 encoding of [[Unicode]] for serializing and transmitting text.
-This enables text direction to be directly determined in simple cases,
-but the mechanism is insufficient for general use.
+The next example expresses the human-readable text "revoked" in the Arabic
+language with a base direction of right-to-left:
       </p>
 
+      <pre class="example nohighlight" title="Arabic text with a base direction of right-to-left">
```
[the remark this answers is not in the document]
dlongley: My understanding (which has changed many times on this issue, so I wouldn't be surprised if it needs more updating)... is that the primary reason for adding "direction" is to handle cases where the language tag is *incorrect*  or missing but there is direction meta data that *is* correct. There is apparently a common pattern where the wrong language tag gets assigned to some data but the direction is proper and should be given preference. Because of this situation, "direction" is being added to capture this otherwise correct and dropped meta data.

For example, the text is Arabic, it is marked with an incorrect tag of "en" but the "direction" is properly "rtl".

I realize it would be odd to add an example with this erroneous information, but isn't that the point of the feature? We don't need it otherwise.

index.html
```diff
@@ -916,7 +916,16 @@ <h3>Concrete Lifecycle Example</h3>
     <span class='comment'>// identifier for the only subject of the credential</span>
     "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
     <span class='comment'>// assertion about the only subject of the credential</span>
-    "alumniOf": "&lt;span lang='en'>Example University&lt;/span>"
+    "alumniOf": {
+      "id": "did:example:c276e12ec21ebfeb1f712ebc6f1",
+      "name": [{
+        "value": "Example University",
+        "language": "en"
```
dlongley: These examples use the alias "language" instead of "lang". Which should it be?

index.html
```diff
@@ -4669,105 +4714,130 @@ <h3>Data First Approaches</h3>
...
     <p>
-There are a number of internationalization considerations implementers should be
-aware of when publishing data described in this specification. As with any web
-standards or protocols implementation, ignoring internationalization makes it
-difficult for data to be produced and consumed across a disparate set of
-languages and societies, which would limit the applicability of the
-specification and significantly diminish its value as a standard.
+There are a number of internationalization considerations implementers
+are advised to be aware of when publishing data described in this specification.
+As with any web standards or protocols implementation, ignoring
```
TallTed: `As with any web standards or protocols implementation, ignoring `
-> `As with implementation of any web standard or protocol, ignoring `

index.html
```diff
@@ -4669,105 +4714,130 @@ <h3>Data First Approaches</h3>
...
+internationalization makes it difficult for data to be produced and consumed
+across a disparate set of languages and societies, which would limit the
+applicability of the specification and significantly diminish its value as
+a standard.
+    </p>
+
+    <p>
+Implementers are strongly advised to read the Strings on the Web: Language and
+Direction Metadata document [[STRING-META]] published by the W3C
```
TallTed:
```
the Strings on the Web: Language and
Direction Metadata document
```
->
```
the <em>Strings on the Web: Language and
Direction Metadata</em> document
```

index.html
```diff
@@ -4669,105 +4714,130 @@ <h3>Data First Approaches</h3>
...
+reliable metadata about text in order to support internationalization.
+Implementers are also urged to read the Implementation Guidance
+[[VC-IMP-GUIDE]] for the latest information related to this topic.
     </p>
 
     <p>
 This section outlines general internationalization considerations to take into
-account when utilizing this data model.
+account when utilizing this data model and is intended to highlight specific
+parts of the Strings on the Web: Language and Direction Metadata document
```
TallTed: `the Strings on the Web: Language and Direction Metadata document`
->
`the <em>Strings on the Web: Language and Direction Metadata</em> document`

index.html
```diff
@@ -4669,105 +4714,130 @@ <h3>Data First Approaches</h3>
...
+When multiple languages, base directions, and/or annotations are used in a
+single natural language string, more complex mechanisms are typically required.
+It is possible to utilize markup languages, such as HTML, to encode
+text with multiple language and/or base directions. It is also
+possible to use the <code>rdf:HTML</code> datatype to encode such values
+accurately in JSON-LD.
+      </p>
+
+      <p>
+Implementers are strongly discouraged from encoding information as HTML
```
TallTed: `Implementers are`
-> `Despite that possibility, implementers are`

index.html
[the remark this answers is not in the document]
r12a: Chaals is correct to point out that the direction information is not essential for a single arabic word, such as used in the example currently.

If the word was "Revoked!"  it _would_ be necessary for the consumer to apply RTL base direction, so that the exclamation mark appears on the correct side of the arabic word. However, if the consumer is using first-strong heuristics to detect direction (which we recommend) then it can determine the direction from that, since the first letter in the arabic word is RTL.

If, however, the first letter in an arabic translation is not RTL, then the base direction needs to be indicated, since otherwise the heuristics will make the wrong decision.  So an example with some initial text in latin script would be a stronger example.

In string-meta we use:
HTML و CSS: تصميم و إنشاء مواقع الويب
which means 
HTML and CSS: Designing and Creating Websites

Btw, there's another issue which is common when dealing with examples of bidi text written in english: How do you show the arabic text in the example?  The way a user would see if correctly displayed is with "CSS و HTML" to the right of the remaining arabic text.  To achieve that in the example, you'd need to use markup under the hood.  However, then it may be less clear to the casual ready of the spec that the string begins with H.  It's not a problem with an easy solution.  Especially in Arabic, where overriding the text to show order of characters in memory produces strange joining behaviour (which is why we use Hebrew for that).

index.html
```diff
@@ -4669,105 +4714,130 @@ <h3>Data First Approaches</h3>
...
 
-    <section class="informative">
-      <h3>Further Information</h3>
       <p>
-The W3C Internationalisation document [[string-meta]] gives more information
-about the need to be able to provide reliable metadata about text to support
-internationalization.
+Implementers considering the use of HTML to encode complex language and/or
+base direction information might consider deconstructing the data into a
+format that does not require complex markup, such as an array of elements
```
r12a: This advice worries me, since it could lead people to think that they should break a sentence up into separate strings everywhere the direction changes.  That would cause problems for translation of the string (see [Working with Composite Messages](https://www.w3.org/International/articles/composite-messages/)), not to mention the fact that an overall base direction (for the whole string) is still necessary, in order to pull together the individual parts in the correct order.

I'm not sure that much can be done for language, without markup.

But for base direction, the answer for _plain text_ strings is to use the Unicode formatting characters (see [How to use Unicode controls for bidi text](https://www.w3.org/International/questions/qa-bidi-unicode-controls)).

aphillips: +1 to this.

Most "multilingual" text actually has a base language. It's true that it is possible to write a sentence that demonstrates how language identification is needed within a single string (particularly when it comes to font selection or language-specific styling). But this is a corner case and is the sort of thing that calls for markup in any event.

index.html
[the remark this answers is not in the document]
aphillips: I slightly disagree with @r12a here. The base direction is needed for other reasons besides just the bidi algorithm. For example, layout, text alignment, or bullet placement in a list might depend on the base direction. Sometimes we need base direction information even when it doesn't affect the text composition.

It's of course helpful to choose examples in which setting the direction has some observable effect, and this always means mixed direction text. The best examples start (as ours does) with a strongly "anti-directional" character and ends (as ours does not) with punctuation (`?` is ideal); even better if one can get a number and some enclosing punctuation (parentheses).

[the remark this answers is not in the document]
r12a: @aphillips, if you're referring to the first para in my comment about a single arabic word, i agree that you'd need an overall base direction to be applied by the consumer if it wasn't inserted into some other text, but the  FS heuristics would get you that base direction.

Btw, a caution about ?, since Arabic text would use ؟ U+061F ARABIC QUESTION MARK, which has the bidi category of AL - Right-to-left Arabic.  In short, it will end up in the right place when dropped into a LTR context if it follows arabic letters. It could be used with Hebrew examples, however, since Hebrew uses the european question mark.

index.html
r12a: I just published an update of the Strings on the Web doc in /TR at https://www.w3.org/TR/string-meta/.  You may want to point to that location, instead of the ED.  I'll try to update regularly using echidna when we add new text to the ED.

index.html
msporny: Replaced all "language" with "lang".

index.html
msporny: Fixed in f0b4f5063bad9633a8d0c5407fac3a2408ea4121.

index.html
msporny: @TallTed done in f0b4f5063bad9633a8d0c5407fac3a2408ea4121.

#r12a done in f0b4f5063bad9633a8d0c5407fac3a2408ea4121.

index.html
msporny: Fixed in f0b4f5063bad9633a8d0c5407fac3a2408ea4121.

index.html
msporny: Fixed in 77e1d9f39ad0b63ad95acb2de2d32c5b518fa2f6.

index.html
msporny: Fixed in 905fec016eafc8f02adca4e71ebdba2e49ef989f.

index.html
msporny: Reverted to this example in 37daadbc5dc6d3e4a165c6e1393dd27d7eadf1ef.

index.html
msporny: Attempted to do this in 9956dfa95f15d8a6823aaa672e3042f023adf0a8.

index.html
msporny: Fixed in 966b34891aba8de46db1b8c1a4e8912ec905b8a7.

index.html
msporny: Done. Removed the bad advice in ea43b2513c9d9b31b6be740a7c52c336256f570b.

msporny:
I've implemented all changes requested by reviewers. This is also in line w/ what seems to be emerging between the JSON-LD WG, i18n community, IETF folks, and the VCWG. 

Merging.

msporny:
Merged. Thanks to all for the help over the past month to get this PR to a point that was acceptable to all communities involved! :)

Issue #518, opened 2019-04-01
JSON-LD verifiable credential with multiple subjects


awoie:
> A verifiable credential contains claims about one or more subjects.

According to the current spec the data model contains a `credentialSubject` which is:
> The value of the credentialSubject property is defined as a set of objects that contain one or more properties that are each related to a subject of the verifiable credential. Each object MAY contain an id as described in Section § 4.2 Identifiers.

Could you please provide an example of a verifiable credential containing multiple subjects.

David-Chadwick:
```
...
"type": ["VerifiableCredential", "MarriageCertificate"],
  "credentialSubject": [{
    "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
    "marriedTo": "did:example:ebc276e12ec21feb1f712ebc6f1"
  },{
    "id": "did:example:ebc276e12ec21feb1f712ebc6f1",
    "marriedTo": "did:example:ebfeb1f712ebc6f1c276e12ec21"
  }],
....
```

awoie:
@David-Chadwick Thank you. In that case I have still another question.

The VC spec contains the following example:
```
{
  ...
  "credentialSubject": {
    "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
    "degree": {
      "type": "BachelorDegree",
      "name": "<span lang='fr-CA'>Baccalauréat en musiques numériques</span>"
    }
  },
  ...
}
```
The `credentialSubject` is defined as a:
> set of objects that contain one or more properties

Note that in JSON a set of properties is mapped onto `Object type`.

The example above is not an **array** of properties either. Is the example wrong, or is this just JSON-LD magic?

awoie:
@dlongley @msporny could you please clarify?

msporny:
> @dlongley @msporny could you please clarify?

Yes. In JSON-LD, all properties can have one or more "objects" associated with them. So, 

```
"foo": { ... }
```

is effectively shorthand for:

```
"foo": [{ ... }]
```
 
So, all of the examples are correct.

In practice, one can use a `credentialSchema` to further restrict the JSON-LD and JSON forms so that you force developers to always use array form, for example.

In the spec, the data model states "a set of objects". A serialization can choose to use the shorthand if there is only one object (and then enforce that there is always only one object by using `credentialSchema`).

Does this answer your question @awoie? If so, can we close this issue?

awoie:
[...]

PR #644, opened 2019-05-26
Add multiple subjects example.

msporny:
Add an example demonstrating how multiple subjects can be associated with the `credentialSubject` property.

Related to #518.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/644.html" title="Last updated on Jun 8, 2019, 3:50 PM UTC (ac18b4d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/644/446c421...ac18b4d.html" title="Last updated on Jun 8, 2019, 3:50 PM UTC (ac18b4d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1467,9 +1467,8 @@ a set of objects that contain one or more properties that are each related to a
 
         <p>
 It is possible to express information related to multiple <a>subjects</a> in a
-<a>verifiable credential</a>. The example below specifies two subjects,
-the first is a recipient of a Bachelor Degree, the second is the university
-that granted the degree. Note the use of array notation to associate
+<a>verifiable credential</a>. The example below specifies two subjects that are
+spouses. Note the use of array notation to associate
 multiple subjects with the <code>credentialSubject</code> property.
         </p>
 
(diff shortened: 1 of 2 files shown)


David-Chadwick:
I would prefer a marriage certificate to be the example of multiple subjects in a VC. 
This is already mentioned in the Note under Figure 6 as being a typical example, so this Note could refer to the example later on. 
(The university degree example seems rather contrived in comparison to a marriage certificate).

awoie:
@msporny That example is interesting. In uPort, we call them issuer credentials and they can be included in the credential itself which is an array of JWTs about the issuer. Technically, it would be possible to define processing rules for a JWT containing issuer credentials to be serialized as a credential with multiple `credentialSubjects`. Do you think we can add that to the implementation guidance?

index.html
```diff
@@ -1465,6 +1465,37 @@ <h3>Credential Subject</h3>
 }
         </pre>
 
+        <p>
+It is possible to express information related to multiple <a>subjects</a> in a
+<a>verifiable credential</a>. The example below specifies two subjects,
+the first is a recipient of a Bachelor Degree, the second is the university
```
TallTed: a little punctuation tweak here.  Also, note that `bachelor's degree` is the generic, compared to `Bachelor of Science Degree in Biology`, `Bachelor of Arts Degree`, or similar for the specific.
```
specifies two subjects,
the first is a recipient of a bachelor's degree, the second
```
-> 
```
specifies two subjects;
the first is a recipient of a bachelor's degree, and the second
```
or -> 
```
specifies two subjects:
the first is a recipient of a bachelor's degree; the second
```

msporny: @TallTed I'm going to simplify this example to a spousal relationship. Were I not to do that, I would have pulled your suggestion in.

msporny:
> @msporny That example is interesting. In uPort, we call them issuer credentials and they can be included in the credential itself which is an array of JWTs about the issuer. Technically, it would be possible to define processing rules for a JWT containing issuer credentials to be serialized as a credential with multiple credentialSubjects. Do you think we can add that to the implementation guidance?

Yes, I think we should add that to the implementation guidance (how to specify credentials about the issuer, like if they're licensed to issue the particular VC they're issuing).

msporny:
> Replace with marriage certificate

Ok, I've updated the example to express a spousal relationship while being considerate of the following points:

* Avoiding calling it a marriage certificate (as some parts of society are not allowed to marry and/or reject the notion of the need to officially recognize the relationship).
* Using statistically gender neutral names.
* Including more than just the bidirectional relationship to make the credential make more sense (by including a name)... understanding that it creates a point of correlation in the example.

msporny:
Merging as all requests have been responded to and/or applied in a way that I don't think will result in objections to merging. Scream if you disagree.

Issue #584, opened 2019-04-30
misleadingly and confusingly named issuanceDate


TallTed:
Regrettably timed, but I only just noticed this.  (previous discussion was in #184)

`issuanceDate` (20,500 Google hits) is a rather odd construct, especially alongside `issuer` (27,500,000 Google hits) and `expirationDate` (1,220,000 Google hits).

The differences in scale of use is even more pronounced if you search for these as phrases, e.g., "issuance date" (736,000 hist) and "expiration date" (41,600,000 hits), rather than as the compound terms.

I know it would cause code changes for all implementors including the test suite, but this would be *much* better changed to `issueDate` (325,000 Google hits) or `issuedOn` (11,500 Google hits) -- potentially also changing `issuer` (27,500,000 Google hits) to `issuedBy` (472,000 Google hits).

(Had I noticed sooner, I would have pushed for `issuedAt` and `expiresAt` which I think also better conveys the datetime nature of these attributes.)

This may get closed out-of-hand as too late, but I felt it better to speak now than not at all.

msporny:
When selecting terms for most Linked Data libraries, we attempted to pick the words based on their common dictionary definition rather than how popular the term is on Google. Google term popularity is one data point, but more importantly, the meaning of the term when searched for in a common dictionary is far more important. Here's what comes up for Merriam-Webster:

> issuance: 1) the action of supplying or distributing something, especially for official purposes, 2) the action of formally making something known.

> expiration: 2)  the fact of coming to an end or the point at which something ends.

I agree that "Date" isn't great (the fields started as pure dates w/ no time, but things changed over time w/ the definitions w/o the property names being updated), but suggest that the two words that we have at present are accurate words to describe the properties.

A change of these property names would require implementations change and would require the test suite to change. That's a substantive change, and our charter extension is such that it would be incredibly difficult to go through another CR at this point... even if that were not the case, this alone doesn't feel like a compelling enough reason to go through a CR. Yes, yes, we shouldn't let W3C Process affect doing "The Right Thing(tm)", but that's where we are today... and if we were to have this discussion, the presumption is that the group would agree to "issuedAt" and "expiresAt" and "issuedBy" without having to result to an Instant Run Off naming poll where we collect all options, run the poll, keep it open for 1-2 weeks, for a total of 4 weeks of time spent on this one issue (which is what these things typically devolve into).

David-Chadwick:
[...]

PR #646, opened 2019-05-26
Add notes for validFrom and validUntil.

msporny:
Reserve validFrom and validUntil, note expected changes in next version of spec.

Related to #584.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/646.html" title="Last updated on Jun 8, 2019, 5:01 PM UTC (c364611)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/646/d31adef...c364611.html" title="Last updated on Jun 8, 2019, 5:01 PM UTC (c364611)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1588,12 +1588,10 @@ associated with the <code>credentialSubject</code> <a>property</a> becomes valid
         </pre>
 
         <p class="note">
-It is expected that the next version of this specification will rename the
-<code>issuanceDate</code> <a>property</a> to <code>validFrom</code> in a way
-that is backwards-compatible. The range of acceptable values will remain the
-same as will the semantic meaning of the <a>property</a>. Implementers are
-advised that the <code>validFrom</code> <a>property</a> is reserved and its use
-for any other purpose is discouraged.
+It is expected that the next version of this specification will add the
+<code>validFrom</code> <a>property</a>. Implementers are advised that the
+<code>validFrom</code> <a>property</a> is reserved and its use for any other
+purpose is discouraged.
         </p>
 
       </section>
@@ -1723,12 +1721,11 @@ date and time the <a>credential</a> ceases to be valid.
         </pre>
 
         <p class="note">
-It is expected that the next version of this specification will rename the
-<code>expirationDate</code> <a>property</a> to <code>validUntil</code> in a way
-that is backwards-compatible. The range of acceptable values will remain the
-same as will the semantic meaning of the <a>property</a>. Implementers are
-advised that the <code>validUntil</code> <a>property</a> is reserved and its use
-for any other purpose is discouraged.
+It is expected that the next version of this specification will add the
+<code>validUntil</code> <a>property</a> in a way that deprecates, but
+preserves backwards compatability with the <code>expirationDate</code>
+<a>property</a>. Implementers are advised that the <code>validUntil</code>
+<a>property</a> is reserved and its use for any other purpose is discouraged.
         </p>
 
       </section>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -1557,6 +1557,16 @@ <h3>Issuance Date</h3>
...
 }
         </pre>
+
+        <p class="note">
+It is expected that the next version of this specification will rename the
+<code>issuanceDate</code> <a>property</a> to <code>validFrom</code> in a way
+that is backwards-compatible. The range of acceptable values will remain the
+same as will the semantic meaning of the <a>property</a>. Implementers are
+advised that the <code>validFrom</code> <a>property</a> is reserved and its use
+for any other purpose is discouraged.
```
dlongley: I think there's a slight problem here for JWT representation. We current express `issuanceDate` as the ["iat" claim](https://tools.ietf.org/html/rfc7519#section-4.1.5) in a JWT. The `validFrom` property should instead map to the ["nbf" claim](https://tools.ietf.org/html/rfc7519#section-4.1.6) in a JWT.

It is true that we intended "issuanceDate" to *mean* `validFrom` in the absence of that property. However, I think *both* properties could be present in a VC with different values and, under that circumstance, `validFrom` should take precedence for determining validity. So `issuanceDate` means the earliest point at which a VC should be considered valid, unless `validFrom` is present. Of course, we can't say that normatively now and we're just reserving this for the future.

So I think it's still true that the range of acceptable values will remain the same and the semantic meaning of the property will be the same but this will not be a "renaming", rather, the presence of `validFrom` would take precedence over `issuanceDate` when determining the VC validity period. Also, as mentioned above, mapping these properties to a JWT would result in using different JWT claims.

cc: @TallTed

TallTed: Good catch.  I think that in the current iteration, we must express `issuanceDate` as `nbf` (as this is and was the intended semantic), and not as `iat` (which was never the intended semantic).  Future versions of the VCDM can change things however is appropriate at that point, including how precedence lays out if/when both are present.

msporny: Ok, updating the spec w/ these changes in mind.

msporny: Fixed in afaedca614da265795c894714bee97b805dd48e6

msporny:
Need a re-review on this one as it too complicated to just merge on its own.

msporny:
@dlongley and @TallTed, can both of you provide a re-review please?

TallTed:
I think associated changes (which I think would be best made in the same merge) are also needed to the JWT mapping section, particularly (in the [msporny-issue-584](https://github.com/w3c/vc-data-model/tree/msporny-issue-584) branch) the following, which are now focused on JWT `exp` and `iat` and do not mention `nbf` --
* [lines 3443-3444](https://github.com/w3c/vc-data-model/blob/c364611ba7c85c2868d19f61bf5c614347306904/index.html#L3443)
* [lines 3450-3451](https://github.com/w3c/vc-data-model/blob/c364611ba7c85c2868d19f61bf5c614347306904/index.html#L3450)
* [lines 3524-3527](https://github.com/w3c/vc-data-model/blob/c364611ba7c85c2868d19f61bf5c614347306904/index.html#L3524)
* [lines 3534-3536](https://github.com/w3c/vc-data-model/blob/c364611ba7c85c2868d19f61bf5c614347306904/index.html#L3534)

TallTed:
As it seems my review is the blocker here...

I cannot approve this without [the changes to the associated areas](https://github.com/w3c/vc-data-model/pull/646#issuecomment-500450041) that I highlighted previously.

I think these would count as bugfix, though they would change normative text, and should therefore not trigger a new PR.

stonematt:
PR was merged to make way for the resolution from WG call on 18 June 19: 
RESOLVED: Change spec text mapping of issuanceDate to use JWT nbf claim (with matching semantics) instead of JWT iat claim (with conflicting semantics) (https://github.com/w3c/vc-data-model/pull/646), in parallel to testsuite change to match (https://github.com/w3c/vc-test-suite/pull/50)

Issue #589, opened 2019-05-03
JSON schema reference


David-Chadwick:
The reference to JSON schema is way out of date. The Internet Draft expired in 2013. We should instead point to 
https://json-schema.org/
which is the home of JSON schema and provides up to date information

msporny:
Looks like this is the latest for JSON Schema at IETF: https://tools.ietf.org/html/draft-handrews-json-schema

W3C Process frowns upon pointing to websites that might disappear when there is an appropriate standards organization document that we could point to instead.

David-Chadwick:
There is a problem pointing to IETF drafts in that they expire in 6 months time. So any reference to them is bound to be soon out of date, which is not good for a Recommendation that will last for years. Referencing an RFC is different since these typically last for years.
So I would prefer us not to reference an IETF draft. I think https://json-schema.org/ is going to have a reasonably long life (it has so far)

msporny:
We should point to the IETF page and submit a PR that does that.

msporny:
PR is in #647. Suggest close after that is merged.

David-Chadwick:
Agreed

burnburn:
WG resolution on 28 May 2019 call:

RESOLUTION: Issue #589 is addressed via PR #647 which updates the non-normative reference to JSON Schema to a more stable IETF URL. Issue #589 should be closed after the PR is merged.

burnburn:
PR merged long ago, submitter agreed.  Closing.

PR #647, opened 2019-05-26
Add latest reference to JSON Schema at IETF.

msporny:
This uses the latest reference to JSON Schema at IETF.

Related to #589.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/647.html" title="Last updated on Jun 8, 2019, 5:07 PM UTC (30304d7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/647/446c421...30304d7.html" title="Last updated on Jun 8, 2019, 5:07 PM UTC (30304d7)">Diff</a>

### DIFF
--- a/common.js
+++ b/common.js
@@ -138,6 +138,13 @@ var vcwg = {
       title: 'InterPlanetary File System (IPFS)',
       href: 'https://en.wikipedia.org/wiki/InterPlanetary_File_System',
       publisher: 'Wikipedia'
+    },
+    'JSON-SCHEMA-2019': {
+      title: 'JSON Schema: A Media Type for Describing JSON Documents',
+      href: 'https://tools.ietf.org/html/draft-handrews-json-schema',
+      authors: ['Austin Wright', 'Henry Andrews'],
+      status: 'Internet-Draft',
+      publisher: 'Internet Engineering Task Force (IETF)'
     }
   }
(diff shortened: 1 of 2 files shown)


common.js
```diff
@@ -138,6 +138,13 @@ var vcwg = {
       title: 'InterPlanetary File System (IPFS)',
       href: 'https://en.wikipedia.org/wiki/InterPlanetary_File_System',
       publisher: 'Wikipedia'
+    },
+    'JSON-SCHEMA-2019': {
```
TallTed: One niggle occurs to me. That draft was issued in March 2018 and expired in September 2018, and we're less than halfway through 2019.  Perhaps the reference `JSON-SCHEMA-2019` should instead be `JSON-SCHEMA-2018` (as there may yet be a 2019-dated revision)?  (I don't know whether these are strictly local identifiers, or if there's some desired persistence/consistency/reuse across W3 or other docs.)

burnburn: The identifiers in this file are local but often get migrated to the specref database.  I agree with considering a JSON-SCHEMA-2018.

msporny: Fixed in 7175fad7f53eaf29ae26051b8bb701e984649e4e.

Issue #653, opened 2019-05-28
The confusing nature of type and presentation


brentzundel:
The text in the spec is unclear about whether it is required for presentations and encapsulated objects to have multiple types. My understanding is that the type field for a verifiable credential MUST have more than one type, but that presentations, encapsulated objects, etc. do not have this requirement.
I recommend text be added to clarify that while `type` is required for presentations and encapsulated objects, it is not required for those objects to have multiple types. Examples in the spec should also be changed to reflect this.

Related to issue #600

burnburn:
From the VCWG call on 11 June 2019:

RESOLUTION: Merge PR #658 as it reflects the consensus of the Working Group. Specifically, more organizations would object to more stringent type requirements than what exists in PR 658.
[10:37] manu: I'll merge that as soon as Brent makes the minor change
[10:38] burn: that resolution only applies after a change has been made.

msporny:
The minor change was made in https://github.com/w3c/vc-data-model/pull/658/commits/12681f9a24bb7669e0f23268b16ef4d318b172c2 and was merged into the specification 9 days ago. Please close.

PR #658, opened 2019-06-05
Clarify type description

brentzundel:
Added text clarifying when more than base types are needed for verifiable credentials, verifiable presentations, and embedded objects; and fixed examples of verifiable presentations.
Resolves issue #653 

***
<a href="https://pr-preview.s3.amazonaws.com/brentzundel/vc-data-model/pull/658.html" title="Last updated on Jun 11, 2019, 4:44 PM UTC (12681f9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/658/d18642b...brentzundel:12681f9.html" title="Last updated on Jun 11, 2019, 4:44 PM UTC (12681f9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -956,7 +956,7 @@ which is then composed into a <a>verifiable presentation</a>. The
     "https://www.w3.org/2018/credentials/v1",
     "https://www.w3.org/2018/credentials/examples/v1"
   ],
-  "type": ["VerifiablePresentation", "CredentialManagerPresentation"],
+  "type": "VerifiablePresentation",
   <span class='comment'>// the verifiable credential issued in the previous example</span>
   "verifiableCredential": [{
     "id": "http://example.edu/credentials/1872",
@@ -1377,7 +1377,8 @@ of this specification who want to support interoperable extensibility, do.
         </p>
 
         <p>
-All <a>credentials</a>, <a>presentations</a>, and encapsulated objects MUST
+If <a>credentials</a>, <a>presentations</a>, and encapsulated objects have
+properties beyond those contained in their respective base types, they MUST
 specify, or be associated with, additional more narrow <a>types</a> (like
 <code>UniversityDegreeCredential</code>, for example) so software systems can
 process this additional information.
@@ -2611,7 +2612,7 @@ in an archive.
     "https://www.w3.org/2018/credentials/examples/v1"
   ],
   "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
-  "type": ["VerifiablePresentation", "CredentialManagerPresentation"],
+  "type": ["VerifiablePresentation", "TermsOfUsePresentation"],
   "verifiableCredential": [{
     "id": "http://example.edu/credentials/3732",
     "type": ["VerifiableCredential", "UniversityDegreeCredential"],
@@ -2909,7 +2910,7 @@ that the <a>holder</a> did not intend to share.
     "https://www.w3.org/2018/credentials/v1",
     "https://www.w3.org/2018/credentials/examples/v1"
   ],
-  "type": ["VerifiablePresentation", "CredentialManagerPresentation"],
+  "type": "VerifiablePresentation",
   "verifiableCredential": [
     {
         "@context": [
@@ -3583,7 +3584,7 @@ header parameter.
       "https://www.w3.org/2018/credentials/v1",
       "https://www.w3.org/2018/credentials/examples/v1"
     ],
-    "type": ["VerifiablePresentation", "CredentialManagerPresentation"],
+    "type": ["VerifiablePresentation"],
     <span class="comment">// base64url-encoded JWT as string</span>
     "verifiableCredential": ["..."]
   }
@@ -5492,8 +5493,8 @@ verifiable credential that was passed to it by the subject">
     "https://www.w3.org/2018/credentials/examples/v1"
   ],
   "id": "did:example:76e12ec21ebhyu1f712ebc6f1z2",
-  "type": ["VerifiablePresentation", "CredentialManagerPresentation"],
-  "credential": [
+  "type": ["VerifiablePresentation"],
+  "verifiableCredential": [
     {
      "@context": [
        "https://www.w3.org/2018/credentials/v1",
--- a/index.html
+++ b/index.html
@@ -1377,11 +1377,14 @@ of this specification who want to support interoperable extensibility, do.
         </p>
 
         <p>
-If <a>credentials</a>, <a>presentations</a>, and encapsulated objects have
-properties beyond those described in their respective base types, they MUST
-specify, or be associated with, additional more narrow <a>types</a> (such as
+All <a>credentials</a>, <a>presentations</a>, and encapsulated objects MUST
+specify, or be associated with, additional more narrow <a>types</a> (like
 <code>UniversityDegreeCredential</code>, for example) so software systems can
-process this additional information.
+process this additional information. The specification of these additional more
+narrow types could be done in the object within which the additional information
+is found, rather than at a higher level such as the <code>type</code>
+<a>property</a> of the <a>credential</a> or <a>presentation</a>. It is also
+possible for the additional more narrow types to be implicit.
         </p>
 
         <p>
--- a/index.html
+++ b/index.html
@@ -2615,7 +2615,7 @@ in an archive.
     "https://www.w3.org/2018/credentials/examples/v1"
   ],
   "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
-  "type": ["VerifiablePresentation", "TermsOfUsePresentation"],
+  "type": ["VerifiablePresentation"],
   "verifiableCredential": [{
     "id": "http://example.edu/credentials/3732",
     "type": ["VerifiableCredential", "UniversityDegreeCredential"],
--- a/index.html
+++ b/index.html
@@ -1378,8 +1378,8 @@ of this specification who want to support interoperable extensibility, do.
 
         <p>
 If <a>credentials</a>, <a>presentations</a>, and encapsulated objects have
-properties beyond those contained in their respective base types, they MUST
-specify, or be associated with, additional more narrow <a>types</a> (like
+properties beyond those described in their respective base types, they MUST
+specify, or be associated with, additional more narrow <a>types</a> (such as
 <code>UniversityDegreeCredential</code>, for example) so software systems can
 process this additional information.
         </p>


index.html
```diff
@@ -1377,8 +1377,9 @@ <h3>Types</h3>
         </p>
 
         <p>
-All <a>credentials</a>, <a>presentations</a>, and encapsulated objects MUST
-specify, or be associated with, additional more narrow <a>types</a> (like
+If <a>credentials</a>, <a>presentations</a>, and encapsulated objects have
+properties beyond those described in their respective base types, they MUST
+specify, or be associated with, additional more narrow <a>types</a> (such as
 <code>UniversityDegreeCredential</code>, for example) so software systems can
 process this additional information.
```
TallTed: Hmmm....
```
If software systems are expected to process information in properties of
<a>credentials</a>, <a>presentations</a>, or encapsulated objects beyond the
properties described in their respective base types, such <a>credentials</a>,
<a>presentations</a>, or encapsulated objects MUST specify, or be associated
with, additional more narrow <a>types</a> in which those additional properties
are described (for example, <code>UniversityDegreeCredential</code>). Lacking
this additional type association or specification, such additional properties
MAY be ignored by systems which process these <a>credentials</a>,
<a>presentations</a>, or encapsulated objects.
```
?

brentzundel: My change makes clear that additional types are only a MUST if there are additional properties not currently defined by a type, thereby softening the requirement in cases where an additional type is nonsensical.
Your recommendation softens the type requirement further than I like, essentially changing it from a MUST to a MAY. It also mentions expected processing, which I am not comfortable adding to the spec.

TallTed: I do not believe I softened anything, though my language appears to have made some things clearer, and exposed an existing softness with which you apparently disagree.

We already have (and you preserved) this great big handwave of "or be associated with" within the MUST, which removes the need to explicitly declare a sub-type within the object at hand, making it that explicit declaration a MAY.

To my knowledge, we do not require that every Validator must be able to understand every property within every VC/VP, nor that every Validator must associate every extra-property-bearing VC/VP with a sub-type of its declared base type -- only effectively that such association be possible in the abstract (i.e., that such a sub-type must be described *somewhere* [and implicitly that such sub-type must be understood by the Issuer and by *some* but not necessarily *all* potential Validators] but not that such description be accessible by any-and-all potential Validators).  Even there, I have not understood this to be a requirement for *every* property but for every *material* property, which materiality can only be determined by the Validator.

Essentially, we've said that _Validators that care_ about the extra property must be able to associate the VC/VP with a sub-type description which describes that property. In other words, Validators must be able to access a property description _for every property with which they're concerned,_ via a description of an associated type for every VC/VP they're processing *which type association **MAY** be declared within the VC/VP, or which subtype association **MAY** be inferred based on the presence of some property/ies within the VC/VP which are not described by the declared base type.*

David-Chadwick: I appreciate that we have a tension between:
- verifiers that do not particularly care about additional properties and will accept any VC regardless of their additional properties and regardless of any constraints the issuer may have placed on the VC, and 
- verifiers that do care and want to understand every property in a VC otherwise they will not accept it.
I think that the former should be outside the scope of the standard, otherwise what is the point of standardising features that the former will happily ignore.
I think the standard should cater for the later, the well behaved eco-system citizens. Thus I do not like @TallTed's addtional text as this is pandering to the former gung-ho citizens.
So to address @TallTed's point "we do not require that every Validator must be able to understand every property within every VC/VP" I think the standard MUST provide mechanisms so that every verifier is able to understand every property in every VC/VP. Then those verifiers that do not care can ignore some properties, but those verifiers that do care can understand every property.

dlongley: Whatever we decide here, we should remember that there may be verifiers that understand particular properties (e.g. "ageOver") that do not care or know about particular VC subtypes.

TallTed: > pandering to the former gung-ho citizens

There is an unfortunate tone here. Frankly, to my eyes, it changes the rest of your comment from a technical argument to a religious one.

To the technical --

* A Verifier whose policy is that it must understand all properties in every VC/VP is free to reject any VC/VP that lacks the explicit sub-type declarations which would deliver that understanding for otherwise unrecognized properties.

* A Verifier who only cares about the properties of the base type is free to accept any VC/VP of that base type, regardless of any additional properties and/or sub-type declaration.

* A Verifier who cares about some specific sub-type is free to require either the explicit declaration _(which remains optional!)_ or the specific additional property/ies (which allow the VC/VP to be associated with that sub-type without any explicit declaration) of that sub-type.

* An Issuer who knows about a sub-type may include the additional property/ies required by that sub-type, with or without explicit declaration of that sub-type, which might have a URI which is only dereferenceable by agents doing so from within a particular IP address range or DNS (sub)domain.

An example...  MyCollege issues a VC about me.  MyCollege includes only the base type declaration (Credential), with properties for the degree name (Bachelor of Science in Biology), honors (Summa Cum Laude), sub-discipline (Reptiles), sub-sub-discipline (Snakes).  The obvious extra properties are implicitly associated with MyCollege's BachelorDegreeDetailed VC type -- which it does not describe outside the university network -- and the BachelorDegree is not sub-typed at all -- even to say it is a DegreeCredential.

The current interpretation suggested by @brentzundel and @David-Chadwick  (which does matches neither the old nor new wording, as I read things) is that this VC is invalid, because it does not explicitly declare its sub-type -- even though it can "be associated with" that sub-type by both its Issuer and any Verifier who knows about MyCollege's sub-type.

brentzundel:
The goal of this PR is to address issue #653, and to answer the question "Must a verifiable presentation have more than one subtype?" 
Since the only extension point within a verifiable presentation which must introduce some other object is the `verifiableCredential` property, and since a verifiable credential already must have multiple types (thereby defining the additional properties in the verifiable credential),   a verifiable presentation that has only the type `verifiablePresentation` would have all of its properties sufficiently defined. Therefore it does not make sense to require that a verifiable presentation have multiple types.

Please judge this PR on the merits of how well it addresses the issue at hand. If there are additional clarifications that others desire to make regarding types, please raise a separate issue for those and raise a separate PR to address them.

TallTed:
@brentzundel - The text at hand does not only discuss verifiable presentations.  It also discusses verifiable credentials and objects which might be encapsulated in either.

Indeed, your comment in issue #653 says (emphasis mine) --

> I recommend text be added to clarify that **_while type is required for presentations and encapsulated objects, it is not required for those objects to have multiple types._** Examples in the spec should also be changed to reflect this.

-- which is not reflected in your suggested text (nor in my expanded rewording of that text).

David-Chadwick:
My opinion is that a VP that only contains embedded VCs (as in Example 12) is a basic VP type (i.e. "type": ["VerifiablePresentation"]) and no subtype is needed. But if the VP contains other properties such as restrictive terms of use, then this needs to be signaled either in the VP type, or the ToU type.

Lets first agree the principles, then wordsmith to conform to them.

brentzundel:
## intro 
I feel like this discussion has gotten out of hand, and would like first to apologize if my comments have caused anyone to be upset. That was not my intention and I am sorry.

I would like to lay out my understanding of the text in question. I assume that the text has been interpreted differently by others. My hope is that we can come to a common understanding of what the original text means, which will enable us to better discuss how best to clarify it.

## The original text of the specification:
```
All credentials, presentations, and encapsulated objects MUST specify, or be associated with, additional more narrow types (like UniversityDegreeCredential, for example) so software systems can process this additional information.
```

## My interpretation of the text
### One of these conditions must hold:
1. All credentials, presentations, and encapsulated objects MUST specify additional more narrow types
2. All credentials, presentations, and encapsulated objects MUST be associated with additional more narrow types

### "additional information" means "properties not defined in the spec"
My interpretation of the original text is that "this additional information" means "the properties inside of the credential, presentation, or encapsulated object that aren't described in `verifiableCredential` or `verifiablePresentation`" 

### intent of the original text
I believe the requirement to have additional more narrow types was intended to ensure that  these properties were described somewhere "so software systems can process this additional information."

### location of more narrow types
My understanding is that either
1. The more narrow types would need to be directly specified (condition 1)
2. The more narrow types should be specified in the credential or presentation within which the "additional information" is found, but may be at a higher level than the object in which the additional information is encapsulated, but they must be specified. That is, the encapsulated objects within the credential do not need to independently specify their types because the verifiable credential already has the additional more narrow type `UniversityDegreeCredential`. The additional information in the object is already associated with `UniversityDegreeCredential`

What is not clear to me, is how an encapsulated object may be considered associated with a type that is not explicitly specified. 

## why presentations may be problem
If my understanding is correct, and the types with which an encapsulated object is associated MUST be in a higher level object, then we have a requirement for presentations (to have an additional more narrow type) that is against the original intent of the text, because a presentation is always the highest level object. 

### why this may not be a problem
This may not actually be a problem if the "Additional more narrow types" may be specified in the objects within which the additional information is found. This would mean that a verifiable presentation could rest easy with a single type, because the encapsulated verifiable credential already specifies the additional more narrow types needed for its additional information.

## My hope for moving forward
If there is a particular aspect of my interpretation of the original text which does not match your interpretation, please let me know. I am looking to understand your point of view, and am open to being wrong.

I hope that we can come to a common interpretation of the text. I will then adjust the language in my pull request to clarify it according to our common understanding.

index.html
David-Chadwick: @dlongley > Whatever we decide here, we should remember that there may be verifiers that understand particular properties (e.g. "ageOver") that do not care or know about particular VC subtypes.

Without knowing the context of a property (which is indicated by the type of the VC), then in general it will make little sense to a verifier to simply know that an isolated property is present. What is this the ageOver property the age over of? Is it the subject, a relative of the subject, or a property of the subject, or a property of some other property etc. (Thought experiment. Replace all additional property names (except ageOver and the base properties that all VCs must contain) by XXXX in any arbitrary VC, and vary the position of ageOver between these XXXX properties, and then see of you can determine what ageOver really means. Clearly you cannot. This is why the type property is essential to understanding the VC.

TallTed:
@David-Chadwick 

> if the VP contains other properties such as restrictive terms of use

Extra properties may have *no* material meaning except to the Issuer -- or possibly to the Holder in the case of a Presentation they've assembled.  Leaping to an example like "restrictive terms of use" is making a big assumption about how implementers/users may proceed in the wild.

If the property has no meaning to the Verifier, has no impact on their decision making -- there is no justification for a VC/VP sub-type.

That said, the current text does not make any distinction between active and inactive (a/k/a material and immaterial) properties.  It says *any* property which is not in the base type description forces a sub-type association -- though it allows that association to be implicit (which it is clear, you would not) through the long-existing text, "or be associated with" (and thereby allows for a "private" sub-type which is only known to the Issuer and their chose "insider" Verifiers).

As I've said several times (and as was part of the origin of that phrase in the spec), such implicit association may be deduced by the Verifier *based on the presence of those properties, without any explicit sub-type declaration.*

TallTed:
@brentzundel - 

As you say, "so software systems can process this additional information" is a key phrase of the original text. The reason for the type and subtype declarations is to allow for understanding of the properties by software.

If some properties are included but not meant to be interpreted by software which doesn't already understand those properties, there's no need for the sub-type -- the software (if any) which *is* meant to understand those properties already understands them -- and if they are not meant to be interpreted by software, then again, understanding doesn't matter.

I believe [my rephrasing above](https://github.com/w3c/vc-data-model/pull/658#discussion_r290896178) allows for all of this, preserves the original intent, and clarifies both what must be done and the reason for that, including the main ramification of leaving out the explicit sub-type declaration (i.e., those extra properties may be ignored).  That said, it may be appropriate to change "ignored" to "ignored or misinterpreted".

brentzundel:
@TallTed I have made some changes based on a better understanding of the original intent of the text in (b3c7f40)[https://github.com/w3c/vc-data-model/pull/658/commits/b3c7f403951446a7262d06f388d60e1647212085]
Please let me know if this works better.

David-Chadwick:
@TallTed My recollection of using the phrase "associated with" was to allow type information of nested objects to be placed on higher objects such as the VC object, rather than forcing each object to contain its type. It was not so that typing could be implicit.

brentzundel:
@David-Chadwick that was my interpretation of the text as well, but I can see potential value in the notion of implicit sub-types as @TallTed has introduced them.

David-Chadwick:
@brentzundel said "What is not clear to me, is how an encapsulated object may be considered associated with a type that is not explicitly specified." This is also not clear to me, and I think the answer is "it cannot".
@TallTed Could you please give me a concrete example of a property that has no meaning to the Verifier, has no impact on its decision making, and yet the issuer feels that it is important enough to place this property in the VC.
@TallTed The part of your argument that I disagree with, if I have correctly understood it, is this "if some properties are included but not meant to be interpreted by software which doesn't already understand those properties, there's no need for the sub-type -- the software (if any) which is meant to understand those properties already understands them". This sentence implies that the issuer and the verifier have a private agreement, that is outside the scope of the standard, in which they agree which properties will be present in a VC. If this is indeed the case, then it should not be discussed in the standard since it is outside the scope of the standard.

index.html

> Without knowing the context of a property (which is indicated by the type of the VC)...

The context is indicated by the `@context` property. Types can be a shortcut for this, but they needn't be. You can define "top level" properties that apply no matter where they are and then use the fact that it's a graph model to understand which subject (in the subject/predicate/object model) to which they apply.

> ...then in general it will make little sense to a verifier to simply know that an isolated property is present. What is this the ageOver property the age over of? Is it the subject, a relative of the subject, or a property of the subject, or a property of some other property etc.

Again, it's the property of whichever node in the graph upon which it appears. This allows verifiers to only care about "ageOver" and look for it on `credentialSubject` (for example) -- without having to care whether it comes from, whether that be a driver's license, a student ID, a birth certificate, etc. The verifier only has to care about "ageOver" and that the VC comes from an issuer they trust. The graph model tells them whether or not it applies to the entity for which they have some requirement.

> (Thought experiment. Replace all additional property names (except ageOver and the base properties that all VCs must contain) by XXXX in any arbitrary VC, and vary the position of ageOver between these XXXX properties, and then see of you can determine what ageOver really means. Clearly you cannot. This is why the type property is essential to understanding the VC.

I feel like there's still some misunderstanding here either about how the graph model works or what the use case is as, if I understand you correctly, this is a trivial exercise:

```js
{
    "...",
    "credentialSubject": {
      "id": "ex:abc",
      "ageOver": "18",
      "XXXX": {
        "id": "ex:def",
        "ageOver": "21",
        "XXXX": {
           "id": "ex:ghi",
           "ageOver": "18",
           "XXXX": /* and so on */
        }
      }
    }
  }
}
```

This will produce triples:

<ex:abc> <ageOver> "18" .
<ex:def> <ageOver> "21" .
<ex:ghi> <ageOver> "18" .

Which mean:

The entity identified by "ex:abc" is over the age of 18.
The entity identified by "ex:def" is over the age of 18.
The entity identified by "ex:ghi" is over the age of 18.

Which tells me exactly which entities the "ageOver" property applies to and I don't have to care in the slightest what the various "XXXX" properties (or types) are. If I can find a statement that says "ex:abc" (or whatever ID I care about) is over the requisite age, then I'm good, that's all I care about. Again, this is one good reason why we adopted a graph model in the first place.

David-Chadwick: Great explanation. But note that the "id" is optional, so the verifier might only be aware that some unknown entity has ageOver. Eg. credentialSubject has XXX has ageOver 18. Is XXX the subject's car? or daughter? or criminalRecord?

TallTed:
@David-Chadwick -- 

> @brentzundel said "What is not clear to me, is how an encapsulated object may be considered associated with a type that is not explicitly specified." This is also not clear to me, and I think the answer is "it cannot".

It certainly can -- by reasoning, inference, and logic.

This is kind of like the idea of a secret handshake between lodge members. Members are informed about this sub-type of handshake, and they can both deliver it and recognize receipt of it -- but they do not want the secret elements to be public, because they do not want non-members to either deliver nor recognize receipt of it.

In this case, explicit specification does exist, but neither the existence nor content of that specification is public. In many cases, the lodge members would even deny existence of such a secret handshake, because simply believing it to exist would likely lead to efforts by non-members to discover and make use of it.


> Could you please give me a concrete example of a property that has no meaning to the Verifier, has no impact on its decision making, and yet the issuer feels that it is important enough to place this property in the VC.

No, and that is precisely my point.  We cannot think of everything that may happen in the future.  We cannot predict what property an Issuer may want to add to a VC (maybe some kind of internal identification code for the VC, distinct from the VC's URI) which may have no meaning to *most* Verifiers, who don't even need to know this property exists, but may have meaning to *some* Verifier(s) (such as the Issuer themselves, who *may* become a Holder or Verifier at any point in the lifespan of a VC).

Does the presence of this property inherently make the VC a sub-type of whatever VC type has all the same properties *except* this extra?  By many lights, yes.  I certainly agree that the presence of this additional property *implicitly* makes a sub-type.

But I do not agree that this sub-type must be described nor identified nor declared explicitly nor publicly -- and that is what your arguments above state -- *and* your arguments above go beyond (and so would require further revision of) all of the current text, the revised texts from @brentzundel, and my own revised text.

I say such sub-type URI and description are only necessary if the Issuer wants that property to be understandable by *any* Verifier. If the Issuer wants that property to be understood only by their predetermined Verifiers, to whom the Issuer has provided a property description, then the property description may reasonably and necessarily be kept private and the sub-type  undeclared.

> This sentence implies that the issuer and the verifier have a private agreement, that is outside the scope of the standard, in which they agree which properties will be present in a VC. If this is indeed the case, then it should not be discussed in the standard since it is outside the scope of the standard.

... and the standard must *allow for this out-of-scope activity.*

That allowance is only delivered by having the sub-type declaration be OPTIONAL.

If the sub-type *MUST* be declared for a VC with extra properties to be a valid VC -- then such private arrangements between Issuer and (select) Verifiers are impossible -- and that, to me, is problematic.

index.html

> But note that the "id" is optional, so the verifier might only be aware that some unknown entity has ageOver. Eg. credentialSubject has XXX has ageOver 18. Is XXX the subject's car? or daughter? or criminalRecord?

Sure. But that's why the verifier asks for a VC with an "ageOver" claim that applies to whomever (i.e, has a matching ID) is authenticating with it. The verifier doesn't have to care what type of VC it is, only that the ID + "ageOver" (or some sensible way to link them) is present. Similarly, if you don't have an "id" that can sensibly be linked to whomever is authenticating it doesn't matter if you have type information or know other properties either.

David-Chadwick:
@TallTed "and the standard must allow for this out-of-scope activity." This is where we fundamentally disagree. All standards specify what is in scope and what is out of scope. Out of scope features are by definition not part of the standard and therefore are not standardised. You appear to want to standardise out of scope features.

"If the sub-type MUST be declared for a VC with extra properties to be a valid VC -- then such private arrangements between Issuer and (select) Verifiers are impossible -- and that, to me, is problematic."
Incorrect. They are not impossible. They are perfectly possible but are out of scope of the standard.
It is logically impossible for an out of scope feature to be part of the standard.

index.html
```diff
@@ -2611,7 +2615,7 @@ <h3>Terms of Use</h3>
     "https://www.w3.org/2018/credentials/examples/v1"
   ],
   "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
-  "type": ["VerifiablePresentation", "CredentialManagerPresentation"],
+  "type": ["VerifiablePresentation", "TermsOfUsePresentation"],
```
msporny: I'd prefer that we don't put this in here as it's probably not going to be best practice to do so. That is, the spec already states that you can specify termsOfUse in a presentation, so no additional type is necessary here. We don't want people copy-pasting this example.

msporny:
We need to discuss this on next Tuesday's call. I think the spec text that @brentzundel is proposing is fine (modulo the small change I mention above). 

Could we bring the discussion back to concrete changes to the text that @brentzundel is making? @David-Chadwick, @TallTed, do either of you have an issue with the text in the current PR?

index.html
msporny: I suggest adding some other funky extension to the presentation to make it more obvious (if we want to show this sort of thing in the specification).

index.html
TallTed: If the Verifier cannot associate `ageOver` with the Subject with which they are concerned, the Verifier is free to (and clearly should) refuse the requested action.  The `"id"` is optional *in the VC Data Model,* but it _may be required by the Verifier's Policy._

Consider the real-world age verification policies of alcohol sellers. Many will accept IDs issued in their local region (e.g., a local driver's license) or by their nation (e.g., a domestic passport), but not IDs issued in *any* region (e.g., a driver's license from a distant state, or a foreign passport) -- because it's relatively easy to know how to check validity of the local documents (basic formatting, watermarks, holographic imagery, etc.), but quite difficult to know how to check validity of every such document from around the world (does Russia include a hologram on their passports? does Moscow include a watermark in the lamination of their driver's license?)...

In other words --

* The VC Data Model is (and should be) flexible.  

* VC Verifier policies need not (and generally should probably not) be (as) flexible.

TallTed:
And I believe that the sentence "It is also possible for the additional more narrow types to be implicit." is implied by the "or be associated with" clause.

A better wording might be "It is also possible for the narrower type association to be implicit."

Note as well -- we do not require that the `subtype` be a URI (as our examples are simple literals), nor that any such URI dereference to anything that defines the properties for either humans or machines.  An Issuer could be in full conformance with this Model by declaring the sub-type to be `myCredentialType` or `http://example.com/fakeCredentialType` and including whatever properties they like -- but Verifiers should probably reject any such Credential.

David-Chadwick:
@TallTed "And I believe that the sentence "It is also possible for the additional more narrow types to be implicit." is implied by the "or be associated with" clause." Unfortunately that is not so. This phrase implied that the type could be in an incapsulating object rather than the object itself. So we should clarify the meaning of "associated with".

dlongley:
@TallTed,

> ...we do not require that the subtype be a URI (as our examples are simple literals)

Note that those subtypes map to URI via `@context`. The [spec says](https://w3c.github.io/vc-data-model/#types):

> The value of the type property MUST be, or map to (via interpretation of the @context property), one or more URIs.

@David-Chadwick,

> This is where we fundamentally disagree. All standards specify what is in scope and what is out of scope. Out of scope features are by definition not part of the standard and therefore are not standardised. You appear to want to standardise out of scope features.

My reading of what @TallTed wrote here:

> ... and the standard must allow for this out-of-scope activity.
> 
> That allowance is only delivered by having the sub-type declaration be OPTIONAL.
> 
> If the sub-type MUST be declared for a VC with extra properties to be a valid VC -- then such private arrangements between Issuer and (select) Verifiers are impossible -- and that, to me, is problematic.

...is that the spec has no business disallowing private agreements precisely because they are out of scope. Similarly, it is out of scope to say what those agreements must be or to define their shape in some way.

To be clear, I believe his argument is that your position is to bring those private agreements into scope by prohibiting them and mandating that some other method of agreement must be used. Rather, since they are out of scope, we must not prohibit or limit them in some specific way. Instead, the standard should allow for them to occur in whatever way those parties decide.

TallTed:
@dlongley 

re URI requirement - quite so, I meant to edit that before posting, as I realized the error on re-reading.  I've struck the erroneous text (as the remainder is still relevant) in [my comment](https://github.com/w3c/vc-data-model/pull/658#issuecomment-500469779).

re your reading of my previous -- Exactly.

burnburn:
From the VCWG call on 11 June 2019:

RESOLUTION: Merge PR 658 as it reflects the consensus of the Working Group. Specifically, more organizations would object to more stringent type requirements than what exists in PR 658.
[10:37] <brent> manu: I'll merge that as soon as Brent makes the minor change
[10:38] <brent> burn: that resolution only applies after a change has been made.

index.html
brentzundel: addressed via [12681f9](https://github.com/w3c/vc-data-model/pull/658/commits/12681f9a24bb7669e0f23268b16ef4d318b172c2)

brentzundel:
The change requested has been made and this PR is ready to merge according to the resolution of the working group.

index.html
```diff
@@ -1380,7 +1380,11 @@ <h3>Types</h3>
...
 <code>UniversityDegreeCredential</code>, for example) so software systems can
-process this additional information.
+process this additional information. The specification of these additional more
+narrow types could be done in the object within which the additional information
+is found, rather than at a higher level such as the <code>type</code>
+<a>property</a> of the <a>credential</a> or <a>presentation</a>. It is also
+possible for the additional more narrow types to be implicit.
         </p>
 
         <p>
```
David-Chadwick: Please add a definition of what an implicit type is in the context of a VC, and please provide an example of a VC with an implicit type in it.
Otherwise remove the concept of an implicit type as I do not know what it is or means.

TallTed: (Given yesterday's resolution, [the above comment](https://github.com/w3c/vc-data-model/pull/658#discussion_r292911089) should perhaps become a new issue, after application merging the PR in its current state.  My suggestion below will still apply.)

Perhaps --
`to be implicit.`
-> `to be implicit. Such implicit types may be implied by the Issuer by including, and inferred by the Verifier by evaluating the presence and/or value(s) of, properties of the implied type which are not described as part of the explicitly declared type(s).`

David-Chadwick: From the above explanation, then I infer that every issuer can issue VCs of just the type VC, and from the properties in the VC, the verifier can infer what type of VC this is. Thus negating the need for the type property to ever be anything other than single valued.

TallTed: Yes.

Issuers *may* Issue VCs with any — or no — other declared (sub-)type.

Verifiers *may* likewise have policies which reject — or require! — VCs with — or lacking! — whatever (sub-)type declaration they like.

Verifiers *may* equally have policies which accept VCs of any — or no! — (sub-)type which were Issued by a particular Issuer and which contain whatever property/ies they like — including properties of a declared (sub-)type of which they know nothing except that property identifier.

----

As I've said previously in this thread and elsewhere, I have always understood the "or be associated with" phrasing to make this declaration (but not the association) optional.

You have said you have always understood otherwise -- that the declaration had to be somewhere nearby.

I submit that your interpretation makes the model less flexible, less usable, and indeed less likely to be widely adopted, and would prevent some kinds of use which I can foresee.

I submit that my interpretation does not prevent your interpretation from being used by anyone who concurs with it.

I believe that all of this leads to one conclusion — which basically accedes to yesterday's resolution, and accepts either yesterday's wording or my small addition above.

David-Chadwick: Your proposed change is a significant normative change, because it invalidates the current text which states

All credentials, presentations, and encapsulated objects MUST specify, or be associated with, additional more narrow types (like UniversityDegreeCredential, for example) so software systems can process this additional information. 

You are now saying that "Issuers may Issue VCs with any — or no — other declared (sub-)type." which contradicts the current CR text.

If your proposed change is accepted this will mean a second CR is needed.

The only solution I can see to this dilemma is by creating a new subtype with the reserved word "Private" whose definition is something like "When the value of type property of the verifiable credential is Private, this means that the subtype has been determined privately between the issuer and the verifier and is not being specified in the type property"

TallTed: My text makes no normative change, and should not require a new CR.

> ... MUST specify, or be associated with, additional ...

-- is semantically equivalent to --

> ... MUST be associated with additional ...

-- as well as the somewhat more explicit, though apparently more confusing --

> ... MUST be associated, which association may be achieved by specific declaration or by inference, with additional ...

Your suggested requirement of a new "Private" subtype, on the other hand, _would be_ a normative change, and _would_ require a fresh CR.

brentzundel: I think that @David-Chadwick's suggestion above should be a separate issue:
> Please add a definition of what an implicit type is in the context of a VC, and please provide an example of a VC with an implicit type in it.

And @TallTed's proposed additional text addressing it should be a separate PR.

We already have WG consensus recorded in a resolution accepting this PR as it currently stands.

David-Chadwick: @brentzundel I fully agree with your original posting "My interpretation of the text". It was always intended that VCs and VPs had explicit typing. This was discussed at length and the WG should refer back to the original meeting minutes where this was discussed and agreed. (To remind you of the primary argument, it was so that a Verifier could quickly check the type to know if it could accept the VC or not, without having to process the entire VC first.) Implicit typing was never intended to be the semantics for the text "be associated with" and was as you said meant to refer to a higher type definition. The paragraph following the one containing "be associated with" makes it clear where this other type information should be found.

Changing explicit to implicit typing is a significant normative change to the spec and removes the benefit to the Verifier of fast processing. Therefore this PR should not be merged without other significant edits in this PR (or a companion one) that together ensure that a normative change is not made to the spec. Accepting this PR without also agreeing either further text in this PR or text in a companion PR will mean that two normative changes to the spec will have to be made in order to cancel each other out.

msporny:
Multiple positive reviews (negative review has a PR forthcoming that will resolve that negative review), non-normative, merging.

msporny:
The follow up to this PR can be found here: https://github.com/w3c/vc-data-model/pull/674

David-Chadwick:
@msporny Unfortunately you have misunderstood our conversation and my proposed resolution. I asserted that the introduction of implicit typing, as this PR does for the first time, is a normative change to the standard and therefore this should **not** be merged "as is". I proposed some alternative text to you that does allow implicit typing in certain cases, whilst mandating explicit typing in other cases. As a result of this, you produced PR#674, which keeps the implicit typing but does not mandate explicit typing for top level properties. Thus I am afraid that this issue still has not been satisfactorily resolved.

msporny:
> As a result of this, you produced PR#674, which keeps the implicit typing but does not mandate explicit typing for top level properties.

Explicit typing would be a normative and substantive change, right?

David-Chadwick:
@msporny . I dont believe so, because the minutes of previous meetings show that explicit typing was intended, as was the original discussion in this PR until people changed their minds. The introduction of implicit typing is the substantive change.

TallTed:
> the minutes of previous meetings show...

``[[citation needed]]``

msporny:
@David-Chadwick wrote:
> The introduction of implicit typing is the substantive change.

The addition of that language is not normative and thus not substantive. I'm quoting from the current specification to help others get a better idea of the text that I think you're referring to:

> The specification of these additional more narrow types could be done in the object within which the additional information is found, rather than at a higher level such as the type property of the credential or presentation. It is also possible for the additional more narrow types to be implicit.

Compare to the text in the CR version of the spec, where that doesn't exist:

https://www.w3.org/TR/verifiable-claims-data-model/#types

So, at this point, if there is going to be objections to this PR, we could just rip this PR out *and* the PR that added the language above, reverting the Types section to what we had at CR and call it a day.

Or do we want to keep going on this PR?

David-Chadwick:
> > the minutes of previous meetings show...
> 
> `[[citation needed]]`

https://www.w3.org/2018/06/19-vcwg-minutes.html

In which @TallTed said "Implicit typing mandates inference/reasoning in tools that consume the data. Explicit typing lowers the bar for verifying, etc. I believe there are other reasons to encourage explicit typing; the only reason I know of to discourage explicit typing is the "extra" triple."

David-Chadwick:
@msporny "

> Or do we want to keep going on this PR?

Given the disagreement in semantics and history between people who have been intimately involved in the production of the CR, then I pity those implementors who have not been. What will they make of it? I think the optimum solution is to keep going so as to have clear unambiguous text in the standard that everyone can agree on. Failing that, we are left with ambiguity where different people interpret fuzzy text to mean different things, which is not good for a standard.

In the latter case we can try to resolve the differences in the Implementors Guide, but if people have opposing irreconcilable views, then even that wont be possible. Then I guess we have to let the market decide (and as we know from VHS vs Betamax that the best technical solution does not always win :-)

TallTed:
> In which @TallTed said "Implicit typing mandates inference/reasoning in tools that consume the data. Explicit typing lowers the bar for verifying, etc. I believe there are other reasons to encourage explicit typing; the only reason I know of to discourage explicit typing is the "extra" triple."

As far as I can glean from re-reading the minutes cited (thank you), we were talking there about the base type -- i.e., VerifiableCredential -- not about sub-types.

Further, nothing in this PR, nor in any of the text I've been advocating, discourages explicit sub-typing, and does in fact encourage it, for the reason I gave in that quote among others.  This text simply does not REQUIRE explicit sub-typing, which requirement I do not believe I've ever advocated.

By all means, *discourage* implicit typing.  But do not *forbid* it.

PR #668, opened 2019-06-19
fixed issue 667 -- renamed iat in nbf

awoie:
As per resolution of our last meeting on June 18th, this is the PR that renames iat in nbf

***
<a href="https://pr-preview.s3.amazonaws.com/awoie/vc-data-model/pull/668.html" title="Last updated on Jul 16, 2019, 3:51 PM UTC (662d256)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/668/d92ceaf...awoie:662d256.html" title="Last updated on Jul 16, 2019, 3:51 PM UTC (662d256)">Diff</a>


.gitignore
```diff
@@ -1,3 +1,7 @@
 *.sw[nop]
 *~
 .DS_Store
+.idea/modules.xml
```
davidlehn: Could collapse all these into one `.idea` rule.

msporny:
After discussing this with the W3C Process team and W3C Management, it is clear that:

1. This was a typo introduced by the original author (@awoie) of the text, who intended to use `nbf` and mistyped it as `iat`.
2. Implementers of JWTs did the right thing in their implementation, which was the natural thing to do -- use `nbf`.
3. The W3C VCWG always intended the use of `nbf` and the typo in the spec was an oversight that was not caught until very late in the process.

In order to confirm that all of the above is true for the W3C PR Transition, we need the following:

* @awoie to confirm that he introduced a typo in the specification by using `iat` and the natural thing to do, and what he intended, was to use `nbf` (as demonstrated by the test suite PR).
* The VCWG to confirm that they had always intended the use of `nbf` as that was the natural thing to do.
* Every implementer that we know of to confirm that `nbf` was the natural thing to use and is what they have implemented.

Once we get all of those confirmations in, we can merge this PR as non-substantive.

David-Chadwick:
Approved

awoie:
@msporny I confirm that I introduced the typo in spec text. The intention was to use `nbf` instead of `iat`. The test suite PR had the `nbf` instead of the `iat` and therefore reflected the correct spec text. 

@msporny As an implementer we (uPort / ConsenSys) always used `nbf` to express the text contained in the specification as this is the natural thing to do.

peacekeeper:
On behalf of Danube Tech, I confirm that our implementation [verifiable-credentials-java](https://github.com/danubetech/verifiable-credentials-java) uses `nbf`.

msporny:
I haven't heard back from the Director yet, and I need to prep the document for publication next week (either as a CR or a PR). The group has decided to make this change whether or not a new CR is required per the 2019-07-16 telecon:

https://www.w3.org/2019/07/16-vcwg-minutes.html#item03

Merging.

PR #670, opened 2019-06-21
fixed #669 -- removed redundant and misleading statement

### LINKED ISSUE #669: Remove unnecessary limitation for alg JWT attribute
The spec currently reads:
> For backward compatibility reasons, the issuer MUST use JWS to represent proofs based on a digital signature.

And it further reads:
> alg MUST be used for RSA and ECDSA-based digital signatures.

Because digital signatures should in general use the JWS, the statement about RSA and ECDSA might be confusing and should be removed. We cannot provide an exhaustive list here, e.g., EdDSA should be included as well.

Proposed non-substantial change (because there is no limitation due to the statement above):
> alg MUST be used for digital signatures.

Furthermore, this change won't have an impact on the test suite results.

awoie:
fixed #669

***
<a href="https://pr-preview.s3.amazonaws.com/awoie/vc-data-model/pull/670.html" title="Last updated on Jun 25, 2019, 2:12 PM UTC (4760916)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/670/5a1f617...awoie:4760916.html" title="Last updated on Jun 25, 2019, 2:12 PM UTC (4760916)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3432,9 +3432,10 @@ specification:
 
             <ul>
               <li>
-<code>alg</code> MUST be used for digital signatures. If
-only the <code>proof</code> property is used, the <code>alg</code> header MUST
-be set to <code>none</code>.
+<code>alg</code> MUST be set for digital signatures. If only the 
+<code>proof</code> property is needed for the chosen signature 
+method (i.e., if there is no choice of algorithm within that method), 
+the <code>alg</code> header MUST be set to <code>none</code>.
               </li>
               <li>
 <code>kid</code> MAY be used if there are multiple keys associated with the
--- a/index.html
+++ b/index.html
@@ -3432,7 +3432,7 @@ specification:
 
             <ul>
               <li>
-<code>alg</code> MUST be used for RSA and ECDSA-based digital signatures. If
+<code>alg</code> MUST be used for digital signatures. If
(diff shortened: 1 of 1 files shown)


dlongley:
#671 touches these lines to fix a small typo -- would be nice to do it all here to avoid the inevitable conflict handling.

index.html
```diff
@@ -3432,7 +3432,7 @@ <h5>JWT Encoding</h5>
 
             <ul>
               <li>
-<code>alg</code> MUST be used for RSA and ECDSA-based digital signatures. If
+<code>alg</code> MUST be used for digital signatures. If
 only the <code>proof</code> attribute is used, the <code>alg</code> header MUST
```
mirceanis: incorporating #671 here:

```suggestion
only the <code>proof</code> property is used, the <code>alg</code> header MUST
```

TallTed:
@awoie -- I stand by my earlier suggested rewording.  I might even go further to --
```
<code>alg</code> MUST be set for digital signatures. If only the 
<code>proof</code> property is needed for the chosen signature 
method (i.e., if there is no choice of algorithm within that method), the <code>alg</code> header MUST be set to <code>none</code>.
```

PR #674, opened 2019-06-30
Add further explanation of types.

msporny:
This is a PR authored by @David-Chadwick to clarify the use of type information. 

I went to great lengths to keep as much of @David-Chadwick's original text as possible while also ensuring we weren't saying anything new that's normative or substantive that we weren't already stating in the spec before CR.

@David-Chadwick, I took *a lot* of time to cross-reference your normative text with stuff that was already in the spec. In two instances, we already had normative text that stated the same thing in the spec. In other cases, I couldn't figure out how to make your MUST statements testable. In others, I didn't know how to accomplish what you wanted us to do normatively without pushing us into another CR. I'm also concerned that this re-states things said elsewhere in the spec, which makes other parts of the spec confusing (because it seems like it's repeating this PR, which it is).

Ultimately, I think this text belongs in the Implementation Guidance document under a section related to creating new types of Verifiable Credentials.

It's also pretty evident that the **Types** section of the spec is now suffering from "too many people touched this section"-itis. It's sloppy, but because it contains so much normative language, it's difficult to clean it up at this point.

This PR is related to issue #653 and PR #658.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/674.html" title="Last updated on Jun 30, 2019, 8:28 PM UTC (19a2caf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/674/2ed130c...19a2caf.html" title="Last updated on Jun 30, 2019, 8:28 PM UTC (19a2caf)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1277,7 +1277,7 @@ containing machine-readable information about the <code>type</code>.
           </dd>
         </dl>
 
-        <pre class="example nohighlight" title="Usage of the type property">
+        <pre class="example nohighlight" title="Usage of the type property in a verifiable credential">
 {
   "@context": [
     "https://www.w3.org/2018/credentials/v1",
@@ -1295,6 +1295,53 @@ containing machine-readable information about the <code>type</code>.
   "proof": { ... }
 }
         </pre>
+
+        <p>
+With respect to this specification, there are at least three different
+categories of type information that implementers are advised to be aware
+of when building solutions. These are:
+        </p>
+
+        <ol>
+          <li>
+The type of the <a>verifiable credential</a>. This signals the top level
+properties that the verifiable credential contains to the <a>verifier</a>.
+The type <code>VerifiableCredential</code> signals all of the properties that
+a basic <a>verifiable credential</a> contains. These properties are all defined
+in this specification and include <code>credentialSubject</code>,
+<code>issuer</code>, <code>proof</code>, and others. If an <a>issuer</a> wants
+to add new top level properties to a <a>verifiable credential</a>, then it
+is strongly advised that the type property be extended with an additional type
+value that indicates that these new properties exist. It is also strongly
+advised that the <code>credentialSchema</code> be utilized to define which of
+these new properties are mandatory and which are optional.
+          </li>
+          <li>
+The type of the credential <a>subject</a>. Since the
+<code>credentialSubject</code> is simply identified by an identifier, the
+type of the <a>subject</a> might not be obvious to a <a>verifier</a>.
+For example, a temperature sensor might be one that reads the temperature of an
+oven, a person, or a refrigerator. Specifying identifier and temperature
+properties only might not allow the verifier to determine whether the sensor
+is operating correctly or not. Specifying the type of sensor will allow the
+<a>verifier</a> to know the expected range of temperature values.
+An <a>issuer</a> is strongly advised to express the type of
+the credential <a>subject</a> unless it might lead to issues such as privacy
+violations of the <a>subject</a>.
+          </li>
+          <li>
+The type that governs the properties of the credential <a>subject</a>.
+For example, a <a>verifiable credential</a> of type <code>UniversityDegree</code>
+might contain the discipline and degree classification, whilst an
+<code>ExtendedUniversityDegree</code> might contain a list of courses studied
+for the degree. An issuer might specify this additional subtype information in
+<code>type</code> property of the <a>verifiable credential</a> or might leave
+the subtype to be implicitly determined by the verifier. For example,
+when the <a>verifier</a> finds additional unexpected properties associated with
+the credential <a>subject</a> in the <a>verifiable credential</a> then it knows
+that the type has been extended.
+          </li>
+        </ol>
 
         <p>
 With respect to this specification, the following table lists the


index.html
```diff
@@ -1296,6 +1296,53 @@ <h3>Types</h3>
...
+oven, a person, or a refrigerator. Specifying identifier and temperature
+properties only might not allow the verifier to determine whether the sensor
+is operating correctly or not. Specifying the type of sensor will allow the
+<a>verifier</a> to know the expected range of temperature values.
+An <a>issuer</a> is strongly advised to express the type of
+the credential <a>subject</a> unless it might lead to issues such as privacy
+violations of the <a>subject</a>.
+          </li>
+          <li>
+The type that governs the properties of the credential <a>subject</a>.
```
dlongley: Since there can be more than one subject:

```suggestion
The type that governs the properties of a credential <a>subject</a>.
```

index.html
```diff
@@ -1296,6 +1296,53 @@ <h3>Types</h3>
...
+For example, a <a>verifiable credential</a> of type <code>UniversityDegree</code>
+might contain the discipline and degree classification, whilst an
+<code>ExtendedUniversityDegree</code> might contain a list of courses studied
+for the degree. An issuer might specify this additional subtype information in
+<code>type</code> property of the <a>verifiable credential</a> or might leave
```
```suggestion
the <code>type</code> property of the <a>verifiable credential</a> or might leave
```

David-Chadwick:
@msporny I appreciate your efforts and I trust we all want to achieve the same long term objectives - a standard that is unambiguous and fit for purpose. Given that this is essentially a security standard (credentials and identity being subsets of information security) then it is very important that we do not write the standard in such a way that a well-intentioned software developer, coupled with similar issuers and verifiers who use this software, believe that they have a standards conformant implementation, but have unwittingly introduced potential vulnerabilities into their services due to ambiguity or wiggle room in the specification. The current text does this by replacing a MUST in my original text with strongly advised in your text, when referring to the addition of new top level properties. This is because top level properties (e.g issuanceDate, termsOfUse etc) invariably constrain the validity of the credential. Thus a conformant issuer may, under the existing text, issue an occasionally constrained VC, but not mention this new property by the addition of a new type value as it is only strongly advised and this is only an occasional constraint, and the verifier may accept the VC when the constraint, if properly understood, would have caused the verifier to reject it. And this is for a well behaved issuer. A badly behaved issuer, or a well behaved issuer using sloppily written though conformant software, may always omit specifying the type value for constraints that always apply, causing verifiers to frequently accept VCs that they should not.

It is thus essential that the following sentence:

"If an issuer wants to add new top level properties to a verifiable credential , then it is strongly advised that the type property be extended with an additional type value that indicates that these new properties exist"

is changed to

"If an issuer wants to add new top level properties to a verifiable credential, then the type property MUST be extended with an additional type value that indicates that these new properties exist"

index.html
David-Chadwick: Agreed

David-Chadwick:
Minor edit

"This signals the top level properties that the verifiable credential contains to the verifier".

to

"This signals to the verifier the top level properties that the verifiable credential contains."

msporny:
> "If an issuer wants to add new top level properties to a verifiable credential, then the type property MUST be extended with an additional type value that indicates that these new properties exist"

How would we test this statement?

dlongley:
@David-Chadwick,

> If an issuer wants to add new top level properties to a verifiable credential, then the type property MUST be extended with an additional type value that indicates that these new properties exist

What if a property is optional? Certainly requiring a new type for every conceivable combination would be a bad idea. What if without the optional properties the VC would look the same as an existing type? It seems to me that this advice is fraught with potential problems.

We should also remember that we have already said that verifiers should understand the `@context` field which defines the types and properties -- and we have provided a `credentialSchema` field that could detail optional properties, etc. It seems like the bases are covered more cleanly already without this advice.

TallTed:
I will be posting some additional comments on the specific wording of the PR as it stands, which comments I am still composing.

@David-Chadwick --

> This is because top level properties (e.g issuanceDate, termsOfUse etc) invariably constrain the validity of the credential.

Top level properties *in VCs **you** Issue or Verify* may "invariably constrain the validity of the credential."  This will not necessarily be the case for VCs Issued or Verified by others!

As just one example -- such other Issuers may well include a top level property indicating the human who pressed the button that caused the VC to be generated -- which property is not relevant to any human Verifier who is not part of the Issuer organization.  Such a property may *not* constrain the validity of the VC, and should therefore not require subtyping of the VC, especially but not only if all other top-level properties conform to the description of the VC's base type.


> causing verifiers to frequently accept VCs that they should not

Note also that your MUST above does not mandate that such sub-type URI be dereferenceable by the (nor any) Verifier.  An Issuer is *perfectly free* to declare a sub-type is a URI which few if any are unable to dereference -- perhaps because the authority component of that URI points to an internal domain of the Issuer, e.g., http://vcs.internal.example.com/.  Your assumption is clearly that Verifiers will choose to refuse action on that basis -- but this is not (and in my strong opinion should not be) mandated by the Data Model.

I have said before and I will say again -- a Verifier is *perfectly free* to set a policy which refuses action based on any VC which includes properties which that Verifier does not understand, whether or not those properties are enumerated in the description of the VC's declared type(s), whether or not the Verifier can dereference (or otherwise access the description of) the declared subtype.  I believe that such blanket refusal based on the presence of any unknown property would be a poor policy in general, though it may be worthwhile in specific cases, but the marketplace may go either way with the model as currently specified -- and specific users may choose to go the other direction as well, if appropriate to their use.


If it is important to a Verifier that the VCs they receive be strongly typed, or that all such VCs be of some limited list of types, or that all properties be explicitly described in the description of those declared types -- that Verifier may refuse action based on any VC which is loosely typed, or strongly typed outside that list, or includes any property not explicitly described by the type descriptions associated with the declared types.

Failure to set appropriate policies would be a failure on the part of the Verifier; it would not be a failure nor weakness nor flaw within the Data Model.

David-Chadwick:
@dlongley The optionality of types is covered in the following sentence which refers to the schema, viz:

It is also strongly advised that the credentialSchema be utilized to define which of these new properties are mandatory and which are optional.

David-Chadwick:
@TallTed . In order to address your comment I suggest we send the CR to the security group and ask them for their specific viewpoint regarding the explicit/implicit typing of top level properties. I am very happy to agree with the decision they make.

David-Chadwick:
> > "If an issuer wants to add new top level properties to a verifiable credential, then the type property MUST be extended with an additional type value that indicates that these new properties exist"
> 
> Yes, I changed this statement because I think it's untestable. How would we test this statement?

I think you can test it as follows. Create a VC with a new top level property "NewProperty". A conformant implementation should reject this when parsing it.

TallTed:
@David-Chadwick --

Consider...

* What of the new top level property `"http://my.example.com/newProperty#this"`, which _may_ be dereferenced (and hence theoretically understood) without a declared VC type (because the property itself is a URI), though of course it will bring no description (because it's based on the unresolvable `example.com`)?

* What of the new declared subtype `"http://my.example.com/newVCType#this"`, which _may_ be dereferenced (and hence its properties theoretically revealed and understood) because it is a URI), though of course it will bring no description and hence no properties (because it's based on the unresolvable `example.com`)?

Your suggested restrictive changes to the spec do not deliver the assurance you clearly believe -- and want others to believe -- they do.  I submit that acting as if that belief were fact would render your more-restricted VCs less secure or reliable than retaining and benefiting from the full flexibility now specified while recognizing its potential weaknesses and countering the latter by policy.

index.html
```diff
@@ -1296,6 +1296,53 @@ <h3>Types</h3>
...
+the subtype to be implicitly determined by the verifier. For example,
+when the <a>verifier</a> finds additional unexpected properties associated with
+the credential <a>subject</a> in the <a>verifiable credential</a> then it knows
+that the type has been extended.
+          </li>
```
TallTed:
```pre
The type that governs the properties of the credential <a>subject</a>.
For example, a <a>verifiable credential</a> of type <code>UniversityDegree</code>
might contain the discipline and degree classification, whilst an
<code>ExtendedUniversityDegree</code> might contain a list of courses studied
for the degree. An issuer might specify this additional subtype information in
<code>type</code> property of the <a>verifiable credential</a> or might leave
the subtype to be implicitly determined by the verifier. For example,
when the <a>verifier</a> finds additional unexpected properties associated with
the credential <a>subject</a> in the <a>verifiable credential</a> then it knows
that the type has been extended.
```
->
```pre
             <p>
The subtype of <a>verifiable credential</a> (communicated as the value of a
<a>type</a> property), which describes which properties of the credential
<a>subject</a> are expected to be found in a <a>verifiable credential</a> of
that subtype. Such properties might be described as mandatory or optional. For
example, a <a>verifiable credential</a> of subtype <code>UniversityDegree</code>
might be described to contain mandatory properties including the discipline and
degree classification, while a <a>verifiable credential</a> of subtype
<code>ExtendedUniversityDegree</code> might optionally contain a list of courses
studied for the degree.
             </p>
             <p>
An issuer might specify this additional subtype information in the <code>type</code>
property of the <a>verifiable credential</a>, or might leave the subtype to be
implicitly determined (or "inferred") by the verifier. For example, when a
<a>verifier</a> finds additional unexpected properties associated with the
credential <a>subject</a> in the <a>verifiable credential</a>, then the verifier
knows that the base type has been extended.  Verifiers might set policies that
refuse all action if a <a>verifiable credential</a> contains properties not
enumerated by its explicitly declared <a>type</a>; or set policies that refuse
any action based on a property not enumerated by an explicitly declared <a>type</a>;
simply ignore such extra properties; or treat the extra properties as just as
authoritative as anything else in the <a>verifiable credential</a>, based on
their treatment of the issuer as universally authoritative. There are other
possibilities.
```

That's pretty extensive.  It probably wants more text, pulling from the extensive discussion threads on this topic.  Or this whole PR could be omitted entirely.

msporny:
Withdrawing this PR as the original author of the text wanted their text to be put forth as a PR without Editorial oversight.

This is done in PR #691.

Issue #537, opened 2019-04-08
JSON-LD Reference


nadalin:
Should this specification actually be built on JSON-LD 1.1 to better align with JSON structures which would align better with JWT

chaals:
This seems like a rehash of #491, and other related issues. The answer to the question has been provided repeatedly:

- https://github.com/w3c/vc-data-model/issues/491#issuecomment-478031097
- https://github.com/w3c/vc-data-model/issues/491#issuecomment-478823560
- https://github.com/w3c/vc-data-model/issues/483#issuecomment-477310938
- https://github.com/w3c/vc-data-model/issues/483#issuecomment-477970532

as well as discussed at the last face to face meeting and various other places.

I believe this issue brings nothing new and should be closed without action.

nadalin:
@chaals not a rehash at all as specification lists JSON-LD 1.0 and my comments is that some issues may be resolved if this specification is based upon JSON-LD 1.1

msporny:
JSON-LD 1.1 isn't a REC yet, only 1.0 is. That said, in reality, most of the implementations that we know are already using the JSON-LD 1.1 features that have shipped in processors for the `@protected`, which would address some of @nadalin's JSON-LD processing concerns.

In a previous WG, we published a REC that noted that it would upgrade to the latest version of a dependent spec when it hit REC. The WG closed, but the W3C staff updated the REC after it had been published and the dependent spec became a REC. So, there is precedent to do this even though the process is a bit shaky.

In any case, the group is already using JSON-LD 1.1 in practice... it would be good if we could align the spec w/ this reality rather than be limited by W3C Process.

nadalin:
@msporny We did this with WebAuthn and the CredMan support as it's not REC yet and we went to the WebAppSec folks and gota statement from them

msporny:
> @msporny We did this with WebAuthn and the CredMan support as it's not REC yet and we went to the WebAppSec folks and gota statement from them

Ok, then happy to do that wrt. the VC spec. I expect the JSON-LD 1.1 WG will support this, we will bring it up wrt. processing this issue during one of the next calls.

[...]

PR #677, opened 2019-07-03
Update JSON-LD 1.0 references to JSON-LD 1.1.

msporny:
Dependency on JSON-LD 1.1 features have been confirmed by the JSON-LD 1.1 WG as stable features.


msporny:
Merging based on WG resolution from VCWG call on 7 May 2019:

Issue #537 should be addressed by referencing JSON-LD 1.1 instead of JSON-LD 1.0. This is a non-substantive change since the Verifiable Credentials Data Model specification included JSON-LD 1.1 features inline for JSON-LD based processors and thus implementers would not have to change their implementations as a result of the updated reference. Issue #537 should be closed after the PR is merged.

PR #685, opened 2019-07-11
fix subclass/superclass typing requirement mismatch

TallTed:
*In the summary table,* change the type requirements of the superclasses (`Credential` and `Presentation`) to match the type requirements of the subclasses (`VerifiableCredential` and `VerifiablePresentation`).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/685.html" title="Last updated on Jul 11, 2019, 2:55 PM UTC (e158b5c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/685/96dc09e...e158b5c.html" title="Last updated on Jul 11, 2019, 2:55 PM UTC (e158b5c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1349,7 +1349,7 @@ objects that MUST have a <a>type</a> specified.
 <a href="#credentials">Credential</a>&nbsp;object
               </td>
               <td>
-<code>VerifiableCredential</code> and a more specific
+<code>VerifiableCredential</code> and, optionally, a more specific
 <a>verifiable credential</a> <a>type</a>. For example,<br>
 <code>"type": ["VerifiableCredential", "UniversityDegreeCredential"]</code>
               </td>
@@ -1371,7 +1371,7 @@ objects that MUST have a <a>type</a> specified.
 <a href="#presentations">Presentation</a>&nbsp;object
               </td>
               <td>
-<code>VerifiablePresentation</code> and a more specific
+<code>VerifiablePresentation</code> and, optionally, a more specific
 <a>verifiable presentation</a> <a>type</a>. For example,<br>
 <code>"type": ["VerifiablePresentation", "CredentialManagerPresentation"]</code>
               </td>


David-Chadwick:
Isnt this a substantive change, as it is effectively changing a MUST into a MAY?

TallTed:
No, because the table is summarizing statements made elsewhere.  The [relevant body text](https://github.com/w3c/vc-data-model/blob/96dc09e869069bcf5f9c36e039f81e32c3cb8b04/index.html#L1282) which is summarized in the table requires only `a type property`, not `a type and a subtype`.

David-Chadwick:
@TallTed And because the statements elsewhere have changed substantially so the table needs to change to sync up.

David-Chadwick:
Guys this is not an editorial, it is a substantive change of a MUST to a MAY. I object to this merging on these grounds

brentzundel:
@David-Chadwick the associated text from the spec:
> All credentials, presentations, and encapsulated objects MUST specify, or be associated with, additional more narrow types (like UniversityDegreeCredential, for example) so software systems can process this additional information. *The specification of these additional more narrow types could be done in the object within which the additional information is found, rather than at a higher level such as the type property of the credential or presentation.*

The bold text makes clear that any additional properties may be specified at a lower level than the `type` property of the verifiable credential or presentation, therefore having a more specific type in the `type` property of a verifiable credential or presentation is optional. The more specific type may be specified in the `type` property of the `credentialSubject`, for example.

msporny:
In addition to what @brentzundel says above, the CR spec published on March 28th states:
> All credentials, presentations, and encapsulated objects MUST specify, **or be associated with**, additional more narrow types (like UniversityDegreeCredential, for example) so software systems can process this additional information.

David-Chadwick:
@msporny . Correct. And the following sentence in March 28th attempts to define what associated with means, viz

"When processing encapsulated objects defined in this specification, (for example, objects **associated with** the credentialSubject object or deeply nested therein), software systems SHOULD use the type information specified in encapsulating objects higher in the hierarchy. Specifically, an encapsulating object, such as credential, SHOULD convey the associated object types so that verifiers can quickly determine the contents of an associated object based on the encapsulating object type."

So the March 28 version makes it very clear that
a) all objects MUST specify their more narrow types or be associated with this type
and
b) associate with refers to the higher encapsulating objects and to nothing else (like an implicit type)

So objects SHOULD specify their types in their higher encapsulating types, but if not, MAY specify them in the objects themselves. But there is NO get out clause saying that objecs do not need to specify their types.

Issue #682, opened 2019-07-10
Implicit Types Insertion


David-Chadwick:
Implicit types have only been inserted into the data model during June/July 2019. Prior to this there was no mention of implicit types in the data model document, and the understanding (at least by many in the group if not all) was that all the types are explicit i.e. the data model is strongly typed as opposed to weakly typed. 

The facts that support this are as follows:

i) the  minutes of 19 June 2018 (https://www.w3.org/2018/06/19-vcwg-minutes.html) where types were discussed, it was clear that explicit typing was to be used. Quote by @TallTed "Implicit typing mandates inference/reasoning in tools that consume the data. Explicit typing lowers the bar for verifying, etc. I believe there are other reasons to encourage explicit typing; the only reason I know of to discourage explicit typing is the "extra" triple."

ii) the original CR text (March 19) of the data model document contains several MUST clauses about typing, and there is no mention of implicit typing or optional typing.

iii) #PR658, and in particular the comment of @brentzundel of 6 June 2019 starting with "Intro" and saying "My interpretation of the text" makes it clear that explicit typing was intended and understood by him. Further in the conversation @brentzundel then said "@David-Chadwick that was my interpretation of the text as well, but I can see potential value in the notion of implicit sub-types as **@TallTed has introduced them**."

So it is clear that @TallTed changed his opinion and introduced the notion of implicit typing in June 2019, a year after he first strongly argued for explicit typing. After that, the majority of the working group accepted the concept of implicit typing, and it has been added to the data model.

This is not to say that implicit typing could not be supported by the data model or is better or worse than explicit typing. But it is a substantive and normative change to the original CR specification, and the inclusion of this concept has not been fully worked through in the specification.

The only clause in the original CR that caused some ambiguity was the phrase "All credentials, presentations, and encapsulated objects MUST specify, **or be associated with**, additional more narrow types (like UniversityDegreeCredential, for example) so software systems can process this additional information". Some members stated that **or be associated with** referred to higher types in the document in line with the explicit typing requirement, whilst others said that it implied weaker typing. The former opinion was supported by the following sentence in the specification, which attempted (rather badly) to describe what **associated with** meant "When processing encapsulated objects defined in this specification, (for example, objects **associated with** the credentialSubject object or deeply nested therein), software systems SHOULD use the type information specified in encapsulating objects higher in the hierarchy".  There was no text in the original CR to support the concept of implicit typing or that **associated with** meant implicit typing.

[...]

PR #691, opened 2019-07-14
Clarify implicit typing. Related to #682.

msporny:
This PR is authored by @David-Chadwick to address the implicit typing issue that he raised in #682.

As requested by @David-Chadwick, the Editors have made no attempt to align the text with other parts of the specification or the W3C Process.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/691.html" title="Last updated on Jul 14, 2019, 8:08 PM UTC (8b7b276)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/691/d92ceaf...8b7b276.html" title="Last updated on Jul 14, 2019, 8:08 PM UTC (8b7b276)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1302,6 +1302,46 @@ in a document containing machine-readable information about the
         </pre>
 
         <p>
+There are three different classes of type information that can be added to the
+verifiable credential type property. These are:
+        </p>
+
+        <ol>
+          <li>
+The type of verifiable credential. This governs the top level properties that
+the verifiable credential contains. The type “verifiableCredential” specifies
+all the properties that a basic verifiable credential must or may contain, and
+these are all defined in this specification. They include properties such as
+termsOfUse, IssuanceDate, credentialSubject, proof etc. If an issuer wants to
+add new top level properties to a verifiable credential, then the type property
+MUST be extended with an additional type value that indicates that these new
+properties exist. The credentialSchema property SHOULD also be present, to
+define which of these new properties are mandatory and which are optional.
+          </li>
+          <li>
+The type of credentialSubject. Because the credentialSubject is simply
+identified by an ID, then it may not be obvious to a verifier what type of
+credentialSubject this is. For example, a temperature sensor may be one that
+reads the temperature of an oven, a person or a refrigerator. Specifying ID and
+temperature properties only may not allow the verifier to determine whether the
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1422,11 +1462,7 @@ <h3>Types</h3>
 All <a>credentials</a>, <a>presentations</a>, and encapsulated objects MUST
 specify, or be associated with, additional more narrow <a>types</a> (like
 <code>UniversityDegreeCredential</code>, for example) so software systems can
-process this additional information. The specification of these additional more
-narrow types could be done in the object within which the additional information
-is found, rather than at a higher level such as the <code>type</code>
-<a>property</a> of the <a>credential</a> or <a>presentation</a>. It is also
```
msporny: I suggest we just strike "It is also possible for the additional more narrow types to be implicit." and call it a day. The CR specification already said "or be associated with" ... and there are MANY ways to do that, not just the ones outlined in this PR.

TallTed:
I generally agree with @msporny's points. This PR 1) includes multiple substantive changes of normative text, some of which are contradictory; 2) would remove much flexibility that was present in CR#1, which I think is important to retain; 3) oversimplifies and/or ignores real-world complexity; and 4) may provide an illusion of security which is not delivered by the additional restrictions.

-----

I would be OK with replacing this entire PR with a strike of "It is also possible for the additional more narrow types to be implicit." *IFF* the preceding sentence were changed to start with `Among other possibilities, the specification of or association with these additional` (rather than `The specification of these additional`).

To be clear, that would replace -- 
```pre
The specification of these additional more
narrow types could be done in the object within which the additional information
is found, rather than at a higher level such as the <code>type</code>
<a>property</a> of the <a>credential</a> or <a>presentation</a>. It is also
possible for the additional more narrow types to be implicit.
```
-- with --
```pre
Among other possibilities, the specification
of or association with these additional more narrow types could be done in the
object within which the additional information is found, rather than at a higher
level such as the <code>type</code> <a>property</a> of the <a>credential</a> or
<a>presentation</a>.
```

index.html
brentzundel: I support Manu's proposal

David-Chadwick: I support it as well

msporny:
Closing in favor of PR #702.

Issue #705, opened 2019-07-23
Verification definition incorrect


jandrieu:
In section 2. Terminology, the spec has the following definition of verification:
> **verification**
> The evaluation of whether a verifiable credential or verifiable presentation complies with this specification.

Unfortunately, this is conformance to the specification, and not verification as meant in the "verifiable" part of "verifiable credentials".

In 5.1 Lifecycle details, we have the following language:
> A verifier verifies the authenticity of the presented verifiable presentation and verifiable credentials. This should include checking the credential status for revocation of the verifiable credentials.

We should use this language (or similar) to correct the glossary term.

stonematt:
PR merged,  Issue resolved.

PR #706, opened 2019-07-23
Update terms.html

jandrieu:
A correction to the glossary term for "verification" to bring it into alignment with the rest of the text.

### DIFF
--- a/terms.html
+++ b/terms.html
@@ -153,8 +153,10 @@ for identifiers.
   </dd>
   <dt><dfn data-lt="verify|verified|verifying|verifiable|verifiability">verification</dfn></dt>
   <dd>
-The evaluation of whether a <a>verifiable credential</a> or 
-<a>verifiable presentation</a> complies with this specification.
+The evaluation of whether a <a>verifiable credential</a> or <a>verifiable presentation</a> 
+    is an authentic and timely statement of the issuer (or presenter for a presentation). 
+    This includes checking the credential status for suspension or revocation of the 
+    verifiable credentials as well as applying the verification method specified.
   </dd>
   <dt><dfn data-lt="verifier|verifiers|verifier's|credential verifiers|credential verifier's">verifier</dfn></dt>
   <dd>
--- a/terms.html
+++ b/terms.html
@@ -155,8 +155,8 @@ for identifiers.
   <dd>
 The evaluation of whether a <a>verifiable credential</a> or <a>verifiable presentation</a> 
 is an authentic and timely statement of the issuer or presenter, respectively.  
-    This means the credential is conformant to the specification, the proof method is
-    satisfied, AND, if present, the status check succeeds.
+    This includes checking that: the credential (or presentation) conforms to the specification; the proof method is
+satisfied; AND, if present, the status check succeeds.
 
   </dd>
   <dt><dfn data-lt="verifier|verifiers|verifier's|credential verifiers|credential verifier's">verifier</dfn></dt>
--- a/terms.html
+++ b/terms.html
@@ -156,7 +156,7 @@ for identifiers.
 The evaluation of whether a <a>verifiable credential</a> or <a>verifiable presentation</a> 
 is an authentic and timely statement of the issuer or presenter, respectively.  
     This includes checking that: the credential (or presentation) conforms to the specification; the proof method is
-satisfied; AND, if present, the status check succeeds.
+satisfied; and, if present, the status check succeeds.
 
   </dd>
   <dt><dfn data-lt="verifier|verifiers|verifier's|credential verifiers|credential verifier's">verifier</dfn></dt>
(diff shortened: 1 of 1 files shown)


TallTed:
This is potentially a very deep rabbit hole. Many terms are overloaded with multiple possible meanings, and several are sometimes if not always synonyms -- including "validation" and "verification".

"Validation" *might* be construed to mean "checking whether the VC/VP is valid, i.e., that it has not expired, and that it's 'date of activation' has passed, and that all the proofs evaluate successfully" *or* to mean "checking whether the VC/VP conforms to the data model specified herein."

"Verification" might likewise be construed to mean either of these things.

***For purposes of this document,*** because we're setting up for future *verifiability*, I think we must use (and define in terms.html) "validation" in the sense of "conforms to the data model" and "verification" in the sense of "dates and proofs are satisfactory" (with that "verification" stage being followed by "application of verifier policy").

Regrettably, doing this properly means reviewing all our documents and ensuring that both "valid"- and "verify"-related words are used consistently in the above senses -- and changing from one word to the other wherever they are currently used in the other sense.

This is very troublesome at this stage, because it's quite time and focus intensive, but it should be purely editorial in action and result.

brentzundel:
I agree with @TallTed 's assessment.

TallTed:
We might want to almost entirely do away with "validation", "valid", and the like, and replace them with "checking conformance with the data model" or similar phrasing, "conformant", and so on...  keeping "valid" only to be specifically associated with date tests against validfrom/validuntil.

Yes, this is terribly timed.  Yes, the docs *can* be used without these changes.

Still, their comprehensibility, and the follow-on arguments about the interpretation of their current ambiguities, will be immeasurably improved by making such changes.

stonematt:
I assert:  conformant != verified != valid, and definitions should reflect that.

The term verify and valid have market meaning.  The concept of "Valid" is often policy driven and contextual, as such we have explicitly called it out of scope (a standard may never tackle this).  The concept of verify is driven by explicit document content and requires a verification agent to complete, which is also out of scope for this group.

@jandrieu and I spent some time this morning on this topic. We think we have a path forward that is a small, incremental change to the current definition of 'verify' that invokes the intent of proofs/signatures and role of the Status method while preserving the differentiation between verify and validate.  Joe is going to present language where verify is conformance based on section 1.4 plus 

The important nuance to this is we are not expressing 'how' verifier would verify a VC/VP (that's in the domain of a user agent/protocol), but more explicitly stating what's meant by verification of a verifiable artifact.  We want the community to understand that an artifact which had a positive verification yesterday may not verify tomorrow, even though the document itself hasn't changed. The ephemeral quality of these must be satisfied before the question of validation and fit for purpose, etc is ever invoked. 

For example, we don't specify a signature suite, but if one exists, a verifier must know how to check the document against the signature in order to know that the claim being read hasn't been tampered with since it was issued.

TallTed:
I think @stonematt's summary fits with my latest comment. Pending review of @jandrieu's language, it sounds like the right direction.

jandrieu:
I added the language. I didn't reference Section 1.4 because that is an external file. Feedback welcome.

terms.html
```diff
@@ -153,8 +153,11 @@
   </dd>
   <dt><dfn data-lt="verify|verified|verifying|verifiable|verifiability">verification</dfn></dt>
   <dd>
-The evaluation of whether a <a>verifiable credential</a> or 
-<a>verifiable presentation</a> complies with this specification.
+The evaluation of whether a <a>verifiable credential</a> or <a>verifiable presentation</a> 
+    is an authentic and timely statement of the issuer (or presenter for a presentation). 
+    This means the credential is conformant to the specification, the proof method is
+    satisfied, AND, if present, the status check succeeds.
+
```
TallTed: I suggest some small tweaks to the whole paragraph...

```pre
The evaluation of whether a <a>verifiable credential</a> (or <a>verifiable presentation</a>) 
is an authentic and timely statement of the issuer (or presenter). This includes checking 
that: the credential (or presentation) conforms to the specification; the proof method is
satisfied; AND, if present, the status check succeeds.
```

The first sentence might be better as --
```pre
The evaluation of whether a <a>verifiable credential</a> or <a>verifiable presentation</a> 
is an authentic and timely statement of the issuer or presenter, respectively. This includes 
checking that: the credential (or presentation) conforms to the specification; the proof 
method is satisfied; AND, if present, the status check succeeds.
```

jandrieu: Agreed.

terms.html
```diff
@@ -153,8 +153,11 @@
   </dd>
   <dt><dfn data-lt="verify|verified|verifying|verifiable|verifiability">verification</dfn></dt>
   <dd>
-The evaluation of whether a <a>verifiable credential</a> or 
-<a>verifiable presentation</a> complies with this specification.
+The evaluation of whether a <a>verifiable credential</a> or <a>verifiable presentation</a> 
+    is an authentic and timely statement of the issuer (or presenter for a presentation). 
+    This means the credential is conformant to the specification, the proof method is
+    satisfied, AND, if present, the status check succeeds.
```
msporny: lowercase AND -> and

stonematt:
addresses issue  #705

stonematt:
@TallTed looks like this one address the comments inline.  Will you give this a final thumbs up so we can pull it in.

PR #707, opened 2019-07-23
Update iss of JWT

David-Chadwick:
U

### DIFF
--- a/index.html
+++ b/index.html
@@ -3499,7 +3499,8 @@ standard <a>verifiable credential</a> counterparts:
 timestamp (<code>NumericDate</code>).
               </li>
               <li>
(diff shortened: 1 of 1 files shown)


awoie:
PR looks good to me

msporny:
From 2019-07-23 telecon: 

RESOLUTION: The issue concerning the JWT iss/holder relationship is resolved by PR #707, which makes a substantive change to the specification. The text should be included in the 2nd Candidate Recommendation publication along with a section marking the JWT feature as at-risk.

https://www.w3.org/2019/07/23-vcwg-minutes.html#resolution03

Substantive, will go into Second CR, merging.

Issue #709, opened 2019-08-06
Underspecification of issuer object format and document obtained if issuer id is dereferenced


nadalin:
I have been trying to implement this specification and having issues, here is one issue I ran into, I'm plodding along to try to finish my implementation.

"The value of the issuer property MUST be either a URI or an object containing an id property. It is RECOMMENDED that the URI in the issuer or its id be one which, if dereferenced, results in a document containing machine-readable information about the issuer that can be used to verify the information expressed in the credential."

I take this is intended to mean the Issuer field specifies the entity that issued the credential, and the value of the field should be a URI that resolve to the data/keys required for verifying the credential, such as a DID/DID Doc, JWK at the specified URL endpoint, etc.

The issue is how do you know what the return document will be, like DID Doc vs JWK vs ___? This  must be clearly specified in the field before resolution; The NOTE  may be trying to vaguely allude to that  https://www.w3.org/TR/vc-data-model/#issuer but does not.

Suggest to use object descriptor as the value that specifically type-identifies what will come back from the resolution of the URI.

issuer: {
  uri: “did:foo:123”,
  format: “did-document”
}

issuer: {
  uri: “https://example.edu/comp-sci/identity-root”,
  format: “jwk”
}

issuer: {
  uri: “ipfs:QmPXME1oRtoT627YKaDPDQ3PwA8tdP9rWuAAweLzqSwAWT”,
  format: “jwk”
}

brentzundel:
I don't think it will be a problem to add an example to the data model that shows an issuer property with a value that is an object.

brentzundel:
It may also be possible to give more detail about this in the implementation guide

nadalin:
@brentzundel Example I don't believe is enough as examples are non-normative. Trying to implement I find the specification under specified in this area. I can't handle a return object properly without knowing what type it is.

brentzundel:
> Suggest to use object descriptor as the value that specifically type-identifies what will come back from the resolution of the URI.

The functionality you desire does not require the spec to be changed in any way. As it is written, the data model supports exactly what you propose.

> I can't handle a return object properly without knowing what type it is.

An implementation (such as yours) is free to specify what return type it expects when dereferencing a URI associated with an issuer.
The data model supports this by allowing the value of in issuer property to be an object (as long as the object has an id property which specifies the URI). The issuer object may also contain properties such as those you describe. The example added via PR #710 shows this.

> Example I don't believe is enough as examples are non-normative. 

[...]

PR #710, opened 2019-08-06
added example of issuer with object as value

brentzundel:
Addresses issue #709 

***
<a href="https://pr-preview.s3.amazonaws.com/brentzundel/vc-data-model/pull/710.html" title="Last updated on Aug 13, 2019, 6:51 PM UTC (3c5b319)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/710/6d8c996...brentzundel:3c5b319.html" title="Last updated on Aug 13, 2019, 6:51 PM UTC (3c5b319)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1614,7 +1614,7 @@ example, <code>"did:example:abfe13f712120431c276e12ecab"</code>).
   "type": ["VerifiableCredential", "UniversityDegreeCredential"],
   <span class="highlight">"issuer": {
       "id": "https://example.com/keys/foo.jwk",
-      "type": "jwk"
+      "type": "urn:example:jwk"
   }</span>,
   "issuanceDate": "2010-01-01T19:23:24Z",
   "credentialSubject": {
--- a/index.html
+++ b/index.html
@@ -1613,8 +1613,8 @@ example, <code>"did:example:abfe13f712120431c276e12ecab"</code>).
   "id": "http://example.edu/credentials/3732",
   "type": ["VerifiableCredential", "UniversityDegreeCredential"],
   <span class="highlight">"issuer": {
-      "id": "https://example.edu/issuers/14",
-      "format": "jwk"
+      "id": "https://example.com/keys/foo.jwk",
+      "type": "jwk"
   }</span>,
   "issuanceDate": "2010-01-01T19:23:24Z",
   "credentialSubject": {
@@ -1624,7 +1624,7 @@ example, <code>"did:example:abfe13f712120431c276e12ecab"</code>).
       "name": "Bachelor of Science and Arts"
     }
   },
-  "proof": { ... }
+  "proof": { <span class="comment">...</span> }
 }
         </pre>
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1604,6 +1604,30 @@ <h3>Issuer</h3>
...
+  }</span>,
+  "issuanceDate": "2010-01-01T19:23:24Z",
+  "credentialSubject": {
+    "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
+    "degree": {
+      "type": "BachelorDegree",
+      "name": "Bachelor of Science and Arts"
+    }
+  },
+  "proof": { ... }
```
TallTed: needs the CSS magic

index.html
```diff
@@ -1604,6 +1604,30 @@ <h3>Issuer</h3>
...
+{
+  "@context": [
+    "https://www.w3.org/2018/credentials/v1",
+    "https://www.w3.org/2018/credentials/examples/v1"
+  ],
+  "id": "http://example.edu/credentials/3732",
+  "type": ["VerifiableCredential", "UniversityDegreeCredential"],
+  <span class="highlight">"issuer": {
+      "id": "https://example.com/keys/foo.jwk",
+      "type": "jwk"
```
msporny: We need to define this in the JSON-LD context or use something like urn:example:jwk.

brentzundel: changed in [3c5b319](https://github.com/w3c/vc-data-model/pull/710/commits/3c5b319b41b618b831c127dfaefea3004ac72b16)

brentzundel: Let me know if this change works.

msporny:
I've been thinking about this PR over the last week or so as I was on the fence about it. Fundamentally, I think the example is an anti-pattern that we don't want people to use, for the following reason:

The PR suggests that people should use public keys to identify issuers. This is a bad practice because you don't get key rotation when you do this, which is one of the things that DIDs achieved (and why so many of us are proceeding along those lines). 

We had started a few years ago ago by using keys as issuers, but we've gone away from doing that over the years because of the bind it puts issuers and verifiers in. Everyone needs to track whose keys are whose and that leads to complex and fragile systems (as exemplified by some of the systems utilizing JWKs as ways to identify issuers).

As such, we came up with the `controller` pattern, where (at least in LD Proofs) the key is associated with a proof, which points to the key via something like (in the Verifiable Credential LD Proof):

```
    "verificationMethod": "did:example:1234567#key-1",
```

When you go and look at that key (in the DID Document), it's expected that it has a controller or some way of associating it with the entity that controls the key. These keys are usually tied to entities in the DID Document, and thus it enables you to do stuff like this (in the Verifiable Credential):

```
  "issuer": "did:example:1234567"
```

... which is great for issuers and verifiers, because that's a long lived identifier that doesn't break when you rotate keys.

This example goes against that very hard earned lesson, and thus I think we should defer pulling this example into the specification.

msporny:
Overcome by PR #715, which has been merged.

PR #711, opened 2019-08-08
Fix typos in JWT encoding decoding

David-Chadwick:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/711.html" title="Last updated on Aug 15, 2019, 7:15 AM UTC (2540dfb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/711/6d8c996...2540dfb.html" title="Last updated on Aug 15, 2019, 7:15 AM UTC (2540dfb)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3478,7 +3478,7 @@ Contained in the JWS signature part.
             <p>
 If no explicit rule is specified, <a>properties</a> are encoded in the same way
 as with a standard <a>verifiable credential</a>, and are added to the
-<code>vc</code> <a>property</a> of the JWT. As with all JWTs, the JWS-based
+<code>vc</code> <a>claim</a> of the JWT. As with all JWTs, the JWS-based
 signature of a <a>verifiable credential</a> represented in the JWT syntax is
 calculated against the literal JWT string value as presented across the wire,
 before any decoding or transformation rules are applied. The following
@@ -3562,18 +3562,11 @@ the Presenter to receive and verify the <a>verifiable presentation</a>).
 
             <p>
 Other JOSE header parameters and <a>claim</a> names not specified herein can be
-used if their use is not explicitly discouraged. Implementers should not use JOSE
+used if their use is not explicitly discouraged. Implementers should not use JWT
 claim names for claims about the <code>credentialSubject</code> such as 
-email, birthdate, etc.
-There are no transformation rules for these and they would get lost after
-transformation. Instead, implementers MUST leave claims about the <a>subject</a> in the
-<code>credentialSubject</code> property of the <code>vc</code> claim, and 
-properties of the
-<a>verifiable credential<a> in
-the <code>vc</code> claim. Additional JOSE claim names such as <code>iat</code>,
-<code>nonce</code>, etc. are
-allowed but they will get lost after transformation unless there is an encoding/decoding
-rule defined. 
+email, birthdate, etc. There are no transformation rules for these and they 
+would get lost after decoding. Instead, implementers MUST add claims about the <a>subject</a> to the <code>credentialSubject</code> property of the JWT
+<code>vc</code> claim. 
             </p>
 
             <p>
@@ -3582,7 +3575,7 @@ the concepts outlined in Section
 <a href="#advanced-concepts">Advanced Concepts</a> (for example,
 <code>refreshService</code>, <code>termsOfUse</code>, and
 <code>evidence</code>). These concepts can be encoded as they are without any
-transformation, and can be added to the <code>vc</code> <a>property</a> of the
+transformation, and can be added to the <code>vc</code> <a>claim</a> of the
 JWT.
             </p>
 
--- a/index.html
+++ b/index.html
@@ -3562,9 +3562,18 @@ the Presenter to receive and verify the <a>verifiable presentation</a>).
 
             <p>
 Other JOSE header parameters and <a>claim</a> names not specified herein can be
-used if their use is not explicitly discouraged. Additional <a>verifiable 
-credential</a> <a>claims</a> MUST
-be added to the <code>credentialSubject</code> <a>property</a> of the JWT.
+used if their use is not explicitly discouraged. Implementers should not use JOSE
+claim names for claims about the <code>credentialSubject</code> such as 
+email, birthdate, etc.
+There are no transformation rules for these and they would get lost after
+transformation. Instead, implementers MUST leave claims about the <a>subject</a> in the
+<code>credentialSubject</code> property of the <code>vc</code> claim, and 
+properties of the
+<a>verifiable credential<a> in
+the <code>vc</code> claim. Additional JOSE claim names such as <code>iat</code>,
+<code>nonce</code>, etc. are
+allowed but they will get lost after transformation unless there is an encoding/decoding
+rule defined. 
             </p>
 
             <p>
--- a/index.html
+++ b/index.html
@@ -3561,12 +3561,16 @@ the Presenter to receive and verify the <a>verifiable presentation</a>).
             </ul>
 
             <p>
-Other JOSE header parameters and <a>claim</a> names not specified herein can be
-used if their use is not explicitly discouraged. Implementers should not use JWT
-claim names for claims about the <code>credentialSubject</code> such as 
-email, birthdate, etc. There are no transformation rules for these and they 
-would get lost after decoding. Instead, implementers MUST add claims about the <a>subject</a> to the <code>credentialSubject</code> property of the JWT
-<code>vc</code> claim. 
+Other JOSE header parameters and JWT <a>claim</a> names not specified herein can be
+used if their use is not explicitly discouraged. Implementers MUST add claims about
+the <a>subject</a> to the <code>credentialSubject</code> property of the JWT
+<code>vc</code> <a>claim</a>, and <a>properties</a> of the
+<a>verifiable credential<a> to the <code>vc</code> <a>claim</a>. Implementers
+are advised not to use JWT <a>claim</a> names (e.g. (<code>birthdate</code>,
+<code>nickname</code>, <code>email</code>, etc.) for <a>claims</a> about the
+<code>credentialSubject</code> because these would be lost after decoding
+unless there are encoding/decoding transformation rules defined for them in
+a subsequent standard or other document available to the implementers. 
             </p>
 
             <p>
--- a/index.html
+++ b/index.html
@@ -3554,14 +3554,15 @@ timestamp (<code>NumericDate</code>).
 in the <a>verifiable credential</a> <a>subject</a>.
               </li>
               <li>
-<code>aud</code> MUST represent the <code>subject</code> <a>property</a> of the
-consumer of the <a>verifiable presentation</a>.
+<code>aud</code> MUST represent the
+consumer of the <a>verifiable presentation</a> (i.e. the verifier).
               </li>
             </ul>
 
             <p>
 Other JOSE header parameters and <a>claim</a> names not specified herein can be
-used if their use is not explicitly discouraged. Additional <a>claims</a> MUST
+used if their use is not explicitly discouraged. Additional <a> verifiable 
+credential </a> <a>claims</a> MUST
 be added to the <code>credentialSubject</code> <a>property</a> of the JWT.
             </p>
 
@@ -3618,8 +3619,7 @@ done:
               <li>
 If <code>exp</code> is present, the UNIX timestamp MUST be converted to an
 [[!RFC3339]] <code>date-time</code>, and MUST be used to set the value
-of the <code>expirationDate</code> <a>property</a> of
-<code>credentialSubject</code> of the new JSON object.
+of the <code>expirationDate</code> <a>property</a> of the new JSON object.
               </li>
               <li>
 If <code>iss</code> is present, the value MUST be used to set the
--- a/index.html
+++ b/index.html
@@ -3565,13 +3565,11 @@ Other JOSE header parameters and JWT <a>claim</a> names not specified herein can
 used if their use is not explicitly discouraged. Implementers MUST add claims about
 the <a>subject</a> to the <code>credentialSubject</code> property of the JWT
 <code>vc</code> <a>claim</a>, and <a>properties</a> of the
-<a>verifiable credential<a> to the <code>vc</code> <a>claim</a>. Implementers
-are advised not to use JWT <a>claim</a> names (e.g. (<code>birthdate</code>,
-<code>nickname</code>, <code>email</code>, etc.) for <a>claims</a> about the
-<code>credentialSubject</code> because these would be lost after decoding
-unless there are encoding/decoding transformation rules defined for them in
-a subsequent standard or other document available to the implementers. 
-            </p>
+<a>verifiable credential<a> to the <code>vc</code> <a>claim</a>. Such <a>claims</a>
+about the <a>subject</a> and <a>properties</a> of the <a>verifiable credential<a>
+MAY be echoed/repeated, but should not be placed solely in the JOSE header parameters
+or JWT <a>claim</a> names not specified herein. 
+           </p>
 
             <p>
 This version of the specification defines no JWT-specific encoding rules for
--- a/index.html
+++ b/index.html
@@ -3554,15 +3554,16 @@ timestamp (<code>NumericDate</code>).
 in the <a>verifiable credential</a> <a>subject</a>.
               </li>
               <li>
-<code>aud</code> MUST represent the
-consumer of the <a>verifiable presentation</a> (i.e. the verifier).
+<code>aud</code> MUST represent (i.e., identify) the intended audience
+of the <a>verifiable presentation</a> (i.e., the Verifier(s) intended by
+the Presenter to receive and verify the <a>verifiable presentation</a>).
               </li>
             </ul>
 
             <p>
 Other JOSE header parameters and <a>claim</a> names not specified herein can be
-used if their use is not explicitly discouraged. Additional <a> verifiable 
-credential </a> <a>claims</a> MUST
+used if their use is not explicitly discouraged. Additional <a>verifiable 
+credential</a> <a>claims</a> MUST
 be added to the <code>credentialSubject</code> <a>property</a> of the JWT.
             </p>
 


index.html
```diff
@@ -3554,14 +3554,15 @@ <h5>JWT Encoding</h5>
...
+<code>aud</code> MUST represent the
+consumer of the <a>verifiable presentation</a> (i.e. the verifier).
               </li>
             </ul>
 
             <p>
 Other JOSE header parameters and <a>claim</a> names not specified herein can be
-used if their use is not explicitly discouraged. Additional <a>claims</a> MUST
+used if their use is not explicitly discouraged. Additional <a> verifiable 
+credential </a> <a>claims</a> MUST
```
davidlehn: remove whitespace after `<a>` and before `</a>`.

TallTed:
The third of these changes (circa lines 3620-3623) does appear to be a copy/paste error (i.e., it appears that `of <code>credentialSubject</code>` was accidentally left in place).

The others are not so clear to me as "typos".

If I understand the second (circa lines 3563-3566) correctly, I suggest your --
```
Additional <a> verifiable 
credential </a> <a>claims</a> MUST
be added to the <code>credentialSubject</code> <a>property</a> of the JWT.
```
-- instead become --
```
Such additional parameters 
and <a>claims</a> MUST be added to the <code>credentialSubject</code> 
<a>property</a> of the JWT.
```

As to the first (circa lines 3557-3558), I suggest --
```
<code>aud</code> MUST represent the
consumer of the <a>verifiable presentation</a> (i.e. the verifier).
```
-- instead become --
```
<code>aud</code> MUST represent (i.e., identify) the intended audience
of the <a>verifiable presentation</a> (i.e., the Verifier(s) intended by
the Presenter to receive and verify the <a>verifiable presentation</a>).
```

(Note -- a comma should always immediately follow `e.g.` and `i.e.`.)

index.html
David-Chadwick: Done, thanks

David-Chadwick:
@TallTed Happy to accept your improved text for the first change (the reason I did not make such a detailed change in the first instance is that I wanted to keep the edits to a minimum).

Concerning the second change, this was necessary due to the ambiguity over the word claim. I read the original text to mean a JWT claim, referring back to the previous sentence, but Oliver, the author, meant it to mean a VC claim. Hence the clarification was necessary to make it clear we are talking about VC claims (and not any other parameter of a VC such as ToU). Thus I would prefer not to include your text as I do not know what these additional parameters that you refer to are.

awoie:
@TallTed The others are language issues rather than typos.

One thing to note, no test cases have to be updated, no new implementation reports have to be submitted.

TallTed:
@David-Chadwick, @awoie -- 

Reading in context (i.e., referring back to the paragraphs which preceded it, particularly the start of the section, circa lines 3462-3476), the `<a>claim</a> names` in this sentence _must_ be `JWT <a>claim</a> names` (albeit they are being used to hold `<a>verifiable credential</a> <a>claims</a>`), since the point of this section is encoding a VC as a JWT.

How is an implementer to determine whether or not any `Other JOSE header parameters and [JWT] <a>claim</a> names` they encounter are `<a>verifiable credential</a> <a>claims</a>` or vice versa?  It appears to me that the answer in both cases must always be that ***they are*** because there's no way to know otherwise.

Further, note that we have previously said, `If no explicit rule is specified, [<a>verifiable credential</a>] <a>properties</a> are encoded [into a JWT] in the same way as with a standard <a>verifiable credential</a>, and are added to the <code>vc</code> <a>property</a> of the JWT.`  But here we appear to be making a contradictory explicit rule for all such previously unspecified `[<a>verifiable credential</a>] <a>properties</a>`, mandating that they instead be added to the `<code>credentialSubject</code> <a>property</a> of the JWT`?

Other readers (@msporny, @dlongley, @burnburn, @brentzundel ?) may be able to help clarify this ... or make other suggestions.

awoie:
@TallTed That is not the intention, not all verifiable credential `properties` will be added to the `credentialSubject`. Note, there is a difference between `properties` and `claims`.  As per the W3C VC spec a claim is defined as follow:

> An assertion made about a subject.

Additionally, the W3C VC spec defines the `credentialSubject` as follows:

> A verifiable credential contains claims about one or more subjects. This specification defines a credentialSubject property for the expression of claims about one or more subjects.

TallTed:
My confusion with this section persists. Contradictory and unclear instructions exist in this PR, and I think also in the current text. I am unable to provide improved wording, as I am unclear on which of the contradictory instructions are correct.

awoie:
@TallTed Implementers should not use JOSE claim names for claims about the `credentialSubject`. There is no transformation rule for that. Instead, implementers will add claims about the subject to the `credentialSubject`, and claims about the VC to the `vc` property. Additional JOSE claim names such as `iat`, `nonce`, etc. are allowed because would get lost between transformation except there was a encoding/decoding rule defined.

David-Chadwick:
@TallTed This is precisely why I initiated this PR - to remove all confusion. So please tell us where your confusion still exists then we can fix it. But if we do not know which bit of the text confuses you we cannot resolve it. Thanks

TallTed:
@David-Chadwick - Incorporating the meat of the paragraph @awoie just posted (["Implementers ... defined."](https://github.com/w3c/vc-data-model/pull/711#issuecomment-520464360)) would be a good start.

David-Chadwick:
@TallTed . Done. Take a look and see what you think.

index.html
```diff
@@ -3554,15 +3554,26 @@ <h5>JWT Encoding</h5>
...
+email, birthdate, etc.
+There are no transformation rules for these and they would get lost after
+transformation. Instead, implementers MUST leave claims about the <a>subject</a> in the
+<code>credentialSubject</code> property of the <code>vc</code> claim, and 
+properties of the
+<a>verifiable credential<a> in
+the <code>vc</code> claim. Additional JOSE claim names such as <code>iat</code>,
+<code>nonce</code>, etc. are
+allowed but they will get lost after transformation unless there is an encoding/decoding
+rule defined.
```
TallTed: The above reads significantly better.  I see some grammar issues, as well as some potentially problematic (we're definitely not doing CR#3!) "should" language (and some very distracting line breaks), so I suggest that this replace that --
```
used if their use is not explicitly discouraged. Implementers are advised to put
<a>claims</a> about the <a>subject</a> in the <code>credentialSubject</code>
<a>property</a> of the <code>vc</code> <a>claim</a>, and <a>properties</a> of the
<a>verifiable credential<a> itself in the <code>vc</code> <a>claim</a> (alongside
the <code>credentialSubject</code> <a>property</a>). Implementers are advised to
not use JOSE <a>claim</a> names (such as <code>email</code>, <code>birthdate</code>,
etc.) for <a>claims</a> about the <code>credentialSubject</code>, because these
would be lost after transformation, as there are no transformation rules for them.
Additional JOSE <a>claim</a> names (such as <code>iat</code>, <code>nonce</code>,
etc.) are allowed, but they will likewise be lost after transformation unless
there is an encoding/decoding rule defined.
```

I find myself wondering where and how the (apparently possible) additional encoding/decoding rule can be defined by the implementer?  Or should that say "unless there is an encoding/decoding rule defined in this document"?  In which case, I suggest --
```
Additional JOSE <a>claim</a> names (<code>nickname</code>, <code>nonce</code>,
etc.) are allowed, but they will likewise be lost after transformation unless
there is an encoding/decoding rule defined.
```


At this point, I realized that `<code>iat</code>` has been explicitly discussed, so it probably does not belong in a list of _"**Additional** JOSE claim names"._  In order to pick another example, I went to the [JSON Web Token (JWT) RFC (#7519)](https://tools.ietf.org/html/rfc7519), where it appears that JOSE does not have Claims, but Header Parameters.  JWT appears to have Claims.  This terminology issue will ripple through all our documents.

I am not an expert on JWT, JOSE, etc.  I am just trying to comprehend and thereby help clarify this writing, and continuing to struggle with it, because of inconsistencies like the above.

awoie: Hi, the IANA claim/header registry maintains a list of these registered claim names and headers. The IANA table is also referenced in the JWT spec: https://www.iana.org/assignments/jwt/jwt.xhtml

TallTed: @awoie - Both the JWT spec and the IANA table refer to **_JWT Claims_** and/or **_JSON Web Token Claims_**; neither mentions **_JOSE Claims_**.  Searching the broader web, there are nearly no uses of *JOSE Claims* -- near enough zero to make me question the pages that *do* use that phrasing.

index.html
```diff
@@ -3554,15 +3554,26 @@ <h5>JWT Encoding</h5>
...
+of the <a>verifiable presentation</a> (i.e., the Verifier(s) intended by
+the Presenter to receive and verify the <a>verifiable presentation</a>).
               </li>
             </ul>
 
             <p>
 Other JOSE header parameters and <a>claim</a> names not specified herein can be
-used if their use is not explicitly discouraged. Additional <a>claims</a> MUST
-be added to the <code>credentialSubject</code> <a>property</a> of the JWT.
+used if their use is not explicitly discouraged. Implementers should not use JOSE
```
msporny: "should not" -> are advised not to

index.html
David-Chadwick: @TallTed. I have now been through the JWT section and tried to make all the terms identical throughout to reduce confusion. Consequently there is no mention of JOSE claims in the revised text (and there never was in the original CR either), JWT claims are always referred to as JWT claims and not as JWT properties (which did occur a couple of times in the original CR), properties refer to VC properties, and the conformance statements (MUSTs) remain the same in number and intention as in the original CR. Can you please check if you are happy with this revised version. Many thanks.

TallTed: @awoie @David-Chadwick - Following today's call, some more revision...

```
Other JOSE header parameters and JWT <a>claim</a> names not specified herein can
be used if their use is not explicitly discouraged. Implementers are advised to
put <a>claims</a> about the <a>subject</a> in the <code>credentialSubject</code>
<a>property</a> of the <code>vc</code> <a>claim</a>, and <a>properties</a> of the
<a>verifiable credential<a> itself directly into the <code>vc</code> <a>claim</a>
(i.e., parallel to the <code>credentialSubject</code> <a>property</a>).
Implementers are advised to not use JWT <a>claim</a> names (<code>birthdate</code>,
<code>email</code>, etc.) for <a>claims</a> about the <code>credentialSubject</code>,
because these would be lost after transformation, as there are no transformation
rules for them. Use of additional JWT <a>claim</a> names (that is, JWT <a>claim</a>
names not explicitly discussed in this specification, such as <code>nickname</code>,
<code>nonce</code>, etc.) is allowed, but they will likewise be lost after
transformation unless there is an encoding/decoding rule defined.
```

I still don't have clarity on the last sentence above -- i.e., *where, when, and how* are these encoding/decoding rules defined?  Can the implementer or some third party define them at some point in the future?  Or does this refer only to _the rules in this specification?_  In the latter case, I would change the last sentence to simply --
```
Use of additional JWT <a>claim</a> names (that is, JWT <a>claim</a>
names not explicitly discussed in this specification, such as <code>nickname</code>,
<code>nonce</code>, etc.) is allowed, but they will likewise be lost after
transformation.
```

David-Chadwick: @TallTed It appears that your only outstanding issue now is extensibility i.e. how can a JWT be extended with new claim names that represent new properties of a VC subject, that have not already been defined in our VC data model recommendation. Is this correct?
If so, this is an issue that Oliver and myself have already discussed, and we saw that this should be a section for the implementation guide, which describes how VCs can be extended with new @context values, new credentialSchema values and new JWT claim names, perhaps registered with IANA. This was mentioned in our weekly call on 6 Aug when Manu asked me if the text that @dlongley had already provided on extensibility was sufficient. Having now read this text, the answer is No, it is not yet sufficient. It is fine for @context, but we are still missing text for extending credentialSchemas and JWT claim names.
So to resolve your outstanding issue (if this is indeed your only one) can I propose the following:
i) add the following to the end of your text "...defined in a subsequent standard or other document available to the VC ecosystem users."
ii) Oliver and myself write this extensibility section for the Implementation Guide

index.html
David-Chadwick: Done, thanks

index.html
TallTed: I cannot promise that new wording (or re-reading of existing wording) will not reveal a new issue; realizations happen when they do.  That said, I try to raise issues to the attention of the group immediately when they appear before me, so whatever I've posted reflects my current concerns.

*That* said, my concern here is not "extensibility" per se.  It is with the wording of this paragraph, which doesn't point to the (non-normative) Implementation Guide nor any other area as the source (extensible or not) of the "definitions" of the "encoding/decoding rules" which are necessary to preserve the values of any `Other JOSE header parameters and JWT <a>claim</a> names not specified herein`.

Adding "defined in a subsequent standard or other document available to the VC ecosystem users" to my previous suggestions will not suffice, as there is no way to assure that any given Verifier or Holder/Presenter will have acquired or adopted such a "subsequent standard or other document" -- so there is no way to assure that such extension `JOSE header parameters and JWT <a>claim</a> names` will be preserved even by fully compliant VC processors.

I think the appropriate advice may be that `<a>claims</a> about the <a>subject</a>` (which we've said should be encoded `in the <code>credentialSubject</code> <a>property</a> of the <code>vc</code> <a>claim</a>`), and `<a>properties</a> of the <a>verifiable credential<a> itself` (which we've said should be encoded `directly into the <code>vc</code> <a>claim</a>`) ***`MAY be echoed/repeated but should not be placed solely in JOSE header parameters or JWT <a>claim</a> names not specified herein, because these will not be preserved through VC/VP transformation to/from JWT encoding.`***

awoie: @David-Chadwick We could put a non-normative note into the document that we expect implementers to use that mechanism for VC claims. 

@TallTed Sounds good to me. That is exactly what happens. I'm happy with the spec text then.

TallTed: @awoie @David-Chadwick - OK. I think this is then the final paragraph, ready to copy-and-paste to the PR branch --
```
Other JOSE header parameters and JWT <a>claim</a> names not specified herein
can be used if their use is not explicitly discouraged. Implementers MUST put
<a>claims</a> about the <a>subject</a> in the <code>credentialSubject</code>
<a>property</a> of the <code>vc</code> <a>claim</a>, and <a>properties</a> of
the <a>verifiable credential<a> itself directly into the <code>vc</code>
<a>claim</a> (i.e., parallel to the <code>credentialSubject</code>
<a>property</a>). Such <a>claims</a> about the <a>subject</a> and
<a>properties</a> of the <a>verifiable credential<a> MAY be echoed/repeated,
but should not be placed solely, in JOSE header parameters or JWT <a>claim</a>
names not specified herein, because these will not be preserved through
transformation to or from JWT encoding.
```

David-Chadwick: Ok, so we are now mandating that all additional parameters must be placed in the vc claim, and optionally may be duplicated as JWT claims. Nevertheless if they are duplicated then transformation rules may need to be defined so that the duplications are exactly equal. This has already been done in the PR to the Implementation Guide for the aud claim. So the last sentence above is incorrect, since they will be preserved if such transformation rules are defined. Therefore I have simply removed everything from ",because....." to the end.

awoie: I would remove the "MAY" and use "can" as it could be just a non-normative statement. 

@David-Chadwick @TallTed Are you now both fine with language?

TallTed: No, I am not yet "fine with the language/spec text" because --

* the transformation rules ARE NOT DEFINED except for the parameters and claim names discussed in the Data Model spec.  Also, we have *never* mandated that "duplications are exactly equal", which is part of why we may encounter *conformant* transformations that add and/or remove JOSE headers and/or JWT claims that we have not explicitly discussed.

* I believe the MAY language is OK, even in this phase of CR, because it doesn't require changing any implementation, regardless of what they now do.

* @David-Chadwick - Note that you did not copy-and-paste my text block, which I provided for exactly that purpose, so you lost a very important comma (following "solely"), some important markup (at least one `<a>claims</a>` is not currently `<a>`-wrapped), and have effectively if not actively made other changes to my carefully drafted text.



So you can compare, here is your latest version (with line wraps adjusted to the github interface for PR discussion, but no other changes from the current PR) --
```
Other JOSE header parameters and JWT <a>claim</a> names not specified
herein can be used if their use is not explicitly discouraged.
Implementers MUST add claims about the <a>subject</a> to the
<code>credentialSubject</code> property of the JWT <code>vc</code>
<a>claim</a>, and <a>properties</a> of the <a>verifiable credential<a>
to the <code>vc</code> <a>claim</a>. Such <a>claims</a> about the
<a>subject</a> and <a>properties</a> of the <a>verifiable credential<a>
MAY be echoed/repeated, but should not be placed solely in the JOSE
header parameters or JWT <a>claim</a> names not specified herein. 
```
-- and my latest version (with two additional fixes, being closing `</a>` where there were opening `<a>`), by which I stand --
```
Other JOSE header parameters and JWT <a>claim</a> names not specified
herein can be used if their use is not explicitly discouraged.
Implementers MUST put <a>claims</a> about the <a>subject</a> in the
<code>credentialSubject</code> <a>property</a> of the <code>vc</code>
<a>claim</a>, and <a>properties</a> of the <a>verifiable credential</a>
itself directly into the <code>vc</code> <a>claim</a> (i.e., parallel
to the <code>credentialSubject</code> <a>property</a>). Such
<a>claims</a> about the <a>subject</a> and <a>properties</a> of the
<a>verifiable credential</a> MAY be echoed/repeated, but should not be
placed solely, in JOSE header parameters or JWT <a>claim</a> names not
specified herein, because these will not be preserved through
transformation to or from JWT encoding.
```

awoie: @TallTed You won't be able to add all these transformation rules. You cannot add rules for properties that are not in the W3C spec. If the spec will use additional properties, additional transformation rules will be added in the future. I think, it should be fine to say that these properties will then get lost in case there is no explicit transformation rule defined. For this reason, people are advised to add claims to the `vc` and `credentialSubject` property. Are you fine with that in general?

awoie: IMO, this is not a bug, just a limitation of the spec.

TallTed: @awoie - "These properties will then get lost in case there is no explicit transformation rule defined."  Right, but, _where are the explicit transformation rules defined?_  If they all must be in the Data Model Specification (as I believe to be the case), then we should say that.  If they might be found somewhere else in future, then we MUST say *that,* and we must tell implementers how to stay up-to-date with any new transformation rules that might be developed to address any new properties -- and how one user/implementation can tell whether another user/implementation is working with a different set of transformation rules.

In other words -- I think the explicit rules are and must remain in the Data Model Spec, so we should simply say "if there's no transformation rule *here* for your property, then that property will get lost on transformation, so don't use it."

Re-re-re-re-re-reading this paragraph, I think the first sentence is redundant to some of what follows, so I have a bit more revision --

```
Implementers MUST put <a>claims</a> about the <a>subject</a> in the
<code>credentialSubject</code> <a>property</a> of the <code>vc</code>
<a>claim</a>, and <a>properties</a> of the <a>verifiable credential</a>
itself directly into the <code>vc</code> <a>claim</a> (i.e., parallel
to the <code>credentialSubject</code> <a>property</a>). Such
<a>claims</a> about the <a>subject</a> and <a>properties</a> of the
<a>verifiable credential</a> MAY be echoed/repeated, but should not be
placed solely, in JOSE header parameters or JWT <a>claim</a> names not
specified herein (so long as their use is not explicitly discouraged),
because these will not be preserved through transformation to or from
JWT encoding.
```

That last sentence may be unclear because of its complexity. It could be split as --

```
Implementers MUST put <a>claims</a> about the <a>subject</a> in the
<code>credentialSubject</code> <a>property</a> of the <code>vc</code>
<a>claim</a>, and <a>properties</a> of the <a>verifiable credential</a>
itself directly into the <code>vc</code> <a>claim</a> (i.e., parallel
to the <code>credentialSubject</code> <a>property</a>). Such
<a>claims</a> about the <a>subject</a> and <a>properties</a> of the
<a>verifiable credential</a>
            </p>
            <ul>
              <li>
MAY be echoed/repeated in JOSE header parameters or JWT <a>claim</a>
names not specified herein, so long as their use is not explicitly
discouraged
              </li>
              <li>
SHOULD NOT be placed solely in JOSE header parameters or JWT
<a>claim</a> names not specified herein, because these will not be
preserved through transformation to or from JWT encoding
              </li>
            </ul>
```

awoie: @TallTed In order to address your first bullet point, what needs to be changed in the spec? IMO, the text above addresses that concern by informing the implementer about it. IMO, no further language change needed. I'm also ok with MAY if there is no concern. @David-Chadwick could you use TallTed's latest version of the text above. Then, I believe we are fine with TallTed's third concern.

awoie: @TallTed @David-Chadwick I'm trying to untangle this conversation because I think it is important that we can agree on the language.

I agree with TallTed and I am fine with keeping --> this addresses his first bullet point:
> because these will not be preserved through transformation to or from JWT encoding

I'm also ok with "MAY" if there is no other implication.

I'm also ok with your remarks in your third bullet point. 

@David-Chadwick could you please update the PR accordingly? I don't see a reason not to. 
@TallTed In case, the PR was updated accordingly would you then be fine with the PR?

awoie: I'm happy with your last text except that I don't see the need for introducing normative language. That normative language won't have an impact on existing or future implementations. We could use the following instead:

> can be echoed/repeated in JOSE header parameters or JWT <a>claim</a>
names not specified herein, so long as their use is not explicitly
discouraged

and 

> It is expected that JOSE header parameters or JWT
<a>claim</a> names not specified herein won't be placed solely, because these will not be
preserved through transformation to or from JWT encoding

I don't see this as a substantial change.

awoie: @msporny @dlongley @brentzundel @burnburn I would love to hear your opinion on this.

David-Chadwick: @TallTed Can you please fill in @awoie's doodle poll so that we can discuss the text by conference call. We are getting nowhere fast with all these textual changes and it is massively time consuming to little good purpose. A short conference call should prove to be much more productive. Thanks

TallTed: I'm not strongly attached to the normative language. I do think it is the best way to express what needs to be said, and that it would not force another CR, but as you both feel otherwise, I've edited it out here --

```
Implementers MUST put <a>claims</a> about the <a>subject</a>
into the <code>credentialSubject</code> <a>property</a> of the
<code>vc</code> <a>claim</a>, and <a>properties</a> of the
<a>verifiable credential</a> itself directly into the
<code>vc</code> <a>claim</a> (i.e., parallel to the
<code>credentialSubject</code> <a>property</a>). Such <a>claims</a>
about the <a>subject</a> and <a>properties</a> of the
<a>verifiable credential</a> can be echoed/repeated in JOSE header
parameters or JWT <a>claim</a> names not specified herein, so long
as their use is not explicitly discouraged. It is expected that
such <a>claims</a> about the <a>subject</a> and <a>properties</a>
of the <a>verifiable credential</a> will not be found only in JOSE
header parameters or JWT <a>claim</a> names not specified herein
(i.e., that they will be echoed as described above), because those
additional JOSE header parameters or JWT <a>claim</a> names would
not be preserved through transformation to and from JWT encoding.
```

index.html
```diff
@@ -3478,7 +3478,7 @@ <h5>JWT Encoding</h5>
             <p>
 If no explicit rule is specified, <a>properties</a> are encoded in the same way
 as with a standard <a>verifiable credential</a>, and are added to the
-<code>vc</code> <a>property</a> of the JWT. As with all JWTs, the JWS-based
+<code>vc</code> <a>claim</a> of the JWT. As with all JWTs, the JWS-based
```
msporny: Changes for property->claim are fine.

index.html
```diff
@@ -3554,24 +3554,30 @@ <h5>JWT Encoding</h5>
 in the <a>verifiable credential</a> <a>subject</a>.
               </li>
               <li>
-<code>aud</code> MUST represent the <code>subject</code> <a>property</a> of the
-consumer of the <a>verifiable presentation</a>.
+<code>aud</code> MUST represent (i.e., identify) the intended audience
```
msporny: This change could be considered substantive, although I expect no implementations would change.

index.html
```diff
@@ -3554,24 +3554,30 @@ <h5>JWT Encoding</h5>
...
               </li>
             </ul>
 
             <p>
-Other JOSE header parameters and <a>claim</a> names not specified herein can be
-used if their use is not explicitly discouraged. Additional <a>claims</a> MUST
-be added to the <code>credentialSubject</code> <a>property</a> of the JWT.
-            </p>
+Other JOSE header parameters and JWT <a>claim</a> names not specified herein can be
+used if their use is not explicitly discouraged. Implementers MUST add claims about
```
msporny: This is a normative change.

index.html
```diff
@@ -3554,24 +3554,30 @@ <h5>JWT Encoding</h5>
...
+the <a>subject</a> to the <code>credentialSubject</code> property of the JWT
+<code>vc</code> <a>claim</a>, and <a>properties</a> of the
+<a>verifiable credential<a> to the <code>vc</code> <a>claim</a>. Such <a>claims</a>
+about the <a>subject</a> and <a>properties</a> of the <a>verifiable credential<a>
+MAY be echoed/repeated, but should not be placed solely in the JOSE header parameters
```
msporny: This is additional normative guidance, and I'm concerned that it'll lead to non-interoperable JWT implementations. It's basically saying "some implementations may echo/repeat JOSE header parameters or JWT claims... but others may not echo/repeat, or may not even echo/repeat the same things." That will inevitably lead to some implementations doing the opposite of what other implementations do. I suggest we don't change this as it doesn't clarify things and only further confuses the issue.

index.html
```diff
@@ -3618,8 +3624,7 @@ <h5>JWT Decoding</h5>
               <li>
 If <code>exp</code> is present, the UNIX timestamp MUST be converted to an
 [[!RFC3339]] <code>date-time</code>, and MUST be used to set the value
-of the <code>expirationDate</code> <a>property</a> of
-<code>credentialSubject</code> of the new JSON object.
+of the <code>expirationDate</code> <a>property</a> of the new JSON object.
```
msporny: This is a normative change as it changes where the data goes.

msporny:
This PR is replaced by PRs #717, #718, #719, and #720.

PR #737, opened 2020-04-06
Fix inconsistent issuanceDate

### LINKED ISSUE #736: Examples issuanceDate is invalid
`issuanceDate` is set to `2010-01-01T19:73:24Z" ` in examples:
1, 2, 11, 45, 46
This is invalid because: 
`73` is not a valid number of minutes.

Other issuanceDate in examples = `2010-01-01T19:23:24Z`

@tplooker @kdenhartog can you take care of this

tagging @msporny for awareness

Liam-Tait:
issuanceDate was invalid, as 73 is not a valid number of minutes

closes #736

Signed-off-by: Liam Tait <liam.tait@mattr.global>

***
<a href="https://pr-preview.s3.amazonaws.com/Liam-Tait/vc-data-model/pull/737.html" title="Last updated on Apr 6, 2020, 5:46 AM UTC (87638fa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/737/0e58fb9...Liam-Tait:87638fa.html" title="Last updated on Apr 6, 2020, 5:46 AM UTC (87638fa)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -757,7 +757,7 @@ usually a digital signature.
 	       Proof Graph on the bottom.  The Credental Graph has
 	       Credential 123 with 4 properties: 'type' of value
 	       AlumniCredential, 'issuer' of Example University,
-	       'issuanceDate' of 2010-01-01T19:73:24Z, and
+	       'issuanceDate' of 2010-01-01T19:23:24Z, and
 	       credentialSubject of Pat, who has an alumniOf property
 	       with value of Example University.  The Proof Graph has
 	       Signature 456 with 5 properties: 'type' of
@@ -925,7 +925,7 @@ discount from a university. In the example below, Pat receives an alumni
   <span class='comment'>// the entity that issued the credential</span>
   "issuer": "https://example.edu/issuers/565049",
   <span class='comment'>// when the credential was issued</span>
-  "issuanceDate": "2010-01-01T19:73:24Z",
+  "issuanceDate": "2010-01-01T19:23:24Z",
   <span class='comment'>// claims about the subjects of the credential</span>
   "credentialSubject": {
     <span class='comment'>// identifier for the only subject of the credential</span>
@@ -990,7 +990,7 @@ the <a>verifier</a> and <a>verified</a>.
     "id": "http://example.edu/credentials/1872",
     "type": ["VerifiableCredential", "AlumniCredential"],
     "issuer": "https://example.edu/issuers/565049",
-    "issuanceDate": "2010-01-01T19:73:24Z",
+    "issuanceDate": "2010-01-01T19:23:24Z",
     "credentialSubject": {
       "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
       "alumniOf": {
@@ -1742,7 +1742,7 @@ the signing date. The example below uses RSA digital signatures.
   "id": "http://example.gov/credentials/3732",
   "type": ["VerifiableCredential", "UniversityDegreeCredential"],
   "issuer": "https://example.edu",
-  "issuanceDate": "2010-01-01T19:73:24Z",
+  "issuanceDate": "2010-01-01T19:23:24Z",
   "credentialSubject": {
     "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
     "degree": {
@@ -2018,7 +2018,7 @@ the desired <a>verifiable presentation</a>.
         <figure>
           <img style="margin: auto; display: block; width: 50%;"
             src="diagrams/claim-example-2.svg" alt="Pat has a property
-            dateOfBirth whose value is 2010-01-01T19:73:24Z">
+            dateOfBirth whose value is 2010-01-01T19:23:24Z">
           <figcaption style="text-align: center;">
 A basic claim expressing that Pat's date of birth is 2010-01-01T19:23:24Z. Date
 encoding would be determined by the schema.
@@ -5769,7 +5769,7 @@ verifiable credential that was passed to it by the subject">
       "id": "http://example.gov/credentials/3732",
       "type": ["VerifiableCredential", "PrescriptionCredential"],
       "issuer": "https://example.edu",
-      "issuanceDate": "2010-01-01T19:73:24Z",
+      "issuanceDate": "2010-01-01T19:23:24Z",
       "credentialSubject": {
         "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
         "prescription": {....}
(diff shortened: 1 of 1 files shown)


iherman:
(Setting the change as editorial, clearing the IPR issue.)

iherman:
> curiously, I can merge this PR. Should I be able to do that @iherman ?

@kdenhartog thanks for noticing that. The current set up has been done automatically by the W3C system team, and we may not want to keep that. I will discuss this separately with @burnburn

burnburn:
@kdenhartog @peacekeeper This should now be fixed.  Please let us know if you still have merge permissions.

OR13:
If I could have merged this PR, I would have :)

tplooker (approved):

OR13 (approved):

iherman:
My apologies not to have noticed a minor formal problem before.

The PR would lead to, eventually, republishing the Rec. Which is fine, but I think there should be a clear set of breadcrumbs to the reason of the changes, more explicitly than buried in (by then closed) GitHub PR. I think two things should happen.

- An official [errata management](https://w3c.github.io/vc-data-model/errata.html) has been set up for this WG/Recommendation, and I think that protocol should be followed even for editorial changes.
- Any change may have to be recorded in the document in a separate 'Changes' section. That is the practice in most W3C official documents (see, e.g., a [similar situation in JSON-LD](https://w3c.github.io/json-ld-syntax/#changes-from-rec1)). Not worth a detailed description when it is not necessary, but it is a good habit to take to have this.

However obvious the changes may be (and this is one), any change on a standard text must be handled with care.

Cc @msporny @burnburn @brentzundel @dlongley

OR13:
@iherman thanks for the explanation!

sounds like we need to do some issue management, and test updates before this can be merged.

I would suggest keeping the invalid date and challenging implementers to comprehend dates correctly... this should have been caught by tests.

msporny (approved):

msporny:
@iherman -- agree with your reading that we need to be careful w/ all future PR merges. That said, this is purely editorial -- just changes examples, yes, we should note that, but I'm hesitant to make the PR process that heavyweight for something being managed by the CCG at present.

In other words, merging isn't going to cause an auto-publish to TR-space, is it? 

What if we do this for the process:

1. Editors can feel free to merge as appropriate, there will be a clear commit history of changes made to spec from a particular date.
2. Editors are required to translate that into errata statements AND into the in-spec changelog since last version before publication to TR-space.

Both items above can happen in parallel to issue processing. 

@iherman -- thoughts on that process?

iherman:
> @iherman -- agree with your reading that we need to be careful w/ all future PR merges. That said, this is purely editorial -- just changes examples, yes, we should note that, but I'm hesitant to make the PR process that heavyweight for something being managed by the CCG at present.

Yes, this one is editorial, but #752 is not. Whatever we discuss here is incomplete for that case.

> 
> In other words, merging isn't going to cause an auto-publish to TR-space, is it?

As far as I can see, ECHIDNA is not set up (and, I believe it would not work for a Rec anyway, but I have not checked).

> 
> What if we do this for the process:
> 
> 1. Editors can feel free to merge as appropriate, there will be a clear commit history of changes made to spec from a particular date.
> 2. Editors are required to translate that into errata statements AND into the in-spec changelog since last version before publication to TR-space.
> 
> Both items above can happen in parallel to issue processing.

Just some comments

- I would prefer to start with the first step of (2) and leave the erratum out there for a few days before a merge is done. Leave time for the community. In other words, as soon as an issue is raised (which will eventually lead to a PR), that issue should be marked as a proposed errata by the editor. Then, if a merge is agreed upon, the corresponding issue should be set to Erratum (it can be closed, the errata process would still refer to it)
- I am not sure that he commit history is enough as a record for the changes. I believe the document itself should have a change section where the changes are marked, albeit in a very concise form (it can be as simple as referring to the erratum issue). Digging through the github commit records is way too demanding for many.

My remaining question is different (and not for me to decide): who makes the decision on a PR being merged or not. Of course, for an editorial one like this one it is not a big deal. But I am not sure what the process on #752 is: I have not seen a larger discussion on the CCG or elsewhere. But this is not for me to decide, it is for the WG. @burnburn?

msporny:
@iherman wrote:
>  I am not sure what the process on #752 is: I have not seen a larger discussion on the CCG or elsewhere.

Ah, now I see the issue. You are probably not aware that the CCG has already worked on and approved a process for updating the VC Data Model specification (and I believe the VCWG delegated the finalization of this process to the W3C CCG): https://github.com/w3c-ccg/w3c-ccg.github.io/blob/master/VCWGmaintprocess.md

What you're suggesting above changes that... which would require us to go through an update to the process (the first time we did this, it took many, many months).

Thoughts, @iherman? :)

iherman:
@msporny,

> @iherman wrote:
> 
> > I am not sure what the process on #752 is: I have not seen a larger discussion on the CCG or elsewhere.
> 
> Ah, now I see the issue. You are probably not aware that the CCG has already worked on and approved a process for updating the VC Data Model specification (and I believe the VCWG delegated the finalization of this process to the W3C CCG): https://github.com/w3c-ccg/w3c-ccg.github.io/blob/master/VCWGmaintprocess.md

Ah. I indeed did not see this document.

(A small rant: the fact that I did not see it may be the result of me getting old and stupid, but may also be the result of not properly advertised among WG members. Or both. We should try to remedy the latter; the former is doomed...)

> What you're suggesting above might change that... 

Not significantly. I have only a two comments (but I think I am repeating myself).

1. I understand, and I agree, that new features that are proposed for inclusion into the new documents, are not really errata. This is clearly the case for #752, for example and, I think, these are the cases concentrated on by that set of rules. But other cases like this issue (#737) is, I believe, an erratum in the current spec and, in my view, should be annotated as such. The document errata is the place where the public at large can check their own understanding of the _current_ spec and discover some known problems. Why else would it be there?

2. As I said in https://github.com/w3c/vc-data-model/pull/737#issuecomment-734082004, in my view any change of the document should also be documented _in the document itself_. Nothing fancy.

> 
> which would require us to go through an update to the process (the first time we did this, it took many, many months). We'll need to iron out what you feel needs to be done as W3C Staff Contact and what the CCG approved as the process. The more we can keep these things in the hands of the Editors / Chairs / staff contact, the better off we'll be.

As I said, I do not think those two remarks fundamentally change anything. (1) is just a clarification on the role of the errata management for, well, errata (and not new features); (2) is just a rule of good behaviour for editors:-). 

---

P.S. You should check the formatting of the bullet points on the process page above. I had to look at the markdown source to make some sense of it.

dlongley (approved):

kdenhartog:
multiple positive reviews, no objections, merging

kdenhartog:

Issue #105, opened 2018-02-06
Holders and Identifiers


David-Chadwick:
The current text in 1.3 says 
Holders SHOULD be able to easily control and own their own identifiers.

This applies equally to subjects, perhaps even more so than holders,
since VCs apply to subjects and contain subject IDs. VCs do not contain holder IDs.
Suggest either replace Holders with Subjects or add the requirement for subjects

Subjects SHOULD be able to easily control and own their own identifiers.

stonematt:
does everyone agree and is this true?

dlongley:
I think it's fine to add:

"Subjects SHOULD be able to easily control and own their own identifiers."

As a requirement. Nothing in the data model would prevent that capability and the spec's agnosticism with respect to identifiers generally supports any such statement, regardless of role.

David-Chadwick:
This text has not yet made it into the latest version (13 Feb 18) so it is not closed yet

David-Chadwick:
This text has not yet made it into the latest version (18 April 18). Worse still the latest text says 

identifier registry
    Mediates the creation and verification of subject identifiers.

So from the above, there is no mention of issuers or holders having identifiers as well. Nevertheless the following text is still present

Holders SHOULD be able to easily control and own their own identifiers.

This will be confusing to new readers of the specification.

burnburn:
@David-Chadwick please take a look and let us know what concerns remain given the current text.

David-Chadwick:
@burnburn I do not believe that the text has yet made it into the data model document.

David-Chadwick:
@msporny This textual change is such a simple clarification, so I wonder why it is taking so long to make it into the data model document?

TallTed:
I think it is good this hasn't gone in yet, as it is not a simple clarification.  I am sorry not to have noticed this sooner.

"Anyone can say anything about anything" rather contradicts the idea that "Subjects SHOULD be able to easily control and own their own identifiers."  

In fact, I think that the existing "Holders SHOULD be able to easily control and own their own identifiers" is also problematic. 

As an *Issuer,* I may Identify the Subjects of my statements however I like, verifiably or otherwise.  Consider the University identifying its Professors, which may know that it has two "John Q Smythe" to deal with, while one of those believes himself to be the only "John Q Smythe" in the universe, and so wants a simpler identifier than the University intends needs to use in order to differentiate.

Putting either Subject or Holder in control of the Subject Identifiers puts an unreasonable burden on Issuers, at best.

[...]

PR #780, opened 2021-06-14
Clarified subtitle of Data Model

TallTed:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/780.html" title="Last updated on Dec 9, 2021, 3:02 AM UTC (473a81a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/780/6c429cf...473a81a.html" title="Last updated on Dec 9, 2021, 3:02 AM UTC (473a81a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -19,7 +19,7 @@
         shortName: "vc-data-model",
 
         // subtitle for the spec
-        subtitle: "Expressing information with verifiable provenance",
+        //subtitle: "Expressing information with verifiable provenance",
 
         // if you wish the publication date to be other than today, set this
 //        publishDate: "2019-09-03",
--- a/index.html
+++ b/index.html
@@ -19,7 +19,7 @@
         shortName: "vc-data-model",
 
         // subtitle for the spec
-        subtitle: "Expressing information with verifiable provenance on the Web",
+        subtitle: "Expressing information with verifiable provenance",
 
         // if you wish the publication date to be other than today, set this
 //        publishDate: "2019-09-03",
--- a/index.html
+++ b/index.html
@@ -19,7 +19,7 @@
         shortName: "vc-data-model",
 
         // subtitle for the spec
-        subtitle: "Expressing verifiable information on the Web",
+        subtitle: "Expressing information with verifiable provenance on the Web",
 
         // if you wish the publication date to be other than today, set this
 //        publishDate: "2019-09-03",


index.html
```diff
@@ -19,7 +19,7 @@
         shortName: "vc-data-model",
 
         // subtitle for the spec
-        subtitle: "Expressing verifiable information on the Web",
+        subtitle: "Expressing information with verifiable provenance on the Web",
```
msporny, line 22: I'm wondering if we can go with something more pedestrian here... for example: "Knowing where your information comes from on the Web" -- "Manufacturer labels for information on the Web" -- "Ensuring tamper resistant publication of information on the Web."

TallTed: "Verifiably expressing the origin and transfer history of information on the Web"?

TallTed: (Maybe we can add a bit of informative text to the body somewhere, defining information provenance as its origin and transfer history, and linking out to PROV-O?)

brentzundel: Since the title is "Verifiable Credentials Data Model v1.0" could the subtitle be "data structures for verifiable information" ?

TallTed: Except it's not verifiable _information_.

I still don't think provenance is that much of a mystery, and _that's_ what it is we're establishing -- a way for a Reader/Verifier to be cryptographically certain that the declared Asserter/Issuer made the Assertions/Claims found in the Verifiable Credential.

brentzundel: "data structures for information provenance" ?

TallTed: Looks a lot like my suggestion above, i.e., "Expressing information with verifiable provenance [on the Web]" where the last phrase could be kept or removed, since I don't think VCs are actually limited to Web use...

dlongley (approved):

msporny (approved):
An improvement... I'm wondering if "on the Web" applies anymore since we're now seeing these things used "off of the Web"... embedded in QR Codes, in retail.

index.html
TallTed, line 22:
```suggestion
        subtitle: "Expressing information with verifiable provenance",
```

msporny: +1 good change, five stars, gets the job done, would buy again.

msporny (approved):

TallTed:
I think provenance is understood well enough.  It's more than source/origin, and I think that's important to communicate, which would make the subtitle much longer with any other wording.  We *could* make it link to a dictionary definition, the [PROV DBpedia entity](https://dbpedia.org/resource/PROV_(Provenance)), the [PROV Wikipedia page](https://en.wikipedia.org/wiki/PROV_(Provenance)), or even the [PROV-O Primer](https://www.w3.org/TR/2013/NOTE-prov-primer-20130430/).

dlongley (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2021-08-11-vcwg#section3-6) on 2021-08-11

- no resolutions were taken

#### 3.6. Clarified subtitle of Data Model (pr vc-data-model#780)

_See github pull request [#780](https://github.com/w3c/vc-data-model/pull/780)._




**Ted Thibodeau Jr.:** fairly self-explanatory, good discussion.  

> *Manu Sporny:* I'm fine w/ it being in there...

**Wayne Chang:** once the 14-day period is up we can merge  

</details>

dhh1128:
I don't like using the word "provenance" in the subtitle; that has a specific meaning about deep sourcing that I don't think VCs fulfill. We're not describing where data comes from, or proving where it comes from, except in the most proximate sense -- it comes from the issuer. But where the issuer got it (what "provenance" suggests) is out of scope. I suggest we reword to "with verifiable source."

Tagging @SmithSamuelM, who may also have an opinion.

TallTed:
I hear you, @dhh1128. Upon further consideration, I think you're right; "verifiable source" is better.  Now editing the PR....

index.html
```diff
@@ -19,7 +19,7 @@
         shortName: "vc-data-model",
 
         // subtitle for the spec
-        subtitle: "Expressing verifiable information on the Web",
+        subtitle: "Expressing information with verifiable provenance",
```
dlongley, line 22: Suggesting this on behalf of @dhh1128. Can this achieve consensus?

```suggestion
        subtitle: "Expressing information with verifiable source",
```

dlongley, line 22: Another option:

```suggestion
        subtitle: "Expressing information with verifiable authorship",
```

msporny: Ooof, this evokes a "What does that mean?" response from me... I don't see how it is any better than provenance.

msporny: This one feels better... except that we don't talk about "authors" in the document anywhere.

TallTed:
I think I'm OK with either "verifiable source" or "verifiable authorship".  I wish there were an easy way to run a quick poll on Github!

index.html
mwherman2000: It's more than source and/or authorship ...there is an an aspect that deals with the content of the claims as well.

A VC is expressing an trustable attestation that a set of claims has not been modified relative to a point in time (issuance of the VC).

Perhaps, "Expressing a set of immutable attributes relative to a point in time".

dhh1128:
I'm aligned with Ted now.

If Manu doesn't like "source" and only tolerates "authorship", and if Michael doesn't like authoriship, here are a few other options just to stimulate thinking (but to be clear, I'm still good with either of Ted's alternatives):

* Expressing verifiably issued information
* Issuing verifiable information
* Expressing information with a verifiable issuer

Etc.

mwherman2000:
+1 for
> Expressing information with a verifiable issuer

but I believe the title of the specification needs to be changed given that it includes so many disconnected concepts and lacks any serious cohesion.  See https://github.com/w3c/vc-data-model/issues/791

msporny:
In an attempt at compromise wrt. "source"... what about:

> Confirming the source of information on the Web

I'm back-tracking on "on the Web" because we really are using Web technologies here... and the Web does flow into our daily lives, even in the physical space, so it's probably fine.

I'm suggesting "Confirming" as a synonym for "verifying"... it's typically a good idea to explain the title of the specification using different words in the subtitle... I'd be fine w/ going back to "Verifying", but suggest we find a good synonym.

... and this includes the word "source" in a way that I think is understandable to the general public (which is one good measure of a good subtitle).

So, @dhh1128, @TallTed, what about?

> Confirming the source of information on the Web

mwherman2000:
There should be no requirement/limitation to use VCs "on the Web".
VCs, as we're learning (e.g. Unbound Credentials), are starting to become viewed as a more general form of verifiable data envelope.  IMO, "on the Web" is old/archaic and unnecessary.

TallTed:
I do not believe that including or excluding "on the Web" in/from the title puts any requirement or limitation on where or how VCs may be used, any more than that phrase has done so in the title of any other W3 TR.

"Confirming/Verifying the source of information [on the Web]" suggests that the spec is applicable to just any information you may encounter [on the Web], but in fact, it only applies to (*some*) information that has been expressed with the specific intent of enabling such confirmation/verification.

"Expressing information with verifiable authorship [on the Web]" works because it puts the necessary action ("Expressing") first.

In other words, Verifiability depends *first* on the Issuer doing the right things.

(Related concerns, such as making false statements [e.g., "fake news" including but not limited to satirical "news" reports] more easily detectable, which I think many thought would be solved by the Verifiable Claims WG, Verifiable Credentials Data Model, Credentials CG, et al, are the focus of the [Credibility CG](https://www.w3.org/community/credibility/), which has been rather moribund in recent years due to a [perceived, correctly or not] lack of achievable goals or implementable projects.)

mwherman2000:
Here's something to think about in terms of the future vision for VCs as well as the current scope of the 1.0 spec: https://youtu.be/IAozsIsrlbU?t=1947
Watch this for 2-3 minutes if you have the time.

kdenhartog:
It appears we may not have consensus on the PR as currently authored at this point. Leaving this one for now

brentzundel:
perchance: "Enabling confirmation of information sources [on the web]"

mwherman2000:
Continuing with @TallTed's feedback...

- Longer version: Verifiable credentials are verifiable containers (or envelopes) for arbitrary digital data.
- Shorter version: Verifiable credentials are verifiable containers for arbitrary digital data.
- Shortest version: Verifiable containers for arbitrary digital data

mwherman2000:
> Any argument with "Structuring verifiable containers for data"?

The addition of _Structuring_ makes it sound like "how to" document rather that a "what" document (IMO not a connotation we're looking for.)

kdenhartog:
I get the sense many of these suggestions which build on "verifiable containers of data" leave it so arbitrary that it leaves more up to the interpretation of the reader and could lead them astray as to the purpose and common ways VCs are used. Some points of consideration in this:

Should we further constrain around the concept of a subject given it's a core concept in the spec? 
 - e.g. "verifiable containers of claims about a subject"
 - e.g. "verifiable claims of data about an entity"
 
Are we expecting to highlight the nature of the structure in any way? Given this is a data model which does constrain things to a greater degree than say just a JWT or a JSON-LD Proof is it worth us calling this out?
 - e.g. "a structured data model for making data verifiable"
 - e.g. "an extensible data model for making verifiable claims"
 
 If I might offer one that includes these two concepts:
 
 "a structured data model for making verifiable claims of data about an entity"

mwherman2000:
@kdenhartog See below. Note in the previous posts for VCs being a _container_ (in particular, a _verifiable container_) for digital data

> e.g. "verifiable containers of claims about a subject"

No - not all VCs have a subject, are metadata about a subject, nor do all VCs need a subject (reference: Unbound Credentials - see above).  We need to ignore, for now, the JSON-LD specific concept of null subjects ...this is JSON-LD #tinsel that none (most?) of the other data platforms want or need.

> e.g. "verifiable claims of data about an entity"

No - misses out on the concept of VCs being a verifiable _container_ (or _envelope_) for digital data

> e.g. "a structured data model for making data verifiable"

No - misses out on the concept of VCs being a verifiable _container_ (or _envelope_) for digital data

> e.g. "an extensible data model for making verifiable claims"

No - misses out on the concept of VCs being a verifiable _container_ (or _envelope_) for digital data. It's not about making the _claims_ verifiable. It's about _container_ for the claims that is verifiable.

> "a structured data model for making verifiable claims of data about an entity"

No^4 ...See above

kdenhartog:
>Check out the series of previous posts that document the growing consensus for VCs being some sort of a container (in particular, a verifiable container) for digital data. Also check out my comments below.

Sure, while there is a growing consensus for that, that doesn't read like a statement that is aligned with the spec as it stands today. In my personal opinion, I think it would be better for us to not link this spec to these new concepts until we've had sufficient time to discuss it as a WG since it would also require us to update quite a few statements in the spec (some normative and some non-normative) in order to align this.

As one example, we currently state in [section 4.4](https://w3c.github.io/vc-data-model/#credential-subject):  

```
A verifiable credential contains claims about one or more subjects. This specification defines a credentialSubject property for the expression of claims about one or more subjects.

A verifiable credential MUST have a credentialSubject property. 
```

If we're going to further abstract VCs to no longer include a subject, then we shouldn't be requiring the usage of the `credentialsSubject` property. However removing this property would constitute a breaking change on V1. So it's my opinion that subtitles that try to introduce the VC spec as an arbitrary "verifiable container" don't reflect the current state of the spec. For this reason I believe that we shouldn't be adding changes like this until we can properly discuss this as a WG.

mwherman2000:
> > growing consensus for VCs being some sort of a container (in particular, a verifiable container) for digital data

> Sure, while there is a growing consensus for that, that doesn't read like a statement that is aligned with the spec as it stands today. In my personal opinion, I think it would be better for us to not link this spec to these new concepts until we've had sufficient time to discuss it as a WG since it would also require us to update quite a few statements in the spec (some normative and some non-normative) in order to align this.

@kdenhartog The fundamental issue underlying this whole discussion (which IMO should be a relatively simple discussion/decision) is because, after 2 years, there has been, until recently, no consensus on the subtitle/byline; and hence, a strong indication there is no consensus on what an actual VC represents.

The fact that there is growing consensus around "VCs being some sort of a container (in particular, a verifiable container) for digital data" is:
1. A good thing because there was no consensus before
2. I believe the recent thinking represents a natural maturing of the conversations and concepts around VCs ... i.e. what they actually are.
3. Other than yourself @kdenhartog , I haven't seen or heard any dissent around the idea of "VCs being some sort of a container (in particular, a verifiable container) for digital data".

These are all good important developments ...IMO fundamental developments.

I don't see how the quote from the recommendation has any bearing on this. 
> A verifiable credential contains claims about one or more subjects. This specification defines a credentialSubject property for the expression of claims about one or more subjects.
> A verifiable credential MUST have a credentialSubject property.

No one, to my knowledge, has suggested that the _credentialSubject_ be removed, made optional, renamed,  or otherwise changed.  Its just _there_ - perhaps its naming is nothing more than a historical artifact.

Further, the _credentialSubject_ is _not_ the determinant of whether a credential _subject_ or not. Again,  _credentialSubject_ is just there to group a collection of properties together (technically, an array of collections of properties). _credentialSubject_

Whether or not, a credential has a _subject_ or not is determined by whether  _credentialSubject_ contains the optional _id_ property.  I believe @msporny also agrees with this point (reference: https://github.com/w3c/vc-data-model/pull/800)

Here's examples of 2 VCs to illustrate these points...

image

Gently, based on the above discussion points, I don't believe the following conclusions are neither justified nor correct:

>If we're going to further abstract VCs to no longer include a subject, then we shouldn't be requiring the usage of the credentialsSubject property. However removing this property would constitute a breaking change on V1. So it's my opinion that subtitles that try to introduce the VC spec as an arbitrary "verifiable container" don't reflect the current state of the spec. For this reason I believe that we shouldn't be adding changes like this until we can properly discuss this as a WG.

TallTed:
@kdenhartog

> * e.g. "verifiable containers of claims about a subject"

This is a partial possibility, because while it accurately expresses that it's the *container* which is verifiable, it limits the VC to a single subject.  If it were "about _one or more subjects_" or "about _one or more entities_", that could work.

That said, the *credential* is only verifiable in that (1) its _issuer_ is confirmed via public/private keypair or similar cryptographic material, and (2) its contents (the _claims_) have not been altered, i.e., they are as issued.

> * e.g. "verifiable claims of data about an entity"
> * e.g. "a structured data model for making data verifiable"
> * e.g. "an extensible data model for making verifiable claims"
> "a structured data model for making verifiable claims of data about an entity"

These suggestions all fail, because the *claims* are not *verifiable* in any way aside from being unchanged since the VC was issued; their veracity is unknown.

kdenhartog:
@mwherman2000 

I may have misunderstood that you were alluding to removing the subject. This has been a concept that I've heard floated around in different discussions where the same term "verifiable container" was used as well. Through discussions and readings about the authentic data containers which also rely upon the similar concepts and as I understood that body of work, they seem to support not having a subject (@brentzundel / @dhh1128 please correct me if I'm wrong). That in combination of your discussions around "unbound credentials" led me to believe that was what was being discussed.

In this case, I've conflated your meaning of "container" and theirs to assume a broader scope. Thanks for clarifying that you're not looking to imply further removal of the subject in this case. Now that I have further clarity around the way you're using it let me see if I can find a synonym that aligns my thinking and yours given "container" may be a bit overloaded here. I'm definitely not as opposed to it now compared to when I originally authored that comment with this additional clarity.

@TallTed said:
> This is a partial possibility, because while it accurately expresses that it's the container which is verifiable, it limits the VC to a single subject. If it were "about one or more subjects" or "about one or more entities", that could work.

Yup, I agree with that assessment.

> These suggestions all fail, because the claims are not verifiable in any way aside from being unchanged since the VC was issued; their veracity is unknown.

I agree with that assessment, although it does beg the question - Why call them "verifiable" credentials then?

Based on your comments what about this?

`a structured data model for making authenticatable claims of data about one or more entities`

SmithSamuelM:
A problem with the word "verifiable" is that it has a common english dictionary definition and a narrow cryptographic definition which are not the same. What may be confusing to any non-technical users of verifiable credentials is that they will tend to use the dictionary definition not the cryptographic definition which conveys the wrong meaning. 

The dictionary definition may be more precisely understood as verifiable veracity of the claims.

Whereas as the cryptographic definition may be more precisely understood as verifiable signatures or other cryptographic proofs of the authenticity of the claims vis-a-vis the issuer.

A better word would be "authenticatable credential" which precisely unambiguously refers to the cryptographic definition and avoids the dictionary definition. 

But alas the word VC may be stuck with us in spite of its ambiguous meaning if for no other reason than inertia.

But since the community already switched from verifiable claims to verifiable credentials it may be worth revisiting the use of the terms verifiable and replace it with authenticatable or authentic.

dlongley:
@kdenhartog,

>> These suggestions all fail, because the claims are not verifiable in any way aside from being unchanged since the VC was issued; their veracity is unknown.
>
> I agree with that assessment, although it does beg the question - Why call them "verifiable" credentials then?

Actually, this is one reason why the WG rejected the terminology "verifiable claims" and instead went with "verifiable credentials". This was meant to help clarify that the claims themselves are not verifiable; the credentials are. Credentials are presentable *containers* of claims issued by third parties. So, what you can verify is the authorship of the credentials -- you can know (with a very high degree of certainty) where the credentials came from and that they have not been tampered with since their creation.

@SamSmith,

> A better word would be "authenticatable credential" which precisely unambiguously refers to the cryptographic definition and avoids the dictionary definition.

I'm afraid that calling something authentic or authenticatable may be not much more likely to evoke the important differentiation you highlight. "Authentic" often just means genuine, true, or accurate -- and many people could just as easily think that authenticating the credentials means they (or the claims therein) are "true". Getting one's credentials "verified" may do a better job of conveying the credentials were simply checked to ensure they come from the issuing authority. But, really, I suspect many people just think of both of those words as synonyms and wouldn't pick up on the nuance.

I do think it's good that we didn't end up using the term "verifiable claims" and instead chose "verifiable credentials" because it does seem to have been helpful thus far.

mwherman2000:
> In this case, I've conflated your meaning of "container" and theirs to assume a broader scope. Thanks for clarifying that you're not looking to imply further removal of the subject in this case. Now that I have further clarity around the way you're using it let me see if I can find a synonym that aligns my thinking and yours given "container" may be a bit overloaded here. I'm definitely not as opposed to it now compared to when I originally authored that comment with this additional clarity.

@kdenhartog Because someone may have abused the word _container_ in the past, IMO it's not a reason to reject the word when it is properly the best term.  IMO, to look for a synonym is a kin to looking for a _second best_ word.  We shouldn't have a "second best" subtitle/byline IMO.

mwherman2000:
> A Credential always has a Subject, whether that Subject is named (i.e., explicitly identified by IRI, by proper noun) or not (i.e., identified implicitly by blank node, by pronoun).

Do you have a specific reference or authority for this claim (;-)) @TallTed ...serious question.

In all of my reading and writing, a Credential is simply a collection of Claims (which in turn are defined to be a collection of Name-Value Pairs).  There is no definition or association with a Subject. However, as clearly seen in the VC Data Model, a Subject can be represented as one of the claims (name-value pairs) in a credential.

image

mwherman2000:
@TallTed Forgive my directness but... references to RDF in these discussions are a bit of a red herring wrt the VC data model.  _Resource Description Framework (RDF)_ is not part of either the normative nor non-normative text in the VC data model recommendation,

mwherman2000:
> a structured container for making provenanced claims of data about one or more entities

@kdenhartog If you think of the imagined intended audience (which are the only people that matter in this discussion), IMO nobody is going to understand this. In fact, IMO, I believe will detract from people wanting to read the VC data model specification.

I believe the subtitle/byline needs to:

> The subtitle reels them in by explaining exactly why they should read the book. (https://scribemedia.com/book-subtitle/)

> A good subtitle provides context for the title. It's informative, letting you know who should read the book and how it will help them. Your subtitle should also tell readers who the book is for, but that doesn't mean you have to name the audience. (https://scribemedia.com/book-subtitle/)

SmithSamuelM:
@dlongley I agree that authentic has a broader meaning that might still overlap a little bit with veracity but certainly less than verifiable does. My first suggestion is to use autheticatable  defined in the dictionary as : Capable of being authenticated.
Which is clear and precise and is more precise and more concise but with the same meaning as 

> So, what you can verify is the authorship of the credentials

TallTed:
@SmithSamuelM 
> @dlongley I agree that authentic has a broader meaning that might still overlap a little bit with veracity but certainly less than verifiable does. My first suggestion is to use **_authenticatable_** defined in the dictionary as : Capable of being authenticated.
> Which is clear and precise and is more precise and more concise but with the same meaning as
> 
> > So, what you can verify is the authorship of the credentials

The clarity and precision you see in _authenticatable_ is not necessarily seen by others. I see it as more ambiguous, in fact, because it leads you straight to the verb _authenticate_, which Google tells me is defined as --

1. prove or show (something, especially a claim or an artistic work) to be true or genuine. "they were invited to authenticate artifacts from the Italian Renaissance"
2. validate.
3. **`COMPUTING`** (of a user or process) have one's identity verified. "it only accepts commands from users who have authenticated as 'clients'"

-- so if people's eyes are drawn to the `COMPUTING`-focused definition we're right back to verification, but oddly limited to identity!  The first definition here is really the most apt, but it prompts a dive in to the claims (which we're *not* labeling as verifiable) and suggests ignoring the container, the _credential_, which holds them!

SmithSamuelM:
@dlongley > so if people's eyes are drawn to the COMPUTING-focused definition we're right back to verification, 

The verification of WHAT is the problem. Both the first and third definitions provide the  WHAT. Verification by itself is ambiguous because it doesn't define WHAT is being verified. Authenticatible implies strongly the WHAT that is is being verified is authorship, identity of the source of information or commands. You have verifiable credential which expands to verifiable container of claims (by your definition). But still leaves open what is being verified. Is is the veracity of the claims in the container or the authorship (to use your word) of the claims in the container. Clearly authenticatible points to authorship much more so than it points to veracity. Autheticatible may not be perfect but its much more precise than verifiable.  But anyway take it for it its worth. As I said above I doubted that the inertia of VC was surmountable in any event. So maybe its a moot point.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2021-09-08-vcwg#section3-9) on 2021-09-08

- no resolutions were taken

#### 3.9. Clarified subtitle of Data Model (pr vc-data-model#780)

_See github pull request [#780](https://github.com/w3c/vc-data-model/pull/780)._




**Brent Zundel:** This is, should we change the subtext of the data model.  
… Does anyone have a subtitle they would like to throw out now, that would end up in a comment in the PR?  
… If you aren't familiar with this conversation, feel free to jump in and make comments in the PR.  

**David Chadwick:** Let's move on...  

**Brent Zundel:** Okay, next.  

</details>

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2021-09-15-vcwg#section3-6) on 2021-09-15

- no resolutions were taken

#### 3.6. Clarified subtitle of Data Model (pr vc-data-model#780)

_See github pull request [#780](https://github.com/w3c/vc-data-model/pull/780)._




> *Brent Zundel:* [https://github.com/w3c/vc-data-model/pull/780](https://github.com/w3c/vc-data-model/pull/780)

**Brent Zundel:** One more... in case anyone has had this brilliant realization: what should the subtitle of our data model be?  
… There has been quite a conversation, refining it...  
… The summary at the end was... let's just keep looking.  
… If no one has anything more to suggest, we'll move on to issues.  

</details>

kdenhartog:
This one is still pending - any new ideas to resolve this yet? In absence of agreement I think we're supposed to stay with what we have which I think isn't quite as good as many of the suggestions being provided here, so let's try to find something that works for everyone and get it updated.

TallTed:
"Expressing data in verifiably authored containers"

dlongley:
"Expressing verifiably authored data" seems shorter/better.

TallTed:
"Expressing verifiably authored data" suggests that the statements within the VC are verifiably authored (issued), not that the collection is...

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2021-11-10-vcwg#section3-1) on 2021-11-10

- no resolutions were taken

#### 3.1. Clarified subtitle of Data Model (pr vc-data-model#780)

_See github pull request [vc-data-model#780](https://github.com/w3c/vc-data-model/pull/780)._




**Manu Sporny:** I prefer we close this PR and try again later.  

> *Dave Longley:* "Expressing verifiably authored data" is shorter.

**Ted Thibodeau Jr.:** it has been a lot of bike shedding because it is nuanced and getting this right is difficult.  

> *Manu Sporny:* Not a strong suggestion to close it.

> *Manu Sporny:* I feel like we need to transform it into an issue and go from there?.

> *Ted Thibodeau Jr.:* "Expressing verifiably authored data" suggests that the statements within the VC are verifiably authored, not that the collection is....

> *Manu Sporny:* I'm fine w/ moving on as well.

</details>

kdenhartog:
This PR is intended to address issue #105

While there has been a fair bit of discussion on this PR itself about the topic, our go to solution is to label the issue rather than the PR, so I'll be removing the Errata and Editorial labels from this PR and making sure the issue is linked. Given that issue has been closed already and this PR is open I believe the intent is for discussion to continue here until we're able to merge it. 

Any issues with this @brentzundel @wyc or @msporny ?

brentzundel:
I agree that this PR does not need an errata label.

TallTed:
"Expressing data in verifiably assembled containers"?

kdenhartog:
"Establishing provenance of data on the web using standardized containers"?

dlongley:
Some other ideas:

1. Expressing data in verifiably authentic containers
1. Expressing data in verifiably authentic documents
1. Expressing data in verifiably authentic digital documents
1. Expressing claims in verifiably authentic containers
1. Expressing claims in verifiably authentic documents
1. Expressing claims in verifiably authentic digital documents

kdenhartog:
> Expressing data in verifiably authentic digital documents

This one seems the most explain like I'm 5 of them all - I'd be good with this one

msporny:
> > Expressing data in verifiably authentic digital documents
> 
> This one seems the most "explain like I'm 5" of them all - I'd be good with this one

+1

dlongley:
> Expressing data in verifiably authentic digital documents

My remaining concern with this subtitle is that it sounds more generic than we intended with the VC data model. A VC is a *credential*, not just any arbitrary digital document. VCs are about expressing the attributes of entities / describing subjects / making claims about subjects. This is more narrow than "expressing data". This is an important distinction that I don't think we want to lose -- and it seems to me that failing to highlight it in the past has caused confusion.

David-Chadwick:
4. from @dlongley's  list is my favourite, followed by 5. I agree with @dlongley that "data" is too generic

msporny:
We need to either close this issue, or rank choice vote these options and put it to the CCG... we're going around in circles.

David-Chadwick:
Giving a short list of options to the CCG for voting on, is one way to resolve this, providing we can agree on the short list of say half a dozen options.

dlongley:
Another option: "Verifiably authentic containers for claims"

TallTed:
> > Expressing data in verifiably authentic digital documents
> 
> [@dlongley] My remaining concern with this subtitle is that it sounds more generic than we intended with the VC data model.

Perhaps more generic than *you* intended, but I absolutely meant it to be more generic than you appear to be describing here.

Further, to my eyes, "claims" doesn't have a clear link to the meaning I think most here would be intending, and "expressing claims" is just as vague.

Once again, the reason we changed from "claims" to "credentials" was that the former suggested to people that the statements (which we intended to be the "claims" under discussion) contained in the were verifiable, not that the utterance (the container, the "credential", the collection of "claims") was verifiably sourced.

-----

As @msporny says, we are going around in the same circles, again.

We might be better off with *no* subtitle, than with continuing to cycle through many if not all of the suggestions made to date and likely to follow.

dlongley:
> We might be better off with no subtitle, than with continuing to cycle through many if not all of the suggestions made to date and likely to follow.

+1 to that. I actually think it's the best proposal -- remove the subtitle. We're trying to communicate too much with too few words -- let people read the introduction / whatever they need to.

index.html
msporny, line 22:
```suggestion
        //subtitle: "Expressing information with verifiable provenance",
```

This change suggestion removes the subtitle of the specification completely, given that we've been at this PR for about six months now without a resolution.

So, the proposal is -- remove the subtitle! Any objections?

kdenhartog: None from me

brentzundel (approved):

kdenhartog (approved):

mwherman2000:
It's all about the outer part of the credential being verifiable and only this part (the credential envelope). The VC "concept" says nothing about the verifiability the inner part of the credential ...i.e. the parts inside the credentialSubject (aka the data part) ...other than it (the envelope and the blob that is inside) hasn't been tampered with once signed. 

Watch https://youtu.be/kM30pd3w8qE and then  https://youtu.be/9RLYS7Xvabc

mwherman2000:
I can approve going with no subtitle

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2021-12-15-vcwg#section2-1) on 2021-12-15

- no resolutions were taken

#### 2.1. Clarified subtitle of Data Model (pr vc-data-model#780)

_See github pull request [vc-data-model#780](https://github.com/w3c/vc-data-model/pull/780)._




**Manu Sporny:** We've got one we should be able to resolve very quickly.  
… we've tried so many things and we've landed on just deleting the subtitle.  
… the suggestion is to just remove it and that seems to have the best consensus.  
… no objections to this solution.  
… so this is a final call for objections.  
… if not I'll merge this on the call.  

**Brent Zundel:** Any objections?.  
… none heard.  

**Manu Sporny:** That was 6 months in the making :).  
… there's multiple that are just editorial that we can just skip.  

</details>

msporny:

PR #783, opened 2021-07-29
adds properties for AlumniCredential and DisputeCredential for valid JSON-LD examples

### LINKED ISSUE #750: Some example credential types not defined in example context
The spec uses `AlumniCredential` as an example credential type in § 4.1 Contexts, and `DisputeCredential` in § 5.9 Disputes. But the example context `https://www.w3.org/2018/credentials/examples/v1` does not define those types. It does define `UniversityDegreeCredential` and `RelationshipCredential`.

There are also additional credential types used in examples not defined in the example context, but those are in non-normative sections so I think that is supposed to be okay.

§ 4.3 Types says, "The value of the type property MUST be, or map to (through interpretation of the `@context` property), one or more URIs." Also, [JSON-LD 1.1 § 3.5 Specifying the Type](https://www.w3.org/TR/json-ld/#specifying-the-type) (non-normative) says, "In Linked Data, types are uniquely identified with an IRI." I think this is needed for LD-Proofs, where the type is expanded to a IRI for converting to normalized N-Quads for hashing.

### LINKED ISSUE #769: AlumniCredential missing from example VC context
The AlumniCredential term is missing in the Verifiable Credentials Example context: https://www.w3.org/2018/credentials/examples/v1

... which causes the value to be interpreted as a relative URL in Example 2, which is wrong:

https://www.w3.org/TR/vc-data-model/#example-2-a-simple-example-of-a-verifiable-presentation

Thanks to @iherman for catching the bug.

kdenhartog:
This is an addition to the example context file which is consider non-normative so that the examples in example 2 and example 26 produce valid JSON-LD.

This PR closes #769 and closes #750

### DIFF
--- a/contexts/credentials/examples/v1
+++ b/contexts/credentials/examples/v1
@@ -18,6 +18,8 @@
     "Mother": "ex:Mother",
     "RelationshipCredential": "ex:RelationshipCredential",
     "UniversityDegreeCredential": "ex:UniversityDegreeCredential",
+    "AlumniCredential": "ex:AlumniCredential",
+    "DisputeCredential": "ex:DisputeCredential",
     "ZkpExampleSchema2018": "ex:ZkpExampleSchema2018",
 
     "issuerData": "ex:issuerData",
@@ -42,6 +44,8 @@
     "evidenceDocument": "ex:evidenceDocument",
     "spouse": "schema:spouse",
     "subjectPresence": "ex:subjectPresence",
-    "verifier": {"@id": "ex:verifier", "@type": "@id"}
+    "verifier": {"@id": "ex:verifier", "@type": "@id"},
+    "currentStatus": "ex:currentStatus",
+    "statusReason": "ex:statusReason",
   }]
 }
--- a/contexts/credentials/examples/v1
+++ b/contexts/credentials/examples/v1
@@ -46,6 +46,6 @@
     "subjectPresence": "ex:subjectPresence",
     "verifier": {"@id": "ex:verifier", "@type": "@id"},
     "currentStatus": "ex:currentStatus",
-    "statusReason": "ex:statusReason",
+    "statusReason": "ex:statusReason"
   }]
 }


kdenhartog:
As I understand the discussion between @msporny and @iherman in #769 this doesn't require an errata label and can be updated to main directly. However for better awareness about the change, I think it would be useful to actually merge this against V1.1 instead and still mark it as an errata just for proper noting about the change even though it.

TallTed:
It seems odd to combine the big process changes in the README with the little `@context` changes in one PR.

kdenhartog:
Thanks for the review. I originally built this off of main instead of V1.1 and then realized these needed to target V1.1 branch in which case that process review PR got added back in. I'll pull that portion out and upgrade V1.1 to include this new process in a separate PR so that these changes are far easier to get reviewed. Thanks for pointing this out.

brentzundel (approved):

contexts/credentials/examples/v1
```diff
@@ -42,6 +44,8 @@
     "evidenceDocument": "ex:evidenceDocument",
     "spouse": "schema:spouse",
     "subjectPresence": "ex:subjectPresence",
-    "verifier": {"@id": "ex:verifier", "@type": "@id"}
+    "verifier": {"@id": "ex:verifier", "@type": "@id"},
+    "currentStatus": "ex:currentStatus",
+    "statusReason": "ex:statusReason",
```
dlongley, line 49: No trailing commas in JSON.

```suggestion
    "statusReason": "ex:statusReason"
```

dlongley (approved):
Approving but not that a minor tweak is needed (see comment).

msporny (approved):

contexts/credentials/examples/v1
TallTed: This comma must be removed before merge!

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2021-08-11-vcwg#section3-5) on 2021-08-11

- no resolutions were taken

#### 3.5. adds properties for AlumniCredential and DisputeCredential for valid JSON-LD examples (pr vc-data-model#783)

_See github pull request [#783](https://github.com/w3c/vc-data-model/pull/783)._




**Wayne Chang:** multiple approvals here  

**Kyle Den Hartog:** small updates to examples so the example contexts are proper  
… so they work as expected  

**Manu Sporny:** this is a change to the example context, which nobody on the planet should be relying on for production. So updating it is not a problem  

**David Chadwick:** this means you could put nonTransferrable into the example context?  

**Manu Sporny:** yes, if we had consensus to do that, but I don't think we do  
… if we agree as a working group that we want others to follow that model, then we could do that, but it would only exist in the example  

> *David Chadwick:* its already in an example, so that is why it should be in the examples context

</details>

kdenhartog:
multiple positive reviews, feedback received and incorporated, merging

kdenhartog:

Issue #776, opened 2021-04-27
Example 22 uses termsOfUse


pchampin:
[Example 22](https://www.w3.org/TR/vc-data-model/#example-22-usage-of-the-termsofuse-property-by-a-holder) uses `termsOfUse` as a property of a veriable presentation.

However, in [the context](https://www.w3.org/2018/credentials/v1), `termsOfUse` is only defined in the local context of `VerifiableCredential`.

msporny:
> [Example 22](https://www.w3.org/TR/vc-data-model/#example-22-usage-of-the-termsofuse-property-by-a-holder) uses `termsOfUse` as a property of a veriable presentation.
> 
> However, in [the context](https://www.w3.org/2018/credentials/v1), `termsOfUse` is only defined in the local context of `VerifiableCredential`.

Yes, this is a bug in the JSON-LD context, nice catch @pchampin!

Noting what the spec [says for ToU](https://www.w3.org/TR/vc-data-model/#terms-of-use):

"Terms of use can be utilized by an issuer or a holder to communicate the terms under which a verifiable credential or verifiable presentation was issued."

We always meant `termsOfUse` to be usable from within a presentation... we just didn't carry that through to the JSON-LD Context.

> cc @msporny for any thoughts on the substantiveness of modifying the VP context defn

Imma spec language lawyer here for a second:

1. We always meant `termsOfUse` to be usable from within a presentation.
2. The JSON-LD Context in the specification is non-normative.

So, technically speaking, we could update the JSON-LD Context in place and fix the bug with little chance that things could go wrong. 

*takes spec lawyer hat off*

That said, I don't think we should change the 2018 VC context and fix it... people are using that context in production. We should treat this like a substantive change and put it in the next JSON-LD Context for VCs. I suggest that even though it's not technically a substantive change to the 2018 context, it's still effectively a substantive change since it would affect implementation behavior (the term would go from being dropped to not being dropped during signing).

Instead, we should fix this in the next version of the VC context... noting that we're going to be making a bunch of other changes at the same time (like factoring the cryptosuites out).

kdenhartog:
I've written a PR to remove this from the specification per @wyc option 1. We'll need to either update this issue afterwards and leave it open to track re-adding the text back in or we'll need to open another issue after it's been merged.

TallTed:
[...]

PR #786, opened 2021-08-03
Recode graph diagrams

chaals:
See #777 - I diverged from that issue and cleaned up the code and thought about the accessibility. (I'll write a different PR that should address #777 directly)

- make text into actual text
- use stylesheet
- simplify source for legibility
- Order similar to the flow, for accessibility

- Make credential-graph.svg a subsection, and note with a comment.
- Add slight background colouring to further delineate each subsection

PR #787, opened 2021-08-04
removes mention of termsOfUse with verifiable presentations

kdenhartog:
Since the verifiable presentation term context did not include the property the solution we've ended up on is to remove this portion
of the text from the current specification and it will be readded back in in Version 2 when we can update the contexts.


kdenhartog:
fixes issue #776

brentzundel:
I wonder if this PR will make the `termsOfUse` section more confusing by removing the presentation example (editorial change, no problem), but doesn't remove the text that explains that presentations can have a `termsOfUse` property (substantive change, maybe v1.2, but probably v2).
I think that rather than removing the example, it may be better to add a note that says there is a bug in the `@context` file such that it doesn't support `termsOfUse` in presentations even though the spec does, so implementers who wish to use this property will need to add `termsOfUse` to their own context for presentations.

kdenhartog:
> I wonder if this PR will make the `termsOfUse` section more confusing by removing the presentation example (editorial change, no problem), but doesn't remove the text that explains that presentations can have a `termsOfUse` property (substantive change, maybe v1.2, but probably v2).
> I think that rather than removing the example, it may be better to add a note that says there is a bug in the `@context` file such that it doesn't support `termsOfUse` in presentations even though the spec does, so implementers who wish to use this property will need to add `termsOfUse` to their own context for presentations.

Yeah that's something I was thinking about as well when I was working on this. You stating that makes me think it's actually probably the better direction. I'll get a second PR open on that today so that we can decide which is preferred method during the review period.

kdenhartog:
I did some investigation into how this would work before creating a PR. As far as I'm able to tell extending the contexts to support the term in the Verifiable Presentation wouldn't work as defined. Since the `termsOfUse` property is already defined within the V1 context it appears to create a term conflict because the `termsOfUse` property is protected in the VC document. If someone can link me to an example that shows a valid JSON-LD document extending to make this property available, I'll go ahead and write up the second PR to support that.

At this point though, I'm not convinced the extension method is a valid path forward.

David-Chadwick:
> I think that rather than removing the example, it may be better to add a note that says there is a bug in the `@context` file such that it doesn't support `termsOfUse` in presentations even though the spec does, so implementers who wish to use this property will need to add `termsOfUse` to their own context for presentations.

This is my favoured approach. The intention was always that ToU should be in presentations. So we should point out there is a bug in the "at context" rather than remove the example and any mention of ToU in presentations

kdenhartog:
> > I think that rather than removing the example, it may be better to add a note that says there is a bug in the `@context` file such that it doesn't support `termsOfUse` in presentations even though the spec does, so implementers who wish to use this property will need to add `termsOfUse` to their own context for presentations.
> 
> This is my favoured approach. The intention was always that ToU should be in presentations. So we should point out there is a bug in the "at context" rather than remove the example and any mention of ToU in presentations

I agree, but I couldn't find a way that "implementers who wish to use this property will need to add `termsOfUse` to their own context for presentations" could actually go about doing this because terms are being protected. If someone can show me a document with it working then I can switch that over. I just didn't want to suggest that route if it's not actually usable.

David-Chadwick:
One solution might be to define PToU (Presentation Terms Of Use) in VCv2 as a differently named (or alias) property for VPs with the same semantics as the existing ToU for VCs.
On a related note, we already have to supplement the existing ToU context definition for VCs, when we add our our properties to ToUs. So clearly the existing ToU is not immutable.

kdenhartog:
> 
> 
> One solution might be to define PToU (Presentation Terms Of Use) in VCv2 as a differently named (or alias) property for VPs with the same semantics as the existing ToU for VCs.
> On a related note, we already have to supplement the existing ToU context definition for VCs, when we add our our properties to ToUs. So clearly the existing ToU is not immutable.

Do you have an example document of this that I can play around with to see it working?

David-Chadwick:
This is the URL that was used in the TRAIN project to define the context for the trust scheme that is embedded in the ToU
Currently we have not used ToUs in VPs, only in VCs.

dlongley:
@kdenhartog,

> As far as I'm able to tell extending the contexts to support the term in the Verifiable Presentation wouldn't work as defined. Since the `termsOfUse` property is already defined within the V1 context it appears to create a term conflict because the `termsOfUse` property is protected in the VC document.

The fact that `termsOfUse` is defined in a type-scoped context for `VerifiableCredential` shouldn't be an issue here. The issue, I'd think, is that `VerifiablePresentation` does not define `termsOfUse` and you can't just redefine `VerifiablePresentation` to include it (and you can't "append" term definitions to an existing type-scoped context).

What you can do (without running afoul of the `@protection` rules) is add a new type like `TermsOfUseVerifiablePresentation`, include a `termsOfUse` term definition in its type-scoped context, and then anyone creating a `VerifiablePresentation` that wants to use the term would need to include the additional type in the VP: `{"type": ["VerifiablePresentation", "TermsOfUseVerifiablePresentation"], ...}`.

kdenhartog:
> @kdenhartog,
> 
> > As far as I'm able to tell extending the contexts to support the term in the Verifiable Presentation wouldn't work as defined. Since the `termsOfUse` property is already defined within the V1 context it appears to create a term conflict because the `termsOfUse` property is protected in the VC document.
> 
> The fact that `termsOfUse` is defined in a type-scoped context for `VerifiableCredential` shouldn't be an issue here. The issue, I'd think, is that `VerifiablePresentation` does not define `termsOfUse` and you can't just redefine `VerifiablePresentation` to include it (and you can't "append" term definitions to an existing type-scoped context).
> 
> What you can do (without running afoul of the `@protection` rules) is add a new type like `TermsOfUseVerifiablePresentation`, include a `termsOfUse` term definition in its type-scoped context, and then anyone creating a `VerifiablePresentation` that wants to use the term would need to include the additional type in the VP: `{"type": ["VerifiablePresentation", "TermsOfUseVerifiablePresentation"], ...}`.

That's what I was thinking as well and I couldn't get it working. I think the reason was actually an issue with how I was looking to extend it. I'll give it another go since your point made it obvious to me now where I was likely making the mistake.

kdenhartog:
Ok I tried this again and still couldn't get it working on JSON-LD Playground, with that suggestion: 

https://json-ld.org/playground/#startTab=tab-expanded&json-ld=%7B%22%40context%22%3A%5B%22https%3A%2F%2Fwww.w3.org%2F2018%2Fcredentials%2Fv1%22%2C%22https%3A%2F%2Fwww.w3.org%2F2018%2Fcredentials%2Fexamples%2Fv1%22%2C%22https%3A%2F%2Fgist.githubusercontent.com%2Fkdenhartog%2Fe3fe193144bfde9901cedde196719cb5%2Fraw%2F9279eb8ebbb0e5b422b71254b961e872f6754dc7%2Fplayground.jsonld%22%5D%2C%22id%22%3A%22did%3Aexample%3Aebfeb1f712ebc6f1c276e12ec21%22%2C%22type%22%3A%5B%22VerifiablePresentation%22%2C%22termsOfUseVpExtension%22%5D%2C%22verifiableCredential%22%3A%5B%7B%22%40context%22%3A%5B%22https%3A%2F%2Fwww.w3.org%2F2018%2Fcredentials%2Fv1%22%2C%22https%3A%2F%2Fwww.w3.org%2F2018%2Fcredentials%2Fexamples%2Fv1%22%5D%2C%22id%22%3A%22http%3A%2F%2Fexample.edu%2Fcredentials%2F3732%22%2C%22type%22%3A%5B%22VerifiableCredential%22%2C%22UniversityDegreeCredential%22%5D%2C%22issuer%22%3A%22https%3A%2F%2Fexample.edu%2Fissuers%2F14%22%2C%22issuanceDate%22%3A%222010-01-01T19%3A23%3A24Z%22%2C%22credentialSubject%22%3A%7B%22id%22%3A%22did%3Aexample%3Aebfeb1f712ebc6f1c276e12ec21%22%2C%22degree%22%3A%7B%22type%22%3A%22BachelorDegree%22%2C%22name%22%3A%22Bachelor%20of%20Science%20and%20Arts%22%7D%7D%2C%22proof%22%3A%7B%7D%7D%5D%2C%22termsOfUse%22%3A%5B%7B%22type%22%3A%22HolderPolicy%22%2C%22id%22%3A%22http%3A%2F%2Fexample.com%2Fpolicies%2Fcredential%2F6%22%2C%22profile%22%3A%22http%3A%2F%2Fexample.com%2Fprofiles%2Fcredential%22%2C%22prohibition%22%3A%5B%7B%22assigner%22%3A%22did%3Aexample%3Aebfeb1f712ebc6f1c276e12ec21%22%2C%22assignee%22%3A%22https%3A%2F%2Fwineonline.example.org%2F%22%2C%22target%22%3A%22http%3A%2F%2Fexample.edu%2Fcredentials%2F3732%22%2C%22action%22%3A%5B%223rdPartyCorrelation%22%5D%7D%5D%7D%5D%2C%22proof%22%3A%7B%7D%7D

wyc:
@kdenhartog to close after the other PR is opened.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2021-08-11-vcwg#section3-2) on 2021-08-11

- no resolutions were taken

#### 3.2. removes mention of termsOfUse with verifiable presentations (pr vc-data-model#787)

_See github pull request [#787](https://github.com/w3c/vc-data-model/pull/787)._




**Kyle Den Hartog:** this needs to be closed after back and forth conversations. I found a way to fix this and will add a PR for that.  

**Brent Zundel:** should close once we have a PR merged  

**Wayne Chang:** I'll add a note for that  

</details>

kdenhartog:
I decided rather than fuss around with closing and opening PRs here, I would just force push to update this branch. Was easy enough to do, so this should be ready for review again taking into account the suggestions from @brentzundel

kdenhartog:
I've re-added the label for this. We didn't say if we needed to resend an email for that, so to error on the side of caution I did so in this email: https://lists.w3.org/Archives/Public/public-credentials/2021Aug/0200.html

It's a one time occurrence because we were going to close this one and open a new one, so I figured that's the best way to handle this scenario. I don't expect that every change to a PR will require an email and a reset to the 14 day merge, but I'll leave that to the chairs to decide.

brentzundel (approved):

msporny (approved):
I'm a bit wary of this change... but fine if we fix in v2.

index.html
```diff
@@ -2758,10 +2758,21 @@ <h3>Terms of Use</h3>
 {
   "@context": [
     "https://www.w3.org/2018/credentials/v1",
-    "https://www.w3.org/2018/credentials/examples/v1"
+    "https://www.w3.org/2018/credentials/examples/v1",
+    {
+        "VerifiablePresentationExtension": {
```
msporny, line 2763:
```suggestion
        "VerifiablePresentationTermsOfUseExtension": {
```

index.html
```diff
@@ -2758,10 +2758,21 @@ <h3>Terms of Use</h3>
...
+          "@id": "https://www.w3.org/2018/credentials#VerifiablePresentationExtension",
```
msporny, line 2764:
```suggestion
          "@id": "https://www.w3.org/2018/credentials/examples#VerifiablePresentationTermsOfUseExtension",
```

We should be wary of putting hacks in the credentials vocabulary.

index.html
```diff
@@ -2758,10 +2758,21 @@ <h3>Terms of Use</h3>
...
+              "@id": "cred:termsOfUse",
+              "@type": "@id"
+            }
+          }
+        }
+    }
   ],
   "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
-  "type": ["VerifiablePresentation"],
+  "type": ["VerifiablePresentation", "VerifiablePresentationExtension"],
```
msporny, line 2775:
```suggestion
  "type": ["VerifiablePresentation", "VerifiablePresentationTermsOfUseExtension"],
```

index.html
```diff
@@ -2778,9 +2789,11 @@ <h3>Terms of Use</h3>
       "name": "Bachelor of Science and Arts"
     }
     },
-    "proof": { <span class="comment">...</span> }
+    "proof": { 
+
+  }
```
msporny, line 2794: -1 to this change, it breaks consistency with the other examples. If you're worrying about `...` being copy-pasted, it won't be... we have CSS rules to prevent it from being copied to the paste buffer.

index.html
```diff
@@ -2758,10 +2758,21 @@ <h3>Terms of Use</h3>
 {
   "@context": [
     "https://www.w3.org/2018/credentials/v1",
-    "https://www.w3.org/2018/credentials/examples/v1"
+    "https://www.w3.org/2018/credentials/examples/v1",
+    {
+        "VerifiablePresentationExtension": {
+          "@id": "https://www.w3.org/2018/credentials#VerifiablePresentationExtension",
+          "@context": {
+            "termsOfUse": {
```
dlongley, line 2766: Examples should always use `@protected`.

```suggestion
            "@protected": true,
            "termsOfUse": {
```

index.html
```diff
@@ -2790,11 +2803,22 @@ <h3>Terms of Use</h3>
       "target": "http://example.edu/credentials/3732",
       "action": ["3rdPartyCorrelation"]
     }]
-  }</span>,
-  "proof": [ ... ]
+  }],
+  "proof": {}
```
msporny, line 2807: Breaks consistency...

index.html
```diff
@@ -2778,9 +2789,11 @@ <h3>Terms of Use</h3>
       "name": "Bachelor of Science and Arts"
     }
     },
-    "proof": { <span class="comment">...</span> }
+    "proof": { 
+
+  }
   }],
-  <span class="highlight">"termsOfUse": [{
+  "termsOfUse": [{
```
msporny, line 2796: Why are we removing the highlight? We highlight the parts of examples to draw people's attention to the feature being focused on in the example.

index.html
dlongley: And add `"@protected": true` at the top of this embedded context too.

msporny (changes requested):
Upon further reading, need a few changes in this PR for it to line up with the rest of the spec.

index.html
kdenhartog: ahh yes, good catch. I didn't mean to put it in that vocabulary. I'll be more careful about that.

index.html
kdenhartog: fixed, thanks for pointing out this issue. I've reverted it back to how it originally was.

index.html
kdenhartog: It was unintentional from quick copying and pasting from JSON-LD playground after confirming the document was valid. I didn't think about the fact I need to make it formatted back into HTML. Thanks for catching these things. I'll be more careful about keeping these things consistent going forward.

kdenhartog:
Thanks @msporny and @dlongley for the reviews. I've updated everything to align with your feedback.

index.html
```diff
@@ -2790,11 +2802,22 @@ <h3>Terms of Use</h3>
...
 
+        <p class="note">
+Warning: The <code>termsOfUse</code> property is improperly defined within the
+<code>VerifiablePresentation</code> scoped context. This is a bug with the
+version 1 context and will be fixed in the version 2 context. In the meantime,
+implementors who wish to use this feature will be required to extend the context
+of the <a>verifiable presentation</a> with an additional term that defines the
+<code>termsOfUse</code> property which can be used along side the <a>verifiable
+  presentation</a> type property in order for the term to be semantically recognized in a
+JSON-LD processor.
```
TallTed, line 2818:
```suggestion
of their <a>verifiable presentation</a> with an additional term that defines the
<code>termsOfUse</code> property, which can then be used alongside the
<a>verifiable presentation</a> type property, in order for the term to be
semantically recognized in a JSON-LD processor.
```

index.html
```diff
@@ -2758,10 +2758,22 @@ <h3>Terms of Use</h3>
 {
   "@context": [
     "https://www.w3.org/2018/credentials/v1",
-    "https://www.w3.org/2018/credentials/examples/v1"
+    "https://www.w3.org/2018/credentials/examples/v1",
+    {
+        "VerifiablePresentationTermsOfUseExtension": {
```
dlongley, line 2763:
```suggestion
        "@protected": true,
        "VerifiablePresentationTermsOfUseExtension": {
```

kdenhartog: Thanks for this addition. I thought we could intentionally drop this one since the term is protected (and only a single term is being defined in here. I'll go ahead and add this in, but for my knowledge what's the reason we need it listed in both places?

dlongley: The `@protected` flag below applies to the scoped context within the `VerifiablePresentationTermsOfUseExtension` term definition, thereby protecting all terms in *that* context. The one above applies to the context that defines `VerifiablePresentationTermsOfUseExtension`, protecting all terms in it. IOW, there are two contexts here -- and each one should set `@protected` to protect all terms in those contexts.

kdenhartog:
awaiting re-review from @msporny to make sure he's satisfied with the changes before merging

chaals:
Noting that the explanatory diagram includes termsOfUse, I'm watching this in relation to #786 but I think that given the grahp doesn't manage namespacing, and is an example, I would leave the termsOfUse there.

Although it is trivial to generate a version with it removed if the group prefers to do that.

kdenhartog:
> Noting that the explanatory diagram includes termsOfUse, I'm watching this in relation to #786 but I think that given the grahp doesn't manage namespacing, and is an example, I would leave the termsOfUse there.

That's where the PR has headed now as well, so I like the idea of leaving it in the graph in #786

kdenhartog:

PR #794, opened 2021-08-17
Making explicit the binding of the holder to a VC

swcurran:
This PR is an attempt to address #789 , adding, where appropriate (at least in my opinion), explicit references to the binding of a the holder (who will often be the subject) of the VC such that in a verifiable presentation the verifier can authenticate that the holder was issued the verifiable credential.

I've tried to follow the use of links to glossary items, but no doubt under- or over-used those links in places.  Likewise for where I have tagged JSON properties with the "code" tag.  Any guidance on that would be appreciated.

My main concern about this PR comes from the @David-Chadwick comment that `credentialSubject.id` may have become the de facto way to authenticate the holder, rather than using `holder.id`.  Starting from scratch, I think it makes far more sense to use `holder.id`, but I wonder if it is too late.  IMHO, I would much prefer that the VC Data Model had no opinion on the contents of the credential (it's just data) vs. explicitly saying there must be a subject and forcing all use cases to map (sometimes artificially) to the structure.  However, I understand that is not appropriate to change in the V1.x context. IMHO the use of `credentialSubject.id` vs.  `holder.id` is a by-product of that decision.

***
<a href="https://pr-preview.s3.amazonaws.com/swcurran/vc-data-model/pull/794.html" title="Last updated on Aug 17, 2021, 4:59 PM UTC (da7b669)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/794/1f574ef...swcurran:da7b669.html" title="Last updated on Aug 17, 2021, 4:59 PM UTC (da7b669)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -335,6 +335,9 @@ specification:
 A role an <a>entity</a> might perform by possessing one or more
 <a>verifiable credentials</a> and generating <a>verifiable presentations</a>
 from them. Example holders include students, employees, and customers.
+Issuers can bind a <a>verifiable credential</a> to the holder to which it is issued,
+and the holder can prove that binding using the <code>proof</code> property of a 
+<a>verifiable presentation</a>.
           </dd>
           <dt><a>issuer</a></dt>
           <dd>
@@ -444,6 +447,10 @@ a tamper-evident and privacy-respecting manner.
 different <a>issuers</a> into a single artifact, a
 <a>verifiable presentation</a>.
           </li>
+          <li>
+<a>Verifiable credentials</a> are often bound by the <a>issuer</a> to a specific
+<a>holder</a> at issuance time.
+          </li>           
           <li>
 <a>Issuers</a> can issue <a>verifiable credentials</a> about any <a>subject</a>.
           </li>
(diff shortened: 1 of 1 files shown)


swcurran:
I've linked my W3C and GitHub accounts. Do I need to submit a new PR? @msporny

agropper:
I would like to keep this issue open. It's confusing for the VC
specification to consider a Holder as separate from a Subject. This
distinction was useful in explaining one way to use VCs but it is not the
only way to use VCs. To the extent that VCs are linked to VPs, that is
still not a reason to introduce Holder as a normative concept.

Please see
https://github.com/decentralized-identity/waci-presentation-exchange/issues/93
for one of the reasons. As we have seen over five years of discussion on
this Holder vs. Subject issue, there are many other considerations.

swcurran:
This is a PR not the issue #789.  I closed this and opened #795 to try to address the IPR issue in this PR, but I still have that problem. I'm contacting @msporny (via email) for help.

agropper:
Whatever the process, I'm hoping to discuss this issue / PR where folks think it most appropriate.

msporny:
@swcurran wrote:
> I closed this and opened #795 to try to address the IPR issue in this PR

We typically don't close issues until they're either addressed by a PR, a WG resolution, or are closed w/o objections.

@agropper has objected to closing this issue with legitimate rationale, thus I'm re-opening.

msporny:
Nevermind, I thought I was on the issue :)

David-Chadwick:
I would like this PR to add a credentialHolder property with identical syntax to the credentialSubject property i.e. the credentialHolder can contain an optional id and optional holder properties. In this way the issuer can identify the holder by either an id (linked to cryptographic material) or via human readable properties (e.g. name, address, passport number) or both.

msporny:
> I would like this PR to add a credentialHolder property with identical syntax to the credentialSubject property i.e. the credentialHolder can contain an optional id and optional holder properties. In this way the issuer can identify the holder by either an id (linked to cryptographic material) or via human readable properties (e.g. name, address, passport number) or both.

Can of worms: Opened. :)

Wouldn't this cause all of the fields in `credentialSubject` to be duplicated in `credentialHolder`? How does this interact with the existing `holder` property? What about the `evidence` property?

@David-Chadwick, also note that this PR is closed, and you might want to comment here: #795

PR #795, opened 2021-08-17
Making explicit the binding of the holder to a VC

swcurran:
This PR is an attempt to address #789 , adding, where appropriate (at least in my opinion), explicit references to the binding of a the holder (who will often be the subject) of the VC such that in a verifiable presentation the verifier can authenticate that the holder was issued the verifiable credential.

I've tried to follow the use of links to glossary items, but no doubt under- or over-used those links in places.  Likewise for where I have tagged JSON properties with the "code" tag.  Any guidance on that would be appreciated.

My main concern about this PR comes from the @David-Chadwick comment that `credentialSubject.id` may have become the de facto way to authenticate the holder, rather than using `holder.id`.  Starting from scratch, I think it makes far more sense to use `holder.id`, but I wonder if it is too late.  IMHO, I would much prefer that the VC Data Model had no opinion on the contents of the credential (it's just data) vs. explicitly saying there must be a subject and forcing all use cases to map (sometimes artificially) to the structure.  However, I understand that is not appropriate to change in the V1.x context. IMHO the use of `credentialSubject.id` vs.  `holder.id` is a by-product of that decision.

***
<a href="https://pr-preview.s3.amazonaws.com/swcurran/vc-data-model/pull/795.html" title="Last updated on Nov 14, 2021, 6:00 PM UTC (6cc2160)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/795/6c429cf...swcurran:6cc2160.html" title="Last updated on Nov 14, 2021, 6:00 PM UTC (6cc2160)">Diff</a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,3 +29,33 @@ mailing list as well.
 
 ### Other useful links
 * [Public group email archive](https://lists.w3.org/Archives/Public/public-vc-wg/)
+
+## Process Overview for VC Data Model Pull Requests
+1. For now, we will focus only on merging new errata PRs into this repository,
+but encourage activity related to new features.
+2. Once a PR is opened, chairs and editors make judgement call on whether
+changes are substantive or editorial.
+<dl>
+  <dt>Editorial</dt>
+  <dd>Mark with "editorial" tag, merge into branch "v1.1"</dd>
+  <dt>Substantive</dt>
+  <dd>Mark with "substantial" tag. Bugfixes are merged into separate branch "v1.2". New Features stay around as an open PR.</dd>
+</dl>
+3. W3C CCG is notified of PRs that will be merged in the next 14 days if there
+are no objections.
+4. When it's determined a new reccomendation should go out, the W3C Verifiable
+Credentials Working Group members meet, review all the PRs that have been
+merged, and make a formal recommendation if agreement is reached.
+
+### Roadmap for 2021
+- 1 editorial update (v1.1?)
+- 1 substantive update (v1.2?)
+- VC Test Suite Refactoring
+- Start planning VC v2 Work, request a rechartering 3-6 months before end of
+  year to keep VC WG functioning.
+
+### Focus areas
+- [v1] Fixing a specific bug
+- [v1] Update examples in the spec to make them modern
+- [v2] VC `@context` needs updating, possibly with security vocab modularized
+  into smaller components instead of all included into a large context file.
--- a/index.html
+++ b/index.html
@@ -335,6 +335,9 @@ specification:
 A role an <a>entity</a> might perform by possessing one or more
 <a>verifiable credentials</a> and generating <a>verifiable presentations</a>
 from them. Example holders include students, employees, and customers.
+Issuers can bind a <a>verifiable credential</a> to the holder to which it is issued,
+and the holder can prove that binding using the <code>proof</code> property of a 
+<a>verifiable presentation</a>.
           </dd>
           <dt><a>issuer</a></dt>
           <dd>
@@ -444,6 +447,10 @@ a tamper-evident and privacy-respecting manner.
 different <a>issuers</a> into a single artifact, a
 <a>verifiable presentation</a>.
           </li>
+          <li>
+<a>Verifiable credentials</a> are often bound by the <a>issuer</a> to a specific
+<a>holder</a> at issuance time.
+          </li>           
           <li>
 <a>Issuers</a> can issue <a>verifiable credentials</a> about any <a>subject</a>.
           </li>
@@ -1032,7 +1039,15 @@ the <a>verifier</a> and <a>verified</a>.
   }
 }
         </pre>
-
+        <p>
+In this example, the <a>issuer</a> has bound the <a>verifiable credential</a> to
+the <a>holder</a> (Pat) using the
+<code>credentialSubject.id</code> property (rather than using the
+<code>holder.id</code> property). The second <code>proof</code> property in the
+<a>verifiable presentation</a> (with <code>proofPurpose</code> of
+<code>"authentication"</code>) is used by Pat to prove their relationship to the
+<a>credentialSubject.id</a> and thus, that they are the intended <a>holder</a>.
+        </p> 
         <p class="note">
 Implementers that are interested in understanding more about the
 <code>proof</code> mechanism used above can learn more in
@@ -1946,9 +1961,13 @@ is expected to be a <a>URI</a> for the entity that is generating the
           </dd>
           <dt><var>proof</var></dt>
           <dd>
-If present, the value of the <code>proof</code> <a>property</a> ensures that
-the <a>presentation</a> is <a>verifiable</a>. For details related to the use of
-this property, see Section <a href="#proofs-signatures"></a>.
+If present, the value of the <code>proof</code> <a>property</a> ensures that the
+<a>presentation</a> is <a>verifiable</a>. Where the issuer has bound the
+<a>verifiable credential</a> being presented to a specific <a>holder</a>, the
+<code>proof</code> <a>property</a>
+contains <a>verifiable</a> data to prove the indicated <a>holder</a> is
+presenting the proof. For details related to the use of this property, see
+Section <a href="#proofs-signatures"></a>.
           </dd>
         </dl>
 
@@ -2070,6 +2089,10 @@ An <a>issuer</a> <dfn data-lt="issue">issues</dfn> a
 any other actions involving a <a>credential</a>.
           </li>
           <li>
+In some cases, a <a>holder</a> might transfer to another <a>holder</a> one or more of its
+<a>verifiable credentials</a> that have not been bound by the <a>issuer</a> to that <a>holder</a>.
+          </li>
+          <li>
 A <a>holder</a> might <dfn data-lt="transfers">transfer</dfn> one or more of
 its <a>verifiable credentials</a> to another <a>holder</a>.
           </li>
@@ -2185,6 +2208,13 @@ trust could be weakened depending on the risk assessment of the <a>verifier</a>.
 All <a>entities</a> trust the <a>verifiable data registry</a> to be
 tamper-evident and to be a correct record of which data is controlled by which
 <a>entities</a>.
+          </li>
+          <li>
+In many cases, the <a>verifier</a> trusts the <a>issuer</a> to issue each
+<a>verifiable credential</a> to its intended <a>holder</a> (often the
+<a>subject</a>) and, where appropriate, to provide a <a>verifiable</a> binding
+in the
+<a>verifiable credential</a> to that <a>holder</a>.
           </li>
           <li>
 The <a>holder</a> and <a>verifier</a> trust the <a>issuer</a> to issue
@@ -2931,7 +2961,7 @@ ability of a <a>holder</a> to:
           <li>
 Combine multiple <a>verifiable credentials</a> from multiple <a>issuers</a> into
 a single <a>verifiable presentation</a> without revealing
-<a>verifiable credential</a> or <a>subject</a> identifiers to the
+<a>verifiable credential</a> or <a>holder</a> identifiers to the
 <a>verifier</a>. This makes it more difficult for the <a>verifier</a> to collude
 with any of the issuers regarding the issued <a>verifiable credentials</a>.
           </li>
@@ -3970,9 +4000,11 @@ the data in a <a>verifiable credential</a> while at rest.
 
       <p>
 <a>Subjects</a> of <a>verifiable credentials</a> are identified using the
-<code>credential.credentialSubject.id</code> field. The identifiers used to
-identify a <a>subject</a> create a greater risk of correlation when the
-identifiers are long-lived or used across more than one web domain.
+<code>credential.credentialSubject.id</code> field. <a>Holders</a> of
+<a>verifiable credentials</a> are identified using the
+<code>credential.holder.id</code> field. The identifiers used to identify a
+<a>subject</a> and/or <a>holder</a> create a greater risk of correlation when
+the identifiers are long-lived or used across more than one web domain.
       </p>
 
       <p>
@@ -4035,10 +4067,10 @@ properties of the cryptography used.
       <h3>Long-Lived Identifier-Based Correlation</h3>
 
       <p>
-<a>Verifiable credentials</a> might contain long-lived identifiers that could
-be used to correlate individuals. These types of identifiers include
-<a>subject</a> identifiers, email addresses, government-issued identifiers,
-organization-issued identifiers, addresses, healthcare vitals,
+<a>Verifiable credentials</a> might contain long-lived identifiers that could be
+used to correlate individuals. These types of identifiers include
+<a>subject</a> and <a>holder</a> identifiers, email addresses, government-issued
+identifiers, organization-issued identifiers, addresses, healthcare vitals,
 <a>verifiable credential</a>-specific JSON-LD contexts, and many other sorts of
 long-lived identifiers.
       </p>
@@ -4181,9 +4213,8 @@ not a concern or would not result in large economic or reputational losses.
 
       <p>
 <a>Verifiable credentials</a> that are <a>bearer credentials</a> are made
-possible by not specifying the <a>subject</a> identifier, expressed using the
-<code>id</code> <a>property</a>, which is nested in the
-<code>credentialSubject</code> <a>property</a>. For example, the following
+possible by not specifying either the <code>subjectCredential.id</code> or
+<code>holder.id</code> identifiers. For example, the following
 <a>verifiable credential</a> is a <a>bearer credential</a>:
       </p>
 
@@ -4405,8 +4436,8 @@ A <a>subject</a> identifier of a <a>credential</a> refers to the same
 <a>subject</a> across multiple <a>presentations</a> or <a>verifiers</a>. Even
 when different <a>credentials</a> are presented, if the <a>subject</a>
 identifier is the same, <a>verifiers</a> (and those with access to
-<a>verifier</a> logs) could infer that the <a>holder</a> of the
-<a>credential</a> is the same person.
+<a>verifier</a> logs) could infer that the <a>subject</a> of the
+<a>credential</a> is the same person. Likewise for <a>holder</a> identifiers.
         </li>
         <li>
 The underlying information in a <a>credential</a> can be used to identify an
@@ -4435,8 +4466,8 @@ by:
 
       <ul>
         <li>
-Using a globally-unique identifier as the <a>subject</a> for any given
-<a>credential</a> and never re-use that <a>credential</a>.
+Using a globally-unique identifier as the <a>subject</a> and/or <a>holder</a>
+for any given <a>credential</a> and never re-use that <a>credential</a>.
         </li>
         <li>
 If the <a>credential</a> supports revocation, using a globally-distributed
@@ -5026,25 +5057,43 @@ data fields in this specification by <a>verifiers</a>.
       </p>
 
       <section class="informative">
-        <h3>Credential Subject</h3>
+        <h3>Holder</h3>
 
         <p>
-In the <a>verifiable credentials</a> presented by a <a>holder</a>, the value
-associated with the <code>id</code> <a>property</a> for each
-<code>credentialSubject</code> is expected to identify a <a>subject</a> to the
-<a>verifier</a>. If the <a>holder</a> is also the <a>subject</a>, then
-the <a>verifier</a> could authenticate the <a>holder</a> if they have
-public key metadata related to the <a>holder</a>. The <a>verifier</a> could then
-authenticate the <a>holder</a> using a signature generated by the <a>holder</a>
-contained in the <a>verifiable presentation</a>. The <code>id</code>
-<a>property</a> is optional. <a>Verifiers</a> could use other <a>properties</a>
-in a <a>verifiable credential</a> to uniquely identify a <a>subject</a>.
+An <a>issuer</a> may indicate that a <a>verifiable credential</a> has been
+issued to a specific <a>holder</a> using the <code>holder</code>
+<a>property</a>. Alternatively, when the <a>holder</a> is the <a>subject</a>,
+the <code>credentialSubject.id</code> <a>property</a> could be used for the same
+purpose, ideally paired with the <code>nonTransferable</code> <a>property</a>.
+When such <a>verifiable credentials</a> are used to produce a <a>verifiable
+presentation</a>, the
+<a>verifier</a> may authenticate the <a>holder</a> (or holder <a>subject</a>)
+using a holder-generated signature in the <code>proof</code> <a>property</a> of
+the <a>verifiable presentation</a>.
+
+Where there is no <a>issuer</a>-designated <a>holder</a> the <a>verifiable
+credential</a> can be transferred to other <a>holders</a>, and no <a>holder</a>
+authentication is necessary by the <a>verifier</a>.
         </p>
 
         <p class="note">
 For information on how authentication and WebAuthn might work with
 <a>verifiable credentials</a>, see the Verifiable Credentials Implementation
 Guidelines [[VC-IMP-GUIDE]] document.
+        </p>
+
+      </section>
+
+      <section class="informative">
+        <h3>Credential Subject</h3>
+
+        <p>
+In the <a>verifiable credentials</a> presented by a <a>holder</a>, the value
+associated with the <code>id</code> property for each
+<code>credentialSubject</code> is expected to identify a <a>subject</a> to the
+<a>verifier</a>. The <code>id</code> property is optional. <a>Verifiers</a>
+could use other properties in a <a>verifiable credential</a> to uniquely
+identify a subject.
         </p>
 
       </section>
--- a/index.html
+++ b/index.html
@@ -448,8 +448,8 @@ different <a>issuers</a> into a single artifact, a
 <a>verifiable presentation</a>.
           </li>
           <li>
-<a>Verifiable credentials</a> are often bound by the <a>issuer</a> to a specific
-<a>holder</a> at issuance time.
+<a>Verifiable credentials</a> are sometimes associated by the <a>issuer</a> with
+a specific <a>holder</a> at issuance time.
           </li>           
           <li>
 <a>Issuers</a> can issue <a>verifiable credentials</a> about any <a>subject</a>.
--- a/index.html
+++ b/index.html
@@ -5068,7 +5068,7 @@ purpose, ideally paired with the <code>nonTransferable</code> <a>property</a>.
 When such <a>verifiable credentials</a> are used to produce a <a>verifiable
 presentation</a>, the
 <a>verifier</a> may authenticate the <a>holder</a> (or holder <a>subject</a>)
-using a holder-generated signature in the <code>proof</code> <a>property</a> of
+using a <a>holder</a>-generated signature in the <code>proof</code> <a>property</a> of
 the <a>verifiable presentation</a>.
 
 Where there is no <a>issuer</a>-designated <a>holder</a> the <a>verifiable
--- a/index.html
+++ b/index.html
@@ -4213,7 +4213,7 @@ not a concern or would not result in large economic or reputational losses.
 
       <p>
 <a>Verifiable credentials</a> that are <a>bearer credentials</a> are made
-possible by not specifying either the <code>subjectCredential.id</code> or
+possible by not specifying either the <code>credentialSubject.id</code> or
 <code>holder.id</code> identifiers. For example, the following
 <a>verifiable credential</a> is a <a>bearer credential</a>:
       </p>
--- a/index.html
+++ b/index.html
@@ -4467,7 +4467,7 @@ by:
       <ul>
         <li>
 Using a globally-unique identifier as the <a>subject</a> and/or <a>holder</a>
-for any given <a>credential</a> and never re-use that <a>credential</a>.
+for any given <a>credential</a>, and never re-using that <a>credential</a>.
         </li>
         <li>
 If the <a>credential</a> supports revocation, using a globally-distributed


index.html
```diff
@@ -5026,19 +5057,23 @@ <h2>Validation</h2>
...
+purpose, ideally paired with the <code>nonTransferable</code> <a>property</a>.
+When such <a>verifiable credentials</a> are used to produce a <a>verifiable
+presentation</a>, the
+<a>verifier</a> may authenticate the <a>holder</a> (or holder <a>subject</a>)
+using a holder-generated signature in the <code>proof</code> <a>property</a> of
+the <a>verifiable presentation</a>.
+
+Where there is no <a>issuer</a>-designated <a>holder</a> the <a>verifiable
+credential</a> can be transferred to other <a>holders</a>, and no <a>holder</a>
+authentication is necessary by the <a>verifier</a>.
```
msporny, line 5076: This implies that one must now always set both credential.holder and credential.credentialSubject.id in order to do holder authentication... which goes contrary to what we say elsewhere in the Validation section. This will need to be reconciled before we can merge.

msporny (changes requested):
I did a quick-scan of this PR and, overall, I'm fine with most of the text. The only exception is the new "Holder" section, which asserts things about the mental model of holders that I don't think we have consensus in the WG on (and thus, it'll need to be discussed).

index.html
```diff
@@ -335,6 +335,9 @@ <h3>Ecosystem Overview</h3>
 A role an <a>entity</a> might perform by possessing one or more
 <a>verifiable credentials</a> and generating <a>verifiable presentations</a>
 from them. Example holders include students, employees, and customers.
+Issuers can bind a <a>verifiable credential</a> to the holder to which it is issued,
+and the holder can prove that binding using the <code>proof</code> property of a
```
dlongley, line 339: I think we need to find some other language here that states that the issuer can *assert* what they saw vs. saying that the issuer can "bind a VC to a holder". At least one plain reading of that statement would render it false as the issuer has no such power. We'd need a clear definition of "bind" that avoided that issue.

What we know is that the issuer can create some kind of assertion about what they saw when issuing the VC -- and that a presentation may include a proof such that some kind of narrowing conclusions can be drawn from it. For example the issuer may have required the holder to prove they have access to some API and then the verifier can then require the presenter (who may not be the holder) to prove that they also have access to that API. Of course, that sort of language would highlight that the holder does not have to be the same as the presenter (they are not actually "bound") and, on its own, would make the feature seem powerless to the reader. This is the other side of the spectrum of what's written in this PR now: that the issuer or this "binding" has more power than it actually does.

Maybe we need to discuss how issuers can make assertions that can create additional constraints for the verifier to take advantage of via presentation requirements -- and that each of these constraints must be weighed carefully so as not to be over-estimated in what they accomplish?

index.html
```diff
@@ -3970,9 +4000,11 @@ <h3>Identifier-Based Correlation</h3>
 
       <p>
 <a>Subjects</a> of <a>verifiable credentials</a> are identified using the
-<code>credential.credentialSubject.id</code> field. The identifiers used to
-identify a <a>subject</a> create a greater risk of correlation when the
-identifiers are long-lived or used across more than one web domain.
+<code>credential.credentialSubject.id</code> field. <a>Holders</a> of
+<a>verifiable credentials</a> are identified using the
+<code>credential.holder.id</code> field. The identifiers used to identify a
```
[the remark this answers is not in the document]
David-Chadwick: I would like this PR to add a credentialHolder property (between sections 4.4 and 4.5) with identical syntax to the credentialSubject property i.e. the credentialHolder can contain an optional id and optional holder properties. In this way the issuer can identify the holder by either an id (linked to cryptographic material) or via human readable properties (e.g. name, address, passport number) or both.

index.html
```diff
@@ -4435,8 +4466,8 @@ <h3>Usage Patterns</h3>
 
       <ul>
         <li>
-Using a globally-unique identifier as the <a>subject</a> for any given
-<a>credential</a> and never re-use that <a>credential</a>.
+Using a globally-unique identifier as the <a>subject</a> and/or <a>holder</a>
+for any given <a>credential</a> and never re-use that <a>credential</a>.
```
TallTed, line 4470:
```suggestion
for any given <a>credential</a>, and never re-using that <a>credential</a>.
```

index.html
```diff
@@ -445,6 +448,10 @@ <h3>Use Cases and Requirements</h3>
 <a>verifiable presentation</a>.
           </li>
           <li>
+<a>Verifiable credentials</a> are often bound by the <a>issuer</a> to a specific
+<a>holder</a> at issuance time.
```
TallTed, line 452:
```suggestion
<a>Verifiable credentials</a> are sometimes associated by the <a>issuer</a> with
a specific <a>holder</a> at issuance time.
```

index.html
```diff
@@ -1032,7 +1039,15 @@ <h3>Concrete Lifecycle Example</h3>
   }
 }
         </pre>
-
+        <p>
+In this example, the <a>issuer</a> has bound the <a>verifiable credential</a> to
+the <a>holder</a> (Pat) using the
+<code>credentialSubject.id</code> property (rather than using the
+<code>holder.id</code> property). The second <code>proof</code> property in the
```
TallTed, line 1046:
```suggestion
In this example, the <a>issuer</a> has bound the <a>verifiable credential</a> to
the <a>holder</a> (Pat) using the
<code>credentialSubject.id</code> property (rather than using the
<code>holder.id</code> property). The second <code>proof</code> property in the
```
Aside from the continuing issue of inappropriate use of "binding", there is NOTHING in the existing doc (unless it snuck by me, which I hope it did not!) that says that the `credentialSubject.id` does double-duty as a `holder.id`, associating a VC with a specific Holder at issuance.  We spent many hours discussing this, and Holder-sameAs-Subject was neither determined to be the most-common nor the default scenario.  If anything, I remember us coming up with more Holder-**NOT**-sameAs-Subject scenarios.

index.html
```diff
@@ -1946,9 +1961,13 @@ <h3>Presentations</h3>
           </dd>
           <dt><var>proof</var></dt>
           <dd>
-If present, the value of the <code>proof</code> <a>property</a> ensures that
-the <a>presentation</a> is <a>verifiable</a>. For details related to the use of
-this property, see Section <a href="#proofs-signatures"></a>.
+If present, the value of the <code>proof</code> <a>property</a> ensures that the
+<a>presentation</a> is <a>verifiable</a>. Where the issuer has bound the
+<a>verifiable credential</a> being presented to a specific <a>holder</a>, the
```
TallTed, line 1966: As above.

index.html
```diff
@@ -2070,6 +2089,10 @@ <h3>Lifecycle Details</h3>
 any other actions involving a <a>credential</a>.
           </li>
           <li>
+In some cases, a <a>holder</a> might transfer to another <a>holder</a> one or more of its
+<a>verifiable credentials</a> that have not been bound by the <a>issuer</a> to that <a>holder</a>.
```
TallTed, line 2093: Binding again.  Also appears to presume that association with Holder is default for VCs.

index.html
```diff
@@ -2187,6 +2210,13 @@ <h2>Trust Model</h2>
 <a>entities</a>.
           </li>
           <li>
+In many cases, the <a>verifier</a> trusts the <a>issuer</a> to issue each
+<a>verifiable credential</a> to its intended <a>holder</a> (often the
+<a>subject</a>) and, where appropriate, to provide a <a>verifiable</a> binding
+in the
+<a>verifiable credential</a> to that <a>holder</a>.
```
TallTed, line 2217: Presumes Holder-sameAs-Subject facts not in evidence.  Also binding, again.

index.html
```diff
@@ -2931,7 +2961,7 @@ <h3>Zero-Knowledge Proofs</h3>
           <li>
 Combine multiple <a>verifiable credentials</a> from multiple <a>issuers</a> into
 a single <a>verifiable presentation</a> without revealing
-<a>verifiable credential</a> or <a>subject</a> identifiers to the
+<a>verifiable credential</a> or <a>holder</a> identifiers to the
```
TallTed, line 2964: You cannot replace `subject` with `holder` like this. It *might* work out if you added `holder` to `subject`, but only if you use `associated` or similar instead of `bound`; there may be other issues with new draft text.

index.html
```diff
@@ -5026,19 +5057,23 @@ <h2>Validation</h2>
       </p>
 
       <section class="informative">
-        <h3>Credential Subject</h3>
+        <h3>Holder</h3>
```
TallTed, line 5060: <strike>These are not interchangeable in this way.  *Maybe* an added `Holder` section will work.  Replacing the `Credential Subject` section does not work for me at all.</strike>

...

Now I've read further along, and I see that you did add a `Holder` section.  You just confused me by putting this new section ahead of the existing `Credential Subject` section, which order doesn't seem logical to me.  What is this ordering based on?

TallTed (changes requested):

mwherman2000:
First, a caveat, I haven't been following this thread in detail but @David-Chadwick and I had a discussion this "morning" that might be relevant/useful.  

It was regarding how the different `id` fields can be used/applied in the context of a business document represented as a VC (e.g. a purchase order originating from an accounting system (the system of record) that gave this particular purchase order a purchase order number (e.g. id) of `2222`).

In a generic purchase order scenario (I haven't coded one up yet), credential id is used as the identifier for the VC _envelope_ that wraps the credentialSubject.  The credentialSubject id represents the _subject_ of the claims that are also contained in the credentialSubject. The credentialSubject is where the purchase order data is stored (e.g. purchase order id identifier plus the purchase order form data represented as claims). Each VC envelope has a unique credential id. Each credentialSubject has a unique credentialSubject id.

In a specific purchase order scenario, credential id might have a value of `did:vcenvelope:1111` and credentialSubject id might have a value of `did:purchaseorder:2222`.  The credentialSubject id (aka purchase order id identifier) is unique and generated by the accounting system of record (as is typical).  Each time a purchase order `did:purchaseorder:2222` is sent to a Holder, it is "placed in a new envelope` that has a new did:vcenvelope identifier (e.g. `did:vcenvelope:1111`, `did:vcenvelope:1121`, `did:vcenvelope:1211`, ...

What about the Holder id part of the discussion? 

First, the Holder id is not the same as the VC envelope or the purchase order credentialSubject. In the purchase order scenario it isn't needed.  The originator of the purchase order (purchaser) will be encoded the purchase order claims data.  

If a Holder id is required for other use case, it can be encoded before or after the `@context` for example (i.e. the top part of the VC) ...e.g. in the _packing label_.

Although the following isn't a purchase order example, I'm including it because IMO it suggests some useful names for the anatomy of a VC.

image

index.html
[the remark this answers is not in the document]
swcurran: @msporny -- that is not what I meant. I would expect that information to be put in by the **issuer** into the VC.

index.html
```diff
@@ -4181,9 +4213,8 @@ <h3>Bearer Credentials</h3>
 
       <p>
 <a>Verifiable credentials</a> that are <a>bearer credentials</a> are made
-possible by not specifying the <a>subject</a> identifier, expressed using the
-<code>id</code> <a>property</a>, which is nested in the
-<code>credentialSubject</code> <a>property</a>. For example, the following
+possible by not specifying either the <code>subjectCredential.id</code> or
```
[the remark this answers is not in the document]
swcurran: Corrected.

index.html
```diff
@@ -5026,19 +5057,23 @@ <h2>Validation</h2>
...
-associated with the <code>id</code> <a>property</a> for each
-<code>credentialSubject</code> is expected to identify a <a>subject</a> to the
-<a>verifier</a>. If the <a>holder</a> is also the <a>subject</a>, then
-the <a>verifier</a> could authenticate the <a>holder</a> if they have
-public key metadata related to the <a>holder</a>. The <a>verifier</a> could then
-authenticate the <a>holder</a> using a signature generated by the <a>holder</a>
-contained in the <a>verifiable presentation</a>. The <code>id</code>
-<a>property</a> is optional. <a>Verifiers</a> could use other <a>properties</a>
-in a <a>verifiable credential</a> to uniquely identify a <a>subject</a>.
+An <a>issuer</a> may indicate that a <a>verifiable credential</a> has been
```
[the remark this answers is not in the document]
swcurran: Changed

index.html
David-Chadwick: We already have the ToU property for the issuer to place constraints on the issued VC. So instead of having a new holder property in the VC, that ranks pari-parsu with the credentialSubject we could place the holder property in a new ToU type.
Remembering of course that the verifier can always accept any VC regardless of who presents it, or what the ToUs may state, or what its validity period is, or what its revocation status is.

msporny: > So instead of having a new holder property in the VC, that ranks pari-parsu with the credentialSubject we could place the holder property in a new ToU type.

I would far prefer that we go down this route. Thank you for suggesting it @David-Chadwick.

I do also think that we need to make sure that no one believes that it's possible to "bind" a holder to a VC and believe that any sort of cryptographic binding solves the problem of "is the person I'm interacting w/ remotely/in person the same as the holder". That problem is not solved by cryptography.

index.html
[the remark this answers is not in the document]
msporny: > @msporny -- that is not what I meant. I would expect that information to be put in by the **issuer** into the VC.

Yes, understood now. When I made the suggestion, I misunderstood what you were going for... but since then, I think we're noting that doing this is problematic. I like the proposal @David-Chadwick is putting forward around "Terms of Use" or "Evidence".

agropper:
From a privacy engineering perspective, introducing Holder as a normative
element is a problem no matter how it's done because it gives the Issuer
power that they do not currently have with legacy credentials, some of
which use biometrics to bind the subject to the verifiable integrity
mechanism of the credential itself.

There will be use-cases where the Subject is not the Holder. A VC that
represents a capability would be one example and this was discussed some
months ago. I believe the consensus that emerged is that there's no good
reason to code capabilities as standard VCs, but I could be wrong.

A real-world example of a capability is the Release of Information (ROI)
form that, for example, gives the Social Security Disability Office (SSDO)
the right to request a health record directly from a hospital. The
patient/subject is asked to configure (hospital name, authorized delegate
name, records date range, HIV exclusion, etc...) and sign the ROI and
return it to the SSDO as the authorized delegate. A similar sequence is
used to delegate a car dealer's bank to access a credit report on the
purchaser.

In these fairly common use-cases, there is no obvious holder. The health
record or credit report VC is issued, by design, directly to the Verifier.
The delegate name on the ROI form is the Verifier.

My point goes back to Manu's statement:

> I do also think that we need to make sure that no one believes that it's
> possible to "bind" a holder to a VC and believe that any sort of
> cryptographic binding solves the problem of "is the person I'm interacting
> w/ remotely/in person the same as the holder". That problem is not solved
> by cryptography.


We know it's possible to bind a Subject to a VC by adding a biometric
attribute or a link to a biometric attribute to the VC. What are the
use-cases where the Issuer needs to "bind" a holder to a VC?

- Adrian

msporny:
> We know it's possible to bind a Subject to a VC by adding a biometric attribute or a link to a biometric attribute to the VC. 

Yes, sorry, what I meant was "I do also think that we need to make sure that no one believes that it's possible to **CRYPTOGRAPHICALLY** "bind" a holder to a **VC AND THEN ONLY DEPEND ON THAT CRYPTOGRAPHIC BINDING**". What you get in that case is something that's completely untraceable and open to wide-scale abuse (almost perfect anonymity, which is ok for some low-risk use cases like movie tickets and single use bus passes -- things greatly constrained by time and space).

agropper:
What about
https://github.com/HIEofOne/Trustee-Community/blob/master/Biometric-Health-Card.pdf
?

David-Chadwick:
On 21/08/2021 19:43, Adrian Gropper
      wrote:
    
    What
      are the
      use-cases where the Issuer needs to "bind" a holder to a VC?
    These can only be use cases in which the holder NE subject
      (because when holder==subject, binding to subject is sufficient to
      simultaneously bind to holder).
    There are many use cases where the issuer gives the VC to a
      holder NE subject, and the verifier needs to know that the
      presenter is the original authorised holder, and not someone who
      copied the VC from the genuine holder and is passing this off as
      belonging to them. Examples include: holder is acting as power of
      attorney (proxy) for the subject, holder is the authorised
      representative of an inanimate object (such as an organisation).
      In other words, where the subject is not able to act on its own
      behalf and cannot perform the delegation/authorisation itself.
    kind regards
    David

mwherman2000:
> There are many use cases where the issuer gives the VC to a holder NE subject, and the verifier needs to know that the presenter is the original authorised holder, and not someone who copied the VC from the genuine holder and is passing this off as belonging to them. 

@David-Chadwick I think most of the business document use cases we've been discussing also fall into this camp (e.g. a Purchase Order).

David-Chadwick:
@mwherman2000 Yes if a purchase order is an inanimate object, given to a buyer by the company (issuer) to hold until next month say, the supplier does not want a fraudsters to present the PO now, causing it to produce the goods, when the buyer did not intend the goods to be bought until next month. One can envisage many other examples like this, where wrongly presented business documents could cause financial damage to either the issuer or the verifier.

mwherman2000:
Agreed but to some extent this depends on the _completeness_ and _immutability_ of the purchase order object represented inside the credentialSubject.  Delivery terms, conditions, dates, payment terms, conditions, parties, etc. can and should be specified in the purchase order object (they are).  Nothing can be added or changed to a signed, verifiable purchase order.

- PO sample schema: http://docs.oasis-open.org/ubl/os-UBL-2.2/UBL-2.2.html#S-INVOICE-SCHEMA
- PO XML example: http://docs.oasis-open.org/ubl/os-UBL-2.2/xml/UBL-Invoice-2.1-Example.xml
- Invoice JSON examples: http://docs.oasis-open.org/ubl/UBL-2.1-JSON/v1.0/cnd02/UBL-2.1-JSON-v1.0-cnd02.html#S-EXAMPLE-UBL-JSON-DOCUMENTS

...or am I missing something?

swcurran:
> A real-world example of a capability is the Release of Information (ROI) form that...

My opinion on this example, is that what is described ("issued...to the Verifier") is not a verifiable credential use case. Unless the recipient (the SSDO) needs to prove the data to another party (an actual verifier), then that is not a vc -- it's just releasing data to the SSDO. If they use a VC data format, that's (perhaps) convenience, but does not make it a (lower case) vc.  If the SSDO does need to prove the data in the future, then it is a vc, and the holder is the SSDO, and the verifier should know that.

swcurran:
@msporny -- `VC AND THEN ONLY DEPEND ON THAT CRYPTOGRAPHIC BINDING` does that statement go back to Dave Longley's article, or is it more than that?  The caps imply that written in stone vs. opinion and I would like to be clear on what you mean.

mwherman2000:
@David-Chadwick 
> Yes if a purchase order is an inanimate object, given to a buyer by the company (issuer) to hold until next month say, the supplier does not want a fraudsters to present the PO now, causing it to produce the goods, when the buyer did not intend the goods to be bought until next month

In this scenario, let's say the delayed purchase order had the value of `credential.credentialSubject.id` set `did:abcgrocery:po:1234`. Are you suggesting that someone could create a fake PO with the same `credentialSubject.id`? It's possible but it won't be verifiable as originating from ABC Grocery. 

(Just in case it helps further the discussion the outer _VC envelope_ would have a `credential.id` of, say, `did:abcgrocery:envelope:123-456-789`.)

msporny:
> does that statement go back to Dave Longley's article, or is it more than that?

Yes, it does.

> The caps imply that written in stone vs. opinion and I would like to be clear on what you mean.

The only thing the caps were meant to imply was clarifying the loose language I used the first time around. The caps were added to try and add some clarity to what I was saying to my initial statement.

swcurran:
Thanks.  I'm definitely not an expert in this, but my understanding is that the claim (in your comment) that such a cryptographic binding is impossible is not considered fact, and opinions differ.  Since it would be extremely helpful if a practical mechanism was found that sufficiently mitigates the risk of using such a cryptographic binding, I don't want anyone to turn away from trying to accomplish that because some think it can't be practical.

David-Chadwick:
@mwherman2000 "Are you suggesting that someone could create a fake PO ". No I never said or implied that. I simply said that anyone having possession of the genuine PO can present it, even if that person was not the holder that the issuer gave it to.

mwherman2000:
RE: I simply said that anyone having possession of the genuine PO can present it

Yes, there's good solutions for that too ...like only accepting business documents over trusted, verifiable channels established between a particular Purchaser (ABC Grocery) and a particular Supplier (David's Cabbages) (e.g. using DIDComm-like or HTTPS communications channels). Worst case, it would be detected as a duplicate PO.

mwherman2000:
There's other VC transmission strategies too. 

For example, instead of the Originator's Agent (Originating Holder's Agent) sending (shooting) over the actual PO VC to the Receiver's Agent (Receiving Holder's Agent), the Originator('s Agent) sends the Receiver(('s Agent) a _Notification VC_ containing the DID of the Originator's Agent and the DID of the particular PO envelope.  

The Receiver then looks up the URI of the Originator (and its Agent's Service Endpoint) on a VDR and uses that trusted URI to call back to the Originator to download the PO envelope containing the PO content (PO number and the claims describing the claims that comprise the PO). The latter may be a recursive process if the PO is represented as a hierarchy or graph of related PO component VCs. ...or something like that. 

Of course, there's all sorts of cross verifications made a long the way.

agropper:
The Power of Attorney document lists the identity of the delegate as a
claim and signs it (often with a witness). The identity of the holder is
completely irrelevant.

An auto registration is an example of your inanimate “thing”. The subject
is the thing. The responsible parties include the owner of the thing (for
transfer of title purposes), the person who registered the thing (to get
the parking tickets), and the insurance company (in case anyone wants to
validate insurance). These are three separate identities as claims and I
don’t see a holder in this case either.

My point remains that identifying a holder is a privacy engineering problem
because it limits the choices of the subject or whoever lease has the VC.

The “double spend” or VC copy issue (a theater ticket) is not mitigated by
identifying a holder or a subject.

What am I missing?

mwherman2000:
> An auto registration is an example of your inanimate “thing”. The subject is the thing. The responsible parties include the owner of the thing (for transfer of title purposes), the person who registered the thing (to get the parking tickets), and the insurance company (in case anyone wants to validate insurance). These are three separate identities as claims and I don’t see a holder in this case either.

I beg to differ @agropper using your own POA example. 

Just like in your POA example, the value of `credentialSubject id` is the identifier for the auto registration document itself; then there is a claim in that document that will reference/deference to, for example, the vehicle's VIN number.

(The value of `credential id` is the identifier associated with the _credential envelope_ that the auto registration document was delivered in.)

swcurran:
@agropper

> ... The identity of the holder is completely irrelevant. ...

The holder is who the issuer intended should have the credential, and we think that is important.  Obviously, when the subject and holder are the same, it's easy.  Where they are not, then the holder matters.  If/when delegation is supported and standardized, the holder of the original VC becomes an issuer to a new holder of a new VC.  I don't understand how delegation can be done by simply transferring a VC to another, with no indication of intent.

Depending on how you model the POA VC, the holder would be extremely if they are one being given the POA for another, and that other is considered the subject.

IMHO VCs are not suitable for solving general "double spend" problems. It works if there is only one verifier (e.g., the movie ticket), but does not work for something like a prescription, where the holder can take it to many independent pharmacies.

index.html
```diff
@@ -5026,19 +5057,23 @@ <h2>Validation</h2>
...
+issued to a specific <a>holder</a> using the <code>holder</code>
+<a>property</a>. Alternatively, when the <a>holder</a> is the <a>subject</a>,
+the <code>credentialSubject.id</code> <a>property</a> could be used for the same
+purpose, ideally paired with the <code>nonTransferable</code> <a>property</a>.
```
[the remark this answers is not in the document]
swcurran: Removed.  I see that the property is only referenced in the (IMHO) very confusing Appendix C, and is probably not the way to go.

index.html
```diff
@@ -5026,19 +5057,23 @@ <h2>Validation</h2>
...
-<a>verifier</a>. If the <a>holder</a> is also the <a>subject</a>, then
-the <a>verifier</a> could authenticate the <a>holder</a> if they have
-public key metadata related to the <a>holder</a>. The <a>verifier</a> could then
-authenticate the <a>holder</a> using a signature generated by the <a>holder</a>
-contained in the <a>verifiable presentation</a>. The <code>id</code>
-<a>property</a> is optional. <a>Verifiers</a> could use other <a>properties</a>
-in a <a>verifiable credential</a> to uniquely identify a <a>subject</a>.
+An <a>issuer</a> may indicate that a <a>verifiable credential</a> has been
+issued to a specific <a>holder</a> using the <code>holder</code>
+<a>property</a>. Alternatively, when the <a>holder</a> is the <a>subject</a>,
```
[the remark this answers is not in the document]
swcurran: This is what I find amazing and is the core of this change.  Figure 1 has holder in it, so someone thought it might be important.  It is included in the [base context](https://www.w3.org/TR/vc-data-model/#base-context).

I've heard this from another person from the original working group, so despite it's inclusion in the context, I gather it is not intended to be used.

index.html
```diff
@@ -5026,19 +5057,23 @@ <h2>Validation</h2>
...
+the <code>credentialSubject.id</code> <a>property</a> could be used for the same
+purpose, ideally paired with the <code>nonTransferable</code> <a>property</a>.
+When such <a>verifiable credentials</a> are used to produce a <a>verifiable
+presentation</a>, the
+<a>verifier</a> may authenticate the <a>holder</a> (or holder <a>subject</a>)
```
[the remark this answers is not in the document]
swcurran: Changed

index.html
swcurran: Changed

index.html
swcurran: Changed.

index.html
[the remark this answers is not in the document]
mwherman2000: > Alternatively, when the holder is the subject, the credentialSubject.id property could be used for the same purpose, ideally paired with the nonTransferable property.

1. _when the holder is the subject_ - how is the prescriptive test for this (in software)? Which properties are being compared?
2. _the credentialSubject.id property could be used for the same purpose_ - can be more direct in stating that the _same purpose_ is ...just write out?

agropper:
@swcurran said:
“The holder is who the issuer intended should have the credential, and we
think that is important.”

So, an example would be a Letter of Recommendation (the subject should not
see the content of the VC) or a Credit Report (the issuer wants to be paid
each time the report is requested). How is the holder involved in these
cases?

Delegated authorization deals with the Letter of Recommendation and Credit
Report issuance effortlessly.

Please provide some real-world examples of holder different from subject or
delegate.

- Adrian

mwherman2000:
I know I'm being terrible simplistic but in this conversation, the "Holder" can/should/could be viewed as anyone who is hold the VC (or a copy of the VC) at any point in its lifecycle from issuance to revocation.

image

David-Chadwick:
@agropper "Please provide some real-world examples of holder different from subject or delegate." I already did. And so did you by adding different claims to a VC that specified the attorney or the owner or other claims identifying a third party who is not the subject. What you are effectively doing is saying 'these claims identify authorised entities' and then the verifier can determine from the VP if the holder is one of these authorised entities. But your solution is specific to the type of VC, and varies for each type, whereas what we are trying to do is define a generic method that is applicable to all types of VC where the holder NE subject.

agropper:
So, the holder becomes a special case of Terms of Use for a credential
where the ToU happens to be an identity?

David-Chadwick:
ToU is one possibility. So is having a Holder property. It is for the W3C group to decide
a) if a standardised generic method is needed for the issuer to signal who it gave the VC to, and
b) if it is, what this mechanism should be.

agropper:
Thanks for the clarification! Either way, the decision looks like an added
privilege of the Issuer and might raise privacy issues.

ToU seems much clearer to me than Holder. In general, for any given type of
credential, as a Subject, I would want any constraints added by the Issuer
to be very obvious.

Holder, on the other hand, is an identity attribute and, as such, dilutes
the minimum privilege benefits of a capabilities architecture. Also, Holder
as an added optional identity in the VC, is unusually confusing as our
years of debate has shown and continues to show.

OR13:
I opened https://github.com/w3c/vc-data-model/issues/923. 

can we close this now?

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2022-09-07-vcwg#section1-1) on 2022-09-07

- no resolutions were taken

#### 1.1. Making explicit the binding of the holder to a VC (pr vc-data-model#795)

_See github pull request [vc-data-model#795](https://github.com/w3c/vc-data-model/pull/795)._




**Manu Sporny:** 795 is going to remain open until we discuss the holder stuff at TPAC.  

</details>

README.md
```diff
@@ -29,3 +29,33 @@ mailing list as well.
...
+changes are substantive or editorial.
+<dl>
+  <dt>Editorial</dt>
+  <dd>Mark with "editorial" tag, merge into branch "v1.1"</dd>
+  <dt>Substantive</dt>
+  <dd>Mark with "substantial" tag. Bugfixes are merged into separate branch "v1.2". New Features stay around as an open PR.</dd>
+</dl>
+3. W3C CCG is notified of PRs that will be merged in the next 14 days if there
+are no objections.
+4. When it's determined a new reccomendation should go out, the W3C Verifiable
```
decentralgabe, line 46:
```suggestion
4. When it's determined a new recommendation should go out, the W3C Verifiable
```

README.md
```diff
@@ -29,3 +29,33 @@ mailing list as well.
...
+but encourage activity related to new features.
+2. Once a PR is opened, chairs and editors make judgement call on whether
+changes are substantive or editorial.
+<dl>
+  <dt>Editorial</dt>
+  <dd>Mark with "editorial" tag, merge into branch "v1.1"</dd>
+  <dt>Substantive</dt>
+  <dd>Mark with "substantial" tag. Bugfixes are merged into separate branch "v1.2". New Features stay around as an open PR.</dd>
+</dl>
+3. W3C CCG is notified of PRs that will be merged in the next 14 days if there
```
decentralgabe, line 44: how?

PR #816, opened 2021-09-09
clarification of schema property

David-Chadwick:


kdenhartog (approved):
Good catch, this seems like an obvious change that just slipped through and won't cause any sort of breaking change. However, it does appear to be adding a normative statement so I'm going to classify it as a V1.2 change which I think is fine to go through during the maintainence WG.

David-Chadwick:
There is an issue about whether a holder should insert a schema property into the VP that it issues, because there are going to be multiple flavours of VPs.
Finally, should we add a note to say that the schema is not intended to be constrained to defining only the properties of the credentialSubject (which some people have wrongly thought to be the case, even though I think the current text is clear that it applies to the entire VC)

index.html
```diff
@@ -2477,10 +2477,11 @@ <h3>Data Schemas</h3>
 data syntax, nor enables the definition of arbitrary encodings to alternate
 representation formats.
         </p>
-
+       
         <p>
 This specification defines the following <a>property</a> for the expression of a
-data schema:
+data schema, which MAY be included by an <a>issuer</a> in
```
msporny, line 2483: I don't disagree with this statement, but keep in mind that non-testable MAY statements are effectively worthless -- you can't assert conformance either way. Sure, someone MAY do this, but they may also do many other things that are unstated. I suggest just changing this from "MAY" to "can" -- don't use normative language for stuff that's not testable.

msporny (changes requested):
Fine with the concept, but no need to use normative language.

index.html
kdenhartog: If we remove this, we can downgrade this to a V1.1 change as well.

David-Chadwick: Happy to replace MAY with can. I thought that because MAY does not affect conformance then the edit would qualify as v1.1.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2021-09-15-vcwg#section3-4) on 2021-09-15

- no resolutions were taken

#### 3.4. clarification of schema property (pr vc-data-model#816)

_See github pull request [#816](https://github.com/w3c/vc-data-model/pull/816)._




**Brent Zundel:** Next: PR 816  

**David Chadwick:** I raised this because it was on a different list - TallTed asked me to point to it here.  
… Some people are creating schemas that cover the credential subject and not the whole credential.  
… I thought this could be because the text is not clear.  
… First thing I found is that there is no statement that the credentialSchema must be present.  
… But that would be a breaking change. So made it a "CAN".  
… I think the text is clear that the schema property is for the entire credential.  
… But there are now two other issues which I've put there: what about verifiable presentations?  
… There could be different flavors of VPs... If the group agrees, we could put schema for VPs.  
… The second issue: Should we put a note in, to specifically say that the schema is not intended to apply just to the subject.  

> *Kyle Den Hartog:* +1 I'm good with adding those to this PR

**Brent Zundel:** I am personally fine with those additional pieces of text being added to this PR, just because it is part of the same conversation we've been having.  
… I'm not opposed to the addition of normative language if it's needed to address errata.  
… I agree with Manu that in this case it would be difficult to say how this would be testable. Making a normative statement may not give us the assurances we are hoping for. And as you said, changing it to a "CAN" is fine.  
… So the normative change would be to add a schema property for use in the VP. I think that is arguably errata and could be incorporated in the revisions we make, as long as it goes in this review period.  
… Any questions, comments, before moving on to the next PR?  

**Manu Sporny:** o/  

</details>

index.html
kdenhartog: My general thinking here is that because we're adding normative language that it would make it a V1.2 change due to the following clause in [6.1.3 class 3](https://www.w3.org/2020/Process-20200915/#correction-classes)

```
clears up an ambiguity or under-specified part of the specification in such a way that data, a processor, or an agent whose conformance was once unclear becomes clearly either conforming or non-conforming.
```

Without the normative keyword addition, I generally see it as following under the following statement instead:

```
clarifying informative use cases or other non-normative text
```

kdenhartog:
```suggestion
data schema, which can be included by an <a>issuer</a> in  
```

kdenhartog:
I've relabeled this to a V1.1 change pending the update (change "MAY" to "can") in accordance with the discussion we had about this PR on the 9/15 call.

David-Chadwick:
I hope I have now accepted MAY to can using the web interface

kdenhartog:
Looks like it didn't go through quite properly when I just reviewed it, but I was able to get it all fixed up easy enough. Looks like everything should be good to go for this PR based on current feedback.

brentzundel (approved):

kdenhartog:
@msporny since you still have requested changes (I think they've all be addressed here at this point) I'll leave this one to you to merge.

index.html
```diff
@@ -2477,10 +2477,11 @@ <h3>Data Schemas</h3>
 data syntax, nor enables the definition of arbitrary encodings to alternate
 representation formats.
         </p>
-
+
```
dlongley, line 2480: Unnecessary addition of whitespace.

dlongley (approved):

msporny (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2021-09-29-vcwg#section1-4) on 2021-09-29

- no resolutions were taken

#### 1.4. clarification of schema property (pr vc-data-model#816)

_See github pull request [#816](https://github.com/w3c/vc-data-model/pull/816)._




**Wayne Chang:** On issue 816, clarification of schema roperty, 3/4 approvals, manu requested changes, but seems like they've been addressed, is that correct?  

**Manu Sporny:** If that was done, we're good to go.  
… Yup, that's done, let me just approve, we're good.  

**Wayne Chang:** Ok, squashing and merging.  
… do we mean to merge into v1.1?  
… Can I change this so it merges to v1.1?  

**David Chadwick:** Yes, please.  

**Wayne Chang:** Seems like they're conflicts, we'll have to fix those.  
… Just README and diagrams are conflicting, this is more of a technical git issue, I can help w/ that.  
… Any other comments on this PR  

</details>

PR #801, opened 2021-08-30
update reference links to CCG and LDP Registries

kdenhartog:
closes #746

This updates the text to no longer point to DVCG work items and instead points to the active versions in the CCG where possible. In the case of LD-Signatures and LDS-RS2018 I've updated the text to instead point to the LDP-Registries which includes links of all of the registered LD Proofs that could be used with proofs.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/801.html" title="Last updated on Sep 9, 2021, 1:44 AM UTC (81d4ac2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/801/ea2a3ab...81d4ac2.html" title="Last updated on Sep 9, 2021, 1:44 AM UTC (81d4ac2)">Diff</a>

Issue #726, opened 2019-10-20
ZKP implementation - why proof and schema mandatory within derived VC?


Iso5786:
Hi,
if within IRMA (https://irma.app/docs/zkp/) at least two attributes from different credentials are disclosed, the computed ZKP (based on the Idemix specification) is one message which also proofs that the credentials belong to the same holder.

If within VC I want to compute a verifiable presentation with ZKP, it states that each derived VC must contain
"Proof, using the proof property, that can be used to derive verifiable presentations that present information contained in the original verifiable credential in zero-knowledge. The zero-knowledge verifiable presentation must not reveal any information not intended to be revealed by the holder."

But also 
"The verifiable presentation MUST contain a proof property to enable the verifier to ascertain that all derived verifiable credentials in the verifiable presentation were issued to the same holder without leaking personally identifiable information that the holder did not intend to share. "

So am not sure if this two statements are valid as within IRMA I just need one proof that will be attached to the presentation, but each derived VC does not necessarily need a proof property to make it work.

Also, I don't completely understand what needs to be contained within a credentialSchema as this is mandatory within a ZKP scenario. 
E.g. the issuers public key will be included in the resolved URI of the issuer so it would be good to maybe add an example of a credentialSchema within the ZKP section? The credential definition is defined via the type to my understanding.

Kind regards,
Daniel Ostkamp

brentzundel:
> If within VC I want to compute a verifiable presentation with ZKP, it states that each derived VC must contain
"Proof, using the proof property, that can be used to derive verifiable presentations that present information contained in the original verifiable credential in zero-knowledge. The zero-knowledge verifiable presentation must not reveal any information not intended to be revealed by the holder."

The proof within the derived credential is a zero-knowledge proof of signature. This proof must necessarily be included in each derived credential, as each credential may have been issued by a different issuer and according to a different schema.

> But also
"The verifiable presentation MUST contain a proof property to enable the verifier to ascertain that all derived verifiable credentials in the verifiable presentation were issued to the same holder without leaking personally identifiable information that the holder did not intend to share. "

The proof contained in the verifiable presentation proves that each of the derived credentials were issued to the same holder. Whether this is done using a link secret, or a holder seal, or simply by proving in zero knowledge that two semantically identical properties contain the same value, this proof belongs in the presentation because it connects all of the derived credentials.

[...]

PR #818, opened 2021-09-10
update normative statements in ZKP section

kdenhartog:
This PR is intended to be a first cut attempt to address the concerns raised in #726 in a way that generalizes the normative language in the ZKP section of the spec. I've attempted to keep all aspects of CL-Signatures system within the scope so that it's not going to exclude the usage of that system, while also trying to call out the more generic patterns of ZKP systems that are shared by BBS+-Sigs and CL-Sigs. 

@brentzundel definitely looking to get your eyes across this soon so that we can hopefully get this in within the allotted time of the charter.

Also worth calling out that because I referenced the LDP-REGISTRY here this PR won't reference the Biblio section properly until #801 is merged onto V1.1 and I can rebase those changes in here. I intentionally left the Biblio portion out and didn't build on #801 in order to keep this PR focused on the content for now. Once we get that other PR through the link should work properly.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/818.html" title="Last updated on Sep 29, 2021, 11:50 PM UTC (c750517)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/818/ea2a3ab...c750517.html" title="Last updated on Sep 29, 2021, 11:50 PM UTC (c750517)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2982,11 +2982,11 @@ how the data model can be used to issue, present, and verify zero-knowledge
         </p>
 
         <p>
-To use zero-knowledge <a>verifiable credentials</a>, the <a>issuer</a> must
-issue a <a>verifiable credential</a> in a manner that enables
-the <a>holder</a> to derive a proof from the originally issued <a>verifiable
-credential</a>, so that the <a>holder</a> can present the information to a
-<a>verifier</a> in a privacy-enhancing manner.
+For a <a>holder</a> to use a zero-knowledge <a>verifiable presentation</a>,
+the <a>issuer</a> must have issued a <a>verifiable credential</a> in a manner
+that enables the <a>holder</a> to derive a proof from the originally issued
+<a>verifiable credential</a>, so that the <a>holder</a> can present the
+information to a <a>verifier</a> in a privacy-enhancing manner.
 This implies that the <a>holder</a> can prove the validity of the
 <a>issuer's</a> signature without revealing the values that were signed, or when
 only revealing certain selected values. The standard practice is to do so by
--- a/index.html
+++ b/index.html
@@ -2985,8 +2985,8 @@ how the data model can be used to issue, present, and verify zero-knowledge
         <p>
 To use zero-knowledge <a>verifiable credentials</a>, the <a>issuer</a> must
 issue a <a>verifiable credential</a> in a manner that enables
-the <a>holder</a> to derive a proof from the original issued <a>verifiable
-credential</a> so that the <a>holder</a> can present the information to a
+the <a>holder</a> to derive a proof from the originally issued <a>verifiable
+credential</a>, so that the <a>holder</a> can present the information to a
 <a>verifier</a> in a privacy-enhancing manner.
 This implies that the <a>holder</a> can prove the validity of the
 <a>issuer's</a> signature without revealing the values that were signed, or when
--- a/index.html
+++ b/index.html
@@ -2983,7 +2983,7 @@ how the data model can be used to issue, present, and verify zero-knowledge
         </p>
 
         <p>
-To use zero-knowledge <a>verifiable credentials</a> the <a>issuer</a> must
+To use zero-knowledge <a>verifiable credentials</a>, the <a>issuer</a> must
 issue a <a>verifiable credential</a> in a manner that enables
 the <a>holder</a> to derive a proof from the original issued <a>verifiable
 credential</a> so that the <a>holder</a> can present the information to a
--- a/index.html
+++ b/index.html
@@ -2942,8 +2942,7 @@ information about the <code>proof</code> <a>property</a>, see Section
         <p>
 A zero-knowledge proof is a cryptographic method where an entity can prove to
 another entity that they know a certain value without disclosing the actual
-value. This is sometimes referred to as selective disclosure in attribute based
-credential systems. A real-world example is proving that an accredited university has
+value. A real-world example is proving that an accredited university has
 granted a degree to you without revealing your identity or any other personally
 identifiable information contained on the degree.
         </p>
@@ -2993,10 +2992,23 @@ This implies that the <a>holder</a> can prove the validity of the
 only revealing certain selected values. The standard practice is to do so by
 proving knowledge of the signature, without revealing the signature itself.
 There are two requirements for <a>verifiable credentials</a> when they are to be
-used in zero-knowledge proof systems. The <a>verifiable credential</a> MUST
-contain a Proof, using the <code>proof</code> <a>property</a>, so that the
-<a>holder</a> can derive a <a>verifiable presentation</a> that reveals only
-the information than they intend to reveal.
+used in zero-knowledge proof systems.
+        </p>
+
+        <ul>
+          <li>
+The <a>verifiable credential</a> MUST contain a Proof, using the
+<code>proof</code> <a>property</a>, so that the <a>holder</a> can derive a
+<a>verifiable presentation</a> that reveals only the information than the
+<a>holder</a> intends to reveal.
+          </li>
+          <li>
+If a <a>Credential</a> definition is being used, the <a>Credential</a>
+definition MUST be defined in the <code>credentialSchema</code> <a>property</a>,
+so that it can be used by all parties to perform various cryptographic
+operations in zero-knowledge.
+          </li>
+        </ul>
 
         <p>
 The following example shows one method of using <a>verifiable credentials</a> in
--- a/index.html
+++ b/index.html
@@ -3058,9 +3058,9 @@ expected to be met:
         <ul>
           <li>
 Each derived <a>verifiable credential</a> within a <a>verifiable
-presentation</a> MUST contain all necessary information to verify the
-<a>verifiable credential</a> either by including it directly within the
-credential or by referencing the necessary information.
+presentation</a> MUST contain all information necessary to verify the
+<a>verifiable credential</a>, either by including it directly within the
+credential, or by referencing the necessary information.
           </li>
           <li>
 A <a>verifiable presentation</a> MUST NOT leak information that would enable
--- a/index.html
+++ b/index.html
@@ -3003,7 +3003,7 @@ The <a>verifiable credential</a> MUST contain a Proof, using the
 <a>holder</a> intends to reveal.
           </li>
           <li>
-If a <a>Credential</a> definition is being used, the <a>Credential</a>
+If a <a>credential</a> definition is being used, the <a>credential</a>
 definition MUST be defined in the <code>credentialSchema</code> <a>property</a>,
 so that it can be used by all parties to perform various cryptographic
 operations in zero-knowledge.
--- a/index.html
+++ b/index.html
@@ -3004,9 +3004,9 @@ zero-knowledge. It makes use of a Camenisch-Lysyanskaya Signature
 [[?CL-SIGNATURES]], which allows the presentation of the <a>verifiable
 credential</a> in a way that supports the privacy of the
 <a>holder</a> and <a>subject</a> through the use of selective disclosure of the
-<a>verifiable credential</a> values. Some other crypto systems which rely upon
-the usage of zero-knowledge proofs in order to selective disclose attributes can
-be found in the [[?LDP-REGISTRY]] as well.
+<a>verifiable credential</a> values. Some other cryptographic systems which rely
+upon zero-knowledge proofs to selectively disclose attributes can be found in the
+[[?LDP-REGISTRY]] as well.
         </p>
 
         <pre class="example nohighlight" title="A verifiable credential that supports CL Signatures">
--- a/index.html
+++ b/index.html
@@ -2942,8 +2942,9 @@ information about the <code>proof</code> <a>property</a>, see Section
         <p>
 A zero-knowledge proof is a cryptographic method where an entity can prove to
 another entity that they know a certain value without disclosing the actual
-value. A real-world example is proving that an accredited university has granted
-a degree to you without revealing your identity or any other personally
+value. This is sometimes referred to as selective disclosure in attribute based
+credential systems. A real-world example is proving that an accredited university has
+granted a degree to you without revealing your identity or any other personally
 identifiable information contained on the degree.
         </p>
         <p>
@@ -2975,45 +2976,38 @@ their issued <a>verifiable credentials</a>.
         </ul>
 
         <p>
-This specification describes a data model that supports zero-knowledge proof
-mechanisms. The examples below highlight how the data model can be used to
-issue, present, and verify zero-knowledge <a>verifiable credentials</a>.
+This specification describes a data model that supports selective disclosure
+with the use of a zero-knowledge proof mechanisms. The examples below highlight
+how the data model can be used to issue, present, and verify zero-knowledge
+<a>verifiable credentials</a>.
         </p>
 
         <p>
 To use zero-knowledge <a>verifiable credentials</a> the <a>issuer</a> must
-issue a <a>verifiable credential</a> in a manner that enables the <a>holder</a>
-to present the information to a <a>verifier</a> in a privacy-enhancing manner.
+issue a <a>verifiable credential</a> in a manner that enables
+the <a>holder</a> to derive a proof from the original issued <a>verifiable
+credential</a> so that the <a>holder</a> can present the information to a
+<a>verifier</a> in a privacy-enhancing manner.
 This implies that the <a>holder</a> can prove the validity of the
 <a>issuer's</a> signature without revealing the values that were signed, or when
 only revealing certain selected values. The standard practice is to do so by
 proving knowledge of the signature, without revealing the signature itself.
 There are two requirements for <a>verifiable credentials</a> when they are to be
 used in zero-knowledge proof systems. The <a>verifiable credential</a> MUST
-contain a:
-        </p>
-
-        <ul>
-          <li>
-<a>Credential</a> definition, using the <code>credentialSchema</code>
-<a>property</a>, that can be used by all parties to perform various
-cryptographic operations in zero-knowledge.
-          </li>
-          <li>
-Proof, using the <code>proof</code> <a>property</a>, that can be used to derive
-<a>verifiable presentations</a> that present information contained in the
-original <a>verifiable credential</a> in zero-knowledge. The zero-knowledge
-<a>verifiable presentation</a> must not reveal any information not intended to
-be revealed by the <a>holder</a>.
-          </li>
-        </ul>
+contain a Proof, using the <code>proof</code> <a>property</a>, in the
+<a>verifiable credential</a> so that the <a>holder</a> can derive a
+<a>verifiable presentation</a> that reveals no more information than they intend
+to.
 
         <p>
 The following example shows one method of using <a>verifiable credentials</a> in
-zero-knowledge. It makes use of a CL Signature, which allows the presentation of
-the <a>verifiable credential</a> in a way that supports the privacy of the
+zero-knowledge. It makes use of a Camenisch-Lysyanskaya Signature
+[[?CL-SIGNATURES]], which allows the presentation of the <a>verifiable
+credential</a> in a way that supports the privacy of the
 <a>holder</a> and <a>subject</a> through the use of selective disclosure of the
-<a>verifiable credential</a> values.
+<a>verifiable credential</a> values. Some other crypto systems which rely upon
+the usage of zero-knowledge proofs in order to selective disclose attributes can
+be found in the [[?LDP-REGISTRY]] as well.
         </p>
 
         <pre class="example nohighlight" title="A verifiable credential that supports CL Signatures">
@@ -3046,7 +3040,6 @@ the <a>verifiable credential</a> in a way that supports the privacy of the
   }</span>
 }
         </pre>
-
         <p>
 The example above provides the <a>verifiable credential</a> definition by using
 the <code>credentialSchema</code> <a>property</a> and a specific proof that is
@@ -3057,17 +3050,18 @@ usable in the Camenisch-Lysyanskaya Zero-Knowledge Proof system.
 The next example utilizes the <a>verifiable credential</a> above to generate a
 new derived <a>verifiable credential</a> with a privacy-preserving proof. The
 derived <a>verifiable credential</a> is then placed in a
-<a>verifiable presentation</a>, which further proves that the entire assertion
-is valid. There are three requirements of most <a>verifiable presentations</a>
-when they are to be used in zero-knowledge systems:
+<a>verifiable presentation</a>, so that the <a>verifiable credential</a>
+discloses only the intended <a>claims</a> and additional credential
+metadata that the <a>holder</a> intended. In order to do this all of the
+following requirements are expected be met:
         </p>
 
         <ul>
           <li>
-Each derived <a>verifiable credential</a> within a
-<a>verifiable presentation</a> MUST have a <code>credentialSchema</code>
-<a>property</a>. This allows the derived <a>verifiable credential</a> to
-reference the credential definition used to generate the derived proof.
+Each derived <a>verifiable credential</a> within a <a>verifiable
+presentation</a> MUST contain all necessary information to verify the
+<a>verifiable credential</a> either by including it directly within the
+credential or by referencing the necessary information.
           </li>
           <li>
 A <a>verifiable presentation</a> MUST NOT leak information that would enable
@@ -3075,8 +3069,8 @@ the <a>verifier</a> to correlate the <a>holder</a> across multiple
 <a>verifiable presentations</a>.
           </li>
           <li>
-The <a>verifiable presentation</a> MUST contain a <code>proof</code>
-<a>property</a> to enable the <a>verifier</a> to ascertain that all derived
+The <a>verifiable presentation</a> SHOULD contain a <code>proof</code>
+<a>property</a> to enable the <a>verifier</a> to check that all derived
 <a>verifiable credentials</a> in the <a>verifiable presentation</a> were issued
 to the same <a>holder</a> without leaking personally identifiable information
 that the <a>holder</a> did not intend to share.
--- a/index.html
+++ b/index.html
@@ -2994,10 +2994,9 @@ only revealing certain selected values. The standard practice is to do so by
 proving knowledge of the signature, without revealing the signature itself.
 There are two requirements for <a>verifiable credentials</a> when they are to be
 used in zero-knowledge proof systems. The <a>verifiable credential</a> MUST
-contain a Proof, using the <code>proof</code> <a>property</a>, in the
-<a>verifiable credential</a> so that the <a>holder</a> can derive a
-<a>verifiable presentation</a> that reveals no more information than they intend
-to.
+contain a Proof, using the <code>proof</code> <a>property</a>, so that the
+<a>holder</a> can derive a <a>verifiable presentation</a> that reveals only
+the information than they intend to reveal.
 
         <p>
 The following example shows one method of using <a>verifiable credentials</a> in
--- a/index.html
+++ b/index.html
@@ -3050,9 +3050,9 @@ The next example utilizes the <a>verifiable credential</a> above to generate a
 new derived <a>verifiable credential</a> with a privacy-preserving proof. The
 derived <a>verifiable credential</a> is then placed in a
 <a>verifiable presentation</a>, so that the <a>verifiable credential</a>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2975,45 +2976,38 @@ <h3>Zero-Knowledge Proofs</h3>
         </ul>
 
         <p>
-This specification describes a data model that supports zero-knowledge proof
-mechanisms. The examples below highlight how the data model can be used to
-issue, present, and verify zero-knowledge <a>verifiable credentials</a>.
+This specification describes a data model that supports selective disclosure
+with the use of a zero-knowledge proof mechanisms. The examples below highlight
```
TallTed, line 2980:
```suggestion
with the use of zero-knowledge proof mechanisms. The examples below highlight
```

index.html
```diff
@@ -2975,45 +2976,38 @@ <h3>Zero-Knowledge Proofs</h3>
...
+how the data model can be used to issue, present, and verify zero-knowledge
+<a>verifiable credentials</a>.
         </p>
 
         <p>
 To use zero-knowledge <a>verifiable credentials</a> the <a>issuer</a> must
```
TallTed, line 2986:
```suggestion
To use zero-knowledge <a>verifiable credentials</a>, the <a>issuer</a> must
```

index.html
```diff
@@ -2975,45 +2976,38 @@ <h3>Zero-Knowledge Proofs</h3>
...
-issue a <a>verifiable credential</a> in a manner that enables the <a>holder</a>
-to present the information to a <a>verifier</a> in a privacy-enhancing manner.
+issue a <a>verifiable credential</a> in a manner that enables
+the <a>holder</a> to derive a proof from the original issued <a>verifiable
+credential</a> so that the <a>holder</a> can present the information to a
```
TallTed, line 2989:
```suggestion
the <a>holder</a> to derive a proof from the originally issued <a>verifiable
credential</a>, so that the <a>holder</a> can present the information to a
```

index.html
```diff
@@ -2975,45 +2976,38 @@ <h3>Zero-Knowledge Proofs</h3>
...
-<a>verifiable presentations</a> that present information contained in the
-original <a>verifiable credential</a> in zero-knowledge. The zero-knowledge
-<a>verifiable presentation</a> must not reveal any information not intended to
-be revealed by the <a>holder</a>.
-          </li>
-        </ul>
+contain a Proof, using the <code>proof</code> <a>property</a>, in the
+<a>verifiable credential</a> so that the <a>holder</a> can derive a
+<a>verifiable presentation</a> that reveals no more information than they intend
+to.
```
TallTed, line 3000: The last two words of this, "to reveal", could be deleted, but that phrasing might confuse more people.
```suggestion
contain a Proof, using the <code>proof</code> <a>property</a>, so that the
<a>holder</a> can derive a <a>verifiable presentation</a> that reveals only
the information than they intend to reveal.
```

index.html
```diff
@@ -2975,45 +2976,38 @@ <h3>Zero-Knowledge Proofs</h3>
...
-zero-knowledge. It makes use of a CL Signature, which allows the presentation of
-the <a>verifiable credential</a> in a way that supports the privacy of the
+zero-knowledge. It makes use of a Camenisch-Lysyanskaya Signature
+[[?CL-SIGNATURES]], which allows the presentation of the <a>verifiable
+credential</a> in a way that supports the privacy of the
 <a>holder</a> and <a>subject</a> through the use of selective disclosure of the
-<a>verifiable credential</a> values.
+<a>verifiable credential</a> values. Some other crypto systems which rely upon
+the usage of zero-knowledge proofs in order to selective disclose attributes can
+be found in the [[?LDP-REGISTRY]] as well.
```
TallTed, line 3010:
```suggestion
<a>verifiable credential</a> values. Some other cryptographic systems which rely
upon zero-knowledge proofs to selectively disclose attributes can be found in the
[[?LDP-REGISTRY]] as well.
```

index.html
```diff
@@ -3057,26 +3050,27 @@ <h3>Zero-Knowledge Proofs</h3>
 The next example utilizes the <a>verifiable credential</a> above to generate a
 new derived <a>verifiable credential</a> with a privacy-preserving proof. The
 derived <a>verifiable credential</a> is then placed in a
-<a>verifiable presentation</a>, which further proves that the entire assertion
-is valid. There are three requirements of most <a>verifiable presentations</a>
-when they are to be used in zero-knowledge systems:
+<a>verifiable presentation</a>, so that the <a>verifiable credential</a>
+discloses only the intended <a>claims</a> and additional credential
+metadata that the <a>holder</a> intended. In order to do this all of the
+following requirements are expected be met:
```
TallTed, line 3056:
```suggestion
discloses only the <a>claims</a> and additional credential metadata that the
<a>holder</a> intended. To do this, all of the following requirements are
expected to be met:
```

index.html
```diff
@@ -3057,26 +3050,27 @@ <h3>Zero-Knowledge Proofs</h3>
...
         <ul>
           <li>
-Each derived <a>verifiable credential</a> within a
-<a>verifiable presentation</a> MUST have a <code>credentialSchema</code>
-<a>property</a>. This allows the derived <a>verifiable credential</a> to
-reference the credential definition used to generate the derived proof.
+Each derived <a>verifiable credential</a> within a <a>verifiable
+presentation</a> MUST contain all necessary information to verify the
+<a>verifiable credential</a> either by including it directly within the
+credential or by referencing the necessary information.
```
TallTed, line 3064:
```suggestion
presentation</a> MUST contain all information necessary to verify the
<a>verifiable credential</a>, either by including it directly within the
credential, or by referencing the necessary information.
```

index.html
```diff
@@ -3057,26 +3050,27 @@ <h3>Zero-Knowledge Proofs</h3>
...
 the <a>verifier</a> to correlate the <a>holder</a> across multiple
 <a>verifiable presentations</a>.
           </li>
           <li>
-The <a>verifiable presentation</a> MUST contain a <code>proof</code>
-<a>property</a> to enable the <a>verifier</a> to ascertain that all derived
+The <a>verifiable presentation</a> SHOULD contain a <code>proof</code>
+<a>property</a> to enable the <a>verifier</a> to check that all derived
 <a>verifiable credentials</a> in the <a>verifiable presentation</a> were issued
 to the same <a>holder</a> without leaking personally identifiable information
```
TallTed, line 3075:
```suggestion
<a>verifiable credentials</a> in the <a>verifiable presentation</a> were issued
to the same <a>holder</a> without leaking personally identifiable information
```
Somehow I missed this in previous reviews.

* Why do the `derived <a>verifiable credentials</a>` have to have been `issued
to the same <a>holder</a>`?
* Are `derived <a>verifiable credentials</a>` issued at all?
* Why are we checking whether the current holder was the original holder of any  `<a>verifiable credentials</a>`?

index.html
```diff
@@ -2942,8 +2942,9 @@ <h3>Zero-Knowledge Proofs</h3>
         <p>
 A zero-knowledge proof is a cryptographic method where an entity can prove to
 another entity that they know a certain value without disclosing the actual
-value. A real-world example is proving that an accredited university has granted
-a degree to you without revealing your identity or any other personally
+value. This is sometimes referred to as selective disclosure in attribute based
+credential systems. A real-world example is proving that an accredited university has
```
dlongley, line 2946: Hmm, this reordering now reads to me as equating selective disclosure with zero knowledge proofs. A reader could be confused into think they are the same; most critically, selective disclosure involves sharing a particular value (and nothing else) as opposed to a ZKP which involves proving one knows a value without sharing anything else (including not sharing the value itself). Selective disclosure protocols may involve ZKPs, of course, but I think the language here needs some tweaking.

agropper:
This confusion is related to the introduction of "holder" as a normative actor in the VC data model. https://github.com/decentralized-identity/waci-presentation-exchange/discussions/96

P1 - A verifiable credential can stand alone separately from its presentation and can participate in ZKPs regardless who "holds" it. 

P2 - Presentation (with or without ZKP) by a particular holder, be they the issuer, subject, or delegate is a matter of identity that is completely separable from the VC data model itself. The boundary between the VC and its derivatives is authorization. 

P3 - Introducing "holder" as a normative element of the VC data model brings with it certain privacy risks such as issuers and verifiers requiring "certified" holders that are therefore not self-sovereign to the VC subject. 

P1-P3 are an attempt at a relatively simple framing of what could be confusing about VCs. Are these statements wrong or controversial? 

See also:
https://lists.w3.org/Archives/Public/public-credentials/2021Sep/0008.html 
https://lists.w3.org/Archives/Public/public-credentials/2021Aug/0001.html

TallTed:
@agropper -- This is the Open World of the Web. Anyone (Issuer) can say (Issue) anything (Claims within Credentials) about anything (Subject) including anyone. The *Subject* has *NO* authority over any Issuer nor Verifier, and, in fact, has no role as an actor of any kind in the ecosystems of the Web or VCs.

jandrieu:
> This confusion is related to the introduction of "holder" as a normative actor in the VC data model. [decentralized-identity/waci-presentation-exchange#96](https://github.com/decentralized-identity/waci-presentation-exchange/discussions/96)
> 
> P1 - A verifiable credential can stand alone separately from its presentation and can participate in ZKPs regardless who "holds" it.
> 
> P2 - Presentation (with or without ZKP) by a particular holder, be they the issuer, subject, or delegate is a matter of identity that is completely separable from the VC data model itself. The boundary between the VC and its derivatives is authorization.
> 
> P3 - Introducing "holder" as a normative element of the VC data model brings with it certain privacy risks such as issuers and verifiers requiring "certified" holders that are therefore not self-sovereign to the VC subject.
> 
> P1-P3 are an attempt at a relatively simple framing of what could be confusing about VCs. Are these statements wrong or controversial?

@agropper I'm afraid this is a continued area where terminology is preventing us from having a reasonable conversation.

The holder is *by definition* the party that physically holds and later presents a credential. Therefore, P1 is not a thing. A credential is ALWAYS presented by a holder. That holder may or may not be the subject, but they are ALWAYS the holder.

Second, VCs do not have derivatives. That isn't a thing. There are some notions of chains of VCs, but those are not part of the VC architecture. You can ALWAYS make a statement about another statement "Joe says that what Adrian said is brilliant" and those types of semantics could be thought of as derivative if that's what you mean, but they really are just two VCs: two statements. One the initial utterance, the second the commentary on the first.

So, P2 is also not valid. Presentation is always by a holder. And delegation doesn't apply precisely because non-holders cannot present. Non-subjects can, and when they do, they are acting as the holder.

Finally, "holder" is not being introduced. It is foundational to the work and its been there since well before it entered Recommendation status. The Holder is a primary role in the system, BY DEFINITION, and attempting to reject that notion rather than work within it is why it is so hard for many of us to work with you to address the meritorious parts of your concerns.

I'm not sure how else to get these points across. I believe we are talking past each other, as I have explained this multiple times.

I believe any attempt to remove Holder at this stage would require such a restructuring of the VC model as to make it an unworkable adjustment for the maintenance working group. Further, you seem to be the only one who perceives that the mere definition of that role creates privacy problems. It doesn't. It clarifies who is doing what at which stage in the VC lifecycle. Now, what holders, verifiers, and issuers are allowed to do can absolutely cause privacy problems, but the definition itself does not.

We need to find a way to get you speaking the same language as the rest of the community or we're going to continue running this thread that only you seem to understand. Repeating conversations because of an unwillingness to adopt the common language is not a recipe for collaborative development.

index.html
brentzundel: Indy anoncreds zkp-credentials, which were used as the exemplar for much of the material in this section, require that issued credentials be bound to a holder. One reason for this is that anoncreds aren't bound to a subject, but are rather issued to the subject, so binding to the holder is a way to accomplish connection of the credential to the subject. Another is to give verifiers more confidence that credentials are linked even when the disclosed claims are kept to a minimum. Binding to a holder also makes sharing credentials more detectable and unwieldy, which is seen as valuable in the anoncreds ecosystem.

That said, there are more zkp-credential schemes now than there were when the section was written, and not all of them require holder binding, even though many still consider that best practice. This is why the MUST is being changed to a SHOULD.

index.html
```diff
@@ -2975,45 +2976,38 @@ <h3>Zero-Knowledge Proofs</h3>
...
 There are two requirements for <a>verifiable credentials</a> when they are to be
 used in zero-knowledge proof systems. The <a>verifiable credential</a> MUST
-contain a:
-        </p>
-
-        <ul>
-          <li>
-<a>Credential</a> definition, using the <code>credentialSchema</code>
-<a>property</a>, that can be used by all parties to perform various
-cryptographic operations in zero-knowledge.
```
brentzundel, line 3000: instead of getting rid of this, I would prefer it changed from MUST to MAY

agropper:
@joeandrieu The reason I may not be “speaking the same language as the rest
of this community” may be that I have a unique role as invited expert on
privacy. As such, I have been an active participant for about as long as
anyone and I have tried to implement the standards to the extent I have
access supported libraries. Unlike most of the folks in our community, I am
working and supporting the implementation out of my own pocket.

I am also a strong supporter of the VC and DID work and am trying to
understand it and feel comfortable explaining it to skeptics in a
convincing way. Thank you for bearing with me.

more inline….

On Fri, Sep 10, 2021 at 3:46 PM Joe Andrieu ***@***.***>
wrote:

> This confusion is related to the introduction of "holder" as a normative
> actor in the VC data model.
> decentralized-identity/waci-presentation-exchange#96
> <https://github.com/decentralized-identity/waci-presentation-exchange/discussions/96>
>
> P1 - A verifiable credential can stand alone separately from its
> presentation and can participate in ZKPs regardless who "holds" it.
>
> P2 - Presentation (with or without ZKP) by a particular holder, be they
> the issuer, subject, or delegate is a matter of identity that is completely
> separable from the VC data model itself. The boundary between the VC and
> its derivatives is authorization.
>
> P3 - Introducing "holder" as a normative element of the VC data model
> brings with it certain privacy risks such as issuers and verifiers
> requiring "certified" holders that are therefore not self-sovereign to the
> VC subject.
>
> P1-P3 are an attempt at a relatively simple framing of what could be
> confusing about VCs. Are these statements wrong or controversial?
>
> @agropper <https://github.com/agropper> I'm afraid this is a continued
> area where terminology is preventing us from having a reasonable
> conversation.
>
I’m doing my best to be reasonable.

> The holder is *by definition* the party that physically holds and later
> presents a credential. Therefore, P1 is not a thing. A credential is ALWAYS
> presented by a holder. That holder may or may not be the subject, but they
> are ALWAYS the holder.
>
By this definition, the Issuer is always a holder as well.

Please reword P1 in any way you would prefer.

> Second, VCs do not have derivatives. That isn't a thing. There are some
> notions of chains of VCs, but those are not part of the VC architecture.
> You can ALWAYS make a statement about another statement "Joe says that what
> Adrian said is brilliant" and those types of semantics could be thought of
> as derivative if that's what you mean, but they really are just two VCs:
> two statements. One the initial utterance, the second the commentary on the
> first.
>
> I’m not hung up on the “derivative” issue. I agree with your statement
above.

So, P2 is also not valid. Presentation is always by a holder. And
> delegation doesn't apply precisely because non-holders cannot present.
> Non-subjects can, and when they do, they are acting as the holder.
>
Again, as per P2, I see the issuer as a potential holder just like the
subject or their delegate. If you can reword the intent of P2 in a way you
agree with, it might help our conversation.

> Finally, "holder" is not being introduced. It is foundational to the work
> and its been there since well before it entered Recommendation status. The
> Holder is a primary role in the system, BY DEFINITION, and attempting to
> reject that notion rather than work within it is why it is so hard for many
> of us to work with you to address the meritorious parts of your concerns.
>
I’m trying my best. I just don’t understand why my drivers license as a
verifiable credential can exist without the concept of “holder”. What am I
missing when the license is changed from a physical tamper-resistant card
to a tamper-resistant digital structure?

> I'm not sure how else to get these points across. I believe we are talking
> past each other, as I have explained this multiple times.
>
> I believe any attempt to remove Holder at this stage would require such a
> restructuring of the VC model as to make it an unworkable adjustment for
> the maintenance working group. Further, you seem to be the only one who
> perceives that the mere definition of that role creates privacy problems.
> It doesn't. It clarifies who is doing what at which stage in the VC
> lifecycle. Now, what holders, verifiers, and issuers are allowed to do can
> absolutely cause privacy problems, but the definition itself does not.
>
Please, help me understand the “holder” relationship in terms that I might
relate to a skeptic. The claims in a VC are bound to the identity of a
subject. If / when a holder is introduced in a normative way, it probably
relates to the protocols for interacting with VCs rather than the VC data
model itself.

> We need to find a way to get you speaking the same language as the rest of
> the community or we're going to continue running this thread that only you
> seem to understand. Repeating conversations because of an unwillingness to
> adopt the common language is not a recipe for collaborative development.
>
I’ve not introduced new language. I’m just trying to keep separate the data
model concerns from the role and protocol concerns. Why is that causing a
problem?

index.html
kdenhartog: I'm good with that, will update it.

index.html
kdenhartog: Yeah I agree with your reading of it. I wanted to highlight that ZKPs are likely being used for selective disclosure which is what led to this language update. However, your point stands that ZKPs and selective disclosure mechanisms are not the same thing, although they often can be conflated. I'll give an attempt at rewording this in order to make sure this point isn't lost on the reader.

kdenhartog:
> 
> 
> This confusion is related to the introduction of "holder" as a normative actor in the VC data model. [decentralized-identity/waci-presentation-exchange#96](https://github.com/decentralized-identity/waci-presentation-exchange/discussions/96)
> 
> P1 - A verifiable credential can stand alone separately from its presentation and can participate in ZKPs regardless who "holds" it.
> 
> P2 - Presentation (with or without ZKP) by a particular holder, be they the issuer, subject, or delegate is a matter of identity that is completely separable from the VC data model itself. The boundary between the VC and its derivatives is authorization.
> 
> P3 - Introducing "holder" as a normative element of the VC data model brings with it certain privacy risks such as issuers and verifiers requiring "certified" holders that are therefore not self-sovereign to the VC subject.
> 
> P1-P3 are an attempt at a relatively simple framing of what could be confusing about VCs. Are these statements wrong or controversial?
> 
> See also:
> https://lists.w3.org/Archives/Public/public-credentials/2021Sep/0008.html
> https://lists.w3.org/Archives/Public/public-credentials/2021Aug/0001.html

I don't see them as wrong for the most part (there's some edge cases I'm sure where they don't hold), but not sure how they affect the language written in the PR. Based on my reading, the language goes to further enhance these points not reduce it. What are you suggesting in terms of updates to the PR here?

index.html
kdenhartog: updated in latest commit. Resolving this comment.

index.html
kdenhartog: I just dropped the additional sentence I had originally added, so that this reads the same as it used to. I figured this would be the easier solution to address your concern. If you have a suggestion to further enhance and untie these concepts in a second round of review I'm happy to add that in as well.

index.html
kdenhartog: This suggestion appears to be the same as the current text in there. I presume this review comment was more about raising the questions which I think @brentzundel addressed. Please let me know if there's any additional changes that I missed to resolve this comment.

TallTed: > this review comment was more about raising the questions

Correct. However, I do not think that the answers provided by @brentzundel remove my concerns with this section.

I am mostly concerned about the "bound"/"binding" terminology in use here, which I think is a significantly stronger relationship than is actually present between these VCs and their Subjects *or* Holders. I think it would be better to say that xyz property/ies in the Credentials under specific discussion are (or SHOULD be or MUST be) set to strings or URIs which identify the Holder who is expected to Present the VC to any Verifiers downstream. Then, the Verifier is expected to check that _**either**_ some other qrs property/ies were (or SHOULD have been or MUST have been) set to strings or URIs which identify the Subject, ***OR*** the Subject is always the same as the Holder identified by the values of xyz property/ies, ... Or something along those lines.

I think that "anoncreds aren't bound to a subject, but are rather issued to the subject, so binding to the holder is a way to accomplish connection of the credential to the subject" translates roughly to "anoncreds are Issued to their Subject, which is defined as a sapient being, and which is not identified as the Subject within the VC. Identifying the Holder is an indirect way to accomplish association of the VC with its Subject," which Subject is not really obscured by this method, since the identified Holder is always the Subject and anyone who can access the value of the field identifying the Holder is thereby able to access to identifier of the Subject. Leading to another concern with this section, since it seems that the Subject anonymity desired to be delivered through ZKP, at least in part by identifying one Holder who must be the Subject, instead of identifying the Subject, isn't actually delivered. 

I'm sure I'm missing something exceedingly clever, and I hope someone can explain it in terms I can understand and that we can swiftly adopt (and/or edit slighly to adopt) for this section.

dlongley: > I am mostly concerned about the "bound"/"binding" terminology in use here, which I think is a significantly stronger relationship than is actually present between these VCs and their Subjects or Holders. I think it would be better to say that xyz property/ies in the Credentials under specific discussion are (or SHOULD be or MUST be) set to strings or URIs which identify the Holder who is expected to Present the VC to any Verifiers downstream.

+1. And if there are cases where there aren't actually any properties in the VC itself with this information because some mechanisms include it via secrets and/or proof data, then that can be stated as an option as well. I think it's important that we indicate that we're talking about setting expectations -- and not assert a level of control that isn't possible. It's also fine to say that there are mechanisms that attempt to impede adversaries that try to subvert these expectations, but such statements should include a friendly warning that they all have limitations that need to be well understood by implementers (and readers should consult external documentation for that).

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2021-09-15-vcwg#section3-2) on 2021-09-15

- no resolutions were taken

#### 3.2. update normative statements in ZKP section (pr vc-data-model#818)

_See github pull request [#818](https://github.com/w3c/vc-data-model/pull/818)._




**Brent Zundel:** PR 818. Update normative statements in the ZKP section. A little more extensive of a PR.  
… Kyle, since you're here, can you walk us through the changes  

**Kyle Den Hartog:** Just trying to update things as best as possible. Particularly in the normative statements, but also to go around addressing smaller fixes to make it more readable  
… to make the ZKP section be able to fit CL signatures, particularly with Anon Credentials  
… but also to allow the BBS+ suite.  
… The main things are the list, basically removing the requirement to use a credentialSchema - and also tweaking the possibility to generate a proof...  
… It says it must contain a proof property... obvious. The additional constraint is so the holder can derive credentials to reveal only what they want to reveal.  
… So I was trying to convey the ZKP must have some ... property  
… The ZKP layer, mainly around the guarantees of the credentialSchema, making sure not to change guarantees around leaking additional information.  
… Also in the proof, the ability to link things to the credentialSchema.  
… Sorry, that was tough...  

**Brent Zundel:** Thank you, we know it is late/early where you are  
… Ted has raised concern about some language around holder binding. I think these are valid concerns and we should have conversations around it.  
… I don't know if this place is the right place to have those conversations.  

> *Dave Longley:* I don't think this PR uses that language (would have to double check)

**Brent Zundel:** The comments were around things in the spec that this PR isn't changing.  
… I wouldn't want to hold up this PR, but would like to work on addressing this concern.  
… What does the group think?  

**Ted Thibodeau Jr.:** I'm fine with spinning it out into a different issue - I just raised the comment there because that's when I noticed it - don't know how it slipped by in past readings.  

**Brent Zundel:** I think this is an important conversation to have, ties in with other issues.  

**Kyle Den Hartog:** To add to that, I think this is an indication that ... broader changes, decoupling the subject/holder binding issue... better to address separately.  

**Brent Zundel:** Ted, if you would open a new issue...  

**Ted Thibodeau Jr.:** I'll spin one up now.  

</details>

index.html
TallTed: This thread continues in #821.

kdenhartog: Do we want to action this within this PR - if so can you propose some suggested changes based on the discussions so far.

kdenhartog:
If people are satisfied with the recent changes can we get some re-approvals on here? If not, please leave some suggestions and I'll plan on getting this updated before the call this week.

index.html
```diff
@@ -2975,45 +2975,50 @@ <h3>Zero-Knowledge Proofs</h3>
...
 
         <p>
-To use zero-knowledge <a>verifiable credentials</a> the <a>issuer</a> must
-issue a <a>verifiable credential</a> in a manner that enables the <a>holder</a>
-to present the information to a <a>verifier</a> in a privacy-enhancing manner.
+To use zero-knowledge <a>verifiable credentials</a>, the <a>issuer</a> must
+issue a <a>verifiable credential</a> in a manner that enables
+the <a>holder</a> to derive a proof from the originally issued <a>verifiable
+credential</a>, so that the <a>holder</a> can present the information to a
+<a>verifier</a> in a privacy-enhancing manner.
```
TallTed, line 2989:
```suggestion
For a <a>holder</a> to use a zero-knowledge <a>verifiable presentation</a>,
the <a>issuer</a> must have issued a <a>verifiable credential</a> in a manner
that enables the <a>holder</a> to derive a proof from the originally issued
<a>verifiable credential</a>, so that the <a>holder</a> can present the
information to a <a>verifier</a> in a privacy-enhancing manner.
```

index.html
TallTed: @kdenhartog -- There are no changes to make here in this PR based on this comment.

TallTed:
Modulo concerns in #821, the (I hope, minor to all eyes) change just requested, and anything I may have missed on re-re-re-re-reading this (#818), I think I'm OK with merging this.

index.html
```diff
@@ -2975,45 +2975,50 @@ <h3>Zero-Knowledge Proofs</h3>
...
+how the data model can be used to issue, present, and verify zero-knowledge
+<a>verifiable credentials</a>.
         </p>
 
         <p>
-To use zero-knowledge <a>verifiable credentials</a> the <a>issuer</a> must
-issue a <a>verifiable credential</a> in a manner that enables the <a>holder</a>
-to present the information to a <a>verifier</a> in a privacy-enhancing manner.
+For a <a>holder</a> to use a zero-knowledge <a>verifiable presentation</a>,
+the <a>issuer</a> must have issued a <a>verifiable credential</a> in a manner
```
kdenhartog, line 2986:
```suggestion
the <a>issuer</a> MUST have issued a <a>verifiable credential</a> in a manner
```

kdenhartog: @msporny @brentzundel thoughts on this change? I'm pretty sure it's testable (test if a credential can be derived), but not sure if we want this in here.

brentzundel:
I think it would be best to use "needs to" or some other non normative word
choice,  rather than add apossibly untestable normative statement.

index.html
```diff
@@ -2975,45 +2975,50 @@ <h3>Zero-Knowledge Proofs</h3>
...
+how the data model can be used to issue, present, and verify zero-knowledge
+<a>verifiable credentials</a>.
         </p>
 
         <p>
-To use zero-knowledge <a>verifiable credentials</a> the <a>issuer</a> must
-issue a <a>verifiable credential</a> in a manner that enables the <a>holder</a>
-to present the information to a <a>verifier</a> in a privacy-enhancing manner.
+For a <a>holder</a> to use a zero-knowledge <a>verifiable presentation</a>,
+the <a>issuer</a> needs to have issued a <a>verifiable credential</a> in a manner
```
TallTed, line 2986:
```suggestion
they need an <a>issuer</a> to have issued a <a>verifiable credential</a> in a manner
```

index.html
```diff
@@ -2975,45 +2975,50 @@ <h3>Zero-Knowledge Proofs</h3>
...
+<a>verifiable presentation</a> that reveals only the information than the
+<a>holder</a> intends to reveal.
           </li>
           <li>
-Proof, using the <code>proof</code> <a>property</a>, that can be used to derive
-<a>verifiable presentations</a> that present information contained in the
-original <a>verifiable credential</a> in zero-knowledge. The zero-knowledge
-<a>verifiable presentation</a> must not reveal any information not intended to
-be revealed by the <a>holder</a>.
+If a <a>Credential</a> definition is being used, the <a>Credential</a>
```
msporny, line 3006:
```suggestion
If a <a>credential</a> definition is being used, the <a>credential</a>
```

msporny:
Normative and substantive, multiple reviews, changes requested and made, no objections, merging.

kdenhartog:
There's a linked issue with the Errata label, removing the Errata and V1.2 substantive labels from the PR

PR #819, opened 2021-09-14
Use URI id for credentialStatus and refreshService

### LINKED ISSUE #748: URIs or URLs for ids
The `id` property for `credentialStatus`, `refreshService`, and `evidence` objects are specified as URLs:

- "Each refreshService value MUST specify its type [...] and its id, which is the URL of the service"
- "The value of the credentialStatus property MUST include the [...] id property, which MUST be a URL."
- "The value of the evidence property MUST be one or more evidence schemes [...] The id property is optional, but if present, SHOULD contain a URL"

Other `id` properties are specified as URIs:
- (§4.2) "The value of the id property MUST be a single URI"
- "The value of the `@context` property MUST be an ordered set where the first item is a URI with the value https://www.w3.org/2018/credentials/v1"
- (credentialSubject) "Each object MAY contain an id, as described in Section § 4.2 Identifiers."
- "The value of the issuer property MUST be either a URI or an object containing an id property. It is RECOMMENDED that the URI in the issuer or its id [...]" 
- "If present, the value of the holder property is expected to be a URI for the entity that is generating the presentation."
- "Each credentialSchema MUST specify its type [...], and an id property that MUST be a URI"
- (presentations) "The id property is optional and MAY be used to provide a unique identifier for the presentation. For details related to the use of this property, see Section § 4.2 Identifiers."

Could the credential status id, refresh service ids, and evidence object ids be just URIs, like the other id properties? Or is it really meant that they should/must be URLs?

Also, should the `id` property of a `termsOfUse` object or of a credential `issuer` object be expected to be a URI?

Section § 4.2 Identifiers specifies the `id` property - but I am not sure if that is meant to apply to all uses of `id` properties, or only ones which specifically reference §4.2 or are described in §4.2 (like in Example 4).

clehner:
Fix #748. This changes the *MUST* requirement for [credentialStatus](https://www.w3.org/TR/vc-data-model/#status) id and
[refreshService](https://www.w3.org/TR/vc-data-model/#refreshing) id to be URI instead of URL. The purpose of this change is to
improve consistency with the rest of the data model, where `id`s are expected
to be URIs (but not necessarily URLs). This should help implementers who want to ensure conformance with the VC Data Model without feeling compelled to check whether or not certain values are URLs, which may be a difficult determination to make without further specification by extensions. For example, there the VC Test Suite doesn't check URL-ness of refresh service ids: https://github.com/w3c/vc-test-suite/blob/593d5a5fb83f89ba60f32c0443d01fced5b4a3db/test/vc-data-model-1.0/21-refresh.js#L59 . Also, this specification doesn't seem to define URL or directly reference [RFC 1738 - Uniform Resource Locators (URL)](https://datatracker.ietf.org/doc/html/rfc1738), although it does have a definition for URI and identifier, and cites [RFC 3986 - Uniform Resource Identifier (URI): Generic Syntax](https://datatracker.ietf.org/doc/html/rfc3986).

The other use of URL, in [evidence](https://www.w3.org/TR/vc-data-model/#evidence) `ids`, I have not changed in this PR, as that is a *SHOULD*, so I understand it would not be necessary to check for strict conformance. Maybe it is more useful to leave that part in to indicate the intended use of it?

As discussed on the [2021-09-08 VCWG call](https://github.com/w3c/vc-data-model/issues/748#issuecomment-915839459), these changes may be substantive, since it changes *MUST*s, but also might not be, if it is considered that checking whether it is a URL might not by programatically possible as currently specified.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/819.html" title="Last updated on Sep 30, 2021, 11:55 AM UTC (c52a913)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/819/41baf90...c52a913.html" title="Last updated on Sep 30, 2021, 11:55 AM UTC (c52a913)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1872,7 +1872,7 @@ include the following:
 
             <ul>
               <li>
-<code>id</code> <a>property</a>, which MUST be a URL.
+<code>id</code> <a>property</a>, which MUST be a URI.
               </li>
               <li>
 <code>type</code> <a>property</a>, which expresses the <a>credential</a> status
@@ -2659,7 +2659,7 @@ more refresh services that provides enough information to the recipient's
 software such that the recipient can refresh the <a>verifiable credential</a>.
 Each <code>refreshService</code> value MUST specify its <code>type</code> (for
 example, <code>ManualRefreshService2018</code>) and its <code>id</code>, which
-is the URL of the service. The precise content of each refresh service is
+is the URI of the service. The precise content of each refresh service is
 determined by the specific <code>refreshService</code> <a>type</a> definition.
           </dd>
         </dl>


dlongley (approved):

peacekeeper (approved):

brentzundel (approved):

dlongley:
Maybe we may want to use URI and add an additional statement(s) that there's an expectation that machine readable information needs to be retrievable either from the URI or via some other field value in the `credentialStatus`/`refreshService` sections. That may resolve the concern around us wanting to be clear that there's some "locator" requirements/expectations here.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2021-09-15-vcwg#section3-1) on 2021-09-15

- no resolutions were taken

#### 3.1. Use URI id for credentialStatus and refreshService (pr vc-data-model#819)

_See github pull request [#819](https://github.com/w3c/vc-data-model/pull/819)._




**David Chadwick:** This is a PR that uses a URI id for `credentialStatus` and refresh service. It has a couple approvals already.  
… This was in response to an issue that said... some say URI, some say URL. This takes the two instances that say must be URL, and change to say must be a URI.  
… Last time, we discussed, this might not actually be a breaking change?  

**Manu Sporny:** This is a normative breaking change. It doesn't matter if it's a subset: we are changing the normative... what's allowed.  
… Even if nobody cares, it's technically a 1.2, thing, and it's labeled as such.  

> *Ted Thibodeau Jr.:* normative, but should not be breaking, because it relaxes requirements. I think it does class as a typo bug.

**Brent Zundel:** Question for group: do we agree that this is a bug in the specification?  
… If it's not, it needs to be deferred to v2.  

**Manu Sporny:** I didn't get a change to review... I recall we limited it on purpose - but don't remember if it was these two things specifically.  
… I want to be careful about this change because I believe we said you want to be actually able to go out and discover this information.  
… I'd like a little more time to review it, to make sure we are not undoing something we wanted to do.  

> *David Chadwick:* @manu +1

> *David Chadwick:* I also remember that it was a URL so that it could be fetched

**Brent Zundel:** That is the primary question. Is this addressing errata. Is it substantive... because we can only make substantive changes in response to errata.  
… If we can't answer that, it will be deferred.  

**Kyle Den Hartog:** The only time I'm aware of this being a concern is with `credentialStatus` and Revocation Status 2020. There is a normative statement that references an id and says it needs to be linkable.  
… They are utilizing the VC Data Model... that could create a conflict... suggest taking a look at it.  

**Manu Sporny:** Now... my memory is being jogged... we absolutely meant URL. So I believe probably the fix is wrong and we shouldn't merge this.  
… But I would like more time to dive in and consider it.  

**Ted Thibodeau Jr.:** I'm fine with having the discussion in the GitHub pages... but having a URL does not guarantee it is dereferencable any more than URI...  
… URL abstracts by location... just a construct that makes it easier for humans to look at it... It's a difference without a difference.  
… Like I said, we can have the argument in the GitHub pages... but it will just go on forever if we need to.  

> *Dave Longley:* there would be no problem with revocation status type imposing additional constraints on top of the vc data model spec

> *Dave Longley:* (in response to kyle)

**Brent Zundel:** Reminder, the conversation minuted here will automatically go into the PR. I agree that conversation should continue there.  

> *Manu Sporny:* I agree with TallTed -- we probably need to say "you have to be able to get information from the URI" or something to that effect

**Brent Zundel:** We'll move to the next Pull Request.  

</details>

wyc:
Needs to be targeting the v1.2 branch instead of main.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2021-09-29-vcwg#section1-5) on 2021-09-29

- no resolutions were taken

#### 1.5. Use URI id for credentialStatus and refreshService (pr vc-data-model#819)

_See github pull request [#819](https://github.com/w3c/vc-data-model/pull/819)._




**Wayne Chang:** One more PR review... This one by cel.  
… this looks like it's good to go, needs to be retargeted to v1.2?  

**Manu Sporny:** yes  

**Charles Lehner:** yes  

**Wayne Chang:** Looks like there is no v1.2 branch, we'll resolve out of band.  

</details>

msporny (approved):

msporny:
Merge conflicts, please fix @clehner. We're trying to get this merged down by the end of the week.

kdenhartog:
cleaned up the merge conflicts on this.

Normative and substantive, multiple reviews, no objections, merging.

clehner:
Thanks @kdenhartog!

kdenhartog:

PR #822, opened 2021-09-15
address inconsistencies in date-time strings

### LINKED ISSUE #782: RFC 3339 vs ISO8601 vs XSD Date Time
IMO, the spec fails to define datetime's correctly.

consider:

> A credential MUST have an issuanceDate property. The value of the issuanceDate property MUST be a string value of an [RFC3339] combined date and time string representing the date and time the credential becomes valid, which could be a date and time in the future.

https://w3c.github.io/vc-data-model/#issuance-date

yet: https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v1#L44

RFC3339 !== XSD Date Time.

See the differences summarized [here](https://atom-syntax.imc.narkive.com/x8nvl4Bh/xsd-datetime-vs-rfc-3339) and [here](https://medium.com/easyread/understanding-about-rfc-3339-for-datetime-formatting-in-software-engineering-940aa5d5f68a)

Most notably:

>  RFC 3339 allows lower- and upper-case 'T' and 'Z', while
xsd:dateTime requires upper-case only. 

> RFC 3339 allows the replacement of the 'T' with a space which
xsd:dateTime does not. [Note: ISO 8601 doesn't either, hence RFC
3339 isn't actually a "profile of ISO 8601" as it claims].

> RFC 3339 does not allow the hour 24 but xsd:dateTime does (in
xsd:dateTime [as in ISO 8601] 00 & 24 both represent midnight, but
24 means the end of one day while 00 means the start of the next)

>  RFC 3339 gives a different semantic to a timezone offset of -00:00 from
+00:00 and Z which xsd:dateTime (and to the best of my knowledge
ISO 8601) doesn't.

In short, I believe it was an error to point to RFC 3339, and instead we should have referenced ISO 8601 directly...

> RFC 3339 is following the ISO 8601 DateTime format. The only difference is RFC allows us to replace “T” with “space”.

This means that VC that is normatively compliant can have an issuance date which is neither valid ISO 8601 nor valid XSD Date Time.

Most credentials I have observed in the wild use `ISO 8601`.

Additionally, https://w3c.github.io/vc-data-model/#jwt-encoding

> For backward compatibility with JWT processors, the following JWT-registered claim names MUST be used ***instead of, or in addition to, their respective standard verifiable credential counterparts***

> nbf MUST represent issuanceDate, encoded as a UNIX timestamp (NumericDate).

yet:

> A credential MUST have an issuanceDate property. 

https://w3c.github.io/vc-data-model/#issuance-date

This produces an obvious logical impossibility for normative implementations. 

(issuanceDate cannot be omitted as the JWT section claims), 

...however the JWT processing section is at least pointing to a datetime representation that is compatible with practical datetime reality... https://www.unixtimestamp.com/

Adding this up, put yourself in the shoes of a spec implementer... You have 2 extreme possibilities:

1. YOLO `issuanceDate` and treat it as an opaque string based on confusion summarized above.
2. Use ISO 8601, XSD Date Time and Unix Time Stamps instead of RFC 3339.

It might seem obvious that you would choose 2, but you will be forced to fail normatively valid credentials, such as these:

```jsonc
{
  "@context": [
    "https://w3id.org/security/suites/jws-2020/v1"
  ],
  "issuer": {
    "id": "did:key:z6MkhRZ7XMcydYWTYGQrVxKXaG9Bpd3LrbA345QGWNMaSxeE"
  },
  // note space is legal to RFC 3339 but not ISO 8601 or XSD Date Time
  "issuanceDate": "2021-07-11 13:32:11.063Z", 
  "credentialSubject": { "id": "did:example:ebfeb1f712ebc6f1c276e12ec21" },
  "proof": {
    "type": "JsonWebSignature2020",
    // legal ISO 8601 AND XSD Date Time
    "created": "2010-01-01T19:23:24Z",
    "verificationMethod": "did:key:z6MkhRZ7XMcydYWTYGQrVxKXaG9Bpd3LrbA345QGWNMaSxeE#z6MkhRZ7XMcydYWTYGQrVxKXaG9Bpd3LrbA345QGWNMaSxeE",
    "jws": "eyJhbGciOiJFZERTQSIsImI2NCI6ZmFsc2UsImNyaXQiOlsiYjY0Il19..WGofUrPdHhJG2rxoVDLDynOHcZaqS2g5PJtCqOsvQ4ePKl_so1jZMytBX3sbkbbrzqG03Nhpe_zgymVfDaX6Bw"
  }
}
```

Consider the equivalent VC-JWT representation that happens to also be valid JSON-LD, you are now forced to convey the issuance date as being "XSD Date Time" in JSON-LD, yet its string representation is RFC 3339 and its additional unix timestamp is equivalent to all 3 (legally) even though its JSON-LD term definition is contradicting the normatively term definition.

Again, the problem is referring to RFC 3339.... 

ISO 8601 + XSD Date Time + Unix Timestamps are all happy friends.... RFC 3339 is a mistake.

kdenhartog:
This closes #782 by changing the normative reference of how date-time strings are expected to represented

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/822.html" title="Last updated on Sep 30, 2021, 12:08 PM UTC (0db4216)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/822/ea2a3ab...0db4216.html" title="Last updated on Sep 30, 2021, 12:08 PM UTC (0db4216)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1648,11 +1648,11 @@ expressing the date and time when a <a>credential</a> becomes valid.
           <dd>
 A <a>credential</a> MUST have an <code>issuanceDate</code> <a>property</a>. The
 value of the <code>issuanceDate</code> <a>property</a> MUST be a string value of
-an [[!RFC3339]] combined date and time string representing the date and time the
-<a>credential</a> becomes valid, which could be a date and time in the future.
-Note that this value represents the earliest point in time at which the
-information associated with the <code>credentialSubject</code> <a>property</a>
-becomes valid.
+an [[!XMLSCHEMA11-2]] combined <code>date-time</code> string representing the
+date and time the <a>credential</a> becomes valid, which could be a date and
+time in the future. Note that this value represents the earliest point in time
+at which the information associated with the <code>credentialSubject</code>
+<a>property</a> becomes valid.
           </dd>
         </dl>
 
@@ -1682,9 +1682,9 @@ It is expected that the next version of this specification will add the
 <code>validFrom</code> <a>property</a> and will deprecate the
 <code>issuanceDate</code> <a>property</a> in favor of a new <code>issued</code>
 <a>property</a>. The range of values for both properties are expected to remain
-as [[?RFC3339]] combined date and time strings. Implementers are advised that
-the <code>validFrom</code> and <code>issued</code> <a>properties</a> are
-reserved and use for any other purpose is discouraged.
+as [[?XMLSCHEMA11-2]] combined <code>date-time</code> strings. Implementers are
+advised that the <code>validFrom</code> and <code>issued</code>
+<a>properties</a> are reserved and use for any other purpose is discouraged.
         </p>
 
       </section>
@@ -1792,7 +1792,7 @@ the expression of <a>credential</a> expiration information.
           <dt><var>expirationDate</var></dt>
           <dd>
 If present, the value of the <code>expirationDate</code> <a>property</a> MUST be
-a string value of an [[!RFC3339]] combined date and time string representing the
+a string value of an [[!XMLSCHEMA11-2]] <code>date-time</code> representing the
 date and time the <a>credential</a> ceases to be valid.
           </dd>
         </dl>
@@ -3679,7 +3679,7 @@ done:
             <ul>
               <li>
 If <code>exp</code> is present, the UNIX timestamp MUST be converted to an
-[[!RFC3339]] <code>date-time</code>, and MUST be used to set the value
+[[!XMLSCHEMA11-2]] <code>date-time</code>, and MUST be used to set the value
 of the <code>expirationDate</code> <a>property</a> of
 <code>credentialSubject</code> of the new JSON object.
               </li>
@@ -3690,7 +3690,7 @@ If <code>iss</code> is present, the value MUST be used to set the
                   </li>
                   <li>
 If <code>nbf</code> is present, the UNIX timestamp MUST be converted to an
-[[!RFC3339]] <code>date-time</code>, and MUST be used to set the value
+[[!XMLSCHEMA11-2]] <code>date-time</code>, and MUST be used to set the value
 of the <code>issuanceDate</code> <a>property</a> of the new JSON object.
                   </li>
                   <li>
--- a/index.html
+++ b/index.html
@@ -1648,9 +1648,10 @@ expressing the date and time when a <a>credential</a> becomes valid.
           <dd>
 A <a>credential</a> MUST have an <code>issuanceDate</code> <a>property</a>. The
 value of the <code>issuanceDate</code> <a>property</a> MUST be a string value of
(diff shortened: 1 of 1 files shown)


OR13 (approved):
This is good, but if its possible to point directly to http://www.w3.org/TR/xmlschema11-2/#dateTime we probably should... not sure if the vc data model can use cite tags... here is an example from the did imp guide.

https://github.com/w3c/did-imp-guide/blob/main/index.html#L138

```
 <a data-cite="DID-CORE#dfn-did-methods">DID method</a>
```

https://github.com/w3c/did-imp-guide/blob/main/index.html#L27
```
      xref: {
        specs: ["DID-CORE", "VC-DATA-MODEL", "DID-CBOR-REPRESENTATION", "DID-RUBRIC", "did-spec-registries", "RFC4086", "RFC4880", "RFC6838", "RFC8259", "RFC8446", "RFC8174", "RFC2119"],
        profile: "web-platform",
      },
```

kdenhartog:
I'll see what can be done to point specifically to that section. Good call out.

brentzundel (approved):

kdenhartog:
Still need to update this

wyc:
@kdenhartog did you mean to merge to v1.2 branch? This is labeled as a substantive change.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2021-09-29-vcwg#section1-3) on 2021-09-29

- no resolutions were taken

#### 1.3. address inconsistencies in date-time strings (pr vc-data-model#822)

_See github pull request [#822](https://github.com/w3c/vc-data-model/pull/822)._




> *Wayne Chang:* [https://github.com/w3c/vc-data-model/pull/822](https://github.com/w3c/vc-data-model/pull/822)

> *Manu Sporny:* I think this has received approval.

**David Chadwick:** Is this a substantive change to v1.2  

**Wayne Chang:** I'll ask Kyle.  

**Manu Sporny:** It is a substantive change.  
… It has to go to v1.2  

**Wayne Chang:** Ok, we don't need another review period, content is fine, just need to retarget to v1.2  

</details>

kdenhartog:
Yup - I need to create a V1.2 branch and re-point this at that new branch which I've carved out some time to do today in order to hit our deadlines.

msporny (approved):

msporny:
> Yup - I need to create a V1.2 branch and re-point this at that new branch which I've carved out some time to do today in order to hit our deadlines.

Just saw this, will let you take this PR from here @kdenhartog. I'm disengaging.

kdenhartog:
Normative and substantive, multiple reviews, changes requested and made, no objections, merging.

kdenhartog:
There's a linked issue with the Errata label, removing the Errata labels from the PR

PR #834, opened 2021-11-11
Add Verifiable Credential digital signature views for multiple examples.

msporny:
This PR adds the [respec-vc](https://github.com/digitalbazaar/respec-vc) extension to the specification and activates it for example #4. You can view the result here:

https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/834.html#example-usage-of-the-id-property

or see the image below:

image

Click the tabs to view the Credential, Verifiable Credential (with proof), and Verifiable Credential (as JWT).

Once this is merged, I can go through and update any other examples that would benefit from this feature. Once that's done, we can close issue #833 and PR #817.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/834.html" title="Last updated on Nov 21, 2021, 5:38 PM UTC (42392a6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/834/df99d22...42392a6.html" title="Last updated on Nov 21, 2021, 5:38 PM UTC (42392a6)">Diff</a>

PR #785, opened 2021-07-29
working on #777

TallTed:
tries to partially (diagram changes are also needed) fix #777, by uncoloring and adjusting the text

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/785.html" title="Last updated on Aug 11, 2021, 3:17 PM UTC (bd8092d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/785/b1ea54d...bd8092d.html" title="Last updated on Aug 11, 2021, 3:17 PM UTC (bd8092d)">Diff</a>

Issue #777, opened 2021-04-29
all-but-illegible text style needs fix for humans


TallTed:
Brought to eye by #775, the `<span class="color-text" style="color:#b4a7d6;">` and similar [here in the source](https://github.com/w3c/vc-data-model/blob/6c429cfc54ebfa6f5b74a95dd51d46eed7856b9e/index.html#L834-L843), and [here in the preview (describing figures 7 and 8)](https://w3c.github.io/vc-data-model/#basic-vp) is all-but-illegible in my browser (Chrome Version 90.0.4430.93 (Official Build) (x86_64) on macOS El Capitan (10.11.6)). I does't seem likely that it would be much more legible in any other browser on any other platform.

<img width="817" alt="illegible_text" src="https://user-images.githubusercontent.com/2559602/116598394-c5dcca80-a8f4-11eb-8de0-34aa945a791b.png">

Regrettably, this styling didn't go through W3 CSS, so it probably counts as an errata for W3 purposes.

I *am* surprised it passed muster for the Accessibility review!

brentzundel:
We agree this needs fixing.

kdenhartog:
Noting that the text coloing is intended to be aligned with the color coordination of the graph below it. Would we prefer to update this to the links of the definitions and include a paraentheses to indicate color association or drop the color association between the text and figure 8 all together?

TallTed:
Tangentially, I would split the descriptions of Figure 7 and Figure 8 into separate paragraphs. (Yes, Fig 7 will be short.  That's OK.)

I would suggest a dotted-line in each color, surrounding the blocks of that color, extending to the right of the illustration, where the textual label for each collection should probably be in black.

With re-re-reading, I am thinking that the text should also be tweaked a bit, and either instead of or in addition to the "{first, second, third, fourth} graph", those graphs should be identified by the name used to label the current dashed-black-line blobs (i.e., "{Presentation, Credential, Credential Proof, Presentation Proof} Graph").  Without so doing, it can appear that the latter names are out of sync with the names used in the textual description -- which makes this already complex section even more difficult to comprehend.  I've taken a whack at this in #785.

TzviyaSiegman:
I haven't looked at this document in a long time, but I did an accessibility review years ago. If the colors are conveying meaning, the meaning needs to be conveyed by more than just color for those who cannot perceive color. (I can perceive color, and I am unsure what is being conveyed by the color, and I find it hard to read.)

chaals:
[...]

PR #835, opened 2021-11-13
Add respec-vc proof/JWT rendering to many other examples.

msporny:
This PR builds on PR #834 by just applying respec-vc to all other examples that it can be applied to at this time.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/835.html" title="Last updated on Nov 14, 2021, 6:07 PM UTC (90d16e3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/835/7790d6c...90d16e3.html" title="Last updated on Nov 14, 2021, 6:07 PM UTC (90d16e3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -933,7 +933,7 @@ discount from a university. In the example below, Pat receives an alumni
     <span class='comment'>// purpose of this proof</span>
     "proofPurpose": "assertionMethod",
     <span class='comment'>// the identifier of the public key that can verify the signature</span>
-    "verificationMethod": "https://example.edu/issuers/565049/keys/1",
+    "verificationMethod": "https://example.edu/issuers/565049#key-1",
     <span class='comment'>// the digital signature value</span>
     "jws": "eyJhbGciOiJSUzI1NiIsImI2NCI6ZmFsc2UsImNyaXQiOlsiYjY0Il19..TCYt5X
       sITJX1CxPCT8yAV-TVkIEq_PbChOMqsLfRoPsnsgw5WEuts01mq-pQy7UJiN5mgRxD-WUc
@@ -989,7 +989,7 @@ the <a>verifier</a> and <a>verified</a>.
       "type": "RsaSignature2018",
       "created": "2017-06-18T21:19:10Z",
       "proofPurpose": "assertionMethod",
-      "verificationMethod": "https://example.edu/issuers/565049/keys/1",
+      "verificationMethod": "https://example.edu/issuers/565049#key-1",
       "jws": "eyJhbGciOiJSUzI1NiIsImI2NCI6ZmFsc2UsImNyaXQiOlsiYjY0Il19..TCYt5X
         sITJX1CxPCT8yAV-TVkIEq_PbChOMqsLfRoPsnsgw5WEuts01mq-pQy7UJiN5mgRxD-WUc
         X16dUEMGlv50aqzpqh4Qktb3rk-BuQy72IFLOqV0G_zS245-kronKb78cPN25DGlcTwLtj
@@ -1199,7 +1199,7 @@ about the <code>id</code>.
         </dl>
 
         <pre class="example nohighlight vc" title="Usage of the id property"
-          data-vc-verification-method="https://example.edu/issuers/565049/keys/1">
+          data-vc-verification-method="https://example.edu/issuers/565049#key-1">
 {
   "@context": [
     "https://www.w3.org/2018/credentials/v1",
@@ -1274,7 +1274,8 @@ in a document containing machine-readable information about the
           </dd>
         </dl>
 
-        <pre class="example nohighlight" title="Usage of the type property">
+        <pre class="example nohighlight vc" title="Usage of the type property"
+          data-vc-verification-method="https://example.edu/issuers/565049#key-1">
 {
   "@context": [
     "https://www.w3.org/2018/credentials/v1",
@@ -1290,8 +1291,7 @@ in a document containing machine-readable information about the
       "type": "BachelorDegree",
       "name": "Bachelor of Science and Arts"
     }
-  },
-  "proof": { <span class="comment">...</span> }
+  }
 }
         </pre>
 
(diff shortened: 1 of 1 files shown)


TallTed (approved):
Looks good at a quick skim.

But I noticed that the paragraph describing Figure 5 and Figure 6 has a few colored-and-outlined words which are near illegible, which problem I thought we resolved months ago with substantial rewrite of this paragraph including splitting it into two... And that makes me very concerned that there may be related and/or other issues that may need re-resolution.

David-Chadwick:
Whilst the JWT example is no doubt correct, it would be more helpful to readers if the additional tab could be (or another one added) which shows the transformation of the credential into the JWT payload (prior to signing and base64 encoding) as in example 28. The reason being is that the transformation is the part that is specified in the VC specification, whilst the signing and base64 encoding are already standardised in referenced RFCs and we do not modify these specs.

msporny:
> Whilst the JWT example is no doubt correct, it would be more helpful to readers if the additional tab could be (or another one added) which shows the transformation of the credential into the JWT payload (prior to signing and base64 encoding) as in example 28. The reason being is that the transformation is the part that is specified in the VC specification, whilst the signing and base64 encoding are already standardised in referenced RFCs and we do not modify these specs.

I've raised an issue to track this since it has more to do with respec-vc:

https://github.com/w3c/vc-data-model/issues/836

Let's take JWT enhancements to the current examples there.

index.html
```diff
@@ -933,7 +933,7 @@ <h3>Concrete Lifecycle Example</h3>
     <span class='comment'>// purpose of this proof</span>
     "proofPurpose": "assertionMethod",
     <span class='comment'>// the identifier of the public key that can verify the signature</span>
-    "verificationMethod": "https://example.edu/issuers/565049/keys/1",
+    "verificationMethod": "https://example.edu/issuers/565049#key-1",
```
kdenhartog, line 936: How are you linking these particular verificationMethods? I was looking over the respec-vc implementation and didn't see you hard coding the private keys in to test this and from the looks of it the keys were being randomly generated on the fly so wasn't sure if people would be able to verify these or if that's acceptable for them to not be.

kdenhartog (approved):
From a spec perspective, everything looks good to me. There's some things I'd like to better understand about the implementation to help maintain it and improve a bit more is all.

iherman (approved):
This comment is just to make your life bit more difficult :-)

See, e.g., example. The "subject" of the example is the `credentialSubject` property, duly highlighted in the example. However, if I look at the signed example, that highlight disappears, and the reader is lost as to exactly what happened: one expects to see the top part of the example to be exactly the same, including the highlight. Would it be possible to make it so?

index.html
msporny: The short of it is that we're not (yet). It's a future enhancement for sure.

I was thinking of adding a section in the appendix of the spec that lists the public/private keys used for all examples as well as the documents located at the verification method URLs. Since they're generated per page refresh, we could either 1) hard code some keys, or 2) be okay with using new keys for every spec. There are benefits and drawbacks for both approaches.

msporny:
> However, if I look at the signed example, that highlight disappears, and the reader is lost as to exactly what happened: one expects to see the top part of the example to be exactly the same, including the highlight. Would it be possible to make it so?

Yes, one could apply a custom diff'ing algorithm to the pre-signed and signed output and then apply some sort of patching algorithm to match the highlights between the unsigned and signed output up. Another option would be to just pluck the `proof` field the signed output and tack it onto the end. The former is probably a week or two of coding, the latter would be a few hours... but then you run the risk of not tacking on the proof at the right place. There would have to be some regexes involved as well, I imagine. It's not entirely straightforward. :)

index.html
kdenhartog: Makes sense - good to know. I think I have a general idea of what is going on with that code now as well after this and think I could help to maintain it as we get specific requests from the WG on it in the future.

iherman:
> > However, if I look at the signed example, that highlight disappears, and the reader is lost as to exactly what happened: one expects to see the top part of the example to be exactly the same, including the highlight. Would it be possible to make it so?
> 
> Yes, one could apply a custom diff'ing algorithm to the pre-signed and signed output and then apply some sort of patching algorithm to match the highlights between the unsigned and signed output up. Another option would be to just pluck the `proof` field the signed output and tack it onto the end. The former is probably a week or two of coding, the latter would be a few hours... but then you run the risk of not tacking on the proof at the right place. There would have to be some regexes involved as well, I imagine. It's not entirely straightforward and so is left as an enhancement for the future. :)

I told you my goal is to make your life more difficult :-)

index.html
msporny: I would love help maintaining that package. Specifically, I want to get a BBS+ signature in there to demonstrate that we're on top of the whole "general purpose solution to selective disclosure" problem space.

TallTed:
The issue [I referenced](https://github.com/w3c/vc-data-model/pull/835#pullrequestreview-805415720) above was #777, which was addressed by PRs #785 and #786, but those no longer appear to have been merged?  Unless maybe this PR is against the wrong base?

msporny:
> The issue [I referenced](https://github.com/w3c/vc-data-model/pull/835#pullrequestreview-805415720) above was #777, which was addressed by PRs #785 and #786, but those no longer appear to have been merged? Unless maybe this PR is against the wrong base?

Hrm, this PR is in the main history: https://github.com/w3c/vc-data-model/commit/a0f486365a2ab83a4bd67093c43b47a7198d129b but this one is not: https://github.com/w3c/vc-data-model/commit/e0ec8de9bcd0c809c78c09310ea201d3d3dc023f 

@TallTed can you please open an issue to track this, we have to make sure we figure out what's going on here. Losing commits like this is a big deal/problem.

msporny:
I'm merging this PR into PR #834 (give that there are no objections, I need to clean up the JWT encoding, and because that other PR isn't going to be merged until all of this is correct). This is an administrative move to merge two PRs into one before we send out the 14 day merge announcement to the CCG.

PR #849, opened 2021-12-09
URL, spelling, markup, formatting fixes

davidlehn:
Various cleanups.  I'd look at the various commits.  Most are basic fixes.

The "Fix sections and indents." commit hits a lot of markup so makes the overall diff a mess.  The intent was to make the markup formatting consistent.  That was a bit of a challenge because the `<hN>` tags had various numbering issues.  I think I fixed them to what they should be.  If that commit should be a different PR, I can do that.

***
<a href="https://pr-preview.s3.amazonaws.com/davidlehn/vc-data-model/pull/849.html" title="Last updated on Dec 9, 2021, 12:58 AM UTC (0d3afac)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/849/4b9bdd4...davidlehn:0d3afac.html" title="Last updated on Dec 9, 2021, 12:58 AM UTC (0d3afac)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/davidlehn/vc-data-model/pull/849.html" title="Last updated on Dec 16, 2021, 12:31 AM UTC (f166f2d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/849/4b9bdd4...davidlehn:f166f2d.html" title="Last updated on Dec 16, 2021, 12:31 AM UTC (f166f2d)">Diff</a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ credentials on the Web in a way that is cryptographically secure,
 privacy respecting, and machine-verifiable.
 
 We encourage contributions meeting the [Contribution
-Guidelines](CONTRIBUTING.md).  While we prefer the creation of issues
+Guidelines](CONTRIBUTING.md). While we prefer the creation of issues
 and Pull Requests in the GitHub repository, discussions often occur
 on the
 [public-vc-wg](http://lists.w3.org/Archives/Public/public-vc-wg/)
--- a/VCDMExplainer.md
+++ b/VCDMExplainer.md
@@ -346,7 +346,7 @@ but does not require it.
 
 ## Features at Risk
 
-We have asked for [preliminary commitments from potential implementers](https://docs.google.com/spreadsheets/d/1SzfAUA0J72-1BORHJEmY4cdZrQ6vmKy4oq_24r_NwB4/edit?usp=sharing).  We expect all features to have at
+We have asked for [preliminary commitments from potential implementers](https://docs.google.com/spreadsheets/d/1SzfAUA0J72-1BORHJEmY4cdZrQ6vmKy4oq_24r_NwB4/edit?usp=sharing). We expect all features to have at
 least two implementations. With current commitments, some features have exactly two.  
 Of course we will continue to recruit for additional implementations.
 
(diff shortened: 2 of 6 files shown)


TallTed:
The `<hN>` here are expected to be cosmetic, as the W3 publishing tools rationalize them based on the `<section>` tags, which don't require humans to track depth at all; only that they're always opened and closed where appropriate.  (Though I'll grant you that the `<hN>` tags can be quite confusing when trying to clean up the HTML source!)

davidlehn:
Yeah, I know the tools handle the header numbers, but I figure if they are there, they should be correct at this stage of spec development.  It is a huge markup hassle to move sections around so I waited a while to even try to do that update.  Ideally there would be a fancy formatting tool that understands the style used here.

msporny (approved):
All changes are editorial, LGTM.

@davidlehn was the formatting done using some automated tooling, or did you do it by hand?

davidlehn:
> @davidlehn was the formatting done using some automated tooling, or did you do it by hand?

Both.

Valid markup structure: I use `tidy -qe FILE.html` and internal vim error markers to see problems. Then fix by hand.  Something like this should be an automatic PR check.  A bit tricky with partial files that get included, but could at least help with main file.

Grammar, period spacing etc: vim flags it.

Spelling: vim spelling checker along with a large list of valid keywords and other local things. Still very annoying with all the json/js etc false positives, but can scan pretty quick.

Markup indents: vim is setup to know how to indent tags based on previous tags.  Can just hit a formatting command on markup lines and it moves things to the right level, assuming previous markup is correct. Can't do that on the left-flush text and some tag attribute continuation lines, so it's still manual to go through the lines.  If all markup in a file is setup a certain way, I think this could just be a single macro.

davidlehn:
@msporny A minor http to https change slipped into a REC file. Should I remove that so the snapshot isn't changed?

msporny:
@davidlehn 
> vim ... vim ... vim ... vim ... vim

So, what you're saying is that we need to figure out a way to run vim in a Github Action, eh? :P 

> @msporny A minor http to https change slipped into a REC file. Should I remove that so the snapshot isn't changed?

Ideally, yes.

davidlehn:
> > @msporny A minor http to https change slipped into a REC file. Should I remove that so the snapshot isn't changed?
> 
> Ideally, yes.

Done.

msporny:

PR #854, opened 2021-12-26
Create SECURITY.md

pavelkorablev6:
w3c:v1.1

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,21 @@
+# Security Policy
+
+## Supported Versions
+
+Use this section to tell people about which versions of your project are
+currently being supported with security updates.
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 5.1.x   | :white_check_mark: |
+| 5.0.x   | :x:                |
+| 4.0.x   | :white_check_mark: |
+| < 4.0   | :x:                |
+
+## Reporting a Vulnerability
+
+Use this section to tell people how to report a vulnerability.
+
+Tell them where to go, how often they can expect to get an update on a
+reported vulnerability, what to expect if the vulnerability is accepted or
+declined, etc.


msporny (changes requested):
Adding a generic SECURITY.md file is probably not helpful at this point in time. This is something that other W3C specifications don't do, AFAIK.

There are only two versions of VCs (1.0 and 1.1), and both are supported at present, AFAIK. Reporting a vulnerability is a good thing to have, but it's noted in the specification (raise an issue).

Largely, -1 to this PR because it's generic and having multiple places w/ the same information can be problematic as the Editors will forget to update all the places over time.

kdenhartog:
-1 from me as well. If the specification itself has an issue it should be raised as an issue and addressed via proper discussions rather than a limited number of people finding out about it IMO. These security documents are more useful for implementations I'd think.

kdenhartog (changes requested):
If the WG does decide this is needed (-1 from me and the other editor so far) then the text needs to be updated to be more specific than the generic text included at the moment.

TallTed:
@pavelkorablev6 -- Different people will approach this project and its information in different ways.  `SECURITY.md` appears to be a GitHub-minded semi-de-facto standard place for such information, but please note that it is not a W3-minded place for this.

Might I suggest, therefore, that @pavelkorablev6 modify the `SECURITY.md` content in this PR to suit this W3 project?  That is, change the boilerplate, which really says nothing right now, to link to the relevant section of [the published Data Model document](https://www.w3.org/TR/vc-data-model/#security-considerations) and [the Data Model issues page](https://github.com/w3c/vc-data-model/issues).

Given those changes, I would think that adding a reciprocal comment to the HTML source of the Data Model document, which points to this `SECURITY.md`, would be sufficient to remind editors to update the latter if the hrefs it targets are changed for whatever unlikely reason.  Maybe an additional check in the GitHub Actions could flag any errors which might result from `SECURITY.md` hrefs targeting URLs which no longer exist?

msporny:
@TallTed wrote:
> Given those changes, I would think that adding a reciprocal comment to the HTML source of the Data Model document, which points to this SECURITY.md, would be sufficient to remind editors to update the latter if the hrefs it targets are changed for whatever unlikely reason.

I'm afraid that I would almost certainly miss that pointer in the HTML source. It's exceedingly easy to miss content like that, which is why we tend to prefer content in one location to make sure we don't screw things up over time. Working memory fades rapidly, and SECURITY.md is not a part of many Editor's workflows at W3C, so unless we can automate this (and we have bigger fish to fry here), I suggest we just drop it entirely. The place for this information is in the Status of the Document section and the README.md file. Almost all of the thousands of W3C repos don't use SECURITY.md to my knowledge.

I appreciate that we're trying to find a way to pull in @pavelkorablev6's PR, but in this instance, I think it does more harm than good.

TallTed:
> That said, I wouldn't be averse to putting this information in `README.md`, but what that should really do is just duplicate the Status of the Document section verbatim... which tells people how to engage w/ issues/PRs/etc.

That seems a reasonable solution, given which I wouldn't mind also having `SECURITY.md` reduced to a pointer to `README.md`.

msporny:
Marking this PR as "close after 7 days". The individual that raised the PR is non-responsive and it does not look like this PR is going to be accepted in it's current form.

kdenhartog:
closing this PR - it's been 10 days since the label was added.

PR #898, opened 2022-07-30
Update README to reflect v2.0 work.

msporny:
This PR updates the README.md to reflect the v2.0 work.

Note some of the minor tweaks to the process, which the Chairs/Staff will want to review and update as appropriate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,8 +44,8 @@ mailing list.
      <dd>Mark with "substantive" tag.</dd>
    </dl>
 3. W3C CCG is automatically notified of PRs as they are raised and discussed.
-4. PRs are merged in 7 days if there is adequate review, as determined by
-   the Chairs and Editors, and no objections.
+4. PRs are usually merged in 7 days if there is adequate review and consensus,
+   as determined by the Chairs and Editors.
 
 ### Roadmap for 2022-2024
 
--- a/README.md
+++ b/README.md
@@ -1,65 +1,60 @@
 ## Verifiable Credentials Data Model and Representations
 
-Credentials are a part of our daily lives; driver's licenses are
-used to assert that we are capable of operating a motor vehicle,
-university degrees can be used to assert our level of education,
-and government-issued passports enable us to travel between countries.
-This specification provides a mechanism to express these sorts of
-credentials on the Web in a way that is cryptographically secure,
+Credentials are a part of our daily lives; driver's licenses are used to assert
+that we are capable of operating a motor vehicle, university degrees can be used
+to assert our level of education, and government-issued passports enable us to
+travel between countries. This specification provides a mechanism to express
+these sorts of credentials on the Web in a way that is cryptographically secure,
 privacy respecting, and machine-verifiable.
 
 We encourage contributions meeting the [Contribution
 Guidelines](CONTRIBUTING.md). While we prefer the creation of issues
-and Pull Requests in the GitHub repository, discussions often occur
+and Pull Requests in the GitHub repository, discussions also occur
 on the
 [public-vc-wg](http://lists.w3.org/Archives/Public/public-vc-wg/)
-mailing list as well.
+mailing list.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,65 +1,60 @@
...
 
 ## Process Overview for VC Data Model Pull Requests
-1. For now, we will focus only on merging new errata PRs into this repository,
-   but encourage activity related to new features.
-2. Once a PR is opened, chairs and editors make judgement call on whether
-   changes are substantive or editorial.
+1. Anyone can open a PR on the repository. Note that for the PR to be merged,
+   the individual must agree to the
+   [W3C Patent Policy](https://www.w3.org/Consortium/Patent-Policy/).
+2. Once a PR is opened on the `v2.0` branch, chairs and editors make judgement
```
iherman, line 38: I expect this may become a source of confusion, because it is not the "usual" setup. That is why I would be in favour of using the 'main' branch for v2.0 for now on.

README.md
```diff
@@ -1,65 +1,60 @@
...
-- VC Test Suite Refactoring
-- Start planning VC v2 Work, request a rechartering 3-6 months before end of
-  year to keep VC WG functioning.
-
-### Focus areas
-- [v1] Fixing a specific bug
-- [v1] Update examples in the spec to make them modern
-- [v2] VC `@context` needs updating, possibly with security vocab modularized
-  into smaller components instead of all included into a large context file.
+3. W3C CCG is automatically notified of PRs as they are raised and discussed.
```
iherman, line 46: Has an automatic procedure been set up for this?

iherman (approved):

README.md
msporny: There is no longer a `main` branch. IIRC, we removed it in the v1.1 days because it was confusing people. 

I'm indifferent wrt. if we add it back by renaming the `v2.0` back to `main`.

We probably need to decide if we are going to use tags to specify versions, or branches -- branches make sense when there is active development (such as the v1.1 work), tags make sense when we cut a release and don't plan to update that particular release (which is the case for v1.0 and v1.1). 

So, we could say, `main` is always the latest ED spec. When we need to make a minor release, we create a branch, work on that branch and then release from that branch. Once the release is finalized, we convert the branch to a tag and delete the branch.

This means, we would:

1. Create tags for: `v1.0` and `v1.1`
2. Delete the branches for `v1.0` and `v1.1`
3. Rename the `v2.0` branch to `main` and set Github Pages to publish from `main`.

The other option would just skip step 3 above. Thoughts?

README.md
msporny: Yep, via [github-notify-ml](https://github.com/w3c/github-notify-ml-config): 

CCG receives weekly updates for all VCWG specs on Mondays: https://lists.w3.org/Archives/Public/public-credentials/2022Jul/0104.html

README.md
iherman: Personally, what I am interested in is the v2.0->main change. I am neutral as for (1) and (2); my lazy nature says just leave things as they are, it does not make too much difference in practice.

brentzundel: I think having our primary branch be called `main` will be less confusing for folks, especially since we are not working on multiple versions concurrently this time around.

README.md
```diff
@@ -1,65 +1,60 @@
...
+4. PRs are merged in 7 days if there is adequate review, as determined by
+   the Chairs and Editors, and no objections.
```
brentzundel, line 48:
```suggestion
4. PRs are usually merged in 7 days if there is adequate review and consensus,
   as determined by the Chairs and Editors.
```

brentzundel (approved):

Sakurann (approved):

README.md
msporny: > I think having our primary branch be called `main` will be less confusing for folks

The main branch has been renamed to `main`.

msporny:

PR #920, opened 2022-08-25
Update the formal vocabulary files

### LINKED ISSUE #770: Bug in the credential vocabulary specification

### LINKED ISSUE #916: Data Model Vocabulary versioning questions
In agreement with @msporny I will maintain the formal [VC Vocabulary definition](https://w3c.github.io/vc-data-model/vocab/credentials/credentials.html) document. Apart from cosmetic changes, this means handling possible bugs like #770, but also adapting it to the v2 version of the model.

There is a practical versioning issue, though, which may need a consensus before moving on. The [VC Vocabulary definition](https://w3c.github.io/vc-data-model/vocab/credentials/credentials.html) has never been published as a W3C Note. That means that the current (ie, dated 26 May 2019) is the _only_ existing version of that document. If I go ahead and make any change whatsoever, history will be lost. The change we would do on the vocabulary definition will lead to a 2nd version of the vocabulary, that may or may not be 100% backward compatible. But I am hesitant to make any change that would indeed lead to a loss of history.

I would therefore propose that would make a "fork" (not necessarily a github fork) of the vocabulary definition. This can take two forms:

1. I make a separate folder in our `vc-data-model` repo `vocab/credential2.0`, make all the changes there and leave the current document intact. The new URL for the vocabulary will become `https://w3c.github.io/vc-data-model/vocab/credentials2.0/credentials.html`
2. I create a separate github repo, `vc-data-model-vocab` and move all the documents there. The new URL for the vocabulary will become `https://w3c.github.io/vc-data-model-vocab`. This may make it easier to publish, if the WG so decides at some point in the future, a bona fide WG Note for the vocabulary; but even without that decision it may avoid minor conflicts with the development of the main data model.

I am (though mildly) in favour of the second alternative.

Note that, as it came up in an discussion on the thread, this WG may want to take over the https://w3c-ccg.github.io/security-vocab/ as well to update and refresh. Whatever we decide here may have an effect on that decision, too, if we want to be consistent...

cc @msporny

iherman:
This PR creates a v2 version of the RDFS vocabulary definition of the data model. Eventually, just as the context v2 will supersede the current v1, this vocabulary definition will supersede the current one.

The plan is to catch up with the evolution of v2 context and keep the two in sync.

Because there is no diff (and there were lots of small changes that would make a diff unusable anyway, here are the main changes on the file(s):

**Editorial changes:**

- updated to latest respec
- updated the title of the document, referring to v2.0
- changed the list of editors
- minor CSS addition to ensure a proper formatting of terms
- removed the generation of the context file (it is maintained elsewhere, having this file here is probably a leftover which only create confusion; we should not have a context at two different places)
- added an explicit version info into the html template (`owl:versionInfo` is not o.k., it would have to take an IRI as an object)
- removed terms that are not specified in this vocabulary from the listing of term definitions in §4
- added (in the HTML version) an `id` attribute to all terms, so that something like `.../#evidence` would properly dereference.

**Handling bugs in the previous version:**

- added the `odrl` namespace (which is used) to the list of used namespaces in the HTML document , and removed `sec` (which is not used)
- added the `VerifiableCredentialGraph` class and changed the domain of the `verifiableCredential` property, to fix #770
- modified mk_vocab.rb not to add an erroneous `dc:imports` triple that had no object

Note that this PR does _not_ yet sync up with the latest version of the context file, as merged in #905. That should come in a new PR if and when this one is accepted and merged.

To make the review easier, here are the previews of the three generated file:

- [HTML version](https://raw.githack.com/w3c/vc-data-model/vocab-v2/vocab/credentials/v2/credentials.html)
- [Turtle version](https://raw.githubusercontent.com/w3c/vc-data-model/vocab-v2/vocab/credentials/v2/credentials.ttl)
- [JSON-LD version](https://raw.githubusercontent.com/w3c/vc-data-model/vocab-v2/vocab/credentials/v2/credentials.jsonld)

Fix #916


vocab/credentials/v2/vocab.csv
```diff
@@ -0,0 +1,23 @@
...
+credentialStatus,rdf:Property,Credential Status,,VerifiableCredential,cred:CredentialStatus,@id,,"The value of the `credentialStatus` property MUST include the `id` property, which MUST be a URL, and the `type` property, which expresses the credential status type (also referred to as the credential status scheme), which MUST provide enough information to determine the current status of the credential (for example, suspended or revoked)."
+credentialSubject,rdf:Property,Credential Subject,,VerifiableCredential,,@id,,An entity about which claims are made.
+evidence,rdf:Property,evidence,,VerifiableCredential,cred:CredentialEvidence,,,The value of the `evidence` property MUST be one or more evidence schemes providing enough information to a verifier to determine whether the evidence gathered meets their requirements for issuing a credential. The precise content of each evidence scheme is determined by the specific evidence type definition.
```
pchampin, line 16: The ranges of `credentialSchema`, `credentialStatus` and `evidence` are classes in the `cred:` namespace, but are not defined in the vocabulary.

pchampin (commented):
The way the documentations of terms are written is ambiguous. Some are focused on the conceptual level (e.g. "The value of the credentialSchema property MUST be one or more data schemas") while other are focused on the JSON-LD syntactic level (e.g. "The value of the credentialStatus property MUST include the id property, which MUST be a URL").

This should be harmonized, all focusing on the conceptual level, if indeed the JSON-LD context is maintained elsewhere.

In the same line, the `Term Definitions` section in the HTML document seems to be mostly focused on JSON-LD, not RDFS, so I would simply remove it.

vocab/credentials/v2/credentials.ttl
```diff
@@ -0,0 +1,106 @@
+@prefix dc: <http;//purl.org/dc/terms/> .
```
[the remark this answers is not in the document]
iherman: Oops. But the error is in the ruby file, this file is generated.

vocab/credentials/v2/credentials.ttl
```diff
@@ -0,0 +1,106 @@
...
+@prefix cred: <https://w3.org/2018/credentials#> .
+@prefix odrl: <http://www.w3.org/ns/odrl/2/> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+
+# CSVM Ontology definition
+cred: a owl:Ontology;
+  dc:title "Verifiable Credentials Vocabulary v2.0"@en;
+  dc:description """This document describes the RDFS vocabulary description used for Verifiable Credentials [[VC-DATA-MODEL]]."""@en;
+  dc:date "2022-08-25"^^xsd:date;
+  rdfs:seeAlso <https://www.w3.org/TR/vc-data-model-2.0/>;
```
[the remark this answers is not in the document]
iherman: The error is in the ruby file, this file is generated.

vocab/credentials/v2/mk_vocab.rb
```diff
@@ -0,0 +1,361 @@
...
+    json = JSON.parse(to_jsonld)
+    eruby = Erubis::Eruby.new(File.read("template.html"))
+    eruby.result(ont: json['@graph'], context: json['@context'].is_a?(Array) ? json['@context'].last : json['@context'])
+  end
+
+  def to_ttl
+    output = []
+
+    prefixes = {
+      "dc"   => {subClassOf: "http;//purl.org/dc/terms/"},
```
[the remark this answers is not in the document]
iherman: Done

vocab/credentials/v2/mk_vocab.rb
```diff
@@ -0,0 +1,361 @@
...
+    prefixes.each {|id, entry| output << "@prefix #{id}: <#{entry[:subClassOf]}> ."}
+
+    output << "\n# CSVM Ontology definition"
+    output << "cred: a owl:Ontology;"
+    output << %(  dc:title "#{TITLE}"@en;)
+    output << %(  dc:description """#{DESCRIPTION}"""@en;)
+    output << %(  dc:date "#{date}"^^xsd:date;)
+    #output << %(  dc:imports #{imports.map {|i| '<' + i + '>'}.join(", ")};)
+    #output << %(  owl:versionInfo <#{commit}>;)
+    output << %(  rdfs:seeAlso #{seeAlso.map {|i| '<' + i + '>'}.join(", ")};)
```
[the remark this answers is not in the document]
iherman: Done

vocab/credentials/v2/vocab.csv
iherman: Oops.

@msporny I would need a one sentence description (label) to be added to the vocabulary for the following classes:

- cred:CredentialSchema
- cred:CredentialStatus
- cred:CredentialEvidence

For the time being I have added the class definition with the label set to T.B.D. but a proper sentence would be good.

iherman:
> The way the documentations of terms are written is ambiguous. Some are focused on the conceptual level (e.g. "The value of the credentialSchema property MUST be one or more data schemas") while other are focused on the JSON-LD syntactic level (e.g. "The value of the credentialStatus property MUST include the id property, which MUST be a URL").
> 
> This should be harmonized, all focusing on the conceptual level, if indeed the JSON-LD context is maintained elsewhere.

I would agree with this. I note that the definitions are usually come as a copy of the data model spec. I still have to familiarize myself more with the model before I could provide a set of alternative texts, and the spec may also change, so I would keep that on a todo list and do that at some point. Again, point well taken.

> In the same line, the Term Definitions section in the HTML document seems to be mostly focused on JSON-LD, not RDFS, so I would simply remove it.

Yes, I was wondering about that myself. This section does not bring anything new and valuable to the reader, and I was tempted to remove it altogether, but I shied away from it for now. @msporny is it o.k. to remove it?

iherman:
@pchampin just to be on the safe side: is the text used for `verifiableCredential` and `VerifiableCredentialGraph` ok? These is the term that I have introduced to handle #770.  I also know that the exact relationship of the URL for a graph and the graph itself is not 100% specified; I want to avoid unnecessary raised eyebrows...

pchampin:
> @pchampin just to be on the safe side: is the text used for `verifiableCredential` and `VerifiableCredentialGraph` ok? These is the term that I have introduced to handle #770.

They look good to me, but in the class description, I would replace "Instances of this class *identify* RDF graphs" with "Instances of this class *are* RDF graphs".

> I also know that the exact relationship of the URL for a graph and the graph itself is not 100% specified; I want to avoid unnecessary raised eyebrows...

The change above keeps us on the safe side, I think. It does not imply any means by which the object of the `verifiableCredential` property is linked to the "content" of the graph.

pchampin:
> Well... a class is a collection of resources, more exactly URI-s to resources, right?

No, I wouldn't put it that way.

Quoting \[[RDF11-MT](https://www.w3.org/TR/rdf11-mt/#rdfs-interpretations)\]:
> It is convenient to state the RDFS semantics in terms of a new semantic construct, a class, i.e. a resource which represents a set of things in the universe (...)

"A set of things in the universe" (a.k.a. resources), not a set of IRIs. The fact that those resources may be identified by IRIs is irrelevant to the class itself.

So, to develop my argument above:
a triple `<#x> cred:verifiableCredential <#y>` implies that
  - `<#x>` identifies an instance of `cred:VerifiablePresentation`
  - `<#y>` identifies an instance of `cred:VerifiableCredentialGraph`, which according to the definition is an RDF graph
but the ontology itself is agnostic about how to find out what triples are contained in the graph identified by `<#y>`

To fill that gap, the common practice with VC is to put the graph *identified* by `<#y>` into a [named graph](https://www.w3.org/TR/rdf11-concepts/#dfn-named-graph) whose graph name is `<#y>`. People might raise their eyebrow at that, but they can't blame the ontology for it -- which I think was your original question.

iherman:
> > Well... a class is a collection of resources, more exactly URI-s to resources, right?
> 
> No, I wouldn't put it that way.
> 
> Quoting [[RDF11-MT](https://www.w3.org/TR/rdf11-mt/#rdfs-interpretations)]:
> 
> > It is convenient to state the RDFS semantics in terms of a new semantic construct, a class, i.e. a resource which represents a set of things in the universe (...)
> 
> "A set of things in the universe" (a.k.a. resources), not a set of IRIs. The fact that those resources may be identified by IRIs is irrelevant to the class itself.

Ok, I stand corrected...

> 
> So, to develop my argument above: a triple `<#x> cred:verifiableCredential <#y>` implies that
> 
> * `<#x>` identifies an instance of `cred:VerifiablePresentation`
> * `<#y>` identifies an instance of `cred:VerifiableCredentialGraph`, which according to the definition is an RDF graph
>   but the ontology itself is agnostic about how to find out what triples are contained in the graph identified by `<#y>`
> 

Based on what you said above, I am fine with this formulation.

> To fill that gap, the common practice with VC is to put the graph _identified_ by `<#y>` into a [named graph](https://www.w3.org/TR/rdf11-concepts/#dfn-named-graph) whose graph name is `<#y>`. People might raise their eyebrow at that, but they can't blame the ontology for it -- which I think was your original question.

Right. But I will not put this into the text itself:-) Just remaining terse.

iherman:
(maybe that deserves a separate issue, you tell me.)

The specification makes a difference between a "credential" and a "verifiable credential", as well as a "presentation" and a "verifiable presentation", that could be roughly characterized by saying that a "verifiable credential" (resp. "verifiable presentation") is a temper evident "credential" (resp. "presentation"). If this distinction is important, then it would deserve a separation in the ontology, by defining a "Credential" class and a "Verifiable Credential" subclass. (resp. "Presentation" and "Verifiable Presentation"). This is easy to add to the ontology.

If we do so, the question is whether the domain of the various properties should be verifiable credentials or just credentials. It strikes that almost none of the properties in this vocabulary depend on the "verifiable" aspect, maybe except for "verifiableCredential" which, at least with its name, is bound to "verifiable". The only property that _MUST_ have a domain of "verifiable ..." would be `proof`, but that is not defined in this vocabulary...

WDYT?

vocab/credentials/v2/vocab.csv
iherman: I make this resolved now, although the descriptions may require review.

vocab/credentials/v2/README.md
```diff
@@ -0,0 +1,3 @@
+# Updating vocabulary
+
+Vocabulary definitions are managed in vocab.csv. Add or change entries within this file. Regenerate `credentials.ttl`, `credentials.jsonld`, and `credentials.html` by running `mk_vocab.rb`.
```
TallTed, line 3:
```suggestion
Vocabulary definitions are managed in `vocab.csv`. Add or change entries within this file. Regenerate `credentials.ttl`, `credentials.jsonld`, and `credentials.html` by running `mk_vocab.rb`.
```

dlongley:
@iherman,

> The specification makes a difference between a "credential" and a "verifiable credential", as well as a "presentation" and a "verifiable presentation", that could be roughly characterized by saying that a "verifiable credential" (resp. "verifiable presentation") is a temper evident "credential" (resp. "presentation"). If this distinction is important, then it would deserve a separation in the ontology, by defining a "Credential" class and a "Verifiable Credential" subclass. (resp. "Presentation" and "Verifiable Presentation"). This is easy to add to the ontology.

IIRC, avoiding adding types in the ontology for this was avoided in the 1.0 work to keep down the complexity. I imagine we'll want to continue doing so as a trade off (simplicity over theoretrical purity), but we'll have to see. For this reason, I don't think we should add it in this PR.

vocab/credentials/v2/credentials.jsonld
```diff
@@ -0,0 +1,385 @@
+{
+    "@context": [
+        {
+            "@version": 1.1
```
dlongley, line 4: We've been deliberately avoiding setting `@version` everywhere -- JSON-LD 1.1 parsers handle this automatically via feature detection.

iherman: Ah. That was a leftover (I presume) from the time we thought the `@version` was required in JSON-LD 1.1. (If you remember, this change came pretty much at the end of the JSON-LD 1.1 development.) Thanks.

Done.

iherman:
> I imagine we'll want to continue doing so as a trade off (simplicity over theoretrical purity), but we'll have to see. For this reason, I don't think we should add it in this PR.

Personally, I would still be in favour of having this (or, alternatively, the spec text should not make a difference between "Credential" and "Verifiable Credential") but I am o.k. postponing this discussion. If we close this PR I may raise an issue just to keep a trace for ourselves.

OR13 (approved):

dlongley (approved):

vocab/credentials/v2/credentials.jsonld
```diff
@@ -0,0 +1,382 @@
+{
+    "@context": [
+        "https://w3id.org/security/v2",
```
OR13, line 3: hmm... prefer to not do this... this will trigger an entire 2 other network requests, since v2 relies on v1... thats 3 network requests to load the complete v2 context...

OR13: This can be avoided by simply copying the terms required by the vc data model... instead of copying everything...for example:

```json
{
  "@context": [{
    "@version": 1.1
  }, "https://w3id.org/security/v1", {
    "AesKeyWrappingKey2019": "sec:AesKeyWrappingKey2019",
    "DeleteKeyOperation": "sec:DeleteKeyOperation",
    "DeriveSecretOperation": "sec:DeriveSecretOperation",
    "EcdsaSecp256k1Signature2019": "sec:EcdsaSecp256k1Signature2019",
    "EcdsaSecp256r1Signature2019": "sec:EcdsaSecp256r1Signature2019",
    "EcdsaSecp256k1VerificationKey2019": "sec:EcdsaSecp256k1VerificationKey2019",
    "EcdsaSecp256r1VerificationKey2019": "sec:EcdsaSecp256r1VerificationKey2019",
    "Ed25519Signature2018": "sec:Ed25519Signature2018",
    "Ed25519VerificationKey2018": "sec:Ed25519VerificationKey2018",
    "EquihashProof2018": "sec:EquihashProof2018",
    "ExportKeyOperation": "sec:ExportKeyOperation",
    "GenerateKeyOperation": "sec:GenerateKeyOperation",
    "KmsOperation": "sec:KmsOperation",
    "RevokeKeyOperation": "sec:RevokeKeyOperation",
    "RsaSignature2018": "sec:RsaSignature2018",
    "RsaVerificationKey2018": "sec:RsaVerificationKey2018",
    "Sha256HmacKey2019": "sec:Sha256HmacKey2019",
    "SignOperation": "sec:SignOperation",
    "UnwrapKeyOperation": "sec:UnwrapKeyOperation",
    "VerifyOperation": "sec:VerifyOperation",
    "WrapKeyOperation": "sec:WrapKeyOperation",
    "X25519KeyAgreementKey2019": "sec:X25519KeyAgreementKey2019",

    "allowedAction": "sec:allowedAction",
    "assertionMethod": {"@id": "sec:assertionMethod", "@type": "@id", "@container": "@set"},
    "authentication": {"@id": "sec:authenticationMethod", "@type": "@id", "@container": "@set"},
    "capability": {"@id": "sec:capability", "@type": "@id"},
    "capabilityAction": "sec:capabilityAction",
    "capabilityChain": {"@id": "sec:capabilityChain", "@type": "@id", "@container": "@list"},
    "capabilityDelegation": {"@id": "sec:capabilityDelegationMethod", "@type": "@id", "@container": "@set"},
    "capabilityInvocation": {"@id": "sec:capabilityInvocationMethod", "@type": "@id", "@container": "@set"},
    "caveat": {"@id": "sec:caveat", "@type": "@id", "@container": "@set"},
    "challenge": "sec:challenge",
    "ciphertext": "sec:ciphertext",
    "controller": {"@id": "sec:controller", "@type": "@id"},
    "delegator": {"@id": "sec:delegator", "@type": "@id"},
    "equihashParameterK": {"@id": "sec:equihashParameterK", "@type": "xsd:integer"},
    "equihashParameterN": {"@id": "sec:equihashParameterN", "@type": "xsd:integer"},
    "invocationTarget": {"@id": "sec:invocationTarget", "@type": "@id"},
    "invoker": {"@id": "sec:invoker", "@type": "@id"},
    "jws": "sec:jws",
    "keyAgreement": {"@id": "sec:keyAgreementMethod", "@type": "@id", "@container": "@set"},
    "kmsModule": {"@id": "sec:kmsModule"},
    "parentCapability": {"@id": "sec:parentCapability", "@type": "@id"},
    "plaintext": "sec:plaintext",
    "proof": {"@id": "sec:proof", "@type": "@id", "@container": "@graph"},
    "proofPurpose": {"@id": "sec:proofPurpose", "@type": "@vocab"},
    "proofValue": "sec:proofValue",
    "referenceId": "sec:referenceId",
    "unwrappedKey": "sec:unwrappedKey",
    "verificationMethod": {"@id": "sec:verificationMethod", "@type": "@id"},
    "verifyData": "sec:verifyData",
    "wrappedKey": "sec:wrappedKey"
  }]
}{
  "@context": [{
    "@version": 1.1
  }, "https://w3id.org/security/v1", {
    "AesKeyWrappingKey2019": "sec:AesKeyWrappingKey2019",
    "DeleteKeyOperation": "sec:DeleteKeyOperation",
    "DeriveSecretOperation": "sec:DeriveSecretOperation",
    "EcdsaSecp256k1Signature2019": "sec:EcdsaSecp256k1Signature2019",
    "EcdsaSecp256r1Signature2019": "sec:EcdsaSecp256r1Signature2019",
    "EcdsaSecp256k1VerificationKey2019": "sec:EcdsaSecp256k1VerificationKey2019",
    "EcdsaSecp256r1VerificationKey2019": "sec:EcdsaSecp256r1VerificationKey2019",
    "Ed25519Signature2018": "sec:Ed25519Signature2018",
    "Ed25519VerificationKey2018": "sec:Ed25519VerificationKey2018",
    "EquihashProof2018": "sec:EquihashProof2018",
    "ExportKeyOperation": "sec:ExportKeyOperation",
    "GenerateKeyOperation": "sec:GenerateKeyOperation",
    "KmsOperation": "sec:KmsOperation",
    "RevokeKeyOperation": "sec:RevokeKeyOperation",
    "RsaSignature2018": "sec:RsaSignature2018",
    "RsaVerificationKey2018": "sec:RsaVerificationKey2018",
    "Sha256HmacKey2019": "sec:Sha256HmacKey2019",
    "SignOperation": "sec:SignOperation",
    "UnwrapKeyOperation": "sec:UnwrapKeyOperation",
    "VerifyOperation": "sec:VerifyOperation",
    "WrapKeyOperation": "sec:WrapKeyOperation",
    "X25519KeyAgreementKey2019": "sec:X25519KeyAgreementKey2019",

    "allowedAction": "sec:allowedAction",
    "assertionMethod": {"@id": "sec:assertionMethod", "@type": "@id", "@container": "@set"},
    "authentication": {"@id": "sec:authenticationMethod", "@type": "@id", "@container": "@set"},
    "capability": {"@id": "sec:capability", "@type": "@id"},
    "capabilityAction": "sec:capabilityAction",
    "capabilityChain": {"@id": "sec:capabilityChain", "@type": "@id", "@container": "@list"},
    "capabilityDelegation": {"@id": "sec:capabilityDelegationMethod", "@type": "@id", "@container": "@set"},
    "capabilityInvocation": {"@id": "sec:capabilityInvocationMethod", "@type": "@id", "@container": "@set"},
    "caveat": {"@id": "sec:caveat", "@type": "@id", "@container": "@set"},
    "challenge": "sec:challenge",
    "ciphertext": "sec:ciphertext",
    "controller": {"@id": "sec:controller", "@type": "@id"},
    "delegator": {"@id": "sec:delegator", "@type": "@id"},
    "equihashParameterK": {"@id": "sec:equihashParameterK", "@type": "xsd:integer"},
    "equihashParameterN": {"@id": "sec:equihashParameterN", "@type": "xsd:integer"},
    "invocationTarget": {"@id": "sec:invocationTarget", "@type": "@id"},
    "invoker": {"@id": "sec:invoker", "@type": "@id"},
    "jws": "sec:jws",
    "keyAgreement": {"@id": "sec:keyAgreementMethod", "@type": "@id", "@container": "@set"},
    "kmsModule": {"@id": "sec:kmsModule"},
    "parentCapability": {"@id": "sec:parentCapability", "@type": "@id"},
    "plaintext": "sec:plaintext",
    "proof": {"@id": "sec:proof", "@type": "@id", "@container": "@graph"},
    "proofPurpose": {"@id": "sec:proofPurpose", "@type": "@vocab"},
    "proofValue": "sec:proofValue",
    "referenceId": "sec:referenceId",
    "unwrappedKey": "sec:unwrappedKey",
    "verificationMethod": {"@id": "sec:verificationMethod", "@type": "@id"},
    "verifyData": "sec:verifyData",
    "wrappedKey": "sec:wrappedKey"
  }]
}
```

OR13: Which then loads v1:

```json
{
  "@context": {
    "id": "@id",
    "type": "@type",

    "dc": "http://purl.org/dc/terms/",
    "sec": "https://w3id.org/security#",
    "xsd": "http://www.w3.org/2001/XMLSchema#",

    "EcdsaKoblitzSignature2016": "sec:EcdsaKoblitzSignature2016",
    "Ed25519Signature2018": "sec:Ed25519Signature2018",
    "EncryptedMessage": "sec:EncryptedMessage",
    "GraphSignature2012": "sec:GraphSignature2012",
    "LinkedDataSignature2015": "sec:LinkedDataSignature2015",
    "LinkedDataSignature2016": "sec:LinkedDataSignature2016",
    "CryptographicKey": "sec:Key",

    "authenticationTag": "sec:authenticationTag",
    "canonicalizationAlgorithm": "sec:canonicalizationAlgorithm",
    "cipherAlgorithm": "sec:cipherAlgorithm",
    "cipherData": "sec:cipherData",
    "cipherKey": "sec:cipherKey",
    "created": {"@id": "dc:created", "@type": "xsd:dateTime"},
    "creator": {"@id": "dc:creator", "@type": "@id"},
    "digestAlgorithm": "sec:digestAlgorithm",
    "digestValue": "sec:digestValue",
    "domain": "sec:domain",
    "encryptionKey": "sec:encryptionKey",
    "expiration": {"@id": "sec:expiration", "@type": "xsd:dateTime"},
    "expires": {"@id": "sec:expiration", "@type": "xsd:dateTime"},
    "initializationVector": "sec:initializationVector",
    "iterationCount": "sec:iterationCount",
    "nonce": "sec:nonce",
    "normalizationAlgorithm": "sec:normalizationAlgorithm",
    "owner": {"@id": "sec:owner", "@type": "@id"},
    "password": "sec:password",
    "privateKey": {"@id": "sec:privateKey", "@type": "@id"},
    "privateKeyPem": "sec:privateKeyPem",
    "publicKey": {"@id": "sec:publicKey", "@type": "@id"},
    "publicKeyBase58": "sec:publicKeyBase58",
    "publicKeyPem": "sec:publicKeyPem",
    "publicKeyWif": "sec:publicKeyWif",
    "publicKeyService": {"@id": "sec:publicKeyService", "@type": "@id"},
    "revoked": {"@id": "sec:revoked", "@type": "xsd:dateTime"},
    "salt": "sec:salt",
    "signature": "sec:signature",
    "signatureAlgorithm": "sec:signingAlgorithm",
    "signatureValue": "sec:signatureValue"
  }
}
```

OR13: A better starting point would be:

```json
{
 "@context": {
    "@version": 1.1,
    "@vocab": "https://w3id.org/security#"
  }
}
```
Since you can see that all the security concerns are actually mapped to w3id.org... not w3c.

msporny: -1 for cascaded loading of security/v2 and security/v1 -- we want to get away from that in this iteration of the work.
-1 for use of @vocab for any JSON-LD context that touches security, we want to be specific about the terms we use, especially when those things are going to be digitally signed.

vocab/credentials/v2/credentials.ttl
```diff
@@ -0,0 +1,120 @@
+@prefix dc: <http://purl.org/dc/terms/> .
+@prefix owl: <http://www.w3.org/2002/07/owl#> .
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix cred: <https://w3.org/2018/credentials#> .
```
OR13, line 5: how come we don't see

```
https://w3id.org/security#
```

In this list....

vocab/credentials/v2/template.html
```diff
@@ -0,0 +1,241 @@
...
+        feature if possible.  For details, see
+        the <a href="https://github.com/w3c/vc-test-suite/">test suite</a> and
+        sample <a href="https://w3c.github.io/vc-test-suite/implementations/">implementation
+        report</a>.
+      </p> -->
+    </section>
+    <section>
+      <h2>Introduction</h2>
+      <p property="dc:description"><%= ont["dc:description"]["en"] %></p>
+      <p>This specification makes use of the following namespaces:</p>
```
OR13, line 141: I really hope we can simplify this list in v2.... especially removing odrl.

vocab/credentials/v2/credentials.jsonld
OR13: > Using `@vocab` causes the new "undefined terms detected" safety features of jsonld.js to not fire, and we definitely don't want to silence those safeties.

Interesting... does jsonld.js not have the ability to detect if `@vocab` is being used to expand a term? 

There are a lot of contexts out there that might include `@vocab`... sounds like it would be safer to allow for an error to be thrown when `@vocab` is used, rather than praying it won't be included in an open world data model... IIRC, the last context to invoke `@vocab` wins... and the entire extensibility model of VCDM 1.1 is based on adding to `@context`... see comment here: 

https://github.com/digitalbazaar/jsonld.js/pull/452#discussion_r840071361

vocab/credentials/v2/credentials.ttl
iherman: Because the vocabulary does not refer/use this namespace.

vocab/credentials/v2/credentials.jsonld
iherman: I think we are discussing the wrong thing here. _This vocabulary definition has nothing to do with the `@context` definition._ In fact, the jsonld version of the vocabulary should be cleaned up to _remove_ most of the context we are talking about. Just look at the Turtle file of the vocabulary; the JSON-LD version should not, I would even say, must not contain more information than that. The JSON-LD `@context` file is maintained elsewhere.

Sorry if I am preaching to the choir, but

1. The vocabulary "just" defines the terms that are defined/used in the cred namespace. It is all it has to do
2. The `@context` is a _mapping_ from a JSON-LD encoded RDF content to this (and other) namespaces.

In fact, I have, on my machine, a separate branch that vastly simplifies the JSON-LD version of the vocabulary to do just that. My intention was to add that as a separate PR once this one is settled; if you want, I can add a PR _against the branch of this PR_ to see if it makes things cleaner.

vocab/credentials/v2/credentials.ttl
iherman: (see also my comment in https://github.com/w3c/vc-data-model/pull/920#discussion_r960290781)

vocab/credentials/v2/template.html
iherman: Note that this PR has not yet looked at the v2 version of the `@context`. In fact, it is only an attempt to clean up the 1.1 vocabulary.

I agree that the ODRL reference may be an overkill.

vocab/credentials/v2/credentials.jsonld
msporny: > Interesting... does jsonld.js not have the ability to detect if `@vocab` is being used to expand a term?

At present, it can detect it (because, nothing would work if it couldn't)... but it doesn't have the option of throwing a warning or an error. It's a good suggestion for a new feature for jsonld.js -- some use cases will want to use `@vocab` (like when you're playing around with some ideas and don't want to define an entire JSON-LD context to do so), and some use cases will probably want to throw when `@vocab` is used to ensure that you're not magic'ing semantically undefined terms into existence.

/cc @dlongley @davidlehn

msporny (changes requested):
I believe the only blocking issue in this PR is the definition of `vocab/credentials/v2/credentials.jsonld` -- ideally, we should leave that (and all generated files), out of being checked in and perhaps generate them during the build process?

iherman:
> I believe the only blocking issue in this PR is the definition of `vocab/credentials/v2/credentials.jsonld` -- ideally, we should leave that (and all generated files), out of being checked in and perhaps generate them during the build process?

I am not sure what you say, @msporny. The `credentials.json` file _is_ generated by a ruby script starting by the vocabulary csv file (just as for the HTML and the Turtle files). The script itself is generating too much insofar as, originally, it seems to have generated a vocabulary file _and_ a `@context` file in one place. This became moot by the fact that the definition of the `@context` file is now done separately. 

As I said, I have created a separate branch (`simplify-json-ld`), branching off this current branch, to simplify this. I have commented out some stuff from the [ruby script](https://github.com/w3c/vc-data-model/blob/simplify-json-ld/vocab/credentials/v2/mk_vocab.rb) which now generates a "pure" [json-ld serialization of the vocabulary](https://github.com/w3c/vc-data-model/blob/simplify-json-ld/vocab/credentials/v2/credentials.jsonld). Personally, I would like to use this in the final version.

vocab/credentials/v2/credentials.jsonld
dlongley: @msporny,

> I think people are reacting to the fact that this PR re-defines the v2 JSON-LD Context...

Does it? I thought that was a totally different file that only the vocab used.

Re: throwing errors when using `@vocab` ...

Using `@vocab` is an explicit signal that the author does not want errors thrown for undefined terms, but instead wants them folded into some vocabulary without knowing what they are. So I don't expect `safe` mode to ever throw for those cases. Using `@vocab` is a sort of middle ground in the security space -- where unknown terms will be signed, but there will be no globally unambiguous definition for their meaning. The introduction of a new context could also "define them later" -- in ways that may contradict their previous (locally understood) meaning. This is the danger of using `@vocab`, but if that danger is understood, i.e., there is *zero* reliability with those terms at all, perhaps that's an acceptable compromise for allowing their use.

We should get a clear understanding for the potential advantage sought by using `@vocab` so it can be weighed transparently and appropriately against the disadvantages. Of course, if `@vocab` is *only* being used as a bad substitute for `safe` mode -- then it seems to me that `safe` mode is clearly better. I think there's more people are looking for though.

All that being said, the implementation of the new `safe` mode feature in jsonld.js (at least) actually uses a more comprehensive internal event system that is able to emit all kinds of events for debugging purposes (including ones around the use of `@vocab`). That feature hasn't been externalized yet with a clean API. But that feature will be a development / debugging tool as opposed to a data validation / protection feature like `safe` mode is.

iherman: > I think people are reacting to the fact that this PR re-defines the v2 JSON-LD Context in a way that 1) we don't want, and 2) is in conflict with [what's already agreed to as a baseline](https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v2) by the group:
>
> https://github.com/w3c/vc-data-model/pull/920/files#diff-0457f943f5c5375c7aa4592b374df978608840ae0f0d3bd13cf3df4d21dda2b1
>
> Everything else you said sounds fine, Ivan. :)

This PR does not re-defines the JSON-LD context (and it should not), it uses a (in my view, bloated) context within the vocabulary definition in JSON-LD.

See my comment elsewhere: https://github.com/w3c/vc-data-model/pull/920#issuecomment-1234357006. Should I take that branch and merge with this one, to get this issue completely out of the way?

Of course, once this PR is settled, we have to look at the differences between the vocabulary and the v2 context to bring them in line. That is the next step, for sure.

iherman: > > I think people are reacting to the fact that this PR re-defines the v2 JSON-LD Context...
> 
> Does it? I thought that was a totally different file that only the vocab used.

Exactly. And it uses it "internally", not for external usage.

msporny: Ah, I see now that the "JSON-LD Context" is just "the JSON-LD representation of the Vocabulary". I think both @OR13 and @mprorock were confused in the same way that I was.

iherman: > Perhaps if we rename these files to vocabulary.(html|ttl|jsonld) it would lead to less confusion? Granted, actually reading the entirety of the file might have done that, but I do remember scanning down quite a ways... but not far enough to get to the vocabulary definitions at the bottom.
> 
> You can ignore all of my suggestions (except the one to rename the files) at this point. :P

I am fine renaming the files; it is probably a good idea. I think I will also go ahead and simplify the json-ld file to the strict minimum to avoid any misunderstandings.

But... I have to call it a day, this will be done tomorrow...

iherman:
@msporny @OR13 @dlongley, I have made the changes, including renaming the files themselves. Forget everything that you saw before 😀; here are the three files that are generated by the ruby script:

- [vocabulary.html](https://raw.githack.com/w3c/vc-data-model/vocab-v2/vocab/credentials/v2/vocabulary.html)
- [vocabulary.ttl](https://github.com/w3c/vc-data-model/blob/vocab-v2/vocab/credentials/v2/vocabulary.ttl)
- [vocabulary.jsonld](https://github.com/w3c/vc-data-model/blob/vocab-v2/vocab/credentials/v2/vocabulary.jsonld)

I hope this is now good to go/merge.

cc @pchampin

msporny:
> I am not sure what you say, @msporny. The `credentials.json` file _is_ generated by a ruby script starting by the vocabulary csv file (just as for the HTML and the Turtle files).

What I'm saying is that if we have auto-generated files, we should use a Github process to auto-generate them instead of manually committing the files into Github. When we manually commit auto-generated files, there is a good chance that they will diverge from the underlying source files. It also unnecessarily pollutes the Github history with a bunch of check ins for auto-generated files. This doesn't have to be fixed in this PR, can be fixed in a future PR, and so we don't need to hold up the merge here to fix the longer-term issue.

.vscode/bookmarks.json
```diff
@@ -0,0 +1,39 @@
+{
+	"files": [
+		{
+			"path": "vocab/credentials/v2/mk_vocab.rb",
+			"bookmarks": [
```
msporny, line 5: We shouldn't commit .vscode files for personal developer bookmarks.

msporny (changes requested):
Only one minor nit for a file that was probably checked in by mistake. Otherwise, LGTM.

.vscode/bookmarks.json
iherman: Oops... I usually have this on .gitignore, but I missed it this time. Sorry about that.

msporny (approved):

msporny:

PR #922, opened 2022-08-29
modify editors list

### LINKED ISSUE #917: Condense editors list to set of current editors
I propose we condense the listed editors down to the set of current editors. Editors of previous versions are still listed under those versions.
If removal of former editors proves too painful for some, I suggest perhaps an "editors of previous versions" section.
We also have an acknowledgements section that could be suitable.

brentzundel:
closes #917 

Signed-off-by: Brent Zundel <brent.zundel@gmail.com>

***
<a href="https://pr-preview.s3.amazonaws.com/brentzundel/vc-data-model/pull/922.html" title="Last updated on Sep 9, 2022, 4:09 PM UTC (d072698)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/922/1789b95...brentzundel:d072698.html" title="Last updated on Sep 9, 2022, 4:09 PM UTC (d072698)">Diff</a>


OR13 (approved):

Sakurann (approved):

dlongley (approved):

selfissued (approved):

msporny (changes requested):
Removing these people from the Editors list will wipe these former editors from the ReSpec bibliography. They will no longer show up for their contributions to the specification, even though many of them made significant contributions to the specification that stand to this day.

I know nobody approving this PR meant it this way, but there is a word for this -- when a group of people are credited for work largely done by other people -- plagiarism. The new Editors (that have done a very small amount of work compared to the former Editors) will get their name on a specification where previous Editor's contributions far outweigh theirs. When we get to the end and re-calculate the contributions, we're going to end up in a situation where we have to put many of these names back, in the same order that they appear today.

This is a bad practice and must not be allowed. We removed Ian Hickson from the HTML5 specification in the same way, and I know that he disagreed strongly given that he wrote the entire document himself for years only to have his name wiped from the top of the spec, and then in the bibliography in the end.

If the problem we are attempting to solve here is the length of the Editor's list, we can fix that with a simple update to ReSpec, which I can volunteer to write (in time). The solution would list the first `n` Editors, and then have an "et. al" that can be expanded to the full list, if the reader wants to do so. That preserves all of the Editors in the cross-standards-body bibliography database while making the header a bit easier to read.

Strong -1 to this PR. We shouldn't transform recognition of the immense work done by previous Editors into a footnote and a removal from the SpecRef bibliography.

brentzundel:
The goal of this PR is to provide an indication to readers who the current set of editors are for V2.

Since we will do a calculation of editorial contribution before this specification becomes a recommendation, which will restore to Editorial status any previous editors whose contents remain in V2, I don't see how any irrevocable harm would be done by clarifying the current set of editors while V2 is an editors draft. Editors of previous versions are still editors of those versions. They may be listed (in the end) as editors of this version, because we take contributions into consideration for producing the final list, but at this point in time they are not editors.

Bad practice would be to eliminate mention of previous contributions and have no plan for re-evaluating according to contributions to the final published spec. This is not that.

kdenhartog:
In that case, I believe reSpec has a built in former editors attribute that I believe can be used. I originally was looking at that option when I started pulling my name off stuff, but couldn't quite figure it out for our tooling usage. That could be a useful way to address both the concerns. I'm a +1 imageof the et al suggestion Manu made as well for addressing the length of the list.

decentralgabe (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2022-09-07-vcwg#section1-2) on 2022-09-07

- no resolutions were taken

#### 1.2. modify editors list (pr vc-data-model#922)

_See github pull request [vc-data-model#922](https://github.com/w3c/vc-data-model/pull/922)._




**Manu Sporny:** 922 is a suggestion that we update the editors' list.  
… I have objected to the change because it removes people from the bibliography.  
… The editors list is really long now.  
… I suggest that we update respec to list the primary editors then et. al.  

**Brent Zundel:** I have no wish to deprive people of credit.  
… But our editors' list is ginormous. I propose move people out of the list who are not actively editing now.  
… Let's move them into a Former Editors section. Respec can do that.  

</details>

brentzundel:
I have modified the original changes this PR suggested. Rather than moving those names from the editors list and placing them in the acknowledgements section, they have been moved to a former editors section.

msporny (approved):
While I'm ok with this PR temporarily, it still wipes the former editors from the ReSpec global bibliography, which still results in the work of the former editors not being acknowledged. 

I'll try to chat w/ Marcos at W3C TPAC to see if we can fix ReSpec behavior once and for all, to make it so we can manually signal what the bibliography list should be.

I expect that the v2.0 REC will attribute the Editors based on merit (actual contributions). Approving and merging.

msporny:

Issue #913, opened 2022-08-18
`issuanceDate/expirationDate` and `validFrom/validUntil`


msporny:
We had added a warning to the VCDM spec that we were eventually going to deprecate `issuanceDate/expirationDate` and replace it with `validFrom/validUntil`.

https://www.w3.org/TR/vc-data-model/#issue-container-generatedID-13
https://www.w3.org/TR/vc-data-model/#issue-container-generatedID-15

The "date the credential was issued" is represented by the `created` property on the proof if using a `DataIntegrityProof` and `iat` if used in a JWT. The "validity period" for a VC used to be from `issuanceDate` to `expirationDate`, but then some reviewers confused those terms for "the date the credential was issued", so the group came to consensus on renaming these properties to `validFrom` and `validUntil` to make it more clear what was going on -- as well as moving away from the "Date" language since the value range was a date-time.

There was [an attempt at applying this change in the v2 context](https://github.com/w3c/vc-data-model/pull/905), but [there were objections](https://github.com/w3c/vc-data-model/pull/905#pullrequestreview-1076017558) to the deprecation of `issuanceDate` and `expirationDate`. 

The purpose of this issue is to determine if we are going to deprecate  `issuanceDate/expirationDate` and rename to `validFrom/validUntil` or if we are going to do something else.

selfissued:
As described in issued #918, I believe this would move us in the wrong direction - towards unnecessary and useless complexity.

awoie:
I would be ok with the following two options:

Option 1: keep a property in the VC data model that represents the time of issuance and expiration:
- deprecate `issuanceDate` since the semantical definition is a bit misleading.
- keep `expirationDate` since it is less controversial and it maps onto `exp` JWT claim.
- add `issued` that is semantically the same as the `iat` JWT claim.
- add `validFrom` that is semantically the same as `nbf` JWT claim.

Option 2: move issuance and expiration dates to Data Integrity and VC-JWT specs
- deprecate `issuanceDate`
- deprecate `expirationDate`
- define `iat` and `nbf` and `exp` in VC-JWT spec
- define `created` in Data Integrity spec (already commonly used)
- define `expirationDate` in Data Integrity spec (to my knowledge this is not always used)

I prefer Option 1 since I see both expiration and issuance part of the core VC Data Model. From an implementers perspective, it would be also easier since it is a common use case to understand when a VC was issued and when it does expire and it should work across proof formats.

Option 1:

```
[...]

PR #927, opened 2022-09-06
Vocabulary synced with v2 context

iherman:
The PR attempts to synchronize with the [V2 context](https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v2). The following changes have been made:

I Have added the properties which were missing from v2, namely:

- `holder`
- `issued`

(I presume the missing `holder` property was actually  a bug in the v1 version of the vocabulary). Note that the domain of `holder` is a bit more complicated because it can either be a Verifiable Credential or a Verifiable Presentation.

Some properties are not in the v2 context anymore. Because the decision is that the v2 version of the vocabulary, when "published", will reside on the same URL as v1, removing those properties from the vocabulary is not an option. What I propose is to formally "deprecate" those terms (and classes, if applicable) in the vocabulary; there is a well defined property, and corresponding classes, in the owl namespace exactly for that. I have therefore modified the underlying script and csv file to include the possibility for deprecation, and I have set as deprecated:

- issuanceDate
- expirationData

I was not sure whether the  `JsonSchemaValidator2018` and `ManualRefreshService2018` should also be deprecated or not; please advise.

I have also made some editorial changes on the HTML output, but those are secondary (but it explains some of the change reports in the diff file below).

For the ease of reviewing, see:

- [HTML version of the proposed V2](https://labs.w3.org/spec-generator/?type=respec&url=https://raw.githack.com/w3c/vc-data-model/vocab/sync-v2-context/vocab/credentials/v2/vocabulary.html)
- [Diff for the HTML Version](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Flabs.w3.org%2Fspec-generator%2F%3Ftype%3Drespec%26url%3Dhttps%3A%2F%2Fw3c.github.io%2Fvc-data-model%2Fvocab%2Fcredentials%2Fv2%2Fvocabulary.html&doc2=https%3A%2F%2Flabs.w3.org%2Fspec-generator%2F%3Ftype%3Drespec%26url%3Dhttps%3A%2F%2Fraw.githack.com%2Fw3c%2Fvc-data-model%2Fvocab%2Fsync-v2-context%2Fvocab%2Fcredentials%2Fv2%2Fvocabulary.html)
- [Turtle version of the proposed V2](https://github.com/w3c/vc-data-model/blob/vocab/sync-v2-context/vocab/credentials/v2/vocabulary.ttl)
- [JSON-LD version of the proposed V2](https://github.com/w3c/vc-data-model/blob/vocab/sync-v2-context/vocab/credentials/v2/vocabulary.jsonld)
- [The "controlling" CSV file for proposed V2](https://raw.githubusercontent.com/w3c/vc-data-model/vocab/sync-v2-context/vocab/credentials/v2/vocab.csv)

(Reviewers: any proposed change on the term descriptions, label, etc, should be made on the CSV file. Everything else is generated...)


iherman:
Note that there is a clear relationship to #913: it is still pending whether issuanceDate and expirationDate should be deprecated. Maybe it is better to hold on the deprecation of those until that issue is settled.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2022-09-07-vcwg#section1-4) on 2022-09-07

- no resolutions were taken

#### 1.4. Vocabulary synced with v2 context (pr vc-data-model#927)

_See github pull request [vc-data-model#927](https://github.com/w3c/vc-data-model/pull/927)._




**Brent Zundel:** Finally, the last one remaining open is 927. Ivan updated the vocabulary to match the VC context.  
… We need one more update of the PR by the editors.  

**Ivan Herman:** For 927, I have a question there that it would be good to have an answer to.  
… I don't know whether the JSON schema should be deprecated.  

**Manu Sporny:** It has been deprecated.  

**Ivan Herman:** After that, then we can merge.  

</details>

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2022-09-07-vcwg#section2-2) on 2022-09-07

- no resolutions were taken

#### 2.2. `issuanceDate/expirationDate` and `validFrom/validUntil` (issue vc-data-model#913)

_See github issue [vc-data-model#913](https://github.com/w3c/vc-data-model/issues/913)._




**Kristina Yasuda:** validFrom ... validUntil.  

**Michael Jones:** There's a discussion, which I understand predates my involvement about possibly deprecating `issuedDate` and replacing it with `validFrom` and possibly deprecating `expirationDate` and replacing it with `validUntil`.  

_See github pull request [vc-data-model#927](https://github.com/w3c/vc-data-model/pull/927)._




**Michael Jones:** There was a useful discussion on the issue about what use cases might exist for having a `validFrom` period that differs from the `issuanceDate`. In particular, the example was given of a future-dated coupon which starts being valid on maybe July 4th 2023 and is valid through July 4th 2024 but you could hand that out to people now. I understand that and it's a reasonable thing to do.  
… That said, I believe the predominant case is that stuff is valid when issued and that's what we should optimize for. That's the same that we did in JWTs.  
… There's typically always an issued at (iat) time and an expiration time. If you want it to be valid from a different time from the issuance then you add the optional claim to indicate that.  
… If it ever valuable to have a valid until that is different from the expiration date -- and no one has given a use case so I'd be happy to just leave `expirationDate` alone and not have that.  

**Manu Sporny:** There are a couple of things wrapped up in this issue.  
… Manu said that people thought that people couldn't include a time.  
… We put in a deprecation warning because of that.  
… We said that we were going to rename it.  
… The first issue is that we named these properties in a misleading way.  
… The second issue is that there's a strict difference between the validity period of a credential and when it was signed.  

> *Orie Steele:* +1 manu, we also want to separate "the data model" from "securing it".

**Manu Sporny:** As Mike said, when a coupon was signed and when it becomes valid are different.  
… Retailers need to distribute digital coupons before they come into play.  
… This is a real retail use case.  
… The other side of that is that the validity period of a credential can be different from the validity period of a digital signature.  
… A passport or driver's license has a validity period independent of the signature validity period.  
… A driver's license that has expired doesn't change a person's birthday.  
… We shouldn't conflate them.  

**David Chadwick:** A UK driver's license expires when you become 65.  
… Whereas the signature will expire well before that.  
… Some of these pertain to the claims and some pertain to the cryptography.  

**Orie Steele:** The validity period for a credential is different from the validity period for crypto.  

> *Orie Steele:* +1 DavidC, all cryptography has a shelf life that does not match always match the life of the claims the issuer is making.

**Michael Jones:** I was going to advertise Orie's presentation at TPAC, which is, I believe he's going to explicitly talk about clear semantic distinctions between properties of the Verifiable Credential and properties of the signatures used to secure them.  

> *Orie Steele:* I will attempt to cover this in my TPAC talk, I make not promises for success.

**Michael Jones:** I think some of this discussion would be well informed by a clear separation of those two concerns. There may be relationships between the two sets of dates but we should be clear about them.  

> *Manu Sporny:* we have been clear in the textual description, some developers don't have time to read specs. :).

**Michael Jones:** Also responding to a comment Manu made a few minutes ago -- the fact that it's called `issuanceDate` and that you couldn't put a time, is a documentation time. We should just be clear in the definition what it means. That's not sufficient for making a breaking change, we should just fix the documentation, not change the property name.  

**Ted Thibodeau Jr.:** It's not a documentation error. The documentation was quite clear.  
… The humans will still make the mistakes.  

> *Manu Sporny:* +1 to what TallTed is saying, we have to take the human element into consideration here.

**Ted Thibodeau Jr.:** Documentation does not fix issues like this.  

**Michael Jones:** If implementers aren't using our specifications, there's not much point in us writing them.  

**Ted Thibodeau Jr.:** Humans are fallible. They do not recall everything they read.  
… We are writing the spec for humans to implement them.  
… What we can fix, we should fix.  

> *Justin Richer:* Naming things is hard. Really hard.

**Ted Thibodeau Jr.:** We can fix things by using better names.  

**Kristina Yasuda:** It seems like people care a lot about this.  
… Please read the specs and suggest ways to make things clearer.  

</details>

iherman:
Based on the [meeting of 2022-09-07](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2022-09-07-vcwg) I have

- deprecated the `JsonSchemaValidator2018` and `ManualRefreshService2018` classes
- _not_ deprecated the `issuanceDate` and `expirationDate` properties; this is pending on the resolution of #913 

@msporny you could now merge this PR...

dlongley (approved):

vocab/credentials/v2/mk_vocab.rb
```diff
@@ -287,13 +296,13 @@ def to_ttl
...
     end
 
-    output << "\n# Instance definitions" unless @instances.empty?
-    @instances.each do |id, entry|
-      output << "cred:#{id} a #{namespaced(entry[:type])};"
-      output << %(  rdfs:label "#{entry[:label]}"@en;)
-      output << %(  rdfs:comment """#{entry[:comment]}"""@en;)
-      output << %(  rdfs:isDefinedBy cred: .)
-    end
+    # output << "\n# Instance definitions" unless @instances.empty?
```
decentralgabe, line 299: why comment these out but not remove?

iherman: My own caution... The background is that the ruby script is not of my making, I inherited it. Furthermore, I am not really familiar with ruby, so I have to be extremely careful about any modification I do on the code because I am driving in the dark. I.e., I put things in comments instead of boldly deleting it. It would be better if I re-wrote the whole script in a language that I am more familiar with, but I decided it is not worth the trouble.

I expect the v2 vocabulary to go through some more iteration in the coming period; once we get to a final point, it is my plan to clean the script of such impurities...

decentralgabe: fair enough!

decentralgabe (approved):

pchampin (approved):
a few minor comments, but overall LGTM

vocab/credentials/v2/mk_vocab.rb
```diff
@@ -121,6 +122,8 @@ def to_jsonld
         'rdfs:label' => {"en" => entry[:label].to_s},
         'rdfs:comment' => {"en" => entry[:comment].to_s},
       }
+      node['owl:deprecated'] = 'true' if entry[:deprecated]
+      node['@type'] = ['rdf:Property','owl:DeprecatedClass'] if entry[:deprecated]
```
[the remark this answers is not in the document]
iherman: Because yours truly is careless and made a copy/paste error...

vocab/credentials/v2/mk_vocab.rb
```diff
@@ -248,6 +253,8 @@ def to_ttl
     output << "\n# Class definitions" unless @classes.empty?
     @classes.each do |id, entry|
       output << "cred:#{id} a rdfs:Class;"
+      output << %(  a owl:DeprecatedClass;) if entry[:deprecated]
+      output << %(  owl:deprecated "true"^^xsd:boolean;) if entry[:deprecated]
```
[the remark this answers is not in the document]
iherman: Because... I forgot:-)

msporny:

Issue #901, opened 2022-07-31
Add DataIntegritySignature to v2 context?


msporny:
Now that the VCWG work is under way and the CCG Data Integrity specification has been selected as the input document for the Data Integrity work, it's time to start nailing down some implementation directions for the next 2+ years.

One of the first of those considerations is streamlining the way we build Data Integrity cryptosuites and include them in the base VC v2 context. Please find proposals linked to the end of this post that that:

* Ideally, eliminates the need to create new JSON-LD Contexts for cryptosuites for 80%+ of the use cases we currently have.
* Continues to ensure the ability to get large compression gains via the use of semantic compression (e.g., CBOR-LD)

The presentation deck containing the proposal is here: https://docs.google.com/presentation/d/16J9TcFc6fX18Cun_Rhj3xqNlnhkCs6MnPkGBlRL3YKg/edit

Archived copies of the proposal are here: https://lists.w3.org/Archives/Public/public-vc-wg/2022Jul/0044.html

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2022-08-03-vcwg#section6-10) on 2022-08-03

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 6.10. Add DataIntegritySignature to v2 context? (issue vc-data-model#901)

_See github issue [vc-data-model#901](https://github.com/w3c/vc-data-model/issues/901)._




**Brent Zundel:** 901, also got this one wrong?.  

**Manu Sporny:** You may be eventually correct though! There is a question about where the context lives... That's in the core repo for now..  

**Brent Zundel:** Removing that link... Last one.  

> *Manu Sporny:* +1 to move nonce to data-integrity :).

</details>

msporny:
This is being addressed by PR #943.

Sakurann:
done! was called `dataIntegrityProof`

TallTed:
Maybe worth changing issue title from `DataIntegritySignature` to match actual addition of `dataIntegrityProof`?

PR #943, opened 2022-10-09
Add DataIntegrityProof to credentials/v2 JSON-LD Context.

msporny:
This PR addresses issue #901 by adding DataIntegrityProof to the credentials/v2 JSON-LD Context. Here's the proposal for streamlining Data Integrity cryptosuites that was discussed at W3C TPAC:

https://docs.google.com/presentation/d/16J9TcFc6fX18Cun_Rhj3xqNlnhkCs6MnPkGBlRL3YKg/edit

This PR establishes the entirety of what we'd need to include in the credentials/v2 context to support the proposal above. This addition should cover around 80%-ish of the data integrity proof use cases we have and ensure that we don't have to keep adding date-stamped cryptosuites to the base VC context. It also makes life easier for developers using Data Integrity and VCs (as they don't have to specify separate JSON-LD Cryptosuite contexts).

Here's an open source library that implements a cryptosuite based on the new DataIntegrityProof type (with a README explaining how to use it):

https://github.com/digitalbazaar/eddsa-2022-cryptosuite

Here's a draft interoperability test suite demonstrating its implementability (for `eddsa-2022`):

https://w3c-ccg.github.io/di-eddsa-2022-test-suite/#conformance

The Data Integrity specification has defined this over the course of the last several weeks with reviews from multiple WG members, changes requested and made, and consensus achieved on the core concepts:

https://w3c.github.io/vc-data-integrity/#how-it-works

... and is defined normatively here:

https://w3c.github.io/vc-data-integrity/#proofs
https://w3c.github.io/vc-data-integrity/#dataintegrityproof

This PR is an alternate proposal to PR #924.

### DIFF
--- a/contexts/credentials/v2
+++ b/contexts/credentials/v2
@@ -78,13 +78,76 @@
         },
         "proof": {
           "@id": "https://w3id.org/security#proof",
-          "@type": "@id", 
+          "@type": "@id",
           "@container": "@graph"
         },
         "verifiableCredential": {
           "@id": "https://www.w3.org/2018/credentials#verifiableCredential",
           "@type": "@id",
           "@container": "@graph"
+        }
+      }
+    },
+
+    "DataIntegrityProof": {
+      "@id": "https://w3id.org/security#DataIntegrityProof",
+      "@context": {
+        "@protected": true,
+        "id": "@id",
+        "type": "@type",
+        "challenge": "https://w3id.org/security#challenge",
+        "created": {
+          "@id": "http://purl.org/dc/terms/created",
+          "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
+        },
+        "domain": "https://w3id.org/security#domain",
+        "expires": {
+          "@id": "https://w3id.org/security#expiration",
+          "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
+        },
+        "nonce": "https://w3id.org/security#nonce",
+        "proofPurpose": {
+          "@id": "https://w3id.org/security#proofPurpose",
+          "@type": "@vocab",
+          "@context": {
+            "@protected": true,
+            "id": "@id",
+            "type": "@type",
+            "assertionMethod": {
+              "@id": "https://w3id.org/security#assertionMethod",
+              "@type": "@id",
+              "@container": "@set"
+            },
+            "authentication": {
+              "@id": "https://w3id.org/security#authenticationMethod",
+              "@type": "@id",
+              "@container": "@set"
+            },
+            "capabilityInvocation": {
+              "@id": "https://w3id.org/security#capabilityInvocationMethod",
+              "@type": "@id",
+              "@container": "@set"
+            },
+            "capabilityDelegation": {
+              "@id": "https://w3id.org/security#capabilityDelegationMethod",
+              "@type": "@id",
+              "@container": "@set"
+            },
+            "keyAgreement": {
+              "@id": "https://w3id.org/security#keyAgreementMethod",
+              "@type": "@id",
+              "@container": "@set"
+            }
+          }
+        },
+        "cryptosuite": "https://w3id.org/security#cryptosuite",
+        "proofValue": {
+          "@id": "https://w3id.org/security#proofValue",
+          "@type": "https://w3id.org/security#multibase"
+        },
+        "verificationMethod": {
+          "@id": "https://w3id.org/security#verificationMethod",
+          "@type": "@id"
         }
       }
     }


dlongley (approved):

decentralgabe (approved):

peacekeeper (approved):

contexts/credentials/v2
```diff
@@ -87,6 +87,69 @@
           "@container": "@graph"
```
OR13, line 87: sigh.... this impacts a lot of how the data will be processed.

contexts/credentials/v2
```diff
@@ -87,6 +87,69 @@
...
+              "@id": "https://w3id.org/security#capabilityInvocationMethod",
+              "@type": "@id",
+              "@container": "@set"
+            },
+            "capabilityDelegation": {
+              "@id": "https://w3id.org/security#capabilityDelegationMethod",
+              "@type": "@id",
+              "@container": "@set"
+            },
+            "keyAgreement": {
```
OR13, line 136: Is encryption in scope for VPs?

contexts/credentials/v2
```diff
@@ -87,6 +87,69 @@
...
+              "@id": "https://w3id.org/security#authenticationMethod",
+              "@type": "@id",
+              "@container": "@set"
+            },
+            "capabilityInvocation": {
+              "@id": "https://w3id.org/security#capabilityInvocationMethod",
+              "@type": "@id",
+              "@container": "@set"
+            },
+            "capabilityDelegation": {
```
OR13, line 131: What does this have to do with credentials ?

contexts/credentials/v2
```diff
@@ -87,6 +87,69 @@
           "@container": "@graph"
         }
       }
+    },
+
+    "DataIntegrityProof": {
+      "@id": "https://w3id.org/security#DataIntegrityProof",
```
OR13, line 93: Link is dead: https://w3id.org/security#DataIntegrityProof

contexts/credentials/v2
```diff
@@ -87,6 +87,69 @@
...
+          }
+        },
+        "cryptosuite": "https://w3id.org/security#cryptosuite",
+        "proofValue": {
+          "@id": "https://w3id.org/security#proofValue",
+          "@type": "https://w3id.org/security#multibase"
+        },
+        "verificationMethod": {
+          "@id": "https://w3id.org/security#verificationMethod",
+          "@type": "@id"
```
OR13, line 150: Probably worth several calls on this topic.... what this means is likely not well understood, by the group.

mprorock: +1 - i would like clarity on this

contexts/credentials/v2
```diff
@@ -87,6 +87,69 @@
...
+            "capabilityInvocation": {
+              "@id": "https://w3id.org/security#capabilityInvocationMethod",
+              "@type": "@id",
+              "@container": "@set"
+            },
+            "capabilityDelegation": {
+              "@id": "https://w3id.org/security#capabilityDelegationMethod",
+              "@type": "@id",
+              "@container": "@set"
+            },
```
mprorock, line 135:
```suggestion
```

mprorock: let's keep this out for now - might make sense, but is a dedicated topic on its own

contexts/credentials/v2
mprorock: is there a corresponding PR on security context?

mprorock (changes requested):

contexts/credentials/v2
```diff
@@ -87,6 +87,69 @@
...
+            "@protected": true,
+            "id": "@id",
+            "type": "@type",
+            "assertionMethod": {
+              "@id": "https://w3id.org/security#assertionMethod",
+              "@type": "@id",
+              "@container": "@set"
+            },
+            "authentication": {
+              "@id": "https://w3id.org/security#authenticationMethod",
```
OR13, line 122: These are for VPs

contexts/credentials/v2
```diff
@@ -87,6 +87,69 @@
...
+        "nonce": "https://w3id.org/security#nonce",
+        "proofPurpose": {
+          "@id": "https://w3id.org/security#proofPurpose",
+          "@type": "@vocab",
+          "@context": {
+            "@protected": true,
+            "id": "@id",
+            "type": "@type",
+            "assertionMethod": {
+              "@id": "https://w3id.org/security#assertionMethod",
```
OR13, line 117: These are for VCs

contexts/credentials/v2
```diff
@@ -87,6 +87,69 @@
...
+        "created": {
+          "@id": "http://purl.org/dc/terms/created",
+          "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
+        },
+        "domain": "https://w3id.org/security#domain",
+        "expires": {
+          "@id": "https://w3id.org/security#expiration",
+          "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
+        },
+        "nonce": "https://w3id.org/security#nonce",
```
OR13, line 108: why we have this and challenge?... prefer 1.

dlongley: They have different semantics (challenge is never chosen by the signer and nonce usually is). That being said, I prefer just having `challenge` and moving any use of nonces that is specific to a particular type of crypto (e.g., BBS) into the `proofValue` via some internal encoding, just like any other crypto / proof-specific parameters or values should be.

In other words, `proofValue` should be self-contained and opaque to applications and only proof verifiers would decode it and break apart its constituent parts and do whatever is needed. This is comparable to a `jws` (that internally has parts like header and payload, etc.). That is where a signer-provided nonce belongs, IMO.

The only reason to expose any values via directly accessible `proof` properties (i.e., not encoded within `proofValue`) is if applications clearly have a need for them. Otherwise it just generates more churn for people writing generalized / layered code. If such a use can be argued for `nonce` then there's a reason to have it -- otherwise I don't think it's necessary.

contexts/credentials/v2
dlongley: The term definitions here are identical to the ones used in the data integrity context because they are intentionally `@protected` to simplify processing. Using different ones would be confusing and would raise a redefinition error if anyone combined the contexts. This is better for developers.

contexts/credentials/v2
dlongley: It means that the `verificationMethod` value will be interpreted as a URL.

contexts/credentials/v2
dlongley: I don't think that's relevant. This PR is about copying a protected term definition from the data integrity context into the credentials context for the convenience of developers. That means copying it so it is exactly the same. What that definition is / should / shouldn't contain is not relevant in this PR.

contexts/credentials/v2
dlongley: See: https://github.com/w3c/vc-data-model/pull/943#discussion_r993892902

dlongley:
Something important to clarify about this PR. This PR copies the term definition for `DataIntegrityProof` from the data integrity context to the credentials v2 context to ease the burden on developers. This PR is not about *what* that definition says. No matter what it says, it needs to be copied precisely or there will be processing problems. If you want to debate what goes into the definition, that should be done in data integrity space. Whatever the definition is -- it needs to be the same in both contexts.

contexts/credentials/v2
msporny: Not yet, no. I was going to do that if/when the Data Integrity JSON-LD Context was merged in the Data Integrity spec repo.

contexts/credentials/v2
```diff
@@ -87,6 +87,69 @@
...
       }
+    },
+
+    "DataIntegrityProof": {
+      "@id": "https://w3id.org/security#DataIntegrityProof",
+      "@context": {
+        "@protected": true,
+        "id": "@id",
+        "type": "@type",
+        "challenge": "https://w3id.org/security#challenge",
```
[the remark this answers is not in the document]
msporny: Yes, the security vocab is in serious need of some love right now. I expect we'll clean it up before or shortly after FPWD. We should certainly make sure that all the term entries exist in the security vocab before FPWD.

contexts/credentials/v2
msporny: If we remove this, we deviate from what the Data Integrity spec's main context. We should have this discussion over there instead of over here.

In the meantime, we might want to think about what it means to not include the non-VC-ecosystem related things in the vc/v2 context. These are values that we expect to exist in `proofPurpose`, which is a Data Integrity thing... I'm not sure mixing and matching is the best approach, so we'd need to explore what happens if we decide to start mixing and matching (and the ways that could all go horribly wrong).

contexts/credentials/v2
msporny: Yeah, I almost removed `nonce` but then thought that folks might complain. `nonce` is currently on the chopping block, we need to see if people see a need to keep it around.

iherman:
> Something important to clarify about this PR. This PR copies the term definition for `DataIntegrityProof` from the data integrity context to the credentials v2 context to ease the burden on developers. This PR is not about _what_ that definition says. No matter what it says, it needs to be copied precisely or there will be processing problems. If you want to debate what goes into the definition, that should be done in data integrity space. Whatever the definition is -- it needs to be the same in both contexts.

Big +1.

Put it another way: this PR (and a `@context` in general) does not _define_ a term; that is what the vocabulary specification is for. The `@context` defines a mapping of a plain text JSON key into a URL that really identifies that term. The mapping may be complex because it is, ahem, context dependent (the term 'context' being used in its general English sense), but that is all. As a consequence, it is perfectly all right if the same term appears in many different `@context` files.

contexts/credentials/v2
OR13: To be clear, this has to do with using this context for things that are not verifiable credentials correct?

Can we at least provide aninformative use case for how this might be related to verifiable credentials?

Some thing like, `capabilityDelegation` can be used to grant access to verifiable credentials stored on remote services?

contexts/credentials/v2
OR13: I think we can be clearer... it means the verificationMethod will be an IRI which can be dereferenced... does it mean it's allowed to be relative or absolute?

contexts/credentials/v2
OR13: Same comment as above, if we keep it in, we need to do a better job of explaining it.... that means adding informative text.

OR13 (changes requested):
In order for me to approve this:

1. All links must lead to term definitions
2. All term definitions must have some informative relationship to verifiable credentials or presentations.

mprorock (approved):
Adjusting - not a huge fan of dead links, but lets make sure we track the corrections and updates

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2022-10-19-vcwg#section2-1) on 2022-10-19

- no resolutions were taken

#### 2.1. Add DataIntegrityProof to credentials/v2 JSON-LD Context. (pr vc-data-model#943)

_See github pull request [vc-data-model#943](https://github.com/w3c/vc-data-model/pull/943)._




**Manu Sporny:** only one PR pending in VCDM.  
… pr#943.  
… we have a couple of objections which we are currently resolving.  
… orie's objection (human readable descriptions) will be added.  
… other one to removing from base @context is not easy to resolve.  

**Michael Prorock:** I appreciate the problem so I will withdraw the change requests.  

**Orie Steele:** dont like PRs to be made that don't lead to anywhere.  
… but don't have a blocking change.  

**Manu Sporny:** thank you guys.  

> *Ivan Herman:* See [Initial work for the vocabulary:](https://github.com/w3c/vc-data-integrity/pull/65).

**Ivan Herman:** there is quite a lot of work still to be done, so if we keep 943 open till resolved it might be quite a long time.  
… I am pushing people to help on vocabulary ASAP.  

</details>

msporny:
> 1. All links must lead to term definitions

This is being addressed in https://github.com/w3c/vc-data-integrity/pull/65.

> 2. All term definitions must have some informative relationship to verifiable credentials or presentations.

We do not have the ability to deviate from the DataIntegrityProof `@protected` definition [defined here](https://github.com/w3c/vc-data-integrity/blob/main/contexts/data-integrity/v1#L11) without creating an incompatibility if that context and the vc/v2 context were to be included in the same document. With respect to the "informative relationship", it is possible to describe and carry an invoked or delegated ZCAP, or a DID Document that has `capabilityDelegation` and `capabilityInvocation` verification relationships, but that language feels like it would be out of place in the VC Data Model given that we've decided to not talk about the specifics of "Securing Verifiable Credentials" in the VC Data Model. Given that the language describing the `DataIntegrityProof` cryptosuite and the verification relationships already exist in the Data Integrity spec, and that spec will be referenced from the vc-data-model spec in time, there is a normative linkage that will exist in time. 

I believe what both @mprorock and @OR13 said during last weeks call was that they were ok w/ the merge of this PR as long as we did item 1 above. I'm proceeding on those grounds and will merge this PR once https://github.com/w3c/vc-data-integrity/pull/65 is merged. Let me know if I've misread the resolutions on the call last week.

contexts/credentials/v2
msporny: There is now: https://github.com/w3c/vc-data-integrity/pull/65

contexts/credentials/v2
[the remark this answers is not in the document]
msporny: There is a PR to address this now: https://github.com/w3c/vc-data-integrity/pull/65

contexts/credentials/v2
msporny: @tplooker had some concerns around removing `nonce` -- thoughts, @tplooker ?

contexts/credentials/v2
msporny: Explanation and plan to move forward is here: https://github.com/w3c/vc-data-model/pull/943#issuecomment-1287491190

contexts/credentials/v2
msporny: Explanation and plan to move forward is here: https://github.com/w3c/vc-data-model/pull/943#issuecomment-1287491190

contexts/credentials/v2
msporny: Explanation and plan to move forward is here: https://github.com/w3c/vc-data-model/pull/943#issuecomment-1287491190

contexts/credentials/v2
msporny: It means it is a URL (relative or absolute). Relative URLs will be resolved relative to the document base, which can be set to any URL via `@base`. This is described in: https://www.w3.org/TR/json-ld11/#base-iri

msporny:
>> 1. All links must lead to term definitions
> 
> This is being addressed in [w3c/vc-data-integrity#65](https://github.com/w3c/vc-data-integrity/pull/65).

https://github.com/w3c/vc-data-integrity/pull/65 has now been merged. 

https://w3id.org/security/ now points to the VCWG-maintained security vocabulary.

The missing links to https://w3id.org/security#DataIntegrityProof and https://w3id.org/security#cryptosuite now exist.

This addresses all known dead links. If there are more, it's now easy to fix and update them.

msporny:

TallTed:
Fixed #901

PR #970, opened 2022-11-02
fix: fixes #762

### LINKED ISSUE #762: Can a credentialSubject be only a string value?
All examples within the VC test suite and in the examples within the specification show the credentialSubject as an object. However there is no normative language which specifies that this property MUST be an object. Is this a case where a normative statement was missed?

The reason for asking is because we've found an edge case where a frame is modifying the credential subject from an object to a string with the value being the value of the `@id` property and we're uncertain what the most interoperable path would be for this.

Checking with editors here: @msporny @dlongley @brentzundel @burnburn

awoie:
Forbids `credentialSubject` represented as JSON-LD string values. I found it more appropriate to add this language to the JSON-LD section since it is JSON-LD related.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1502,10 +1502,9 @@ A <a>verifiable credential</a> MUST have a <code>credentialSubject</code>
           <dt>credentialSubject</dt>
           <dd>
 The value of the <code>credentialSubject</code> <a>property</a> is defined as
-a set of objects that contain one or more properties that are each related to a
+a set of objects that MUST contain one or more properties that are each related to a
 <a>subject</a> of the <a>verifiable credential</a>. Each object MAY contain an
-<code>id</code>, as described in Section <a href="#identifiers"></a>. Furthermore,
-each object MUST contain at least one or more <a>claims</a> about the subject.
+<code>id</code>, as described in Section <a href="#identifiers"></a>.
           </dd>
         </dl>
         
--- a/index.html
+++ b/index.html
@@ -1502,7 +1502,7 @@ A <a>verifiable credential</a> MUST have a <code>credentialSubject</code>
           <dt>credentialSubject</dt>
           <dd>
 The value of the <code>credentialSubject</code> <a>property</a> is defined as
-a set of objects that MUST contain one or more properties that are each related to a
+a set of objects that contain one or more properties that are each related to a
 <a>subject</a> of the <a>verifiable credential</a>. Each object MAY contain an
 <code>id</code>, as described in Section <a href="#identifiers"></a>.
           </dd>
@@ -3456,6 +3456,11 @@ that as long as the same <code>@context</code> declaration is made at the top of
 a <a>verifiable credential</a> or <a>verifiable presentation</a>,
 interoperability is guaranteed for all terms understood by users of the data
 model whether or not they use a [[!JSON-LD]] processor.
+            </li>
+            <li>
+A <code>credentialSubject</code> MUST NOT be represented as a single string
+value. This includes cases where the <code>credentialSubject</code> contains
+only an <code>id</code> property.
             </li>
           </ul>
         </section>
--- a/index.html
+++ b/index.html
@@ -1504,10 +1504,11 @@ A <a>verifiable credential</a> MUST have a <code>credentialSubject</code>
 The value of the <code>credentialSubject</code> <a>property</a> is defined as
 a set of objects that contain one or more properties that are each related to a
 <a>subject</a> of the <a>verifiable credential</a>. Each object MAY contain an
-<code>id</code>, as described in Section <a href="#identifiers"></a>.
+<code>id</code>, as described in Section <a href="#identifiers"></a>. Furthermore,
+each object MUST contain at least one or more <a>claims</a> about the subject.
           </dd>
         </dl>
-
+        
         <pre class="example nohighlight vc"
           title="Usage of the credentialSubject property"
           data-vc-vm="https://example.edu/issuers/565049#key-1">


OR13:
its not related to JSON-LD... its the core data model.

mprorock:
> its not related to JSON-LD... its the core data model.

+1 - this should get moved up to the core data model section, otherwise great

awoie:
I changed the PR, now the language makes "one or more properties" mandatory which implies it cannot be a string value.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2022-11-02-vcwg#section1-3) on 2022-11-02

- no resolutions were taken

#### 1.3. Forbids credentialSubject objects represented as string values, also fixes #762 (pr vc-data-model#970)

_See github pull request [vc-data-model#970](https://github.com/w3c/vc-data-model/pull/970)._




**Manu Sporny:** next item up is PR 970.  
… oliver raised a fix to forbid credentialSubjects being string values.  

**Oliver Terbu:** I put in JSON-LD section because of specific reasons related to JSON-LD, representing an object that has `id` property as single string that's specific to JSON-LD. Thought it was more appropriate to JSON-LD section, could also put it in credentialSubject section. Why is it needed there, though? One or more properties that are related to subject of credential..  
… It felt a little bit redundant in data model section..  

**Orie Steele:** It's an improvement, but it is a core data model thing..  
… In both representations we don't want value part to be a string. I agree with comment on credential subject phrasing being weird, suggest we not conflate adding this requirement to cleaning up that section, but perhaps we can add an issue to revise that section in its entirety, that section is confusing to read. File issue to update credential subject and move to accept PR as is..  

> *Orie Steele:* I can implement "sets" with strings.....

**Oliver Terbu:** The spec currently says value of credential subject is defined as object that is related to subject of credential -- that's not normative text at the moment, but it could say MUST say, would that fix that issue?.  

**Manu Sporny:** that's it for 970, please provide feedback.  
… moving on to data integrity.  

</details>

TallTed:
> [@awoie] implies it cannot be a string value

I don't see any such implication in the revised text.

iherman:
At the moment [credentialSubject](https://w3c.github.io/vc-data-model/vocab/credentials/v2/vocabulary.html#credentialSubject) has no restriction as to its range. If the data model changes to be in line with the PR proposal, then there should be a restriction whereby the property must be an "object property", i.e., the value must either be a blank node (which corresponds to a map in the JSON version) or an IRI.

It is one line to be added to the YAML definition of the property:

```
range : IRI
```

David-Chadwick:
+1 to this change

msporny (approved):

awoie:
@TallTed Can you give an example of a `credentialSubject` that is "an object with one or more properties that are each related to a [subject](https://www.w3.org/TR/vc-data-model/#dfn-subjects)"? IMO, this includes ids and claims. The PR makes this normative and mandatory. The emphasis is on "related to a subject". We could be more precise and say that it must contain one or more claims.

awoie:
> @TallTed Can you give an example of a `credentialSubject` that is "an object with one or more properties that are each related to a [subject](https://www.w3.org/TR/vc-data-model/#dfn-subjects)"

...  **where the `credentialSubject` is a string**.

iherman:
@msporny before you merge this: w.r.t https://github.com/w3c/vc-data-model/pull/970#issuecomment-1308461052 should I then issue a separate PR for the change in the vocabulary, or should we ask @awoie to do that as part of this PR?

msporny (changes requested):
@TallTed wrote:
> The emphasis is on "related to a subject". We could be more precise and say that it must contain one or more claims and/or an id.

I'd prefer we do this. While it's valid, per the data model, to have a bunch of objects containing only IDs, doing that is effectively a "null statement". All you're saying is "there exists a credentialSubject with this ID"... don't know if that's useful.

If we, instead say that it MUST be a set of objects that contain one or more claims, then I think we address the initial issue raised by making it crystal clear that "a string", "a number", "a null value", etc. are not valid entries for `credentialSubject`.

TallTed:
@msporny -- In https://github.com/w3c/vc-data-model/pull/970#pullrequestreview-1175838159, please s/`@TallTed wrote:`/`@awoie wrote:`/  :-)

Otherwise, I think I agree with your suggested revision, i.e., that **_`credentialSubject` MUST be a set of objects that contain one or more claims._**

index.html
```diff
@@ -1502,7 +1502,7 @@ <h3>Credential Subject</h3>
           <dt>credentialSubject</dt>
           <dd>
 The value of the <code>credentialSubject</code> <a>property</a> is defined as
-a set of objects that contain one or more properties that are each related to a
+a set of objects that MUST contain one or more properties that are each related to a
```
msporny, line 1505:
```suggestion
a set of objects that MUST contain one or more claims that are each related to a
```

msporny: @awoie @TallTed Does the language above work for both of you?

awoie:
I committed @msporny suggestions.

awoie:
I believe the latest changes are also fine with @TallTed now, so we can merge in the WG meeting if nobody objects.

msporny (approved):

msporny:
> I believe the latest changes are also fine with @TallTed now, so we can merge in the WG meeting if nobody objects.

We can merge now, it's been 19 days, you've addressed all the feedback AFAICT, there have been multiple reviews, and there are no standing objections. Thank you for the PR @awoie ! :)

msporny:

PR #977, opened 2022-11-12
Add JSON Schema for Verifiable Credentials

decentralgabe:
- Added a json schema for the VCDM (not VPs yet...) using the latest draft
- Added a test against a select number of existing examples

TODO:
- Add a schema for VPs
- Add language to the spec around how this schema can be used (or should this be in another PR?)

Questions/callouts:
- I wasn't certain how or if I should restrict formatting (e.g. for date-time values, I do not know if the `date-time` schema value is acceptable so I avoided it).
- More generally not having all types strictly validated could give a false sense of security, so should have language about that when added to the spec
- I included a schema for "proof" as an option. I'm wondering if I should keep this or not. Notably, there's an error with some examples not having a `proofValue` field which is required in the latest Data Integrity spec, so I made this field not required for now.
- Are all `type` properties either a string or an array of strings? I made this the case...but uncertain

Welcoming to all feedback and revisions.


schema/verifiable-credential-schema.json
```diff
@@ -0,0 +1,244 @@
+{
+  "$id": "https://www.w3.org/2022/credentials/v2/json-schema.json",
```
decentralgabe, line 2: uncertain what to put here

iherman:
@decentralgabe, thanks for having started this!

I think this is a perfect occasion to sync up the various places where vocabularies are defined. After all, there should be a 100% sync among the following documents:

- [Data model specification](https://www.w3.org/TR/vc-data-model-2.0)
- [Data model vocabulary (V2)](https://w3c.github.io/vc-data-model/vocab/credentials/v2/vocabulary.html) 
- [Data Integrity specification](https://www.w3.org/TR/vc-data-integrity/)
- [Data Integrity vocabulary](https://w3c.github.io/vc-data-integrity/vocab/security/vocabulary.html) (knowing that this one is still in a messy state; working on it...)
- [The json schema in this PR](https://github.com/w3c/vc-data-model/blob/add-json-schemas/schema/verifiable-credential-schema.json)

I think that we (meaning you, me for the formal vocabularies, and the editors of the specifications) should work on this sync. Here are the discrepancies that I see right now:

- The schema does not include the following properties:
  - `holder` and `verifiableCredential`: these are VP properties, so these may come later if I understand right.
  - `issued`: is already in the [Data model vocabulary (V2)](https://w3c.github.io/vc-data-model/vocab/credentials/v2/vocabulary.html) but only [mentioned in a spec note](https://www.w3.org/TR/vc-data-model-2.0/#issuance-date); not sure what its fate is
  - `validFrom` and `validUntil`: has been added to the [Data model vocabulary (V2)](https://w3c.github.io/vc-data-model/vocab/credentials/v2/vocabulary.html) but not yet into the specification; it was the subject of a PR that got in conflict with another one and should come soon now that the conflicts are solved
  - `serviceEndpoint`: in the [Data model vocabulary (V2)](https://w3c.github.io/vc-data-model/vocab/credentials/v2/vocabulary.html), but not mentioned in the [Data model specification](https://www.w3.org/TR/vc-data-model-2.0). Should it be removed from the vocabulary?
  - `proofChain`: is listed in the [Data Integrity specification](https://www.w3.org/TR/vc-data-integrity/) specification (not yet in the [Data Integrity vocabulary](https://w3c.github.io/vc-data-integrity/vocab/security/vocabulary.html)). It has a similar structure than `proof`
  - `challenge`: is listed in the [Data Integrity specification](https://www.w3.org/TR/vc-data-integrity/) specification
- In par with the introduction of `validFrom` and `validUntil` the `expirationDate` and `issuanceDate` will be taken out of the spec. In the [Data model vocabulary (V2)](https://w3c.github.io/vc-data-model/vocab/credentials/v2/vocabulary.html) these are still listed as deprecated properties, and we have to decide whether they should be removed from the schema. Personally, I would propose to remove them; after all, `"additionalProperties":true` means that older VC-s will pass the schema, so there is no problem with backward compatibility.
- The value of `verificationMethod` also has some restrictions, per [spec](https://www.w3.org/TR/vc-data-integrity/#verification-methods). I would think it is worth folding those into the schema, too.

I may have missed some more, and may have made a mistake in the list above, but I think this is a perfect occasion to get things right... In any case, once we have these discrepancies solved, we should all do the changes in the documents we are "responsible" of, ie, the specs, the vocabulary, and the schema...

cc @msporny @OR13 @awoie @selfissued @mprorock @dlongley

iherman:
There are a number of places where the spec specifies that the value of the property must be a URL (e.g., `id`) or when the value is either a URL or an object (e.g., `@context` or `credentialSubject`). I think it would be good to add that to the schema, playing with `"oneOf"` and the `"uri"` data format. 

Similarly, a number of properties define their values as dates and we may think about using the `"date-time"` format (but I am not 100% sure that the schema specification of date-time is identical its XMLSCHEMA11-2 equivalent).

schema/verifiable-credential-schema.json
msporny: This is fine for now, the WG needs to decide what they want the new base URL to be.

msporny (approved):
Aside from the comments that @iherman has made, this PR is better than what we have in the specification right now, so we should merge it and improve it in future PRs.

schema/test/test.js
```diff
@@ -0,0 +1,93 @@
+const fs = require('fs');
+const assert = require('assert');
+const Ajv2020 = require("ajv/dist/2020")
+const ajv = new Ajv2020()
+
+const verifiableCredentialSchema = "/../verifiable-credential-schema.json"
```
dlongley, line 6: This file has a mix of semi-colon / non-semi-colon lines ... it should be normalized to use semi-colons.

dlongley (approved):
+1 to Ivan's comments on improvements to be made, but think we can wait as Manu said.

msporny:
@decentralgabe excellent work, thank you for this PR!

> * Added a test against a select number of existing examples

I'll note that the test suite is hard coded to specific examples, which is fine for now. The problem this creates is that once the examples are updated, they might get out of sync w/ the test suite. In time, we might want 1) a test suite that tests every file in a particular directory, and 2) an extension to ReSpec that pulls the examples in from disk, or pulls the examples from the ReSpec file (so things don't drift).

> * Add a schema for VPs

+1

> * Add language to the spec around how this schema can be used (or should this be in another PR?)

Do this in another PR. This PR is already big enough, don't overload it with other things that might cause people to object. We can raise issues for future work that we intend to do (like spec language on how the schema can be used).

> * I wasn't certain how or if I should restrict formatting (e.g. for date-time values, I do not know if the `date-time` schema value is acceptable so I avoided it).

We should probably put in a regex for XML Schema 1.1 date-times... we found an issue recently during JFF Plugfest #2 where one implementer was not putting in any timezone offset, and when you don't do that, some implementations assume UTC and others assume local timezone. Clearly, that will lead to interop issues, so we might want to make timezone specification mandatory (even though XML Schema 1.1 says it optional). All that is pointing to specifying a date-time regex.

> * More generally not having all types strictly validated could give a false sense of security, so should have language about that when added to the spec

Yes, in fact, (and I know this is currently under debate), because we have decentralized extensibility, JSON Schema is better than nothing, but it will also give a false sense of security for VCs that don't have a JSON Schema associated with them (or with developers that provide lax JSON Schema files. That is, there will always be a false sense of security... JSON-LD and JSON Schema is a "gut check", but can never be fully relied upon for correctness.

> * I included a schema for "proof" as an option. I'm wondering if I should keep this or not. Notably, there's an error with some examples not having a `proofValue` field which is required in the latest Data Integrity spec, so I made this field not required for now.

Yes, we should have a schema for proof, and you made the right call wrt. `proofValue` not all Data Integrity cryptosuites use it. We're trying to ratchet down optionality with Data Integrity, but given the decentralized extensibility nature of that technology, we can't fully outlaw certain combinations. What we can do is create a JSON Schema for `DataIntegrityProof` specifically -- that requires `proofValue` and is hopefully what most future cryptosuites will start converging towards.

> * Are all `type` properties either a string or an array of strings? I made this the case...but uncertain

That's still under debate. Per JSON-LD, you can have full URL values. We might want to restrict that in VCs to only be a string or an array of strings (that are URLs). 

> Welcoming to all feedback and revisions.

Again, great work @decentralgabe -- looking forward to this making it into the repository (and used in the new VC test suite).

iherman:
The `proof` (and, obviously, `proofChain`) properties values can be an object or an array of objects. The shema, right now, only accepts the former.

schema/test/test.js
decentralgabe: thanks I try to avoid JS at all costs so consider me a noob

schema/verifiable-credential-schema.json
```diff
@@ -0,0 +1,244 @@
...
+        "type": {
+          "$ref": "#/$defs/type"
+        }
+      },
+      "required": [
+        "type"
+      ],
+      "additionalProperties": true
+    },
+    "proof": {
```
dlongley, line 73: Ivan is right -- this needs to allow for an array of objects (with the same constraints) as well.

decentralgabe:
@iherman a response to some of your points...

I [opened up a new issue](https://github.com/w3c/vc-data-model/issues/979) to track the discrepancies you mentioned.

> * The schema does not include the following properties:  
>   * `holder` and `verifiableCredential`: these are VP properties, so these may come later if I understand right.
I was planning on making VPs a separate schema, or do you think they should be a part of the VC one too?

>   * `issued`: is already in the [Data model vocabulary (V2)](https://w3c.github.io/vc-data-model/vocab/credentials/v2/vocabulary.html) but only [mentioned in a spec note](https://www.w3.org/TR/vc-data-model-2.0/#issuance-date); not sure what its fate is

Going to leave this out until there's some clarity on the fate. I'll note this in the [issue above](https://github.com/w3c/vc-data-model/issues/979)

>   * `validFrom` and `validUntil`: has been added to the [Data model vocabulary (V2)](https://w3c.github.io/vc-data-model/vocab/credentials/v2/vocabulary.html) but not yet into the specification; it was the subject of a PR that got in conflict with another one and should come soon now that the conflicts are solved

I will keep an eye out for this PR. Perhaps the submitter can update the schemas too.

>   * `serviceEndpoint`: in the [Data model vocabulary (V2)](https://w3c.github.io/vc-data-model/vocab/credentials/v2/vocabulary.html), but not mentioned in the [Data model specification](https://www.w3.org/TR/vc-data-model-2.0). Should it be removed from the vocabulary?

Going to leave this out and note this in the [issue above](https://github.com/w3c/vc-data-model/issues/979).

>   * `proofChain`: is listed in the [Data Integrity specification](https://www.w3.org/TR/vc-data-integrity/) specification (not yet in the [Data Integrity vocabulary](https://w3c.github.io/vc-data-integrity/vocab/security/vocabulary.html)). It has a similar structure than `proof`

Added.

>   * `challenge`: is listed in the [Data Integrity specification](https://www.w3.org/TR/vc-data-integrity/) specification

Added.

> * In par with the introduction of `validFrom` and `validUntil` the `expirationDate` and `issuanceDate` will be taken out of the spec. In the [Data model vocabulary (V2)](https://w3c.github.io/vc-data-model/vocab/credentials/v2/vocabulary.html) these are still listed as deprecated properties, and we have to decide whether they should be removed from the schema. Personally, I would propose to remove them; after all, `"additionalProperties":true` means that older VC-s will pass the schema, so there is no problem with backward compatibility.

Noting this in the [new issue](https://github.com/w3c/vc-data-model/issues/979).

> * The value of `verificationMethod` also has some restrictions, per [spec](https://www.w3.org/TR/vc-data-integrity/#verification-methods). I would think it is worth folding those into the schema, too.

Every example in the VCDM spec uses `verificationMethod` a single string value, which seems to conflict with how the property is defined in the Data Integrity spec. I will switch this to an array for now and note it in the new issue. Similarly the `type` property in the `verificationMethod`'s definition says it must be a string value, not an array of strings.

decentralgabe:
Addressed some feedback (noted above), added VP schema and tests.
Think this is ready to merge. Will address spec language and other issues in follow-up PRs.

One callout. The VP schema relies upon a reference to the VC schema. I'm not sure how to get a stable identifier here so folks can start using the schemas.

schema/README.md
```diff
@@ -0,0 +1,25 @@
...
+# Schemas
+
+There are two schemas provided, each with associated tests:
+
+- [Verifiable Credential JSON Schema](verifiable-credential/verifiable-credential-schema.json)
+- [Verifiable Presentation JSON Schema](verifiable-presentation/verifiable-presentation-schema.json)
+
+# Tests
+
+A limited set of tests are included against examples in the specification. To run these tests you can use the following command, after `npm i`:
```
TallTed, line 19: How many times must `npm i` be run?  Before each `npm run test-schemas`, once per log in session, once ever?  The answer to this will guide the appropriate rephrasing.
```suggestion
A limited set of tests against examples in the specification are included. To run these tests, you can use the following command, after `npm i`:
```

schema/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+```bash
+npm run test-schemas
+```
+
+The tests rely on examples featured in the spec. Currently these examples are copy-pasted and there is no guarantee there are in sync with the spec. This should be improved!
```
TallTed, line 25:
```suggestion
The tests rely on examples featured in the spec. Currently these examples have been copy-pasted, and there is no guarantee they are in sync with the spec. This should be improved!
```

schema/verifiable-presentation/verifiable-presentation-schema-test.js
```diff
@@ -0,0 +1,45 @@
...
+      const data = JSON.parse(fs.readFileSync(__dirname + '/example-2.json'));
+      const validate = ajv.compile(schema);
+      const valid = validate(data);
+
+      assert.equal(null, validate.errors);
+      assert.equal(true, valid);
+    });
+
+    // Note: this example fails because the dependent VC is not valid:
+    // It is missing an issuanceDate. Both the VP and VC proofs are missing
```
TallTed, line 21:
```suggestion
    // Note: this example fails because the dependent VC is not valid;
    // it is missing an issuanceDate. Both the VP and VC proofs are missing
```

schema/README.md
decentralgabe: once ever

sbutterfield (approved):

Issue #862, opened 2022-01-19
[Tracking Issue - Proposed Corrections Feedback] URL to URI


brentzundel:
@jyasskin I've raised this issue to track the WG response to the feedback you provided above and to provide a place where we might engage in a conversation as we explore possible responses.

From [feedback on VC Data Model v1.1](https://lists.w3.org/Archives/Public/public-new-work/2022Jan/0002.html)
``` 
* Correction 2, changing "URL" to "URI", moves from a value that can be
fetched to a value that can't necessarily be fetched. The meaning of these
`id` values is undefined both before and after the change, but it moves
from the possibility of defining a general meaning for the fetched
resource, to a need to define the meaning of particular URIs in a table.
That seems like the wrong direction. The WG could cite
https://url.spec.whatwg.org/ if their goal is simply to provide a normative
definition of the "URL" term.
```

The WG also experienced some frustration in attempting to package the proposed corrections in such a way as to make them more accessible for reviewers. We would have definitely benefited from some additional tooling.
The following contain records of the issue and the WG's conversation which led to the corrections related to the above concerns:
- The issue that drove this correction is #748 
- The PR that proposed the changes is #819

jyasskin:
Thanks! @dlongley's comment in https://github.com/w3c/vc-data-model/pull/819#issuecomment-920144413 seems to match my concern here, and provide a plausible fix by adding

> there's an expectation that machine readable information needs to be retrievable [] from the URI

I think "via some other field value" would re-introduce a similar concern around how implementations could figure out what other field value to retrieve.

kdenhartog:
Added the errata label to track this in the [errata document](https://w3c.github.io/vc-data-model/errata.html). I think it could just as well be removed based on the tracking of the additional linked issues and PRs as well. @brentzundel feel free to remove it if you think it's not needed.

brentzundel:
Might as well track it, thanks Kyle.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2022-01-26-vcwg#section4-2) on 2022-01-26

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 4.2. [Tracking Issue - Proposed Corrections Feedback] URL to URI (issue vc-data-model#862)

_See github issue [vc-data-model#862](https://github.com/w3c/vc-data-model/issues/862)._




[...]

PR #994, opened 2022-12-11
Change reference to URI/IRI to use WHATWG URL specification.

msporny:
This PR addresses issue #862 by referencing the WHATWG URL specification instead of the URI/IRI specification from IETF. An explanation as to why we're making this change is included in the PR.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/994.html" title="Last updated on Jan 7, 2023, 5:23 PM UTC (eb0f404)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/994/943d4a6...eb0f404.html" title="Last updated on Jan 7, 2023, 5:23 PM UTC (eb0f404)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -119,7 +119,7 @@
             w3cid: 46156
           }
         ],
-
+        xref: ["url"],
         maxTocLevel: 2,
         inlineCSS: true
       };
@@ -1193,7 +1193,7 @@ The <code>id</code> <a>property</a> MUST NOT have more than one value.
           </li>
           <li>
 The value of the <code>id</code> <a>property</a> MUST be a <a>URL</a> which
-MAY be dereferenceable.
+MAY be dereferenced.
           </li>
         </ul>
 
@@ -1808,7 +1808,7 @@ include the following:
             <ul>
               <li>
 <code>id</code> <a>property</a>, which MUST be a <a>URL</a> which MAY be
-dereferenceable.
+dereferenced.
               </li>
               <li>
 <code>type</code> <a>property</a>, which expresses the <a>credential</a> status
@@ -2352,10 +2352,10 @@ specifically not the domain of this specification.
 
         <p>
 Developers are urged to ensure that extension JSON-LD contexts are highly
-available. Implementations that cannot fetch a context will produce an error.
-Strategies for ensuring that extension JSON-LD contexts are always available
-include using content-addressed URLs for contexts, bundling context documents
-with implementations, or enabling aggressive caching of contexts.
+available. Implementations that cannot dereference a context will produce an
+error. Strategies for ensuring that extension JSON-LD contexts are always
+available include using content-addressed URLs for contexts, bundling context
+documents with implementations, or enabling aggressive caching of contexts.
         </p>
 
         <p>
@@ -5220,7 +5220,7 @@ libraries.
         </li>
 
         <li>
-Loosen the requirement to allow non-dereferenceable <a>URLs</a> in the
+Loosen the requirement to allow <a>URLs</a> that cannot be dereferenced in the
 <code>id</code> property of the <code>credentialStatus</code> and
 <code>refreshService</code> sections of the data model.
         </li>
--- a/index.html
+++ b/index.html
@@ -1065,11 +1065,11 @@ mean the same thing to each other. This might be referred to as
         </p>
         <p>
 <a>Verifiable credentials</a> and <a>verifiable presentations</a> have many
-attributes and values that are identified by <a>URIs</a> [[RFC3986]]. However,
-those <a>URIs</a> can be long and not very human-friendly. In such cases,
+attributes and values that are identified by <a>URLs</a> [[URL]]. However,
+those <a>URLs</a> can be long and not very human-friendly. In such cases,
 short-form human-friendly aliases can be more helpful. This specification uses
 the <code>@context</code> <a>property</a> to map such short-form aliases to the
-<a>URIs</a> required by specific <a>verifiable credentials</a> and <a>verifiable
+<a>URLs</a> required by specific <a>verifiable credentials</a> and <a>verifiable
 presentations</a>.
        </p>
   <p class="note">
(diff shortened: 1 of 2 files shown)


TallTed (approved):
WHAT WG's writing notwithstanding, I'm not convinced that using URL everywhere (including "nondereferenceable URL" and/or "nondereferencable URL", of which I prefer the former, because the pronunciation ends in "subble", not "cubble" or "cable") is going to be less confusing than using URI and/or IRI where previously appropriate.

I have never understood the difficulty in discerning between Identifiers (URI, IRI), Locators (URL), and Names (URN), nor in comprehending that the former (URI/IRI) include both of the latter (URL, URN).  Much like all Squares are Rectangles (four sides and four 90° anges), but not all Rectangles are Squares (all four sides are same length), all URLs and URNs are URIs, but not all URIs are URNs or URLs.

All that said... Given that we are bowing to WHAT WG's pressure, however, this PR seems to do the job.

terms.html
```diff
@@ -169,8 +169,12 @@
...
   </dd>
-  <dt><dfn data-lt="URI|URIs">URI</dfn></dt>
+  <dt><dfn data-lt="URL|URLs">URL</dfn></dt>
   <dd>
-A Uniform Resource Identifier, as defined by [[RFC3986]].
+A Uniform Resource Locator, as defined by [[URL]]. This specification does
+not use the term URI or IRI because those terms have been deemed to be confusing
+to Web developers. This specification uses the phrase "nondereferenceable URL"
+when referring to the concept of a URI or IRI, as defined in [[?RFC3986]] and
+[[?RFC3987]], respectively.
```
brentzundel, line 178: Why are these references non-normative?
[[?RFC3986]] and [[?RFC3987]]

csarven, line 178: The perception of confusion is based on "search result popularity contest" [URL] from a particular search engine's users - on the entire planet.

I'd presume that the strange subset of Web developers finding URI or IRI confusing but not URL, may simply find all sorts of other things in vc-data-model to be unpenetrable.

Having said that, if URL is intended for conformance, there may not be a need to mention URI or IRI.

It may be best to not introduce phrases like "nonderefenceable URL" in vc-data-model where it may be more suitable elsewhere, if so desired, e.g., [URL].

index.html
```diff
@@ -1578,8 +1579,8 @@ <h3>Issuer</h3>
           <dt><var>issuer</var></dt>
           <dd>
 The value of the <code>issuer</code> <a>property</a> MUST be either a
-<a>URI</a> or an object containing an <code>id</code> <a>property</a>. It is
-RECOMMENDED that the <a>URI</a> in the <code>issuer</code> or its
+<a>URL</a> or an object containing an <code>id</code> <a>property</a>. It is
```
OR13, line 1582: are DIDs URLs?

OR13 (dismissed):
These changes appear to prevent a DID from being a valid `id` in many locations.

index.html
iherman: Yes. I believe we checked that in the lifetime of the DID WG.

dlongley: A [DID is a URI](https://www.w3.org/TR/did-core/#did-syntax), which [is a URL per the WHATWG spec](https://url.spec.whatwg.org/#goals). Additionally, every DID is also a DID URL, [per the syntax rules in the DID core spec](https://www.w3.org/TR/did-core/#did-url-syntax). Note a DID with an empty path, no optional query, and no optional fragment is both a DID and a DID URL.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2022-12-14-vcwg#section3-5) on 2022-12-14

- no resolutions were taken

#### 3.5. Change reference to URI/IRI to use WHATWG URL specification. (pr vc-data-model#994)

_See github pull request [vc-data-model#994](https://github.com/w3c/vc-data-model/pull/994)._




**Manu Sporny:** PR994, a request to talk about URLs in the spec - slight bit of push back but no objections. A DID is a URL but it can be referenced in the issue.  

</details>

terms.html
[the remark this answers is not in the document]
dlongley: I don't think "non-deferenceable URL" is very helpful.

Even URNs are not "non-dereferenceable". Every URI "references" something, since the point of a URI is to identify  (reference) a logical or physical resource. But not every URI is a "locator". That means you can't always send a message to a commonly resolvable place specified within the URI using a protocol specified within the URI to "locate" the referenced resource. Maybe none of that is all that useful either and people just get lost in the details.

I do think it's more likely to be helpful to say whether there's an expectation that a URI (now just "URL") "resolves" to something (and what that something is expected to be) where necessary. Otherwise, all that matters is that we say "this is where you put an identifier of a specific (and standard) format, i.e., a "URL", to reference something". If we aren't restricting "what you can do / should be able to with that identifier" then it doesn't really matter and we don't need to say anything about "locators", etc.

msporny:
@OR13 wrote:
> These changes appear to prevent a DID from being a valid `id` in many locations.

I don't believe that is correct, @OR13 -- please confirm: https://github.com/w3c/vc-data-model/pull/994#discussion_r1048741793

Requesting a re-review from you.

terms.html
msporny: The WHATWG URL specification supersedes the URI and IRI specifications and therefore we no longer have any normative dependence on the IETF URI and IRI specifications. Yes, the horror :), but W3C and IETF lost that fight years ago and now WHATWG is the maintainer of "how to process a URI/IRI/URL".

We now depend on the WHATWG URL specification for all normative language on how to treat and process URLs (and IRIs and URIs). The WHATWG URL specification is what's implemented by the vast majority of modern URI/IRI/URL processors/resolvers, AFAIK.

msporny: While I agree with the concerns you're raising, the ship sailed years ago and IETF and W3C lost control of the definition of URL to WHATWG. That's where the most up to date spec resides today, that's what the browser vendors and URL library maintainers implement, and that's what we're trying to align with in this PR.

Looks like a number of people reviewing this PR don't like the "nonderefenceable URL" language here, so I'll try to find alternate language for that.

index.html
msporny: @OR13 asked:
> are DIDs URLs?

Yes, they are. (citation above)

terms.html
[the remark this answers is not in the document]
msporny: > I don't think "non-deferenceable URL" is very helpful.

Ok, I'll make another pass on this PR with that in mind. The  "non-deferenceable URL" is not popular. :)

OR13 (approved):

terms.html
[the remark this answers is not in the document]
msporny: Fixed in eb0f404d62f99fa01c35af74072be5869996596f.

I went down a rabbit hole trying to see if we should pull the [WHATWG Fetch spec](https://fetch.spec.whatwg.org/#url) in before realizing that if we did that, we would definitely step on the landmine that @OR13 was concerned about stepping on -- turns out WHATWG Fetch does not define "fetching" for anything other than a handful of schemes (http(s), ws(s), data, file, etc.). So, I kept it generic by saying "dereference" and DID NOT use the term "fetch", and said that "how you dereference is covered by the URL scheme specification".

msporny: Fixed in eb0f404.

Issue #978, opened 2022-11-14
Requesting content type for `credential` (without proof).


OR13:
### Update

I made 3 related proposals to the list:

- https://lists.w3.org/Archives/Public/public-vc-wg/2022Dec/0032.html
- https://lists.w3.org/Archives/Public/public-vc-wg/2022Dec/0011.html
- https://lists.w3.org/Archives/Public/public-vc-wg/2022Nov/0034.html

All 3 proposals request a JSON media type `application/credential+json`, based on the convention established by activity streams.

### Previous Description

This would allow for a W3C Credential to become a [SCITT claim](https://datatracker.ietf.org/doc/draft-birkholz-scitt-architecture/) (terminology differs).

This would also allow for COSE Sign1 to be used.

Essentially a SCITT Claim could become a Verifiable Credential, when used with COSE Sign1 with a header that looks like this:

```

{
  iss: did:example:123,
  kid: #0,
  cty: application/w3c-credential+json <-- this is what I need
}

{ w3c-credential json as "payload" }

signature

```

Note that this would not be a "vc-cwt"... and yet, its still very useful... 

We might propose a `vc-cwt` with a mapping similar to `vc-jwt` for consistency, but that would not be a solution to this use case.

mprorock:
Getting a content type like this would be awesome and very helpful in SCITT and elsewhere.

Perhaps the following as a starting place?
```
application/verifiable-credential+json
application/verifiable-credential+json-ld
application/verifiable-credential+jwt
```

OR13:
@mprorock following the convention from DID WG, we would have the following:

```
application/verifiable-credential+json
application/verifiable-credential+ld+json
```

Both would be examples of "verifiable credentials with external proofs" similar to how VC-JWT is a "VC with external proof".

mprorock:
> @mprorock following the convention from DID WG, we would have the following:
> 
> ```
> application/verifiable-credential+json
> application/verifiable-credential+ld+json
> ```
> 
> Both would be examples of "verifiable credentials with external proofs" similar to how VC-JWT is a "VC with external proof".

That make total sense and works for me

letmaik:
+1 though it seems like the media type would be for the credential itself, and the complete structure with signing envelope is the verifiable credential. I think it's important to distinguish this.

OR13:
@letmaik yes, thats what I proposed, but due to this language in the existing TR, that concept may be confusing to readers:

[...]

PR #1000, opened 2022-12-14
Add media type registration request

OR13:
Addresses https://github.com/w3c/vc-data-model/issues/978#issue-1448307139

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1000.html" title="Last updated on Jan 16, 2023, 9:47 PM UTC (3ce7e6e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1000/6bae9b7...3ce7e6e.html" title="Last updated on Jan 16, 2023, 9:47 PM UTC (3ce7e6e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5247,7 +5247,7 @@ Extensions of Verifiable Credentials Data Model 1.0</a>
     <tr>
       <td>Encoding considerations: </td>
       <td>
-        Resources that use the "<code>application/credential+json</code>"
+        Resources that use the "<code>application/credential+ld+json</code>"
         Media Type are required to conform to all of the requirements
         for the "<code>application/json</code>" Media Type and are
         therefore subject to the same encoding considerations specified
--- a/index.html
+++ b/index.html
@@ -5224,6 +5224,79 @@ Extensions of Verifiable Credentials Data Model 1.0</a>
         </li>
       </ul>
 
+<section id="media-type">
+  <h2>The <code>application/credential+json</code> Media Type</h2>
+  <p>
+    This specification registers the <code>application/credential+json</code> MIME Media Type specifically for
+    identifying documents
+    conforming to the Verifiable Credentials format.
+  </p>
+  <table>
+    <tr>
+      <td>Type name: </td>
+      <td>application</td>
+    </tr>
+    <tr>
+      <td>Subtype name: </td>
+      <td>credential+json</td>
+    </tr>
+    <tr>
+      <td>Required parameters: </td>
+      <td>None</td>
+    </tr>
+    <tr>
+      <td>Optional parameters: </td>
+      <td>
+        profile: The profile parameter for the application/credential+json
+        media type allows one or more profile URIs to be specified. These
+        profile URIs have the identifier semantics defined in [[!RFC6906]].
+        The "profile" media type parameter MUST be quoted. It contains a
+        non-empty list of space-separated URIs (the profile URIs).
+        <pre>profile-param = "profile=" profile-value
+profile-value = <"> profile-URI 0*( 1*SP profile-URI ) <">
+profile-URI   = URI</"></"></pre>
+        The "URI" in the above grammar refers to the "URI" as defined in
+        Section 3 of [[!RFC3986]].
+      </td>
+    </tr>
+    <tr>
+      <td>Encoding considerations: </td>
+      <td>
+        Resources that use the "<code>application/credential+json</code>"
+        Media Type are required to conform to all of the requirements
+        for the "<code>application/json</code>" Media Type and are
+        therefore subject to the same encoding considerations specified
+        in Section 11 of [[!RFC7159]].
+      </td>
+    </tr>
+    <tr>
+      <td>Security considerations: </td>
+      <td>As defined in this specification.</td>
+    </tr>
+    <tr>
+      <td>Contact: </td>
+      <td>
+        Orie Steele &lt;<a href="mailto:orie@transmute.industries">orie@transmute.industries</a>&gt;
+      </td>
+    </tr>
+  </table>
+
+  <p>
+    Note that while the Verifiable Credentials format uses JSON-LD
+    conventions, there are a number of constraints and additional
+    requirements for Verifiable Credential implementations that justify
+    the use of a specific media type.
+  </p>
+
+  <p>
+    Because Verifiable Credentials can be considered a restricted profile
+    of JSON-LD, Implementations SHOULD consider the
+    `application/ld+json; profile="https://www.w3.org/ns/credentials"`
+    media type as being equivalent to `application/credential+json`.
+  </p>
+</section>
+
+
     </section>
 
     <section>
--- a/index.html
+++ b/index.html
@@ -5273,12 +5273,6 @@ Extensions of Verifiable Credentials Data Model 1.0</a>
     the use of a specific media type.
   </p>
 
-  <p>
-    Because Verifiable Credentials can be considered a restricted profile
-    of JSON-LD, Implementations SHOULD consider the
-    `application/ld+json; profile="https://www.w3.org/ns/credentials"`
-    media type as being equivalent to `application/credential+json`.
-  </p>
 </section>
 
 
--- a/index.html
+++ b/index.html
@@ -5261,7 +5261,7 @@ Extensions of Verifiable Credentials Data Model 1.0</a>
     <tr>
       <td>Contact: </td>
       <td>
-        W3C Verifiable Claims Working Group <a href="mailto:public-vc-wg@w3.org">public-vc-wg@w3.org</a>
+        W3C Verifiable Credentials Working Group <a href="mailto:public-vc-wg@w3.org">public-vc-wg@w3.org</a>
       </td>
     </tr>
   </table>
--- a/index.html
+++ b/index.html
@@ -5238,7 +5238,7 @@ Extensions of Verifiable Credentials Data Model 1.0</a>
     </tr>
     <tr>
       <td>Subtype name: </td>
-      <td>credential+json</td>
+      <td>credential+ld+json</td>
     </tr>
     <tr>
       <td>Required parameters: </td>
--- a/index.html
+++ b/index.html
@@ -5249,7 +5249,7 @@ Extensions of Verifiable Credentials Data Model 1.0</a>
       <td>
         Resources that use the "<code>application/credential+ld+json</code>"
         Media Type are required to conform to all of the requirements
-        for the "<code>application/json</code>" Media Type and are
+        for the "<code>application/ld+json</code>" Media Type and are
         therefore subject to the same encoding considerations specified
         in Section 11 of [[!RFC7159]].
       </td>
--- a/index.html
+++ b/index.html
@@ -5227,7 +5227,7 @@ Extensions of Verifiable Credentials Data Model 1.0</a>
 <section id="media-type">
   <h2>The <code>application/credential+ld+json</code> Media Type</h2>
   <p>
-    This specification registers the <code>application/credential+json</code> MIME Media Type specifically for
+    This specification registers the <code>application/credential+ld+json</code> MIME Media Type specifically for
     identifying documents
     conforming to the Verifiable Credentials format.
   </p>
--- a/index.html
+++ b/index.html
@@ -5225,7 +5225,7 @@ Extensions of Verifiable Credentials Data Model 1.0</a>
       </ul>
 
 <section id="media-type">
-  <h2>The <code>application/credential+json</code> Media Type</h2>
+  <h2>The <code>application/credential+ld+json</code> Media Type</h2>
   <p>
     This specification registers the <code>application/credential+json</code> MIME Media Type specifically for
     identifying documents
@@ -5245,21 +5245,6 @@ Extensions of Verifiable Credentials Data Model 1.0</a>
       <td>None</td>
     </tr>
     <tr>
-      <td>Optional parameters: </td>
-      <td>
-        profile: The profile parameter for the application/credential+json
-        media type allows one or more profile URIs to be specified. These
-        profile URIs have the identifier semantics defined in [[!RFC6906]].
-        The "profile" media type parameter MUST be quoted. It contains a
-        non-empty list of space-separated URIs (the profile URIs).
-        <pre>profile-param = "profile=" profile-value
-profile-value = <"> profile-URI 0*( 1*SP profile-URI ) <">
-profile-URI   = URI</"></"></pre>
-        The "URI" in the above grammar refers to the "URI" as defined in
-        Section 3 of [[!RFC3986]].
-      </td>
-    </tr>
-    <tr>
       <td>Encoding considerations: </td>
       <td>
         Resources that use the "<code>application/credential+json</code>"
@@ -5276,7 +5261,7 @@ profile-URI   = URI</"></"></pre>
     <tr>
       <td>Contact: </td>
       <td>
-        Orie Steele &lt;<a href="mailto:orie@transmute.industries">orie@transmute.industries</a>&gt;
+        W3C Verifiable Claims Working Group <a href="mailto:public-vc-wg@w3.org">public-vc-wg@w3.org</a>
       </td>
     </tr>
   </table>


index.html
```diff
@@ -5224,6 +5224,79 @@ <h2>IANA Considerations</h2>
...
+      </td>
+    </tr>
+    <tr>
+      <td>Security considerations: </td>
+      <td>As defined in this specification.</td>
+    </tr>
+    <tr>
+      <td>Contact: </td>
+      <td>
+        Orie Steele &lt;<a href="mailto:orie@transmute.industries">orie@transmute.industries</a>&gt;
```
TallTed, line 5279: I wonder whether this shouldn't be VCWG (and successors) and/or W3C? @ivan?

OR13: yes, probably.

index.html
```diff
@@ -5224,6 +5224,79 @@ <h2>IANA Considerations</h2>
...
+  <p>
+    Note that while the Verifiable Credentials format uses JSON-LD
+    conventions, there are a number of constraints and additional
+    requirements for Verifiable Credential implementations that justify
+    the use of a specific media type.
+  </p>
+
+  <p>
+    Because Verifiable Credentials can be considered a restricted profile
+    of JSON-LD, Implementations SHOULD consider the
```
Sakurann, line 5293: > Because Verifiable Credentials can be considered a restricted profile
    of JSON-LD
What does this mean? VCs are rather application of JSON-LD rather than it's profile... or did you want to mean VCs use restricted features of JSON-LD?

index.html
```diff
@@ -5224,6 +5224,79 @@ <h2>IANA Considerations</h2>
...
+        profile URIs have the identifier semantics defined in [[!RFC6906]].
+        The "profile" media type parameter MUST be quoted. It contains a
+        non-empty list of space-separated URIs (the profile URIs).
+        <pre>profile-param = "profile=" profile-value
+profile-value = <"> profile-URI 0*( 1*SP profile-URI ) <">
+profile-URI   = URI</"></"></pre>
+        The "URI" in the above grammar refers to the "URI" as defined in
+        Section 3 of [[!RFC3986]].
+      </td>
+    </tr>
```
msporny, line 5261:
```suggestion
```

It's not clear what use case this parameter is intended to address. There should probably be a section about it in the main part of the specification. It's not clear to me what a profile URI is intended to do if supplied to a VC processor.

index.html
```diff
@@ -5224,6 +5224,79 @@ <h2>IANA Considerations</h2>
...
+    <tr>
+      <td>Encoding considerations: </td>
+      <td>
+        Resources that use the "<code>application/credential+json</code>"
```
msporny, line 5265:
```suggestion
        Resources that use the "<code>application/vc+json</code>"
```

index.html
```diff
@@ -5224,6 +5224,79 @@ <h2>IANA Considerations</h2>
...
+    conventions, there are a number of constraints and additional
+    requirements for Verifiable Credential implementations that justify
+    the use of a specific media type.
+  </p>
+
+  <p>
+    Because Verifiable Credentials can be considered a restricted profile
+    of JSON-LD, Implementations SHOULD consider the
+    `application/ld+json; profile="https://www.w3.org/ns/credentials"`
+    media type as being equivalent to `application/credential+json`.
```
msporny, line 5295:
```suggestion
    media type as being equivalent to `application/vc+json`.
```

index.html
msporny, line 5293: Hmm, this seems like it would block the registration of `application/vc+ld+json`? To put it another way, this text seems to be saying "treat a VC profile document as application/vc+json instead of application/vc+ld+json"... which seems really weird.

index.html
msporny: Agree with @TallTed, change controller should be the VCWG.

msporny (changes requested):
The media type name won't survive review, please change to `credential` to `vc`. The change controller should be the VCWG. I've requested clarification on the purpose of the profile parameter as well as the equivalence statement, which seems off.

index.html
OR13: I'm happy to merge this change suggestion after comments from chairs.

index.html
OR13: I think this suggestion further conflates `credential` and `verifiable credential`... I would prefer to be able to uniquely identify "content of the data model" from "content of the data model with protection".

OR13: However, that ship may have sailed in which case perhaps we should adopt your suggested change.

index.html
OR13: Its copied from activity streams, I am happy to modify it... I don't think we need `vc+ld+json` unless we agree to define `vc+json` or `credential+json` in a way that is not compatible... I suspect this was the intention of the text in the original activity pub context.

index.html
selfissued: I agree that we should not conflate the media types for a credential (the data structure to be secured) and a verifable credential (a secured credential).  "credential" is correct for the former and "verifiable-credential" or "vc" is correct for the latter.

index.html
OR13: Can we propose an email for me to use in this PR?

index.html
OR13: Its pulled from the activity streams registration, which seems like a good "starting place".

index.html
```diff
@@ -5224,6 +5224,79 @@ <h2>IANA Considerations</h2>
         </li>
       </ul>
 
+<section id="media-type">
+  <h2>The <code>application/credential+json</code> Media Type</h2>
```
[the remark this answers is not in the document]
OR13: I don't think I can accept this change, because the intention is to define the media type for the "data model" not the "secured data model".... unless we say that a "verifiable credential" can be "unsecured" similar to how "alg: none" is used... If thats ok, I am happy to accept this change, see the proposals on the mailing lists for the intended use of this `cty` with COSE and JOSE.

index.html
OR13: See comment here: https://github.com/w3c/vc-data-model/pull/1000#discussion_r1065927379

index.html
Sakurann: thanks for clarifying!

index.html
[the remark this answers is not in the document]
Sakurann: definitions of the JOSE header parameters are very clear that

`cty` which is used to [`declare the media type of the secured content (the payload)`](https://www.rfc-editor.org/rfc/rfc7515.html#section-4.1.10) , should be `credential+xxx` and not `verifiable-credential+xxx`.

`typ` which is used to [`declare the media type of this complete JWS`](https://www.rfc-editor.org/rfc/rfc7515.html#section-4.1.9) should be `verifiable-credential+xxx`.

the ship that vc-data-model spec defined `credential` in a different way than identity industry used to define/use it has sailed. with `credential` term being more and more accepted as it is defined in vc-data-model spec, with good arguments, registration of `credential+xxx` should be possible.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-01-11-vcwg#section2-3) on 2023-01-11

- no resolutions were taken

#### 2.3. Add media type registration request (pr vc-data-model#1000)

_See github pull request [vc-data-model#1000](https://github.com/w3c/vc-data-model/pull/1000)._




**Manu Sporny:** Next PR 1000 is about media types registration..  

**Michael Jones:** There is an important distinction between the type that is signed versus the type that is the entire thing..  

> *Manu Sporny:* [https://datatracker.ietf.org/doc/html/draft-ietf-mediaman-suffixes](https://datatracker.ietf.org/doc/html/draft-ietf-mediaman-suffixes).

**Manu Sporny:** Mike's comments will be useful input to the PR. Mike will provide more info in writing..  

**Michael Jones:** IETF may need multiple plus signs..  

**Manu Sporny:** This will have impact on the W3C VC efforts as well..  
… We need more reviewers what to do to with this..  

</details>

index.html
brentzundel: We could use member-vc-wg@w3.org which is the WG member-only mailing list.

index.html
brentzundel: If no one complains about removing the section (which hasn't happened in the 3 weeks since it was suggested) and @OR13 agrees with the change, then I think it is fine to commit the suggestion.

index.html
quartzjer: I would also "+1" for `application/credential` being the data model, and `application/vc` being a secured credential.

index.html
selfissued: I support removing this optional parameter.  I've created/registered/used a lot of content types and I've never seen a parameter used.  It seems like unnecessary complexity.

index.html
```diff
@@ -5224,6 +5224,79 @@ <h2>IANA Considerations</h2>
...
+    conforming to the Verifiable Credentials format.
+  </p>
+  <table>
+    <tr>
+      <td>Type name: </td>
+      <td>application</td>
+    </tr>
+    <tr>
+      <td>Subtype name: </td>
+      <td>credential+json</td>
```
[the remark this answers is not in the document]
selfissued: I suggest that the right name for this is `credential+ld+json` - paralleling the syntax used in the DID spec.  As discussed on today's VC call, this is a media type for the credential body - not a media type for the secured (verifiable) credential, so the `credential` part is right.

index.html
OR13, line 5228:
```suggestion
  <h2>The <code>application/credential+ld+json</code> Media Type</h2>
```

index.html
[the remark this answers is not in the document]
OR13: I made your suggestion here: https://github.com/w3c/vc-data-model/pull/1000/files#r1067595321

TallTed:
This PR has a LOT of disconnected discussion of `application/credential+...` vs `application/vc+...` vs `application/verifiable-credential+...`.  I would feel better if all that were brought into an issue, discussed there in detail, and then the results applied to this PR and our media type registration request(s)...

index.html
selfissued: Thanks - let's do this.

index.html
OR13: This section has not been added yet, this PR is entirely new.

index.html
OR13: I think we should use: public-vc-wg@w3.org

selfissued: I agree with using the public list.  For what it's worth, that's what IANA registrations from OpenID specs do.

OR13, line 5279:
```suggestion
        W3C Verifiable Claims Working Group &lt;<a href="mailto:public-vc-wg@w3.org">public-vc-wg@w3.org</a>&gt;
```

selfissued: I agree with this suggestion.

OR13: its been updated to the public working group list.

msporny:
@TallTed wrote:
> I would feel better if all that were brought into an issue, discussed there in detail, and then the results applied to this PR and our media type registration request(s)...

Yes, agreed. There seem to be multiple parallel discussions going on here, with some suggestions that we're going to have even more media types in the future. At present, I can't tell which one of these media types are going to eventually be a valid media type for things associated with this specification:

* `application/credential` (presumably a JSON-LD document w/o any sort of proof? or with one?)
* `application/credential+json` (given that we haven't resolved #947, this is still a possibility?)
* `application/credential+ld+json (Is this protected or not?)
* `application/credential+jwt (Is this a VC-JWT? Or a VC-JOSE/VC-COSE media type?)
* `application/verifiable-credential` (presumably a JSON-LD document w/ a proof?)
* `application/verifiable-credential+json` (given that we haven't resolved #947, this is still a possibility?)
* `application/verifiable-credential+ld+json (protected JSON-LD, I presume?)
* `application/verifiable-credential+jwt (Is it this one or the one above?)

It's hard to understand where the current proposal fits in the options above, and it's not clear that making the distinction between a "credential" and "verifiable credential" via the media type really buys us enough value.

@OR13, what's the plan here -- which one of the above are you intending as options?

index.html
msporny: We can ask the Editor, @cwebber, what the intent was. If what we're trying to accomplish is something akin to ActivityPub, I could see that working at the most basic level:

https://w3c.github.io/activitypub/#retrieving-objects

... but then have questions when it comes to structured suffixes like `+ld+json` and `+jwt` and `+cwt`.

OR13:
@msporny the intention is implemented in the PR, the proposals here explain one potential use of this type in JOSE and COSE:

https://github.com/w3c/vc-data-model/pull/1000#discussion_r1065931004

Looking forward to discussing on a special topic call.

index.html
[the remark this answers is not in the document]
msporny: > with credential term being more and more accepted as it is defined in vc-data-model spec, with good arguments, registration of credential+xxx should be possible.

Alright, if the VCWG is going to get behind that argument, it might be doable. We'll have to see what the media types list says.

msporny:
@OR13 wrote:
>  There are a lot of places where credential and not verifiable-credential show up... and because of the security differences it is critical they be identified as unambiguously as possible.

I'm on board with the concept of having two different media types, and also dubious that developers will use them properly for the same reason that `did:web` is recommending that the DID Document can be served as `application/json`. We know developers get this stuff wrong, and the more media types we have, the higher the chances are that they'll pick the wrong media type.

> I don't think we need to have all answers in at once

Neither do I, but I do think that we have to have a better understanding of the path forward than what we have right now. We don't want to paint ourselves into a corner either. My hope is that we can get to that plan in the upcoming Special Topic call.

> rather than moving forward with one or two items we know we need, and then hashing out details as we go on the other items

Yes, you can make iterative progress that way... and it's also how one paints themselves into a corner. That's the concern here.

All that said, here's a cut at a unified philosophy for doing media type processing:

1. There will be at least two media types, one for unsecured credentials and another for secured credentials.
2. The unsecured credential media type will be `application/credential+json`, which will be equivalent to `application/ld+json` with a profile set to `https://www.w3.org/ns/credentials`.
3. The secured credential media type will be `application/verifiable-credential+json`, which will be equivalent to `application/ld+json` with a profile set to `https://www.w3.org/ns/credentials`.

Once we define these media types, we'll have to talk about what happens when things go wrong, such as:

1. What happens when the client just requests `application/json` for a known good `application/credential+json` resource?
2. What happens when the client requests `application/credential+json` but the server responds with `application/json`?
3. Same questions above, but for `application/ld+json` instead of `application/json`?
4. What is the content type for a data integrity protected `application/credential+json`? A JWT-protected one? A JWS-protected one? A CWT-protected one?

I suggest the answers to the questions above are:

1. Return the same content as for `application/credential+json` using the IANA Media Type Structured Suffix algorithm.
2. Use rules (yet to be written) to do deterministic content sniffing on the payload (look for `@context` with specific values (for v1 and v2) in the first array element).
3. Same as answer to 2 above.
4. `application/verifiable-credential+json`, `application/verifiable-credential+jwt`, `application/verifiable-credential+jws`, and `application/verifiable-credential+cwt`, respectively.

Thoughts?

OR13:
Yes to 1, and 4... maybe 2 and 3... I think we should cover negotiation and the `Accept` parameter separately from defining `Content-Type`.

dlongley:
Cross-posting this here from #978:

Also, if people want to publish VCs in web pages, which media type should they use? Is it going to make things more challenging for this use case if we have extra media types that mean the same thing? People [publish structured data](https://developers.google.com/search/docs/appearance/structured-data/intro-structured-data#how-structured-data-works-in-google-search) using `application/ld+json` today and it's consumed by major search engines.

dlongley:
For JOSE (JWS-secured) containers, it seems one could do `application/jose` / `application/jose+json` with a `cty` for the unsecured payload being `application/json` / `application/ld+json` with a profile of `https://www.w3.org/ns/credentials` -- this means no new core media types would need to be invented. If there's a need for something new (and we could be clear about what it would accomplish and what the trade offs would be for the added complexity), it seems the unified processing model @msporny [proposed](https://github.com/w3c/vc-data-model/pull/1000#issuecomment-1384311704) could be explored as a potential solution.

OR13:
@dlongley I would guess that both `application/credential+ld+json` and `application/verifiable-credential+ld+json` would be compatible with `application/ld+json`...  and `application/json`...

Indeed, your comment about using `application/ld+json` in a `cty` is exactly why we request the registration of `application/credential+ld+json`... Implementers want to be able to switch on `cty` when processing JOSE headers.

And they like the ability to be as specific as possible.

index.html
```diff
@@ -5224,6 +5224,64 @@ <h2>IANA Considerations</h2>
...
+    requirements for Verifiable Credential implementations that justify
+    the use of a specific media type.
+  </p>
+
+  <p>
+    Because Verifiable Credentials can be considered a restricted profile
+    of JSON-LD, Implementations SHOULD consider the
+    `application/ld+json; profile="https://www.w3.org/ns/credentials"`
+    media type as being equivalent to `application/credential+json`.
+  </p>
```
msporny, line 5281:
```suggestion
```

In order to get this PR in more quickly, let's get rid of profiles for now, we can revisit profiles in a future PR.

index.html
msporny: Ok, so this PR is just about `application/credential+json` which is a JSON-LD compatible object. Withdrawing my suggestion for doing `application/vc` in this PR.

index.html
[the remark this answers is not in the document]
msporny: Ok w/ doing another PR if needed.

index.html
```diff
@@ -5224,6 +5224,58 @@ <h2>IANA Considerations</h2>
...
+    conforming to the Verifiable Credentials format.
+  </p>
+  <table>
+    <tr>
+      <td>Type name: </td>
+      <td>application</td>
+    </tr>
+    <tr>
+      <td>Subtype name: </td>
+      <td>credential+json</td>
```
dlongley, line 5241:
```suggestion
      <td>credential+ld+json</td>
```

index.html
```diff
@@ -5224,6 +5224,58 @@ <h2>IANA Considerations</h2>
...
+      <td>credential+ld+json</td>
+    </tr>
+    <tr>
+      <td>Required parameters: </td>
+      <td>None</td>
+    </tr>
+    <tr>
+      <td>Encoding considerations: </td>
+      <td>
+        Resources that use the "<code>application/credential+json</code>"
```
OR13, line 5250:
```suggestion
        Resources that use the "<code>application/credential+ld+json</code>"
```

index.html
```diff
@@ -5224,6 +5224,58 @@ <h2>IANA Considerations</h2>
...
+        Media Type are required to conform to all of the requirements
+        for the "<code>application/json</code>" Media Type and are
```
OR13, line 5252:
```suggestion
        for the "<code>application/ld+json</code>" Media Type and are
```

index.html
```diff
@@ -5224,6 +5224,58 @@ <h2>IANA Considerations</h2>
         </li>
       </ul>
 
+<section id="media-type">
+  <h2>The <code>application/credential+ld+json</code> Media Type</h2>
+  <p>
+    This specification registers the <code>application/credential+json</code> MIME Media Type specifically for
```
dlongley, line 5230:
```suggestion
    This specification registers the <code>application/credential+ld+json</code> MIME Media Type specifically for
```

OR13, line 5230:
```suggestion
    This specification registers the <code>application/credential+ld+json</code> MIME Media Type specifically for
```

index.html
```diff
@@ -5224,6 +5224,58 @@ <h2>IANA Considerations</h2>
...
+      </td>
+    </tr>
+    <tr>
+      <td>Security considerations: </td>
+      <td>As defined in this specification.</td>
+    </tr>
+    <tr>
+      <td>Contact: </td>
+      <td>
+        W3C Verifiable Claims Working Group <a href="mailto:public-vc-wg@w3.org">public-vc-wg@w3.org</a>
```
msporny, line 5264:
```suggestion
        W3C Verifiable Credentials Working Group <a href="mailto:public-vc-wg@w3.org">public-vc-wg@w3.org</a>
```

dlongley (approved):

msporny (approved):
Having had a discussion with @OR13 about this, I understand that this PR is quite focused in what it's attempting to achieve and I agree with what this PR is attempting to achieve, which is at least one registered media type that can be used in `cty`, there may be others that the group registers in the future and those should go in separate PRs.

mkhraisha (approved):

index.html
selfissued: Yes, please apply this correction.

quartzjer (approved):

decentralgabe (approved):

msporny:

OR13:
For folks wondering how to use this new media type, here is an example: https://github.com/transmute-industries/vc-jws/pull/5/files

OR13:
Here is the media type in use with cose sign 1:

- https://github.com/transmute-industries/vc-cose/pull/10/files
- https://transmute-industries.github.io/vc-cose/

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-01-17-vcwg-special#section1) on 2023-01-17

- no resolutions were taken

### 1.  Add media type registration request (pr data-model#1000)

_See github pull request [vc-data-model#1000](https://github.com/w3c/vc-data-model/pull/1000)._




**Orie Steele:** Any objections to merging PR #1000?.  

> *Phillip Long:* +1 for merging.

**Kristina Yasuda:** I think we should merge it..  

> *Dave Longley:* +1 to merging.

**Orie Steele:** There are more approvals than I've seen on any PR since we resumed working..  
… There were change requests that were accepted, which became approvals..  

**Brent Zundel:** Please give us a summary.  

> *Kerri Lemoie:* Yes - overview would be helpful..

**Brent Zundel:** If there are no objections, I believe it would be appropriate to merge it.  

**Orie Steele:** A credential becomes a Verifiable Credential when it contains an embedded or external proof.  
… The media type describes this core data model type without or possibly with a proof.  
… We can take a step closer to resolving ambiguity in the specification.  
… We can use this content-type to describe the content that is secured in the specification.  
… There are two proof formats in the spec: embedded and external.  
… Data integrity proofs use embedded, VC-JWTs use external.  
… The purpose of this PR is to describe the thing that we're securing.  
… It can become a Verifiable Credential by securing it.  
… Having a media type gives us a way to talk about what it is.  

**Michael Jones:** Merge it.  

**Manu Sporny:** The only thing that prevents divergence is talking to one another across the specs.  

> *Michael Prorock:* +1 manu - editor comms, working group eyes - same as any other media type or external pointer (e.g. to alg combinations).

**Manu Sporny:** It's always possible to register too many media types or create non-interoperability.  
… We need to talk to one another.  
… When there are other media types, we need to ensure that they're aligned with our philosophy across the specifications.  
… If they diverge, they need to do so for very good reasons.  

**Michael Prorock:** There's an important lesson from this PR. The quickness of its approval indicates that this was needed..  
… If things get fuzzy later, at that point we can step back.  
… For this one, we quickly got to something useful.  

**Brent Zundel:** At this stage, we first wanted to talk about PR 1000.  
… We have a lot of approvals. It would be entirely appropriate to merge it..  

**Manu Sporny:** Please provide something for the disposition comments when you merge things..  
… I will add the disposition comments.  

**Brent Zundel:** Can the associated issues be closed?.  

</details>

index.html
TallTed: @msporny -- Please confirm that there are open issues for everything meant to be revisited, as it is likely that at least some will fall off the table otherwise.

msporny: I can confirm that there are not open issues for everything meant to be revisited, given it was not clear what needed to be revisited. I do plan to raise a number of issues in the future regarding media type, but at present, those are just in my head. I'll try to find some time to write these issues down at some point later today (if I get some spare cycles).

mwherman2000:
Can someone add or point to what the definition/purpose/goals of a "media type" is/are to this conversation?

The above is quite abstract if it isn't anchored in an agreed upon definition/purpose/goals of what a media type is.

Issue #953, opened 2022-10-19
Add `@vocab` for `private claims` in `/credentials/v2` `@context`


OR13:
PROPOSAL: Add an `@vocab` to the `/credentials/v2` `@context`.

Reasoning:

We (as editors of the did spec registries, and reviewers of various VC-JWT implementations) have seen a consistent failure to leverage `@context` properly in DID Core and Verifiable Credentials... due to the "term not defined in the context" issue.

The problem is not that "people don't want to define semantics"... its the opposite, they are eager to rely on semantics defined by the standards, but they don't know how to define new terms, or they don't want to take responsibility for hosting a vocabulary for their new terms as the first step in using or leveraging a new term... this is a barrier to adoption created by the W3C Working Groups,  based on an argument for semantic purity... that has in my opinion, greatly harmed the adoption of semantic technologies and standards, including Decentralized Identifiers and Verifiable Credentials.

By adding `@vocab` to the core data model context, we can give these developers the same experience they have with JWT private claims, which you should read about here: [rfc7519#section-4.3
](https://www.rfc-editor.org/rfc/rfc7519#section-4.3)

If this proposal is adopted, developers will be allowed to do this:

```jsonc
{
  "@context": [
    // "https://www.w3.org/2018/credentials/v1",
    "https://www.w3.org/ns/credentials/v2",
    // "https://www.w3.org/2018/credentials/examples/v1"
    // { "@vocab": "https://www.w3.org/ns/credentials#" }
  ],
  "id": "http://example.edu/credentials/1872",
  "type": ["VerifiableCredential", "NewCredentialType"],
  "issuer": { 
    "id": "did:example:123", 
     "type": ["Organization", "OrganizationType"] 
   },
  "issuanceDate": "2010-01-01T19:23:24Z",
  "credentialSubject": {
    "id": "did:example:456", 
    "type": ["Person", "JobType"],
    "claimName": "claimValue"
  }
}
```

and `claimName` will get a default IRI, assigned by `https://www.w3.org/ns/credentials/v2`... something like:

~`https://www.w3.org/ns/credentials/#claimName`.~

`https://www.w3.org/ns/credentials/claims/private#claimName`

This will produce beautiful and valid JSON-LD, which can immediately be processed in an interoperable way, using SPARQL, RDF, or other semantic standards developed at W3C and elsewhere, including conversion to Cypher, GQL or Labeled Property Graphs (Automatically).

[...]

PR #1001, opened 2022-12-14
Add `@vocab` to v2 context

OR13:
Addresses: https://github.com/w3c/vc-data-model/issues/953

### DIFF
--- a/contexts/credentials/v2
+++ b/contexts/credentials/v2
@@ -1,7 +1,7 @@
 {
   "@context": {
     "@protected": true,
-    "@vocab": "https://www.w3.org/2018/credentials#private-claim-",
+    "@vocab": "https://www.w3.org/ns/credentials/issuer-dependent#",
 
     "id": "@id",
     "type": "@type",
--- a/contexts/credentials/v2
+++ b/contexts/credentials/v2
@@ -1,6 +1,7 @@
 {
   "@context": {
     "@protected": true,
+    "@vocab": "https://www.w3.org/2018/credentials#private-claim-",
 
     "id": "@id",
     "type": "@type",


contexts/credentials/v2
```diff
@@ -1,6 +1,7 @@
 {
   "@context": {
     "@protected": true,
+    "@vocab": "https://www.w3.org/2018/credentials#private-claim-",
```
OR13, line 4: It's sad to see `2018` stuck in here forever.

I've opted to provide vocab expansions with a fragment prefix, so that no changes to W3C directories is needed.

This will lead to term definitions that look like this:

```
https://www.w3.org/2018/credentials#private-claim-failedToDefineThisTermInASecondContext
```

For cases like this:

 {
  "@context": "https://www.w3.org/2018/credentials/v2",
  "id": "http://example.edu/credentials/1872",
  "issuer": "did:example:123",
  "credentialSubject": {
    "id": "did:example:456",
    "failedToDefineThisTermInASecondContext": "🧠💎"
  }
}

dlongley: > It's sad to see `2018` stuck in here forever.

I agree that `2018` is not a good value there and wished we could have gone another direction in the 1.0 WG. I  also believe there's been discussion around whether we may be able to change that and replace it with `ns`. So it's conceivable that we wouldn't be stuck with it forever, but if we're going to make that change, it should definitely be before 2.0 goes to REC.

dlongley, line 4: I think using a `#` here may be the wrong thing to do unless the plan is to dynamically generate anchors in the vocab page. I think it would be better to use a slash to give us more options on what is served / rendered when the URL is hit. We probably want to render some text like "The definition for the term 'foo' is issuer dependent. Please see the issuer's documentation for a definition of this term."

I also think we should call this out as being "issuer dependent". So one possible alternative is:

```suggestion
    "@vocab": "https://www.w3.org/2018/credentials/issuer-dependent-vocab/",
```

OR13: I'm not sure, but I feel like maybe this is exploitable with path traversal? fragment seems safer.

OR13: I filed https://github.com/w3c/vc-data-model/issues/1002#issue-1497487224 to track that discussion, I would love to fix this before 2.0 goes to REC.

OR13: I am happy to do the client side parsing and anchoring part if we're allowed to make that experience nice, it feels like maybe a good thing to not reflect "private claims / issuer defined claims" through a W3C web server logs.

mprorock (approved):
This is an awesome path - i particularly like the `private-claim` approach for labeling

mkhraisha (approved):

decentralgabe (approved):

contexts/credentials/v2
iherman: I agree with @dlongley. Mixing these ad-hoc references with `https://www.w3.org/2018/credentials` is misleading. The V2 version of the vocabulary did everything to make the `#` behave properly, and this proposal would destroy it.

I would actually prefer to choose a distinct URL which is _not_ "part of" the official URL. We could do something like `https://www.w3.org/ns/private-vc-term#`. The advantage is that we could then create a single HTML file at that URL that would explain what these things are; if people really try to dereference `https://www.w3.org/ns/private-vc-term#failedToDefineThisTermInASecondContext` then they will hit something meaningful.

msporny (changes requested):
In general, supportive of the PR as long as the following change is made (see @iherman and @dlongley's comments as to why):

The `@vocab` URL should point to something that doesn't confuse these class of terms with the terms in the well defined vocabulary, so something like `https://www.w3.org/ns/credentials/WORD#` (in the case where we get rid of `2018`). This will make it easy to have a separate page that caters to the intent of setting a global base vocabulary. The word we use for `WORD` will need to be debated, which was entirely predictable, given that some want to use `@vocab` to elicit something that's "undefined" (which is true for some definition of 'undefined'), while others want to use it for "private claims" (which is a misleading 'cause it can be used for public, but negotiated out of band, claims), while others want to use it for "issuer-dependent-vocabulary" (which is more accurate, but a mouthful).

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-01-11-vcwg#section2-4) on 2023-01-11

- no resolutions were taken

#### 2.4. Add `@vocab` to v2 context (pr vc-data-model#1001)

_See github pull request [vc-data-model#1001](https://github.com/w3c/vc-data-model/pull/1001)._




**Manu Sporny:** Next PR is 1001. It's about add vocabulary. We need more reviewers on this..  

</details>

vongohren:
Quick question, adding this vocab, technically means that all VCs will pass JSON-LD validation? So if you do a mistake it still passes and you have to look for private-claim to verify what is unreferenced?

msporny:
> Quick question, adding this vocab, technically means that all VCs will pass JSON-LD validation? 

Yes, correct. Now, in order to detect errors, you'll have to look for terms that map to the "undefined" namespace.

> So if you do a mistake it still passes and you have to look for private-claim to verify what is unreferenced?

Yep, you got it.

contexts/credentials/v2
OR13: @iherman @dlongley I don't agree that we are "mixing ad hoc references"... we are defining undefined terms in a way that ensures processing succeeds.

Look at the context definition for 1.1... it does not use `path` based names for anything (which schema.org does).

Some options to consider:

1. `https://www.w3.org/2018/credentials/issuer-dependent-vocab/` + `term used by an issuer`
1. `https://www.w3.org/ns/credentials/private-claim/` + `term used by an issuer`
1. `https://www.w3.org/ns/credentials/private-claim#` + `term used by an issuer`
1. `https://www.w3.org/ns/credentials#private-claim-` + `term used by an issuer`
1. `https://www.w3.org/ns/credentials#issuer-defined-` + `term used by an issuer`
1. `https://www.w3.org/ns/credentials#underspecified-` + `term used by an issuer`

mkhraisha (approved):

contexts/credentials/v2
msporny: This one looks good: https://www.w3.org/ns/credentials/issuer-dependent#<TERM_USED_BY_ISSUER>

OR13, line 4:
```suggestion
    "@vocab": "https://www.w3.org/ns/credentials/issuer-dependent#",
```

OR13: accepted.

OR13:
@msporny please re-review.

msporny (approved):

vongohren:
> > Quick question, adding this vocab, technically means that all VCs will pass JSON-LD validation? 
> 
> Yes, correct. Now, in order to detect errors, you'll have to look for terms that map to the "undefined" namespace.
> 
> > So if you do a mistake it still passes and you have to look for private-claim to verify what is unreferenced?
> 
> Yep, you got it.

Cool cool. It sounds scary but I assume as a compromise all json ld referencers will add a warning? Or is it up to the verifiers and issuers to handle this?

contexts/credentials/v2
```diff
@@ -1,6 +1,7 @@
 {
   "@context": {
     "@protected": true,
+    "@vocab": "https://www.w3.org/ns/credentials/issuer-dependent#",
```
iherman, line 4: This works for me.

iherman (approved):

msporny:
> Cool cool. It sounds scary but I assume as a compromise all json ld referencers will add a warning? 

Nope, that is unlikely to happen.

> Or is it up to the verifiers and issuers to handle this?

Undefined. :)

This is why some of us were warning that putting `@vocab` in the base context was not an ideal solution. It basically turns off error processing. Some view turning off error processing as a good thing, leading to an easier adoption path for developers. Others view turning off error processing as silencing errors that could have catastrophic consequences for some use cases.

I know that Digital Bazaar is going to have to invest in adding a feature to our JSON-LD processors that will have a callback when certain types of triples are generated now (such as those that are in the "undefined namespace") in order to be able to throw linting warnings to developers that care about the use of `@vocab`.

Make no mistake, this is a "hold your nose" compromise for some of us. :)

contexts/credentials/v2
OR13: @iherman yes... I discussed this with @msporny... I had same thought... This was the proposal that came from that... If this discussion balloons because of that choice, I suggest we revert to 2018.

dlongley (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-01-18-vcwg#section2-2) on 2023-01-18

- no resolutions were taken

#### 2.2. Add `@vocab` to v2 context (pr vc-data-model#1001)

_See github pull request [vc-data-model#1001](https://github.com/w3c/vc-data-model/pull/1001)._




**Manu Sporny:** next item is remove zkp, we will address in special topic call later..  
… next is PR 1001, it covers the `@vocab` URL, and provides terms for anything that are related to the issuer..  
… we intend to provide better developer and user experience by for folks built on this feature.  
… there are no objections, we intend to merge by end of day.  

</details>

msporny:

PR #1013, opened 2023-01-19
Added Presentation as a possible domain for termsOfUse to the formal vocabulary

iherman:
Syncing with the latest change of the v2 context

### DIFF
--- a/vocab/credentials/v2/vocabulary.yml
+++ b/vocab/credentials/v2/vocabulary.yml
@@ -136,7 +136,7 @@ property:
 
   - id: termsOfUse
     label: terms of use
-    domain: cred:VerifiableCredential
+    domain: cred:VerifiableCredential,cred:VerifiablePresentation
     range: odrl:Policy
     comment: |
       If specified, the value of the `termsOfUse` property MUST specify one or more terms of use policies under which the creator issued the credential or presentation. If the recipient (a holder or verifier) is not willing to adhere to the specified terms of use, then they do so on their own responsibility and might incur legal liability if they violate the stated Terms of Use. Each `termsOfUse` MUST specify its type, for example, `IssuerPolicy`, and optionally, its instance `id`. The precise contents of each term of use is determined by the specific `TermsOfUse` type definition.
--- a/vocab/credentials/v2/vocabulary.yml
+++ b/vocab/credentials/v2/vocabulary.yml
@@ -139,7 +139,7 @@ property:
     domain: cred:VerifiableCredential,cred:VerifiablePresentation
     range: odrl:Policy
     comment: |
-      If specified, the value of the `termsOfUse` property MUST specify one or more terms of use policies under which the creator issued the credential or presentation. If the recipient (a holder or verifier) is not willing to adhere to the specified terms of use, then they do so on their own responsibility and might incur legal liability if they violate the stated Terms of Use. Each `termsOfUse` MUST specify its type, for example, `IssuerPolicy`, and optionally, its instance `id`. The precise contents of each term of use is determined by the specific `TermsOfUse` type definition.
+      If specified, the value of the optional `termsOfUse` property MUST specify one or more terms of use policies under which the issuer issued the credential or presentation. Each `termsOfUse` policy MUST specify its type (for example, `IssuerPolicy`) and MAY specify its instance `id`. The precise content of each term of use is determined by the specific `TermsOfUse` type definition. If the recipient (a holder or verifier) violates the specified terms of use, the responsibility is their own, and such violation may incur legal liability.
 
   - id: validFrom
     label: Valid from


iherman:
(Should be merged only when the automatic vocabulary generation is operational.)

msporny (approved):

msporny:
> (Should be merged only when the automatic vocabulary generation is operational.)

Automatic vocabulary generation is operational.

iherman:
@msporny I presume that you have the right and honor to merge PR-s... and I am not supposed to do it

msporny:
> @msporny I presume that you have the right and honor to merge PR-s... and I am not supposed to do it

Yes, I'm waiting for more people to see it/review it before merging. I almost merged it, noting that it was an "Editorial" change based on a previous PR that went in... but then thought I'd leave it floating out there for a bit because sometimes these sorts of PRs trigger people to make comments about other things we should be looking out for (or to assert that we're making a mistake for X or Y reason). Unless you object, I was going to give it the standard 7 day delay before merging.

vocab/credentials/v2/vocabulary.yml
```diff
@@ -136,7 +136,7 @@ property:
 
   - id: termsOfUse
     label: terms of use
-    domain: cred:VerifiableCredential
+    domain: cred:VerifiableCredential,cred:VerifiablePresentation
     range: odrl:Policy
     comment: |
       If specified, the value of the `termsOfUse` property MUST specify one or more terms of use policies under which the creator issued the credential or presentation. If the recipient (a holder or verifier) is not willing to adhere to the specified terms of use, then they do so on their own responsibility and might incur legal liability if they violate the stated Terms of Use. Each `termsOfUse` MUST specify its type, for example, `IssuerPolicy`, and optionally, its instance `id`. The precise contents of each term of use is determined by the specific `TermsOfUse` type definition.
```
TallTed, line 142:
```suggestion
      If specified, the value of the optional `termsOfUse` property MUST specify one or more terms of use policies under which the issuer issued the credential or presentation. Each `termsOfUse` value MUST specify its type (for example, `IssuerPolicy`) and MAY specify its instance `id`. The precise content of each term of use is determined by the specific `TermsOfUse` type definition. If the recipient (a holder or verifier) violates the specified terms of use, the responsibility is their own, and such violation may incur legal liability.
```

iherman: I am personally ok with the change, @TallTed, thanks. But I would prefer a domain expert to give the thumbs up before taking this over. My only slight reservation is that this document is, formally, not a recommendation, so I am not sure whether a _MAY_ is appropriate instead of a "may".

iherman: (I realize that _MUST_ was already part of the text, b.t.w., so my question is general, and not closely related to your change proposal.)

TallTed: Given that the original text was `MUST ... optionally`, I can't imagine any domain expert having an issue with the `MAY`, though they (and I, on rereading) may prefer `MAY additionally`.

jandrieu:
I'm confused. Doesn't @context already solve this problem?

brentzundel (approved):

TallTed:
@jandrieu — How does "`@context` already solve this problem"? And what do you see as "this problem" that `@context` somehow solves?

(Also, please edit https://github.com/w3c/vc-data-model/pull/1013#issuecomment-1403861073 and code fence `@context`...)

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-01-25-vcwg#section4-4) on 2023-01-25

- no resolutions were taken

#### 4.4. Added Presentation as a possible domain for termsOfUse to the formal vocabulary (pr vc-data-model#1013)

_See github pull request [vc-data-model#1013](https://github.com/w3c/vc-data-model/pull/1013)._




**Manu Sporny:** straightforward.  
… terms of use PR, syncing with latest changes.  
… follow on, mostly editorial.  

</details>

iherman:
> I'm confused. Doesn't `@context` already solve this problem?

No. `@context` is not a vocabulary specification, it is a _mapping_ from JSON to a vocabulary that is to be specified separately. The vocabulary specification, for example, defines constraints that RDF tools can check, draw conclusions of, etc, all the features that are not in the realm of a `@context`.

In this specific case, the `@context` file does define the mapping for a `termOfUse` property when used in a presentation but, in fact, this conflicts with the current status of the vocabulary which defines `termOfUse` as being used on credentials only. An RDF reasoner may then deduce, in a specific case, that a presentation is in fact a credential...

The proposed PR takes care of this.

vocab/credentials/v2/vocabulary.yml
msporny: Changes here look fine to me (speaking as a domain expert).

msporny:

Issue #950, opened 2022-10-14
Verification of multiple credential schemas


matthieubosquet:
In the [Data Schema section](https://www.w3.org/TR/vc-data-model-2.0/#data-schemas), the `credentialSchema` definition contains the following statement:

> The value of the `credentialSchema` [property](https://www.w3.org/TR/vc-data-model-2.0/#dfn-property) MUST be one or more data schemas that provide [verifiers](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verifier) with enough information to determine if the provided data conforms to the provided schema.

If more than one schema is provided, then verifiers should probably check the credential against every schema provided rather than making sure that at least one of the schemas provided conforms because it would allow for composability of schemas, for example, a credentialSchema property could point to `https://vc-schema.example.org/UniversityDegree` and `https://vc-schema.example.org/NonTransferable`.

Maybe the statement could be altered slightly to make the intended processing of multiple credential schemas (all valid vs at least one valid) clearer:

> The value of the `credentialSchema` [property](https://www.w3.org/TR/vc-data-model-2.0/#dfn-property) MUST be one or more data schemas that provide [verifiers](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verifier) with enough information to determine if the provided data conforms. Each `credentialSchema` MUST match the provided data for it to be determined as conforming by verifiers.

mark-moir:
I agree that a) it must be clear whether the data is required to conform with *all* schemas or if just one is sufficient, and b) that the natural choice is "all".  In case more flexibility is required for a particular use case, this could be achieved using, for example, [Schema Composition](https://json-schema.org/understanding-json-schema/reference/combining.html) if the schema type is `JSONSchemaValidator*`.

decentralgabe:
Related to #933 agree this should be adjusted to support multiple schemas

OR13:
See:

https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v1#L23

seems its not an `@container: @graph`use case... like:

https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v1#L45

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2022-09-15-vcwg#section5-4) on 2022-09-15

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 5.4. Verification of multiple credential schemas (issue vc-data-model#950)

[...]

PR #1023, opened 2023-02-01
Add language clarifying multiple data schemas

decentralgabe:
Addresses #950

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Feb 6, 2023, 6:15 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fvc-data-model%2Fdb7d236885dab37a099c65dbace861a697f1850c%2Findex.html%3FisPreview%3Dtrue)

```

😭  Sorry, there was an error generating the HTML. Please report this issue!
Specification: http://labs.w3.org/spec-generator/uploads/0GVPUy/index.html?isPreview=true%3FisPreview%3Dtrue&publishDate=2023-02-06
ReSpec version: 32.7.0
File a bug: https://github.com/w3c/respec/
Error: Error: Evaluation failed: Timeout: document.respec.ready didn't resolve in 28323ms.
    at ExecutionContext._ExecutionContext_evaluate (/u/spec-generator/node_modules/puppeteer-core/lib/cjs/puppeteer/common/ExecutionContext.js:229:15)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async ExecutionContext.evaluate (/u/spec-generator/node_modules/puppeteer-core/lib/cjs/puppeteer/common/ExecutionContext.js:107:16)
    at async generateHTML (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:221:12)
    at async toHTML (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:92:18)
    at async Object.generate [as respec] (file:///u/spec-generator/generators/respec.js:15:44)
    at async file:///u/spec-generator/server.js:252:48

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/vc-data-model%231023.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2468,12 +2468,14 @@ to determine if the provided data conforms to the provided schema object. Each
 or <code><a href="https://transmute-industries.github.io/vc-credential-schema-open-api-specification/">
 OpenApiSpecificationValidator2022</a></code>), and an <code>id</code> <a>property</a>
 which MUST be a <a>URL</a> identifying the schema file. </p>
+
 <p>The precise contents of each data schema object is determined by its 
 <code>type</code> definition. Different <code>type</code> implementations
 of data schema objects may choose to support different processing mechanisms 
 such as enabling the use of multiple data schemas for a single credential.
 Types may also enable specific processing rules (e.g. if there are multiple 
-schemas, all must be valid, some must be valid, and so on).</p>
+schemas, all must be valid, some must be valid, this schema applies to this
+portion of a credential, and so on).</p>
           </dd>
         </dl>
 
--- a/index.html
+++ b/index.html
@@ -2463,8 +2463,8 @@ The value of the <code>credentialSchema</code> <a>property</a> MUST be a single
 data schema object which provides <a>verifiers</a> with enough information 
 to determine if the provided data conforms to the provided schema object. Each
 <code>credentialSchema</code> MUST specify its <code>type</code> (e.g.,
-<code><a href="https://w3c-ccg.github.io/vc-json-schemas/">CredentialSchema2022</a></code>)
-, and an <code>id</code> <a>property</a> which MUST be a <a>URL</a> identifying
+<a href="https://w3c-ccg.github.io/vc-json-schemas/">`CredentialSchema2022`</a>),
+and an `id` <a>property</a> which MUST be a <a>URL</a> identifying
 the schema file. </p>
 
 <p>The precise contents of each data schema object is determined by its 
--- a/index.html
+++ b/index.html
@@ -2458,14 +2458,22 @@ the <a>verifiable credentials</a> that it issues:
 
         <dl>
           <dt><var>credentialSchema</var></dt>
-          <dd>
-The value of the <code>credentialSchema</code> <a>property</a> MUST be one or
-more data schemas that provide <a>verifiers</a> with enough information to
-determine if the provided data conforms to the provided schema. Each
-<code>credentialSchema</code> MUST specify its <code>type</code> (for example,
-<code>JsonSchemaValidator2018</code>), and an <code>id</code> <a>property</a>
-that MUST be a <a>URL</a> identifying the schema file. The precise contents of
-each data schema is determined by the specific type definition.
+          <dd><p>
+The value of the <code>credentialSchema</code> <a>property</a> MUST be a single
+data schema object which provides <a>verifiers</a> with enough information 
+to determine if the provided data conforms to the provided schema object. Each
+<code>credentialSchema</code> MUST specify its <code>type</code> (e.g.,
+<code>JsonSchemaValidator2018</code>, 
+<code><a href="https://w3c-ccg.github.io/vc-json-schemas/">CredentialSchema2022</a></code>,
+or <code><a href="https://transmute-industries.github.io/vc-credential-schema-open-api-specification/">
+OpenApiSpecificationValidator2022</a></code>), and an <code>id</code> <a>property</a>
+which MUST be a <a>URL</a> identifying the schema file. </p>
+<p>The precise contents of each data schema object is determined by its 
+<code>type</code> definition. Different <code>type</code> implementations
+of data schema objects may choose to support different processing mechanisms 
+such as enabling the use of multiple data schemas for a single credential.
+Types may also enable specific processing rules (e.g. if there are multiple 
+schemas, all must be valid, some must be valid, and so on).</p>
           </dd>
         </dl>
 
--- a/index.html
+++ b/index.html
@@ -2461,7 +2461,7 @@ the <a>verifiable credentials</a> that it issues:
           <dd><p>
 The value of the <code>credentialSchema</code> <a>property</a> MUST be a single
 data schema object which provides <a>verifiers</a> with enough information 
-to determine if the provided data conforms to the provided schema object. Each
+to determine whether the provided data conforms to the provided schema object. Each
 <code>credentialSchema</code> MUST specify its <code>type</code> (e.g.,
 <a href="https://w3c-ccg.github.io/vc-json-schemas/">`CredentialSchema2022`</a>),
 and an `id` <a>property</a> which MUST be a <a>URL</a> identifying
--- a/index.html
+++ b/index.html
@@ -2473,9 +2473,10 @@ which MUST be a <a>URL</a> identifying the schema file. </p>
 <code>type</code> definition. Different <code>type</code> implementations
 of data schema objects might choose to support different processing mechanisms 
 such as enabling the use of multiple data schemas for a single credential.
-Types may also enable specific processing rules (e.g. if there are multiple 
-schemas, all must be valid, some must be valid, this schema applies to this
-portion of a credential, and so on).</p>
+Types might also enable specific processing rules. For example, if there are multiple 
+schemas, the processing rules might state that all schemas are expected to be valid, 
+some are expected to be valid, or that a certain schema applies to a certain 
+portion of a credential.</p>
           </dd>
         </dl>
 
--- a/index.html
+++ b/index.html
@@ -2457,26 +2457,24 @@ the <a>verifiable credentials</a> that it issues:
         </p>
 
         <dl>
-          <dt><var>credentialSchema</var></dt>
+          <dt><var>`</var></dt>
           <dd><p>
 The value of the <code>credentialSchema</code> <a>property</a> MUST be a single
 data schema object which provides <a>verifiers</a> with enough information 
 to determine if the provided data conforms to the provided schema object. Each
 <code>credentialSchema</code> MUST specify its <code>type</code> (e.g.,
-<code>JsonSchemaValidator2018</code>, 
-<code><a href="https://w3c-ccg.github.io/vc-json-schemas/">CredentialSchema2022</a></code>,
-or <code><a href="https://transmute-industries.github.io/vc-credential-schema-open-api-specification/">
-OpenApiSpecificationValidator2022</a></code>), and an <code>id</code> <a>property</a>
-which MUST be a <a>URL</a> identifying the schema file. </p>
+<code><a href="https://w3c-ccg.github.io/vc-json-schemas/">CredentialSchema2022</a></code>)
+, and an <code>id</code> <a>property</a> which MUST be a <a>URL</a> identifying
+the schema file. </p>
 
 <p>The precise contents of each data schema object is determined by its 
 <code>type</code> definition. Different <code>type</code> implementations
 of data schema objects might choose to support different processing mechanisms 
 such as enabling the use of multiple data schemas for a single credential.
-Types might also enable specific processing rules. For example, if there are multiple 
-schemas, the processing rules might state that all schemas are expected to be valid, 
-some are expected to be valid, or that a certain schema applies to a certain 
-portion of a credential.</p>
+Types might also enable specific processing rules. For example, if there are
+multiple schemas, the processing rules might state that all schemas are
+expected to be valid, some are expected to be valid, or that a certain schema
+applies to a certain portion of a credential.</p>
           </dd>
         </dl>
 
--- a/index.html
+++ b/index.html
@@ -2471,7 +2471,7 @@ which MUST be a <a>URL</a> identifying the schema file. </p>
 
 <p>The precise contents of each data schema object is determined by its 
 <code>type</code> definition. Different <code>type</code> implementations
-of data schema objects may choose to support different processing mechanisms 
+of data schema objects might choose to support different processing mechanisms 
 such as enabling the use of multiple data schemas for a single credential.
 Types may also enable specific processing rules (e.g. if there are multiple 
 schemas, all must be valid, some must be valid, this schema applies to this


index.html
```diff
@@ -2458,14 +2458,24 @@ <h3>Data Schemas</h3>
...
-The value of the <code>credentialSchema</code> <a>property</a> MUST be one or
-more data schemas that provide <a>verifiers</a> with enough information to
-determine if the provided data conforms to the provided schema. Each
-<code>credentialSchema</code> MUST specify its <code>type</code> (for example,
-<code>JsonSchemaValidator2018</code>), and an <code>id</code> <a>property</a>
-that MUST be a <a>URL</a> identifying the schema file. The precise contents of
-each data schema is determined by the specific type definition.
+          <dd><p>
+The value of the <code>credentialSchema</code> <a>property</a> MUST be a single
+data schema object which provides <a>verifiers</a> with enough information
```
msporny, line 2463: Hmm, this makes it not possible to provide more than one type of schema... for example, one schema to do data validation (e.g., JSON Schema) and another schema to do data transformation (e.g., CL Signatures, Gordian Envelopes, etc.).

Limiting to ONE and only one schema is probably not a good idea. 

Limiting to one schema type in the list of `credentialSchema`s is probably better.

index.html
```diff
@@ -2458,14 +2458,24 @@ <h3>Data Schemas</h3>
...
-more data schemas that provide <a>verifiers</a> with enough information to
-determine if the provided data conforms to the provided schema. Each
-<code>credentialSchema</code> MUST specify its <code>type</code> (for example,
-<code>JsonSchemaValidator2018</code>), and an <code>id</code> <a>property</a>
-that MUST be a <a>URL</a> identifying the schema file. The precise contents of
-each data schema is determined by the specific type definition.
+          <dd><p>
+The value of the <code>credentialSchema</code> <a>property</a> MUST be a single
+data schema object which provides <a>verifiers</a> with enough information 
+to determine if the provided data conforms to the provided schema object. Each
```
msporny, line 2464: "Each" suggests multiple schemas?

index.html
```diff
@@ -2458,14 +2458,24 @@ <h3>Data Schemas</h3>
...
+<code>credentialSchema</code> MUST specify its <code>type</code> (e.g.,
+<code>JsonSchemaValidator2018</code>,
```
msporny, line 2466: We should remove `JsonSchemaValidator2018` since it never existed in practice and was just suggested as an example, IIRC.

index.html
```diff
@@ -2458,14 +2458,24 @@ <h3>Data Schemas</h3>
...
-<code>JsonSchemaValidator2018</code>), and an <code>id</code> <a>property</a>
-that MUST be a <a>URL</a> identifying the schema file. The precise contents of
-each data schema is determined by the specific type definition.
+          <dd><p>
+The value of the <code>credentialSchema</code> <a>property</a> MUST be a single
+data schema object which provides <a>verifiers</a> with enough information 
+to determine if the provided data conforms to the provided schema object. Each
+<code>credentialSchema</code> MUST specify its <code>type</code> (e.g.,
+<code>JsonSchemaValidator2018</code>, 
+<code><a href="https://w3c-ccg.github.io/vc-json-schemas/">CredentialSchema2022</a></code>,
```
msporny, line 2467: We should link to this one non-normatively.

index.html
```diff
@@ -2458,14 +2458,24 @@ <h3>Data Schemas</h3>
...
-determine if the provided data conforms to the provided schema. Each
-<code>credentialSchema</code> MUST specify its <code>type</code> (for example,
-<code>JsonSchemaValidator2018</code>), and an <code>id</code> <a>property</a>
-that MUST be a <a>URL</a> identifying the schema file. The precise contents of
-each data schema is determined by the specific type definition.
+          <dd><p>
+The value of the <code>credentialSchema</code> <a>property</a> MUST be a single
+data schema object which provides <a>verifiers</a> with enough information 
+to determine if the provided data conforms to the provided schema object. Each
+<code>credentialSchema</code> MUST specify its <code>type</code> (e.g.,
```
msporny, line 2465: Past Internationalization reviewers have asked us to not use "e.g.," and use "for example" instead. "e.g.," is difficult for some non-native English readers (or readers that are not familiar with Latin). I don't remember if we've done a scan of the spec to remove all examples of "e.g.," with "for example".

index.html
```diff
@@ -2458,14 +2458,24 @@ <h3>Data Schemas</h3>
...
+<code>JsonSchemaValidator2018</code>, 
+<code><a href="https://w3c-ccg.github.io/vc-json-schemas/">CredentialSchema2022</a></code>,
+or <code><a href="https://transmute-industries.github.io/vc-credential-schema-open-api-specification/">
+OpenApiSpecificationValidator2022</a></code>), and an <code>id</code> <a>property</a>
```
msporny, line 2469: We shouldn't link to vendor organization Github repos in standards-track specs because 1) not all vendor organizations stick around for as long as these specs do and the links break over time, and 2) competing vendors complain of low-key favoritism when we do this.

Ideally, all external links should point to stable documents that we expect to last for a decade or more.

index.html
```diff
@@ -2458,14 +2458,24 @@ <h3>Data Schemas</h3>
...
+which MUST be a <a>URL</a> identifying the schema file. </p>
+
+<p>The precise contents of each data schema object is determined by its 
+<code>type</code> definition. Different <code>type</code> implementations
+of data schema objects may choose to support different processing mechanisms
```
msporny, line 2474:
```suggestion
of data schema objects choose to support different processing mechanisms 
```

msporny: Avoid using normative language such as "MUST/SHOULD/MAY" as lowercase text. Inevitably reviewers get confused and ask the Editors: "Do you mean 'MAY' here?" or worse "You're not testing this normative MUST requirement, it's lowercase, you need to make it uppercase." We've discovered that over time, that eats into Editor's time maintaining the spec, so we try to avoid lowercase normative language if at all possible by using "expected to/might/can" and other language that can't be confused with a normative statement.

index.html
```diff
@@ -2458,14 +2458,24 @@ <h3>Data Schemas</h3>
...
+OpenApiSpecificationValidator2022</a></code>), and an <code>id</code> <a>property</a>
+which MUST be a <a>URL</a> identifying the schema file. </p>
+
+<p>The precise contents of each data schema object is determined by its 
+<code>type</code> definition. Different <code>type</code> implementations
+of data schema objects may choose to support different processing mechanisms 
+such as enabling the use of multiple data schemas for a single credential.
+Types may also enable specific processing rules (e.g. if there are multiple 
+schemas, all must be valid, some must be valid, this schema applies to this
+portion of a credential, and so on).</p>
```
msporny, line 2478:
```suggestion
Types might also enable specific processing rules. For example, if there are multiple 
schemas, the processing rules might state that all schemas are expected to be valid, 
some are expected to be valid, or that a certain schema applies to a certain 
portion of a credential.</p>
```

Clean up grammar, lowercase normative language, and "i.e.," usage.

msporny (changes requested):
Mostly editorial changes requested. The one item that's concerning is the limitation of only being able to associate a single schema with a VC, which removes composability of the `credentialSchema` feature.

index.html
OR13: > Limiting to ONE and only one schema is probably not a good idea.

If we don't limit to one, we must provide examples that make use of more than one, and tests...

I suggest limiting to one.

OR13 (approved):

index.html
OR13: The other reference is to the W3C CCG which also relies on "GitHub" for hosting.

OR13: Additional context, there are at least 4 companies using this approach today, its the same approach used by 
- https://github.com/w3c-ccg/traceability-vocab

index.html
OR13: 100% agree.

index.html
OR13: Yes, confused by this, and your other comments... I think we should start with 1, since thats closest to what we had in 1.1

OR13:
Assuming issues are filed for disagreements, this is an improvement and should be accepted.

index.html
decentralgabe: thanks, I saw there's a mix of e.g. and for example. I'll change this one.

index.html
decentralgabe: will remove.

index.html
decentralgabe: Sorry, had a hard time with this language. My intention is to say the "credentialSchema" property can only list one, though the value it resolves to can list 1 or many.

index.html
decentralgabe: Yes - suggest limiting to one schema type in the list of credential schema, which when resolved can contain multiple schemas

index.html
decentralgabe: How can I link to this non-normatively?

What about keeping it non-normative until v2 is shored up (in progress).

index.html
decentralgabe: @OR13 can you move it into the CCG and I'll add it back then?

index.html
brianorwhatever: I'm still having troubles parsing what you are saying.

are you saying this should be valid

```
"credentialSchema": [{"type": "CredentialSchema2022", ...}, {"type": "JsonSchemaValidator2018", ...}]
```

or this

```
"credentialSchema": {"type": ["CredentialSchema2022", "JsonSchemaValidator2018"], ...}
```

or something else entirely?

decentralgabe: neither. I'm saying

```
"credentialSchema": {"type": ["CredentialSchema2022"]}"
```

and the value of the `id` for `CredentialSchema2022` can contain multiple schemas

brianorwhatever: how can an `id` contain multiple things? The document that resolves from that `id` can have multiple schemas?

decentralgabe: yes, the document that resolves from the `id` can have multiple schemas as defined by the type

index.html
TallTed: Something similar should be done with instances of `i.e.`. One possible replacement is `in other words`, either parenthetically or (typically) preceded by a semicolon and followed by a comma.

index.html
```diff
@@ -2457,15 +2457,24 @@ <h3>Data Schemas</h3>
...
-that MUST be a <a>URL</a> identifying the schema file. The precise contents of
-each data schema is determined by the specific type definition.
+          <dt><var>`</var></dt>
+          <dd><p>
+The value of the <code>credentialSchema</code> <a>property</a> MUST be a single
+data schema object which provides <a>verifiers</a> with enough information 
+to determine if the provided data conforms to the provided schema object. Each
+<code>credentialSchema</code> MUST specify its <code>type</code> (e.g.,
+<code><a href="https://w3c-ccg.github.io/vc-json-schemas/">CredentialSchema2022</a></code>)
+, and an <code>id</code> <a>property</a> which MUST be a <a>URL</a> identifying
```
TallTed, line 2467:
```suggestion
<a href="https://w3c-ccg.github.io/vc-json-schemas/">`CredentialSchema2022`</a>),
and an `id` <a>property</a> which MUST be a <a>URL</a> identifying
```

index.html
decentralgabe: Thanks both, created https://github.com/w3c/vc-data-model/issues/1025 to track

David-Chadwick: I am having difficult determining where we have  got to with this clarification. First, I had no problem understanding the original language. Second, we already have examples e.g. JFF plugfest, or using different schema languages (i.e.types) to define the same VC data structure. So if type does equal the schema language, then each type should have a single ID (URL) where the schema can be downloaded from. So what exactly is the change that is being proposed now? Since I cannot see how one URL/ID can or should point to different schema files.

David-Chadwick (changes requested):

index.html
decentralgabe: @David-Chadwick see this for an example https://github.com/w3c-ccg/vc-json-schemas/issues/121

index.html
OR13: @decentralgabe we are not required to "move things to CCG" by W3C Process. 

We can adopt them directly as work items, as a working group.

I suggest adding a note saying it will be removed in future draft if its not adopted as a work item, to both links.

index.html
TallTed:
```suggestion
to determine whether the provided data conforms to the provided schema object. Each
```

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-02-01-vcwg#section3-5) on 2023-02-01

- no resolutions were taken

#### 3.5. Add language clarifying multiple data schemas (pr vc-data-model#1023)

_See github pull request [vc-data-model#1023](https://github.com/w3c/vc-data-model/pull/1023)._




**Manu Sporny:** New one PR 1023 from yesterday, about credential schemas. This PR says you can have only one credential schema, and then there's some clarifying language. Please jump in if you have opinions on credential schema..  
… This is all about VC data model..  

> *Manu Sporny:* [https://github.com/w3c/vc-status-list-2021/pulls](https://github.com/w3c/vc-status-list-2021/pulls).

**Manu Sporny:** Moving on to vc-status-list-2021 PRs.  

</details>

matthieubosquet (approved):
Delegating composability to an hypothetic future implementation type of schema validation seems ok to me.

I’m not sure it requires an entire paragraph (since cardinality is clear, I think it would be ok to remove that note/the second paragraph altogether).

The cardinality of 1 schema is clear to me.

index.html
decentralgabe: Waiting for you and @msporny to have consensus before proceeding here

David-Chadwick:
yes happy to

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2022-09-15-vcwg#section5-4) on 2022-09-15

- no resolutions were taken

#### 5.4. Verification of multiple credential schemas (issue vc-data-model#950)

_See github issue [vc-data-model#950](https://github.com/w3c/vc-data-model/issues/950)._




**Gabe Cohen:** There is a PR.  

_See github pull request [vc-data-model#1023](https://github.com/w3c/vc-data-model/pull/1023)._




</details>

index.html
aphillips: The I18N WG is not aware of our having--now or in the past--a policy against (mostly Latin derived) abbreviated terms common to English such as `i.e.`, `e.g.`, `cf.`, `etc.` and so forth. If one or another of us made this comment in the past, this is to say: we didn't mean it/sorry.

I do note that some writers or organizations have style guides that prefer spelled out English terms (for clarity in English), but I18N does not take a position on this. I double-checked the [W3C Manual of Style](https://www.w3.org/Guide/manual-of-style/) and it is silent on this topic, so I suppose it is not a W3C policy either.

msporny: *cries* it came up during the VC 1.0 review and I had to go through the entire spec to make the conversion, IIRC. I'd have to go and track it down... which, as you state above, would probably not change the guidance at this point.

So, now this is a VCWG / Editor's decision -- how do we want to be consistent here... always use abbreviated latin terms "e.g.," and "i.e.," or always use the spelled out English equivalents?

decentralgabe: I think it's simpler to use latin. There will be less variation since we'll need to pick common english replacements which don't necessarily come to mind without looking at examples.

decentralgabe: resolved by the discussion in #1045

decentralgabe:
After today's call I'm ready to close this PR and move back to discussion in the issue #950 

Do you think that's reasonable - @dlongley @dmitrizagidulin @OR13 ?

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-05-03-vcwg#section3-2) on 2023-05-03

- no resolutions were taken

#### 3.2. Add language clarifying multiple data schemas (pr vc-data-model#1023)

_See github pull request [vc-data-model#1023](https://github.com/w3c/vc-data-model/pull/1023)._




**Brent Zundel:** nobody wanted to speak to that.  
… next up is PR 1023.  
… gabe, what needs to be done to move this forward?  

> *Kristina Yasuda:* PR 1035 should be closed when rendering gets added to the reserved table.

**Gabe Cohen:** some discussion whether there should be a single credential schema possible per credential.  
… or something else.  
… i want to have one defined in the VCDM.  
… but need to discuss what people think.  

**Dave Longley:** we already have mechanism for starting a set.  
… for the case where we have more than one thing in the data model.  
… if we are saying if we want to support either way, ability to have multiple schemas in a VC, i would expect to use that mechanism.  

> *Orie Steele:* The current JSON-LD context does not define credentialSchema as a container... [https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v2#L18](https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v2#L18).

**Dave Longley:** if not, if different credential schema types can become multitype schemas then i would expect that somebody creates a multi purpose mechanism.  
… i don't know if we need to invent a new thing to do that.  
… not sure about the arguments and what are tradeoffs?  

**Gabe Cohen:** not exactly an accurate representation.  
… arguing for keeping complexitiy in specific type specifications.  
… that is the purpose of having types for defining processing logic.  
… want to avoid complextiy in data model.  

**Orie Steele:** agree with gabe.  
… dave is aruging that json-ld expansion can handle all of that.  
… how to deal with polymorphism.  
… let's not do any json-ld processing.  

**Brent Zundel:** if i understand correctly if multiple schemas are present and the processing rules that then apply that we limit schemas to a single schema differentiated by types, so processing rules for single schema can be defined by the type and we also recognize that multiple schemas can be developed by having a composed schema.  
… fair summary?  
… is this something we can come to some agreement?  

> *Gabe Cohen:* correct brent.

**Dave Longley:** i don't know where i am at on this issue in particular.  
… json-ld is sort of red herring.  
… json-ld processors will find out a way to handle that no matter what we put into the spec.  

> *Orie Steele:* +1 dlongley.

**Dave Longley:** whether or not schemas are used is up to verifier.  
… we can keep the data model simple and we can still put specific rules in specific type spec.  
… and then up to verifier if follows rule or not.  

> *Orie Steele:* We are basically saying what is legal value for `credentialSchema`... object, array string or all of them.

> *Orie Steele:* Sounds like we don't need to say anything if the verifier just knows how to do JSON-LD processing.

**Oliver Terbu:** I am in favor of keeping complexity on the `type` side.  

**Dmitri Zagidulin:** my initial thought was that multiple items in cred schema was initially confusing. but then realized that.  
… easy to understand example is different schema languages where multiple schemas would make sense.  

> *Dave Longley:* +1 to dmitriz, multiple schema languages.

**Dmitri Zagidulin:** shacl schema or json schema or whatever schema language other come up with.  

> *Gabe Cohen:* dmitri, that can easily be handled *within* a type.

> *Dave Longley:* decentralgabe: a new "multi" type will just need to be invented.

**Dmitri Zagidulin:** question is if we need to specify and or or semantic for that field.  

> *Orie Steele:* sure, so we say `credentialSchema` supports array, object or string... and let `type` handle multiple langauges?

> *Orie Steele:* and the core data model does not comment on the complexity?

> *Gabe Cohen:* yes, multi type `->` complex rules in the data model.

> *Dmitri Zagidulin:* Orie - that seems reasonable.

> *Dave Longley:* data model doesn't need those rules regardless.

> *Orie Steele:* seems like we are really just saying "the core data model is JSON-LD" with extra words.

**Brent Zundel:** encourage folks to have that conversation.  

> *Dmitri Zagidulin:* decentralgabe - re "multiple schemas can be handled within a type" -- yeah, that's true too.

**Brent Zundel:** we would like to have this PR to move forward.  

</details>

brentzundel:
> After today's call I'm ready to close this PR and move back to discussion in the issue #950
> 
> Do you think that's reasonable - @dlongley @dmitrizagidulin @OR13 ?

I going to mark this as `pending close` and close in one week if no one objects.

OR13 (changes requested):
This PR has conflicts and based on the discussion in the last working group call, it seems better to just let JSON-LD norms build how this property can be used.

This means the following will all remain valid:

```json
"credentialSchema": "IRI",
```

```json
"credentialSchema": [ { "id": "IRI", "type": "Class", ...}, { "id": "IRI-2", "type": "Class-2", ...} ],
```

```json
"credentialSchema": { "id": "IRI", "type": "Class", ...},
```

decentralgabe:
@OR13 I'm open to this change but then I believe we need to add language such as
> If multiple credential schema types are provided they must be validated in order. Any single failure results in a validation failure for the `credentialSchema` property.

OR13:
@decentralgabe we don't need to add that language, because it's already assumed people understand how JSON-LD works.

We say the base media type is JSON-LD and the the standard uses compact form.

If people don't understand what that means, they cannot implement the spec.

If they understand it, they can do anything that is legal to a parser that processes the contexts

decentralgabe:
@OR13 I'm not sure I fully understand how JSON-LD works. If we cannot guarantee processing order, will we still need to state how to interpret the results of processing multiple credential schemes? If not - how will we make sure verifiers have consistent results?

iherman:
> This PR has conflicts and based on the discussion in the last working group call, it seems better to just let JSON-LD norms build how this property can be used.
> 
> This means the following will all remain valid:
> 
> ```json
> "credentialSchema": "IRI",
> ```
> 
> ```json
> "credentialSchema": [ { "id": "IRI", "type": "Class", ...}, { "id": "IRI-2", "type": "Class-2", ...} ],
> ```
> 
> ```json
> "credentialSchema": { "id": "IRI", "type": "Class", ...},
> ```

While this works and is correct, I am not sure it is the best choice. The usage of `type` has a very specific meaning in JSON-LD, and would require, if we play by the rules, to define a unique URL (or add it to one of our vocabularies) for `Class`, `Class-2`, etc. This is doable, but sounds like an overkill to me. 

We may, instead, use a different key there (`schema_type`?), which is a property we can define and we could say that it has some predefine string values (`json schema`, `json type definition`, `shacl`, etc).

iherman:
>>@OR13 I'm open to this change but then I believe we need to add language such as
>>> If multiple credential schema types are provided they must be validated in order. Any single failure results in a validation failure for the `credentialSchema` property.
>
> @decentralgabe we don't need to add that language, because it's already assumed people understand how JSON-LD works.
> 
> We say the base media type is JSON-LD and the the standard uses compact form.
> 
> If people don't understand what that means, they cannot implement the spec.
> 
> If they understand it, they can do anything that is legal to a parser that processes the contexts

I am not sure I understand this remark. JSON-LD is oblivious to this, which also means that it does not provide any guidance per se.

If `credentialSchema` is "just" used without further ado, the order of the schemas listed in the right hand side does not count. If `credentialSchema` is defined with range being a List (ie, it is defined, in JSON-LD terminology, as a list container) then the order counts. It is the specification's job to say what we want, then we can map the intention to JSON-LD.

(That being said, do we really want such a strong requirement, @decentralgabe? Alternatively, we _could_ say that if any of the schema validate, we are good...)

decentralgabe:
Closing since there isn't consensus. Let's please continue the discussion here https://github.com/w3c/vc-data-model/issues/950#issuecomment-1536493084

PR #1014, opened 2023-01-20
Add normative requirements regarding proof

OR13:

PR #1033, opened 2023-02-10
Add guidance for `credential+ld+json` for `external`proof

OR13:
Remove proposed guidance that lacked consensus per @dlongley on: https://github.com/w3c/vc-data-model/pull/1014

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1033.html" title="Last updated on Mar 6, 2023, 10:30 PM UTC (b0d7938)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1033/034e5d0...b0d7938.html" title="Last updated on Mar 6, 2023, 10:30 PM UTC (b0d7938)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5133,7 +5133,7 @@ Extensions of Verifiable Credentials Data Model 1.0</a>
   </p>
 
   <p>
-    This media type MUST NOT be used with <dfn>embedded proof</dfn>.
+    This media type MUST NOT be used to describe a verifiable credential with an <dfn>embedded proof</dfn>.
   </p>
   <p>
     This media type MAY be used with <dfn>external proof</dfn>.
--- a/index.html
+++ b/index.html
@@ -5135,7 +5135,7 @@ Extensions of Verifiable Credentials Data Model 1.0</a>
   <p>
     This media type can be used with credentials secured using an <a>external proof</a>.
   </p>
-  <p>If a [[JSON-LD]] context is present in the body of the document, and indicated by the presence of `ld+json` in the media type, the credential MUST be a valid <a href="https://www.w3.org/TR/json-ld11/#dfn-json-ld-document">JSON-LD document</a>.</p>
+  <p>A [[JSON-LD]] context is expected to be present in the body of the document, and as indicated by the presence of `ld+json` in the media type, the credential is expected to be a valid <a href="https://www.w3.org/TR/json-ld11/#dfn-json-ld-document">JSON-LD document</a>.</p>
 </section>
 
 
--- a/index.html
+++ b/index.html
@@ -5133,9 +5133,6 @@ Extensions of Verifiable Credentials Data Model 1.0</a>
   </p>
 
   <p>
-    This media type MUST NOT be used to describe a verifiable credential with an <dfn>embedded proof</dfn>.
-  </p>
-  <p>
     This media type MAY be used with <dfn>external proof</dfn>.
   </p>
 </section>
--- a/index.html
+++ b/index.html
@@ -5139,7 +5139,6 @@ Extensions of Verifiable Credentials Data Model 1.0</a>
   <p>
     This media type MAY be used with <dfn>external proof</dfn>.
   </p>
-
 </section>
 
 
--- a/index.html
+++ b/index.html
@@ -5133,7 +5133,7 @@ Extensions of Verifiable Credentials Data Model 1.0</a>
   </p>
 
   <p>
-    This media type can be used with credentials secured using an <dfn>external proof</dfn>.
+    This media type can be used with credentials secured using an <a>external proof</a>.
   </p>
   <p>If a [[JSON-LD]] context is present in the body of the document, and indicated by the presence of `ld+json` in the media type, the credential MUST be a valid <a href="https://www.w3.org/TR/json-ld11/#dfn-json-ld-document">JSON-LD document</a>.</p>
 </section>
--- a/index.html
+++ b/index.html
@@ -5135,7 +5135,6 @@ Extensions of Verifiable Credentials Data Model 1.0</a>
   <p>
     This media type MUST NOT be used with <dfn>embedded proof</dfn>.
   </p>
-
   <p>
     This media type MAY be used with <dfn>external proof</dfn>.
   </p>
--- a/index.html
+++ b/index.html
@@ -5132,6 +5132,14 @@ Extensions of Verifiable Credentials Data Model 1.0</a>
     the use of a specific media type.
   </p>
 
+  <p>
+    This media type MUST NOT be used with <dfn>embedded proof</dfn>.
+  </p>
+
+  <p>
+    This media type MAY be used with <dfn>external proof</dfn>.
+  </p>
+
 </section>
 
 
--- a/index.html
+++ b/index.html
@@ -5133,7 +5133,7 @@ Extensions of Verifiable Credentials Data Model 1.0</a>
   </p>
 
   <p>
-    This media type MAY be used with <dfn>external proof</dfn>.
+    This media type can be used with credentials secured using an <dfn>external proof</dfn>.
   </p>
 </section>
 
--- a/index.html
+++ b/index.html
@@ -5135,6 +5135,7 @@ Extensions of Verifiable Credentials Data Model 1.0</a>
   <p>
     This media type can be used with credentials secured using an <dfn>external proof</dfn>.
   </p>
+  <p>If a [[JSON-LD]] context is present in the body of the document, and indicated by the presence of `ld+json` in the media type, the credential MUST be a valid <a href="https://www.w3.org/TR/json-ld11/#dfn-json-ld-document">JSON-LD document</a>.</p>
 </section>
 
 


dlongley (approved):

index.html
```diff
@@ -5132,6 +5132,9 @@ <h2>The <code>application/credential+ld+json</code> Media Type</h2>
     the use of a specific media type.
   </p>
 
+  <p>
+    This media type MAY be used with <dfn>external proof</dfn>.
```
dlongley, line 5136: The improved suggestion from @msporny in the other PR:

```suggestion
    This media type can be used with credentials secured using an <dfn>external proof</dfn>.
```

index.html
```diff
@@ -5132,6 +5132,9 @@ <h2>The <code>application/credential+ld+json</code> Media Type</h2>
     the use of a specific media type.
   </p>
 
+  <p>
```
mprorock, line 5135:
```suggestion
  <p>
    This media type MUST NOT be used to describe a verifiable credential with an <dfn>embedded proof</dfn>.
  </p>
  <p>
```

index.html
mprorock, line 5136:
```suggestion
    This media type MAY be used with credentials secured using an <dfn>external proof</dfn>.
```

mprorock (changes requested):

mprorock:
added some suggestions to make it very clear what can and can't be done to make things very simple and hopefully avoid 1) confusion, 2) resulting type confusion issues (i.e. like what we keep seeing in x509 and related openssl CVEs)

dlongley:
> added some suggestions to make it very clear what can and can't be done to make things very simple and hopefully avoid 1) confusion, 2) resulting type confusion issues (i.e. like what we keep seeing in x509 and related openssl CVEs)

Of course, adding restrictions do define what can and can't be done. However, the specific restrictions added should have a clear rationale and be technologically defensible.

So, the specific suggestion that you cannot use this media type with a verifiable credential that has an embedded proof ... serves what purpose? Where is that useful and what are the pros / cons?

index.html
OR13: We have similar text with strong support on https://github.com/w3c/vc-data-model/pull/1014

I created this PR assuming @dlongley and @msporny might accept it if this text was not present.

OR13: I won't be accepting this suggestion.

mprorock (approved):

mprorock:
> > added some suggestions to make it very clear what can and can't be done to make things very simple and hopefully avoid 1) confusion, 2) resulting type confusion issues (i.e. like what we keep seeing in x509 and related openssl CVEs)
> 
> Of course, adding restrictions do define what can and can't be done. However, the specific restrictions added should have a clear rationale and be technologically defensible.
> 
> So, the specific suggestion that you cannot use this media type with a verifiable credential that has an embedded proof ... serves what purpose? Where is that useful and what are the pros / cons?

converting to approve - did not mean to take discussion out of #1014

msporny (approved):

index.html
```diff
@@ -5132,6 +5132,9 @@ <h2>The <code>application/credential+ld+json</code> Media Type</h2>
...
+    This media type can be used with credentials secured using an <dfn>external proof</dfn>.
+  </p>
```
mprorock, line 5137:
```suggestion
  </p>
  <p>If a [[json-ld]] context is present and indicated by the presence of `+ld` in a media type, the credential MUST be valid <a href="https://www.w3.org/TR/json-ld11/#dfn-json-ld-document">JSON-LD document</a>.</p>
```

mprorock: might not be the right place for this..

OR13: Seems like the correct place per: https://www.w3.org/TR/json-ld11/#interpreting-json-as-json-ld

Screen Shot 2023-02-11 at 10 08 36 AM

OR13: I wonder if we need to mirror the guidance on the link header here as well.

msporny: Three things:

1. We should say "by the presence of `+ld+json`", because there is no such thing as a `+ld` anything, IIRC.
2. We haven't registered [a structured suffix](https://www.iana.org/assignments/media-type-structured-suffix/media-type-structured-suffix.xhtml) for `+ld+json` yet... I don't think that's an issue, but the IETF structured suffix lawyers (or like minded individuals) might grumble.
3. Someone will insist that we test the `MUST` statement (as we do for all of our testable MUST statements), or insist that to be conformant that implementations MUST perform JSON-LD processing... I don't think we want either of those things, so suggest something like "The credential will pass a JSON-LD document conformance checker." (we need better text than that... the point being not to invoke the "How are you going to test this in the test suite!? Does this mean that implementations must perform JSON-LD conformance checking!?" line of questioning).

Don't know if we really need this sort of language... but I'm not opposed to it being somewhere in the spec.

msporny (approved):
LGTM. Don't know about @mprorock's addition... on the fence about that one.

index.html
TallTed: Pulling together @mprorock's initial insert, with the comments by @OR13 and @msporny, plus a bit of my own special sauce:
```suggestion
  </p>
  <p>If a [[JSON-LD]] context is present in the body of the document, and indicated by the presence of `ld+json` in the media type, the credential MUST be a valid <a href="https://www.w3.org/TR/json-ld11/#dfn-json-ld-document">JSON-LD document</a>.</p>
```

OR13: I would like to commit this suggestion, are there any objections to doing so?

msporny: No objections in 5 days, though I'm concerned about the number of people that have /not/ reviewed this PR.

iherman (changes requested):
This PR seems to be unfinished. It refers to the term of "external proof", but it does not define what this means. Until that is properly defined, this is incomplete and should not be merged imho.

OR13:
@iherman external proof is defined here: 

https://w3c.github.io/vc-data-model/#proofs-signatures

However I agree that `proof` is in general not defined sufficiently in the core data model.

Which is why people can't understand its relationship to `credential` vs `verifiable credential`.

index.html
TallTed: @msporny — (Re-)request more reviews?

index.html
```diff
@@ -5132,6 +5132,10 @@ <h2>The <code>application/credential+ld+json</code> Media Type</h2>
     the use of a specific media type.
   </p>
 
+  <p>
+    This media type can be used with credentials secured using an <dfn>external proof</dfn>.
```
iherman, line 5136:
```suggestion
    This media type can be used with credentials secured using an <a>external proof</a>.
```

iherman (changes requested):

iherman:
> @iherman external proof is defined here:
> 
> https://w3c.github.io/vc-data-model/#proofs-signatures
> 

I stand corrected. But https://github.com/w3c/vc-data-model/pull/1033#discussion_r1117883393 explains why I went wrong: in the new section the term "external proof" appeared as a definition and not as a reference.

OR13:
@iherman accepted your change suggestion, can you please re-review

iherman (approved):

index.html
```diff
@@ -5132,6 +5132,10 @@ <h2>The <code>application/credential+ld+json</code> Media Type</h2>
     the use of a specific media type.
   </p>
 
+  <p>
+    This media type can be used with credentials secured using an <a>external proof</a>.
+  </p>
+  <p>If a [[JSON-LD]] context is present in the body of the document, and indicated by the presence of `ld+json` in the media type, the credential MUST be a valid <a href="https://www.w3.org/TR/json-ld11/#dfn-json-ld-document">JSON-LD document</a>.</p>
```
dlongley, line 5138: Given recent resolutions, I don't think we need this statement at all:

```suggestion
```

We could alternatively change it to:

```suggestion
  <p>A [[JSON-LD]] context is expected to be present in the body of the document, and as indicated by the presence of `ld+json` in the media type, the credential MUST be a valid <a href="https://www.w3.org/TR/json-ld11/#dfn-json-ld-document">JSON-LD document</a>.</p>
```

OR13: @dlongley your changes seem to make `context` optional, since they change the text from `required` to `expected`.

Is this your intention?

brentzundel (approved):

index.html
dlongley: > your changes seem to make `context` optional, since they change the text from `required` to `expected`.

No, it doesn't make it optional, it just acknowledges that you cannot expect someone who sends you data tagged with media type `application/anything` to have complied with the requirements of that media type. The fact that it's expected is because it's required. It's a more terse way of writing "Since it's required, parsers will expect it, so make sure you include it as a sender. Of course, as a parser, you should be aware that not every sender will comply, so be ready to reject bad data".

OR13: I would prefer something to the effect of:

"A context MUST be present in any VCDM media types ending in `+ld+json`."

dlongley: @OR13,

I'm fine with that `MUST` language in a normative section of the spec. That the document is expected to be a valid JSON-LD document comes with the `+ld+json` suffix, but if we don't get that for free (yet), it's also fine to say (or we could use MUST language there as well). People will have opinions around what tests would need to come with this. I think, at this point, we're meaning the same things and we just need everyone to agree on the right way and place to say it.

OR13: I am going to merge your suggestion after a while if there are no objections.

OR13: The suggestion we are threaded on: https://github.com/w3c/vc-data-model/pull/1033#discussion_r1123362533

msporny: I'm trying to process this PR, at present, I don't know how long "after a while if there are no objections" is.

I also don't know if @dlongley's concrete language is the change proposal above, or if it could be considered a part of #1055, or something else.

Help me understand 1) what the final language for this PR is going to be, and 2) when we can merge it.

OR13: I was waiting to press the "accept suggestion" button, which I just pressed: https://github.com/w3c/vc-data-model/pull/1033#discussion_r1123362533

It automatically marks the thread as resolved.

msporny:

PR #1042, opened 2023-02-14
Replace JsonSchemaValidator2018 (undefined) with CredentialSchema2022

decentralgabe:
This is a replacement of an unknown draft to a current CCG draft. Separately, I am interested in in bringing CredentialSchema2022 into this working group.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1042.html" title="Last updated on Mar 1, 2023, 11:58 PM UTC (8961924)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1042/fe54b80...8961924.html" title="Last updated on Mar 1, 2023, 11:58 PM UTC (8961924)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2463,7 +2463,7 @@ The value of the <code>credentialSchema</code> <a>property</a> MUST be one or
 more data schemas that provide <a>verifiers</a> with enough information to
 determine if the provided data conforms to the provided schema. Each
 <code>credentialSchema</code> MUST specify its <code>type</code> (for example,
-<code><a href="https://w3c-ccg.github.io/vc-json-schemas/">CredentialSchema2022</a></code>), and an <code>id</code> <a>property</a>
+<code>CredentialSchema2022</code>), and an <code>id</code> <a>property</a>
 that MUST be a <a>URL</a> identifying the schema file. The precise contents of
 each data schema is determined by the specific type definition.
           </dd>
--- a/index.html
+++ b/index.html
@@ -2463,7 +2463,7 @@ The value of the <code>credentialSchema</code> <a>property</a> MUST be one or
 more data schemas that provide <a>verifiers</a> with enough information to
 determine if the provided data conforms to the provided schema. Each
 <code>credentialSchema</code> MUST specify its <code>type</code> (for example,
-<code>JsonSchemaValidator2018</code>), and an <code>id</code> <a>property</a>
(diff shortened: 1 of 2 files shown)


OR13 (approved):

TallTed (approved):
LGTM

index.html
```diff
@@ -2463,7 +2463,7 @@ <h3>Data Schemas</h3>
 more data schemas that provide <a>verifiers</a> with enough information to
 determine if the provided data conforms to the provided schema. Each
 <code>credentialSchema</code> MUST specify its <code>type</code> (for example,
-<code>JsonSchemaValidator2018</code>), and an <code>id</code> <a>property</a>
+<code><a href="https://w3c-ccg.github.io/vc-json-schemas/">CredentialSchema2022</a></code>), and an <code>id</code> <a>property</a>
```
msporny, line 2466:
```suggestion
<code>CredentialSchema2022</code>), and an <code>id</code> <a>property</a>
```

This link is likely to break over the years, so the suggestion is to not link to it. It's somewhat unavoidable to link to it non-normatively in the appendix... with the assumption that the bibliography will be updated over time, and eventually will go into the ReSpec bibliography database, which can then point people to a newer spec if the link dies.

Either that, or wait until there is a stable link at W3C for CredentialSchema2022 and then point it that.

msporny (changes requested):
Overall, LGTM with a minor nit on hand-specifying links to non-normative documents (ideally, use the bibliography to do this so you only have to update the link in one place when it changes... like when it moves from CCG to VCWG).

The other thing to consider is that the spec is really about JSON Schema validation... but that said, it doesn't seem like there is any other schema language validator vying for attention this year in the VC ecosystem. It might become a problem if CBOR Schema validation is the new hotness in 2-3 years, but if that happens, you'd probably see a "CborCredentialSchema2026" spring into existence. Non-blocking food for thought.

index.html
decentralgabe: Agree, will update this.

decentralgabe: removed href

brentzundel (approved):
while Credential Schema 2022 remains non-normative, shouldn't the references to it also all be non-normative?
I see that multiple references to JSON Schema 2018 are done normatively (i.e., without the leading '?'). I believe this was an oversight that we currently have the opportunity to correct along with the other changes made in this PR.

decentralgabe:
Good call @brentzundel I didn't know about the `?` trick. Just updated.

msporny (approved):

msporny:

Issue #935, opened 2022-09-17
Add "credential boilerplates" as examples


OR13:
Developers copy spec examples, we should plan to have better boilerplates in v2, here is an example:

```json
{
  "@context": [
    "https://www.w3.org/ns/credentials/v2",
    { "@vocab": "https://www.w3.org/ns/credentials#" }
  ],
  "id": "http://example.edu/credentials/1872",
  "type": ["VerifiableCredential", "NewCredentialType"],
  "issuer": "did:example:123",
  "issuanceDate": "2010-01-01T19:23:24Z",
  "credentialSubject": {
    "id": "did:example:456",
    "claimName": "claimValue"
  }
}
```

OR13:
Here is the minimal VCDM 1.1 boilerplate for the record:

```json
{
  "@context": "https://www.w3.org/2018/credentials/v1",
  "id": "http://example.edu/credentials/1872",
  "type": ["VerifiableCredential"],
  "issuer": "did:example:123",
  "issuanceDate": "2010-01-01T19:23:24Z",
  "credentialSubject": {
    "id": "did:example:456"
  }
}
```

David-Chadwick:
Referring to my diagram in issue #982 , the first four are credential metadata properties and the last one is the credential.

OR13:
I believe this is now valid:


```json
{
  "@context": [
    "https://www.w3.org/ns/credentials/v2"
  ],
  "id": "http://example.edu/credentials/1872",
  "type": ["VerifiableCredential", "NewCredentialType"],
  "issuer": { 
    "id": "did:example:123", 
     "type": ["Organization", "OrganizationType"] 
   },
  "issuanceDate": "2010-01-01T19:23:24Z",
  "credentialSubject": {
    "id": "did:example:456", 
    "type": ["Person", "JobType"],
    "claimName": "claimValue"
  }
}
```

OR13:
- https://www.w3.org/ns/credentials/v2 (404)... cc @iherman @pchampin @msporny

pchampin:
> * https://www.w3.org/ns/credentials/v2 (404)... cc @iherman @pchampin @msporny

According to our CVS log, this URL has never yet returned anything. If I'm wrong, then I'm missing something big time.

I'm assuming the plan is to put there the content generated from https://github.com/w3c/vc-data-model/tree/main/vocab/credentials/v2, right?

OR13:
[...]

Issue #758, opened 2020-12-12
Fix the vc-data-model namespace with https://www.w3.org/ns/vc/v2


OR13:
In DID Core we have `https://www.w3.org/ns/did/v1`... in vc-data-model we have `https://www.w3.org/2018/credentials/v1`... this is confusing for developers.

I propose we fix this with vc-data-model v2 by creating:

`https://www.w3.org/ns/vc/v2`

kdenhartog:
This would break a lot of things in validating VC's today. I'm going to propose that we address this in V2 instead and leave it outside the scope of the maintenance WG.

msporny:
This PR is related: https://github.com/w3c/vc-data-model/pull/927

OR13:
This PR is addressed by: https://github.com/w3c/vc-data-model/pull/927

OR13:
Should be closed pending PR above.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2022-09-07-vcwg#section3-1) on 2022-09-07

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 3.1. Fix the vc-data-model namespace with https://www.w3.org/ns/vc/v2 (issue vc-data-model#758)

_See github issue [vc-data-model#758](https://github.com/w3c/vc-data-model/issues/758)._




**Kristina Yasuda:** Issue: Fix the data model namespace.  

**Orie Steele:** It seems that someone has fixed some parts of this recently.  
… There are term definitions that we're working to create.  
… We want to be able to read the definitions for those terms.  
… You can read them in the registry where they're specified.  
… You want to be able to click the link for a term definition and read the definition.  
… We should make sure that our term links are clickable.  

**Ivan Herman:** This is exactly what the vocabulary work does.  
… Now it's such that if you take the URL of a term, it will go there.  
… Temporarily, the links go to the V1 vocabulary.  
… Eventually we will point to the official V2 vocabulary.  

> *Orie Steele:* I suggest closing after PR is merged.

**Ivan Herman:** I think this issue can be closed.  

> *Manu Sporny:* We should wait until PR is merged.

**Kristina Yasuda:** We will close it after the PR is merged.  

</details>

msporny:
In an attempt to resolve this issue once and for all:

[...]

PR #1049, opened 2023-02-17
Setting up a user dependent namespace document

iherman:
The latest version of the `@context` file includes a reference, via a `@vocab`, to the `https://www.w3.org/ns/credentials/issuer-dependent` URL. At present, this URL leads to a 404, which is a big no-no. This PR creates a short "namespace document" alongside the VCDM vocabulary at `https://github.com/w3c/vc-data-model/tree/main/vocab/credentials/v2/issuer-dependent.html` (the file can be [previewed here](https://raw.githack.com/w3c/vc-data-model/issuer-dependent-namespace-document/vocab/credentials/v2/issuer-dependent.html)) which provides a simple landing place.

Although not part, physically, of the PR, I also have on my local disc a `.htaccess` file that is ready to be pushed onto `https://www.w3.org/ns/credentials/` and which looks as follows:

```
RewriteEngine On
RewriteBase /ns/credentials/
AddType application/ld+json .jsonld
AddType text/turtle .ttl


RewriteRule ^v2$ https://w3c.github.io/vc-data-model/contexts/credentials/v2 [E=json,P]
RewriteRule ^issuer-dependent$ https://w3c.github.io/vc-data-model/vocab/credentials/v2/issuer-dependent.html [P]
```

which takes care of mapping the `@vocab` URL to the file set up by this PR and, incidentally, also makes the mapping of the latest context file from `/ns/`, as requested in #935 (and as requested by @OR13). Note that the `.htaccess` filed does ***not*** map the core vocabulary file from `https://www.w3.org/ns/credentials/`, because that is still a pending issue #758. If the decision in #758 is to make that change, then the aforementioned `.htaccess` file can be extended accordingly.

### DIFF
--- a/vocab/credentials/v2/issuer-dependent.html
+++ b/vocab/credentials/v2/issuer-dependent.html
@@ -66,13 +66,13 @@
     <section>
         <h2>Issuer Dependent Terms</h2>
         <p>
-            The <a href="https://www.w3.org/TR/vc-data-model-2.0/#base-context">base context</a> of a Verifiable Credential, as defined in [vc-data-model], provides a mapping from the standard terms used in credentials to the vocabulary items defined by the Verifiable Credentials Working Group through different vocabularies, for example the <a href="https://www.w3.org/2018/credentials/">Verifiable Credentials Vocabulary</a>. Applications using additional, user-dependent vocabularies of their own are encouraged to create their own formal vocabularies, and corresponding JSON-LD context definitions, to ensure the interoperability of the credentials they produce and/or consume.
+            The <a href="https://www.w3.org/TR/vc-data-model-2.0/#base-context">base context</a> of a Verifiable Credential, as defined in [vc-data-model], provides a mapping from the standard terms used in credentials to the vocabulary items defined by the Verifiable Credentials Working Group through different vocabularies, such as the <a href="https://www.w3.org/2018/credentials/">Verifiable Credentials Vocabulary</a>. Applications using additional, user-dependent vocabularies of their own are encouraged to create their own formal vocabularies, and corresponding JSON-LD context definitions, to ensure the interoperability of the credentials they produce and/or consume.
         </p>
         <p>
-            There may be cases, however, when the creation of such formal vocabularies and context definitions is too much of a burden; this is the case for closed applications, experiments, or examples. The way the <a href="https://www.w3.org/TR/vc-data-model-2.0/#base-context">base context</a> has been constructed is such that issuers may use new, hitherto undefined terms without further ado, and these terms will be mapped on URL-s of the form <code>https://www.w3.org/ns/credentials/issuer-dependent#TERM</code>.
+            There may be cases, however, when the creation of such formal vocabularies and context definitions is too much of a burden, such as for closed applications, experiments, or examples. The <a href="https://www.w3.org/TR/vc-data-model-2.0/#base-context">base context</a> has been constructed in such a way that issuers may use new, hitherto undefined terms without further ado, and these terms will be mapped onto URLs of the form <code>https://www.w3.org/ns/credentials/issuer-dependent#TERM</code>.
         </p>
         <p>
-            This document provides a proper landing place for those URL-s, without any additional definitions for the terms themselves. 
+            This document provides a proper landing place for those URLs, without any additional definitions for the terms themselves. 
         </p>
     </section>
 
--- a/vocab/credentials/v2/issuer-dependent.html
+++ b/vocab/credentials/v2/issuer-dependent.html
@@ -1,0 +1,80 @@
+<html lang="en">
+  <head>
+    <meta charset='utf-8'/>
+    <title>Verifiable Credentials Issuer-Dependent Terms</title>
+    <script class="remove" src="https://www.w3.org/Tools/respec/respec-w3c"></script>
+    <script>
+      function remove_status_remark() {
+          const sotd = document.getElementById("sotd");
+          const p = sotd.getElementsByTagName('p')[0];
+          sotd.removeChild(p);
+      }
+    </script>
+    <script class="remove">
+      var respecConfig = {
+        localBiblio: {
+          "vc-data-model": {
+            title: "Verifiable Credentials Data Model v2.0",
+            href: "https://www.w3.org/TR/vc-data-model-2.0/",
+            authors: [
+              "Manu Sporny", "Grant Noble", "Dave Longley",
+              "Daniel C. Burnett", "Brent Zundel", "Kyle Den Hartog",
+              "Orie Steel", "Michael B. Jones", "Gabe Cohen"
+            ],
+            publisher: "W3C"
+          }
+        },
+        specStatus:  "base",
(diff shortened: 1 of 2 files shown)


iherman:
Note, b.t.w., that the VDCM specification has an outdated context file in the appendix, which does not seem to include the `@vocab` entry. This PR does not touch that file.

OR13 (approved):
Thank you!

vocab/credentials/v2/issuer-dependent.html
```diff
@@ -0,0 +1,80 @@
...
+            <a href="mailto:public-vc-comments@w3.org">public-vc-comments@w3.org</a>
+            (<a href="mailto:public-vc-comments-request@w3.org?subject=subscribe">subscribe</a>,
+            <a href="https://lists.w3.org/Archives/Public/public-vc-comments/">archives</a>).    
+        </p>
+    </section>
+    <section id="conformance"></section>
+    <section>
+        <h2>Issuer Dependent Terms</h2>
+        <p>
+            The <a href="https://www.w3.org/TR/vc-data-model-2.0/#base-context">base context</a> of a Verifiable Credential, as defined in [vc-data-model], provides a mapping from the standard terms used in credentials to the vocabulary items defined by the Verifiable Credentials Working Group through different vocabularies, for example the <a href="https://www.w3.org/2018/credentials/">Verifiable Credentials Vocabulary</a>. Applications using additional, user-dependent vocabularies of their own are encouraged to create their own formal vocabularies, and corresponding JSON-LD context definitions, to ensure the interoperability of the credentials they produce and/or consume.
```
TallTed, line 69:
```suggestion
            The <a href="https://www.w3.org/TR/vc-data-model-2.0/#base-context">base context</a> of a Verifiable Credential, as defined in [vc-data-model], provides a mapping from the standard terms used in credentials to the vocabulary items defined by the Verifiable Credentials Working Group through different vocabularies, such as the <a href="https://www.w3.org/2018/credentials/">Verifiable Credentials Vocabulary</a>. Applications using additional, user-dependent vocabularies of their own are encouraged to create their own formal vocabularies, and corresponding JSON-LD context definitions, to ensure the interoperability of the credentials they produce and/or consume.
```

vocab/credentials/v2/issuer-dependent.html
```diff
@@ -0,0 +1,80 @@
...
+        </p>
+        <p>
+            There may be cases, however, when the creation of such formal vocabularies and context definitions is too much of a burden; this is the case for closed applications, experiments, or examples. The way the <a href="https://www.w3.org/TR/vc-data-model-2.0/#base-context">base context</a> has been constructed is such that issuers may use new, hitherto undefined terms without further ado, and these terms will be mapped on URL-s of the form <code>https://www.w3.org/ns/credentials/issuer-dependent#TERM</code>.
```
TallTed, line 72:
```suggestion
            There may be cases, however, when the creation of such formal vocabularies and context definitions is too much of a burden, such as for closed applications, experiments, or examples. The <a href="https://www.w3.org/TR/vc-data-model-2.0/#base-context">base context</a> has been constructed in such a way that issuers may use new, hitherto undefined terms without further ado, and these terms will be mapped onto URLs of the form <code>https://www.w3.org/ns/credentials/issuer-dependent#TERM</code>.
```

vocab/credentials/v2/issuer-dependent.html
```diff
@@ -0,0 +1,80 @@
...
+        </p>
+        <p>
+            This document provides a proper landing place for those URL-s, without any additional definitions for the terms themselves.
```
TallTed, line 75:
```suggestion
            This document provides a proper landing place for those URLs, without any additional definitions for the terms themselves. 
```

TallTed (changes requested):
Minor tweaks.

msporny (approved):
LGTM, might need a tiny bit of editorial cleanup after the merge, but that would be ... editorial... no need to do that here, in the first PR.

melvincarvalho:
Normally in Linked Data a term in a vocab will dereference so that it's possible to learn more about the term, for example the type of the item (string, number, date etc.) the human readable label and so on

Looking at this vocab, that seems not to be the case here wrt dereferening

If the term will not dereference would it not be better to remove the centralized http dependency, and extra round trip by using a URN?

RFC 8141 provides for a documentation to be given for URNs without requiring the round trips that many libraries (e.g. solidOS) will fire off automatically.

It would be good to understand this pattern from a Linked Data point of view.  

If the aim is to provide a placeholder for json terms that are not registers, then there could be more benefit from a general solution that many systems could use, solving the problem of uniftying JSON and JSON-LD once and for all

iherman:
> Normally in Linked Data a term in a vocab will dereference so that it's possible to learn more about the term, for example the type of the item (string, number, date etc.) the human readable label and so on
> 
> Looking at this vocab, that seems not to be the case here wrt dereferening
> 
> If the term will not dereference would it not be better to remove the centralized http dependency, and extra round trip by using a URN?
> 
> RFC 8141 provides for a documentation to be given for URNs without requiring the round trips that many libraries (e.g. solidOS) will fire off automatically.
> 
> It would be good to understand this pattern from a Linked Data point of view.
> 
> If the aim is to provide a placeholder for json terms that are not registered, then there could be more benefit from a general solution that many systems could use, solving the problem of uniftying JSON and JSON-LD once and for all

I find the argument of using a URN for this case compelling. This should probably be raised as a separate issue.

@OR13 @msporny ?

OR13:
@iherman I think we can create a better experience for users by retaining URL structure. 

I don't think using URNs for this case would be an improvement.

iherman:
Just for the sake of arguments: we are talking about users who do not care about URL-s, Linked Data, etc; this is a fallback for that community. Which is fine, but then I do not understand your point. They would not care about the format of the identifier, whether it would start with a `https` or a `urn`...

OR13:
@iherman I think the intention is to provide term definitions that gently encourage issuer's to do a better job at linked data... These users are not intending to ignore linked data, let us not assume why they have failed to define terms in the context.

Making the term's URNs seems like the opposite of helping or educating.

Providing a web page that gives gentle guidance and instruction is better than spitting out unresolvable URNs, because the latter provides no guidance or instruction... no incentive for improvement and therefor it is likely to encourage even lower quality linked data.

melvincarvalho:
Thank @iherman I guess it does warrant its own issue.  I'll just reply with some added information.

RFC 8141 provides a means for documentation, namely

https://www.iana.org/assignments/urn-namespaces/urn-namespaces.xhtml

On the topic a general solution to this problem.  Might it perhaps be useful to register `urn:jsonld` with IANA?

msporny:
Editorial, multiple reviews, changes requested (some applied, one tracked in a new issue), no objections, merging.

iherman:
For the records: the .htaccess file has been installed; these two URL-s should work now as required:

- https://www.w3.org/ns/credentials/v2
- https://www.w3.org/ns/credentials/issuer-dependent

cc @msporny @OR13

OR13:
Awesome!

Issue #779, opened 2021-05-15
Is an Verifiable credential in JSON-LD expand form valid?


kdenhartog:
We've recently been toying around with the idea of how to enable verifiable credential validation in offline scenarios and have hit a bit of a snag around retrieval of arbitrary context files. One of the ways we've looked to get around this is to include the semantics by passing a VC in the expanded form and including a list of contexts so that the credentials can be mapped back to a valid compact form VC which matches many of the normative statements within the spec.

With this in mind, I think it may make sense for us to look at how we could identify versions of a context using a hash for example `https://www.w3.org/2018/credentials/ab4ddd9a531758807a79a5b450510d61ae8d147eab966cc9a200c07095b0cdcc` and then combining this with passing of the expanded form credential so that the semantics of the credential don't need to be resolved before verification.

Thoughts?

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2022-08-03-vcwg#section6-1) on 2022-08-03

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 6.1. Is a verifiable credential in JSON-LD expand form valid? (issue vc-data-model#779)

_See github issue [vc-data-model#779](https://github.com/w3c/vc-data-model/issues/779)._




**Brent Zundel:** First one we are looking at is #779. I anticipate some of these may be a little less clear, because of the way the original data model was written..  

**Manu Sporny:** The answer is no, an expanded document is not valid..  
… But this has more to do with vc-data-model than data-integrity; would suggest leave it.  

**Brent Zundel:** OK, anyone disagree?.  
… Removing the "data-integrity?" label..  

</details>

jandrieu:
@kdenhartog said
> We've recently been toying around with the idea of how to enable verifiable credential validation when it's not possible to retrieve arbitrary context files. 

My answer: Don't.

Not to be snarky, but if the software doesn't recognize a context it should NOT process the VC. There is no time in a production system where you should dereference an unknown context. Either the software was written to handle the context, or it wasn't. You can't expect that the software can magically interpret a previously unseen context.

It is a known security failing to dereference contexts in production. If you follow that guidance, in production, you should never worry about "unretreivable contexts" because you shouldn't be retrieving them anyway.

kdenhartog:
[...]

PR #1050, opened 2023-02-17
Restrict JSON-LD expression to compact document form.

msporny:
Addresses PR #779 by limiting JSON-LD expressions to compact document form.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1050.html" title="Last updated on Mar 15, 2023, 4:26 PM UTC (e63c069)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1050/8639d7c...e63c069.html" title="Last updated on Mar 15, 2023, 4:26 PM UTC (e63c069)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3363,6 +3363,13 @@ side. The associated vocabulary document for the Verifiable Credentials Data
 Model is available at <code>https://www.w3.org/2018/credentials</code>.
         </p>
 
+        <p>
+This specification restricts the usage of JSON-LD representations. JSON-LD <a
+href="https://www.w3.org/TR/json-ld/#compacted-document-form">compact document
+form</a> MUST be utilized for all expressions of the data model described by
+this specification.
+        </p>
+
         <section>
           <h3>Syntactic Sugar</h3>
 
--- a/index.html
+++ b/index.html
@@ -3364,7 +3364,8 @@ Model is available at <code>https://www.w3.org/2018/credentials</code>.
         </p>
 
         <p>
-This specification restricts the usage of JSON-LD representations of Verifiable Credentials and Verifiable Presentations. JSON-LD <a
+This specification restricts the usage of JSON-LD representations of 
+<a>credentials</a> and <a> presentations</a>. JSON-LD <a
 href="https://www.w3.org/TR/json-ld/#compacted-document-form">compact document
 form</a> MUST be utilized for all representations of the data model described by
 this specification.
--- a/index.html
+++ b/index.html
@@ -3364,9 +3364,9 @@ Model is available at <code>https://www.w3.org/2018/credentials</code>.
         </p>
 
         <p>
-This specification restricts the usage of JSON-LD representations. JSON-LD <a
+This specification restricts the usage of JSON-LD representations of Verifiable Credentials and Verifiable Presentations. JSON-LD <a
 href="https://www.w3.org/TR/json-ld/#compacted-document-form">compact document
-form</a> MUST be utilized for all expressions of the data model described by
+form</a> MUST be utilized for all representations of the data model described by
 this specification.
         </p>
 
--- a/index.html
+++ b/index.html
@@ -3364,11 +3364,11 @@ Model is available at <code>https://www.w3.org/2018/credentials</code>.
         </p>
 
         <p>
-This specification restricts the usage of JSON-LD representations of 
-<a>credentials</a> and <a> presentations</a>. JSON-LD <a
+This specification restricts the usage of JSON-LD representations of
+the data model. JSON-LD <a
 href="https://www.w3.org/TR/json-ld/#compacted-document-form">compact document
-form</a> MUST be utilized for all representations of the data model described by
-this specification.
+form</a> MUST be utilized for all representations of the data model in the
+base media type, `application/vc+ld+json`.
         </p>
 
         <section>


index.html
```diff
@@ -3363,6 +3363,13 @@ <h3>JSON-LD</h3>
 Model is available at <code>https://www.w3.org/2018/credentials</code>.
         </p>
 
+        <p>
+This specification restricts the usage of JSON-LD representations. JSON-LD <a
```
OR13, line 3367:
```suggestion
This specification restricts the usage of JSON-LD representations of Verifiable Credentials and Verifiable Presentations. JSON-LD <a
```

index.html
```diff
@@ -3363,6 +3363,13 @@ <h3>JSON-LD</h3>
...
+href="https://www.w3.org/TR/json-ld/#compacted-document-form">compact document
+form</a> MUST be utilized for all expressions of the data model described by
```
OR13, line 3369: what is the difference between "expression" and "representation" ?

OR13 (changes requested):
Prefer more explicit wording, not sure how the normative statement will be testable in its current form.

dlongley (approved):

TallTed (approved):
LGTM. Will be improved by @OR13's suggestion and question.

iherman (approved):

kdenhartog (approved):

index.html
msporny: Historically, I think we've tried to use "expression" to talk about something in the abstract (the data model) and "representation" to talk about something concrete (a serialization). I mixed them up here, and will change both to speak to "representation" -- does that work for you, @OR13 ?

msporny, line 3369:
```suggestion
form</a> MUST be utilized for all representations of the data model described by
```

index.html
```diff
@@ -3363,6 +3363,13 @@ <h3>JSON-LD</h3>
 Model is available at <code>https://www.w3.org/2018/credentials</code>.
         </p>
 
+        <p>
+This specification restricts the usage of JSON-LD representations of Verifiable Credentials and Verifiable Presentations. JSON-LD <a
```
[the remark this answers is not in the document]
msporny: Hmm... are we only limiting those? We're limiting credentials, right? Specifically, the base media type?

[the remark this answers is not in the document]
OR13: I suppose it would be clearer to say `presentation` and `credential`.

restricting the shape of `proof` does not belong in this spec, imo.

OR13 (changes requested):
I like this suggested change, https://github.com/w3c/vc-data-model/pull/1050#discussion_r1113610884

But I would also take a change dropping `verifiable`.

msporny:
> But I would also take a change dropping `verifiable`.

Ok, dropping `verifiable` for now since it seems to be more accurate wrt. the media type being discussed.

index.html
```diff
@@ -3363,6 +3363,14 @@ <h3>JSON-LD</h3>
 Model is available at <code>https://www.w3.org/2018/credentials</code>.
         </p>
 
+        <p>
+This specification restricts the usage of JSON-LD representations of 
+<a>credentials</a> and <a> presentations</a>. JSON-LD <a
+href="https://www.w3.org/TR/json-ld/#compacted-document-form">compact document
+form</a> MUST be utilized for all representations of the data model described by
```
OR13, line 3370: "all representations of the data model"... this seems not correct based on day 3 resolution.

we agreed that this only needs to be true for the "base media type" aka

`application/credential+ld+json` or `application/presentation+ld+json`

TallTed: > "all representations of the data model"... this seems not correct based on F2F day 3 resolution.

You're taking that phrase out of context.  The stricture is on "all representations of the data model **_described by this specification_**" which I believe we expect to be few, so your concern seems premature to me.

OR13: > Resolution #1: The base media type for the VCDM is credential+ld+json. `@context` is required (MUST) in the base media type; other media types MAY choose to include `@context`. Serializations in other media types (defined by the VCWG) MUST be able to be transformed into the base media type. Another media type MUST identify if this transformation is one-directional or bi-directional. Bi-directional transformation MUST preserve `@context`. Transformation rules MUST be defined, but not necessarily by this WG..

https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-02-16-vcwg#resolution1

https://github.com/w3c/vc-data-model/issues/947#issuecomment-1434506542

I feel `credential+ld+json` uses "compact form" would be a clearer way of saying the same thing.

Seems the intention is aligned with resolution, and data integrity proofs can indeed use "non compact form" for VCs.

TallTed: I think we would do well by rewording the sentence —
```
JSON-LD <a
href="https://www.w3.org/TR/json-ld/#compacted-document-form">compact document
form</a> MUST be utilized for all representations of the data model described by
```
— to —
```
JSON-LD <a
href="https://www.w3.org/TR/json-ld/#compacted-document-form">compact document
form</a> MUST be utilized for all JSON-LD representations of the data model described by
```
— if that is indeed what was intended (which I don't think it was).

Given that we're defining the base media type, named `credential+ld+json` (or `vc+ld+json`), to which all other representations must be losslessly transformable, I think it would be better to say —
```
JSON-LD <a
href="https://www.w3.org/TR/json-ld/#compacted-document-form">compact document
form</a> MUST be utilized for all representations of the data model in the base 
media type, vc+ld+json (or credential+ld+json).
```

OR13 (approved):
Approving only based on the resolution today, regarding `vc+ld+json`.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-03-14-vcwg-special#section2-3) on 2023-03-14

- no resolutions were taken

#### 2.3. Restrict JSON-LD expression to compact document form. (pr vc-data-model#1050)

_See github pull request [vc-data-model#1050](https://github.com/w3c/vc-data-model/pull/1050)._




**Manu Sporny:** This one, PR 1050, was just waiting on what we decided today. I'm not worried about this PR.  
… We can merge this based on our conversation today.  

</details>

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-03-14-vcwg-special#section1) on 2023-03-14

List of resolutions:

- ***[Resolution No. 1](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-03-14-vcwg-special#resolution1):  We will change `application/credential+ld+json` to `application/vc+ld+json` to use as the media type for verifiable credentials and add language to the spec that indicates the media type does not signal how the VC has been signed.***

### 1. Media types `vc+json+ld` vs. `credential+json+ld`.

**Manu Sporny:** Could we tackle the PRs about media type first? And I ask because there are about five PRs we could potentially resolve if we get to a conclusion there and we are maybe close.  

**Ted Thibodeau Jr.:** I'm hesitant to go along with changing special topic on the special topic call — might have shown up if another topic is discussed.  

**Brent Zundel:** With those comments, let's stick w/ PRs.  

**Manu Sporny:** I'm wondering — So I've got a proposal for the group and wanting to see if we can get to consensus... seeking feedback on it.  
… We have been talking about ... we have decided that there will be one base media type for the VC data model. That's super good and useful.  
… Then, we started talking about ... that was `application/credential+ld+json` as the base media type, unsecured, our base media type. Then we have a number of media types around `application/vc+something` for secured things but there has been disagreement around the requirement of `proof` and we were trying to figure out how to do that through media types. We now have a section in the spec for the media types.  
… Then Orie raised a really good question — about how these two media types `application/credential+ld+json` and `application/vc+ld+json` ... people are arguing that `proof` is optional in both and we have two media types that describe the same thing. Skipping past a lot of the back and forth.  
… I wonder if we can make the base media type `application/vc+ld+json` and `proof` is optional in that. Then in the media types section we make internal/external proofs clear and so on.  
… The proposal would be for the base media type to be `application/vc+ld+json`, `proof` is optional, but we don't say that in the media type registration but in the spec. What do people think?  

**Ted Thibodeau Jr.:** Yes, in general theory, I'm with you.  
… The media type for a zip archive doesn't change whether it's password protected, or one or a hundred files, or anything, really. You deal with it as a zip archive.  

> *Dave Longley:* +1 to TallTed.

**Ted Thibodeau Jr.:** In our case, we are dealing with credentials which are verifiable, and the failure could be that proof is missing, or proof fails (you messed up the crypto), or something along those lines.  
… It's taken a while to reach that point as a group, we might be a step away where we out to be. That might just be ld+json. I want there to be an explicit difference between `vc+ld+json` and `ld+json`.  
… There has to be something that a ld+json loses when it doesn't process as `vc+ld+json`.  

**Kristina Yasuda:** I'm not with you yet, Manu. The proposal I think I made in previous call was well received — can we completely remove proof property from VC Data Model so only media type in main becomes credential+ld+json, so we don't talk about proof property. Then vc+jwt and data integrity can define their own media types vc+??? if needed, is that an option, is that on the table still?  

**Michael Jones:** I'm going to try to react to all three comments.  
… I agree with Manu that de-facto, the WG hasn't been able to, for JSON-LD data model, been able to make any distinctions between credential and verifiable credential in the media type.  
… The attempts to say that proof must not be present for `credential+ld+json` have not succeeded, and in order to make progress, I'm willing to see single VC data model be `application/vc+ld+json`.  
… To Ted's comment, I agree that there is and should be a meaningful distinction between `ld+json` and `vc+ld+json`, the difference is the required members of VC data model are present and are used in ways specified.  

**Manu Sporny:** Yes, MikeJ, exactly!.  

**Michael Jones:** I think we can agree to that.  
… I agree with you that proof property should be moved into securing spec, that's what defines its properties, I don't think that Kristina's comment blocks us from registering the media type, and from brent's earlier comments, if we can unblock the media type registrations, we'll be a good bit further forward today.  

> *Orie Steele:* +1 to compact form being a constraint.

**Dave Longley:** I got on queue to agree with MikeJ — the constraints are the data model — and compact form must be used. any one of those things establishes a specific new constraint, that's the difference.  
… with respect to Kristina's comment — if we move proof back into another spec, doesn't resolve particular matter at hand, proof being optional, we'd end up in that situation, those issues are orthogonal. That's a different question from what we're trying to resolve here. We want the base media type to be `application/vc+ld+json` as everyone else has spoken to.  

> *Phillip Long:* +1 that moving the proof to data integrity is a different question and not the one that we're talking about here.

**Orie Steele:** I agree with the comments from Manu and Mike Jones about `application/vc+ld+json` media type. I also am not sure that proof predicate should be removed from core data model context. I want to talk about what's in v1.1 spec, if we can't achieve consensus to change it, it's goin gto stay the same.  
… In v1.1, proof is a predicate that exists. There is no definition of Credential (in RDF sense), only Verifiable Credential. The proof types are the RDF classes that the securing formats tended to define. A proof type example would be `Ed25519Signature2018`. In v1.1 we shipped proof property in core context. In v2.0, we cleaned some stuff up and avoided having to talk about types. If we say base media type is `application/vc+ld+json`, we commit ourselves to have conversations about proof being in media type.  
… It would be good to get agreement on the base media type on what we see in v1.1, there is a higher chance of doing that with `application/vc+ld+json`, vs coming to consensus on what `application/credential+ld+json`.  

**Kristina Yasuda:** If you are talking about media type for v1.1, `vc+ld+json` might be ok, but to use that for v2.0, I'd be against that. To v2.0 is to separate payload from how it's being secured. Saying the base media type is "vc+" — what is the verifiable part of that payload if that's the base media type?  
… Does that mean now we'd only have one media type per signed credential that has to express both what is a payload and what is signed? To be more specific, one example I can think of, in SD-JWT, with current separation of `application/credential+ld+json` — use that as a payload, use SD-JWT to secure it.  
… when you're signing, apply digest as defined in SD-JWT and you're good to go. I don't know what that means when we say `application/vc+ld+json`.  

**Ted Thibodeau Jr.:** First, assuming we continue down this road, how are we going to convince IANA that there will not be a cascade of special definitions of `ABC/xyz+ld+json` each with their own set of members and usages? I don't have a concrete differentiator, we could say "you have to have these members in order to be a VC".  

> *Orie Steele:* partial answer to Ted's question, IANA expects registration for types that use registered suffixes. That is why we define structured suffixes.

**Ted Thibodeau Jr.:** Consider RFC822 messages and attachments, where media types came from, there is an envelope and a payload and metadata, these things are separate, signature on signed mail, on unsigned main no signature, not mentioned. these are things that didn't require new media types, media types for those things existed in other usages, those media types are applied in enclosures. we're using media types in a way that's not intended to be used, it'll bite us if it's not already doing so.  

> *Michael Jones:* My response to Ted will reference [https://www.rfc-editor.org/rfc/rfc8725.html#use-typ](https://www.rfc-editor.org/rfc/rfc8725.html#use-typ).

**Dave Longley:** I wanted to respond to Kristina, differentiating between credentials and verifiable crednetials — to a certain extend, we've made a number of things more complicated than before in ways we didn't get more value out of that complication. The difference might not come through an incompatible syntax. data integrity adds proof, which adds types to a proof property. that fits in w/ existing data model. That other syntaxes might have incompatible syntax, but when that happens, you need a new media type. I think all that's ok.  
… I think Mike's on the queue to respond to what Ted said, I'll end my comments there.  

> *Dave Longley:* i was also going to say . i don't think SD-JWT will have to change.

**Michael Jones:** I did want to respond to TallTed's comment... how are we going to convince IANA to say there's not going to be a multiplication of mediatypes `*+ld+json`, by analogy, there was an RFC written 8725 — JWT best practices, and one of the best practices listed was "Use an explicit media type for every type of JWT object" — explicit typing and that can prevent a confusion for one type of JWT for another. That's part of a more general move in security community.  

> *Orie Steele:* +1 selfissued.

**Michael Jones:** "Use explicit typing" to prevent confusion, not a bug that there might be a multiplication of JSON-LD subtypes, it's a feature, because it enables explicit typing. Using `vc+ld+json` is an example for explicit typing, that's a good thing.  

**Brent Zundel:** Initial round was prompted by a proposal, folks spoke up for certain aspects, resolve the media type aspects first then continue conversation where it needs to go.  
… How would we need to change that proposal to move forward?  

**Dave Longley:** This might be a slight tweak, keep proof where it is right now and figure out if we move it later.  

> *Orie Steele:* +1 dlongley.

**Kristina Yasuda:** I still haven't heard an answer.  
… How do we know if it's signed or not?  

**Orie Steele:** In case signing is used — proof is used, RDF says the type... if external is used, then proof isn't there or is ignored and external signing mechanism applies information necessary for signature. So, for JWT, alg secures content type for application/vc+ld+json — and if you look at that content type, youg et back JSON, verify signature in external proof case. If you see proof in data model, and you don't understand it, you ignore it.  

> *Dave Longley:* +1 to Orie.

**Orie Steele:** People might profile away proof property, but there are scenarios where they want it to be present and might want to tunnel secured data in proof. And add external proof format. Short answer is in case of external proof, you're going to verify the external proof and then you get payload... if payload contains internal proof, you might ignore it, or if you do understand it, you verify according to proof type.  

> *Dave Longley:* +1 to Orie.

> *Phillip Long:* =1.

> *Orie Steele:* For the record, I don't really care to "block" the "credential tunneling" feature.

> *Phillip Long:* +1.

**Kristina Yasuda:** Thanks Orie, that makes sense, I think. I'll be ok given that is clearly stated in VCDM. I don't know what concrete proposal language should be, someone could point me to that? I think what I heard was base media type says this is a signed credential using JSON-LD and how to determine how it's signed is specific to signing mechanism, which is fine.  
… Make sure that's loud and clear, you have to depend on checking mechanism.  

> *Orie Steele:* And the current "proof" section essentially says this already... it says "at least one, proof format"... [https://w3c.github.io/vc-data-model/#proofs-signatures](https://w3c.github.io/vc-data-model/#proofs-signatures).

> *Dave Longley:* +1 to kristina to provide clear guidance.

**Kristina Yasuda:** For VC-JWT, for example, other specifications can define their content type, credential is signed ... if my understanding is correct, I'm ok w/ direction.  

> *Manu Sporny:* (Brent deftly composes a proposal, much like Beethoven writes a symphony.).

**Orie Steele:** media type `vc+ld+json` might have external proof, you won't see evidence in internal content type, because external proof is external.  
… You could have internal proof, and that would say you have an internal proof. I would avoid trying to smith proof related changes into resolution. We will have to address proof related issues as a group, don't see need to add more language to this particular proposal.  

**Brent Zundel:** Kristina, are you ok w/ moving forward noting that we need to have proof conversation separately?  

> *Manu Sporny:* Kristina makes noises of disapproval.

**Brent Zundel:** Or we could add...  

**Kristina Yasuda:** The media type does not dictate how credential is signed?  

> *Dave Longley:* +1 to Kristina.

> *Orie Steele:* +1 to the "media type does not dictate how the credential is signed".

> *Kerri Lemoie:* +1 Kristina.

> *Phillip Long:* +1.

> *Orie Steele:* that is what I was trying to say regarding the current proof section of the spec.

**Dave Longley:** we should say "to use as the base media type".  

> *Phillip Long:* +1 to Dave's suggestion.

> **Proposed resolution: We will change `credential+ld+json` to `vc+ld+json` to use as the media type for verifiable credentials and add language to the spec that indicates the media type does not signal how the VC has been signed.** *(Brent Zundel)*

> *Brent Zundel:* +1.

> *Michael Jones:* +1.

> *SongPu Ai:* +1.

> *Joe Andrieu:* +1.

> *Gabe Cohen:* +1.

> *Phillip Long:* +1.

> *Shigeya Suzuki:* +1.

> *Orie Steele:* +1.

> *Kerri Lemoie:* +1.

> *Dave Longley:* +1 `application/` prefix is presumed :).

**Manu Sporny:** +1.  

> *Kristina Yasuda:* 0.

**Dave Longley:** +1.  

> *Ted Thibodeau Jr.:* +1 with application/ :-).

**Brent Zundel:** This will go out for review before it's officially resolved.  
… I'm going to add application/ to be clear that that's what we resolved.  

> ***Resolution #1: We will change `application/credential+ld+json` to `application/vc+ld+json` to use as the media type for verifiable credentials and add language to the spec that indicates the media type does not signal how the VC has been signed.***

_See github pull request [vc-data-model#1034](https://github.com/w3c/vc-data-model/pull/1034)._




**Manu Sporny:** With that, I think we can merge Orie's 1034 PR.  
… We may / want to modify that PR to just remove the `application/credential+ld+json` to make it clear there's just one base media type.  
… We should do that and that will be 1034. Then 1050 can refer to `application/vc+ld+json` and we can pull in 1050.  
… 1014 I believe gets addressed by this.  

_See github pull request [vc-data-model#1014](https://github.com/w3c/vc-data-model/pull/1014)._




_See github pull request [vc-data-model#1050](https://github.com/w3c/vc-data-model/pull/1050)._




**Manu Sporny:** We're going very far towards resolving it at least.  

</details>

index.html
msporny: Fixed in e63c0692c44b02f3d679250bcb3c2b48b7dd9e3d.

msporny:

Issue #956, opened 2022-10-24
Protecting integrity of `@context`


mark-moir:
[Section 8.2](https://w3c.github.io/vc-data-model/#content-integrity-protection) (Content Integrity Protection) highlights the dangers of depending on remote content that is not content-integrity protected, and suggests that developers use schemes such as [Cryptographic Hyperlinks](https://datatracker.ietf.org/doc/draft-sporny-hashlink/) or [IPFS](https://en.wikipedia.org/wiki/InterPlanetary_File_System) to protect them.

[Section B1](https://w3c.github.io/vc-data-model/#base-context) (Base Context) includes the content of ([required](https://w3c.github.io/vc-data-model/#contexts)) https://www.w3.org/2018/credentials/v1 and provides its SHA-256 digest.  However, this content itself includes links that are **not** content-integrity protected (e.g., https://w3id.org/security#).

Doesn't [Section B1](https://w3c.github.io/vc-data-model/#base-context) therefore create a false sense of security?

msporny:
> Doesn't [Section B1](https://w3c.github.io/vc-data-model/#base-context) therefore create a false sense of security?

It's debatable if it does or not. :)

Ideally, yes, everything would be content protected all the way down to the human readable text... but we're just not there yet (from an infrastructure standpoint).

The security vocabulary has been brought under the purview of the VCWG: https://github.com/w3c/vc-data-integrity/tree/main/vocab/security

So, the sorts of attacks we're talking about are attacks against humans or semantic reasoners. At present, no one is using semantic inference on these documents, and if they do, I'd expect them to keep a static copy of all machine-readable vocabulary documents. That leaves the attack on the human-readable text, in this case, it's probably unlikely that someone could trick a human into mis-implementing something based on what we've written in the vocabulary (because there are test suites that text for implementation correctness/conformance).

The important thing is that everyone is using the same URLs for the terms, and in order to ensure that, all we need to do is embed the context in the spec and provide a SHA-256 digest for it.

Does that help?

dlongley:
> Doesn't [Section B1](https://w3c.github.io/vc-data-model/#base-context) therefore create a false sense of security?

No, I don't think so. But if you think so, you should draw out a concrete scenario where it's a significant issue so that it could be discussed and mitigations could be considered if it seems worthwhile to do so.

[...]

Issue #1044, opened 2023-02-15
Do we have to introduce the `Credential` class in the VCDM vocabulary?


iherman:
(This discussion may belong to the comment thread of #1014, but it is not commenting on the main line of the discussion. I have therefore decided to create a separate issue.)

As a result of the introduction of `credentials+ld+json` there is now a discrepancy between the media types used in the spec and the [vcdm vocabulary](https://w3c.github.io/vc-data-model/vocab/credentials/v2/vocabulary.html). I do not think that is healthy.

At present, there is no class for a `Credential` in the vocabulary (although it is there between the lines, i.e., in the term definitions). We only have a [`VerifiableCredential`](https://w3c.github.io/vc-data-model/vocab/credentials/v2/vocabulary.html#VerifiableCredential) term. Now that we have a a separate media type for credentials, i.e., a (non-verifiable) credential is now a separate "thing", we may want to consider introducing a separate `Credential` class to the vocabulary.

However, if we do this, we have several issues that we have to consider. These issues reflect the discussions we have in #1014. (Note that this is not only boring question of RDF; the vocabulary is a (simple) ontology, i.e., it forces us to pin down the exact relationships of the various terms we are using.)

The fundamental question is: is a `VerifiableCredential` a subclass of `Credential` or not? This is, essentially, the counterpart of the question whether (if we go down that route) we would have `verifiable+credential+ld+json` or `verifiable-credential+ld+json`, respectively. The alternatives on the vocabulary are:

1. **`VerifiableCredential` is a subclass of `Credential`**  
    From the vocabulary point of view, this is a relatively simple thing. Most properties ([`credentialStatus`](https://w3c.github.io/vc-data-model/vocab/credentials/v2/vocabulary.html#credentialStatus), [`validFrom`](https://w3c.github.io/vc-data-model/vocab/credentials/v2/vocabulary.html#validFrom), etc.) have `Credential` as a domain; by virtue of subclassing using those terms on `VerifiableCredential` instances is perfectly fine.
[...]

PR #1051, opened 2023-02-17
Add warning related to protecting integrity of @context

msporny:
This PR addresses #956 by adding text suggested by the issue creator, with modifications to align it with the rest of the specification.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1051.html" title="Last updated on Apr 3, 2023, 6:51 PM UTC (2dfb826)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1051/dfa4884...2dfb826.html" title="Last updated on Apr 3, 2023, 6:51 PM UTC (2dfb826)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4869,8 +4869,37 @@ The base context, located at
 <strong><code>944167aaabd904ea9e35c98fd7e8794eb6dd42ae4666b036b171e87fc34cc7cc</code></strong>,
 can be used to implement a local cached copy. It is possible to confirm the
 SHA-256 digest by running the following command from a modern Unix command
-interface line: `curl -s https://www.w3.org/ns/credentials/v2 | sha256sum`. For
-convenience, the base context is also provided below.
+interface line: `curl -s https://www.w3.org/ns/credentials/v2 | sha256sum`.
+          </p>
+          <p>
+This section serves as a reminder of the importance of ensuring that, when
+verifying <a>verifiable credentials</a> and <a>verifiable presentations</a>, the
+<a>verifier</a> has information that is consistent with what the <a>issuer</a>
+had when securing the <a>credential</a> or <a>presentation</a>. This information
+might include at least: 1) contents of the credential itself; 2) any content
+included by reference in the credential; 3) any content transitively included by
+reference.
+          </p>
+          <p>
+Mechanisms used to secure <a>credentials</a> and <a>presentations</a>, such
+as digital signatures, address securing the contents of the credential itself
+(item #1 in the previous paragraph). The SHA-256 digest of the Base Context in
+this section addresses securing the context information (item #2 in the
+previous paragraph).
+          </p>
+          <p>
+However, it does not cryptographically protect other content referenced such as
+other JSON-LD Contexts or files referenced via URLs. It is considered a best
+practice to ensure that the same sorts of protections are provided for other
+JSON-LD Contexts and URLs through the use of permanently cached files or
+cryptographic hashes for linked information that is critical to a specific use
+case. See the Content Integrity section of the Verifiable Credential
+Implementation Guide to verify that content it either fetches remotely or caches
+locally is consistent with what the <a>issuer</a> intended.
+          </p>
+          <p>
+For convenience, the base context for the Verifiable Credential Data Model v2.0
+is provided below.
           </p>
 
 <pre class="informative">
--- a/index.html
+++ b/index.html
@@ -4889,7 +4889,7 @@ mechanisms such as [[VC-JWT]] and [[VC-DATA-INTEGRITY].
             </li>
             <li>
 The content in a credential whose meaning depends on a link to an external URL,
-such as a JSON-LD Context, which can be secured by using a locally cached copy,
+such as a JSON-LD Context, which can be secured by using a local static copy,
 or a cryptographic digest of the file.
             </li>
           </ol>
--- a/index.html
+++ b/index.html
@@ -4866,12 +4866,13 @@ franchise. Policy information expressed by the <a>issuer</a> in the
           <p>
 The base context, located at <code>https://www.w3.org/ns/credentials/v2</code>
 with a SHA-256 digest of
-<strong><code>944167aaabd904ea9e35c98fd7e8794eb6dd42ae4666b036b171e87fc34cc7cc</code></strong>,
+<strong><code>ff1fdfa8d3f07547ec149b6652fe3f5a4276b9f0c48344dc29400763ba7b44fa</code></strong>,
 can be used to implement a local cached copy. It is possible to confirm the
 SHA-256 digest by running the following command from a modern Unix command
-interface line: `curl -s https://www.w3.org/ns/credentials/v2 | sha256sum`. It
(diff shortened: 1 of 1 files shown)


peacekeeper (approved):
I think there may be some use cases where a claim includes a URL, and the issuer really only wants to attest to the URL itself, not the content behind it. In fact, in such a use case both the issuer and verifier may be perfectly happy with the content behind the URL changing from time to time. They just want to make sure that the URL itself is correct.

Anyway, this PR doesn't preclude that and in general feels very valuable, especially with regard to `@context`.

index.html
```diff
@@ -4813,13 +4813,45 @@ <h2>Contexts, Types, and Credential Schemas</h2>
...
+          <p>
+However, other data that is referenced from within a credential, such as
+resources that are linked to via URLs, are not cryptographically protected by
+default. It is considered a best practice to ensure that the same sorts of
+protections are provided for any URL that is critical to the security of the
+credential through the use of permanently cached files or cryptographic hashes.
+See the <a href="https://w3c.github.io/vc-imp-guide/#content-integrity">Content
+Integrity</a> section of the Verifiable Credential Implementation Guide to
+verify that content it either fetches remotely or caches locally is consistent
+with what the <a>issuer</a> or <a>holder</a> intended.
```
dlongley, line 4850: Fix grammar and use "loaded from a local copy" to cover both the case of ephemeral caching and loading more permanent local copies.

```suggestion
Integrity</a> section of the Verifiable Credential Implementation Guide to
verify that content that is either fetched remotely or loaded from a local
copy is consistent with what the <a>issuer</a> or <a>holder</a> intended.
```

index.html
```diff
@@ -4813,13 +4813,45 @@ <h2>Contexts, Types, and Credential Schemas</h2>
       <section class="informative">
         <h3>Base Context</h3>
           <p>
-The base context, located at
-<code>https://www.w3.org/ns/credentials/v2</code> with a SHA-256 digest of
+The base context, located at <code>https://www.w3.org/ns/credentials/v2</code>
+with a SHA-256 digest of
 <strong><code>944167aaabd904ea9e35c98fd7e8794eb6dd42ae4666b036b171e87fc34cc7cc</code></strong>,
```
dlongley, line 4818: I think we should consider also adding a hash of the JCS-canonized version of the context to eliminate any differences from whitespace, etc.

index.html
```diff
@@ -4813,13 +4813,45 @@ <h2>Contexts, Types, and Credential Schemas</h2>
...
 can be used to implement a local cached copy. It is possible to confirm the
 SHA-256 digest by running the following command from a modern Unix command
-interface line: `curl -s https://www.w3.org/ns/credentials/v2 | sha256sum`. For
-convenience, the base context is also provided below.
+interface line: `curl -s https://www.w3.org/ns/credentials/v2 | sha256sum`. It
```
OR13, line 4821: `sha256sum` seems less desirable than using openssl...

index.html
```diff
@@ -4813,13 +4813,45 @@ <h2>Contexts, Types, and Credential Schemas</h2>
...
+JSON-LD Contexts, as well as locally cached copies, address item #2 in the
+previous paragraph.
+          </p>
+          <p>
+However, other data that is referenced from within a credential, such as
+resources that are linked to via URLs, are not cryptographically protected by
+default. It is considered a best practice to ensure that the same sorts of
+protections are provided for any URL that is critical to the security of the
+credential through the use of permanently cached files or cryptographic hashes.
+See the <a href="https://w3c.github.io/vc-imp-guide/#content-integrity">Content
```
OR13, line 4847: can we use `data-cite` instead?

index.html
dlongley: The non-canonized hash I get if I use the current `v2` context in the repo is this:

`e41d1743c941b8003a4689e6e844b7f6ae0b9970ed81aa918bd781977bdba8b2`

The canonized hash I get is this:

`d64fce2a2f642ddf9f74a6aa01eb16efdaa822d0380d4891d310f3a1e085e671`

Neither of these matches the value here, which may just be out of date.

dlongley (approved):
Thanks! Some suggestions.

index.html
TallTed: @OR13 — Suggest the alternative command?

index.html
```diff
@@ -4813,13 +4813,45 @@ <h2>Contexts, Types, and Credential Schemas</h2>
...
+or <a>holder</a> had when securing the <a>credential</a> or <a>presentation</a>.
+This information might include at least: 1) contents of the credential itself;
+2) any content linked to via a URL in the credential; 3) any content linked to
+via a URL, which then links to more content via a URL (and so on).
+          </p>
+          <p>
+Mechanisms used to secure <a>credentials</a> and <a>presentations</a>, such
+[[VC-JWT]] and [[VC-DATA-INTEGRITY]], address securing the contents of the
+credential itself (item #1 in the previous paragraph). The SHA-256 digest of
+JSON-LD Contexts, as well as locally cached copies, address item #2 in the
```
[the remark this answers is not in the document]
TallTed: > It would be nice to not have to jump back / refer back to read the intention. 

Yes. I would suggest bumping the 1-3 into a list, with each information item immediately followed by the mechanism used to address it. Item #3 would need some suggestion of how to address it.

iherman (approved):
I approve the PR, but I would flag some thoughts for later. The recent WG discussions make a clear difference between _credentials_ and _verifiable credentials_. I think once the dust settles, we will have to go through the spec with a fine comb to see when we use one term and when the other. I presume this new section is a typical case where the warning should also apply to a _credential_ and not to a _verifiable credential_ only (e.g., in case the verification is done by encapsulating the credential in a JWT).

(see also #1044)

mark-moir:
I commented as requested [here](https://github.com/w3c/vc-data-model/issues/956#issuecomment-1437699105)

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-03-14-vcwg-special#section2-4) on 2023-03-14

- no resolutions were taken

#### 2.4. Add warning related to protecting integrity of @context (pr vc-data-model#1051)

_See github pull request [vc-data-model#1051](https://github.com/w3c/vc-data-model/pull/1051)._




**Manu Sporny:** This one ... people gave comments on the PR and in issues and in my email inbox... I have to reconcile all that first.  
… For PR 1051.  
… I'll deal with all the various inputs.  

</details>

dmitrizagidulin (approved):
👍

index.html
msporny: Fixed in aa4d74c ... which has a different hash than the two items above.

index.html
msporny: Fixed in 1e5723e9636c8f41615bae939109cf4677d78829.

index.html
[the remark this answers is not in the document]
msporny: Fixed in cb1d4558d1b6e2108a238c74530e1147851aba3f.

PR #1053, opened 2023-02-23
VC.name term definition

nissimsan:
"name" pointing at https://schema.org/description doesn't seem right.

### DIFF
--- a/contexts/credentials/v2
+++ b/contexts/credentials/v2
@@ -47,7 +47,7 @@
           "@type": "@id"
         },
         "name": {
-          "@id": "https://schema.org/description",
+          "@id": "https://schema.org/name",
           "@type": "http://www.w3.org/2001/XMLSchema#string"
         },
         "proof": {


nissimsan:
I connected github and w3c, so I think the ipr fail should be resolved now.

jandrieu:
Seems to me "name" should not be in the context file at all. 

It's not part of the VCDM.

OR13 (approved):

OR13:
@jandrieu 

- https://github.com/w3c/vc-data-model/pull/752
- https://github.com/w3c/vc-data-model/pull/905

dlongley (approved):

mprorock (approved):

msporny:
> Seems to me "name" should not be in the context file at all.

`name` became a part of the VCDM in PR #905 (with broad support). Implementers were struggling with being able to simply name and describe a credential using text. It was decided that schema.org's `name` and `description` were useful terms to do this and broadly applicable.

msporny (approved):
@nissimsan wrote:
> "name" pointing at https://schema.org/description doesn't seem right.

Yes, good catch, I hit exactly this problem today while running some code and it puzzled me for a good 15 minutes before I saw the issue.

Thanks for the fix, @nissimsan !

PR #1055, opened 2023-03-01
Add section describing use of media types 

mprorock:
This is intended to be a work in progress PR to help achieve consensus on the right location for describing requirements and best practices around use of media types with VCs.  I believe that this PR should form a basis to close without merging #1014 as the normative language in that PR is probably in the wrong section and does not have consensus.  This PR also may provide guidance around how to proceed with #1034

***
<a href="https://pr-preview.s3.amazonaws.com/mesur-io/vc-data-model/pull/1055.html" title="Last updated on Mar 7, 2023, 2:18 PM UTC (fee2e8c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1055/ab0345e...mesur-io:fee2e8c.html" title="Last updated on Mar 7, 2023, 2:18 PM UTC (fee2e8c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3426,14 +3426,10 @@ model whether or not they use a [[!JSON-LD]] processor.
           Any media types associated with the core data model are listed in the section on
           <a href="#iana-considerations">IANA Considerations</a> for registration with IANA.
         </p>
-        <p>
-          When defining a media type for use with verifiable credentials — for instance, in a 
-          specification that defines a specific syntax — use of the term `verifiable` implies
-          the presence of a <a href="#proofs-signatures">proof</a> with the credential.  
-          This also applies when the term is abbreviated, as in other specifications, 
-          such as in the [[VC-JWT]] specification's use of `vc+ld+json`.
-          Use of media types identified as `verifiable` in this way MUST correspond to
-          the use of one or more proofs with the credential.
+        <p class="issue" data-number="1060">
+          Presence of the term "verifiable" (or a shortened version) in the media type may 
+          or may not imply a proof.  Original proposed language is tracked in the issue 
+          corresponding to this note.
         </p>
         <p>
           Use of the term `credential` in a media type related to a syntax of verifiable 
--- a/index.html
+++ b/index.html
@@ -3432,7 +3432,7 @@ model whether or not they use a [[!JSON-LD]] processor.
           the presence of a <a href="#proofs-signatures">proof</a> with the credential.  
           This also applies when the term is abbreviated, as in other specifications, 
           such as in the [[VC-JWT]] specification's use of `vc+ld+json`.
-          Use of media type identified as `verifiable` in this way MUST correspond with
+          Use of media types identified as `verifiable` in this way MUST correspond to
           the use of one or more proofs with the credential.
         </p>
         <p>
--- a/index.html
+++ b/index.html
@@ -3414,7 +3414,7 @@ model whether or not they use a [[!JSON-LD]] processor.
 
         <p>
           Media types as defined in [[RFC6838]] serve a useful purpose with verifiable 
-          credentials, especially in regards to identifcation of a particular syntax 
+          credentials, especially regarding identification of a particular syntax 
           that is in use with a verifiable credential.
         </p>
         <p>
@@ -3427,16 +3427,17 @@ model whether or not they use a [[!JSON-LD]] processor.
           <a href="#iana-considerations">IANA Considerations</a> for registration with IANA.
         </p>
         <p>
-          When defining a media type for use with verifiable credentials, for instance in a 
-          specification that defines a specific syntax, use of the term `verifiable` implies
+          When defining a media type for use with verifiable credentials — for instance, in a 
+          specification that defines a specific syntax — use of the term `verifiable` implies
           the presence of a <a href="#proofs-signatures">proof</a> with the credential.  
-          This is also the case if the term is abbreviated in accordance with other specs, such as in `vc+ld+json`. 
+          This also applies when the term is abbreviated, as in other specifications, 
+          such as in the [[VC-JWT]] specification's use of `vc+ld+json`.
           Use of media type identified as `verifiable` in this way MUST correspond with
           the use of one or more proofs with the credential.
         </p>
         <p>
           Use of the term `credential` in a media type related to a syntax of verifiable 
-          credentials without the corresponding use of `verifiable` as noted above does not 
+          credentials <i>without</i> the corresponding use of `verifiable` as described above <i>does not</i> 
           imply the presence of a proof with the <a href="#dfn-credential">credential</a>.
         </p>
         <p>
@@ -3444,7 +3445,7 @@ model whether or not they use a [[!JSON-LD]] processor.
           model: `application/credential+ld+json` and `application/verifiable+credential+ld+json`.
           Other specifications such as [[VC-JWT]] define additional media types as may be 
           required to allow for proper interpretation of the syntax of verifiable credentials 
-          encoded in the respective syntax described by the specification.
+          encoded in the syntax described by the respective specification.
         </p>
       </section>
 
--- a/index.html
+++ b/index.html
@@ -3431,8 +3431,8 @@ model whether or not they use a [[!JSON-LD]] processor.
           specification that defines a specific syntax, use of the term `verifiable` implies
           the presence of a <a href="#proofs-signatures">proof</a> with the credential.  
           This is also the case if the term is abbreviated in accordance with other specs, such as in `vc+ld+json`. 
-          Use of media type identified as `verifiable` in this way MUST correspond with use 
-          of one or more proofs with the credential.
+          Use of media type identified as `verifiable` in this way MUST correspond with
+          the use of one or more proofs with the credential.
         </p>
         <p>
           Use of the term `credential` in a media type related to a syntax of verifiable 
--- a/index.html
+++ b/index.html
@@ -3443,7 +3443,7 @@ model whether or not they use a [[!JSON-LD]] processor.
           At the time of this writing, there are two media types associated with the core data 
           model: `application/credential+ld+json` and `application/verifiable+credential+ld+json`.
           Other specifications such as [[VC-JWT]] define additional media types as may be 
-          required to allow for proper interpertation of the syntax of verifiable credentials 
+          required to allow for proper interpretation of the syntax of verifiable credentials 
           encoded in the respective syntax described by the specification.
         </p>
       </section>
--- a/index.html
+++ b/index.html
@@ -3430,7 +3430,7 @@ model whether or not they use a [[!JSON-LD]] processor.
           When defining a media type for use with verifiable credentials, for instance in a 
           specification that defines a specific syntax, use of the term `verifiable` implies
           the presence of a <a href="#proofs-signatures">proof</a> with the credential.  
-          This is also the case if the term is abbreviated, such as in `vc+ld+json`. 
+          This is also the case if the term is abbreviated in accordance with other specs, such as in `vc+ld+json`. 
           Use of media type identified as `verifiable` in this way MUST correspond with use 
           of one or more proofs with the credential.
         </p>
--- a/index.html
+++ b/index.html
@@ -3419,7 +3419,7 @@ model whether or not they use a [[!JSON-LD]] processor.
         </p>
         <p>
           Syntaxes SHOULD be identified by a media type, and certain conventions as 
-          outlined in this section should be followed when defining or using media types 
+          outlined in this section SHOULD be followed when defining or using media types 
           with verifiable credentials.
         </p>
         <p>
--- a/index.html
+++ b/index.html
@@ -3409,6 +3409,44 @@ model whether or not they use a [[!JSON-LD]] processor.
         </section>
 
       </section>
+      <section>
+        <h3>Media Types</h3>
+
+        <p>
+          Media types as defined in [[RFC6838]] serve a useful purpose with verifiable 
+          credentials, especially in regards to identifcation of a particular syntax 
+          that is in use with a verifiable credential.
+        </p>
+        <p>
+          Syntaxes SHOULD be identified by a media type, and certain conventions as 
+          outlined in this section should be followed when defining or using media types 
+          with verifiable credentials.
+        </p>
+        <p>
+          Any media types associated with the core data model are listed in the section on
+          <a href="#iana-considerations">IANA Considerations</a> for registration with IANA.
+        </p>
+        <p>
+          When defining a media type for use with verifiable credentials, for instance in a 
+          specification that defines a specific syntax, use of the term `verifiable` implies
+          the presence of a <a href="#proofs-signatures">proof</a> with the credential.  
+          This is also the case if the term is abbreviated, such as in `vc+ld+json`. 
+          Use of media type identified as `verifiable` in this way MUST correspond with use 
+          of one or more proofs with the credential.
+        </p>
+        <p>
+          Use of the term `credential` in a media type related to a syntax of verifiable 
+          credentials without the corresponding use of `verifiable` as noted above does not 
+          imply the presence of a proof with the <a href="#dfn-credential">credential</a>.
+        </p>
+        <p>
+          At the time of this writing, there are two media types associated with the core data 
+          model: `application/credential+ld+json` and `application/verifiable+credential+ld+json`.
+          Other specifications such as [[VC-JWT]] define additional media types as may be 
+          required to allow for proper interpertation of the syntax of verifiable credentials 
+          encoded in the respective syntax described by the specification.
+        </p>
+      </section>
 
       <section>
         <h3>Proof Formats</h3>


index.html
```diff
@@ -3409,6 +3409,44 @@ <h3>Syntactic Sugar</h3>
...
+        <p>
+          Use of the term `credential` in a media type related to a syntax of verifiable 
+          credentials without the corresponding use of `verifiable` as noted above does not 
+          imply the presence of a proof with the <a href="#dfn-credential">credential</a>.
+        </p>
+        <p>
+          At the time of this writing, there are two media types associated with the core data 
+          model: `application/credential+ld+json` and `application/verifiable+credential+ld+json`.
+          Other specifications such as [[VC-JWT]] define additional media types as may be 
+          required to allow for proper interpertation of the syntax of verifiable credentials
```
dlongley, line 3446:
```suggestion
          required to allow for proper interpretation of the syntax of verifiable credentials 
```

index.html
```diff
@@ -3409,6 +3409,44 @@ <h3>Syntactic Sugar</h3>
...
+          of one or more proofs with the credential.
+        </p>
+        <p>
+          Use of the term `credential` in a media type related to a syntax of verifiable 
+          credentials without the corresponding use of `verifiable` as noted above does not 
+          imply the presence of a proof with the <a href="#dfn-credential">credential</a>.
+        </p>
+        <p>
+          At the time of this writing, there are two media types associated with the core data 
+          model: `application/credential+ld+json` and `application/verifiable+credential+ld+json`.
```
dlongley, line 3444: Notably, we haven't actually defined the latter one yet, but this is just an informative thing that can be changed as needed.

index.html
```diff
@@ -3409,6 +3409,44 @@ <h3>Syntactic Sugar</h3>
...
+        </p>
+        <p>
+          Any media types associated with the core data model are listed in the section on
+          <a href="#iana-considerations">IANA Considerations</a> for registration with IANA.
+        </p>
+        <p>
+          When defining a media type for use with verifiable credentials, for instance in a 
+          specification that defines a specific syntax, use of the term `verifiable` implies
+          the presence of a <a href="#proofs-signatures">proof</a> with the credential.  
+          This is also the case if the term is abbreviated, such as in `vc+ld+json`.
```
dlongley, line 3433: We might want a ref to the JWT spec here to be clear on what this abbreviation language is about.

```suggestion
          This is also the case if the term is abbreviated in accordance with other specs, such as in `vc+ld+json`. 
```

index.html
```diff
@@ -3409,6 +3409,44 @@ <h3>Syntactic Sugar</h3>
...
+          Any media types associated with the core data model are listed in the section on
+          <a href="#iana-considerations">IANA Considerations</a> for registration with IANA.
+        </p>
+        <p>
+          When defining a media type for use with verifiable credentials, for instance in a 
+          specification that defines a specific syntax, use of the term `verifiable` implies
+          the presence of a <a href="#proofs-signatures">proof</a> with the credential.  
+          This is also the case if the term is abbreviated, such as in `vc+ld+json`. 
+          Use of media type identified as `verifiable` in this way MUST correspond with use 
+          of one or more proofs with the credential.
```
dlongley, line 3435:
```suggestion
          Use of media type identified as `verifiable` in this way MUST correspond with
          the use of one or more proofs with the credential.
```

dlongley (approved):
+1!

Thanks for this PR, @mprorock, I think it's an appropriate way to provide guidance around media types and, from my perspective, the guidance it provides is good, clear, and does not violate any of my expectations.

index.html
mprorock: exactly - especially given the discussion on #1034 re `verifiable-credential` vs `verifiable+credential` this gives us a good way to get aligned

index.html
```diff
@@ -3409,6 +3409,44 @@ <h3>Syntactic Sugar</h3>
         </section>
 
       </section>
+      <section>
+        <h3>Media Types</h3>
+
+        <p>
+          Media types as defined in [[RFC6838]] serve a useful purpose with verifiable 
+          credentials, especially in regards to identifcation of a particular syntax
```
TallTed, line 3417:
```suggestion
          credentials, especially regarding identification of a particular syntax 
```

index.html
```diff
@@ -3409,6 +3409,44 @@ <h3>Syntactic Sugar</h3>
...
+        <h3>Media Types</h3>
+
+        <p>
+          Media types as defined in [[RFC6838]] serve a useful purpose with verifiable 
+          credentials, especially in regards to identifcation of a particular syntax 
+          that is in use with a verifiable credential.
+        </p>
+        <p>
+          Syntaxes SHOULD be identified by a media type, and certain conventions as 
+          outlined in this section should be followed when defining or using media types
```
TallTed, line 3422: Ought this `should` be `SHOULD`, or a different word to avoid confusion?
```suggestion
          outlined in this section should be followed when defining or using media types 
```

index.html
```diff
@@ -3409,6 +3409,44 @@ <h3>Syntactic Sugar</h3>
...
+          outlined in this section should be followed when defining or using media types 
+          with verifiable credentials.
+        </p>
+        <p>
+          Any media types associated with the core data model are listed in the section on
+          <a href="#iana-considerations">IANA Considerations</a> for registration with IANA.
+        </p>
+        <p>
+          When defining a media type for use with verifiable credentials, for instance in a 
+          specification that defines a specific syntax, use of the term `verifiable` implies
```
TallTed, line 3431:
```suggestion
          When defining a media type for use with verifiable credentials — for instance, in a 
          specification that defines a specific syntax — use of the term `verifiable` implies
```

index.html
```diff
@@ -3409,6 +3409,44 @@ <h3>Syntactic Sugar</h3>
...
+          the presence of a <a href="#proofs-signatures">proof</a> with the credential.  
+          This is also the case if the term is abbreviated in accordance with other specs, such as in `vc+ld+json`.
```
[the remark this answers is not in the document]
TallTed:
```suggestion
          This also applies when the term is abbreviated, as in other specifications, 
          such as in the [[VC-JWT]] specification's use of `vc+ld+json`.
```

index.html
```diff
@@ -3409,6 +3409,44 @@ <h3>Syntactic Sugar</h3>
...
+          Any media types associated with the core data model are listed in the section on
+          <a href="#iana-considerations">IANA Considerations</a> for registration with IANA.
+        </p>
+        <p>
+          When defining a media type for use with verifiable credentials, for instance in a 
+          specification that defines a specific syntax, use of the term `verifiable` implies
+          the presence of a <a href="#proofs-signatures">proof</a> with the credential.  
+          This is also the case if the term is abbreviated in accordance with other specs, such as in `vc+ld+json`. 
+          Use of media type identified as `verifiable` in this way MUST correspond with
+          the use of one or more proofs with the credential.
```
TallTed, line 3435:
```suggestion
          Use of media types identified as `verifiable` in this way MUST correspond to
          the use of one or more proofs with the credential.
```

index.html
```diff
@@ -3409,6 +3409,44 @@ <h3>Syntactic Sugar</h3>
...
+        </p>
+        <p>
+          Use of the term `credential` in a media type related to a syntax of verifiable 
+          credentials without the corresponding use of `verifiable` as noted above does not 
+          imply the presence of a proof with the <a href="#dfn-credential">credential</a>.
```
TallTed, line 3440:
```suggestion
          Use of the term `credential` in a media type related to a syntax of verifiable 
          credentials <i>without</i> the corresponding use of `verifiable` as described above <i>does not</i> 
          imply the presence of a proof with the <a href="#dfn-credential">credential</a>.
```

index.html
```diff
@@ -3409,6 +3409,44 @@ <h3>Syntactic Sugar</h3>
...
+        </p>
+        <p>
+          At the time of this writing, there are two media types associated with the core data 
+          model: `application/credential+ld+json` and `application/verifiable+credential+ld+json`.
+          Other specifications such as [[VC-JWT]] define additional media types as may be 
+          required to allow for proper interpretation of the syntax of verifiable credentials 
+          encoded in the respective syntax described by the specification.
```
TallTed, line 3447:
```suggestion
          encoded in the syntax described by the respective specification.
```

TallTed (changes requested):
This is definitely several steps in the right direction. Just a few, relatively minor, tweaks suggested.

index.html
mprorock: i think `SHOULD` but that was part of the reason i wanted this as a draft first - happy to adjust it so.

OR13: I feel this paragraph should speak of "serializations / bytes"... also less might be more on this front.

index.html
```diff
@@ -3409,6 +3409,44 @@ <h3>Syntactic Sugar</h3>
...
+          with verifiable credentials.
+        </p>
+        <p>
+          Any media types associated with the core data model are listed in the section on
+          <a href="#iana-considerations">IANA Considerations</a> for registration with IANA.
+        </p>
+        <p>
+          When defining a media type for use with verifiable credentials, for instance in a 
+          specification that defines a specific syntax, use of the term `verifiable` implies
+          the presence of a <a href="#proofs-signatures">proof</a> with the credential.
```
OR13, line 3432: Yikes to this... I might defer this language to a separate PR to avoid blocking this valuable other changes.

`proof` is NOT required to make a `VerifiableCredential`...  asserting that `proof` makes a `VerifiableCredential` implies data integrity proofs as a defacto standard representation / serialization for the core data model.

index.html
OR13, line 3435: I get the intention here, but I don't think `verifiable` is correct... I think `verifiable-credential` is correct. 

It would be nice to straw poll on `verifiable` vs `verifiable-credential` registration options. Just to see the preferences.

I think there are several folks in favor of registering `verifiable+credential+ld+json`.

index.html
mprorock: yeah - this is one reason why i wanted this as a draft first - we do i think want to note that some cryptographic proof is there, not necessarily that it is in the credential as an embedded proof (as it often is not) but that it should be present according to the securing method in use

index.html
mprorock: > I think there are several folks in favor of registering `verifiable+credential+ld+json`

yes - if we register `credential` then the `+` syntax would make likely the most sense

index.html
```diff
@@ -3409,6 +3409,45 @@ <h3>Syntactic Sugar</h3>
...
+        <h3>Media Types</h3>
+
+        <p>
+          Media types as defined in [[RFC6838]] serve a useful purpose with verifiable 
+          credentials, especially regarding identification of a particular syntax 
+          that is in use with a verifiable credential.
+        </p>
+        <p>
+          Syntaxes SHOULD be identified by a media type, and certain conventions as 
+          outlined in this section should be followed when defining or using media types
```
mprorock, line 3422:
```suggestion
          outlined in this section SHOULD be followed when defining or using media types 
```

index.html
mprorock: @OR13 agree - i think `serializations` is better but given that the `syntaxes` section is where this is at the moment I followed the current language of the spec.  would you be open to an issue on the topic of terminology related to syntax vs serialization, etc and how to handle that done separately?

mprorock: @TallTed does [this suggestion](https://github.com/w3c/vc-data-model/pull/1055#discussion_r1122204405) work? if so i will commit it.

index.html
```diff
@@ -3409,6 +3409,45 @@ <h3>Syntactic Sugar</h3>
...
+          with verifiable credentials.
+        </p>
+        <p>
+          Any media types associated with the core data model are listed in the section on
+          <a href="#iana-considerations">IANA Considerations</a> for registration with IANA.
+        </p>
+        <p>
+          When defining a media type for use with verifiable credentials — for instance, in a 
+          specification that defines a specific syntax — use of the term `verifiable` implies
+          the presence of a <a href="#proofs-signatures">proof</a> with the credential.
```
mprorock, line 3432:
```suggestion
          the presence of a <a href="#proofs-signatures">proof</a> with the credential either embedded or external to the credential.  
```

@OR13 note that i linked directly to the section that calls out both embedded AND external proofs intentionally here.  perhaps some additional language as in this suggestion would help

index.html
OR13: The registrations have to be complete and independent.... 

If we register `verifiable+credential+ld+json` we would have to register `credential+ld+json` if we wanted to be able to distinguish between bytes.

If WG members persist in asserting that `proof` is allowed in `credential+ld+json` I suggest we only register `verifiable+credential+ld+json`, and that we NOT attempt to define `credential+ld+json`... since `proof` will always be allowed  (but not required) in `verifiable+credential+ld+json`....

index.html
OR13: > At least one proof mechanism, and the details necessary to evaluate that proof, MUST be expressed for a credential or presentation to be a verifiable credential or verifiable presentation; that is, to be verifiable.

I feel like this spec text is being disputed by @iherman @TallTed @dlongley @msporny and possibly others.

> Media types MAY be registered with more than one suffix appended to
> the base subtype name.  The suffixes MUST be interpreted as ordered.
> Valid media type names containing a structured suffix are built from
> right to left (not left to right).  Characters on the left-most side
> of the left-most "+" in a subtype name specify the base subtype name.
> Characters to the right of each "+" in a subtype name denote
> additional structured syntax suffixes.

> Media types with more than one suffix MUST be registered according to
> the procedure defined in [RFC6838].  A new base subtype name MUST
> only be registered with suffix combinations that are already
> registered in their own right in the Structured Syntax Suffixes
> registry (https://www.iana.org/assignments/media-type-structured-suffix/media-type-structured-suffix.xhtml).

- https://datatracker.ietf.org/doc/draft-ietf-mediaman-suffixes/

> While it is possible for a given media type to be assigned additional
> names, the use of different names to identify the same media type is
> discouraged.

- https://www.rfc-editor.org/rfc/rfc6838#section-4.2

(🔥 let us not register 2 media types that can be used for the "same bytes").

> Media types that make use of a named structured syntax SHOULD use the
> appropriate registered "+suffix" for that structured syntax when they
> are registered.  By the same token, media types MUST NOT be given
> names incorporating suffixes for structured syntaxes they do not
> actually employ. "+suffix" constructs for as-yet unregistered
> structured syntaxes SHOULD NOT be used, given the possibility of
> conflicts with future suffix definitions.

- https://www.rfc-editor.org/rfc/rfc6838#section-4.2.8

(🔥 let us make sure things are actually verifiable or credentials).

Based on this comment: https://github.com/w3c/vc-data-model/issues/1044#issuecomment-1446805038

> `VerifiableCredential` might contain `proof` (modified since `proof` is actually optional in `VerifiableCredential`.)
>  `VerifiableCredential` is a subtype of `Credential`

processing from right to left...

is the `credential` ... `json` ?
is the `credential ... `ld+json`?
is the `credential ... `verifiable+ld+json`? (does this violate the MUST NOT or SHOULD NOT from the RFC?)

Compared to...

is the `verifiable` ... `json` ?
is the `verifiable` ... `ld+json` ?
is the `verifiable` ... `credential+ld+json` ? (does this violate the MUST NOT or SHOULD NOT from the RFC?)

index.html
TallTed: > If we register `verifiable+credential+ld+json` we would have to register `credential+ld+json` if we wanted to be able to distinguish between bytes.

If we're to register `application/verifiable+credential+ld+json`, we would have to register `application/credential+ld+json` *because it's required by the RFC which enables multiple `+` in registered Media Types*, whether or not we want to be able to distinguish between bytes.

If we're to register `application/verifiable-credential+ld+json`, we would *not* have to register `application/credential+ld+json` *because this is permitted by the RFC which enables multiple `+` in registered Media Types.*

> If WG members persist in asserting that `proof` is allowed in `credential+ld+json` I suggest we only register `verifiable+credential+ld+json`, and that we NOT attempt to define `credential+ld+json`... since `proof` will always be allowed (but not required) in `verifiable+credential+ld+json`....

Regardless of whether `proof` is allowed in `application/credential+ld+json`, if we register `application/verifiable+credential+ld+json` we *must* register (which means we *must* define) `application/credential+ld+json` *because it's required by the RFC which enables multiple `+` in registered Media Types.*

If we "persist" in permitting `proof` in `credential+ld+json`, we'll have no need for `application/verifiable+credential+ld+json` (nor `application/verifiable-credential+ld+json`) because the questions of whether there's a `proof` and whether it "proves" the credential comes down to business logic (which may include the application of various cryptomath).

index.html
TallTed: >>At least one proof mechanism, and the details necessary to evaluate that proof, MUST be expressed for a credential or presentation to be a verifiable credential or verifiable presentation; that is, to be verifiable.
>
> [@OR13] [I feel like this spec text is being disputed by @iherman @TallTed @dlongley @msporny and possibly others.](https://github.com/w3c/vc-data-model/pull/1055#discussion_r1122357911)

What a strange feeling... as I am not disputing that sentence at all.

I cannot make heads or tails of the remainder of your comment. Maybe try writing full sentences, including what brought you to quote the chunks of draft and ratified RFCs?

A reminder: "Classes" and "Types" do not _require_ Media Types to be defined and useful. A "tiger" is a subtype of "cat", which is a subtype of "mammal", etc. None of these have Media Types, but they are all useful descriptors. A "CompactDisc" is disjoint from "CassetteTape" and both are disjoint from "VinylLP", but all are subtypes of "Audio Recording Media", so all might be considered Media Types ... but this typing is not handled by RFC nor IANA.

OR13: @TallTed The line of argument you have here: https://github.com/w3c/vc-data-model/pull/1055#discussion_r1123655680

Regarding `proof`, is what leads me to believe that you do not see a difference in information between `credential` and `verifiable credential`.

dlongley: @OR13,

> let us not register 2 media types that can be used for the "same bytes"

This is literally what the `+` syntax does in media types. You can describe the string `{}` using either `application/json` OR `application/ld+json` today. There are numerous other examples with `+xml` suffixes and others. This subclassing is done on purpose to enable many different use cases.

@TallTed has done an excellent job in various threads explaining these rules -- but I fear there continue to be misunderstandings in various places.

OR13: I believe that @dlongley @TallTed @iherman and @msporny are asserting that assuming we register `credential+ld+json`, `verifiable+credential+ld+json` or `verifiable-credential+ld+json`, they would all / both be valid content types for the following:

```json
{
  "@context": [
    "https://w3id.org/security/suites/bls12381-2020/v1"
  ],
  "issuer": "https://example.edu/issuers/14",
  "credentialSubject": {
    "id": "did:example:ebfeb1f712ebc6f1c276e12ec21"
  },
  "proof": {
    "type": "BbsBlsSignature2020",
    "created": "2010-01-01T19:23:24Z",
    "verificationMethod": "did:example:123#key-0",
    "proofValue": "uBncn..."
  }
}
```

My point is that we don't need to register 3 media types for the same bytes, we only need to register `verifiable+credential+ld+json` or `verifiable-credential+ld+json` .

Then the whole "is proof allowed in `credential+ld+json` question" goes away, since `credential+ld+json` is no longer a valid media type.

The alternative proposal is `proof` is not allowed in `credential+ld+json`  and is optional in `verifiable+credential+ld+json` and we must register both, this was my original proposal, let us call it the "2 media type proposal"

@msporny indicated he was in favor of just registering `verifiable+credential+ld+json` and not registering `credential+ld+json`, let us call this the "one media type proposal".

Here are some proposals to run in the next working group call:

PROPOSAL: The 1 media type for vc proposal, the working group will register `verifiable+credential+ld+json` and the working group will not register `credential+ld+json`.

PROPOSAL: The 2 media type for vc proposal, the working group will register `verifiable+credential+ld+json` and the working group will register `credential+ld+json`.

TallTed: > I fear there continue to be some misunderstandings.

I share your fear. In spades.

andresuribe87: Given the comment from @TallTed https://github.com/w3c/vc-data-model/pull/1055#discussion_r1123655680 , then I would suggest we amend the proposals to be the following: 

PROPOSAL: The 1 media type for vc proposal, the working group will register `verifiable-credential+ld+json` and the working group will not register `credential+ld+json`.

PROPOSAL: The 2 media type for vc proposal, the working group will register `verifiable+credential+ld+json` and the working group will register `credential+ld+json`.

index.html
Sakurann: > If we "persist" in permitting proof in credential+ld+json, we'll have no need for application/verifiable+credential+ld+json (nor application/verifiable-credential+ld+json) because the questions of whether there's a proof and whether it "proves" the credential comes down to business logic (which may include the application of various cryptomath).

this is exactly why putting proof in the credential and leaving a decision whether to validate a signature or not to a technical layer but to a business layer does not make sense to me. 
entity receiving a thing needs to know whether it is required to verify a signature or not. "I will not validate a signature because i don't need to do so for my business, regardless of whether the thing is signed or not" sounds pretty dangerous.

dlongley: @Sakurann,

Whether to check a signature or not is always a business/application rules decision -- and one that you need to configure ahead of time in your software, before you receive anything from a party that might be an attacker. An attacker will just tell you not to check the signature. Media types don't solve this problem.

Also, there are also plenty of applications that don't need to check signatures -- like simple JSON viewers. It would be a pretty significant annoyance to a developer if they weren't able to even see some JSON to do some debugging because a signature checker *had* to be run ahead of time and it was failing.

msporny: @Sakurann wrote:
> "I will not validate a signature because i don't need to do so for my business, regardless of whether the thing is signed or not" sounds pretty dangerous.

Allowing the attacker to pick when/how the receiver checks the signature is even more dangerous. 

The approach being described here, to trigger certain types of checking based on media type, is analogous to the `alg=none` and HMAC flaws in the JOSE stack (that enable the attacker to decide how and what the server checks):

* https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/
* https://medium.com/@phosmet/forging-jwt-exploiting-the-none-algorithm-a37d670af54f
* https://cyberpolygon.com/materials/security-of-json-web-tokens-jwt/
* https://www.youtube.com/watch?v=zWVRHK3ykfo

This isn't about technical layer vs. business layer... this is all "security layer" stuff, and the more you enable an attacker to pick the parameters for checking (such as via media type), the greater the attack surface.

index.html
msporny: I would be a +1 for the first proposal directly above this comment (initially proposed by @OR13 and then refined outlined by @andresuribe87), with the caveat that the name of the media type is `vc+ld+json` (where the bikeshedding thereof doesn't need to hold up the decision).

OR13: @msporny I'm not arguing you can't have different media types for the same bytes.... I am arguing against registered multiple media types for the same bytes with indistinguishable definitions in our spec.

+1 to `vc+ld+json`

TallTed: > I am arguing against registered multiple media types for the same bytes with indistinguishable definitions in our spec.

The first question should be whether there's value in having the second media type, whether or not it is a supertype of the type we're primarily focused on.  If there is such value (as I believe there is), then if it is a supertype (as I believe it is), I think we should register both, _or only the supertype_.  (If it were a subtype, we might not register it, but there would be good reason to speak of it and to suggest a media type for it.)

Remember that media types are not actively restrictive nor actively prescriptive in most applications, but are primarily *hints* about how to process the bytes within the resource labeled with that media type.  In the world of the web, they're helpful for various tools to make known to a web server what form of resource those tools prefer such that the server can provide that, if available.  (See [RFC2295](https://www.rfc-editor.org/rfc/rfc2295), among others.)  It's perfectly valid for my super app to process resources of *every* media type in exactly the same way, if users of my super app find value in what my app outputs based on those inputs.

OR13: @TallTed does this 3rd proposal capture your alternate single registration proposal? :

PROPOSAL:The 1 media type for `credential` proposal, the working group will register `credential+ld+json` and the working group will not register `verifiable+credential+ld+json`.

index.html
OR13: Can we remove this paragraph on "proof" ? and move it to an issue to discuss?

OR13 (changes requested):
Let's strip down what we can.

Specifically this part: https://github.com/w3c/vc-data-model/pull/1055/files#r1127041668

^ lets move that to an issue to discuss.

index.html
mprorock: Marking resolved based on commit and thumbs up.  thanks @TallTed

index.html
mprorock: #1060 opened, paragraph removed to refine language

index.html
OR13: @msporny has requested to use `vc+ld+json` elsewhere, which I agree with, but we can leave it like this for now I think.

OR13 (approved):

index.html
```diff
@@ -3409,6 +3409,41 @@ <h3>Syntactic Sugar</h3>
...
+          corresponding to this note.
+        </p>
+        <p>
+          Use of the term `credential` in a media type related to a syntax of verifiable 
+          credentials <i>without</i> the corresponding use of `verifiable` as described above <i>does not</i> 
+          imply the presence of a proof with the <a href="#dfn-credential">credential</a>.
+        </p>
+        <p>
+          At the time of this writing, there are two media types associated with the core data 
+          model: `application/credential+ld+json` and `application/verifiable+credential+ld+json`.
```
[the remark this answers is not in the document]
OR13: I support this change.

TallTed (approved):

msporny:
Normative, multiple reviews, changes requested and made, Joe's suggestion needs to be merged to pull this in, some editorial changes will be made after the merge, further normative changes might be made based on discussion on the call today, no outstanding objections, merging (in a few hours).

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-03-07-vcwg-special#section1-2) on 2023-03-07

- no resolutions were taken

#### 1.2. Add section describing use of media types  (pr vc-data-model#1055)

_See github pull request [vc-data-model#1055](https://github.com/w3c/vc-data-model/pull/1055)._




**Brent Zundel:** In PR #1055, it adds a section describing the use of media types. There's some conversation, a little back and forth, but in general, this adds some normative guidance around media types in general..  
… I encourage people to get on the queue to say what needs to be changed in this PR to move it forward..  

**Manu Sporny:** Generally supportive of this PR. One of the good things it does, before we were making normative statements in the IANA registration section, that was dicey and this is better. This just talks in general of what's expected of certain media types. I think we can more easily put normative statements in here and there are some arguments around what's testable or not..  
… I'm +1 on merging the PR because they aren't just doing MUST do this or that. In general +1 for what this PR is doing which is moving normative language from the media type registrations to the core of the document which allows us to speak more freely about expectations in using media types..  

**Orie Steele:** I'm supportive of merging the PR. I'm several others that I'm the author of that I would close and I would move content from them into a new section or open new PRs to do that. Merging this will unblock several PRs that are currently blocked in my opinion..  

**Brent Zundel:** I will note that this PR has been open for a week. I will note that there is one request for changes from TallTed, I believe those changes have been made..  
… Ted, could you speak to that?.  

**Ted Thibodeau Jr.:** I don't think there's anything remaining, this PR makes substantial improvements from what was there before and even if I need to make a new PR to adjust, I'm good..  

**Joe:** How is this related to the single base media type? The language I'm reading doesn't clarify that one of those two is the base media type and one of those should be transformable into the base..  
… I'm concerned we're losing some of what we talked about in Miami..  

**Orie Steele:** I don't think this PR addresses what happened at the F2F in Miami, but it opens the door for it. It does mention the base media type by name. I think there are so many other PRs open about this topic of media types that I think it would be better to merge this PR and open a new one that's dedicated to summarizing the outcome from the F2F..  
… I am supportive of getting text of what we discussed in Miami, but subsequent to merging this..  

> *Orie Steele:* +1 Brent.

**Brent Zundel:** This PR creates a media type section, it moves some text from IANA registrations into this section and seeks to begin the normative guidance for use of media types. I think Mike Prorock did a pretty good job of avoiding normative statements -- setting the stage for those without making them so much yet..  
… From my perspective..  

**Ted Thibodeau Jr.:** Yeah, I just want to voice that I've typed a few times now. There was and there often is a substantial push to "make decisions" at a F2F, because theoretically having people in the room means better decisions. I disagree with that that is always the case and I think blind adherence to those sessions can be to the detriment of the spec and all of us..  
… I do think we resolved to a base media type and it was X, but we shouldn't necessarily be bound to that forever, we may hit things that indicate that was the wrong decision and we need to be able to revise it..  
… I think all decisions in all calls -- are not final, I forget the right wording there, until a week later. The idea is the same with F2F..  

**Joe Andrieu:** Politely, I want to push back on that. I think we need to make decisions and move forward. I think as a group we have the base media type `application/credential+ld+json`. I don't think this other media type is something we agreed to -- because of this date we haven't agreed to it..  

> *Joe Andrieu:* +1 to adding media types as we get consensus on what they would be and mean.

**Phillip Long:** The way I heard TallTed talking about the issue was relevant to a comment he made in the PR a few days ago. He reminded us that media types aren't actively prescriptive or restrictive. I think adding a media type doesn't preclude us from adding or removing another one later. I think this is better as it moves the text from the IANA registration to the spec. And I think, as Orie said, this frees up more PR work to move forward..  

> *Orie Steele:* I like Joe's suggestion, if there are no objections to it, I would like to see it merged..

**Phillip Long:** As long as we're clear that we can add/remove/change media types, this is a good step forward..  

> *Orie Steele:* +1 Manu.

**Manu Sporny:** I think Joe's correct, I think it's totally appropriate to just list the one we decided on in the WG call and we're just having the second media type discussion shortly..  

**Brent Zundel:** Ok, we wouldn't pull it in now, I think it's appropriate for MikeP to see Joe's comments on the PR..  
… Anyone else objecting to merging this PR with Joe's change?.  

> *Phillip Long:* No objections to Joe's change from here.

**Brent Zundel:** I'm not hearing any objections to merging the PR with Joe's change..  

</details>

PR #1056, opened 2023-03-01
Editorial break out of definitions for clarity

mprorock:
the definition of verifiable credential and verifiable presentation were both embedded in other definitions.  this PR simply moves those definitions to top level terminology

### DIFF
--- a/terms.html
+++ b/terms.html
@@ -9,11 +9,7 @@ An assertion made about a <a>subject</a>.
   </dd>
   <dt><dfn data-lt="credential|credentials">credential</dfn></dt>
   <dd>
-A set of one or more <a>claims</a> made by an <a>issuer</a>. A
-<dfn data-lt="verifiable credentials">verifiable credential</dfn> is a
-tamper-evident credential that has authorship that can be cryptographically
-verified. Verifiable credentials can be used to build
-<a>verifiable presentations</a>, which can also be cryptographically verified.
+A set of one or more <a>claims</a> made by an <a>issuer</a>. 
 The <a>claims</a> in a credential can be about different <a>subjects</a>.
   </dd>
   <dt><dfn>data minimization</dfn></dt>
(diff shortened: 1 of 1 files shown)


TallTed (commented):
Looks good at this reading. It might be worth adding something along the lines of "The class (a/k/a type) of _verifiable credential_ is a subclass (a/k/a subtype) of _credential_" and "The class (a/k/a type) of _verifiable presentation_ is a subclass (a/k/a subtype) of _presentation_" to make explicit these often misunderstood implied subclass a/k/a subtype relation.

mprorock:
> Looks good at this reading. It might be worth adding something along the lines of "The class (a/k/a type) of _verifiable credential_ is a subclass (a/k/a subtype) of _credential_" and "The class (a/k/a type) of _verifiable presentation_ is a subclass (a/k/a subtype) of _presentation_" to make explicit these often misunderstood implied subclass a/k/a subtype relation.

agreed - i think calling that out would be good, but might be better in a separate PR since that would add content as opposed to just being a restructure of existing text

OR13 (approved):

decentralgabe (approved):

dlongley (approved):
I'm fine with this change provided that we're continuing down the path of explicitly separating how we talk about credential and verifiable credential. There have been proposals elsewhere to reverse that trend as possible simplifications. However, describing the concepts like this can probably work either way -- we'll just need to settle on what the tangible technical pieces are that most developers will actually come into contact with when implementing.

awoie (approved):
lgtm

brentzundel:
I am fully on board with the separation we've determined between a credential and a verifiable credential. I am not as sold on a separation between a presentation and a verifiable presentation.
In my mind there is only a verifiable presentation, which is verifiable because it is constructed from verifiable credentials.
Looking for clarity.

mprorock:
> I am fully on board with the separation we've determined between a credential and a verifiable credential. I am not as sold on a separation between a presentation and a verifiable presentation. In my mind there is only a verifiable presentation, which is verifiable because it is constructed from verifiable credentials. Looking for clarity.

key difference is that the presentation of the array of credentials can also be signed - this is not always required for all use cases

msporny (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-03-07-vcwg-special#section1-1) on 2023-03-07

- no resolutions were taken

#### 1.1. Editorial break out of definitions for clarity (pr vc-data-model#1056)

_See github pull request [vc-data-model#1056](https://github.com/w3c/vc-data-model/pull/1056)._




**Manu Sporny:** So this is an editorial fix to the spec. A number of people have been asking us to separate the definitions of Verifiable Credential and Credential into different entries. It has multiple reviews, it didn't change any content, it's been out there for a week, I think we can merge this with no objections..  

**Brent Zundel:** I take back not merging anything today, this one is ready. Nothing but approvals, been out for a week, editorial..  
… Are there any objections to merging this PR? This PR is just taking existing definitions and breaking them into two, no new language, just shifting..  

**Ivan Herman:** A minor thing, Manu can you wait about an hour to merge -- because my scripts go mad when I try to get content from a PR that has been closed. That's all..  

**Brent Zundel:** Not hearing any objections..  

**David Chadwick:** One of the things that came up with myself and TallTed, but a Verifiable Credential is not necessarily Verifiable because the verification could fail for all kinds of reasons, only one good path and hundreds to fail..  

**Brent Zundel:** You're talking about a different PR..  

**David Chadwick:** Yes, but we should change the "can" to MAY on verifiable..  

**Brent Zundel:** This isn't normative, this is an editorial PR..  

> *Manu Sporny:* +1 brent.

**Brent Zundel:** We aren't adding normative language to this PR..  

**Ted Thibodeau Jr.:** I hear the concern, there might be some further text massaging..  

> *Orie Steele:* +1 Brent.

**Ted Thibodeau Jr.:** That verification may fail..  

**Brent Zundel:** That's a great suggestion for another PR. I don't want to open that can of worms on this PR..  

**David Chadwick:** Ok, we can do subsequent PRs for that change..  

**Brent Zundel:** Yes, very much encouraged..  
… I don't hear merging on this PR, when Ivan's script is ready we can merge it..  

</details>

PR #1064, opened 2023-03-13
update json schema to accurately reflect context defn

### LINKED ISSUE #1059: Invalid definition of context in schema
Context is defined in the spec as an array of string or objects (https://www.w3.org/TR/vc-data-model/#contexts):

> The value of the `@context` [property](https://www.w3.org/TR/vc-data-model/#dfn-property) MUST be an ordered set where the first item is a [URI](https://www.w3.org/TR/vc-data-model/#dfn-uri) with the value https://www.w3.org/2018/credentials/v1. For reference, a copy of the base context is provided in Appendix [B.1 Base Context](https://www.w3.org/TR/vc-data-model/#base-context). Subsequent items in the array MUST express context information and be composed of any combination of [URIs](https://www.w3.org/TR/vc-data-model/#dfn-uri) or objects. It is RECOMMENDED that each [URI](https://www.w3.org/TR/vc-data-model/#dfn-uri) in the `@context` be one which, if dereferenced, results in a document containing machine-readable information about the `@context`.

However, the schema restricts `@context` to be only an array of strings (https://github.com/w3c/vc-data-model/blob/main/schema/verifiable-credential/verifiable-credential-schema.json#L139-L148):

```json
"@context": {
      "type": "array",
      "items": {
        "type": "string"
      },
      "contains": {
        "const": "https://www.w3.org/2018/credentials/v1"
      },
      "minItems": 1
    },
```

Schema should be fixed to allow also objects in the array.

decentralgabe:
fixes #1059

### DIFF
--- a/schema/verifiable-credential/example-1.json
+++ b/schema/verifiable-credential/example-1.json
@@ -1,7 +1,7 @@
 {
   "@context": [
-    "https://www.w3.org/2018/credentials/v1",
-    "https://www.w3.org/2018/credentials/examples/v1"
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
   ],
   "id": "http://example.edu/credentials/1872",
   "type": ["VerifiableCredential", "AlumniCredential"],
--- a/schema/verifiable-credential/example-11.json
+++ b/schema/verifiable-credential/example-11.json
@@ -1,7 +1,7 @@
 {
   "@context": [
-    "https://www.w3.org/2018/credentials/v1",
-    "https://www.w3.org/2018/credentials/examples/v1"
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
   ],
   "id": "http://example.gov/credentials/3732",
   "type": ["VerifiableCredential", "UniversityDegreeCredential"],
--- a/schema/verifiable-credential/example-12.json
+++ b/schema/verifiable-credential/example-12.json
@@ -1,7 +1,7 @@
 {
   "@context": [
-    "https://www.w3.org/2018/credentials/v1",
-    "https://www.w3.org/2018/credentials/examples/v1"
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
   ],
   "id": "http://example.edu/credentials/3732",
   "type": ["VerifiableCredential", "UniversityDegreeCredential"],
--- a/schema/verifiable-credential/example-13.json
+++ b/schema/verifiable-credential/example-13.json
@@ -1,7 +1,7 @@
 {
   "@context": [
-    "https://www.w3.org/2018/credentials/v1",
-    "https://www.w3.org/2018/credentials/examples/v1"
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
   ],
   "id": "http://example.edu/credentials/3732",
   "type": ["VerifiableCredential", "UniversityDegreeCredential"],
--- a/schema/verifiable-credential/example-20.json
+++ b/schema/verifiable-credential/example-20.json
@@ -1,7 +1,7 @@
 {
   "@context": [
-    "https://www.w3.org/2018/credentials/v1",
-    "https://www.w3.org/2018/credentials/examples/v1"
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
   ],
   "id": "http://example.edu/credentials/3732",
   "type": ["VerifiableCredential", "UniversityDegreeCredential"],
--- a/schema/verifiable-credential/example-21.json
+++ b/schema/verifiable-credential/example-21.json
@@ -1,7 +1,7 @@
 {
   "@context": [
-    "https://www.w3.org/2018/credentials/v1",
-    "https://www.w3.org/2018/credentials/examples/v1"
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
   ],
   "id": "http://example.edu/credentials/3732",
   "type": ["VerifiableCredential", "UniversityDegreeCredential"],
--- a/schema/verifiable-credential/example-4.json
+++ b/schema/verifiable-credential/example-4.json
@@ -1,7 +1,7 @@
 {
   "@context": [
-    "https://www.w3.org/2018/credentials/v1",
-    "https://www.w3.org/2018/credentials/examples/v1"
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
   ],
   "id": "http://example.edu/credentials/3732",
   "type": ["VerifiableCredential", "UniversityDegreeCredential"],
--- a/schema/verifiable-presentation/example-2.json
+++ b/schema/verifiable-presentation/example-2.json
@@ -1,14 +1,13 @@
 {
   "@context": [
-    "https://www.w3.org/2018/credentials/v1",
-    "https://www.w3.org/2018/credentials/examples/v1"
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
   ],
   "type": "VerifiablePresentation",
-  
   "verifiableCredential": [{
     "@context": [
-      "https://www.w3.org/2018/credentials/v1",
-      "https://www.w3.org/2018/credentials/examples/v1"
+      "https://www.w3.org/ns/credentials/v2",
+      "https://www.w3.org/ns/credentials/examples/v2"
     ],
     "id": "http://example.edu/credentials/1872",
     "type": ["VerifiableCredential", "AlumniCredential"],
--- a/schema/verifiable-presentation/example-25.json
+++ b/schema/verifiable-presentation/example-25.json
@@ -1,14 +1,14 @@
 {
   "@context": [
-    "https://www.w3.org/2018/credentials/v1",
-    "https://www.w3.org/2018/credentials/examples/v1"
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
   ],
   "type": "VerifiablePresentation",
   "verifiableCredential": [
     {
       "@context": [
-        "https://www.w3.org/2018/credentials/v1",
-        "https://www.w3.org/2018/credentials/examples/v1"
+        "https://www.w3.org/ns/credentials/v2",
+        "https://www.w3.org/ns/credentials/examples/v2"
       ],
       "type": ["VerifiableCredential", "UniversityDegreeCredential"],
       "credentialSchema": {
--- a/schema/verifiable-presentation/example-45.json
+++ b/schema/verifiable-presentation/example-45.json
@@ -1,15 +1,15 @@
 {
   "@context": [
-    "https://www.w3.org/2018/credentials/v1",
-    "https://www.w3.org/2018/credentials/examples/v1"
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
   ],
   "id": "https://example.com/VP/0987654321",
   "type": ["VerifiablePresentation"],
   "verifiableCredential": [
     {
-     "@context": [
-       "https://www.w3.org/2018/credentials/v1",
-       "https://www.w3.org/2018/credentials/examples/v1"
+      "@context": [
+        "https://www.w3.org/ns/credentials/v2",
+        "https://www.w3.org/ns/credentials/examples/v2"
       ],
       "id": "http://pharma.example.com/credentials/3732",
       "type": ["VerifiableCredential", "PrescriptionCredential"],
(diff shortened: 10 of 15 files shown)


schema/verifiable-credential/example-1.json
```diff
@@ -1,7 +1,7 @@
 {
   "@context": [
-    "https://www.w3.org/2018/credentials/v1",
-    "https://www.w3.org/2018/credentials/examples/v1"
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
```
OR13, line 4:
```suggestion
```

schema/verifiable-credential/example-11.json
```diff
@@ -1,7 +1,7 @@
 {
   "@context": [
-    "https://www.w3.org/2018/credentials/v1",
-    "https://www.w3.org/2018/credentials/examples/v1"
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
```
OR13, line 4:
```suggestion
```

schema/verifiable-credential/example-12.json
```diff
@@ -1,7 +1,7 @@
 {
   "@context": [
-    "https://www.w3.org/2018/credentials/v1",
-    "https://www.w3.org/2018/credentials/examples/v1"
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
```
OR13, line 4:
```suggestion
```

schema/verifiable-credential/example-13.json
```diff
@@ -1,7 +1,7 @@
 {
   "@context": [
-    "https://www.w3.org/2018/credentials/v1",
-    "https://www.w3.org/2018/credentials/examples/v1"
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
```
OR13, line 4:
```suggestion
```

schema/verifiable-credential/example-20.json
```diff
@@ -1,7 +1,7 @@
 {
   "@context": [
-    "https://www.w3.org/2018/credentials/v1",
-    "https://www.w3.org/2018/credentials/examples/v1"
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
```
OR13, line 4:
```suggestion
```

schema/verifiable-credential/example-21.json
```diff
@@ -1,7 +1,7 @@
 {
   "@context": [
-    "https://www.w3.org/2018/credentials/v1",
-    "https://www.w3.org/2018/credentials/examples/v1"
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
```
OR13, line 4:
```suggestion
```

schema/verifiable-credential/example-4.json
```diff
@@ -1,7 +1,7 @@
 {
   "@context": [
-    "https://www.w3.org/2018/credentials/v1",
-    "https://www.w3.org/2018/credentials/examples/v1"
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
```
OR13, line 4:
```suggestion
```

schema/verifiable-presentation/example-2.json
```diff
@@ -1,14 +1,13 @@
 {
   "@context": [
-    "https://www.w3.org/2018/credentials/v1",
-    "https://www.w3.org/2018/credentials/examples/v1"
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
```
OR13, line 4:
```suggestion
```

schema/verifiable-presentation/example-2.json
```diff
@@ -1,14 +1,13 @@
...
   ],
   "type": "VerifiablePresentation",
-  
   "verifiableCredential": [{
     "@context": [
-      "https://www.w3.org/2018/credentials/v1",
-      "https://www.w3.org/2018/credentials/examples/v1"
+      "https://www.w3.org/ns/credentials/v2",
+      "https://www.w3.org/ns/credentials/examples/v2"
```
OR13, line 10:
```suggestion
```

schema/verifiable-presentation/example-25.json
```diff
@@ -1,14 +1,14 @@
 {
   "@context": [
-    "https://www.w3.org/2018/credentials/v1",
-    "https://www.w3.org/2018/credentials/examples/v1"
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
```
OR13, line 4:
```suggestion
```

schema/verifiable-presentation/example-25.json
```diff
@@ -1,14 +1,14 @@
...
   ],
   "type": "VerifiablePresentation",
   "verifiableCredential": [
     {
       "@context": [
-        "https://www.w3.org/2018/credentials/v1",
-        "https://www.w3.org/2018/credentials/examples/v1"
+        "https://www.w3.org/ns/credentials/v2",
+        "https://www.w3.org/ns/credentials/examples/v2"
```
OR13, line 11:
```suggestion
```

schema/verifiable-presentation/example-45.json
```diff
@@ -1,15 +1,15 @@
 {
   "@context": [
-    "https://www.w3.org/2018/credentials/v1",
-    "https://www.w3.org/2018/credentials/examples/v1"
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
```
OR13, line 4:
```suggestion
```

schema/verifiable-presentation/example-45.json
```diff
@@ -1,15 +1,15 @@
...
   "id": "https://example.com/VP/0987654321",
   "type": ["VerifiablePresentation"],
   "verifiableCredential": [
     {
-     "@context": [
-       "https://www.w3.org/2018/credentials/v1",
-       "https://www.w3.org/2018/credentials/examples/v1"
+      "@context": [
+        "https://www.w3.org/ns/credentials/v2",
+        "https://www.w3.org/ns/credentials/examples/v2"
```
OR13, line 12:
```suggestion
```

schema/verifiable-presentation/example-45.json
```diff
@@ -38,8 +38,8 @@
     },
     {
       "@context": [
-        "https://www.w3.org/2018/credentials/v1",
-        "https://www.w3.org/2018/credentials/examples/v1"
+        "https://www.w3.org/ns/credentials/v2",
+        "https://www.w3.org/ns/credentials/examples/v2"
```
OR13, line 42:
```suggestion
```

decentralgabe:
@OR13 I don't believe your comment is suitable for this PR, as I have followed the convention already in use by the current draft. See https://github.com/w3c/vc-data-model/blob/main/index.html#L980 for an example.

If you wish to modify the examples context that should come separately.

OR13:
@msporny @decentralgabe this PR adds examples that are malformed, because:

- https://www.w3.org/ns/credentials/examples/v2 is 404.

There is no consensus to make "https://www.w3.org/ns/credentials/examples/v2" not 404, and as you can probably tell, I am strongly opposed to repeating the mistakes made in v1.1 again in v2, because it has been a source of implementation burden, it injects bias, and it leads to needlessly complicated examples and tests.

Given the current PR would introduce broken examples, please accept my change suggestion and file an issue to discuss "making https://www.w3.org/ns/credentials/examples/v2 not 404".

I remain opposed to merging this PR without my suggestions being accepted.

OR13:
Related PR: https://github.com/w3c/vc-data-model/pull/1066

iherman:
Didn't we introduce the `@vocab` feature into the `@context` so that, for example, `https://www.w3.org/ns/credentials/examples/v2 ` would not be necessary any more? Maintaining properly yet another vocabulary seems an unnecessary burden to me in the current situation...

msporny:
> Didn't we introduce the `@vocab` feature into the `@context` so that, for example, `https://www.w3.org/ns/credentials/examples/v2 ` would not be necessary any more? Maintaining properly yet another vocabulary seems an unnecessary burden to me in the current situation...

Sure, but you're talking about the base examples that people see and work with. Specifically, all the examples, without an `examples/v2` context, are already confusing people... "I'm getting a bunch of undefined terms... why? Is that supposed to happen? What am I doing wrong?" -- which is the sort of experience we want people to have and then improve... but when our base examples in the core spec do that (same for Data Integrity), implementers get (rightly) concerned.

msporny:
@msporny wrote:
> If this is not acceptable

At this point, given the confusion that only having the base context has created (in the few weeks that we've tried to work with examples that only contain it), I expect DB would be a -1 for having the core examples in the specification not reflect what things like plugfests are using.

> then a proper vocabulary specification should be prepared for all the example terms, using the yaml2vocab tool, and then a redirection in the /ns space. (Is it worth it?)

An alternative would be to use existing JSON-LD Contexts that are being used for pilot or production settings. These include the Open Badges, GS1, Traceability, and Citizenship contexts.

decentralgabe:
@OR13 I'd like to note that the changes I made are copies of the examples in the spec itself, so they should not be controversial whatsoever.

If the spec changes, then this examples are downstream and should be updated. This is not the place to discuss them. As it stands the examples I had are in conflict with the spec.

OR13:
Your examples and the core spec are both malformed. 

I am requesting they be well formed. 

I opened a PR to fix the core spec.

There is more than one way to "fix" this, but I think we should try the simpler solutions, before we try the more complicated ones.

gkellogg:
> B.t.w., the [JSON-LD 1.1 specification](https://www.w3.org/TR/json-ld11/) did the nice tricks of creating a tab for each JSON-LD example, linking to the code snippet in the JSON-LD Playground. Maybe good to follow the example...

The other thing the JSON-LD specs do is to validate, both syntactically and semantically, the examples to make sure that different variations actually said the same thing. This is done using a GitHub action. To many specs end up having syntax issues in their examples.

iherman:
> Since this PR is only about JSON Schema, is it perhaps ok that the JSON-LD is wrong, 

you mean "ok" for now, right?

> but I would like a separate issue filed to ensure that the JSON-LD is corrected in the future.

+1

mprorock (changes requested):
I have no opposition to adding correct links into examples, etc.  I would however like all examples to reflect workable examples with correct URLs to prevent developer confusion, especially as we move to CR - I would like to see this PR merge change requests from @OR13 and get in so that we can move forward with less broken items and then add more work back in.

decentralgabe:
I'd like to restate my earlier comment
>  ... the changes I made are copies of the examples in the spec itself, so they should not be controversial whatsoever.
> If the spec changes, then this examples are downstream and should be updated. This is not the place to discuss them. As it stands the examples I had are in conflict with the spec.

It is strange to rate hole here instead of on a new issue (#1066) that addresses the root cause of the issue. 

It would be unfit for this PR's scope to be expanded to (1) fix all examples and then (2) update json schema examples 

(1) needs to be done in #1066. It does not seem productive to hold off merging this until _all examples_ are fixed. This PR moves from a broken to a less broken state, even if there are changes to be had for examples.

brentzundel (approved):

brentzundel:
@OR13 and @mprorock in light of the discussion on today's call and the closing of #1066 and opening of #1110, are the changes requested here still necessary?

OR13:
This PR can be merged after: https://github.com/w3c/vc-data-model/pull/1110

Which makes the examples that are currently not valid RDF, valid RDF.

decentralgabe:
@OR13 @mprorock since #1110 is merged is this OK to merge too?

mprorock (approved):

OR13 (approved):
ignore the change requests

decentralgabe:
Merging. Dependent PRs addressed, no outstanding conflicts.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-05-10-vcwg#section1-2) on 2023-05-10

- no resolutions were taken

#### 1.2.  update json schema to accurately reflect context defn (pr vc-data-model#1064)

_See github pull request [vc-data-model#1064](https://github.com/w3c/vc-data-model/pull/1064)._




**Manu Sporny:** continuing with other PRs: next item "update json schema to accurately reflect context definition".  

**Gabe Cohen:** this is blocked by a PR that's now been merged, which fixed the example context, which should reserve opposition.  

> *Michael Prorock:* +1.

**Orie Steele:** Manu opened a PR an example context that makes it no longer return a 404.  

**Brent Zundel:** The PR has been merged that makes it a 404.  

> *Michael Prorock:* i have approved.

**Orie Steele:** changing to no objection.  

_See github pull request [vc-data-model#1111](https://github.com/w3c/vc-data-model/pull/1111)._




**Gabe Cohen:** can this be merged with 1111?  

> *Michael Prorock:* please!!!!

**Orie Steele:** can be merged this.  

**Michael Prorock:** yes to merge this and with 1111.  

**Brent Zundel:** 1064 should be merged? No opposition. Merge it.  

**Gabe Cohen:** it's merged.  

</details>

PR #1066, opened 2023-03-15
Remove broken links

OR13:
Relies on issuer defined terms, which is the simplest way to define terms, and does not require the implementer to download several separate JSON-LD context files to make use of the examples in the spec.

- https://www.w3.org/ns/credentials/examples/v2 

^ this link is 404, and causes the examples in the spec to be invalid.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1066.html" title="Last updated on Mar 15, 2023, 5:18 PM UTC (cd4a854)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1066/e231ee3...cd4a854.html" title="Last updated on Mar 15, 2023, 5:18 PM UTC (cd4a854)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -914,8 +914,7 @@ discount from a university. In the example below, Pat receives an alumni
   <span class='comment'>// set the context, which establishes the special terms we will be using
   // such as 'issuer' and 'alumniOf'.</span>
   "@context": [
-    "https://www.w3.org/ns/credentials/v2",
-    "https://www.w3.org/ns/credentials/examples/v2"
+    "https://www.w3.org/ns/credentials/v2"
   ],
   <span class='comment'>// specify the identifier for the credential</span>
   "id": "http://example.edu/credentials/1872",
@@ -976,8 +975,7 @@ the <a>verifier</a> and <a>verified</a>.
         <pre class="example nohighlight" title="A simple example of a verifiable presentation">
 {
   "@context": [
-    "https://www.w3.org/ns/credentials/v2",
-    "https://www.w3.org/ns/credentials/examples/v2"
+    "https://www.w3.org/ns/credentials/v2"
   ],
   "type": "VerifiablePresentation",
   <span class='comment'>// the verifiable credential issued in the previous example</span>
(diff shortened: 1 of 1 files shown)


decentralgabe (approved):

msporny (changes requested):
Objection to the PR, it would make just about every example in the spec use the "undefined terms" property, which would set a bad precedent. 

Alternatives that could achieve consensus are being discussed here:

https://github.com/w3c/vc-data-model/issues/1071

and here:

https://github.com/w3c/vc-data-model/pull/1064#issuecomment-1472127345

It is highly unlikely that there will be consensus to downgrade almost every example in the specification to, what is effectively, a bad practice in general.

OR13:
> Objection to the PR, it would make just about every example in the spec use the "undefined terms" property, which would set a bad precedent.

Actually this fixes that problem (which exists in the spec today), it defines the terms using the default vocab.

The current examples are malformed, and this PR is the simplest possible fix them (since if defines their terms using the default context).

> Alternatives that could achieve consensus include:
> 
> 1. Create a "Getting Started" section that outlines how to get started by using the base context, and then layer semantic meaning after the developer has stabilized the terms they want to use.

This does not fix the problem that the examples in the spec are malformed currently.

Happy to address this in a separate PR.

> 2. Use JSON-LD Contexts that are being used in pilot and production settings.

Introduces further bias, and makes the examples in the spec more complicated.

> 3. Upgrade the base context to include terms like "name" and "website", or other generally useful properties for a credential subject, which could then be used in examples.

This seems like maybe a path forward, make the examples simpler, so you don't need context complexity in them.

You can also use the issuer defined terms to achieve this, that is exactly what the current PR does.

> 4. Use JSON-LD Contexts from the VC Specs Directory.

Introduces further bias, and makes the examples in the spec more complicated.

selfissued (approved):
I agree that all the hyperlinks in our specification should be valid.

OR13:
@Sakurann @brentzundel I would like some call time to discuss this and the JSON Schema PR: 

- https://github.com/w3c/vc-data-model/pull/1064

OR13:
I am fine with the working group adding links that have consensus back to the examples, we can keep having that conversation here:

https://github.com/w3c/vc-data-model/issues/1071

It seems better to remove broken links first, and then add non broken links back.

The links that are removed are not required and the examples are non normative.

Sakurann (approved):

Sakurann:
> It seems to me harder to discover where those broken links were once they're removed, 

why does the WG need to discover old broken links? the one who is interested in fixing it will know where they are

OR13:
I'm closing this PR in favor of https://github.com/w3c/vc-data-model/pull/1111/files

brentzundel:
> I'm closing this PR in favor of https://github.com/w3c/vc-data-model/pull/1111/files

did you mean #1110 ?

PR #1068, opened 2023-03-15
Add 'render' hint property to v2 context

dmitrizagidulin:
This standalone PR adds the `render` hint property to the VC 2.0 `@context`. (Note: this separate from PR #1035).

The `render` hint allows for linking to, or embedding, issuer display preferences / hints in a flexible manner (see the [Rendering Verifiable Credentials](https://github.com/WebOfTrustInfo/rwot11-the-hague/blob/master/draft-documents/rendering-vcs-snapshot-9-27-22.md) paper from RWOT 11 for more info and examples).

The intention is to add examples of `render` usage to the VC Spec directory.

### DIFF
--- a/contexts/credentials/v2
+++ b/contexts/credentials/v2
@@ -63,6 +63,11 @@
           "@id": "https://www.w3.org/2018/credentials#refreshService",
           "@type": "@id"
         },
+        "render": {
+          "@id": "https://www.w3.org/2018/credentials#render",
+          "@type": "@id",
+          "@container": "@graph"
+        }
         "termsOfUse": {
           "@id": "https://www.w3.org/2018/credentials#termsOfUse",
           "@type": "@id"
--- a/contexts/credentials/v2
+++ b/contexts/credentials/v2
@@ -65,8 +65,7 @@
         },
         "render": {
           "@id": "https://www.w3.org/2018/credentials#render",
-          "@type": "@id",
-          "@container": "@graph"
+          "@type": "@id"
         }
         "termsOfUse": {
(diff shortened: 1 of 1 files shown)


contexts/credentials/v2
```diff
@@ -63,6 +63,11 @@
           "@id": "https://www.w3.org/2018/credentials#refreshService",
           "@type": "@id"
         },
+        "render": {
+          "@id": "https://www.w3.org/2018/credentials#render",
+          "@type": "@id",
+          "@container": "@graph"
```
msporny, line 69:
```suggestion
          "@container": "@graph"
```

Why does `render` point to a different graph?

brentzundel:
I assume this PR depends on agreement around #1035. I'm not understanding why we would add `render` to the V2 `@context` otherwise.
Is that correct?

dmitrizagidulin:
> I assume this PR depends on agreement around #1035. I'm not understanding why we would add `render` to the V2 `@context` otherwise. Is that correct?

This is an alternative, lighter-weight proposal to PR #1035 that reserves the term so that extension specifications can use it.

brentzundel:
without spec text indicating that the term is being reserved, I don't think adding `render` to the `@context` file would be appropriate.

OR13 (changes requested):
I do not believe we should add the render property to the core v2 context.

selfissued (changes requested):
I agree with @brentzundel that it's unnecessary to add this property.

Aisp-GitHub:
Sorry it was a mistaken operation about approving the change on adding render property.

dmitrizagidulin:
No consensus was able to be reached, closing PR.

PR #1073, opened 2023-03-31
fix name id

suutaku:
typo

***
<a href="https://pr-preview.s3.amazonaws.com/suutaku/vc-data-model/pull/1073.html" title="Last updated on Apr 3, 2023, 6:54 PM UTC (6cee2d0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1073/dfa4884...suutaku:6cee2d0.html" title="Last updated on Apr 3, 2023, 6:54 PM UTC (6cee2d0)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4922,8 +4922,7 @@ convenience, the base context is also provided below.
           "@type": "@id"
         },
         "name": {
-          "@id": "https://schema.org/name",
-          "@type": "http://www.w3.org/2001/XMLSchema#string"
+          "@id": "https://schema.org/name"
         },
         "proof": {
           "@id": "https://w3id.org/security#proof",
--- a/index.html
+++ b/index.html
@@ -4922,7 +4922,7 @@ convenience, the base context is also provided below.
           "@type": "@id"
         },
         "name": {
-          "@id": "https://schema.org/description",
+          "@id": "https://schema.org/name",
           "@type": "http://www.w3.org/2001/XMLSchema#string"
         },
         "proof": {


index.html
```diff
@@ -4922,7 +4922,7 @@ <h3>Base Context</h3>
           "@type": "@id"
         },
         "name": {
-          "@id": "https://schema.org/description",
+          "@id": "https://schema.org/name",
           "@type": "http://www.w3.org/2001/XMLSchema#string"
         },
```
dlongley, line 4927: It seems this should just be:

"name": "https://schema.org/name",

dlongley (commented):
The hash at the top of the section should be fixed along with this. Notably, the use of `type` with xsd:string is also unnecessary and could be simplified (same applies to `description`).

index.html
```diff
@@ -4922,7 +4922,7 @@ <h3>Base Context</h3>
           "@type": "@id"
         },
         "name": {
-          "@id": "https://schema.org/description",
+          "@id": "https://schema.org/name",
           "@type": "http://www.w3.org/2001/XMLSchema#string"
```
msporny, line 4926:
```suggestion
```

index.html
msporny: Yes, agree (and change applied).

msporny (approved):

OR13:
hmm https://github.com/w3c/vc-data-model/pull/1053

Feels like maybe a data-include would have been a better approach to this... Why is the whole context embedded in html?

PR #1054, opened 2023-02-27
Add confidence method to VCDM

awoie:
Potentially fixes #789, #882 

This PR includes normative changes that define a new property `confirmationMethod` (aka holder binding):
- [x] Add confirmation method to Advanced Concepts Section
- [x] Add confirmation method to Types Section
- [x] Add confirmation method to Extension Points Section

Out-of-scope:
- Privacy and security considerations which requires a separate PR
- Validation section which requires a separate PR

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1054.html" title="Last updated on May 9, 2023, 6:33 PM UTC (05b4034)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1054/fe54b80...05b4034.html" title="Last updated on May 9, 2023, 6:33 PM UTC (05b4034)">Diff</a>

Issue #1080, opened 2023-04-07
(Editorial) systematically adding an anchor to each property


iherman:
At the moment, there is no way to have hyperlinks to properties that are defined in the VCDM specification, e.g., `validFrom` or `credentialSchema`. More specifically, the "formal" specification of these terms are done through a 

```
<dl>
  <dt><var>validFrom</var></dt>
  ...
```

construct; I believe the `<dt>` must have an `@id` property following some clear naming scheme (e.g., `dfn-validFrom`). Even better, maybe using a `<dfn>` would be better; it may allow the VDCM spec to be incorporated into the xref toolkit of other W3C specifications.

Why? I believe it is not a good idea that, e.g., the `credentialSubject` is defined _**both**_ in the VCDM spec and in the vocabulary specification (e.g., at https://www.w3.org/2018/credentials//#credentialSchema). What should happen, in my view, is that the vocabulary specification should refer to the VCDM specification's normative description whenever this makes sense, with only a few words in the vocabulary specification. But the these anchors are the first necessary steps to do so.

(To make it short: the vocabulary specification should be DRY...)

---

- P.S.1. Because I would have to do the updates of the vocabulary definition I am happy to do this in a separate PR on the spec, but only if, in principle, we have an agreement to do this. Hence my auto-assignment to this issue.
- P.S.2. The same editorial principle should be used on _all_ our specs. Not only the data integrity, but _all_ our specs that define new terms, e.g., class instances that are referred to from a VC.

cc @msporny

msporny:
Yes, agreed on this approach. 

I'm less sure of the use of `<dfn>` in the main spec; I've tried to only use `<dfn>` for "terminology associated with prose" and not "literal variable names in code and their associated normative statements". I realize that those two classes of information are *very close*, but they're not directly overlapping. I remember there being some sort of markup issues w/ older version of ReSpec. 

That said, we should experiment and see if we cane make `<dfn>` work everywhere... and if that fails (for any reason), we can fall back to just asserting `id` values by hand in the HTML (yes, I know, not ideal, but that can be the fallback).

iherman:
Ok. I would let you experiment with the `<dfn>` and, once you have decided which one to use I am happy to take over from you filling in the id-s (and linking from the vocabulary).

Sakurann:
should take decisions in #1103 into account.

iherman:
[...]

PR #1082, opened 2023-04-07
Add table of reserved properties

msporny:
This PR attempts to address PRs #987, #1035, #1054, #1068, and other potential future PRs of a similar nature by establishing a table of reserved properties.

The group is currently struggling to achieve consensus around extension points, mostly due to workload, that some would argue as being good for subsets of the ecosystem. Similarly, the group is concerned around defending extension points that it has previously standardized, but for which there are not a minimum of two demonstrably interoperable implementations.

In an effort to provide these extension points, but without needing to standardize implementations for each extension point, the current PR establishes a "table of reserved properties". The expectation is that each of these reserved properties would be listed in the [vc-specs-dir](https://w3c.github.io/vc-specs-dir/#property-extensions) as a potential extension point category, but without there being any normative guidance in the core specification regarding their usage (the guidance would be in the extension specifications). 

If an extension point becomes used over time and can demonstrate a minimum of two demonstrably interoperable implementations, then it would be considered for inclusion in the core specification. If an extension point ends up not being used, then the next WG can make a decision to remove the reserved property or keep it reserved for another implementation cycle.

If this PR is merged, then it's possible for PRs #987, #1035, #1054, and #1068 to be closed and be replaced by reserved property registrations for each property. The upsides being:

1. Reduced possibility of a failure to exit Candidate Recommendation phase.
2. Reduced chance of Formal Objections due to insufficient implementations of extension points.
3. Ability to establish a common extension point while providing more time for the market to sort out the details around the extension point.
4. A more clear process around when an extension point feature should be included in the core VCDM specification (an extension point needs two demonstrably interoperable implementations before normative guidance is added about the extension point).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1082.html" title="Last updated on Apr 22, 2023, 4:34 PM UTC (2d75b77)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1082/4af7219...2d75b77.html" title="Last updated on Apr 22, 2023, 4:34 PM UTC (2d75b77)">Diff</a>

### DIFF
--- a/common.js
+++ b/common.js
@@ -139,7 +139,7 @@ var vcwg = {
       status: 'ED',
       publisher: 'W3C Verifiable Credentials Working Group'
     },
-    'VC-SPECS-DIR': {
+    'VC-SPECS': {
       title: 'Verifiable Credentials Specifications Directory',
       href: 'https://w3c.github.io/vc-specs-dir/',
       authors: ['Manu Sporny'],
--- a/common.js
+++ b/common.js
@@ -138,6 +138,13 @@ var vcwg = {
       authors: ['Orie Steele', 'Michael Jones'],
       status: 'ED',
       publisher: 'W3C Verifiable Credentials Working Group'
+    },
+    'VC-SPECS-DIR': {
+      title: 'Verifiable Credentials Specifications Directory',
+      href: 'https://w3c.github.io/vc-specs-dir/',
+      authors: ['Manu Sporny'],
+      status: 'ED',
+      publisher: 'W3C Verifiable Credentials Working Group'
     }
   }
 };
--- a/index.html
+++ b/index.html
@@ -3270,6 +3270,13 @@ Implementers SHOULD NOT use these properties without a publicly disclosed
 specification describing their implementation.
         </p>
 
+        <p>
+In order to avoid collisions regarding how the following properties are used,
+implementations MUST specify a `type` property in the value associated with the
+reserved property. For more information related to adding `type` information,
+see Section <a href="#types"></a>.
+        </p>
+
         <p class="issue" title="Reserved properties under debate">
 The following extension point properties are under consideration for being
 marked as reserved: `presentationSchema`, `credentialRefresh`, `termsOfUse`,
--- a/index.html
+++ b/index.html
@@ -3255,9 +3255,9 @@ on top of this specification.
         <h3>Reserved Extension Points</h3>
 
         <p>
-This specification reserves a number of extension point <a>properties</a> that
-implementers signalled interest in, where their inclusion in this specification
-was considered to be premature, and where these extension points might be more
+This specification reserves a number of <a>properties</a> serving as possible extension points. While
+some implementers signaled interest in these, their inclusion in this specification
+was considered to be premature; these extension points might be more
 formally defined in future versions of this specification. While these extension
 points are reserved in the base context, their usage is undefined by this
 specification and implementers are cautioned that usage of these properties is
--- a/index.html
+++ b/index.html
@@ -3252,22 +3252,22 @@ on top of this specification.
       </section>
 
       <section class="informative">
-        <h3>Reserved Properties</h3>
+        <h3>Reserved Extension Points</h3>
 
         <p>
-This specification reserves a number of <a>properties</a> that implementers
-signalled interest in, where their inclusion in this specification was
-considered to be premature, and where these properties might be more formally
-defined in future versions of this specification. While these properties are
-reserved in the base context, their usage is undefined by this specification
-and implementers are cautioned that usage of these properties is considered
-experimental.
+This specification reserves a number of extension point <a>properties</a> that
+implementers signalled interest in, where their inclusion in this specification
+was considered to be premature, and where these extension points might be more
+formally defined in future versions of this specification. While these extension
+points are reserved in the base context, their usage is undefined by this
+specification and implementers are cautioned that usage of these properties is
+considered experimental.
         </p>
 
         <p class="issue" title="Reserved properties under debate">
-The following properties are under consideration for being marked as
-reserved: `presentationSchema`, `credentialRefresh`, `termsOfUse`, `evidence`,
-`render`, and `confirmationMethod`.
+The following extension point properties are under consideration for being
+marked as reserved: `presentationSchema`, `credentialRefresh`, `termsOfUse`,
+`evidence`, `render`, and `confirmationMethod`.
         </p>
 
         <table class="simple">
@@ -3291,8 +3291,15 @@ An example entry for a reserved property.
           </tbody>
         </table>
 
+        <p>
+A list of specifications that utilize the extension points defined in this
+specification, as well as the reserved extension points listed above, can
+be found in the Verifiable Credentials Specifications Directory
+[[VC-SPECS-DIR]].
+        </p>
+
+      </section>
     </section>
-  </section>
 
     <section>
       <h2>Syntaxes</h2>
--- a/index.html
+++ b/index.html
@@ -3251,19 +3251,21 @@ on top of this specification.
         </p>
       </section>
 
-      <section class="informative">
+      <section class="normative">
         <h3>Reserved Extension Points</h3>
 
         <p>
 This specification reserves a number of <a>properties</a> serving as possible extension points. While
 some implementers signaled interest in these, their inclusion in this specification
 was considered to be premature; these extension points might be more
-formally defined in future versions of this specification. While these extension
-points are reserved in the base context, their usage is undefined by this
+formally defined in future versions of this specification. It is important to note that they are not defined by this
 specification and implementers are cautioned that usage of these properties is
 considered experimental.
         </p>
-
+Implementers MAY use these properties, but SHOULD expect these terms and/or
+their meanings to change during the process to normatively specify them.
+Implementers SHOULD NOT use these terms without a publicly disclosed
+specification describing their implementation.
         <p class="issue" title="Reserved properties under debate">
 The following extension point properties are under consideration for being
 marked as reserved: `presentationSchema`, `credentialRefresh`, `termsOfUse`,
@@ -3292,10 +3294,11 @@ An example entry for a reserved property.
         </table>
 
         <p>
-A list of specifications that utilize the extension points defined in this
-specification, as well as the reserved extension points listed above, can
-be found in the Verifiable Credentials Specifications Directory
-[[VC-SPECS-DIR]].
+A non-normative list of specifications that stem from the extension points defined
+in this specification, as well as the reserved extension points, can be found in
+the Verifiable Credentials Specifications Directory [[VC-SPECS-DIR]]. Items in
+the directory that refer to reserved extension points SHOULD be treated as
+experimental.
         </p>
 
       </section>
--- a/index.html
+++ b/index.html
@@ -3252,6 +3252,27 @@ on top of this specification.
       </section>
     </section>
 
+    <section class="informative">
+      <h3>Reserved Properties</h3>
+
+      <p>
+This specification reserves a number of <a>properties</a> that implementers
+signalled interest in, where their inclusion in this specification was
+considered to be premature, and where these properties might be more formally
+defined in future versions of this specification. While these properties are
+reserved in the base context, their usage is undefined by this specification
+and implementers are cautioned that usage of these properties is considered
+experimental.
+      </p>
+
+      <p class="issue" title="Reserved properties under debate">
+The following properties are under consideration for being marked as
+reserved: `presentationSchema`, `credentialRefresh`, `termsOfUse`, `evidence`,
+`render`, and `confirmationMethod`.
+      </p>
+    </section>
+  </section>
+
     <section>
       <h2>Syntaxes</h2>
 
--- a/index.html
+++ b/index.html
@@ -3267,7 +3267,7 @@ considered experimental.
         <p class="issue" title="Reserved properties under debate">
 The following extension point properties are under consideration for being
 marked as reserved: `presentationSchema`, `credentialRefresh`, `termsOfUse`,
-`evidence`, `render`, and `confirmationMethod`.
+`evidence`, `render`, `refreshService` and `confirmationMethod`.
         </p>
 
         <table class="simple">
--- a/index.html
+++ b/index.html
@@ -3255,17 +3255,21 @@ on top of this specification.
         <h3>Reserved Extension Points</h3>
 
         <p>
-This specification reserves a number of <a>properties</a> serving as possible extension points. While
-some implementers signaled interest in these, their inclusion in this specification
-was considered to be premature; these extension points might be more
-formally defined in future versions of this specification. It is important to note that they are not defined by this
-specification and implementers are cautioned that usage of these properties is
-considered experimental.
+This specification reserves a number of <a>properties</a> to serve as possible
+extension points. While some implementers signaled interest in these properties,
+their inclusion in this specification was considered to be premature; these
+extension points might be more formally defined in future versions of this
+specification. It is important to note that these properties are not defined by
+this specification and implementers are cautioned that usage of these properties
+is considered experimental.
         </p>
-Implementers MAY use these properties, but SHOULD expect these terms and/or
+        <p>
+Implementers MAY use these properties, but SHOULD expect them and/or
 their meanings to change during the process to normatively specify them.
-Implementers SHOULD NOT use these terms without a publicly disclosed
+Implementers SHOULD NOT use these properties without a publicly disclosed
 specification describing their implementation.
+        </p>
+
         <p class="issue" title="Reserved properties under debate">
 The following extension point properties are under consideration for being
 marked as reserved: `presentationSchema`, `credentialRefresh`, `termsOfUse`,
@@ -3294,11 +3298,11 @@ An example entry for a reserved property.
         </table>
 
         <p>
-A non-normative list of specifications that stem from the extension points defined
-in this specification, as well as the reserved extension points, can be found in
-the Verifiable Credentials Specifications Directory [[VC-SPECS-DIR]]. Items in
-the directory that refer to reserved extension points SHOULD be treated as
-experimental.
+An unofficial list of specifications that are associated with the extension
+points defined in this specification, as well as the reserved extension points
+defined in this section, can be found in the Verifiable Credentials
+Specifications Directory [[?VC-SPECS]]. Items in the directory that refer to
+reserved extension points SHOULD be treated as experimental.
         </p>
 
       </section>
--- a/index.html
+++ b/index.html
@@ -3280,7 +3280,8 @@ see Section <a href="#types"></a>.
         <p class="issue" title="Reserved properties under debate">
 The following extension point properties are under consideration for being
 marked as reserved: `presentationSchema`, `credentialRefresh`, `termsOfUse`,
-`evidence`, `render`, `refreshService` and `confirmationMethod`.
+`evidence`, `renderMethod`, `refreshService`, and
+`confirmationMethod/confidenceMethod`.
         </p>
 
         <table class="simple">
--- a/index.html
+++ b/index.html
@@ -3250,12 +3250,11 @@ this specification and is pursuing that work as an architectural layer built
 on top of this specification.
         </p>
       </section>
-    </section>
 
-    <section class="informative">
-      <h3>Reserved Properties</h3>
+      <section class="informative">
+        <h3>Reserved Properties</h3>
 
-      <p>
+        <p>
 This specification reserves a number of <a>properties</a> that implementers
 signalled interest in, where their inclusion in this specification was
 considered to be premature, and where these properties might be more formally
@@ -3263,13 +3262,35 @@ defined in future versions of this specification. While these properties are
 reserved in the base context, their usage is undefined by this specification
 and implementers are cautioned that usage of these properties is considered
 experimental.
-      </p>
+        </p>
 
-      <p class="issue" title="Reserved properties under debate">
+        <p class="issue" title="Reserved properties under debate">
 The following properties are under consideration for being marked as
 reserved: `presentationSchema`, `credentialRefresh`, `termsOfUse`, `evidence`,
 `render`, and `confirmationMethod`.
-      </p>
+        </p>
+
+        <table class="simple">
+          <thead>
+            <tr>
+              <th>Reserved&nbsp;Property</th>
+              <th>Description</th>
+            </tr>
+          </thead>
+
+          <tbody>
+            <tr>
+              <td>
+`example`
+              </td>
+              <td>
+An example entry for a reserved property.
+              </td>
+            </tr>
+
+          </tbody>
+        </table>
+
     </section>
   </section>
 


index.html
```diff
@@ -3246,6 +3246,55 @@ <h3>Authorization</h3>
...
       </section>
+
+      <section class="informative">
+        <h3>Reserved Extension Points</h3>
+
+        <p>
+This specification reserves a number of extension point <a>properties</a> that
+implementers signalled interest in, where their inclusion in this specification
+was considered to be premature, and where these extension points might be more
+formally defined in future versions of this specification. While these extension
```
iherman, line 3257: This is purely a stylistic comment and, as such, is a matter of taste...

```suggestion
This specification reserves a number of <a>properties</a> serving as possible extension points. While
some implementers signaled interest in these, their inclusion in this specification
was considered to be premature; these extension points might be more
formally defined in future versions of this specification. While these extension
```

iherman (approved):
The question is: what should be the fate of these properties in view of the vocabulary? If they appear in the base context, that means their URL must be dereferencable, which suggests that they should appear in the vocabulary file (note: some of them already do, like `termsOfUse`).

My suggestion is that 

1. They should be added to the formal vocabulary as bona fide (i.e., not as deprecated) properties
2. Their specification text in the vocabulary should signal that they are reserved but not yet fully adopted (and the text should point back to the table that will be created).
3. (If we want to play it really nice, ie, if I am gently pressed...) I can massage the yml2vocab tool to list those in a separate section, just as it does now with deprecated features. 

Note, b.t.w., that schema.org uses the term "pending" for essentially the same issue, see https://schema.org/docs/pending.home.html. We may want to consider to align our terminology and use that term for the same concept.

Final comment, not related to this specification: the current vocabulary definition for the security vocabulary is full of "deprecated" properties and classes, inherited from the CCG days. I wonder whether (1) we should not introduce exactly the same concept in the data integrity spec and (2) declare all the "old" (ie, currently undefined) terms in the security vocabulary into this category rather than simply consider them as deprecated.

msporny:
> The question is: what should be the fate of these properties in view of the vocabulary?

Yes, excellent question...

> 1. They should be added to the formal vocabulary as bona fide (i.e., not as deprecated) properties

+1

> 2. Their specification text in the vocabulary should signal that they are reserved but not yet fully adopted (and the text should point back to the table that will be created).

+1

> 3. (If we want to play it really nice, ie, if I am gently pressed...) I can massage the yml2vocab tool to list those in a separate section, just as it does now with deprecated features.

+1, though this is less of a priority *and* we might want to consider that `deprecated` and `reserved` are two important distinctions that we'd want to have in any vocabulary.

`deprecated` means that we know of the property, it was used in the past, and future usage of the property is discouraged.

`reserved` means that we know of the property, it is expected to be used in the future, and future usage of the property is unknown.

> Final comment, not related to this specification: the current vocabulary definition for the security vocabulary is full of "deprecated" properties and classes, inherited from the CCG days. I wonder whether (1) we should not introduce exactly the same concept in the data integrity spec and (2) declare all the "old" (ie, currently undefined) terms in the security vocabulary into this category rather than simply consider them as deprecated.

Yes, agree that we should introduce the same concept, but in that specification, we definitely want to have a separation between terms that are `deprecated` such as `RsaSignature2018` and terms that are `reserved` like `digestMultibase`.

iherman:
> > The question is: what should be the fate of these properties in view of the vocabulary?
> 
> Yes, excellent question...
> 
> > 1. They should be added to the formal vocabulary as bona fide (i.e., not as deprecated) properties
> 
> +1
> 
> > 2. Their specification text in the vocabulary should signal that they are reserved but not yet fully adopted (and the text should point back to the table that will be created).
> 
> +1
> 
> > 3. (If we want to play it really nice, ie, if I am gently pressed...) I can massage the yml2vocab tool to list those in a separate section, just as it does now with deprecated features.
> 
> +1, though this is less of a priority _and_ we might want to consider that `deprecated` and `reserved` are two important distinctions that we'd want to have in any vocabulary.
> 
> `deprecated` means that we know of the property, it was used in the past, and future usage of the property is discouraged.
> 
> `reserved` means that we know of the property, it is expected to be used in the future, and future usage of the property is unknown.

I am not sure I understand your question. The PR itself makes a clear difference between "deprecated" and "reserved". Simply because the term "deprecated" has the universally accepted meaning that you write above, see, e.g.

> In several fields, especially [computing](https://en.wikipedia.org/wiki/Computing), deprecation is the discouragement of use of some terminology, feature, design, or practice, typically because it has been superseded or is no longer considered efficient or safe, without completely removing it or prohibiting its use.

from [Wikipedia](https://en.wikipedia.org/wiki/Deprecation), which, as we all known, is the repository of the ultimate truth... :-)

In other words, my feeling is that if this PR is accepted by the WG to be merged, then these two categories do exist, there are no further questions there.

But I may miss some nuance.

> 
> > Final comment, not related to this specification: the current vocabulary definition for the security vocabulary is full of "deprecated" properties and classes, inherited from the CCG days. I wonder whether (1) we should not introduce exactly the same concept in the data integrity spec and (2) declare all the "old" (ie, currently undefined) terms in the security vocabulary into this category rather than simply consider them as deprecated.
> 
> Yes, agree that we should introduce the same concept, but in that specification, we definitely want to have a separation between terms that are `deprecated` such as `RsaSignature2018` and terms that are `reserved` like `digestMultibase`.

+1

msporny:
@iherman wrote:
> I am not sure I understand your question. The PR itself makes a clear difference between "deprecated" and "reserved".

Ah, then I just misunderstood what you were suggesting. I thought you were suggesting "rename 'deprecated' to 'reserved' and only have one category". Instead, what you are saying is "add 'reserved' as a category, in addition to the existing category 'deprecated'" -- which is exactly what we want, IMHO.

All good, we agree on all points. :)

index.html
```diff
@@ -3246,6 +3246,55 @@ <h3>Authorization</h3>
...
+points are reserved in the base context, their usage is undefined by this
+specification and implementers are cautioned that usage of these properties is
+considered experimental.
+        </p>
+
+        <p class="issue" title="Reserved properties under debate">
+The following extension point properties are under consideration for being
+marked as reserved: `presentationSchema`, `credentialRefresh`, `termsOfUse`,
+`evidence`, `render`, and `confirmationMethod`.
```
dlongley, line 3266: As mentioned elsewhere, we might want to use `renderHint` both to be clear that it's just a suggestion for rendering and to follow the convention of using nouns.

```suggestion
`evidence`, `renderHint`, and `confirmationMethod`.
```

I continue to think the name `confirmationMethod` has some problems but that is being considered elsewhere as well and we could change it here later as needed.

dlongley (approved):

index.html
Sakurann, line 3266:
```suggestion
`evidence`, `render`, `refreshService` and `confirmationMethod`.
```
refreshService should also be included

Sakurann (changes requested):

decentralgabe (approved):
thanks for putting this together

David-Chadwick:
I think we should differentiate between terms that have already been standardised in V1.1 and new terms that have only recently been proposed. The rationale for this is simple. The previously standardised terms have already been used by various implementors/implementations, but probably with different values e.g. for the type field. No formal interworking has been performed on them yet because a) there weren't any tests in the test suite for them, and/or b) the implementations were only using them between different components of their own products. Does for example a wallet and issuer, or issuer and verifier from the same company qualify as different implementations?  They might be written in different languages and be written by different teams. But they have proven interworking.

index.html
David-Chadwick: Agreed

davidlehn, line 3266: Consider putting these items sorted in a `<ul>`.

iherman:
> I think we should differentiate between terms that have already been standardised in V1.1 and new terms that have only recently been proposed. The rationale for this is simple. The previously standardised terms have already been used by various implementors/implementations, but probably with different values e.g. for the type field. No formal interworking has been performed on them yet because a) there weren't any tests in the test suite for them, and/or b) the implementations were only using them between different components of their own products. Does for example a wallet and issuer, or issuer and verifier from the same company qualify as different implementations? They might be written in different languages and be written by different teams. But they have proven interworking.

This is actually in line with the discussion we were having on the formal (ie, vocabulary) aspect of this change in, e.g., https://github.com/w3c/vc-data-model/pull/1082#issuecomment-1501139867. There is a difference between a term being _deprecated_ and _experimental_.

(Without getting into bike-shedding, I am not sure I like the term "experimental". "Reserved" or "pending" might be better. But I won't lie down the road on this...)

common.js
```diff
@@ -138,6 +138,13 @@ var vcwg = {
       authors: ['Orie Steele', 'Michael Jones'],
       status: 'ED',
       publisher: 'W3C Verifiable Credentials Working Group'
+    },
+    'VC-SPECS-DIR': {
+      title: 'Verifiable Credentials Specifications Directory',
+      href: 'https://w3c.github.io/vc-specs-dir/',
+      authors: ['Manu Sporny'],
+      status: 'ED',
```
OR13, line 146: Does this have to become a NOTE / TR in order for us to reference it informatively?

common.js
```diff
@@ -138,6 +138,13 @@ var vcwg = {
       authors: ['Orie Steele', 'Michael Jones'],
       status: 'ED',
       publisher: 'W3C Verifiable Credentials Working Group'
+    },
+    'VC-SPECS-DIR': {
+      title: 'Verifiable Credentials Specifications Directory',
+      href: 'https://w3c.github.io/vc-specs-dir/',
```
OR13, line 144: I think this changes to a TR link when this becomes a note, right?

index.html
OR13: Does this mean that `credentialSchema` and `credentialStatus` are the only extension points we intend to both reserve and define?

index.html
```diff
@@ -3246,6 +3246,55 @@ <h3>Authorization</h3>
...
+        </p>
+
+        <table class="simple">
+          <thead>
+            <tr>
+              <th>Reserved&nbsp;Property</th>
```
OR13, line 3272: This seems like an embedded registry, sorta like what we have for did spec registries.

Why do we need to repeat the registry in the core spec, why can't we just refer to it where it will exist, like we did with did core?

> Some examples contain terms, both property names and values, that are not defined in this specification. These are indicated with a comment (// external (property name|value)). Such terms, when used in a [DID document](https://www.w3.org/TR/did-core/#dfn-did-documents), are expected to be registered in the DID Specification Registries [[DID-SPEC-REGISTRIES](https://www.w3.org/TR/did-core/#bib-did-spec-registries)] with links to both a formal definition and a JSON-LD context.

- https://www.w3.org/TR/did-core/#conformance

OR13: Seems relevant to our education vertical specific examples as well.

Perhaps with this change, we could make the examples in the spec reflect real world examples.

Or we could include real world examples in an appendix and update the examples in the core spec to be simpler, and more relevant to the sections they are in, instead of education related details.

index.html
OR13: Why do we need to reserve these in the core spec? 

Can't we just define them in the vc spec directory?

Thats what we did with DID core, the did core context does not define things that are in the did spec registries...

https://github.com/w3c/did-spec-registries/blob/main/vocabs/v1/context.jsonld

```
{
  "@context": {
    "@protected": true,
    "id": "@id",
    "type": "@type",

    "alsoKnownAs": {
      "@id": "https://www.w3.org/ns/activitystreams#alsoKnownAs",
      "@type": "@id"
    },
    "assertionMethod": {
      "@id": "https://w3id.org/security#assertionMethod",
      "@type": "@id",
      "@container": "@set"
    },
    "authentication": {
      "@id": "https://w3id.org/security#authenticationMethod",
      "@type": "@id",
      "@container": "@set"
    },
    "capabilityDelegation": {
      "@id": "https://w3id.org/security#capabilityDelegationMethod",
      "@type": "@id",
      "@container": "@set"
    },
    "capabilityInvocation": {
      "@id": "https://w3id.org/security#capabilityInvocationMethod",
      "@type": "@id",
      "@container": "@set"
    },
    "controller": {
      "@id": "https://w3id.org/security#controller",
      "@type": "@id"
    },
    "keyAgreement": {
      "@id": "https://w3id.org/security#keyAgreementMethod",
      "@type": "@id",
      "@container": "@set"
    },
    "service": {
      "@id": "https://www.w3.org/ns/did#service",
      "@type": "@id",
      "@context": {
        "@protected": true,
        "id": "@id",
        "type": "@type",
        "serviceEndpoint": {
          "@id": "https://www.w3.org/ns/did#serviceEndpoint",
          "@type": "@id"
        }
      }
    },
    "verificationMethod": {
      "@id": "https://w3id.org/security#verificationMethod",
      "@type": "@id"
    }
  }
}
```

If it is in the vc data model, it belongs in the v2 context, if it is in the vc specs directory and not the vc-data-model, it does not belong in the vc-data-model v2 context.

OR13 (changes requested):
I would prefer to reference the vc-specs directory, and not maintain a registry internal to our spec that points to it.

But I like the overall direction of the PR.

msporny:
@OR13 wrote:
> I would prefer to reference the vc-specs directory, and not maintain a registry internal to our spec that points to it.

I guess the question that would raise is: Who gets to reserve extension properties in the core vocabulary?

common.js
msporny: > I think this changes to a TR link when this becomes a note, right?

Yes, correct.

index.html
msporny: > Why do we need to reserve these in the core spec?

Because the term will be defined/reserved in the base context.

> Can't we just define them in the vc spec directory?

We can't, for the reasons above.

There is an important difference between: "The WG is reserving this property" and "Some random person on the Internet is reserving this property". I'd argue that we don't want to give random people on the Internet the ability to reserve terms in the core context (nor could they, the WG probably wouldn't allow that).

Reserving a property by the WG signals to the ecosystem that the WG believes that there is a subset of the group that believes that an extension point is reserved, but the WG didn't have time to work on it in this iteration of the specification.

msporny: `refreshService` already has at least two independent demonstrable implementations (that are locked up behind NDAs, unfortunately). `credentialSchema` and `credentialStatus` don't yet, but both are WG work items.

So, perhaps the rule here is: 

1. If it's not a WG work item, or it doesn't already have two independent demonstrable interoperable implementations, it goes on the "reserved extension points" list. OR
2. If it's not a WG work item, it goes on the "reserved extension points" list.

msporny: > Does this mean that `credentialSchema` and `credentialStatus` are the only extension points we intend to both reserve and define?

Yes, seems so.

msporny: > Consider putting these items sorted in a `<ul>`.

These are in an issue marker. This PR is for the concept of having reserved extension points. If this PR goes through, then we'll create a table of extension points with possibly some descriptions of the expected purpose of the extension point along with links to the VC-SPECS-DIR section on each extension point.

index.html
msporny: > Seems relevant to our education vertical specific examples as well.

The examples in the spec feel like an orthogonal issue. The purpose of this section, to establish known extension points that a subset of the group feels are important, is not the same as the text you quoted from the DID Spec.

> Perhaps with this change, we could make the examples in the spec reflect real world examples.

No, that's not the purpose of this PR. We should deal w/ the examples in the spec in a separate PR.

> Or we could include real world examples in an appendix and update the examples in the core spec to be simpler, and more relevant to the sections they are in, instead of education related details.

Yes, +1 to moving away from education-related examples throughout the entire spec. The education-related examples dominated in v1.0 because VCs were largely being adopted by the education sector at the time and a fair number of the people in the group were coming from education-related backgrounds. Additionally, education-related examples seemed to be the least controversial at the time.

In any case, +1 to improving the examples throughout the spec, but we should do that in a separate PR with separate concerns.

index.html
msporny: > As mentioned elsewhere, we might want to use `renderHint` both to be clear that it's just a suggestion for rendering and to follow the convention of using nouns.

-1 to `renderHint` -- they're not hints, they're complete instructions on how to render the credential that the issuer has provided. 

What about `renderer` instead? That's the noun form of `render`, and it is aligned with usage in the field of computer science (graphics model renderer, audio model renderer, 3d print renderer).

> I continue to think the name `confirmationMethod` has some problems but that is being considered elsewhere as well and we could change it here later as needed.

Agree that `confirmationMethod` is problematic and needs to be bike shedded.

OR13:
@msporny 

> I guess the question that would raise is: Who gets to reserve extension properties in the core vocabulary?

Certainly the answer is the VCWG under the consensus and process guidelines of W3C.

There are 2 cases to consider:

1. The term is defined in the TR with normative language, and must be tested.
2. The term is reserved in the TR, but has no normative language and can therefore not be tested.
3. The term is reserved in the TR, has normative language, but you need to understand the vc specs dir to actually use the term or test it.

It feels like we preparing to handle:

case 1 for `credentialStatus` and `credentialSchema`

case 2 for `evidence`, `termsOfUse` and `renderHint`, etc...

For case 2, do we really need to "reserve" these things in the core data model, if we are not going to test them?

iherman:
> @OR13 wrote:
> 
> > I would prefer to reference the vc-specs directory, and not maintain a registry internal to our spec that points to it.
> 
> I guess the question that would raise is: Who gets to reserve extension properties in the core vocabulary?

I try to be a bit formal but, at the same time, practical...

I agree with @OR13 that there is a pretty fine line between these experimental terms (classes or properties) and external registry items. And the difference between the former and the latter might whether the term has been defined/used in the VCDM vocabulary namespace or not. *If* the experimental term has been used as part of the vocabulary namespace (e.g., by having been used via a `@context` that maps it to the VCDM space), then we are supposed to give _some_ description to it in the VCDM specification, because that is the authoritative specification for the vcdm vocabulary (see also #1080 which aims at making this is much cleaner than it is today). For deprecated terms it is probably o.k. to say that the VCDM does not give _any_ description (although, maybe, we could document the deprecated terms as well in a separate table), you would certainly expect _some_ description/definition in the spec for experimental terms. After all, these are terms that is o.k. to use, albeit with caution as for their future evolution.

We should check where we stand with the terms that are listed in the example. What I do know is that [`termOfUse`](https://w3c.github.io/vc-data-model/vocab/credentials/v2/vocabulary.html#termsOfUse) _is_ in the vcdm vocabulary right now, so if we do not want to make it fully normative then, imho, it should be listed in the spec as experimental.

---

Note that, although this discussion has concentrated on the VCDM, the same should apply to the security vocabulary.

index.html
dlongley: I could live with `renderer` even though that sounds more like an application or functional component whereas the values for this field would be parameters or information (or "complete instructions" as you said) describing how to render the credential.

TallTed: `howToRender`?
`renderGuidance`?
`renderMethod`?
`renderTemplate`?

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-04-12-vcwg#section3-1) on 2023-04-12

- no resolutions were taken

#### 3.1. Add table of reserved properties (pr vc-data-model#1082)

_See github pull request [vc-data-model#1082](https://github.com/w3c/vc-data-model/pull/1082)._




**Brent Zundel:** time box 10 mins. for the PRs.  

**Manu Sporny:** PR1082 - please take a look at it. Has to do with PRs that are old and haven't received attention.  

</details>

msporny:
@OR13 wrote:
> case 1 for credentialStatus and credentialSchema
> case 2 for evidence, termsOfUse and renderHint, etc...

AFAICT, the cases are slightly different than what's listed above. I'll use letters in an attempt to not confuse your cases w/ the ones below:

* Case A: The term is defined in a TR published by the WG with normative language, and must be tested. `proof`, `credentialStatus`, `presentationSchema`, and `credentialSchema` fall into this category.
* Case B: The term is reserved in a TR published by the WG, because the term did have normative language that was tested in v1.0 and v1.1, but is being removed from the specification because the WG does not have a current REC-track work item for the extension point. `evidence` and `termsOfUse` fall into this category.
* Case C: The term is reserved in a TR published by the WG, but has no normative language because a REC-track work item for the extension point doesn't exist in the WG. `confirmationMethod`  and `renderer` fall into this category.

> For case 2, do we really need to "reserve" these things in the core data model, if we are not going to test them?

For two reasons:

1. Establishment of properties in the VCDM, fully defined or reserved, is how the "categories" are established in [vc-specs-dir](https://w3c.github.io/vc-specs-dir/#property-based-extensions).
2. If we don't reserve "Case C" properties in the VC vocabulary, they will be defined elsewhere and VCWG will have no authority over those definitions (forever). We're already in this situation w/ `https://schema.org/name` and `https://schema.org/description`, which we're probably ok with... the same level of comfort does not apply to `confirmationMethod` and `renderer`.

@iherman and @OR13, if we don't take the approach being suggested in this PR, then how do we address the two items above?

OR13:
@msporny I know the context URL is normative, but is the context document normative as well?

These term definition issues don't even show up until you apply the context and remove it.

So I wonder if it is appropriate to discuss the fully qualified IRIs for any of these terms in the context of the core data model.

> 1. Establishment of properties in the VCDM, fully defined or reserved, is how the "categories" are established in [vc-specs-dir](https://w3c.github.io/vc-specs-dir/#property-based-extensions).

The document you are referencing is not even a WG note... we don't know what it will be, but it seems odd that it would be driving requirements for the core data model... it should be other way round, if anything.

> 2\. If we don't reserve "Case C" properties in the VC vocabulary, they will be defined elsewhere and VCWG will have no authority over those definitions (forever). We're already in this situation w/ `https://schema.org/name` and `https://schema.org/description`, which we're probably ok with... the same level of comfort does not apply to `confirmationMethod` and `renderer`.

I'm not sure I agree, many implementers may never even notice that schema.org is being taken as a normative dependency in the v2 context, why should it matter if different term definitions are used for other optional data elements?

I don't think we need to reserve terms in the context, especially if there is not consensus to do so... regardless of which web origin controls the term definition.

You don't need to "reserve terms" to have JSON-LD based interoperability, you get that from conformance to JSON-LD.

iherman:
> @iherman and @OR13, if we don't take the approach being suggested in this PR, then how do we address the two items above?

@msporny I am not against the PR... sorry if I said something you misunderstood.

Using a (imho) somewhat clearer terminology, all three of your term categories (A), (B), and (C) are terms that are part of our vocabulary:

> * Case A: The term is defined in a TR published by the WG with normative language, and must be tested. `proof`, `credentialStatus`, `presentationSchema`, and `credentialSchema` fall into this category.

I would call them, for the sake of this discussion, "stable" terms, part of the spec normatively.

> * Case B: The term is reserved in a TR published by the WG, because the term did have normative language that was tested in v1.0 and v1.1, but is being removed from the specification because the WG does not have a current REC-track work item for the extension point. `evidence` and `termsOfUse` fall into this category.

I would call them "deprecated". There was a normative language, ie, they _MUST_ stay in the vocabulary (we do not want to create 404-s if someone dereferences that URL).

> * Case C: The term is reserved in a TR published by the WG, but has no normative language because a REC-track work item for the extension point doesn't exist in the WG. `confirmationMethod`  and `renderer` fall into this category. We prefer implementer do not use these any more.

I would call them "unstable", you called them "experimental": in the future they may become either "stable" or "deprecated", or stay "unstable". Implementers may use them at their own peril. They are not part of the formal standard.

Clearly, it is a bit iffy when a term lands into the "unstable" bin, and we should probably avoid them. But if there are already implementation out there that use JSON-LD with these terms being mapped onto our vocabulary, then they should be here.

OR13:
Labeling the terms seems to be a good path forward, a registry can mark things deprecated, we have that in did spec registries, we used it for publicKeyHex.

I do think we can also choose not to register things, and let the open world model and JSON-LD handle interop without us.

For example, if something is being used in the wild, that does not mean it needs to be registered.

We can register the terms, not use them and not write tests for them... But that seems like it maybe won't get consensus given similar conversations elsewhere.

I don't think adding normative weight to things is always a good idea, especially experimental terms, or mappings.

Id also be ok silently dropping previously normative terms, that we believe have never been used and which don't seem to be valuable enough to mark deprecated.

I don't think treating specs like blockchains and keeping things forever in the latest version is a good thing.

msporny:
@iherman wrote:
> @msporny I am not against the PR... 

Ok, thanks for clearing that up, @iherman.

I'd be fine w/ unstable, stable, and deprecated terminology.

@OR13 wrote:
> Labeling the terms seems to be a good path forward, a registry can mark things deprecated, we have that in did spec registries, we used it for publicKeyHex.

+1

> I do think we can also choose not to register things, and let the open world model and JSON-LD handle interop without us.

+1, in general, though, not the point of contention.

The question in front of the group is: Do we want to explicitly give up control of `confirmationMethod` and `renderer` given that we have multiple organizations in the WG that intend to implement to both of those extension points.

> For example, if something is being used in the wild, that does not mean it needs to be registered.

+1, but not the point of contention.

> We can register the terms, not use them and not write tests for them... But that seems like it maybe won't get consensus given similar conversations elsewhere.

At present, the only one that seems to be objecting is you. Keep in mind that if this PR fails, we're back to square one with figuring out what to do with PRs https://github.com/w3c/vc-data-model/pull/987, https://github.com/w3c/vc-data-model/pull/1035, https://github.com/w3c/vc-data-model/pull/1054, and https://github.com/w3c/vc-data-model/pull/1068.

> I don't think adding normative weight to things is always a good idea, especially experimental terms, or mappings.

+1, but not what's being proposed and not the point of contention.

> Id also be ok silently dropping previously normative terms, that we believe have never been used and which don't seem to be valuable enough to mark deprecated.

We would be dropping terms that organizations have specifically used as extension points, and registered those usages in the VC-SPECS-DIR.

> I don't think treating specs like blockchains and keeping things forever in the latest version is a good thing.

No one is suggesting this.

@iherman and @OR13 -- neither of you addressed what to do as an alternative, namely:

* How are the property extension categories in the VC-SPECS-DIR created?
* What is WG's intention with `evidence`, `termsOfUse`, `confirmationMethod`, and `renderer` .

Please provide a concrete counter-proposal that will address the two items above.

iherman:
@msporny as I said, I am fine with the PR as proposed. My only additional comment (beyond the bike shedding on the terms) isthat these categories should be reflected in the formal vocabulary as well. Actually, I have now a branch on yml2vocab that implements what I propose (not yet a PR, I am waiting for this PR to complete).

index.html
David-Chadwick: @msporny  credentialSchema already has several multiple independent implementations since this was a required property of the JFF plugfest. Multiple suppliers issued VCs with the credentialSchema property in it, and multiple suppliers provided wallets that validated the credentialSchema property. These are not locked up in NDAs and the interworking table of all the suppliers was made public available here https://docs.google.com/spreadsheets/d/1VNsQT_S_dQ1JnStvIJOGo5uuLtjI8m0z1YGS9q55YYk/edit#gid=152952696

OR13:
@msporny 

> No, they show up if you are doing JSON-only processing -- whether or not you recognize and process `evidence`, for example.

Probably better to move this to a thread, unless you mean "evidence" shows up, not https://www.w3.org/2018/credentials#evidence shows up.

In which case, that was the point I was trying to make... It make not impact the core data model, if there is no registered IRI for "evidence" anymore.

> We know that it will be published, either as an evergreen NOTE or a REGISTRY (but called a directory). It will exist in TR space, and it must have registration criteria listed. That's all we need to require us to define what those processes are... which is what this PR does.

Perhaps I missed this, but I think we need to resolve to publish it as NOTE, maybe we did that already... But that does not necessarily solve the issue, since we can't point to a NOTE normatively... can we?

> ... because the WG will have zero input on vocabulary documents that it doesn't control. The question to the WG is, does it want to cede these definitions to organizations other than VCWG. It might very well say "yes"... but once it says "yes", it shouldn't be upset if `confirmationMethod` and `renderer` are defined in ways that it doesn't like and then can not control in the future.

Right, but we have both "VC Vocab" and "Security Vocab" controlled by this working group.

We currently only have a single context which includes references to both.

... and as I noted above, you only see full IRIs after that context (non normative) has been applied.

I am trying to understand what an implementer will be told they MUST understand here.

currently it seems they only need to understand JSON terms like "evidence" and "credentialStatus", not IRIs like https://www.w3.org/2018/credentials#evidence



> Do we want other groups to define these two specific extension points, given that we have at least two organizations per feature in the WG today, that want to implement in ways that are broadly applicable to the ecosystem. It costs the WG almost nothing to reserve the terms (this section + a few lines in a JSON-LD Context file) vs. handing control over who gets to define those features to the broader ecosystem.

Agreed, this is the contention point. I don't think we can control how other groups interpret "evidence" unless we make the context normative, since that is when "this working group" gains control over the term.

https://www.w3.org/2018/credentials#evidence

> At present, the only one that seems to be objecting is you. Keep in mind that if this PR fails, we're back to square one with figuring out what to do with PRs #987, #1035, #1054, and #1068.

Agreed, but... The issue is mostly how these changes will impact implementers, and to what degree they will be required to "care" about the content of the PRs.

I suspect some version of this PR will go through, the important part is being clear on normative requirements, so that implementers don't need to read the NOTE to understand the core spec.

> We would be dropping terms that organizations have specifically used as extension points, and registered those usages in the VC-SPECS-DIR.

Is this really a normative change, if it only shows up after context is applied?

Or is context actually being treated as "normative" in practice, just not explicitly in the spec?

> * How are the property extension categories in the VC-SPECS-DIR created?

I don't see a WG NOTE being the only place to create or advertise JSON-LD extensions, I did answer this point, JSON-LD is an open world model.

> * What is the WG's intention with `evidence`, `termsOfUse`, `confirmationMethod`, and `renderer`?

What are the normative requirements regarding these properties, for implementers when reading only the `vc-data-model` TR ?

v2 context has a vocab, so these are already understood in the context of the issuer, what additional normative processing of them is required?

If we can't be clear, they should be removed, that is my proposal.

mprorock (changes requested):
I would like a simple table of reserved property name and purpose and a statement that these terms are reserved in advance of future standardization or based on prior usage.  A statement that these properties should only be used at the users discretion, and that normative implementation at a later date may come into conflict with usage today.

We should avoid any normative references to to the specs directory or similar.

index.html
```diff
@@ -3246,6 +3246,55 @@ <h3>Authorization</h3>
...
+            </tr>
+
+          </tbody>
+        </table>
+
+        <p>
+A list of specifications that utilize the extension points defined in this
+specification, as well as the reserved extension points listed above, can
+be found in the Verifiable Credentials Specifications Directory
+[[VC-SPECS-DIR]].
```
brentzundel, line 3294: I think it important to clarify the nature of the relationship between this spec and the vc-specs-directory and to provide normative guidance.

```suggestion
A non-normative list of specifications that utilize the extension points defined
in this specification, as well as the reserved extension points, can be found in
the Verifiable Credentials Specifications Directory [[VC-SPECS-DIR]]. Items in
the directory referring to reserved extension points SHOULD be treated as
experimental.
```

brentzundel (changes requested):
Think this should be a normative section so that we can more clearly guide usage of these reserved terms.

I think that whether the terms should also be included in the base context should be a separate conversation.

index.html
dlongley: `renderMethod` sounds good and falls inline with `confidenceMethod`.

TallTed:
> unstable, stable, and deprecated

I like this set.

And @brentzundel's suggestions.

index.html
```diff
@@ -3246,6 +3246,55 @@ <h3>Authorization</h3>
...
+        <p>
+This specification reserves a number of <a>properties</a> serving as possible extension points. While
+some implementers signaled interest in these, their inclusion in this specification
+was considered to be premature; these extension points might be more
+formally defined in future versions of this specification. While these extension
+points are reserved in the base context, their usage is undefined by this
+specification and implementers are cautioned that usage of these properties is
+considered experimental.
+        </p>
+
```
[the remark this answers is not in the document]
mprorock: or:
`Implementer MAY use these properties, but SHOULD expect these terms to change as the process to normatively specify them proceeds. Implementers SHOULD NOT implement without a publicly disclosed specification describing their implementation.`

index.html
```diff
@@ -3246,6 +3246,55 @@ <h3>Authorization</h3>
...
+              <td>
+`example`
+              </td>
+              <td>
+An example entry for a reserved property.
+              </td>
+            </tr>
+
+          </tbody>
+        </table>
```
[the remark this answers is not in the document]
dlongley: IMO, `confirmationMethod` should either be changed to `confidenceMethod` in this suggestion or removed for now.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-04-18-vcwg-special#section2) on 2023-04-18

List of resolutions:

- ***[Resolution No. 1](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-04-18-vcwg-special#resolution1):  the VCWG would like to add a table of reserved properties to the VC Data Model.***

### 2. Add table of reserved properties (pr vc-data-model#1082)

_See github pull request [vc-data-model#1082](https://github.com/w3c/vc-data-model/pull/1082)._




**Brent Zundel:** this pr was raised by manu, everyone seems to be in favor of the general idea.  
… my hope was to discuss proposals for adjustments to it.  

**Manu Sporny:** some background, we have other PRs that are "stuck".  
… they are all trying to add new properties to the Verifiable Credential RDF class in the vc data model.  
… there has been discussion on too much work, and maybe we should not add more work / other items.  
… "what work should we take on".  
… "what should we delay to a future working group".  
… the reason for the PR, is there a way for the group to signal, there is interest in these properties, we want to reserve them.  
… we might want to address things in the past or future, but the working group doesn't want to say anything normative about them.  
… the PR is a way to signal that properties are important, but we are not going to say more about them.  
… seems people are in favor of saying something, we don't know what goes in the table.  
… so that seems to be holding things up.  

**Brent Zundel:** that is pretty close, but i got other feedback.  

**Orie Steele:** I agree with what Manu said, we need something like this PR, I'm glad Manu is saying there are no normative requirements. The thing I'd like clarity on is: some of these properties have existing language in the spec, and our spec has staetments about these properties, if we reserve them and put them in the table, what requirements are on the implementer? What are we reserving, a JSON member, a JSON Member and an IRI, or a name a JSOn member and it's presence in the JSON-LD Context? What is the impact on implementers on reading v2 of the spec. Clarify relationship between this table and VC Specs directory, we could run proposals to get clean consensus..  

**Brent Zundel:** chair hat off, i read the PR, I like the direction... but the main points I have concern with are, if there is a table like this, there should be some normative guidance related to it, what should an implementer do.  
… I wanted to clarify if this PR is just reserving JSON key words, or also IRIs, context. stuff.  
… my feeling is that properties should be in the table or in the spec... not both..  
… I have proposed some changes.  

**Ivan Herman:** my 2 cents, ories questions should be answered, we should reserve. both json members and URIs, I am neutral on the context..  
… implementers should essentially do nothing.  
… other than not use these terms without the URIs.  
… implementers should avoid the terms.  

> *Dave Longley:* +1 to ivan.

**Manu Sporny:** with editor hat on, we have vc-specs-dir with categories.  
… [property based extensions](https://w3c.github.io/vc-specs-dir/#property-based-extensions), supports credentialStatus, credentialSchema, termsOfUse.  
… where did these come from.  
… traditionally they came from the VC Data Model.  
… editor hat on, I am trying to figure out what compels the editor of the vc specs dir, to create new categories.  
… the core data model should provide this guidance to the vc specs dir.  
… editor hat off, I thought it might help to walk people through how we could use the mechanism.  
… lets look at render / renderHint property.  
… one way we could address it, multiple organizations want to do something and there is a RWoT paper for it.  
… lets reserve the term, and say "the wg might do stuff with this term later".  
… then the rendering people can work on it in the CCG, as an extension point the VCWG has defined.  
… we can of course do that anyway.  
… since JSON-LD is open world model.  
… wouldn't it be nice if everyone agreed to how to manager extension points?.  
… render goes in the table reserved properties.  
… work happens in the CCG, using the vocab defined in the VCWG.  
… this would let us move other extension points out of the working group, and to the community group.  

**Michael Prorock:** I like brent's additions, I would accept the PR with those changes.  
… I think it is ok to tack stuff in for an editors draft.  
… if things are well defined, we can remove them from the reserved list.  
… lets get a base line, and let the community do the rest, non normatively.  
… lets use do cleanup to the core spec, which needs to be done.  

**David Chadwick:** I think there is a difference between extension points already defined, and new ones that have never been written in the data model.  
… especially regarding "stay clear of them", they are already defined, and have been used... so we can't provide guidance to stay clear.  
… since they have been defined since v1, and that is totally different from new properties.  
… what about other properties, in what category will "isuee" be reserved.  

**Dave Longley:** we need to be careful doing the wrong thing, telling the CCG not to experiment with these.  

> *Michael Prorock:* quoting from the suggestion: "Implementers SHOULD NOT use these reserved terms outside of experimental situations.".

**Dave Longley:** we want people to work on them, to feed those experiments back to the working group.  
… you should expect use of these to be unstable.  

**Manu Sporny:** agree with dlongley, there is just one line that you have added that is problematic.  
… you "SHOULD NOT use these" is problematic.  
… I agree with [mprorocks suggestion](https://github.com/w3c/vc-data-model/pull/1082#discussion_r1170186655).  
… we should just say: "we are looking for implementer feedback" and "you can use these extension points"..  
… I would support warning people to be careful, and that they are not stable.  
… people should be able to rely on the vocabulary term existing, and that its definition or use might change.  
… +1 to Mike's table.  
… I don't like the normative language telling people to avoid using them.  

> *Ted Thibodeau Jr.:* perhaps "use with informed caution" rather than "SHOULD NOT use".

**Brent Zundel:** implementers should not use these reserved terms outside of experimental situations, I am happy to add additional text to elaborate.  
… it is important to say "Don't use them in production".  
… they are "experimental" we can't let people just do whatever they want with "reserved stuff"..  
… I am not attached to specific language.  

> *Ted Thibodeau Jr.:* +1 brent.

**Brent Zundel:** we all seem to agree in principle, lets get some language.  

**Michael Prorock:** we do need clear normative guidance.  
… you can't assume people will use these things.  
… they are fine for development, but they should remain cautioned until they are defined.  
… I would be happy to add additional words of caution, regarding incubation, testing and interoperability, and you should be aware there is not normative guidance for implementers at this time.  
… people need to be aware of guidance.  

> *Dave Longley:* "if you use these in production, you accept the risk that new versions of the VCDM could be incompatible with your usage".

**Ivan Herman:** as an editor of the formal vocab document... I would like to here answers to orie's question from Manu and others.  
… does introducing a member to the table mean introducing a URL?.  
… the script will need to be aware of "reserved properties" in ontologies.  
… will the URL also be reserved?.  

**Manu Sporny:** yes, it goes in the vocab... I also prefer to have it in the context.  
… I wanted to address Brent's comment, I agree with your proposed language.  
… don't ship this to production language is concerning, the working group will end, and then maintenue mode, and maybe 4 years from now.  
… its highly problematic to reserve something for experimentation for 4 years.  
… for render, we do intend to ship to production before the 4 year timeline.  
… maybe we can point to vc-specs-dir?.  
… it does not mean that it might change, the production use could be broken.  
… the guidance should be, VCWG has identified useful properties for experiments and pilots, and it is ok to go to production, but be aware the WG may change the meaning of the property in the future.  
… based on the table we see today, I can't imagine us making changes.  
… the nuance is important.  
… we should let people use them in production.  

**Brent Zundel:** can other comment on the suggested changes related to the meaning of the terms?.  
… would the swap be fine?.  

> *Michael Prorock:* "Implementer MAY use these properties, but SHOULD expect these terms to change as the process to normatively specify them proceeds. Implementers SHOULD NOT implement without a publicly disclosed specification describing their implementation.".

> *David Chadwick:* should expect -> MAY expect.

> *Dmitri Zagidulin:* +1 to brent's update.

**Michael Prorock:** I wanted to answer Orie's question.  
… I have also added a new suggested text, and I added specification required to it.  
… I play around with evidence, is different from I have a spec that people are using.  
… I think clear normative guidance is important.  
… should they be in vocab? yes. Should they be in the context? I don't have preference.  

**David Chadwick:** I think we should clarify difference between extension points with Type and reserved terms.  
… for example issuee, is a property.  

**Ivan Herman:** reacting to what david says, I am wondering what you mean by type.  
… type has a specific meaning.  
… reacting to what Manu said, the 4 year delay thing... when we end the WG and plan for maintenance, we have the right to declare that this document is "living standard" a.k.a. "evergreen".  
… which means the maintenance WG can add normative changes.  
… these extension points could be added here.  

**Manu Sporny:** I like evergreen, but other don't.  
… the argument against evergreen, is that the smaller group can modify large normative sections of the spec.  
… we can try it again, maybe the group has more trust now.  
… regarding what david teams by type.  
… I think he means `@type` aliased to "type" and with a value of an RDF class..  

> *David Chadwick:* correct.

**Manu Sporny:** it would be normative guidance to add the "type" requirement, and decentralized extensibility is supported by type.  
… can we run proposals?.  
… the working group desires a reserved properties table, the second one could start with the values mprorock has provided.  
… the third one could be on guidance for implementers regarding reserved status.  
… encouraging people to use them, but with caution, that they might change in the future.  

> *Dave Longley:* for people reading JSON i think we're suggesting that types are required like this: `"extensionPoint": {"type": "MustDefineThis", ...}` or `"extensionPoint": [{"type": "MustDefineThis", ...}, ...]`.

> *Dave Longley:* +1 to requiring a type for the extension points.

> *Dave Longley:* (for the values of the extension points).

**Ivan Herman:** DavidC has convinced me, that we have to put these entries in the context.  
… because of `@vocab`.  
… if the term has a URL in the vocabulary, then we MUST put it in the context file, or the vocab will destroy the consistency.  

> *Dave Longley:* +1 to Ivan.

> *Manu Sporny:* +1 to Ivan.

**Orie Steele:** Similar thing to what Ivan said, guidance around warning that term definition would change is equivalent to what term definition would be to be expanded by vocabulary. If any term is not defined in context, it's defined in way that issuer intended it to be defined, if issuer doesn't know, then vocab expansion for that term is correct. It's not with a reserved property for what evidence is... semantic confusion around risk of URL changing would be and also find it concerning to add terms to core context since it's non-normative, experimental context don't need to be added to normative document and shouldn't harm implementers perspective... in general supportive of reserving things and experiment, not supportive of being in core spec and blending w/ normative terms, normative requirements, non-normative, and no guarantee of consistency and no guarantee that term will change in the future, future WGs change definition, putting it in context is saying it won't change, that's the issue..  

> *Michael Prorock:* +1 brent.

> *Manu Sporny:* +1 brent.

**Brent Zundel:** the reason I suggested the changes I did, I think we can address the context issues in separate PR.  

> *Dave Longley:* +1 to brent.

> *David Chadwick:* are these reserved terms or reserved extensions points? they are different.

> *Dave Longley:* maybe we want to call them unstable or prospective terms? ... since "reserved" isn't quite right per our conversation.

> **Proposed resolution: the VCWG would like to add a table of reserved properties to the VC Data Model.** *(Brent Zundel)*

> *Manu Sporny:* +1.

**David Chadwick:** I want a separate table.  

> *Michael Prorock:* +1.

> *Ivan Herman:* +1.

> *Dave Longley:* +1.

> *Chris Abernethy:* +1.

> *Orie Steele:* 0.

> *David Chadwick:* -.001.

> *Dmitri Zagidulin:* +1.

> *Gregory Natran:* +1.

> *Ted Thibodeau Jr.:* +1.

**David Chadwick:** it seems unclear, I am not sure what it means.  

> *Charles Lehner:* 0.

> ***Resolution #1: the VCWG would like to add a table of reserved properties to the VC Data Model.***

**Brent Zundel:** will you object formally?.  

**Manu Sporny:** I will prepare the separate proposal.  

> *Dave Longley:* recommends we use "prospective" terms instead of "reserved" to encourage incubation vs. discourage it..

> *Dave Longley:* want "confirmationMethod" to say "confidenceMethod" ... or leave it off for now.

> *David Chadwick:* -1.

**David Chadwick:** it seems it is trying to do too much at once.  
… we need to define things, with resolutions, before we can list them.  

**Brent Zundel:** anything else?.  

> *David Chadwick:* then we should call the table "extension points".

**Orie Steele:** Warning of terms in existing context -- brent, you said you wanted to separate context conversation from term conversation, WG needs to be aware that v2 context contains these terms. The correct thing to say here is, if you want to separate the conversations, v2 context already contains, we need to resolve removing terms from v2 context. To make table more concrete, modified version of table currently defined in core context and removing terms would be a good starting point..  

> *Gregory Natran:* +1 to dlongley's suggestion on confirmation vs. confidence method..

**Ivan Herman:** i disagree with orie, the context is just a mapping.  
… it is not the "core thing".  
… if we agreed, that these terms have a URL in the vocab, the context is the right thing too do.  

> *Manu Sporny:* I also disagree w/ Orie's statements about using `@vocab` to establish these terms, and removing things like evidence, and termsOfUse, from the v2 context..

> *Dave Longley:* -1 to removing from the core context, these prospective terms should NOT be "issuer defined" via `@vocab`, they should be treated as terms that will be defined by a future WG, that's the whole point..

> *David Chadwick:* +1 to Ivan.

> *Michael Prorock:* note, my suggestions says: "Reserved Property" - these are JSON properties.

**Ivan Herman:** otherwise people might just use the JSON term.  
… without the URL.  
… the context is just a mapping.  
… I don't understand what you mean, by "extension" point.  
… we are talking about "Classes" or "Properties".  

**Brent Zundel:** we need to end the meeting.  
… this specific PR seems to have general favor.  
… for parts.  
… and we should raise issues for extension points vs terms and vocabulary and context.  

> *David Chadwick:* an extension point is a subclass of a json term.

> *Ivan Herman:* DavidC, I still do not understand. Json term does not have a subclass....

> *David Chadwick:* conceptually it may have as there can be different types of json terms.

> *Dave Longley:* ivan: i believe DavidC is referring to the range of the term.

> *Manu Sporny:* my understanding as well ^.

---


</details>

index.html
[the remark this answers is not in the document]
TallTed: A healthy blend of the text above from @brentzundel & @mprorock and then some...
```suggestion
Implementers MAY use these properties, but SHOULD expect these terms and/or their meanings to change during the process to normatively specify them. Implementers SHOULD NOT use these terms without a publicly disclosed specification describing their implementation.
```

OR13:
It is getting pretty hard to review this PR with so many suggestions.

Would it be possible to open smaller competing PRs to compare?

index.html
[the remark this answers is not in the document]
OR13: I suggest we not include it, if there is not yet even consensus on the json member name.

index.html
OR13: I like this language.

index.html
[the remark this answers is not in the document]
OR13: > Implementers SHOULD NOT implement without a publicly disclosed specification
describing their implementation.

to:

Registrations of a `type` for an extension `property` MUST be defined in a specification that is publicly accessible in order to be included in the vc-specs-dir.

index.html
```diff
@@ -3246,6 +3246,55 @@ <h3>Authorization</h3>
 on top of this specification.
         </p>
       </section>
+
+      <section class="informative">
```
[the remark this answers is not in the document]
OR13: I agree, implementers need to understand, what they don't need to understand and vice versa... we are also writing this section for "registrants", and they MUST understand these requirements, or the vc specs dir is truly a blank check.

index.html
```diff
@@ -3246,6 +3246,55 @@ <h3>Authorization</h3>
...
+        <h3>Reserved Extension Points</h3>
+
+        <p>
+This specification reserves a number of <a>properties</a> serving as possible extension points. While
+some implementers signaled interest in these, their inclusion in this specification
+was considered to be premature; these extension points might be more
+formally defined in future versions of this specification. While these extension
+points are reserved in the base context, their usage is undefined by this
```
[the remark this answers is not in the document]
OR13: I agree with this change.

index.html
```diff
@@ -3246,6 +3246,55 @@ <h3>Authorization</h3>
 on top of this specification.
         </p>
       </section>
+
+      <section class="informative">
+        <h3>Reserved Extension Points</h3>
```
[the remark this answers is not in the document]
TallTed: I don't think "Prospective" suggests anything like "consensus in the group that they will eventually be included & defined" (or is your issue with the text embellishment that is part of the discussion within this PR, but not intended to be part of the document text?), but I also don't have a problem with "Reserved".

index.html
msporny: I'm not opposed to `renderMethod`

msporny:
> It is getting pretty hard to review this PR with so many suggestions.
> 
> Would it be possible to open smaller competing PRs to compare?

No, please don't. There was fairly clear guidance during the call yesterday that would result in a merge-able PR. I plan to make those changes today to this PR.

OR13:
@msporny ok, can you resolve conversations so it is easier to review just the things you want in your change set?

index.html
brentzundel: I'm fine with Ted's version

index.html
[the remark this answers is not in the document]
brentzundel: I don't think the text should be changed to point to VC Specs Dir here.

I support Ted's suggestion as written.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-04-19-vcwg#section3-2) on 2023-04-19

- no resolutions were taken

#### 3.2. Add table of reserved properties (pr vc-data-model#1082)

_See github pull request [vc-data-model#1082](https://github.com/w3c/vc-data-model/pull/1082)._




**Brent Zundel:** I'll time box it and see how it's going at 5 minutes and stop it at 10 so we have time for the other work items.  
… We had a good conversation about this yesterday and the impression I got is that there were changes suggested and if a number of those get committed to the PR then the resulting text would be acceptable, I believe, to the three people who have formally requested changes.  
… I think we're in a pretty good state.  

**Manu Sporny:** Agree with that.  
… I was going to try and highlight the specific changes to try to apply that will hopefully result in the PR going in. I think Brent want a table of reserved "something or others" in the document. So that section should exist. I believe the changes that a number of people want ... are we should tell people what they should expect and implementers need to do something with what's in there.  
… So that section should be normative.  
… There is some language that Brent proposed that should go in there with normative statements. That will go in there.  
… I think David Chadwick said we should have to say whether "type" is expected or not -- make it mandatory -- I think that's a safe thing to say. You also wanted to say something else I'd like to handle in another PR.  
… Orie mentioned context and vocab values -- and I think we should handle that in a separate PR.  
… Mike Prorock expected a table with values and with my editor hat off I agree but I don't think people were ready for that in this PR yet.  
… I think that's the delta to be made to this PR to merge it in -- the question to the group is if those changes were made would anyone object?.  

**Orie Steele:** That matches my understanding.  

**Brent Zundel:** That matches my understanding.  

**Orie Steele:** There are a bunch of things you could try and include in your revised PR that might snag it again. The current terms in the core data model or current terms in the core context -- if there isn't consensus to how to handle those you shouldn't put those in your PR.  

> *Manu Sporny:* yes "+1 to we are going to make a table and we're going to figure it out".

**Orie Steele:** Specific URLs same thing. The types piece ... if you register a type for an extension property then you must provide a spec.  
… Specifics will snag the PR.  

> *Manu Sporny:* yes "+1 to saying you have to specify a type, but not specify specific types".

**Orie Steele:** I think it's worth moving normative processing should be moved to another PR.  
… If you're careful to revise the text and careful not to pull things in related to those items I think it will sail through, fingers crossed.  

> *Manu Sporny:* yes "+1 to not pulling in relationships to `@context` and Vocabulary URLs, etc.".

**David Chadwick:** I wanted to bring clarity -- we want to talk about the property names and types and we need both. All our extension points have a type -- and if you have a subtype of the extension point.  
… I know Ivan came back and said he didn't know what the extension point was and people need some clarity perhaps on simple properties and extension point types.  

> *Orie Steele:* this is where increased knowledge of `@type` and `@container`... would be helpful to the group... and hence the need to *understand* the context..

**David Chadwick:** My understanding is that an extension point is like "termsOfUse" or whatever and then there's a type for the value that is registered with that and you need a subtype for it.  
… I dont' think we're going to change those -- we may tweak those, we aren't going to change for example an "evidence" might contain a "revocation list" or the status extension point will contain "terms of use".  
… We may have to tweak them in a minor way but that's really it, those were my comments.  

**Manu Sporny:** I broadly agree with what David said -- I think we're going to have a table and I don't think anything we put in there will change in a drastic way.  
… Going back to David's initial question, when we say extension point ... this PR is just going to say the extension point is going to have a type. That puts things that don't have a type in limbo. In the RDF world those are just literal values. If we wanted to specify something that's always a text string we won't be able to do that with this next PR.  
… Then we have to have a discussion around whether we want the table to include those things.  

> *Orie Steele:* I note the utility of what Manu said regarding, in RDF we call these a "literal value"... and our core data model in JSON-LD, which compels implementers to understand this..

**Manu Sporny:** Ivan, tell me if I'm wrong here, but I think you were saying this is an RDF property that can point to anything, a literal value like a text string, or another object out there -- if we don't give it a very specific range of values it can be. That's the default.  
… With the "issuee" thing, I think... you're concerned ... I would argue with that, that can be something that points to something with a "type". We should have that discussion in an issue in a different PR.  
… I think we should just say in this PR that these extension points have types associated with them and define that in a future PR.  

**Brent Zundel:** The things we are concerned with reserving, to my understanding, and one of the reasons for this table is so that we have a place to put extension points. Because those are the things that are sort of normatively there and yet we anticipate that will get us into trouble in v2.0 of the spec.  

> *Orie Steele:* +1 brent.

> *Manu Sporny:* +1 brent.

**Brent Zundel:** We want to include things that don't have a way of normatively specifying what they are ... but we reserve them is the purpose of this.  
… Because of the nature of the things we're reserving is that they are unspecified which necessarily means they might change.  

**Ivan Herman:** I'm still looking for something that I understand, my mind seems to be narrow minded. I think in terms of RDF here. Which has classes and properties.  
… My understanding is that we are here defining placeholders for classes and properties that are there and we leave them open and any one of those can play a role as an "extension point" if I take it as an English term.  
… I don't know how that term "extension point" can be explained in RDF because from my perspective at the end of the day this has to be reflected in a vocabulary. We don't have to do that now, we can do that later.  
… We should clarify what extension points are later in a separate PR.  

**Brent Zundel:** I believe Orie has some things that he might want discussed here so I wanted to give you time here.  
… To go over PRs.  

</details>

index.html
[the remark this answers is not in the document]
msporny: Noting that we agreed to delay the inclusion of this table to a future PR, I'm going to not merge this commit. There will be a future PR that fills out the table, almost certainly using @mprorock's suggestion above as the base for that PR.

common.js
msporny: Resolving since this comment is not a request to change this PR (and the question has been answered).

index.html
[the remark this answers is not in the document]
msporny: Resolving since this change request is being put off to a future PR.

index.html
msporny: Resolving since this conversation will not result in changes to this PR.

brentzundel (approved):

David-Chadwick (changes requested):
Some (most) of these properties have already been formally defined in v1.0 and v1.1 with the words MUST  etc. being part of their definitions. It is not clear to me what it means when v2 says "might be more formally defined in future versions". More formally than what? Does it mean the MUST might be changed to MAY, or further MUSTs added, or simply that the semantics might be tweaked somewhat? There is a big difference in these two extremes.

awoie (approved):
This looks good to me.

index.html
```diff
@@ -3250,6 +3250,70 @@ <h3>Authorization</h3>
...
+their inclusion in this specification was considered to be premature; these
+extension points might be more formally defined in future versions of this
+specification. It is important to note that these properties are not defined by
+this specification and implementers are cautioned that usage of these properties
+is considered experimental.
+        </p>
+        <p>
+Implementers MAY use these properties, but SHOULD expect them and/or
+their meanings to change during the process to normatively specify them.
+Implementers SHOULD NOT use these properties without a publicly disclosed
```
OR13, line 3269: Might want to move this to the directory, as a registration requirement.

OR13 (approved):

selfissued (approved):
I would delete "Implementers SHOULD NOT use these properties without a publicly disclosed
specification describing their implementation." but I won't block on this.

mprorock (approved):

index.html
```diff
@@ -3250,6 +3250,70 @@ <h3>Authorization</h3>
...
+reserved property. For more information related to adding `type` information,
+see Section <a href="#types"></a>.
+        </p>
+
+        <p class="issue" title="Reserved properties under debate">
+The following extension point properties are under consideration for being
+marked as reserved: `presentationSchema`, `credentialRefresh`, `termsOfUse`,
+`evidence`, `renderMethod`, `refreshService`, and
+`confirmationMethod/confidenceMethod`.
+        </p>
```
awoie, line 3285: I'd remove the concrete reserved properties for now and only merge the PR with an empty list. A first stab on the list can then follow. Otherwise looks good to me.

awoie (changes requested):

index.html
brentzundel: In the previous conversations we've held there has been wide agreement that this PR should not contain any entries in the table, so it doesn't.
 PR #1097 has been raised to introduce them.

msporny:
@David-Chadwick wrote:
> Some (most) of these properties have already been formally defined in v1.0 and v1.1 with the words MUST etc. being part of their definitions. It is not clear to me what it means when v2 says "might be more formally defined in future versions". More formally than what? Does it mean the MUST might be changed to MAY, or further MUSTs added, or simply that the semantics might be tweaked somewhat? There is a big difference in these two extremes.

I raised the following issue to track this question: https://github.com/w3c/vc-data-model/issues/1098

I expect we will need further discussion in the WG to clarify the language.

msporny:
At this point, everyone that requested changes that achieved consensus in the last topic call have had changes applied and re-reviewed (in a positive way). @David-Chadwick asked a question that needs further discussion and is being tracked in an issue. @awoie requested changes, but was informed by @brentzundel that a separate PR is meant to address his concerns.

Therefore:

Normative, multiple reviews, changes requested and made, changes requiring further discussion converted to issues or future PRs, no objections, merging.

PR #1083, opened 2023-04-11
Add issuer dependent term handling

OR13:
This PR attempts to provide guidance when an issuer or holder has failed to define all terms in a JSON-LD context.

### DIFF
--- a/vocab/credentials/v2/issuer-dependent.html
+++ b/vocab/credentials/v2/issuer-dependent.html
@@ -72,7 +72,7 @@
             as defined in [vc-data-model], provides a mapping from the standard terms used in credentials to the vocabulary 
             items defined by the Verifiable Credentials Working Group through different vocabularies, 
             such as the <a href="https://www.w3.org/2018/credentials/">Verifiable Credentials Vocabulary</a>. 
-            Applications using additional, issuer-dependent vocabularies of their own are encouraged to 
+            Applications using additional, issuer-independent vocabularies of their own are encouraged to 
             create their own formal vocabularies, and corresponding JSON-LD context definitions, 
             to ensure the interoperability of the credentials they produce and/or consume.
         </p>
--- a/vocab/credentials/v2/issuer-dependent.html
+++ b/vocab/credentials/v2/issuer-dependent.html
@@ -85,40 +85,22 @@
         <p>
             This document provides a proper landing place for those URLs, without any additional definitions for the terms themselves. 
         </p>
-        <section id="dynamic-term" style="color: #FF5C5C;"></section>
+        <section style="color: #FF5C5C;">
+          <p>
+          The term you are looking for was not defined in a JSON-LD context.
+          </p>
+          <p>
+          Please consider contacting the <a href="https://www.w3.org/2018/credentials/#issuer">issuer</a> 
+          or <a href="https://www.w3.org/2018/credentials/#holder">holder</a> if you believe this is an error.
+          </p>
+          <p>You may also consider reviewing:</p>
+          <ul>
+            <li><a href="https://www.w3.org/TR/json-ld11/#terms">JSON-LD v1.1: Terms</a></li>
+            <li><a href="https://www.w3.org/2018/credentials">The Verifiable Credential Vocabulary</a></li>
+            <li><a href="https://github.com/w3c/vc-specs-dir">The Verifiable Credentials Specification Directory</a></li>
+          </ul>
+        </section>
     </section>
-    <script>
-      function htmlEncode(str){
-          return String(str).replace(/[^\w. ]/gi, function(c){
-              return '&#'+c.charCodeAt(0)+';';
-          });
-      }
-      (async ()=>{
-        // handle issuer / holder defined terms in fragment
-        if (window.location.hash.length){
-          const [_, term] = window.location.hash.split('#')
-          let el = document.getElementById('dynamic-term')
-          el.id = term;
-          const encodedTerm = htmlEncode(term);
-          el.innerHTML = `    
-<h2>${encodedTerm}</h2>
-<p>
-The term <code><strong>${encodedTerm}</strong></code> (<code>https://www.w3.org/ns/credentials/issuer-dependent#${encodedTerm}</code>)
-was not defined in a JSON-LD context.
-</p>
-<p>
-Please consider contacting the <a href="https://www.w3.org/2018/credentials/#issuer">issuer</a> 
-or <a href="https://www.w3.org/2018/credentials/#holder">holder</a> if you believe this is an error.
-</p>
-<p>You may also consider reviewing:</p>
-<ul>
-  <li><a href="https://www.w3.org/TR/json-ld11/#terms">JSON-LD v1.1: Terms</a></li>
-  <li><a href="https://www.w3.org/2018/credentials">The Verifiable Credential Vocabulary</a></li>
-  <li><a href="https://github.com/w3c/vc-specs-dir">The Verifiable Credentials Specification Directory</a></li>
-</ul>
-`
-        }
-      })()
-    </script>
+
 </body>
 </html>
--- a/vocab/credentials/v2/issuer-dependent.html
+++ b/vocab/credentials/v2/issuer-dependent.html
@@ -44,6 +44,8 @@
         noLegacyStyle: false
       };
     </script>
+
+    
   </head>
   <body>
     <section id="abstract">
@@ -66,15 +68,58 @@
     <section>
         <h2>Issuer Dependent Terms</h2>
         <p>
-            The <a href="https://www.w3.org/TR/vc-data-model-2.0/#base-context">base context</a> of a Verifiable Credential, as defined in [vc-data-model], provides a mapping from the standard terms used in credentials to the vocabulary items defined by the Verifiable Credentials Working Group through different vocabularies, such as the <a href="https://www.w3.org/2018/credentials/">Verifiable Credentials Vocabulary</a>. Applications using additional, user-dependent vocabularies of their own are encouraged to create their own formal vocabularies, and corresponding JSON-LD context definitions, to ensure the interoperability of the credentials they produce and/or consume.
+            The <a href="https://www.w3.org/TR/vc-data-model-2.0/#base-context">base context</a> of a Verifiable Credential, 
+            as defined in [vc-data-model], provides a mapping from the standard terms used in credentials to the vocabulary 
+            items defined by the Verifiable Credentials Working Group through different vocabularies, 
+            such as the <a href="https://www.w3.org/2018/credentials/">Verifiable Credentials Vocabulary</a>. 
+            Applications using additional, user-dependent vocabularies of their own are encouraged to 
+            create their own formal vocabularies, and corresponding JSON-LD context definitions, 
+            to ensure the interoperability of the credentials they produce and/or consume.
         </p>
         <p>
-            There may be cases, however, when the creation of such formal vocabularies and context definitions is too much of a burden, such as for closed applications, experiments, or examples. The <a href="https://www.w3.org/TR/vc-data-model-2.0/#base-context">base context</a> has been constructed in such a way that issuers may use new, hitherto undefined terms without further ado, and these terms will be mapped onto URLs of the form <code>https://www.w3.org/ns/credentials/issuer-dependent#TERM</code>.
+            Not all users desire to define JSON-LD vocabularies or to make use of them with W3C Verifiable Credentials. 
+            The <a href="https://www.w3.org/TR/vc-data-model-2.0/#base-context">base context</a> has been constructed 
+            in such a way that issuers may use new, hitherto undefined terms without further ado, 
+            and these terms will be mapped onto URLs of the form <code>https://www.w3.org/ns/credentials/issuer-dependent#TERM</code>.
         </p>
         <p>
             This document provides a proper landing place for those URLs, without any additional definitions for the terms themselves. 
         </p>
+        <section id="dynamic-term" style="color: #FF5C5C;"></section>
     </section>
-
+    <script>
+      function htmlEncode(str){
+          return String(str).replace(/[^\w. ]/gi, function(c){
+              return '&#'+c.charCodeAt(0)+';';
+          });
+      }
+      (async ()=>{
+        // handle issuer / hodler defined terms in fragment
+        if (window.location.hash.length){
+          const [_, term] = window.location.hash.split('#')
+          let el = document.getElementById('dynamic-term')
+          el.id = term;
+          const encodedTerm = htmlEncode(term);
+          el.innerHTML = `    
+<h2>${encodedTerm}</h2>
+<p>
+The term <code><strong>${encodedTerm}</strong></code> (<code>https://www.w3.org/ns/credentials/issuer-dependent#${encodedTerm}</code>)
+was not defined in a JSON-LD context.
+</p>
+<p>
+Please consider contacting the <a href="https://www.w3.org/2018/credentials/#issuer">issuer</a> 
+or <a href="https://www.w3.org/2018/credentials/#holder">holder</a> if you believe this is an error.
+</p>
+<p>You may also consider reviewing:</p>
+<ul>
+  <li><a href="https://www.w3.org/TR/json-ld11/#terms">www.w3.org/TR/json-ld11/#terms</a></li>
+  <li><a href="https://www.w3.org/2018/credentials">www.w3.org/2018/credentials</a></li>
+  <li><a href="https://github.com/w3c/vc-specs-dir">w3c/vc-specs-dir</a></li>
+</ul>
+<p>You might also consider searching for the term in a search engine, possibly with "+jsonld" in the query.</p>
+`
+        }
+      })()
+    </script>
 </body>
 </html>
--- a/vocab/credentials/v2/issuer-dependent.html
+++ b/vocab/credentials/v2/issuer-dependent.html
@@ -72,7 +72,7 @@
             as defined in [vc-data-model], provides a mapping from the standard terms used in credentials to the vocabulary 
             items defined by the Verifiable Credentials Working Group through different vocabularies, 
             such as the <a href="https://www.w3.org/2018/credentials/">Verifiable Credentials Vocabulary</a>. 
-            Applications using additional, user-dependent vocabularies of their own are encouraged to 
+            Applications using additional, issuer-dependent vocabularies of their own are encouraged to 
             create their own formal vocabularies, and corresponding JSON-LD context definitions, 
             to ensure the interoperability of the credentials they produce and/or consume.
         </p>
--- a/vocab/credentials/v2/issuer-dependent.html
+++ b/vocab/credentials/v2/issuer-dependent.html
@@ -79,7 +79,7 @@
         <p>
             Not all users desire to define JSON-LD vocabularies or to make use of them with W3C Verifiable Credentials. 
             The <a href="https://www.w3.org/TR/vc-data-model-2.0/#base-context">base context</a> has been constructed 
-            in such a way that issuers may use new, hitherto undefined terms without further ado, 
+            in such a way that issuers may use new, as-yet undefined terms without additional effort, 
             and these terms will be mapped onto URLs of the form <code>https://www.w3.org/ns/credentials/issuer-dependent#TERM</code>.
         </p>
         <p>
--- a/vocab/credentials/v2/issuer-dependent.html
+++ b/vocab/credentials/v2/issuer-dependent.html
@@ -94,7 +94,7 @@
           });
       }
       (async ()=>{
-        // handle issuer / hodler defined terms in fragment
+        // handle issuer / holder defined terms in fragment
         if (window.location.hash.length){
           const [_, term] = window.location.hash.split('#')
           let el = document.getElementById('dynamic-term')
--- a/vocab/credentials/v2/issuer-dependent.html
+++ b/vocab/credentials/v2/issuer-dependent.html
@@ -79,7 +79,7 @@
         <p>
             Not all document authors desire to define JSON-LD vocabularies or to make use of them with W3C Verifiable Credentials. 
             The <a href="https://www.w3.org/TR/vc-data-model-2.0/#base-context">base context</a> has been constructed 
-            in such a way that issuers may use new, as-yet undefined terms without additional effort, 
+            in such a way that issuers might use new, as-yet undefined terms without additional effort, 
             and these terms will be mapped onto URLs of the form <code>https://www.w3.org/ns/credentials/issuer-dependent#TERM</code>.
         </p>
         <p>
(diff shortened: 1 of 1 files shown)


decentralgabe (approved):

vocab/credentials/v2/issuer-dependent.html
```diff
@@ -66,15 +68,58 @@
...
+            items defined by the Verifiable Credentials Working Group through different vocabularies, 
+            such as the <a href="https://www.w3.org/2018/credentials/">Verifiable Credentials Vocabulary</a>. 
+            Applications using additional, user-dependent vocabularies of their own are encouraged to 
+            create their own formal vocabularies, and corresponding JSON-LD context definitions, 
+            to ensure the interoperability of the credentials they produce and/or consume.
         </p>
         <p>
-            There may be cases, however, when the creation of such formal vocabularies and context definitions is too much of a burden, such as for closed applications, experiments, or examples. The <a href="https://www.w3.org/TR/vc-data-model-2.0/#base-context">base context</a> has been constructed in such a way that issuers may use new, hitherto undefined terms without further ado, and these terms will be mapped onto URLs of the form <code>https://www.w3.org/ns/credentials/issuer-dependent#TERM</code>.
+            Not all users desire to define JSON-LD vocabularies or to make use of them with W3C Verifiable Credentials.
```
OR13, line 80: I changed this, because the previous content seems to be dismissive  / assume the intention of the issuer / holder.

I think we should assume their intention is to use the features we have built, which includes default expansion.

vocab/credentials/v2/issuer-dependent.html
```diff
@@ -66,15 +68,58 @@
...
+              return '&#'+c.charCodeAt(0)+';';
+          });
+      }
+      (async ()=>{
+        // handle issuer / hodler defined terms in fragment
+        if (window.location.hash.length){
+          const [_, term] = window.location.hash.split('#')
+          let el = document.getElementById('dynamic-term')
+          el.id = term;
+          const encodedTerm = htmlEncode(term);
```
OR13, line 102: some fear of reflected XSS here... anyone want to correct this?

vocab/credentials/v2/issuer-dependent.html
```diff
@@ -66,15 +68,58 @@
...
+            Applications using additional, user-dependent vocabularies of their own are encouraged to 
+            create their own formal vocabularies, and corresponding JSON-LD context definitions, 
+            to ensure the interoperability of the credentials they produce and/or consume.
         </p>
         <p>
-            There may be cases, however, when the creation of such formal vocabularies and context definitions is too much of a burden, such as for closed applications, experiments, or examples. The <a href="https://www.w3.org/TR/vc-data-model-2.0/#base-context">base context</a> has been constructed in such a way that issuers may use new, hitherto undefined terms without further ado, and these terms will be mapped onto URLs of the form <code>https://www.w3.org/ns/credentials/issuer-dependent#TERM</code>.
+            Not all users desire to define JSON-LD vocabularies or to make use of them with W3C Verifiable Credentials. 
+            The <a href="https://www.w3.org/TR/vc-data-model-2.0/#base-context">base context</a> has been constructed 
+            in such a way that issuers may use new, hitherto undefined terms without further ado,
```
TallTed, line 82:
```suggestion
            in such a way that issuers may use new, as-yet undefined terms without additional effort, 
```

OR13: When I read this, I knew I could expect a comment from you! lol.

vocab/credentials/v2/issuer-dependent.html
msporny, line 80:
```suggestion
            Not all document authors desire to define JSON-LD vocabularies or to make use of them with W3C Verifiable Credentials. 
```

An alternative here would be "implementers".

"users" tend to be construed as "people that use the technology, in a consumption capacity"... who are different than people that author content on the Web, who are different than people that implement software.

vocab/credentials/v2/issuer-dependent.html
```diff
@@ -66,15 +68,58 @@
...
+            The <a href="https://www.w3.org/TR/vc-data-model-2.0/#base-context">base context</a> of a Verifiable Credential, 
+            as defined in [vc-data-model], provides a mapping from the standard terms used in credentials to the vocabulary 
+            items defined by the Verifiable Credentials Working Group through different vocabularies, 
+            such as the <a href="https://www.w3.org/2018/credentials/">Verifiable Credentials Vocabulary</a>. 
+            Applications using additional, user-dependent vocabularies of their own are encouraged to
```
msporny, line 75:
```suggestion
            Applications using additional, issuer-dependent vocabularies of their own are encouraged to 
```

vocab/credentials/v2/issuer-dependent.html
msporny: -1, we shouldn't assume intention -- we don't know what it is. This landing place is specifically here because we have no idea why they're not defining their terms... all we know is that they didn't do it and so no one should be surprised when there is a term conflict in a large ecosystem.

msporny: We should strike this sentence entirely, we can't assume intent here.

vocab/credentials/v2/issuer-dependent.html
```diff
@@ -66,15 +68,58 @@
...
     </section>
-
+    <script>
+      function htmlEncode(str){
+          return String(str).replace(/[^\w. ]/gi, function(c){
+              return '&#'+c.charCodeAt(0)+';';
+          });
+      }
+      (async ()=>{
+        // handle issuer / hodler defined terms in fragment
```
msporny, line 97:
```suggestion
        // handle issuer / holder defined terms in fragment
```

vocab/credentials/v2/issuer-dependent.html
```diff
@@ -66,15 +68,58 @@
...
+</p>
+<p>
+Please consider contacting the <a href="https://www.w3.org/2018/credentials/#issuer">issuer</a> 
+or <a href="https://www.w3.org/2018/credentials/#holder">holder</a> if you believe this is an error.
+</p>
+<p>You may also consider reviewing:</p>
+<ul>
+  <li><a href="https://www.w3.org/TR/json-ld11/#terms">www.w3.org/TR/json-ld11/#terms</a></li>
+  <li><a href="https://www.w3.org/2018/credentials">www.w3.org/2018/credentials</a></li>
+  <li><a href="https://github.com/w3c/vc-specs-dir">w3c/vc-specs-dir</a></li>
```
msporny, line 117:
```suggestion
  <li><a href="https://www.w3.org/TR/json-ld11/#terms">JSON-LD v1.1: Terms</a></li>
  <li><a href="https://www.w3.org/2018/credentials">The Verifiable Credential Vocabulary</a></li>
  <li><a href="https://github.com/w3c/vc-specs-dir">The Verifiable Credentials Specification Directory</a></li>
```

vocab/credentials/v2/issuer-dependent.html
```diff
@@ -66,15 +68,58 @@
...
+</ul>
+<p>You might also consider searching for the term in a search engine, possibly with "+jsonld" in the query.</p>
```
msporny, line 119:
```suggestion
```

It's not clear what would result from that suggestion either now, or over time. Don't think the suggestion is useful beyond the advice given above.

vocab/credentials/v2/issuer-dependent.html
msporny: The `htmlEncode` function you provided does the wrong thing wrt. the following test values: `foo-bar-1`, `foo bar`, `f@@bar`, and so on. 

You could try `encodeURIComponent` instead, which still fails for stuff like `f@@bar`, but works in expected ways for the other two test cases, and `<script>` injection.

msporny (changes requested):
+1, in general to the PR.

The following items need to be addressed before merging:

* Do not presume intent; terms can be undefined for a variety of reasons. We should non-exhaustively list some of the reasons and then leave it at that. The current PR presumes intent, which we can't do on terms that are undefined.
* Fix the script injection mechanism.
* Be more accurate when it comes to identifying the entity that is using this feature.

vocab/credentials/v2/issuer-dependent.html
TallTed:
```suggestion
        // handle issuer- or holder-defined terms in fragment
```

vocab/credentials/v2/issuer-dependent.html
OR13: Let us not assume that the term will ever be defined or that the issuer made a mistake then.

OR13:
@msporny @TallTed thanks for the feedback.

2 items remain

- https://github.com/w3c/vc-data-model/pull/1083#discussion_r1163394975 (I am in favor of more neutral language than what was there before, but not attached to this specific language).

- the reflected XSS stuff, I feel it might be better to just make the advisement permanent, and not reflect.

vocab/credentials/v2/issuer-dependent.html
```diff
@@ -66,15 +68,57 @@
...
+            The <a href="https://www.w3.org/TR/vc-data-model-2.0/#base-context">base context</a> of a Verifiable Credential, 
+            as defined in [vc-data-model], provides a mapping from the standard terms used in credentials to the vocabulary 
+            items defined by the Verifiable Credentials Working Group through different vocabularies, 
+            such as the <a href="https://www.w3.org/2018/credentials/">Verifiable Credentials Vocabulary</a>. 
+            Applications using additional, issuer-dependent vocabularies of their own are encouraged to
```
dlongley, line 75: This is a typo, right?

```suggestion
            Applications using additional, issuer-independent vocabularies of their own are encouraged to 
```

vocab/credentials/v2/issuer-dependent.html
OR13: Outdating, resolving, please re review

OR13:
@msporny I have attempted to address all feedback, please let me know if you have additional change requests.

vocab/credentials/v2/issuer-dependent.html
```diff
@@ -66,15 +68,57 @@
...
+            The <a href="https://www.w3.org/TR/vc-data-model-2.0/#base-context">base context</a> of a Verifiable Credential, 
+            as defined in [vc-data-model], provides a mapping from the standard terms used in credentials to the vocabulary 
+            items defined by the Verifiable Credentials Working Group through different vocabularies, 
+            such as the <a href="https://www.w3.org/2018/credentials/">Verifiable Credentials Vocabulary</a>. 
+            Applications using additional, issuer-dependent vocabularies of their own are encouraged to 
+            create their own formal vocabularies, and corresponding JSON-LD context definitions, 
+            to ensure the interoperability of the credentials they produce and/or consume.
```
dlongley, line 77: We should highlight an important reason for using issuer-independent vocabularies and contexts here:

```suggestion
            to ensure the interoperability of the credentials they produce and/or consume. Using
            issuer-independent vocabularies reduces the burden on consumers of verifiable
            credentials, who do not need to update software when different issuers reuse the
            same shared vocabularies.
```

dlongley (changes requested):

vocab/credentials/v2/issuer-dependent.html
```diff
@@ -66,14 +68,38 @@
...
         </p>
         <p>
-            There may be cases, however, when the creation of such formal vocabularies and context definitions is too much of a burden, such as for closed applications, experiments, or examples. The <a href="https://www.w3.org/TR/vc-data-model-2.0/#base-context">base context</a> has been constructed in such a way that issuers may use new, hitherto undefined terms without further ado, and these terms will be mapped onto URLs of the form <code>https://www.w3.org/ns/credentials/issuer-dependent#TERM</code>.
+            Not all document authors desire to define JSON-LD vocabularies or to make use of them with W3C Verifiable Credentials. 
+            The <a href="https://www.w3.org/TR/vc-data-model-2.0/#base-context">base context</a> has been constructed 
+            in such a way that issuers may use new, as-yet undefined terms without additional effort, 
+            and these terms will be mapped onto URLs of the form <code>https://www.w3.org/ns/credentials/issuer-dependent#TERM</code>.
```
dlongley, line 83:
```suggestion
            and these terms will be mapped onto URLs of the form <code>https://www.w3.org/ns/credentials/issuer-dependent#TERM</code>.
            Authors that use this feature should be aware that consumers
            are expected to interpret the meaning of their terms relative to
            the issuer.

**Comment by iherman** [comment]
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-05-03-vcwg#section3-3) on 2023-05-03

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 3.3. Add issuer dependent term handling (pr vc-data-model#1083)

_See github pull request [vc-data-model#1083](https://github.com/w3c/vc-data-model/pull/1083)._




**Brent Zundel:** next up is PR 1083.  

**Orie Steele:** default vocab expansion in v2 context. it is there to make v2 context usage easier. this PR is about vocab expansion.  
… we are trying to give good guidance.  
… cleaning up PR.  

**Brent Zundel:** feels like this PR is making progress. what are next steps?  

**Orie Steele:** additional reviews would be helpful.  
… want to remove JS from page.  
… to prevent xss.  
… no json-ld IRIs on the page.  

</details>

File: vocab/credentials/v2/issuer-dependent.html (hunk 10 above)
Comment (dlongley) [reply vocab/credentials/v2/issuer-dependent.html:83]: I'm not sure where we want to indicate that the above can pose challenges, in the three party model, for digital wallet providers. They have to keep up with a list of issuers for which they understand the terms -- which will inevitably be kept small, containing only the largest VC issuers.

File: vocab/credentials/v2/issuer-dependent.html (hunk 9 above)
Comment (OR13) [reply vocab/credentials/v2/issuer-dependent.html:77]: I agree with the intention, but not this specific text.

It seems if would be better to no mention "updating software" here.

Perhaps a citation on the using widely adopted vocabularies such as schema.org would be better.

File: vocab/credentials/v2/issuer-dependent.html (hunk 10 above)
Comment (OR13) [reply vocab/credentials/v2/issuer-dependent.html:83]: I don't this we need to say more here, perhaps you can provide a reference to a standard that would be helpful for the reader to review regarding this? 

Like a section of another W3C TR? 

If this is a real issue, it must have been addressed in previous work, and we should be able to refer readers to a best practice.

File: vocab/credentials/v2/issuer-dependent.html (hunk 9 above)
Comment (dlongley) [reply vocab/credentials/v2/issuer-dependent.html:77]: I think it's important to highlight the burden on consumers who will have to make adjustments for every different issuer when shared (issuer-independent) vocabularies are not used. If the above language doesn't work for you -- could you suggest something else that communicates this?

File: vocab/credentials/v2/issuer-dependent.html [hunk 11]
```diff
@@ -66,14 +68,38 @@
...
+            Applications using additional, issuer-independent vocabularies of their own are encouraged to 
+            create their own formal vocabularies, and corresponding JSON-LD context definitions, 
+            to ensure the interoperability of the credentials they produce and/or consume.
         </p>
         <p>
-            There may be cases, however, when the creation of such formal vocabularies and context definitions is too much of a burden, such as for closed applications, experiments, or examples. The <a href="https://www.w3.org/TR/vc-data-model-2.0/#base-context">base context</a> has been constructed in such a way that issuers may use new, hitherto undefined terms without further ado, and these terms will be mapped onto URLs of the form <code>https://www.w3.org/ns/credentials/issuer-dependent#TERM</code>.
+            Not all document authors desire to define JSON-LD vocabularies or to make use of them with W3C Verifiable Credentials. 
+            The <a href="https://www.w3.org/TR/vc-data-model-2.0/#base-context">base context</a> has been constructed 
+            in such a way that issuers may use new, as-yet undefined terms without additional effort,
```
msporny, line 82:
```suggestion
            in such a way that issuers might use new, as-yet undefined terms without additional effort, 
```

vocab/credentials/v2/issuer-dependent.html
dlongley: Explaining why there are two different options (why one choice might be made over the other) is important, especially when the lead in is: "Not all document authors desire to have to do X." I would expect readers to think "Oh, why would I do extra stuff if I don't have to?" The reason is that they are shifting the burden onto consumers and helping create market failures via concentrated power. We should at least say something about the impact on consumers if we aren't willing to be clear about the ultimate outcomes that are made more likely -- so readers understand why they would reuse existing vocabularies or collaborate to create new shared ones.

msporny: >  perhaps you can provide a reference to a standard that would be helpful for the reader to review regarding this? Like a section of another W3C TR?

Hmm, well, one of the design goals for VCs was exactly to avoid this problem. The base issuer-dependent vocabulary re-introduces it to a certain degree. So, we created a spec to eliminate a problem that we are re-introducing in v2.0 (that you don't know what the issuer meant when they used a particular term, might never know, and you're guaranteed to conflict with others doing the same).

We should say something about the dangers of depending on issuer-dependent terms... the dangers were so bad that we had to create a new global standard to avoid the problem (Verifiable Credentials). We should speak to the danger if this default is used w/o thinking.

msporny (changes requested):
We should speak to the downsides of depending on this vocabulary.

vocab/credentials/v2/issuer-dependent.html
dlongley: @OR13,

> I don't this we need to say more here, perhaps you can provide a reference to a standard that would be helpful for the reader to review regarding this?
> 
> Like a section of another W3C TR?
> 
> If this is a real issue, it must have been addressed in previous work, and we should be able to refer readers to a best practice.

I should also say that following this argument prevents any new work from mentioning the problem -- which means no work will mention the problem. So -- we should mention the problem. Mentioning it will also mean people who have a similar conversation about this problem in the future on some other spec will have something in TR space to link to.

OR13: I maybe misunderstood your original point then, I thought you were asserting there was a documented problem, that is caused by not using "issuer independent" contexts.

I was hoping to refer to that documentation be reference.

Are you saying we will be the first WG at W3C to describe this problem?

dlongley: @OR13,

> Are you saying we will be the first WG at W3C to describe this problem?

In context with the work we're doing here in a way people would readily understand -- I don't know the answer to your question. I know VCs were literally invented to help solve that problem. I'm happy for you to include a reference to another spec in TR space if you know of one, either in addition to the language I proposed, or if you find something that's very duplicative, we can just reference it.

You may be able to point to this generally as an informative reference around avoiding centralization: https://www.ietf.org/archive/id/draft-nottingham-avoiding-internet-centralization-09.html.

vocab/credentials/v2/issuer-dependent.html
OR13: > I think it's important to highlight the burden on consumers who will have to make adjustments for every different issuer when shared (issuer-independent) vocabularies are not used. 

I don't think I understand what you are saying here... I can imagine `consumers` never seeing an IRI for any of the terms the core data model uses, since the base media type is JSON-LD, and the IRIs only show up to consumers of RDF.

> If the above language doesn't work for you -- could you suggest something else that communicates this?

I can try:

If you plan on consuming Verifiable Credentials as RDF, for example, after converting from `application/vc+ld+json` to `application/n-quads`, please be aware that the core data model is JSON-LD and that the issuer may not have intended for `application/n-quads` to be consumed when they added integrity protection to their Verifiable Credential.

This can cause the predicates and types that you process to be potentially ambiguous, in cases where multiple issuers used the same terms in JSON-LD, without defining them, in `issuer independent` contexts.

OR13:
@msporny does this text work for you? https://github.com/w3c/vc-data-model/pull/1083#discussion_r1185498747

vocab/credentials/v2/issuer-dependent.html
dlongley: @OR13,

> I don't think I understand what you are saying here... I can imagine consumers never seeing an IRI for any of the terms the core data model uses, since the base media type is JSON-LD, and the IRIs only show up to consumers of RDF.

This has little to do with IRIs and much more to do with "context". If a VC does not use its own `@context`, then the only "context" available to consumers is who the issuer is. That means the terms, whatever they may be, are issuer-dependent.

It means consumers must understand N "issuers + their credential types". This is a much higher burden than understanding N credential types (via `@context` values), regardless of the issuer.

When you write digital wallet software that is supposed to help users make decisions ... it must check a well-known list of issuers and their VC types to determine what the meaning of a VC is. This is a different class of problem than having to check a well-known list of JSON-LD contexts. Every new JSON-LD context that a wallet recognizes onboards any issuer that uses it. Every VC that uses issuer-independent terms requires the wallet to understand who that issuer is and what they meant by that specific type of credential. Additionally, they must track down the documentation for understanding the terms in issuer-priorietary ways.

The latter case increases the burden considerably for digital wallet providers. One can only expect them to put in the effort for the largest VC issuers -- which will lead to market concentration problems. People who care about this should be aware of this -- so that they might help avoid it. Of course, market players that want to do it on purpose will not be dissuaded, but they are not the target audience of the language here.

So, this has nothing to do with "if you plan on using VCs as RDF" or "n-quads", etc. This has to do with enabling software to determine types based on JSON-LD `@context` -- including writing software that uses VCs as JSON.

TallTed: > Every VC that uses issuer-independent terms requires the wallet to understand who that issuer is

You meant `issuer-dependent` here, I think?

dlongley: @TallTed, yes -- thanks I've applied the edit.

msporny (approved):
I'm good w/ this PR as long as @dlongley and @TallTed's concerns are addressed. It's an improvement on what exists today. We might want to say more about issuer-dependent terms in the implementation guide, or perhaps we should say it here, where most people will end up if they "follow their nose" via the term URLs.

mprorock (approved):
We can build on this after merge in my mind

dlongley:
@mprorock,

> We can build on this after merge in my mind

My change request is blocking. I'm ok with if we want to add my suggestions and merge those -- and then build from there with additional tweaks, however.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-05-10-vcwg#section1-4) on 2023-05-10

- no resolutions were taken

#### 1.4. Add issuer dependent term handling (pr vc-data-model#1083)

_See github pull request [vc-data-model#1083](https://github.com/w3c/vc-data-model/pull/1083)._




**Manu Sporny:** 1083 manu is good with if Tallted's suggestions are addressed. Move on to another PR.  
… 1100 and 1101.  

**Kevin Dean:** the discussion of these will merit more time to see if we can get through others.  

**Manu Sporny:** special topic call needed for 1100 and 1101.  

</details>

msporny:
@OR13, thoughts on how to process this PR? Are you going to apply some variation of @dlongley and @TallTed's suggestions, or should I mark this as DO NOT MERGE and schedule call time to discuss?

OR13:
Closing the PR, does not look like we are going to gain consensus.

PR #1087, opened 2023-04-16
Remove section on Disputes; no one has implemented it.

msporny:
This PR removes the section on "Disputes" because no one has implemented it that I know of. IIRC, I originally wrote the section to address WG concern about "false claims" and what to do about them -- I vaguely remember discussing it at some depth with @David-Chadwick. I don't plan to move this section to the Implementer Guide because it's pure fiction.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1087.html" title="Last updated on Apr 27, 2023, 5:34 PM UTC (23f5c84)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1087/0abf1f2...23f5c84.html" title="Last updated on Apr 27, 2023, 5:34 PM UTC (23f5c84)">Diff</a>


Sakurann (approved):

David-Chadwick:
@msporny I agree that nobody that we know of has implemented this feature. So we should remove it from the DM. And its not really an Implementation issue today either.
Nevertheless the concept of erroneous VCs that either the subject or a third party will want to dispute, will become an increasingly important issue as VCs become more prevalent. So I guess folks can refer to DM1.1 in the future to see what our suggestions were for handling it.

OR13 (approved):
This is the kind of PR I love to see.

TallTed:
I would strongly prefer that this section be put AT RISK until CR's broader review and call for implementations by the broader community, who may well not be implementing (or mentioning their implementation) until the spec is closer to PR/TR, and therefore not deleted unless there are still not two independent implementations when the doc is being prepped for the CR→PR transition.

At such point as the section *is* deleted from the main doc, I strongly suggest that we put it into a NOTE (which doesn't need much more content than the section cut from index.html), as I cannot imagine a future which doesn't include a Dispute mechanism, and this should be a good start thereon.

TallTed (changes requested):
I would strongly prefer that this section be put AT RISK until CR's broader review and call for implementations by the broader community, who may well not be implementing (or mentioning their implementation) until the spec is closer to PR/TR, and therefore not deleted unless there are still not two independent implementations when the doc is being prepped for the CR→PR transition.

At such point as the section is deleted from the main doc, I strongly suggest that we put it into a NOTE (which doesn't need much more content than the section cut from index.html), as I cannot imagine a future which doesn't include a Dispute mechanism, and this should be a good start thereon.

brentzundel (approved):

dlongley (approved):
I think @TallTed's comments here are important to consider, though I also don't think significant effort has been put into the disputes modeling nor am I aware of anyone who is interested in doing so at this time. So what we see there may not even be the best thing for anyone to implement even as an at-risk feature.

msporny:
@David-Chadwick wrote:
> I guess folks can refer to DM1.1 in the future to see what our suggestions were for handling it.

Yes, that is the plan.

@TallTed wrote:
> I would strongly prefer that this section be put AT RISK until CR's broader review and call for implementations by the broader community

We have gone through CR phases (v1.0 and v1.1) that resulted in zero implementations for this section. There is no normative language in this section to implement, so there won't be a test suite as a result of it for v2. IOW, "At Risk" is for sections that have normative content, this section is informative.

> At such point as the section is deleted from the main doc, I strongly suggest that we put it into a NOTE (which doesn't need much more content than the section cut from index.html), as I cannot imagine a future which doesn't include a Dispute mechanism, and this should be a good start thereon.

Remember that we have a v1.0 and a v1.1 REC that contains this language already. It's already been published and has a stable reference. If we need to refer to this work, we can refer to those refs: 

https://www.w3.org/TR/2022/REC-vc-data-model-20220303/#disputes

At this point, @TallTed and @David-Chadwick, I believe we've addressed your concerns and I'd like to merge this PR.

However, I'm requesting a re-review from both of you to ensure that you agree with the merge.

David-Chadwick (approved):

TallTed:
> We have gone through CR phases (v1.0 and v1.1) that resulted in zero implementations for this section.

@msporny — OK, you're right, "at risk" is not appropriate.  However, neither is full deletion.

I suggest the section be reduced to something like "The Dispute feature has been dropped in v2 due to lack of implementation. Interested implementers may refer to prior versions [v1] and [v1.1] and contact the VCWG to discuss implementation of such an extension for future versions."

David-Chadwick:
@TallTed I think what you propose is a good compromise

msporny:
> @TallTed I think what you propose is a good compromise

Yes, I could live with that as well... would anyone that has reviewed this PR object to the change that @TallTed is suggesting?

awoie (approved):
LGTM. I agree with @TallTed that we should be careful in what we remove but as @msporny pointed out we have gone through CR phases (v1.0 and v1.1) that resulted in zero implementations. IMO, it is safe to remove this from the spec.

decentralgabe (approved):
🔥

brentzundel:
I assume the published VCDM 2.0 will include a summary of changes from previous versions, and such a summary will already indicate that properties such as this have been removed.
I agree that we shouldn't just remove it and say nothing, but would prefer to use a changelog over keeping the section with a comment about how it has been removed.

TallTed:
> I agree that we shouldn't just remove it and say nothing, but would prefer to use a changelog over keeping the section with a comment about how it has been removed.

I can live with such an explicit statement in the "changes since 1.1" section.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-04-26-vcwg#section2-1) on 2023-04-26

- no resolutions were taken

#### 2.1. Remove section on Disputes; no one has implemented it. (pr vc-data-model#1087)

_See github pull request [vc-data-model#1087](https://github.com/w3c/vc-data-model/pull/1087)._




**Manu Sporny:** initial PR was to remove section entirely, after pushback new suggestion is to leave title with link to specification.  

> *Orie Steele:* +1 Brent.

**Brent Zundel:** we can remove as long as changelog reflects.  

**Ted Thibodeau Jr.:** concerned that we are removing for lack of implementation, but there should have been independent implementations.  

**Manu Sporny:** that was all non-normative.  
… will change PR to remove section and put that in the changelog - will merge if no objections.  

> *Chris Abernethy:* (no objections were voiced).

</details>

msporny:
@TallTed @David-Chadwick -- I have added an entry in the "Revision History" section that states that the "Disputes" section was removed, with a link to the v1.1 section on Disputes. This will make finding the old section easier for those that are curious to see what was there in v1.0 and v1.1.

Fixed in: 23f5c840e2129dd66b2c5ed76633be3f65023edd

msporny:

PR #1088, opened 2023-04-16
Remove JSON section in lieu of `application/vc+ld+json`.

msporny:
Based on the resolution of the WG to use one base media type (`application/vc+ld+json`), this PR removes the old "JSON" section in the spec, which was always of very questionable value. The prose has been updated to specify that the base media type is `application/vc+ld+json`, and other syntaxes can be used as long as they can map to the base media type.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1088.html" title="Last updated on Apr 22, 2023, 4:30 PM UTC (3ef0175)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1088/5ae6a29...3ef0175.html" title="Last updated on Apr 22, 2023, 4:30 PM UTC (3ef0175)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3261,11 +3261,11 @@ The data model as described in Sections
 <a href="#advanced-concepts"></a> is the canonical structural representation of
 a <a>verifiable credential</a> or <a>verifiable presentation</a>. All
 serializations are representations of that data model in a specific format. This
-section specifies how the data model is realized in JSON-LD for the base
-media type (`application/vc+ld+json`).
-Although syntactic mappings are provided for JSON-LD,
+section specifies how the data model is realized in JSON-LD for
+`application/vc+ld+json`, the base media type for Verifiable Credentials.
+Although syntactic mappings are only provided for JSON-LD,
 applications and services can use any other data representation syntax (such as
-XML, YAML, or CBOR) that is capable of being mapped back to the base media type.
+XML, YAML, or CBOR) that is capable of being mapped back to `application/vc+ld+json`.
 As the <a>verification</a> and <a>validation</a> requirements are defined in terms
 of the data model, all serialization syntaxes have to be deterministically
 translated to the data model for processing, <a>validation</a>, or comparison.
--- a/index.html
+++ b/index.html
@@ -3261,15 +3261,14 @@ The data model as described in Sections
 <a href="#advanced-concepts"></a> is the canonical structural representation of
 a <a>verifiable credential</a> or <a>verifiable presentation</a>. All
 serializations are representations of that data model in a specific format. This
-section specifies how the data model is realized in JSON-LD and plain JSON.
-Although syntactic mappings are provided for only these two syntaxes,
+section specifies how the data model is realized in JSON-LD for the base
+media type (`application/vc+ld+json`).
+Although syntactic mappings are provided for JSON-LD,
 applications and services can use any other data representation syntax (such as
-XML, YAML, or CBOR) that is capable of expressing the data model. As the
-<a>verification</a> and <a>validation</a> requirements are defined in terms of
-the data model, all serialization syntaxes have to be deterministically
+XML, YAML, or CBOR) that is capable of being mapped back to the base media type.
+As the <a>verification</a> and <a>validation</a> requirements are defined in terms
+of the data model, all serialization syntaxes have to be deterministically
 translated to the data model for processing, <a>validation</a>, or comparison.
-This specification makes no requirements for support of any specific
-serialization format.
       </p>
 
       <p>
@@ -3299,47 +3298,6 @@ or an array of values.
       </p>
 
       <section>
-        <h3>JSON</h3>
-
-        <p>
-The data model, as described in Section <a href="#core-data-model"></a>, can be
-encoded in JavaScript Object Notation (JSON) [[!RFC8259]] by mapping property
-values to JSON types as follows:
-        </p>
-
-        <ul>
-          <li>
-Numeric values representable as IEEE754 SHOULD be represented as a Number type.
-          </li>
-          <li>
-Boolean values SHOULD be represented as a Boolean type.
-          </li>
-          <li>
-Sequence value SHOULD be represented as an Array type.
-          </li>
-          <li>
-Unordered sets of values SHOULD be represented as an Array type.
-          </li>
-          <li>
-Sets of <a>properties</a> SHOULD be represented as an Object type.
-          </li>
-          <li>
-Empty values SHOULD be represented as a null value.
-          </li>
-          <li>
-Other values MUST be represented as a String type.
-          </li>
-        </ul>
-
-        <p class="note">
-As the transformations listed herein have potentially incompatible
-interpretations, additional profiling of the JSON format is required to provide
-a deterministic transformation to the data model.
-        </p>
-
-      </section>
-
-      <section>
         <h3>JSON-LD</h3>
 
         <p>
@@ -3354,9 +3312,8 @@ storage engines.
 
         <p>
 [[!JSON-LD]] is useful when extending the data model described in this
-specification. Instances of the data model are encoded in [[!JSON-LD]] in the
-same way they are encoded in JSON (Section <a href="#json"></a>), with the
-addition of the <code>@context</code> <a>property</a>. The
+specification. Instances of the data model are encoded in [[!JSON-LD]]
+and include the <code>@context</code> <a>property</a>. The
 <a href="https://www.w3.org/TR/json-ld/#the-context">JSON-LD context</a>
 is described in detail in the [[!JSON-LD]] specification and its use is
 elaborated on in Section <a href="#extensibility"></a>.
--- a/index.html
+++ b/index.html
@@ -3312,8 +3312,8 @@ storage engines.
 
         <p>
 [[!JSON-LD]] is useful when extending the data model described in this
-specification. Instances of the data model are encoded in [[!JSON-LD]]
-and include the <code>@context</code> <a>property</a>. The
+specification. Instances of the data model are encoded in JSON-LD compact
+form [[!JSON-LD]] and include the <code>@context</code> <a>property</a>. The
 <a href="https://www.w3.org/TR/json-ld/#the-context">JSON-LD context</a>
 is described in detail in the [[!JSON-LD]] specification and its use is
 elaborated on in Section <a href="#extensibility"></a>.


iherman (approved):

index.html
```diff
@@ -3257,15 +3257,14 @@ <h2>Syntaxes</h2>
 <a href="#advanced-concepts"></a> is the canonical structural representation of
 a <a>verifiable credential</a> or <a>verifiable presentation</a>. All
 serializations are representations of that data model in a specific format. This
-section specifies how the data model is realized in JSON-LD and plain JSON.
-Although syntactic mappings are provided for only these two syntaxes,
+section specifies how the data model is realized in JSON-LD for the base
```
OR13, line 3260: what does the word "base" do for us here? feels a bit confusing.

index.html
```diff
@@ -3257,15 +3257,14 @@ <h2>Syntaxes</h2>
...
+media type (`application/vc+ld+json`).
+Although syntactic mappings are provided for JSON-LD,
 applications and services can use any other data representation syntax (such as
-XML, YAML, or CBOR) that is capable of expressing the data model. As the
-<a>verification</a> and <a>validation</a> requirements are defined in terms of
-the data model, all serialization syntaxes have to be deterministically
+XML, YAML, or CBOR) that is capable of being mapped back to the base media type.
```
OR13, line 3264:
```suggestion
XML, YAML, or CBOR) that is capable of being mapped back to `application/vc+ld+json`.
```

index.html
```diff
@@ -3294,47 +3293,6 @@ <h2>Syntaxes</h2>
...
-          <li>
-Empty values SHOULD be represented as a null value.
-          </li>
-          <li>
-Other values MUST be represented as a String type.
-          </li>
-        </ul>
-
-        <p class="note">
-As the transformations listed herein have potentially incompatible
```
OR13, line 3331: felicia-when-she-left-the-village

index.html
```diff
@@ -3350,9 +3308,8 @@ <h3>JSON-LD</h3>
 
         <p>
 [[!JSON-LD]] is useful when extending the data model described in this
-specification. Instances of the data model are encoded in [[!JSON-LD]] in the
-same way they are encoded in JSON (Section <a href="#json"></a>), with the
-addition of the <code>@context</code> <a>property</a>. The
+specification. Instances of the data model are encoded in [[!JSON-LD]]
+and include the <code>@context</code> <a>property</a>. The
```
OR13, line 3312: Do we want to comment on compact JSON-LD here?

OR13 (approved):
Feels like we might make some minor wording changes, but overall this is great.

index.html
TallTed: Please translate for those of us who aren't as hip to the modern meme-lingo... (Really, please don't post things like this in technical discussions. They don't tend to advance the conversation, and they are often needlessly offensive.)

index.html
```diff
@@ -3257,15 +3257,14 @@ <h2>Syntaxes</h2>
 <a href="#advanced-concepts"></a> is the canonical structural representation of
 a <a>verifiable credential</a> or <a>verifiable presentation</a>. All
 serializations are representations of that data model in a specific format. This
-section specifies how the data model is realized in JSON-LD and plain JSON.
-Although syntactic mappings are provided for only these two syntaxes,
+section specifies how the data model is realized in JSON-LD for the base
+media type (`application/vc+ld+json`).
+Although syntactic mappings are provided for JSON-LD,
```
TallTed, line 3262:
```suggestion
section specifies how the data model is realized in JSON-LD for
`application/vc+ld+json`, the base media type for Verifiable Credentials.
Although syntactic mappings are only provided for JSON-LD,
```

index.html
TallTed: I am concerned about the lack of "losslessly" in these paragraphs, as that has been a key factor in our discussions.

index.html
TallTed: Probably.

TallTed (changes requested):

index.html
brentzundel: I think it reads the same without the word 'base'
I'd support removing.

brentzundel (approved):

dlongley (approved):
Thanks! LGTM and +1 to @TallTed's comments.

index.html
msporny: 'base' has been removed in 654b694fd41580ba7724d1593f5c1906de3180e5

index.html
msporny: Fixed in 3ef0175a4931bafc30dd73f893889e6f6ce25c8b.

msporny:

alenhorvat:
@msporny @OR13 this means VCDM v2 discards JWT-VC in a form
{
<jwt claims>
vc: <vc claims>
}
?

OR13:
See the w3c/vc-jwt spec.

alenhorvat:
Answer is: Yes, VCDM v2 support JWT-VC?

Sakurann:
W3C JWT-VC spec describes how to sign using JWS a payload that is expressed using W3C VCDM.

msporny:
> @msporny @OR13 this means VCDM v2 discards JWT-VC in a form

In the case that @alenhorvat is asking a more specific question, that is: Are v1.0 and v1.1 style VC-JWTs supported in v2.0, I believe the answer to that question is: *No*, VC-JWTs of the form you mention above are no longer supported by the new VC-JWT specification. Instead, the new VC-JWT specification places the VC in the payload directly with no transformations.

alenhorvat:
Thank you. Clear. 
To rephrase: 
q1: VCDM v2.0 recognises https://github.com/w3c/vc-jwt as a valid VC?

"_The data model as described in Sections [3. Core Data Model](https://www.w3.org/TR/vc-data-model-2.0/#core-data-model), [4. Basic Concepts](https://www.w3.org/TR/vc-data-model-2.0/#basic-concepts), and [5. Advanced Concepts](https://www.w3.org/TR/vc-data-model-2.0/#advanced-concepts) is the canonical structural representation of a [verifiable credential](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verifiable-credential) or [verifiable presentation](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verifiable-presentation). All serializations are representations of that data model in a specific format. This section specifies how the data model is realized in JSON-LD for application/vc+ld+json, the base media type for Verifiable Credentials. Although syntactic mappings are only provided for JSON-LD, applications and services can use any other data representation syntax (such as XML, YAML, or CBOR) that is capable of being mapped back to application/vc+ld+json. As the [verification](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verify) and [validation](https://www.w3.org/TR/vc-data-model-2.0/#dfn-credential-validation) requirements are defined in terms of the data model, all serialization syntaxes have to be deterministically translated to the data model for processing, [validation](https://www.w3.org/TR/vc-data-model-2.0/#dfn-credential-validation), or comparison."_

q2: Is my understanding, that VCs can be treated also as JSON (when -LD is not required by the use case) correct? Or one could define a VC in an XML format, for example, and it would be a valid VC according to the VCDM v2.0?

q3: If additional JWT claims are required (because of the signature), are those claims added on the top level?
e.g., iat, exp along with validFrom, validUntil?

msporny:
> q1: VCDM v2.0 recognises https://github.com/w3c/vc-jwt as a valid VC?

The WG is trying to be very careful with the language used around the question "What is a VC?". At the moment, the way to think of the VC-JWT specification is: "A mechanism for securing the Verifiable Credentials Core Data Model using JWTs as described in the VC-JWT specification."

> q2: Is my understanding, that VCs can be treated also as JSON (when -LD is not required by the use case) correct?

This is a point of contention in the WG. The original design of VCs allowed for VCs to be treated as JSON as long as you checked/did certain things. The WG has since said that, no, there is only one media type for VCs and that is `application/vc+ld+json` <-- that is the core data model serialization and it can be secured using a variety of mechanisms (VC-JWT, Data Integrity, etc.). It is that  finding that led to the removal of the JSON section from the specification.

In reality, many systems process the VC Data Model as JSON and don't invoke JSON-LD processing until later in the process. For example, an incoming VC might be checked using JSON Schema (JSON processing) before a Data Integrity proof is checked (and in some cases, JCS still counts as JSON processing). It is possible to hard code a JSON-only processing of incoming VCs if the application is specific enough. That said, that is up to implementers to decide when/how to do JSON vs. JSON-LD processing.

> Or one could define a VC in an XML format, for example, and it would be a valid VC according to the VCDM v2.0?

At this point in time, there is only ONE serialization format for the VC data model and that is `application/vc+ld+json`. That said, there are mechanisms that can cleanly round-trip from the VCDM -> some other format -> VCDM. For example, CBOR-LD is capable of taking `application/vc+ld+json`, converting it to CBOR, transmitting it, and the other end of the connection transforming the compressed binary encoding, losslessly, back into `application/vc+ld+json`. This is how the TruAge system transmits digitally signed Verifiable Credentials in QR Codes (as CBOR-LD).

> q3: If additional JWT claims are required (because of the signature), are those claims added on the top level? e.g., iat, exp along with validFrom, validUntil?

I don't know where the Editors are on this particular item. Perhaps @OR13, @mprorock, and @selfissued can weigh in on this one.

alenhorvat:
_It is possible to hard code a JSON-only processing of incoming VCs if the application is specific enough. That said, that is up to implementers to decide when/how to do JSON vs. JSON-LD processing._

Then VC can be JSON only and essentially an application/vc+json could be completely valid?

In our case, some UC process only JSON, others use JSON-LD (signatures are always JWS/JAdES, where JSON is compared against the JSON schema and no JCS or any other additional processing). Some (most) UC don't define @context for various reasons.

isomasa:
> The WG has since said that, no, there is only one media type for VCs and that is `application/vc+ld+json`

I think that the WG should bring the evidence and reasons of why this is beneficial for the ecosystem. I would also like to raise that LD is based on JSON, thus LD is the possible extension.

> At this point in time, there is only ONE serialization format for the VC data model and that is application/vc+ld+json.

Requiring JSON-LD to be THE serialisation format will remove all adoption. This second quote will force JSON-LD context to exist, or the attribute will drop during the serialisation. This is not healthy for adoption of VC. 

> That said, that is up to implementers to decide when/how to do JSON vs. JSON-LD processing.

This spec should be reworded/fixed if the recommendation is to ignore it. Most implementors will not have JSON-LD definitions, and if everyone will just use bogus @json, it still adds zero value.

So my single question is, what is the value for the ecosystem to force JSON-LD in? this can be handled in the https://www.w3.org/TR/vc-data-integrity/.

msporny:
> _It is possible to hard code a JSON-only processing of incoming VCs if the application is specific enough. That said, that is up to implementers to decide when/how to do JSON vs. JSON-LD processing._
> 
> Then VC can be JSON only and essentially an application/vc+json could be completely valid?

At this point in time, the answer is "No". There is no `application/vc+json` media type because the decentralized semantics of that media type is not defined, nor did the WG come to any sort of consensus of how we'd define that media type or the decentralized semantics around the media type. 

The WG has debated this endlessly at this point: https://github.com/w3c/vc-data-model/issues/947

This was the most contentious issue the WG has ever discussed and it took close to 9+ months to come to a resolution on just that point. The resolution was to trim the number of media types down to ONE media type, which is: `application/vc+ld+json`. You can use a single `"@context": ["https://www.w3.org/ns/credentials/v2"] (though, that is frowned upon) and not have to define any other `@context` value. All properties become "issuer defined", which is to say that what you have at that point is a bespoke point-to-point integration that requires coordination between the verifier and issuer. Again, not recommended that you take that route because it leads to interop/market failures.

> In our case, some UC process only JSON, others use JSON-LD (signatures are always JWS/JAdES, where JSON is compared against the JSON schema and no JCS or any other additional processing). Some (most) UC don't define @context for various reasons.

Like I mention above, you can do some variation of that, but it's not recommended by a subset of the WG. Other WG members will tell you that it's fine to not have any defined semantics, use JSON Schema to define the semantics, or do some other proprietary thing. The choice is yours, but you shouldn't expect broad interoperability with your solution if you take that path.

msporny:
> > The WG has since said that, no, there is only one media type for VCs and that is `application/vc+ld+json`
> 
> I think that the WG should bring the evidence and reasons of why this is beneficial for the ecosystem. I would also like to raise that LD is based on JSON, thus LD is the possible extension.

The reasons are included in this mega-thread, that took 9+ months to resolve: https://github.com/w3c/vc-data-model/issues/947

> Requiring JSON-LD to be THE serialisation format will remove all adoption. This second quote will force JSON-LD context to exist, or the attribute will drop during the serialisation. This is not healthy for adoption of VC.

The WG has debated these points for many years and has decided to go in the direction stated above (for all the reasons that were discussed in that issue thread above). 

We tried the alternate path that you are suggesting in the DID WG (multiple media types) and it resulted in a bad outcome as well. The vast majority of implementers implemented with the `@context` value specified.

> So my single question is, what is the value for the ecosystem to force JSON-LD in? this can be handled in the https://www.w3.org/TR/vc-data-integrity/.

The value is that there is ONE media type. The WG came to consensus on the point that having a single base data model syntax was beneficial to the ecosystem because optionality, when it comes to global standards, tends to fragment the ecosystem. Therefore, this is one place where we could reduce optionality  while meeting all of the use cases in front of the WG.

alenhorvat:
VC semantics is great and is well defined. Assumption is that use cases agree to follow the semantics as defined in the specifications.

However, if VC is not processed as LD, application/vc+json is more a appropriate format. Media type is used by the use cases since they need to signal how the content should be processed.
Since many use cases do not have a need for the -ld capabilities, wouldn't it be easier to achieve global interoperability also for these use cases by defining the type correctly in one specification than leave it to the market to do whatever they want.

Question is about vc+json, not vc+jwt (where semantic interop issues appear)

alenhorvat:
I agree with you and everything you wrote. I fully understand the interop and scalability issues. Proposal/question is very simple:
If VC has the format and structure as defined in the standard for application/vc+json+ld (with all the claims, etc) but the context is not processed, would it make sense to use application/vc+json

If answer is no, this simply means we need to define a profile for it (definition of the profile would essentially be the statement above). I'm arguing that it is more beneficial to have it in the specification.

To clarify: the proposal is
- to not make the @context optional
- to not change any of the defined claims (e.g., replace them by JWT claims)
and the goal is to NOT reopen any of the painful discussions.

People will use and process VCs as JSON for various reasons. Suggestion here is to take what's been defined and the WG agreed upon and just provide guidelines or show that if UC is not defining @context for the claims and types they introduce (e.g., the UC does not require global interoperability) and they are processing it as JSON, use application/vc+json

alenhorvat:
We're not trying to solve the semantics problem. We're just trying to find a way how to express what it seems we agree: VC (as defined today) can be treated and processed as JSON.
Question is: how to express this in a machine readable way?

mime type? any other way?

We don't want to use custom way to express that in some cases VC should be processed as LD and in other cases as JSON.

isomasa:
Thank you @msporny for the time and devotion, much appreciated. 

I can see from the answers that JSON-LD as requirement is not really a requirement at all. All deterministic, bi-directional, and lossless serialisations and transformations can be done with default context of "https://www.w3.org/ns/credentials/v2". Without extending this, it will work exactly like JSON, and supports both, JSON-LD and JSON processors.

Media type of `application/ld+vc+json` in this way is also equal to `application/vc+json`, but the latter one should not be used to have just one media type for the VC. There's just the extra boilerplate of context with the default vocab.

mtaimela:
Thank you @TallTed.

I fully agree that having 100% semantic web, where all data and transactions are machine readable is where we want to be. But I also have to embrace the fact of legacy and allow the natural adoption of semantics. This is not an easy task and I do not want this to hinder other achievements like SSI and give the control of data to the owners of it.

I wrote a small summary, for everyone else in the world reading this thread. Could it be possible to respond if summary the below is how it was intended.

VCWG decided:
- @context is required (MUST) in the base media type, other media types MAY choose to include @context.
- Bi-directional transformation MUST preserve @context
- Transformation rules MUST be defined, but not necessarily by this WG.

Specification states:
- Although syntactic mappings are only provided for JSON-LD, applications and services can use any other data representation syntax (such as XML, YAML, or CBOR) that is capable of being mapped back to application/vc+ld+json
- default JSON-LD @context is https://www.w3.org/ns/credentials/v2, .... for the base media type of application/vc+ld+json.

Thread agrees:
JSON can be converted into "Issuer defined JSON-LD", by applying the default @context. This is frowned upon as it doesn't improve semantic web.

confirmation needed:
VCDM 2.0 compatible VCs are "any data representation syntax" when the serialisation towards JSON-LD is deterministic, bi-directional, and lossless, as long as it may be converted into JSON-LD with the original @context (custom or default). The serialisation must be documented.

msporny:
@mtaimela wrote:
> I fully agree that having 100% semantic web, where all data and transactions are machine readable is where we want to be. But I also have to embrace the fact of legacy and allow the natural adoption of semantics. This is not an easy task and I do not want this to hinder other achievements like SSI and give the control of data to the owners of it.

JSON-LD is designed to use as much of it as is useful to you. The VC Data Model attempts to follow the same approach, with a few important nuances.

> I wrote a small summary, for everyone else in the world reading this thread. Could it be possible to respond if summary the below is how it was intended.

Yes, answers below:

> * @context is required (MUST) in the base media type, other media types MAY choose to include @context.

Yes, with the warning that "other media types" is probably not going to be defined very clearly in this iteration of the VCWG (as we are running out of time in our charter and will enter complete and utter feature freeze in September). We are currently dealing with too many other things, and the "transformations" language has proven to be contentious (thus it is unlikely that we'll be able to achieve consensus on EXACTLY what that means... though I can try to provide insight into where the VCWG might be heading in future iterations).

> * Bi-directional transformation MUST preserve @context

Yes (but I expect that we will not see that text appear in v2.0, possibly during v3.0 once the market has settled down a bit on the transformations stuff). 

I will note that CBOR-LD is an example of something that exists today that transforms to a completely different binary data stream, but is able to round-trip and thus preserve `@context`. So, the VCWG is well aware that this is possible for organizations that act in good faith. It's the organizations that might not act in good faith that are the concern, which means we would have to be very careful about what language we write as writing the wrong language will have organizations claiming that they are "100% compliant W3C VCs", when they are not (or are doing the bare minimum, like using "issuer defined terms" without defining an `@context` which can lead to interop failures at scale).

> * Transformation rules MUST be defined, but not necessarily by this WG.

Yes, but again, the VCWG will almost certainly not include this text in the current specification and wait for market feedback to drive the sort of language that will lead to a healthy, interoperable ecosystem.

> * Although syntactic mappings are only provided for JSON-LD, applications and services can use any other data representation syntax (such as XML, YAML, or CBOR) that is capable of being mapped back to application/vc+ld+json
> * default JSON-LD @context is https://www.w3.org/ns/credentials/v2, .... for the base media type of application/vc+ld+json.
> 
> Thread agrees: JSON can be converted into "Issuer defined JSON-LD", by applying the default @context. This is frowned upon as it doesn't improve semantic web.

Yes, though this isn't about the "semantic web". It's an information modelling problem that exists in any system with locally defined terms that were never meant to be used globally. Think of every proprietary JSON HTTP API out there now, the data objects that they publish largely only work at one domain. 

If something is issuer-defined, then you have to have an out of band communication with the issuer to understand the proper usage of their terms, which will almost certainly conflict with other issuer-defined terms. If one issuer says "name" means "the name of your website", and another issuer says "name" means your full name, then the issuers mean very different things and you can't count on interoperability between those issuers. If you look at this problem at scale, it would require you to have to either 1) contact every issuer out of band to understand how they're using their terms, or 2) centralize the market to a very small set of issuers that have an agreement on which terms they are using (which would be a market competition failure).

> confirmation needed: VCDM 2.0 compatible VCs are "any data representation syntax" when the serialisation towards JSON-LD is deterministic, bi-directional, and lossless, as long as it may be converted into JSON-LD with the original @context (custom or default). The serialisation must be documented.

Yes, close, and here's where you're not going to get much more guidance from the VCWG. The following is my interpretation of what's probably going to happen in time:

*digital credentials* that are compatible with the W3C Verifiable Credential Data Model are credentials where they can be transformed to the VCDM and where the interpretation of those transformed VCs remains the same as their original interpretation. That is, one could say that their digital credential is VCDM 2.0 compatible if:

The digital credential can be unidirectionally transformed to the VCDM and can be clearly interpreted in the way it was originally intended to be interpreted (before it was transformed).

Bonus points if the digital credential can be bidirectionally transformed to the VCDM and can be clearly interpreted in the way it was originally intended to be interpreted (before it was transformed AND after it is re-transformed back into the original digital credential).

This would enable people to say "mdocs are compatible with the W3C VCDM" if such a unidirectional or bidirectional transformation mechanism was documented /somewhere/. That's what the VCWG was going for, and again, this is my interpretation of what could be the best outcome given the various opinions in the WG. It's quite a large and active WG, so coming to consensus on topics such as these has started to increasingly take more time than it did in the early days (for better or worse).

Does that help explain the current state of play more @alenhorvat and @mtaimela?

alenhorvat:
Thank you @msporny and others for these valuable responses.

Yes, it answers my questions and resolves my concerns. It is a nice evolution from the VCDM v1.1.

IMO the answers should/could be compiled into a blog post or FAQ or similar, since we see the question re-appearing over and over.

Thank you!

mtaimela:
Thank you @msporny very much for your valuable time and efforts on the topic and for the work on the VC. This answered all the questions I had, it is great concept and looking forward for the evolution of future v3.

msporny:
@mtaimela and @alenhorvat, two new PRs have been raised to clarify the questions that both of you had. I have tried to incorporate your feedback in both of them:

https://github.com/w3c/vc-data-model/pull/1202

https://github.com/w3c/vc-data-model/pull/1203

Please let us know your thoughts and if both of those PRs address your concerns.

Issue #1090, opened 2023-04-18
Normative guidance regarding `@context` is not clear


OR13:
> The value of the `@context` [property](https://w3c.github.io/vc-data-model/#dfn-property) MUST be an ordered set where the first item is a [URL](https://w3c.github.io/vc-data-model/#dfn-url) with the value https://www.w3.org/ns/credentials/v2. For reference, a copy of the base context is provided in Appendix [B.1 Base Context](https://w3c.github.io/vc-data-model/#base-context). Subsequent items in the array MUST express context information and be composed of any combination of [URLs](https://w3c.github.io/vc-data-model/#dfn-url) or objects. It is RECOMMENDED that each [URL](https://w3c.github.io/vc-data-model/#dfn-url) in the `@context` be one which, if dereferenced, results in a document containing machine-readable information about the `@context.`

- https://w3c.github.io/vc-data-model/#contexts

However, the current context that is resolved contains non normative members and data integrity proof specific values:

- https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v2

This makes the guidance to implementers ambiguous.

This is one of several issues I am filing related to the discussions on:

https://github.com/w3c/vc-data-model/pull/1082

I propose the following:

1. The v2 context only contain JSON-LD terms definitions for "JSON Members" that are normative, in the core data model.
2. The v2 context should not contain definitions for "reserved" or "proposed" terms, I am happy for those to also be hosted by W3C though, perhaps in a new context for "experimental features", @msporny had proposed something similar previously.
3. The W3C vocabulary for VCs should define only normative terms, it does not need to "reserve URLs" when their meaning is not yet established by the group, for this case, the correct term URL is the "issuer dependent one" since the issuer has no other reference to use.

iherman:
I repeat some of the points I made in https://github.com/w3c/vc-data-model/pull/1093

> The @context file does not define anything whatsoever. The term in question is defined in the VCDM specification and the VCDM vocabulary document. In the sense of the specification, the @context file is merely a helpful tool for JSON-LD implementers to make it sure that when they use the JSON key termOfUse, they indeed use the one the WG has agreed upon. Fundamentally, that is al it does. If the item is removed from the @context, by virtue of the @vocab that we introduced in @context, the term will instead be turned into https://www.w3.org/ns/credentials/issuer-dependent#termOfuse, which is not what we would want.

[...]

PR #1091, opened 2023-04-18
Remove DataIntegrityProof from v2 context

OR13:
Addresses part of #1090

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1091.html" title="Last updated on Apr 18, 2023, 4:40 PM UTC (4263186)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1091/fdbf89e...4263186.html" title="Last updated on Apr 18, 2023, 4:40 PM UTC (4263186)">Diff</a>

### DIFF
--- a/contexts/credentials/v2
+++ b/contexts/credentials/v2
@@ -90,69 +90,6 @@
           "@type": "@id"
         }
       }
-    },
-
-    "DataIntegrityProof": {
-      "@id": "https://w3id.org/security#DataIntegrityProof",
-      "@context": {
-        "@protected": true,
-        "id": "@id",
-        "type": "@type",
-        "challenge": "https://w3id.org/security#challenge",
-        "created": {
-          "@id": "http://purl.org/dc/terms/created",
-          "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
-        },
-        "domain": "https://w3id.org/security#domain",
-        "expires": {
-          "@id": "https://w3id.org/security#expiration",
-          "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
-        },
-        "nonce": "https://w3id.org/security#nonce",
-        "proofPurpose": {
-          "@id": "https://w3id.org/security#proofPurpose",
-          "@type": "@vocab",
-          "@context": {
-            "@protected": true,
-            "id": "@id",
-            "type": "@type",
-            "assertionMethod": {
-              "@id": "https://w3id.org/security#assertionMethod",
-              "@type": "@id",
-              "@container": "@set"
-            },
-            "authentication": {
-              "@id": "https://w3id.org/security#authenticationMethod",
-              "@type": "@id",
-              "@container": "@set"
-            },
-            "capabilityInvocation": {
-              "@id": "https://w3id.org/security#capabilityInvocationMethod",
-              "@type": "@id",
-              "@container": "@set"
-            },
-            "capabilityDelegation": {
-              "@id": "https://w3id.org/security#capabilityDelegationMethod",
-              "@type": "@id",
-              "@container": "@set"
-            },
-            "keyAgreement": {
-              "@id": "https://w3id.org/security#keyAgreementMethod",
-              "@type": "@id",
-              "@container": "@set"
-            }
-          }
-        },
-        "cryptosuite": "https://w3id.org/security#cryptosuite",
-        "proofValue": {
-          "@id": "https://w3id.org/security#proofValue",
-          "@type": "https://w3id.org/security#multibase"
-        },
-        "verificationMethod": {
-          "@id": "https://w3id.org/security#verificationMethod",
-          "@type": "@id"
-        }
-      }
     }
   }
 }


contexts/credentials/v2
```diff
@@ -90,69 +90,6 @@
...
-              "@id": "https://w3id.org/security#keyAgreementMethod",
-              "@type": "@id",
-              "@container": "@set"
-            }
-          }
-        },
-        "cryptosuite": "https://w3id.org/security#cryptosuite",
-        "proofValue": {
-          "@id": "https://w3id.org/security#proofValue",
-          "@type": "https://w3id.org/security#multibase"
```
OR13, line 149: Related PR in Data Integrity

https://github.com/w3c/vc-data-integrity/pull/92

dlongley (changes requested):
-1 to doing this since we have said that it makes developer's lives easier for many common cases in a variety of other discussions.

If we don't want to do this here, perhaps we should also reconsider, for example, separating out the core context into two: one that has `@vocab` and one that doesn't -- to avoid impacting those that do not want to deal with its usage.

I think we should figure out which we're prioritizing: ideological / abstraction boundary purity or ease of common use. It seems these things are sometimes in conflict with one another.

msporny (changes requested):
-1 for the following reasons:

* We [already had consensus](https://github.com/w3c/vc-data-model/pull/943) to merge this a while ago.
* We changed the way we did Data Integrity, based on @OR13's concerns, to make it easier, which resulted in this definition in the core context, so we could move toward having a single context.
* There is very little cost (just a few extra bytes in a context that's perma-cached) to having this in the core context, and then only to those that use JSON-LD.

Overall, this feels like an appeal to "academic purity", which overrides the practical benefits of having this in here and it would make those using Data Integrity have to do unnecessary, extra work for no practical benefit.

OR13:
My proposed changes result in a consistent experience for data integrity implementers, and they save cost for non data integrity implementers.

## v1 and v1.1
```json
"@context": [
    "https://w3id.org/security/data-integrity/v1"
  ],
```
## v2
```json
"@context": [
    "https://w3id.org/security/data-integrity/v1"
  ],
```

## using data integrity to secure arbitrary json
```json
"@context": [
    {"title": "https://schema.org/title"},
    "https://w3id.org/security/data-integrity/v1"
]
```

TallTed (changes requested):
I'm with @msporny [see [1](https://github.com/w3c/vc-data-model/pull/1091#pullrequestreview-1390757249)] and @dlongley [see [2](https://github.com/w3c/vc-data-model/pull/1091#pullrequestreview-1390598800)]

OR13:
Also https://github.com/w3c/vc-data-integrity/pull/92

mprorock (approved):
This section of the context (or an equivalent) seems like is should be in the data integrity context, not in the core context

Approving, contingent on the data integrity side having coverage for this in their context

selfissued (approved):
Since DataIntegrityProof isn't needed for many VCs, it makes sense to remove it.  In general, we should be aligning the base context with the elements that are intended for use in all VCs.

decentralgabe (approved):
agree with @selfissued's comments

msporny:
> Since DataIntegrityProof isn't needed for many VCs

Please provide proof of this statement given that @OR13, @mprorock, @decentralgabe, and @selfissued's entire argument for this PR hinges on it.

While it might be true that VCs that your organizations are issuing don't require `DataIntegrityProof`, it's also true that there are other organizations issuing VCs that plan to make significant use of `DataIntegrityProof`.

As was mentioned on the call today, this property is included (as well as `name` and `description`) because it helps a subset of developers and it is not a great burden to developers that don't plan to use `DataIntegrityProof`.

In an attempt to put this in perspective, the same argument that argues for the removal of `DataIntegrityProof` applies to `credentialSchema`. We could assert that there is no demonstration of widespread usage of the feature and that only a very small subset of the community is going to find the property useful; therefore, we should remove it from the core context. This is, of course, a questionable argument given that the cost for developers that are not going to use `credentialSchema` is quite low.

If the WG is going to argue for the removal of `DataIntegrityProof` on these grounds, we should be consistent and argue for the removal of `credentialSchema` as well. Doing so, of course, would be inadvisable.

peacekeeper (commented):
Considering that the base media type is `application/vc+ld+json` and that `proof` is a core property, I think it also makes sense to keep `DataIntegrityProof` in the v2 context, in order to be able to easily add proofs to a JSON-LD VC.

Removing it I think could potentially also lead to interoperability problems between Data Integrity proofs and VC-JWT, since one type of proof would require an additional context and the other one wouldn't.

iherman:
As someone put it, in this sounds like an "academic purity". I would say that even in terms of academic purity it does not hold, because the `@context` file _does not define_ anything. _It is a tool to make the life of developers, who care about the Linked Data aspect of VC-s easier_, whilst it is totally opaque for those who care of a pure JSON only.

To put things to extremes, "academic purity" might be to say: do not use an external `@context` file at all. Instead, along the lines proposed in https://github.com/w3c/vc-data-model/pull/1091#issuecomment-1514806914, use

```json
{
    "@context" : {
        "title": "https://schema.org/title",
        "issuer" :  {
            "@id" : "https://www.w3.org/2018/credentials#issuer",
            "@type": "@id",
        },
        "DataIntegrityProof" : "https://w3id.org/security#DataIntegrityProof",
         …
        "@vocab": "https://www.w3.org/ns/credentials/issuer-dependent#"
    }
}
```

requiring all developers to list all terms they use in a specific VC instance. Linked Data-wise, that would be equivalent to the usage of the external `@context` file. It would also make the life of developers miserable. At best, each developer would create his/her own `@context` file, at worst, they would drop JSON-LD altogether.

-1 to this proposal.

contexts/credentials/v2
```diff
@@ -90,69 +90,6 @@
...
-              "@container": "@set"
-            },
-            "keyAgreement": {
-              "@id": "https://w3id.org/security#keyAgreementMethod",
-              "@type": "@id",
-              "@container": "@set"
-            }
-          }
-        },
-        "cryptosuite": "https://w3id.org/security#cryptosuite",
```
OR13, line 146: Keeping this in the v2 context, will allow anyone to create new crypto suites with no W3C review after the TR is published... here are some examples:

```json
{
    "type": "DataIntegrityProof",
    "cryptosuite": "bitcoin-2023",
    "created": "2020-11-05T19:23:24Z",
    "verificationMethod": "https://ldi.example/issuer/1#z6MkjLrk3gKS2nnkeWcmcxi
      ZPGskmesDpuwRBorgHxUXfxnG",
    "proofValue": "z4oey5q2M3XKaxup3tmzN4DRFTLVqpLMweBrSxMY2xHX5XTYVQeVbY8nQA
      VHMrXFkXJpmEcqdoDwLWxaqA3Q1geV6"
  }
  ```
  
```json
{
    "type": "DataIntegrityProof",
    "cryptosuite": "ethereum-2023",
    "created": "2020-11-05T19:23:24Z",
    "verificationMethod": "https://ldi.example/issuer/1#z6MkjLrk3gKS2nnkeWcmcxi
      ZPGskmesDpuwRBorgHxUXfxnG",
    "proofValue": "z4oey5q2M3XKaxup3tmzN4DRFTLVqpLMweBrSxMY2xHX5XTYVQeVbY8nQA
      VHMrXFkXJpmEcqdoDwLWxaqA3Q1geV6"
  }
  ```
  
```json
{
    "type": "DataIntegrityProof",
    "cryptosuite": "ecdsa-p521",
    "created": "2020-11-05T19:23:24Z",
    "verificationMethod": "https://ldi.example/issuer/1#z6MkjLrk3gKS2nnkeWcmcxi
      ZPGskmesDpuwRBorgHxUXfxnG",
    "proofValue": "z4oey5q2M3XKaxup3tmzN4DRFTLVqpLMweBrSxMY2xHX5XTYVQeVbY8nQA
      VHMrXFkXJpmEcqdoDwLWxaqA3Q1geV6"
  }
```

Similar to all predicate extension points in the core data model, `cryptosuite` is an extension point, unlike other predicates, this one is not bound to any RDF types.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-05-24-vcwg#section1-6) on 2023-05-24

- no resolutions were taken

#### 1.6. Add JSON Web Token Examples (pr vc-status-list-2021#60)

_See github pull request [vc-status-list-2021#60](https://github.com/w3c/vc-status-list-2021/pull/60)._




**Orie Steele:** We seem to have a difference of opinion regarding adding examples to the specification. And I would like to merge the examples as-is. I don't agree with the perspective that every example in our spec needs multiple context values.  
… I don't agree with that and in light of bundling the Data Integrity context.  
… This adds informative examples, can we merge this, please, thank you.  

> *Greg Berstein:* Please!

**Manu Sporny:** -1 to merge until the WG discusses it. With all the examples it keeps coming up and we need to have a consistent way that we're doing examples and figure out what it is so the editors can apply that uniformly across all the specs.  

> *Phillip Long:* is this a special topics call?

**Kristina Yasuda:** I'm up for spending 5-10 minutes on this call on this.  

> *Orie Steele:* lets remove the DataIntegrityProof RDF class from the core context... and make the examples include multiple contexts.

**Manu Sporny:** I don't know if we'll get through it in 5 minutes. The proposal is to just use the example in the core spec.  
… Use it as an example across all the specs that need to use an example. That way we're just using an example that we've agreed to use across all the specs and we don't need to do variations of that in every other specification.  

_See github pull request [vc-data-model#1091](https://github.com/w3c/vc-data-model/pull/1091)._




**Kristina Yasuda:** Are you fundamentally not ok with the example, Manu?  

**Manu Sporny:** Yes.  

**Kristina Yasuda:** Anyone else who has strong opinions about what examples we use?  

**Orie Steele:** All the examples conformant to the core data model.  

**Michael Prorock:** My question is along the same line... what precisely is wrong with the examples?  
… At least by my read and going and running the example code it all checks out fine.  

**Manu Sporny:** One of the things is that we have a suspension false entry in here, I don't know where that property came from, I think Orie said that's an artifact in the PR and it's not in the data model.  
… There's a concern over the use of the undefined / issuer-independent context / vocab.  
… Now we're using the issuer-dependent terms in the example.  
… There are a number of people that said in the other PR that's not a good practice.  
… We've had people say that they would rather the examples use the examples context instead of the issuer-defined one.  
… I don't know why we're doing in status list when core does something else.  

**Dave Longley:** A lot of developers will come along and look at examples to get things started, and there are such things as good examples and bad examples.  

> *Orie Steele:* IMO a good example is one that uses the minimal required fields.  
> it is bad form to include optional stuff in example, and bad form to include DataIntegrityProof in the core v2 context.

**Dave Longley:** You can create examples that are bad examples, which would be conformant, but wouldn't look good. It's not enough to say "We created an example and it validates so it's fine." If we get people started off on he wrong foot, or encouraging market failures, we're making mistakes. We should not be flippant about the examples we put in the spec. They're important, developers will look at these examples and maybe not the text.  
… Saying "this verifies" is not enough, these are entry points into the spec and possibly exit points.  

**Kristina Yasuda:** Sounds like we're leaving this issue open for now.  
… Anything on the BBS cryptosuite stuff?  

</details>

PR #1092, opened 2023-04-18
Remove evidence

OR13:
Remove proposed term from v2 context

Addresses #1090

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1092.html" title="Last updated on Apr 18, 2023, 4:45 PM UTC (935862d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1092/fdbf89e...935862d.html" title="Last updated on Apr 18, 2023, 4:45 PM UTC (935862d)">Diff</a>

### DIFF
--- a/contexts/credentials/v2
+++ b/contexts/credentials/v2
@@ -29,10 +29,6 @@
         "description": {
           "@id": "https://schema.org/description"
         },
-        "evidence": {
-          "@id": "https://www.w3.org/2018/credentials#evidence",
-          "@type": "@id"
-        },
         "validFrom": {
           "@id": "https://www.w3.org/2018/credentials#validFrom",
           "@type": "http://www.w3.org/2001/XMLSchema#dateTime"


msporny (changes requested):
-1 for the following reasons:

* This term has [already been used](https://w3c.github.io/vc-specs-dir/#evidence) as an extension point in the VC Specifications Directory.
* There is active interest in this extension point (as evidenced by the discussion on the calls and the registration of an extension specification).

iherman:
-1, see https://github.com/w3c/vc-data-model/pull/1093#issuecomment-1514232247

OR13:
As I stated on the call.

If the working group has agreed to this term, as a normative requirement, it should not be reserved, and it should be defined in the TR AND in the JSON-LD context.

We should avoid bundling non normative stuff into the JSON-LD context, since they are cached, they can be split by normative and non normative easily, and then implementers can more easily signal their intention to use experimental extensions.

dlongley (changes requested):
-1 This will lead to mistakes because of `@vocab` being in the core context.

David-Chadwick:
-1 from me also

awoie:
-1 from my end as well since OIDF eKYC folks are using it.

brentzundel:

OR13:

msporny:
> I thought we had agreement, that we would be reserving this URL and term name, and not defining it or using it in the v2 spec?

Removing the declaration from the v2 context (what this PR dos) is something that we did not have consensus on, and are unlikely to get consensus on it. It's been a week since being marked pending close. Closing.

PR #1094, opened 2023-04-18
Remove refreshService

OR13:
Addresses #1090

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1094.html" title="Last updated on May 4, 2023, 9:12 PM UTC (5d1eb5c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1094/fdbf89e...5d1eb5c.html" title="Last updated on May 4, 2023, 9:12 PM UTC (5d1eb5c)">Diff</a>

### DIFF
--- a/contexts/credentials/v2
+++ b/contexts/credentials/v2
@@ -53,10 +53,6 @@
           "@type": "@id",
           "@container": "@graph"
         },
-        "refreshService": {
-          "@id": "https://www.w3.org/2018/credentials#refreshService",
-          "@type": "@id"
-        },
         "termsOfUse": {
           "@id": "https://www.w3.org/2018/credentials#termsOfUse",
           "@type": "@id"


msporny (changes requested):
-1 for the following reasons:

* This term has [already been used](https://w3c.github.io/vc-specs-dir/#refresh-service), by [two extension specifications](https://w3c-ccg.github.io/vc-refresh-2021/), as an extension point in the VC Specifications Directory.
* The TruAge system, which has shipped, uses this extension point.
* There is active interest in this extension point (as evidenced by the two specifications and deployment above that utilize the extension point).

OR13:
@msporny if there are 2 independent implementations of this, it should be defined normatively, and not `reserved`.

Once that has happened, it should be included in the core context, not before.

dlongley (changes requested):
-1 This will lead to mistakes because of `@vocab` being in the core context.

David-Chadwick:
Having the refresh service in the VC is an invitation to verifiers to call home, so should be removed from the possibility of being in the VC and only allowed in the VP, thereby allowing the holder to decide if the verifier can see this field or not. The issuer can always send this information to the holder external to the VC (e.g. in the protocol or in a VP)

msporny:
> Having the refresh service in the VC is an invitation to verifiers to call home

Verifiers don't use the `refreshService`, the holder does. If a verifier were to try to use the `refreshService` they'd be denied.

> so should be removed from the possibility of being in the VC and only allowed in the VP

That's an interesting approach... will have to think about it more.

> thereby allowing the holder to decide if the verifier can see this field or not. The issuer can always send this information to the holder external to the VC (e.g. in the protocol or in a VP)

If it's sent out of band, there is no standard way to do it.

> @msporny if there are 2 independent implementations of this, it should be defined normatively, and not reserved.

TruAge is one organization, Digital Bazaar is the other (but doesn't count since TruAge uses our implementations), there is a third (but we'll have to work through some NDA issues to get them to speak publicly about it). I don't know what the WG's bar is to "proving it's implemented by multiple independent parties". 

But, whatever happens above, @dlongley and @iherman's statements stand... if we don't define it in the core context, `@vocab` will kick in and default it to the wrong vocabulary URL.

David-Chadwick:
`If a verifier were to try to use the refreshService they'd be denied.`
that misses the point. It means that the issuer knows where the holder has used the VC, so it is privacy invasive.
`If it's sent out of band, there is no standard way to do it.`
When there is standard issuing protocol, such as OIDC4VCI, then there is a standard way.

dlongley:
@David-Chadwick,

> that misses the point. It means that the issuer knows where the holder has used the VC, so it is privacy invasive.

It seems that if the issuer has any means by which the verifier could communicate with it, that the verifier could send a VC ID (or any other unique identifier from a VC) to it. Clearly, this can't be the bar by which we evaluate what can be in a VC. Now, I'm sure that's not what you intend here, but I'm trying to sort out if it ends up being the same thing when we follow the logic of it.

I agree that it's best to discourage bad behavior, but we can't control the communication between verifiers and issuers. We do want issuers that *want* to protect the privacy of those people that they issue VCs to to be able to do that to the best of their ability -- so I think we need to frame the discussion around that goal instead. I think this also implies we shouldn't restrict something that doesn't actually significantly mitigate a privacy problem as it may cause other problems or hinder innovation (including potentially better privacy preserving refresh mechanisms!).

> When there is standard issuing protocol, such as OIDC4VCI, then there is a standard way.

I'm not sure this helps here. If there's a standard protocol out there that could be followed by a verifier up until the point where they must reveal the specific VC they want refreshed (e.g., by ID), then they can always send whatever value they want to there. It won't be honored or accepted by the issuer, presumably, but the VC ID will be transmitted to them. That seems to make any standard protocol that allows this to fall into the same category flagged here as a potential privacy problem.

If that standard protocol doesn't allow transmission of the VC ID, let's say, for example, it only allows refresh by credential type, then that protocol won't allow a user to have N-many different VCs of the same type -- when coupled with the ability to refresh any particular one by choice. This could become a problem for that protocol in the future and I would expect it to be changed, circumventing the principle we tried to apply here -- likely without people realizing it or understanding the circuitous logic behind it.

Is your suggestion that we specifically add normative language that says a holder can't have N-many different VCs of the same type -- unless they cannot refresh any specific one of them? I doubt it (it seems an odd thing to restrict and one we couldn't test). This suggests to me that preventing a refresh service from being listed in a VC on the basis that it will stop a verifier from abusing it seems flawed. Maybe we should just have some language in the privacy considerations section instead?

Additionally, we should consider how people may want to port VCs between various different wallets. Does OID4VCI allow this without having to go back to the issuer? So, for example, the user's private choice of wallet is preserved? Does it specify how to ensure the user can continue to refresh? If not, we should expect users to find that protocol frustrating and ask for updates to it or they may gravitate towards something else.

It's likely beneficial to simplicity, tooling, and interoperability if everything the user needs to at least start engaging in a refresh protocol for a specific VC is encapsulated in the VC itself -- but that does imply that if they transmit that VC entirely to any other party, that party will be able to see that information.

We could also point to selective disclosure mechanisms as a means by which presenters could omit information such as refresh service data from what they present to verifiers. This approach would prevent verifiers from seeing the information and would still keep everything self-contained for portability use cases. In fact, this may be a more privacy preserving approach in the end, if the refresh service is a URL capability -- instead of using some discovery protocol(s) from other information that the verifier could learn or send unique VC information to.

In short, I'm not convinced that preventing a refresh service from appearing in a VC is either the best way to protect privacy -- nor that it will even sufficiently mitigate any threats there.

msporny:
@David-Chadwick wrote:
> It means that the issuer knows where the holder has used the VC, so it is privacy invasive.

Any URL provided by the issuer, that is fetched by a Verifier, would be "privacy invasive" by that definition.

I agree that we want to minimize URLs of that nature, and we do that quite aggressively in the TruAge program (such URLs do not exist by design, signatures are not repeated by design, all VCs are single-use bearer tokens, etc.). That said, that's at one (fairly extreme) end of the privacy-enhancing spectrum.

In the middle are VCs that have no design requirements to not be tracked back to the issuer... For example, an issuer logo provided in the `issuer` field would fall into the same category as the use case you're concerned about above, and as @dlongley said -- that's probably not the bar we're trying to define here.

I take your point that we should avoid URLs of this nature, but `refreshService` isn't such a URL. At least in our deployments, `refreshService` points to a general URL that the entire population uses... that is, hitting the URL doesn't impact herd privacy... and besides, the verifier is definitely not expected to hit that URL (and in the TruAge scenario, the verifier is not even given access to that URL -- the VC that contains the `refreshService` is designed to never be handed over to a 3rd party... ever).

It is true that `refreshService` could be included in a VC in such a way that the issuer uses it as a tracking vector... but honestly, there are "better" ways to do that sort of privacy-invasive thing (like individual-bound tracking image URLs, for instance). 

We're going to have to have some guidance here wrt. tracking URLs AND possibly create standards for holder's to discover these sorts of tracking URLs in a decentralized fashion. For example, a service that helps you submit hashes for URL strings to see if anyone else out there has the same URL... if they do, the liklihood that you're dealing w/ a tracking URL is low... if you're the only one that has one of those URLs in your VC... well, then, it's likely that someone has injected a tracking URL in your VC (and your digital wallet should probably warn you about that credential, and possibly the issuer).

David-Chadwick:
`It seems that if the issuer has any means by which the verifier could communicate with it, that the verifier could send a VC ID (or any other unique identifier from a VC) to it. Clearly, this can't be the bar by which we evaluate what can be in a VC.`
I think it is more nuanced than that. Clearly any issuer could add any proprietary feature that allows a verifier to call home. What we should **not** be doing is standardising such a feature, as this implies we are condoning it. 

`If there's a standard protocol out there that could be followed by a verifier up until the point where they must reveal the specific VC they want refreshed (e.g., by ID), then they can always send whatever value they want to there`
I would argue that the first step of any standard issuing protocol should be to authenticate the caller. So revealing the specific VC should not be an issue since the verifier will never get to this step (unless the verifier is also a customer of the issuer and can authenticate - but this is surely an edge case).

`how people may want to port VCs between various different wallets.`
This is not possible when the private key is stored in the device's hardware since it cannot be copied. The issuer must be involved in this case. So assuming the issuer is involved is the safer approach, rather than assuming that users can copy their keys and VCs to anywhere they wish to.

`selective disclosure mechanisms as a means by which presenters could omit information such as refresh service data from what they present to verifiers`
I agree that this is a mitigating step, but it is only applicable if the wallet and verifier support the particular algorithm/mechanism. The better mitigating step is to not have the refresh property in the first place and to send the refresh information to the wallet in the issuing protocol (which both the wallet and issuer must support if they are to interwork).

brentzundel:

OR13:

OR13:
(sry for close, too close to comment button).

OR13:
Closing this PR in favor of https://github.com/w3c/vc-data-model/pull/1113

PR #1107, opened 2023-04-27
Clarify relationship bx securing mechanisms and media types

brentzundel:
This PR attempts to add some clarity around the relationship between securing mechanisms and media types.

It primarily uses language suggested in Issue #1065. 
It adds that media types defined that allow for transformation into application/vc+ld+json should begin with application/vc+
It states that application/vc+ld+json does not imply any securing mechanism
it states that other media types should not imply any securing mechanism
I removes the note about Issue #1065

fixes #1065

***
<a href="https://pr-preview.s3.amazonaws.com/brentzundel/vc-data-model/pull/1107.html" title="Last updated on May 4, 2023, 9:48 PM UTC (1e350e3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1107/c964156...brentzundel:1e350e3.html" title="Last updated on May 4, 2023, 9:48 PM UTC (1e350e3)">Diff</a>

Issue #1102, opened 2023-04-27
Single Extension Point


David-Chadwick:
Here is a radical proposal which simultaneously simplifies the data model and removes all arguments about which extension points to keep and which to move to a reserved list.

DELETE THEM ALL AND REPLACE THEM WITH A SINGLE EXTENSION POINT

The single extension point can be regarded as the supertype of all the existing extension points. The only mandatory parameter of the extension point is the type, and this must be a URI. After that, all the following parameters are determined by the type. The existing defined types for all the existing V1 extension points can still be used in the new generic extension point, since they are globally unique URIs. Hence their existing definitions will not change. All that changes is the definition of the new wrapper parameter. Instead of being ToU, or Evidence or whatever, it becomes extensionPoint. In this way there is no need to ever define a new extension point. The extensionPoint parameter can be used for all of them.

P.S. This is not a new idea, it's the existing standard X.509v3 model which has been proven to work.

decentralgabe:
Could you give an example @David-Chadwick? How is this different than using `@context` as a single extension point?

selfissued:
I appreciate you think about alternatives and simplifications, David.  However in this case, I think that developers would be more confused by mixing together fields with different characteristics into a single extension point.  I'd rather that each field retain its own JSON member name and description.

David-Chadwick:
Yes I mean having a single extension point called`extensionPoint`. This is not confusing to implementors as the X.509 experience indicates. Rather it is more confusing to implementors to have several different flavours of extension points because then they have to decide which flavour to use for their new extension. When there is only one, then this dilemma does not arise. All extensions go in the same extension point. They are differentiated as now by their `type` parameter. This is in fact much easier for implementors to use because they simply need to look at the type of the extension and compare it their list of known extension types.  A URI match means they understand the extension. An unknown URI means they do not. Today implementors have to repeat this code for each of the different flavours of extension point so is in fact more code than what I am proposing.
@selfissued If the descriptions of the extension points were solid and agreed upon then we would not be needing to add the wording that their definitions may change in future, as is currently being proposed for the reserved properties table. So I assert that having just one extension point with solid wording is much less confusing for everyone.

OR13:
I am not in favor of this change.

msporny:
-1 to an "extension point bucket" :P

iherman:
[...]

PR #1108, opened 2023-04-28
Add `renderMethod` as reserved extension point.

msporny:
This PR builds upon PR 1097 and adds three new properties (and their reserved URLs) to the reserved properties table: 

* `confirmationMethod` / `confidenceMethod`
* `presentationSchema`
* `renderMethod`

There is now also an example specification that demonstrates how these reserved properties could be leveraged by external specifications: [Verifiable Credential Rendering Methods](https://digitalbazaar.github.io/vc-render-method/)

The specification notes that it [utilizes the reserved property](https://digitalbazaar.github.io/vc-render-method/#the-rendermethod-property) in the VC Data Model, and then [defines a type](https://digitalbazaar.github.io/vc-render-method/#svgrenderingtemplate2023) that can be used with the extension point.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1108.html" title="Last updated on May 13, 2023, 5:53 PM UTC (d4d935b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1108/354036f...d4d935b.html" title="Last updated on May 13, 2023, 5:53 PM UTC (d4d935b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3255,13 +3255,34 @@ see Section <a href="#types"></a>.
               <th>Description</th>
             </tr>
           </thead>
+
           <tbody>
+            <tr>
+              <td>`confirmationMethod` / `confidenceMethod`</td>
+              <td>
+A property used for specifying how a <a>verifier</a> can confirm, or
+gain more confidence in, the presentation of the credential being associated
+with a particular <a>subject</a>. The associated vocabulary URL MUST be
+`https://www.w3.org/2018/credentials#confirmationMethodORconfidenceMethod`.
+
+                <p class="issue" title="Bikeshedding confirmation/confidence">
+The name of this property is currently being debated by the Working Group.
+                </p>
+              </td>
+            </tr>
             <tr>
               <td>`evidence`</td>
               <td>
 A property used for specifying the evidence that was presented in order to
 issue the credential. The associated vocabulary URL MUST be
 `https://www.w3.org/2018/credentials#evidence`.
+              </td>
+            </tr>
+            <tr>
+              <td>`presentationSchema`</td>
+              <td>
+A property used for specifying the schema for presentations. The associated
+vocabulary URL MUST be `https://www.w3.org/2018/credentials#presentationSchema`.
               </td>
             </tr>
             <tr>
@@ -3273,14 +3294,24 @@ associated vocabulary URL MUST be
               </td>
             </tr>
             <tr>
+              <td>`renderMethod`</td>
+              <td>
+A property used for specifying how to render a credential into a visual,
+auditory, or haptic format. The associated vocabulary URL MUST be
+`https://www.w3.org/2018/credentials#renderMethod`.
+              </td>
+            </tr>
+            <tr>
               <td>`termsOfUse`</td>
               <td>
+An example entry for a reserved property.
 A property used for specifying the terms of use for a credential. The associated
 vocabulary URL MUST be `https://www.w3.org/2018/credentials#termsOfUse`.
               </td>
             </tr>
           </tbody>
         </table>
+
 
         <p>
 An unofficial list of specifications that are associated with the extension
--- a/index.html
+++ b/index.html
@@ -3252,7 +3252,7 @@ see Section <a href="#types"></a>.
 The following reserved properties are at risk and will be removed from the
 specification if at least one specification with two independent implementations
 are not demonstrated by the end of the Candidate Recommendation Phase:
-`confirmationMethod` / `confidenceMethod`, `presentationSchema`, and
+`confidenceMethod`, `presentationSchema`, and
 `renderMethod`.
         </p>
 
@@ -3266,16 +3266,13 @@ are not demonstrated by the end of the Candidate Recommendation Phase:
 
           <tbody>
             <tr>
-              <td>`confirmationMethod` / `confidenceMethod`</td>
+              <td>`confidenceMethod`</td>
               <td>
-A property used for specifying how a <a>verifier</a> can confirm, or
-gain more confidence in, the presentation of the credential being associated
-with a particular <a>subject</a>. The associated vocabulary URL MUST be
-`https://www.w3.org/2018/credentials#confirmationMethodORconfidenceMethod`.
+A property used for specifying how a <a>verifier</a> can gain more confidence
+in, the presentation of the credential being associated with a particular
+<a>subject</a>. The associated vocabulary URL MUST be
+`https://www.w3.org/2018/credentials#confidenceMethod`.
 
-                <p class="issue" title="Bikeshedding confirmation/confidence">
-The name of this property is currently being debated by the Working Group.
-                </p>
               </td>
             </tr>
             <tr>
--- a/index.html
+++ b/index.html
@@ -3312,7 +3312,6 @@ auditory, or haptic format. The associated vocabulary URL MUST be
             <tr>
               <td>`termsOfUse`</td>
               <td>
-An example entry for a reserved property.
 A property used for specifying the terms of use for a credential. The associated
 vocabulary URL MUST be `https://www.w3.org/2018/credentials#termsOfUse`.
               </td>
--- a/index.html
+++ b/index.html
@@ -3258,37 +3258,11 @@ see Section <a href="#types"></a>.
 
           <tbody>
             <tr>
-              <td>`confidenceMethod`</td>
-              <td>
-A property used for specifying how a <a>verifier</a> can gain more confidence
-in, the presentation of the credential being associated with a particular
-<a>subject</a>. The associated vocabulary URL MUST be
-`https://www.w3.org/2018/credentials#confidenceMethod`.
-                <p class="issue" title="(AT RISK) Reservation depends on implementations">
-This reserved property is at risk and will be removed from the
-specification if at least one specification with two independent implementations
-are not demonstrated by the end of the Candidate Recommendation Phase.
-                </p>
-              </td>
-            </tr>
-            <tr>
               <td>`evidence`</td>
               <td>
 A property used for specifying the evidence that was presented in order to
 issue the credential. The associated vocabulary URL MUST be
 `https://www.w3.org/2018/credentials#evidence`.
-              </td>
-            </tr>
-            <tr>
-              <td>`presentationSchema`</td>
-              <td>
-A property used for specifying the schema for presentations. The associated
-vocabulary URL MUST be `https://www.w3.org/2018/credentials#presentationSchema`.
-                <p class="issue" title="(AT RISK) Reservation depends on implementations">
-This reserved property is at risk and will be removed from the
-specification if at least one specification with two independent implementations
-are not demonstrated by the end of the Candidate Recommendation Phase.
-                </p>
               </td>
             </tr>
             <tr>
--- a/index.html
+++ b/index.html
@@ -3248,6 +3248,14 @@ reserved property. For more information related to adding `type` information,
 see Section <a href="#types"></a>.
         </p>
 
+        <p class="issue" title="(AT RISK) Reserved properties">
+The following reserved properties are at risk and will be removed from the
+specification if at least one specification with two independent implementations
+are not demonstrated by the end of the Candidate Recommendation Phase:
+`confirmationMethod` / `confidenceMethod`, `presentationSchema`, and
+`renderMethod`.
+        </p>
+
         <table class="simple">
           <thead>
             <tr>
--- a/index.html
+++ b/index.html
@@ -3248,14 +3248,6 @@ reserved property. For more information related to adding `type` information,
 see Section <a href="#types"></a>.
         </p>
 
-        <p class="issue" title="(AT RISK) Reserved properties">
-The following reserved properties are at risk and will be removed from the
-specification if at least one specification with two independent implementations
-are not demonstrated by the end of the Candidate Recommendation Phase:
-`confidenceMethod`, `presentationSchema`, and
-`renderMethod`.
-        </p>
-
         <table class="simple">
           <thead>
             <tr>
@@ -3272,7 +3264,11 @@ A property used for specifying how a <a>verifier</a> can gain more confidence
 in, the presentation of the credential being associated with a particular
 <a>subject</a>. The associated vocabulary URL MUST be
 `https://www.w3.org/2018/credentials#confidenceMethod`.
-
+                <p class="issue" title="(AT RISK) Reservation depends on implementations">
+This reserved property is at risk and will be removed from the
+specification if at least one specification with two independent implementations
+are not demonstrated by the end of the Candidate Recommendation Phase.
+                </p>
               </td>
             </tr>
             <tr>
@@ -3288,6 +3284,11 @@ issue the credential. The associated vocabulary URL MUST be
               <td>
 A property used for specifying the schema for presentations. The associated
 vocabulary URL MUST be `https://www.w3.org/2018/credentials#presentationSchema`.
+                <p class="issue" title="(AT RISK) Reservation depends on implementations">
+This reserved property is at risk and will be removed from the
+specification if at least one specification with two independent implementations
+are not demonstrated by the end of the Candidate Recommendation Phase.
+                </p>
               </td>
             </tr>
             <tr>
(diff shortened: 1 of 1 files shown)


awoie (approved):
lgtm

msporny:
> Can we split this PR up, so we can address each of these independently?

We can just refactor this PR after we have a discussion in it. The changes are small enough that a refactor after we get consensus will be easier than having a parallel discussion on each one.

> The reason for my objection is that both of these extensions when leveraged by an issuer, impact complexity on the holder / verifier, which I believe might lead to fragmentation and harm adoption.

Hmm, that doesn't make sense to me, perhaps there is a confusion of some kind. Could you explain this a bit more? 

The impact that `confirmationMethod/confidenceMethod` would be the exact same for `renderMethod` and `presentationSchema`. That is, they're optional properties that Holders/Verifiers can ignore, unless they want to process them.

Also, note the issue marker here: https://github.com/w3c/vc-data-model/pull/1108/files#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051R3200-R3206

Which states that all three are at risk unless a specification and two interoperable implementations exist.

dlongley (approved):

index.html
```diff
@@ -3221,16 +3250,26 @@ <h3>Reserved Extension Points</h3>
...
+              <td>
+A property used for specifying how to render a credential into a visual,
+auditory, or haptic format. The associated vocabulary URL MUST be
+`https://www.w3.org/2018/credentials#renderMethod`.
+              </td>
+            </tr>
             <tr>
               <td>`termsOfUse`</td>
               <td>
+An example entry for a reserved property.
```
davidlehn, line 3264: Should this be here?

msporny: No, it shouldn't... bad merge. Fixed in cabcea159857e6f3cfa75cc92cd3d4aed1c0da8f. Thank you!

gobengo (approved):
these seem useful to define now to enable adoption. I dont know of any other vocabulary that these short names would collide with.

index.html
```diff
@@ -3197,14 +3197,36 @@ <h3>Reserved Extension Points</h3>
 see Section <a href="#types"></a>.
         </p>
 
+        <p class="issue" title="(AT RISK) Reserved properties">
+The following reserved properties are at risk and will be removed from the
+specification if at least one specification with two independent implementations
+are not demonstrated by the end of the Candidate Recommendation Phase:
+`confirmationMethod` / `confidenceMethod`, `presentationSchema`, and
+`renderMethod`.
+        </p>
```
brentzundel, line 3206: It isn't clear to me if this means only from the pertinent sections of the spec, or from this table as well.
Would it be better to put this text before each  at risk property?

msporny: > It isn't clear to me if this means only from the pertinent sections of the spec, or from this table as well. 

Yeah, good point, I wasn't thinking about the existing sections of the spec.

We should probably just add at-risk markers to `evidence`, `refreshService`, and `termsOfUse` to be crystal clear wrt. their current state (at risk) as we enter CR. I'll do that in a separate PR.

> Would it be better to put this text before each at risk property?

Yes, I can do that. It felt like it would be repetitive to copy/paste the same text to each entry. In the grand scheme of things, it doesn't really matter because the entry and/or text would be removed as we move into PR. I'll do the revision in an update to this PR.

David-Chadwick:
Can we please discuss issue #1102 along with this issue please. Thanks.

decentralgabe (approved):

index.html
```diff
@@ -3197,11 +3197,12 @@ <h3>Reserved Extension Points</h3>
...
-        <p class="issue" title="Reserved properties under debate">
-The following extension point properties are under consideration for being
-marked as reserved: `presentationSchema`, `credentialRefresh`, `termsOfUse`,
-`evidence`, `renderMethod`, `refreshService`, and
-`confirmationMethod/confidenceMethod`.
+        <p class="issue" title="(AT RISK) Reserved properties">
+The following reserved properties are at risk and will be removed from the
+specification if at least one specification with two independent implementations
+are not demonstrated by the end of the Candidate Recommendation Phase:
+`confirmationMethod` / `confidenceMethod`, `presentationSchema`, and
```
brentzundel, line 3204:
```suggestion
`confidenceMethod`, `presentationSchema`, and
```

index.html
```diff
@@ -3214,17 +3215,60 @@ <h3>Reserved Extension Points</h3>
 
           <tbody>
             <tr>
+              <td>`confirmationMethod` / `confidenceMethod`</td>
```
brentzundel, line 3218:
```suggestion
              <td>`confidenceMethod`</td>
```

index.html
```diff
@@ -3214,17 +3215,60 @@ <h3>Reserved Extension Points</h3>
...
               <td>
-`example`
+A property used for specifying how a <a>verifier</a> can confirm, or
+gain more confidence in, the presentation of the credential being associated
+with a particular <a>subject</a>. The associated vocabulary URL MUST be
+`https://www.w3.org/2018/credentials#confirmationMethodORconfidenceMethod`.
```
brentzundel, line 3223:
```suggestion
A property used for specifying how a <a>verifier</a> can gain more confidence
in, the presentation of the credential being associated with a particular
<a>subject</a>. The associated vocabulary URL MUST be
`https://www.w3.org/2018/credentials#confidenceMethod`.
```

index.html
```diff
@@ -3214,17 +3215,60 @@ <h3>Reserved Extension Points</h3>
...
+
+                <p class="issue" title="Bikeshedding confirmation/confidence">
+The name of this property is currently being debated by the Working Group.
+                </p>
```
brentzundel, line 3227:
```suggestion
```

brentzundel (approved):
This week's special topic call resulted in (I believe) consensus around `confidenceMethod`
I have made suggestions to reflect that.

index.html
msporny: Fixed in https://github.com/w3c/vc-data-model/pull/1108/commits/a86f02053c1063164671a57d7962b868a916b90b.

index.html
```diff
@@ -3211,7 +3211,22 @@ <h3>Reserved Extension Points</h3>
               <th>Description</th>
             </tr>
           </thead>
+
           <tbody>
+            <tr>
+              <td>`confidenceMethod`</td>
```
OR13, line 3217:
```suggestion
              <td>`confirmationMethod`</td>
```

index.html
```diff
@@ -3211,7 +3211,22 @@ <h3>Reserved Extension Points</h3>
...
+              <td>
+A property used for specifying how a <a>verifier</a> can gain more confidence
+in, the presentation of the credential being associated with a particular
+<a>subject</a>. The associated vocabulary URL MUST be
+`https://www.w3.org/2018/credentials#confidenceMethod`.
```
OR13, line 3222:
```suggestion
`https://www.w3.org/2018/credentials#confirmationMethod `.
```

OR13: It would be good to not invent new security terminology that can so easily be confused with existing RFCs that doe exactly the same thing:

> By including a "cnf" (confirmation) claim in a JWT, the issuer of the
> JWT declares that the presenter possesses a particular key and that
> the recipient can cryptographically confirm that the presenter has
> possession of that key.
   
https://datatracker.ietf.org/doc/html/rfc7800#section-3

OR13 (changes requested):
Adding confirmation like this is going to cause interoperability problems with existing securing formats, for example JWTs.

Reserving these properties is not necessary, since JSON-LD supports extension of credential attributes using `@context`, and there is no working group consensus on how these terms will be used in the future, we should allow for implementation experience to drive the standards development process.

This PR groups 3 experimental properties into one change to the document, I think the entire PR is a mistake, but my only blocking feedback is regarding confirmation method.

Sakurann (changes requested):
For the `confidenceMethod`, I would like to see #1054 merged first.

awoie (changes requested):
I would want to get PR #1054 to get merged before or not include confidence method as a reserved property but instead make it part of the core VCDM.

awoie:
> I would want to get PR #1054 to get merged before or not include confidence method as a reserved property but instead make it part of the core VCDM.

Also the current description of confidence method in this PR is different from the description in PR #1054

msporny:
@OR13 wrote:
> Adding confirmation like this is going to cause interoperability problems with existing securing formats, for example JWTs.

Assertion stated without explanation. It seems as if this was discussed at length during the VCWG special topic call and the group came to consensus that using `cnf` was the wrong thing to do:

https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-05-02-vcwg-special

> Reserving these properties is not necessary, since JSON-LD supports extension of credential attributes using `@context`, and there is no working group consensus on how these terms will be used in the future, we should allow for implementation experience to drive the standards development process.

This is going back on the consensus position of the group to have reserved properties. There doesn't have to be WG consensus beyond that the terms are reserved and that there is a simple statement related to what the term does.

> This PR groups 3 experimental properties into one change to the document, I think the entire PR is a mistake, but my only blocking feedback is regarding confirmation method.

It seems as if you were not present at the meeting that settled the topic on `confidenceMethod` being the preferred term: 

https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-05-02-vcwg-special

Are you stating that consensus was not reached because you are objecting to the consensus reached on the call?

msporny:
@Sakurann wrote:
> For the `confidenceMethod`, I would like to see #1054 merged first.

Is this a technical objection to the PR, or a request that #1054 is merged before this PR?

msporny:
@awoie wrote:
> I would want to get PR #1054 to get merged before 

We could have done that, but #1054 is now blocked by @OR13.

> or not include confidence method as a reserved property but instead make it part of the core VCDM.

`confidenceMethod` is on the same footing as `evidence`, `termsOfUse`, and `refreshService`... that is, there has yet to be a demonstration of two implementations.  `evidence`, `termsOfUse`, and `refreshService` are listed in the reserved properties section today (even though they are listed in the specification):

https://w3c.github.io/vc-data-model/#reserved-extension-points

Why should `confidenceMethod` achieve a higher status than `evidence`, `termsOfUse`, and `refreshService` in the spec today?

To reiterate the purpose of this PR:

* Any extension point that has a definition in the specification (`evidence`, `termsOfUse`, `refreshService`, and -- presumably -- `confidenceMethod`, if #1054 is merged), but does not have two implementations for it, has its section listed in the spec as "at risk" and is listed in the reserved extension points table. If we don't get enough implementations by the end of CR, we delete the section from the spec and leave the reserved properties table entry for the item alone. If we get two implementations, we remove the entry from the reserved properties table (and the section remains in the spec).
* Any extension point that does not have a definition in the specification (`presentationSchema`, `renderMethod`) gets an entry in the table and a term URL assigned to it. If we get enough implementations by the end of CR, we create a definition for the extension point in the specification (just as there is one for `evidence`, `termsOfUse`, and `refreshService` now) by lifting it's definition from the external spec that defines the extension point. Example for [renderMethod](https://digitalbazaar.github.io/vc-render-method/#the-rendermethod-property).

> Also the current description of confidence method in this PR is different from the description in PR https://github.com/w3c/vc-data-model/pull/1054

We can update the description to match #1054.

index.html
TallTed: Except that they *don't* do exactly the same thing. The `cnf` claim (and, I think, the related `confirmationMethod`) is specifically tied to key possession, while the `confidenceMethod` is explicitly broader and non-specific, though it might *include* key possession.

OR13: @TallTed you are correct they don't do the same thing, and one of them is more dangerous because it has an unbound domain... https://en.wikipedia.org/wiki/Confidence_trick

index.html
```diff
@@ -3211,7 +3211,22 @@ <h3>Reserved Extension Points</h3>
               <th>Description</th>
             </tr>
           </thead>
+
           <tbody>
+            <tr>
+              <td>`confidenceMethod`</td>
+              <td>
+A property used for specifying how a <a>verifier</a> can gain more confidence
+in, the presentation of the credential being associated with a particular
```
David-Chadwick, line 3220: two suggestions. First, the comma after in is not needed. Second, 'being associated with a particular subject' is either ambiguous or unnecessary. All credentials are associated with a subject as they have a subject property, so does this mean that the presenter is associated with a particular subject or not?
I suggest simply remove 'being associated with a particular subject' or clarify to 'can gain more confidence
that the presenter of the credential is associated with the subject"

TallTed: Definitely, remove the extraneous comma.
```suggestion
in the presentation of the credential being associated with a particular
```

mprorock (approved):

index.html
mprorock, line 3217:
```suggestion
              <td>`confidenceMethod`</td>
               <td>`confirmationMethod`</td>
```

dlongley: -1 to bringing back `confirmationMethod`.

mprorock: how would you suggest we map `cnf` then?

dlongley: @mprorock,

> how would you suggest we map `cnf` then?

A specific, single-key-style proof-of-possession type of `confidenceMethod` (which someone will need to define) can be mapped to `cnf`. Other types clearly wouldn't map, for example:

```
{
  ...,
  "confidenceMethod": [{
    "type": "FooSingleKeyProofOfPossessionConfirmationThing",
    "cnf": { /* same kind of stuff that goes in 'cnf' }
  }, {
    "type": "DoesNotWorkWithCnf",
    ...
  }]
}
```

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-05-10-vcwg#section1-6) on 2023-05-10

- no resolutions were taken

#### 1.6. Add new v2.0 reserved properties. (pr vc-data-model#1108)

_See github pull request [vc-data-model#1108](https://github.com/w3c/vc-data-model/pull/1108)._




**Manu Sporny:** PR 1108 - Add new v2.0 reserved properties.  
… 4 objections with change requests from Orie, David Chandwick, Oliver and Kristina.  

> *Orie Steele:* FWIW i think [https://github.com/w3c/vc-data-model/pull/1107](https://github.com/w3c/vc-data-model/pull/1107) can be merged outside of call time, I was the only objector.

**Manu Sporny:** think Oliver and Kristina wanted to see confidence method merged before this PR is merged, which we're not doing today.  
… what are you thoughts on this Orie?  

**Orie Steele:** seems like render method would go in by itself. there's work going in the CCG and people would agree with this.  
… would approve render method and has presentation schema reservations.  

> *Samuel Smith:* In the academic literature the word confidence is extremely common in expert systems and automated reasoning to provide a measure of uncertainty on a given predicate.

> *Samuel Smith:* See [https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5378479/#:~:text=Nevertheless%2C%20there%20is%20a%20fundamental,commitment%20is%20overt%20or%20covert).](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5378479/#:~:text=Nevertheless%2C%20there%20is%20a%20fundamental,commitment%20is%20overt%20or%20covert).).

**Orie Steele:** today Orie is ok with the confidence method in the table. Not a fan of it in any format and prefer to have it reviewed for terms independently.  

**Manu Sporny:** refactor PR to only refer to render method - any objections to that?  

**Brent Zundel:** no objections heard.  

> *Shigeya Suzuki:* I share same thought on presentationSchema and renderMethod with Orie.

> *Orie Steele:* I would be a +1 to reserving render method, since their is a ccg spec which I can read that might eventually get 2 independent implementations.

> *Dave Longley:* SamSmith: (and that's a match for its usage in VCs).

> *Phillip Long:* Manu wil redo 1108 to only include render method.

</details>

msporny:
I have reworked this PR to only include `renderMethod`. This clears all registered objections with the PR, namely:

* Both @Sakurann and @awoie's request to merge #1054 first no longer applies, as #1054 is now blocked (by @OR13), and the property is no longer a part of this PR.
* @David-Chadwick's request to change `confidenceMethod` no longer applies due to the previous bullet point.
* The remaining property in this PR, `renderMethod` did not receive any objections in this PR, nor during a call for objections during the last call.
* @OR13 and @mprorock confirmed that they will not object to a `renderMethod`-only PR being merged.

Therefore:

Sakurann:
I am not ok with this being merged over my request for changes, saying that the PR #1054 is blocked. Please revert. 

I am against merging properties that were not in v1.1 as extension points like this without much clarity what it means, especially given the direction PR #1054 is taking.

Issue #981, opened 2022-11-20
ManualRefreshService2018 - what exactly is it?


AFlowOfCode:
I hope it is not incorrect to ask this here - if so feel free to delete/close & I'll ask on one of the mailing lists.

The only information I can find anywhere is in the [vocabulary](https://www.w3.org/2018/credentials/): A type of refresh service that must be interacted with in a manual fashion.

However the specificity of the term (particularly the inclusion of the year) to me suggests that this type of refresh service may have a specification hidden somewhere away from search engines. The above definition is extremely open-ended and causes me to wonder why just setting the value to `"manual"` couldn't suffice.

So to rephrase the title of this issue, does `ManualRefreshService2018` imply anything beyond the meaning of the word "manual" (in contrast to "automatic")? I'd like to make sure I'm not misusing this term if I include it in reference to a custom method of refreshing a credential.

Thank you.

OR13:
I suggest this example be removed until such time as there is a TR or equivalent level of standard to refer to.

This issue should be converted to a request to remove the example.

msporny:
We could also point to the CCG refresh service mechanism, which would be an editorial change.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-01-17-vcwg#section2-2) on 2024-01-17

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 2.2. ManualRefreshService2018 - what exactly is it? (issue vc-data-model#981)

_See github issue [vc-data-model#981](https://github.com/w3c/vc-data-model/issues/981)._




**Brent Zundel:** This is assigned to you, manu, is this still an issue?  

**Manu Sporny:** No, unfortunately, it's still in the spec. We do have a CCG credential refresh spec that we can point to.  
… I know there are two implementations of the CCG credential refresh mechanism in production around the TruAge program.  
… The proposal here would be to use that value because that value is in production, would that be enough to address this issue?  

**Brent Zundel:** Refresh service is a reserved term, not defined?  

> *Manu Sporny:* See [term definition](https://w3c.github.io/vc-data-model/#refreshing).

**Manu Sporny:** We have a section for it, it's marked at-risk.  

**Brent Zundel:** Unless we can get an equivalent set of specs, then we're going to remove it?  

[...]

PR #1113, opened 2023-05-04
Remove reserved concept of `refreshService` from the spec

OR13:
Now that it has been reserved, we can remove the section from the spec and close https://github.com/w3c/vc-data-model/issues/981

Alternatively, we can retain the section and remove `refreshService` from the "reserved properties" table.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1113.html" title="Last updated on May 4, 2023, 10:19 PM UTC (146f792)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1113/d167254...146f792.html" title="Last updated on May 4, 2023, 10:19 PM UTC (146f792)">Diff</a>

### DIFF
--- a/contexts/credentials/v2
+++ b/contexts/credentials/v2
@@ -53,10 +53,6 @@
           "@type": "@id",
           "@container": "@graph"
         },
-        "refreshService": {
-          "@id": "https://www.w3.org/2018/credentials#refreshService",
-          "@type": "@id"
-        },
         "termsOfUse": {
           "@id": "https://www.w3.org/2018/credentials#termsOfUse",
           "@type": "@id"
--- a/index.html
+++ b/index.html
@@ -2369,7 +2369,7 @@ documents with implementations, or enabling aggressive caching of contexts.
         <p>
 Implementers are advised to pay close attention to the extension points in this
 specification, such as in Sections <a href="#proofs-signatures"></a>,
-<a href="#status"></a>, <a href="#data-schemas"></a>,<a href="#refreshing"></a>,
+<a href="#status"></a>, <a href="#data-schemas"></a>,
 <a href="#terms-of-use"></a>, and <a href="#evidence"></a>. While this
 specification does not define concrete implementations for those extension
 points, the Verifiable Credential Specifications Directory [[?VC-SPECS]]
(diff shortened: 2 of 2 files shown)


brentzundel (changes requested):
This PR should also include a mention in the Revision History section about the change.

msporny (changes requested):
> Alternatively, we can retain the section and remove `refreshService` from the "reserved properties" table.
> 
> [Preview](https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1113.html) | [Diff](https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1113/d167254...146f792.html)

Yeah, we're in a weird place w/ the "legacy but reserved properties". There are a few options here:

1. Mark them as at risk (in the section they're about), and to be removed by the end of CR, if there are not at least two implementations for a single type. Don't add them to the reserved properties table.
2. Do item 1 above, AND add them to the reserved properties table and say that the property will remain in the reserved properties table if there are not at least two implementations for a single type. If we do this, we just end up deleting either the section, or the entry in the reserved properties table. That is, we just remove text instead of having to shift text (which sometimes raises eyebrows during CR (and then you have to explain why you're not trying to pull something over on the W3C, that this was an intentional addition to the spec).
3. Do what's being done in this PR, and delete the section, and add it to the reserved properties table.

If we do item 3, we are (possibly) reducing the chances that it gets implemented because the spec doesn't say what to implement. One could argue: "Wait a second, you're expecting people to implement something that's not spec'd?" -- the counter-argument to that could be: "Well, it's spec'd in v1.1 -- do that." -- that whole exchange is a bit shaky on both sides of the exchange.

If we do item 1, then we have to add text after CR, which always requires quite a bit of explanation as to why you're making a normative change so late in the process.

If we do item 2, we just delete something that we warned people we might delete during CR. This, from a process perspective, feels the easiest to me.

Based on that, -1 to merging this PR for two reasons: 1) I'd prefer we do item 2 above (from a process perspective), and 2) I already know that there are at least two independent implementations of this feature -- TruAge, and a digital wallet vendor that can probably speak more openly about this in a couple of months.

decentralgabe:
+1 @msporny I prefer an "at risk" approach

Issue #946, opened 2022-10-11
what is attribute ID  in the verifiable credential?


Muthurajj:
1.what is the purpose of the id ?
2.How to create the id url?
3.If the id is refer to identity for the claims in VC then we have storage the claims in server to identify gobally?

OR13:
`id` must be an IRI, see the spec:

- https://www.w3.org/TR/vc-data-model/#identifiers
- https://webmasters.stackexchange.com/questions/98569/what-is-the-use-of-id-in-json-ld-syntax


`@id` is aliased it `id` in https://www.w3.org/2018/credentials/v1

msporny:
We should add text that provides guidance around what an "id" is, whether or not it is required, what is expected to be there, must it be globally unique, there are cases where there isn't an ID for a VC, etc.

OR13:
related issue: https://github.com/w3c/vc-data-model/issues/1031

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2022-09-15-vcwg#section5-5) on 2022-09-15

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 5.5. what is attribute ID  in the verifiable credential? (issue vc-data-model#946)

_See github issue [vc-data-model#946](https://github.com/w3c/vc-data-model/issues/946)._




**Kristina Yasuda:** This needs clarification in the spec.  
… Is anyone willing to write the clarifying text.  

**Manu Sporny:** I can try.  

**Orie Steele:** He's asking about what you should expect it to be and how should you handle it.  
… This is related to the URL vs. URI confusion.  

**David Chadwick:** I have an issue with the "parts of" wording.  

> *Dmitri Zagidulin:* (the question that almost all devs new to VCs ask is - 1) is 'id' required? 2) is it /supposed/ to be globally unique? Or can I use like a database row # for the id?.

**David Chadwick:** The credential schema should be one thing.  
… There's a disagreement there.  
… I don't think there's agreement that Gabe's proposal should be accepted.  

**Dmitri Zagidulin:** Maybe this text belongs more in the Implementation Guide.  
… I think we want to add it in the spec.  
… Is it required and must it be globally unique?.  
… We need guidance to say that it's not required and that it should be globally unique.  
… If absent, use the signature string as an ID.  

**Joe Andrieu:** We call them URIs. We refer to RFC 3986..  

[...]

Issue #60, opened 2017-07-18
What would a counter-signed RWoT claim look like?


kimdhamilton:
_From @ChristopherA on July 14, 2017 22:43_

Using my DID (the latest value from Shannon's code), and signing a verifiable claim (created from the verifiable claims playground and signed with my correct testnet private key I used to create the DID transaction) I get:

```
{
  "@context": [
    "https://schema.org/",
    "https://w3id.org/security/v1"
  ],
  "claim": {
    "id": "did:btcr:xyv2-xzyq-qqm5-tyke",
    "alternate-name": "ChristopherA",
    "knows": {
      "id": "did:btcr:rzqq-qqqq-uvlj-ez",
      "alternate-name": "Alice"
    }
  },
  "signature": {
    "type": "EcdsaKoblitzSignature2016",
    "created": "2017-07-14T22:38:36Z",
    "creator": "ecdsa-koblitz-pubkey:02b97c30de767f084ce3080168ee293053ba33b235d7116a3263d29f1450936b71",
    "signatureValue": "IGbDlNVSLIk7//rnuEEw7vtRKQ1HFLb+/K0cFHZ80zcVbFkF6vQj+jtDANc0MD9o0nhL5ykGFTd1VoP7a7B2iqM="
  }
}
```

This should be (hopefully and eventually) inspectable, validatable, verifiable and confirmable.

However, under the BTCR method, we are extremely self-sovereign centric, so this claim is not considered to be valid until it is countersigned by Alice.

@dlongley What would that extra countersignature look like?

_Copied from original issue: WebOfTrustInfo/btcr-hackathon#45_

kimdhamilton:
I think the signatures accumulate in a single array at the end, like this. But @dlongley would know better.

https://github.com/WebOfTrustInfo/btcr-did-tools-js/blob/master/claims/kimh-countersigned.jsonld

msporny:
@kimdhamilton @ChristopherA Multisig for LDS is defined here: https://w3c-dvcg.github.io/ld-signatures/#multiple-signatures

There are two options... set signatures and chained signatures. If the order of signatures matters, you should use chained signatures. Note that we have set signatures implemented, but chained signatures are not implemented and we should discuss some of the concerns before going that route.

In any case, hopefully the spec link above helps shed some light on the options available.

burnburn:
@kimdhamilton please take a look at this.  May be blocked same as #32.

kimdhamilton:
[...]

PR #1114, opened 2023-05-06
Add usage guidance related to the `id` property.

msporny:
This PR addresses issue #946 by describing what the `id` property is used for, some example values for it, and whether or not it should be optional.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1114.html" title="Last updated on May 13, 2023, 6:49 PM UTC (4af336e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1114/d167254...4af336e.html" title="Last updated on May 13, 2023, 6:49 PM UTC (4af336e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1176,7 +1176,7 @@ allows for the expression of statements about specific things in the
 objects in a <a>verifiable credential</a> or a <a>holder</a> when expressing
 objects in a <a>verifiable presentation</a>. Example <code>id</code> values
 include UUIDs (`urn:uuid:0c07c1ce-57cb-41af-bef2-1b932b986873`), HTTP URLs
-(`https://id.example/123`), and DIDs (`did:example:1234abcd`).
+(`https://id.example/things#123`), and DIDs (`did:example:1234abcd`).
         </p>
 
         <p>
--- a/index.html
+++ b/index.html
@@ -1166,14 +1166,17 @@ This concept is further expanded on in Section <a href="#extensibility"></a>.
         <h3>Identifiers</h3>
 
         <p>
-When expressing statements about a specific thing, such as a person, product,
-or organization, it is often useful to use some kind of identifier so that
-others can express statements about the same thing. This specification defines
-the optional <code>id</code> <a>property</a> for such identifiers. The
-<code>id</code> <a>property</a> is intended to unambiguously refer to an object,
-such as a person, product, or organization. Using the <code>id</code>
-<a>property</a> allows for the expression of statements about specific things in
-the <a>verifiable credential</a>.
+When expressing statements about a specific thing, such as a person, product, or
+organization, it is often useful to use a globally unique identifier for it.
+Globally unique identifiers are useful so that others can express statements
+about the same thing. This specification defines the optional <code>id</code>
+<a>property</a> for such identifiers. Using the <code>id</code> <a>property</a>
+allows for the expression of statements about specific things in the
+<a>verifiable credential</a> and is set by an <a>issuer</a> when creating
+objects in a <a>verifiable credential</a> or a <a>holder</a> when creating
+objects in a <a>verifiable presentation</a>. Examples of <code>id</code>s
+include UUIDs (`urn:uuid:0c07c1ce-57cb-41af-bef2-1b932b986873`), HTTP URLs
+(`https://id.example/123`), and DIDs (`did:example:1234abcd`).
         </p>
 
         <p>
@@ -1202,7 +1205,8 @@ where pseudonymity is required. Developers are encouraged to read Section
 scenarios. There are also other types of correlation mechanisms documented in
 Section <a href="#privacy-considerations"></a> that create privacy concerns.
 Where privacy is a strong consideration, the <code>id</code> <a>property</a>
-MAY be omitted.
+MAY be omitted. Not all use cases require the usage of the <code>id</code>
+<a>property</a>.
         </p>
 
         <dl>
--- a/index.html
+++ b/index.html
@@ -1167,14 +1167,14 @@ This concept is further expanded on in Section <a href="#extensibility"></a>.
 
         <p>
 When expressing statements about a specific thing, such as a person, product, or
-organization, it is often useful to use a globally unique identifier for it.
-Globally unique identifiers are useful so that others can express statements
+organization, it can be useful to use a globally unique identifier for that thing.
+Globally unique identifiers enable others to express statements
 about the same thing. This specification defines the optional <code>id</code>
-<a>property</a> for such identifiers. Using the <code>id</code> <a>property</a>
+<a>property</a> for such identifiers. The <code>id</code> <a>property</a>
 allows for the expression of statements about specific things in the
-<a>verifiable credential</a> and is set by an <a>issuer</a> when creating
-objects in a <a>verifiable credential</a> or a <a>holder</a> when creating
-objects in a <a>verifiable presentation</a>. Examples of <code>id</code>s
+<a>verifiable credential</a> and is set by an <a>issuer</a> when expressing
+objects in a <a>verifiable credential</a> or a <a>holder</a> when expressing
+objects in a <a>verifiable presentation</a>. Example <code>id</code> values
 include UUIDs (`urn:uuid:0c07c1ce-57cb-41af-bef2-1b932b986873`), HTTP URLs
 (`https://id.example/123`), and DIDs (`did:example:1234abcd`).
         </p>
@@ -1205,7 +1205,7 @@ where pseudonymity is required. Developers are encouraged to read Section
 scenarios. There are also other types of correlation mechanisms documented in
 Section <a href="#privacy-considerations"></a> that create privacy concerns.
 Where privacy is a strong consideration, the <code>id</code> <a>property</a>
-MAY be omitted. Not all use cases require the usage of the <code>id</code>
+MAY be omitted. Some use cases do not require, or explicitly require omitting, the <code>id</code>
 <a>property</a>.
         </p>
 


index.html
```diff
@@ -1166,14 +1166,17 @@ <h3>Contexts</h3>
...
-When expressing statements about a specific thing, such as a person, product,
-or organization, it is often useful to use some kind of identifier so that
-others can express statements about the same thing. This specification defines
-the optional <code>id</code> <a>property</a> for such identifiers. The
-<code>id</code> <a>property</a> is intended to unambiguously refer to an object,
-such as a person, product, or organization. Using the <code>id</code>
-<a>property</a> allows for the expression of statements about specific things in
-the <a>verifiable credential</a>.
+When expressing statements about a specific thing, such as a person, product, or
+organization, it is often useful to use a globally unique identifier for it.
```
dlongley, line 1170:
```suggestion
organization, it can be useful to use a globally unique identifier for it.
```

index.html
```diff
@@ -1166,14 +1166,17 @@ <h3>Contexts</h3>
...
+Globally unique identifiers are useful so that others can express statements
+about the same thing. This specification defines the optional <code>id</code>
+<a>property</a> for such identifiers. Using the <code>id</code> <a>property</a>
+allows for the expression of statements about specific things in the
+<a>verifiable credential</a> and is set by an <a>issuer</a> when creating
+objects in a <a>verifiable credential</a> or a <a>holder</a> when creating
```
dlongley, line 1176:
```suggestion
<a>verifiable credential</a> and is set by an <a>issuer</a> when expressing
objects in a <a>verifiable credential</a> or a <a>holder</a> when expressing
```

dlongley (approved):

index.html
```diff
@@ -1166,14 +1166,17 @@ <h3>Contexts</h3>
...
+objects in a <a>verifiable presentation</a>. Examples of <code>id</code>s
+include UUIDs (`urn:uuid:0c07c1ce-57cb-41af-bef2-1b932b986873`), HTTP URLs
+(`https://id.example/123`), and DIDs (`did:example:1234abcd`).
```
melvincarvalho, line 1179: While the current text in the pull request suggests using HTTP URLs as identifiers (e.g., `https://id.example/123`), I would like to argue that using hash URIs with a fragment identifier is a more appropriate approach for identifying a person, place, or thing.

The W3C has addressed this topic in a TAG resolution called HR14 (Hash In URI). The resolution suggests that HTTP URIs without fragment identifiers should be used for documents, while hash URIs with fragment identifiers should be used for identifying resources other than documents, such as people, places, or things. You can find more information about this resolution in the following W3C document: https://www.w3.org/2001/tag/doc/hash-in-uri

The main rationale behind using hash URIs with fragment identifiers is that it allows for better separation between the document that describes the resource and the resource itself. This approach helps in avoiding confusion and allows for more precise statements about the resource.

Considering this, I suggest updating the example in the pull request to include a hash URI with a fragment identifier for identifying a person, place, or thing, such as `https://example.com/people#alice`. This way, the identifier clearly points to a non-document resource, in line with the W3C HR14 resolution.

Here's a revised version of the text:

Globally unique identifiers are useful so that others can express statements about the same thing. This specification defines the optional `id` property for such identifiers. Using the `id` property allows for the expression of statements about specific things in the verifiable credential and is set by an issuer when creating objects in a verifiable credential or a holder when creating objects in a verifiable presentation. Examples of `id`s include UUIDs (`urn:uuid:0c07c1ce-57cb-41af-bef2-1b932b986873`), HTTP URLs with fragment identifiers (`https://example.com/people#alice`), and DIDs (`did:example:1234abcd`).

index.html
TallTed:
```suggestion
organization, it can be useful to use a globally unique identifier for that thing.
```

index.html
```diff
@@ -1166,14 +1166,17 @@ <h3>Contexts</h3>
...
-or organization, it is often useful to use some kind of identifier so that
-others can express statements about the same thing. This specification defines
-the optional <code>id</code> <a>property</a> for such identifiers. The
-<code>id</code> <a>property</a> is intended to unambiguously refer to an object,
-such as a person, product, or organization. Using the <code>id</code>
-<a>property</a> allows for the expression of statements about specific things in
-the <a>verifiable credential</a>.
+When expressing statements about a specific thing, such as a person, product, or
+organization, it is often useful to use a globally unique identifier for it.
+Globally unique identifiers are useful so that others can express statements
```
TallTed, line 1171:
```suggestion
Globally unique identifiers enable others to express statements
```

index.html
```diff
@@ -1166,14 +1166,17 @@ <h3>Contexts</h3>
...
+about the same thing. This specification defines the optional <code>id</code>
+<a>property</a> for such identifiers. Using the <code>id</code> <a>property</a>
```
TallTed, line 1173:
```suggestion
<a>property</a> for such identifiers. The <code>id</code> <a>property</a>
```

index.html
```diff
@@ -1202,7 +1205,8 @@ <h3>Identifiers</h3>
 scenarios. There are also other types of correlation mechanisms documented in
 Section <a href="#privacy-considerations"></a> that create privacy concerns.
 Where privacy is a strong consideration, the <code>id</code> <a>property</a>
-MAY be omitted.
+MAY be omitted. Not all use cases require the usage of the <code>id</code>
```
TallTed, line 1208:
```suggestion
MAY be omitted. Some use cases do not require use of the <code>id</code>
```

TallTed (changes requested):
Minor tweaks requested above. I'm not against @melvincarvalho's suggested paragraph revision, but my tweaks to the existing paragraph should also be applied to his paragraph, if it's merged into the text.

index.html
TallTed: @melvincarvalho — I am somewhat concerned that [the referenced W3C document](https://www.w3.org/2001/tag/doc/hash-in-uri) has internal links to ["This version"](http://www.w3.org/TR/2009/WD-hash-in-url-20090415/) and ["Latest Version"](http://www.w3.org/TR/hash-in-url/), both in `www.w3.org/TR/`, but both `404`. I also see that the _Status of this Document_ is that it was "published as a Public Working Draft" in 2009 and hasn't been updated since. This significantly lowers my inclination to follow its guidance.

decentralgabe (approved):
generally agree, though I don't see much of an issue with local uniqueness under an issuer as a namespace. better than it was, so approving.

OR13 (approved):
This is an improvement over the existing text, but I don't see why keeping `id` optional while requiring JSON-LD is an improvement over JSON Web Tokens.

Because JSON-LD is automatically assigning non globally unique `id` values (blank node identifiers) when a globally unique one is omitted.... which is worse, than requiring a unique `id` value.

index.html
brentzundel, line 1208:
```suggestion
MAY be omitted. Some use cases require omitting the <code>id</code>
```

mprorock (approved):

index.html
msporny, line 1208:
```suggestion
MAY be omitted. Some use cases do not require, or explicitly require omitting, the <code>id</code>
```

index.html
```diff
@@ -1166,14 +1166,17 @@ <h3>Contexts</h3>
...
+organization, it can be useful to use a globally unique identifier for that thing.
+Globally unique identifiers enable others to express statements
+about the same thing. This specification defines the optional <code>id</code>
+<a>property</a> for such identifiers. The <code>id</code> <a>property</a>
+allows for the expression of statements about specific things in the
+<a>verifiable credential</a> and is set by an <a>issuer</a> when expressing
+objects in a <a>verifiable credential</a> or a <a>holder</a> when expressing
+objects in a <a>verifiable presentation</a>. Example <code>id</code> values
+include UUIDs (`urn:uuid:0c07c1ce-57cb-41af-bef2-1b932b986873`), HTTP URLs
+(`https://id.example/123`), and DIDs (`did:example:1234abcd`).
```
msporny, line 1179:
```suggestion
(`https://id.example/people#alice`), and DIDs (`did:example:1234abcd`).
```

@TallTed I'm making the change above in order to avoid another HTTP Range-14 perma-thread. :)

msporny:
@decentralgabe wrote:
> generally agree, though I don't see much of an issue with local uniqueness under an issuer as a namespace. 

If the issuer namespace is global, then local uniqueness is global uniqueness.

If the issuer namespace can conflict with another issuer's namespace, such as, the namespace is literally the empty string... then, the issue is `id` conflicts if random identifiers in a sufficiently large namespace aren't utilized.

msporny:
> This is an improvement over the existing text, but I don't see why keeping `id` optional while requiring JSON-LD is an improvement over JSON Web Tokens.

It's an improvement because when `id` is optional, and not specified, then the internal identifier utilized is scoped to the document (it has document-scoped semantics). JWT, and  JSON, leaves the semantics for identifiers undefined while at the same time some libraries suggest equivalence if the contents of an object are byte-wise the same as in another document.

IOW, the identifier story for JSON-LD (and RDF) for unlabeled nodes is a well understood and well defined thing... it is completely undefined for JSON and JWT.

> Because JSON-LD is automatically assigning non globally unique `id` values (blank node identifiers) when a globally unique one is omitted.... which is worse, than requiring a unique `id` value.

You've stated that something is worse without supporting why it is worse with a well-reasoned argument. Why is it worse?

msporny:

index.html
melvincarvalho: Thank you, @TallTed, for your feedback. I understand your reservations given the accessibility issues with the referenced W3C document and its dated status. The 404 links have now been fixed.  I believe the essence of the TAG resolution encapsulates remains relevant, particularly in the context of our work here.

The resolution outlines a distinction between identifiers for documents versus non-document resources, a principle that I believe could contribute to the clarity and accuracy of our identifiers. While the document itself may no longer be maintained, the principles it outlines remain sound and adopted in various contemporary practices.

That said, I would encourage aligning with the TAG finding, particularly the use of hash URIs with fragment identifiers for non-document resources, despite the age of the document. After all, the longevity of these principles attests to their utility and relevance.

TallTed: @melvincarvalho — Please note that while the linked document now has different (working) "This version" and "Latest Version" targets, neither is accurate. Note that the Abstract says, "This draft finding is being prepared in response to [TAG issue #60](http://www.w3.org/2001/tag/group/track/issues/60)." (That would certainly have been better noted in the SotD.)

Going to that TAG issue, I find that while it links to the above document, it also says that (emphasis mine), "_Current_ work is: [Repurposing the Hash Sign for the New Web; W3C Working Draft: Putative TAG Finding 28 February 2011](http://www.w3.org/2001/tag/2011/02/HashInURI-20110228)" which is two years more recent, but still outdated.

The first exterior link found in both as an example, in [Addressing Into Multimedia Streams -- CNN](https://www.w3.org/2001/tag/2011/02/HashInURI-20110228.html#MultimediaStreamsCNN), is `http://www.cnn.com/video/#/video/tech/2008/02/19/vo.aus.sea.spider.ap` which was probably accurate in 2011, but it is not today, as that link gets redirected to the fragmentless `https://www.cnn.com/videos/tech/2008/02/19/vo.aus.sea.spider.ap` (which is also now a 404).

I submit that the DRAFT TAG finding should be more carefully considered and potentially updated, given the changes in the wild.

melvincarvalho: @TallTed , your comments seem to divert from the core issue at hand. It's crucial to focus on the present topic and not get lost in the minutiae of details from documents. I respectfully suggest that you update your understanding with the reading materials before making assertions. This will ensure we're all on the same page, and that our discussions are productive and relevant. Let's work together to address the important issues at hand rather than getting sidetracked.

melvincarvalho:
While I am not raising any objections at this moment, I am concerned that the topic about the distinction between documents and things might have been overlooked amidst other discussions. It's worth noting that the TAG has promptly addressed the broken link issue.

Refer: https://lists.w3.org/Archives/Public/www-tag/2023May/0001.html

Although I could make further corrections based on @TallTed's suggestions, the core content of the document remains unchanged. I interact with developers daily who often confuse between things and documents on the web. Examples in REC track work may inadvertently solidify this misunderstanding, leading developers to believe that pointers and contents are interchangeable. This could potentially act as a significant barrier to semantic interoperability, and might be worth reconsidering in future if the goal is to produce REC quality examples.

TallTed:
@melvincarvalho — This confusion/belief "that pointers and contents are interchangeable" is long-running and is frequently referred to as [HTTPRange14](https://en.wikipedia.org/wiki/HTTPRange-14). Programmers _should_ understand the difference, as they work with memory pointers and the values found there, but somehow the introduction of HTTP tends to twist many minds.

melvincarvalho:
@TallTed well said!  Ends.

PR #1119, opened 2023-05-09
Remove Blockchain

OR13:
Seems there is no need to comment on proof types which are not supported in the vc specs directory or the core spec, and which unnecessarily jeopardize the work item to formal objections.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1119.html" title="Last updated on May 16, 2023, 2:15 PM UTC (bd5b90a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1119/80cefee...bd5b90a.html" title="Last updated on May 16, 2023, 2:15 PM UTC (bd5b90a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4711,7 +4711,7 @@ the validity period for a <a>verifiable credential</a> is not in the future.
 The cryptographic mechanism used to prove that the information in a
 <a>verifiable credential</a> or <a>verifiable presentation</a> was not
 tampered with is called a <em>proof</em>. There are many types of cryptographic
-proofs including, but not limited to, digital signatures, zero-knowledge proofs. 
+proofs including, but not limited to, digital signatures and zero-knowledge proofs. 
 In general, when verifying proofs, implementations are expected to ensure:
         </p>
 
--- a/index.html
+++ b/index.html
@@ -4711,9 +4711,8 @@ the validity period for a <a>verifiable credential</a> is not in the future.
 The cryptographic mechanism used to prove that the information in a
 <a>verifiable credential</a> or <a>verifiable presentation</a> was not
 tampered with is called a <em>proof</em>. There are many types of cryptographic
-proofs including, but not limited to, digital signatures, zero-knowledge proofs,
-Proofs of Work, and Proofs of Stake. In general, when verifying proofs,
-implementations are expected to ensure:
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -4711,9 +4711,8 @@ <h3>Proofs (Signatures)</h3>
 The cryptographic mechanism used to prove that the information in a
 <a>verifiable credential</a> or <a>verifiable presentation</a> was not
 tampered with is called a <em>proof</em>. There are many types of cryptographic
-proofs including, but not limited to, digital signatures, zero-knowledge proofs,
-Proofs of Work, and Proofs of Stake. In general, when verifying proofs,
-implementations are expected to ensure:
+proofs including, but not limited to, digital signatures, zero-knowledge proofs.
```
OR13, line 4714:
```suggestion
proofs including, but not limited to, digital signatures & zero-knowledge proofs. 
```

OR13: Noting that we have yet to document any credential proof formats that use zero knowledge proofs.

decentralgabe (approved):

mprorock (approved):

msporny (approved):

msporny (approved):
I'm a bit lukewarm on this PR. The strongest argument for it seems to be for political reasons ("people might formally object because they're conflating PoW and PoS with the word 'blockchain'! Those people are having a visceral reaction to one technology and using that as reason to FO this spec -- even though this spec mentions the technology non-normatively -- thus, the FO won't hold"). That said, those political reasons do exist... though I doubt changing this sentence is going to save us from FOs from people that conflate "all types of PoW" (e.g. PoW used in captchas vs. blockchain PoWs).

dlongley (approved):
Can live with this -- we could have also mentioned "timestamp proofs" or something to avoid the blockchain politics but to also indicate other classes of proofs in this informative section. It may be a good idea to do so given that, from time to time, people forget that other proof classes are possible and this can lead to design trouble. Of course, "but not limited to" already covers this, just less explicitly so.

msporny:
Editorial, multiple reviews, changes suggested but not made, no objections, merging.

PR #1127, opened 2023-05-09
Add issue markers

OR13:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1127.html" title="Last updated on May 9, 2023, 11:41 PM UTC (36b1ab3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1127/80cefee...36b1ab3.html" title="Last updated on May 9, 2023, 11:41 PM UTC (36b1ab3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1336,7 +1336,7 @@ MUST have a <a>type</a> specified.
               <td>
 <code>VerifiableCredential</code> and, optionally, a more specific
 <a>verifiable credential</a> <a>type</a>. For example,<br>
-<code>"type": ["VerifiableCredential", "UniversityDegreeCredential"]</code>
+<code>"type": ["VerifiableCredential", "IdentityCredential"]</code>
               </td>
             </tr>
 
@@ -1347,10 +1347,10 @@ MUST have a <a>type</a> specified.
               <td>
 <code>VerifiableCredential</code> and, optionally, a more specific
 <a>verifiable credential</a> <a>type</a>. For example,<br>
-<code>"type": ["VerifiableCredential", "UniversityDegreeCredential"]</code>
+<code>"type": ["VerifiableCredential", "IdentityCredential"]</code>
               </td>
             </tr>
-
+<p class="issue" data-number="1126"></p>
             <tr>
               <td>
 <a>Verifiable presentation</a>&nbsp;object <br />(a subclass of a
@@ -1360,6 +1360,7 @@ MUST have a <a>type</a> specified.
 <code>VerifiablePresentation</code> and, optionally, a more specific
 <a>verifiable presentation</a> <a>type</a>. For example,<br>
 <code>"type": ["VerifiablePresentation", "CredentialManagerPresentation"]</code>
+<p class="issue" data-number="1125"></p>
               </td>
             </tr>
 
@@ -1371,6 +1372,7 @@ MUST have a <a>type</a> specified.
 <code>VerifiablePresentation</code> and, optionally, a more specific
 <a>verifiable presentation</a> <a>type</a>. For example,<br>
 <code>"type": ["VerifiablePresentation", "CredentialManagerPresentation"]</code>
+<p class="issue" data-number="1125"></p>
               </td>
             </tr>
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1336,7 +1336,7 @@ <h3>Types</h3>
               <td>
 <code>VerifiableCredential</code> and, optionally, a more specific
 <a>verifiable credential</a> <a>type</a>. For example,<br>
-<code>"type": ["VerifiableCredential", "UniversityDegreeCredential"]</code>
+<code>"type": ["VerifiableCredential", "IdentityCredential"]</code>
```
OR13, line 1339: I made this more neutral, since our spec does not define `UniversityDegreeCredential`, the example here is misleading.

Obviously since the context is omitted, this term is expected to be defined as:

https://www.w3.org/ns/credentials/issuer-dependent#IdentityCredential

OR13: I can revert this change if we get stuck on this attempt to improve this currently invalid example.

brentzundel (approved):

mprorock (approved):

index.html
msporny: `IdentityCredential` is going to confuse people given the charter in progress for the "Identity Credentials" work by Google, Apple, etc. We should avoid that terminology for two reasons:

1. It will get confused by the upcoming W3C WG that will be similarly named.
2. "Identity" is a lightning rod word that we should avoid in the specification, especially in examples.
3. Using an issuer-dependent term as the core example is not something that will achieve consensus. Continuing to raise PRs that use issuer-dependent terms in core examples will continue to be objected to.

Alternate suggestion: `ExampleCredential` ... but, honestly, we have an examples v2 context now... we should just use it for the examples in the specification.

index.html
```diff
@@ -1390,7 +1393,9 @@ <h3>Types</h3>
               </td>
               <td>
 A valid <a>credential</a> status <a>type</a>. For example,<br>
-<code>"type": "StatusList2021Entry"</code>
+<code>"type": "StatusList2021Entry"</code>.
+
+<p>See <a href="https://www.w3.org/TR/vc-status-list/">vc-status-list</a>.</p>
```
msporny, line 1398: This should be an informative ref to the status list spec.

index.html
```diff
@@ -1401,6 +1406,8 @@ <h3>Types</h3>
               <td>
 A valid terms of use <a>type</a>. For example,<br>
 <code>"type": "OdrlPolicy2017"</code>)
+
+<p class="issue" data-number="1123"></p>
```
msporny, line 1410: Not sure that I'm a fan of adding issue markers throughout the spec. We have 65+ issues where we'd have to do that, only to go back through and remove them. It's unnecessary churn for the Editors (without good reason for the churn), and this isn't being done consistently, AFAICT, from the latest issue marker additions.

msporny (changes requested):
This PR isn't just adding issue markers, it's attempting to also change the core example in the specification to language that is going to be problematic given the new W3C WG being formed around identity credentials. The naming is going to create unnecessary conflict.

index.html
TallTed: It seems to me that issue markers are appropriate for significant publication milestones, like CRs. Less so, though arguable, for FPWDs. Far less so for ED-per-commit and/or daily-ED, though editor's prerogative probably holds.

Given that, I'd suggest that there be some rule of thumb that issue markers be added based on exceeding some number of comments and/or duration of discussion (I don't have a suggestion for either trigger), on the presumption that the issue is less likely to be closed quickly, and so more likely to persist beyond the next milestone publication and/or transition, which is likely to bring more unfamiliar readers for whom the issue markers are intended.

index.html
OR13: I think we are not a working group dedicated to the education vertical... and I find t he consistent bias towards academic use cases harmful to the standard... I hope we can find more neutral examples.

TallTed: Perhaps `ExampleAchievementCredential` which could include merit badges from scouting-like organizations; trophies/medals from athletic competitions; "top salesperson" from employer, professional org, etc.; and so on?

Issue #1121, opened 2023-05-09
Update diagrams to not use RSA, and generally be accurate


OR13:
Probably use `DataIntegrityProof` since it is the only proof format that is also a graph format.

- https://w3c.github.io/vc-data-model/#info-graph-vc

- https://w3c.github.io/vc-data-model/#info-graph-vp

msporny:
PR https://github.com/w3c/vc-data-model/pull/1129 has been raised to address this issue. Once PR #1129 has been merged, this issue will be closed.

Issue #967, opened 2022-10-28
In Example 1, `value` and `lang` are not aliased to `@value` and `@language`


pchampin:
... and therefore, the two schema:name's of the example university are empty objects, and rendered in RDF as blank nodes.
Example 1: 
* [in the spec](https://www.w3.org/TR/vc-data-model/#example-a-simple-example-of-a-verifiable-credential)
[...]

Issue #1126, opened 2023-05-09
Address "Credential" vs "VerifiableCredential"


OR13:
https://w3c.github.io/vc-data-model/#types

<img width="844" alt="Screen Shot 2023-05-09 at 6 31 46 PM" src="https://github.com/w3c/vc-data-model/assets/8295856/23f87cdc-b8cb-45ca-9f2a-f6ead787786a">

Table indicates these are separate `types`, but the working group has since resolved not to define them separately, and every property that is valid for `credential` is also valid for `verifiable credential`.

OR13:
See https://github.com/w3c/vc-data-model/pull/1057

awoie:
This means, we should remove "Credential" entirely from the VCDM? IMO, it is confusing to have both in the VCDM, especially since the securing mechanisms are no longer part of the VCDM.

OR13:
@awoie see https://github.com/w3c/vc-data-model/pull/1129

Data Integrity proofs are in the v2 context, and in all the examples, I think your assessment is not accurate... the current spec has defects in both directions of the argument.

awoie:
@OR13 I fully agree we should fix that.

David-Chadwick:
There is clearly a difference between an object with a proof, meaning that it is verifiable, and an object without a proof, meaning that it is not verifiable. Since we have both of these objects in our specifications then we should be clear which term we apply to each of these objects. We cannot use the same term to apply to two objects because this introduces ambiguity.
I thought we already had an agreement that the object without the proof was to be termed "credential" and the object with the proof was to be termed "verifiableCredential"?

OR13:
I agree with ted.

I think we are making a mistake in the current spec, referring to credential and verifiable credentials separately... 


Both can have proof and neither might verify, they are not useful terms for distinguishing attributes of the data model, since they support the exact same properties.

We should avoid "credential" and focus on "claims" and "verifiable credential", or we should drop the word "verifiable" from everything but the spec title, and just have claims and credentials...

I was rereading WebAuthN yesterday, and I think we have missed a huge opportunity to align better with it, due to our terminology coming from RDF instead of security vocabulary.

The minimal VerifiablePresentation is just a proof from an authenticator (signature from a credential). 

The difference is web authn credentials are bound to a single RP (2 party model) and VCDM is not like that, and has more serious privacy issues because of this.

Our concept of "holder binding / confidence method" is their concept of allowedCredentials but scoped again to only authentication.

[...]

Issue #1103, opened 2023-04-27
Vocabulary normative, context isn't?


iherman:
I was not present at the [special call on 2023-04-25](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-04-25-vcwg-special) but, by reading the minutes, I have the impression that one of the controversy comes from the status of the JSON-LD `@context` file, namely, whether it is normative or not. That file being normative seems to clash with the desire to provide a context file that developers can use without hassle in cases where the JSON-LD behaviour is necessary, ie, if the URDNA canonicalization is used. 

So here is a proposal that, though maybe controversial, may make the situation cleaner:

1. The JSON-LD `@context` file is _not_ normative. Its goal is, among other things, is to make the life of developers easy insofar as all the "usual" terms are there, mapping to the VCDM and the various security vocabulary terms. 
2. The VCDM vocabulary _is_ normative. To be a little bit more precise: _the [JSON-LD representation of the vocabulary](https://w3c.github.io/vc-data-model/vocab/credentials/v2/vocabulary.jsonld) is normative_ and should be copied (automatically) into the VCDM specification. The rule (which is already the case) is that the vocabulary contains a URL reference of _all_ terms defined by the VCDM _and only those_. 

That approach may cleanly separate the issues once and for all.  The `@context` file's role is "just" to provide a mapping from the (standard) VCDM terms in JSON to the (standard) Vocabulary URLs, making the Linked Data representation of the VCDM following the standard. The JSON-LD representation of the vocabulary gives a clean specification for all VCDM terms by providing its URL and some of the essential characteristics (type, range, domain). And it does it in a machine readable way, assuring that the VCDM also plays following the rules established for Linked Data.

(The same approach should be used for the Security vocabulary, with the particularity that the terms appearing in the security vocabulary may be defined by the Data Integrity spec _or_ one of the cryptosuite specs.)  

@msporny @dlongley @OR13 @decentralgabe

decentralgabe:
I like the proposal, @iherman. It makes sense to me and seems like a reasonable compromise. 
@msporny had mentioned on the call it's best not to make the context normative because it could be tougher to make changes if bugs are found. Would this be a problem if we make the vocabulary normative?

iherman:
> @msporny had mentioned on the call it's best not to make the context normative because it could be tougher to make changes if bugs are found. Would this be a problem if we make the vocabulary normative?

[...]

PR #1129, opened 2023-05-14
Modernize examples in specification

msporny:
This PR modernizes the examples in the specification by:

* Updating the old `RsaSignature2018` examples to use `DataIntegrityProof`.
* Updating the Alumni and Degree examples to use the examples/v2 context (and changing the type names to be more explicit about being examples).
* Removing the complex `@lang`/`@value` examples and using simpler text instead.

This PR addresses #967 and #1121.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1129.html" title="Last updated on May 20, 2023, 6:49 PM UTC (d57867a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1129/a57fd93...d57867a.html" title="Last updated on May 20, 2023, 6:49 PM UTC (d57867a)">Diff</a>


dlongley (approved):

iherman:
A comment on the "Getting Started" example, (Example 3 in §4.1). An attentive reader's reaction may be that the example is incorrect JSON-LD, because there is no definition for, e.g., `MyPrototypeCredential`. We _know_ this isn't true by virtue of the `@vocab` in the VCDM core `@context`, but the presence of that is not mentioned anywhere in the document. I realize we do not want to give a high level of publicity for it, because it may drive implementers to a wrong path, ie, misusing what is essentially a security measure, but maybe it is worth adding an editorial note in this paragraph mentioning this.

iherman:
Looking at the examples, there is still room for confusion for the user. The term definitions come from (at least) four different places:

- terms defined in this specification (e.g., `issuer`)
- terms defined in another specification in the "family" of specs published by this WG (e.g., `verificationMethod`, or `eddsa-2022`). We also know that, somewhat deceptively, the `proof` property is _not_ formally defined by this specification either, although it looks that way.
- terms taken over from external vocabularies (e.g., `name` coming from schema.org)
- example terms (e.g., `alumniOf`).

I know from my own experience that all this is extremely confusing, and makes it very difficult to understand the spec.

One way of handling this may be:

- outline what I just wrote in the spec
- (this is where it is relevant for the PR) systematically use some ways of differentiating among these categories (different color, usage of italics/bold, different fonts; I am not sure what is the best from an accessibility point of view).

---
I am fine if the comment is "Ivan, let us do this in a different PR". But I consider it imperative to do it at some point.

TallTed (commented):
The changes look OK, as far as they go. I share @iherman's concerns, and think there should be an issue opened for them if they are not addressed in this PR, so we don't forget to address them.

andresuribe87 (approved):

brentzundel (approved):

OR13 (changes requested):
These diagrams don't match the consensus regarding "proof" and "credential".

We had some really long threads about how "proof" is part of both "credential" and "verifiable credential"

These diagrams show proof as separate graph object.

This does correspond to what the JSON-LD context does when it is applied, but the context is not normative, so the diagrams are not reflective of normative requirements...  and they seem to be contradicting consensus on "anything can have a proof".

These diagrams might be better in vc-data-integrity.

Technically "proof" is defined inside of "VerifiableCredential" and "VerifiablePresentation": 

https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v2#L51

I am requesting the following changes:

1. Use "VerifiableCredential" not "Credential".
2. Add examples for JWT (or file an issue and link to tools so I can provide equivalent diagrams).

OR13:
Related issue regarding aligning informative examples, with normative requirements... https://github.com/w3c/vc-data-model/issues/1103#issuecomment-1547372227

OR13:
Related to the "Credential" vs "VerifiableCredential" part... https://github.com/w3c/vc-data-model/issues/1126#issuecomment-1549524438

OR13:
Related issue: https://github.com/w3c/vc-data-model/issues/881

Discussed on the call today.

msporny:
@OR13 wrote:
> I am requesting the following changes:
> 1. Use "VerifiableCredential" not "Credential".

This was done in https://github.com/w3c/vc-data-model/pull/1129/commits/6156f485c5b5f7f88ab0432bb2fc8d313cac52a1 and https://github.com/w3c/vc-data-model/pull/1129/commits/96920dd8966fa9c87efdb69520879231960c2bec.

> 2. Add examples for JWT (or file an issue and link to tools so I can provide equivalent diagrams).

The Editors of the VC-JWT specification can provide those. I've raised issue#1135 for them to process.

I believe this addresses both of your change requests @OR13, please re-review.

OR13:
See this PR which attempts to update the examples, and the related tooling currently being deployed... https://github.com/digitalbazaar/respec-vc/pull/8

I think it would be better to simply switch to mermaid at this point... but I can do that in a subsequent PR.

OR13 (approved):

OR13:
I agree with @iherman 's comments about the proof values, we expect to see `cryptosuite`, we don't expect to see nice human readable names for verification method.

It would be good to make the diagrams consistent, with the actual nquads.

longpd:
+1 to Ivan's suggestion for future attention, and the resolution to the naming issues Manu has responded to from Orie. If I were an approver I'd agree, too!

msporny:

PR #1137, opened 2023-05-23
Add section on schema validity

decentralgabe:
Also updated reference to the FPWD of the Credential Schema 2023 spec

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1137.html" title="Last updated on May 24, 2023, 3:43 PM UTC (0cb2dad)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1137/f26b43e...0cb2dad.html" title="Last updated on May 24, 2023, 3:43 PM UTC (0cb2dad)">Diff</a>

### DIFF
--- a/common.js
+++ b/common.js
@@ -118,12 +118,12 @@ var vcwg = {
       href: 'https://en.wikipedia.org/wiki/InterPlanetary_File_System',
       publisher: 'Wikipedia'
     },
-    'CREDENTIAL-SCHEMA-2022': {
-      title: 'Verifiable Credentials JSON Schema 2022',
-      href: 'https://w3c-ccg.github.io/vc-json-schemas/',
+    'CREDENTIAL-SCHEMA-2023': {
+      title: 'Verifiable Credentials JSON Schema 2023',
+      href: 'https://w3c.github.io/vc-json-schema/',
       authors: ['Gabe Cohen', 'Orie Steele'],
-      status: 'CG-DRAFT',
-      publisher: 'Credentials Community Group'
+      status: 'FPWD',
+      publisher: 'W3C Verifiable Credentials Working Group'
     },
     'JSON-LD': {
       title: 'JSON-LD 1.1: A JSON-based Serialization for Linked Data',
--- a/common.js
+++ b/common.js
@@ -120,7 +120,7 @@ var vcwg = {
     },
     'CREDENTIAL-SCHEMA-2023': {
       title: 'Verifiable Credentials JSON Schema 2023',
-      href: 'https://w3c.github.io/vc-json-schema/',
+      href: 'https://www.w3.org/TR/vc-json-schema/',
       authors: ['Gabe Cohen', 'Orie Steele'],
       status: 'FPWD',
       publisher: 'W3C Verifiable Credentials Working Group'
--- a/index.html
+++ b/index.html
@@ -4899,7 +4899,7 @@ If the <code>credentialSchema</code> property is available, the schema of a
 <a>verifiable credential</a> is expected to be evaluated by the <a>verifier</a>
 according to the <code>credentialSchema</code> <a>type</a> definition for the
 <a>verifiable credential</a> and the <a>verifier's</a> own schema evaluation
-criteria. For example, if the <a>credentialSchema</a>'s' <a>type</a> value is
+criteria. For example, if the <code>credentialSchema</code>'s <code>type</code> value is
 [[?CREDENTIAL-SCHEMA-2023]], then a <a>verifier</a> can ensure a credential's
 data is valid against the given JSON Schema.
         </p>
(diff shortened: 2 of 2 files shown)


common.js
```diff
@@ -118,12 +118,12 @@ var vcwg = {
       href: 'https://en.wikipedia.org/wiki/InterPlanetary_File_System',
       publisher: 'Wikipedia'
     },
-    'CREDENTIAL-SCHEMA-2022': {
-      title: 'Verifiable Credentials JSON Schema 2022',
-      href: 'https://w3c-ccg.github.io/vc-json-schemas/',
+    'CREDENTIAL-SCHEMA-2023': {
+      title: 'Verifiable Credentials JSON Schema 2023',
+      href: 'https://w3c.github.io/vc-json-schema/',
```
decentralgabe, line 123:
```suggestion
      href: 'https://www.w3.org/TR/vc-json-schema/',
```

index.html
```diff
@@ -4891,6 +4891,20 @@ <h3>Status</h3>
...
 
+      <section class="informative">
+        <h3>Schema</h3>
+
+        <p>
+If the <code>credentialSchema</code> property is available, the schema of a
+<a>verifiable credential</a> is expected to be evaluated by the <a>verifier</a>
+according to the <code>credentialSchema</code> <a>type</a> definition for the
+<a>verifiable credential</a> and the <a>verifier's</a> own schema evaluation
+criteria. For example, if the <a>credentialSchema</a>'s' <a>type</a> value is
```
msporny, line 4902:
```suggestion
criteria. For example, if the <a>credentialSchema</a> <a>type</a> value is
```

index.html
```diff
@@ -4891,6 +4891,20 @@ <h3>Status</h3>
         </p>
       </section>
 
+      <section class="informative">
+        <h3>Schema</h3>
+
+        <p>
+If the <code>credentialSchema</code> property is available, the schema of a
+<a>verifiable credential</a> is expected to be evaluated by the <a>verifier</a>
```
msporny, line 4899:
```suggestion
<a>verifiable credential</a> can be evaluated by the <a>verifier</a>
```

index.html
msporny, line 4902:
```suggestion
criteria. For example, if the <code>credentialSchema</code>'s' <a>type</a> value is
```

msporny: You're switching between `<code></code>` markup, and links to sections normatively defining the term, and it's not clear why you're doing that. Try to be consistent. If you feel like you need to refer to the sections on type, use the `<a></a>` setup, if you just want to refer to the property used in the document, use the `<code></code>` markup (with a fair warning that we're moving from `<code></code>` to ```` markup (the latter being ReSpec syntactic sugar for the former).

index.html
msporny: (Company representative hat on): Our software will probably not utilize `credentialSchema` in a significant number of use cases -- the reasoning being that the issuer wouldn't have issued an VC with an invalid schema, or they're doing bad schema management and we don't want `credentialSchema` mismatches to get in the way of the VC being utilized. IOW, I'm not sure that we're positive or negative on this feature yet and therefore do not want any expectations set on verifier's using this feature. It's fine if it's there, it's fine if it's optional, so the request here is to not set any expectations that the verifier is expected to use the schema property (unless they want to).

msporny (approved):
+1 on this PR as long as one mandatory change is made: Do not set an expectation that Verifiers will use this property, it's optional. The other changes are minor editorial changes, but ones that need to be made in a consistent way.

index.html
decentralgabe: my intention was to convey if the property was there it is expected to be evaluated. if the issuer puts it there, they intend for it to be used.

decentralgabe: if the issuer puts it there and wants to say it's optional - I'm not sure how they convey that. I view the properties in the VCDM as a (loose) contract between an issuer and verifier

msporny: Nevermind, I failed to realize that the section you added is in the Validation section of the spec. :)

If the intention is to make processing the `credentialSchema` field mandatory, then we need to discuss that as a WG. 

With my company representative hat on, we would be a -1 to mandatory processing of `credentialSchema`. We believe it should be an option for the verifier to set because it is, ultimately, their decision. The feature should mirror how we do `credentialStatus`.

With my editor hat on, having language like "is expected to be", when what you really mean is "MUST", is dicey, because people will read it as a MAY instead of a MUST. Even SHOULD is dicey if you are suggesting that organizations will process the field by default.

To provide an analogy, the `credentialStatus` field processing is optional, the spec doesn't say you MUST process it. It just says what the field is used for and what the field should contain:

https://w3c.github.io/vc-data-model/#status

And then we use the  "is expected to be" language in the section on Validation... which is what this PR does (IIUC).

msporny:
Editorial, multiple reviews, changes requested (but not made due to change request being retracted), no objections, merging.

index.html
decentralgabe: thanks @msporny I'm open to a future change to keep the language more consistent with `credentialStatus`

PR #1149, opened 2023-06-11
Add graph node identifiers for registered claims

OR13:
Since the v2 context contains terms specific to data integrity proofs.

And I have closed the PR attempting to remove those terms: https://github.com/w3c/vc-data-model/pull/1091

It stands to reason we will need to add terms related to all proof formats to the v2 context, in the spirit of making things easier for developers.

### DIFF
--- a/contexts/credentials/v2
+++ b/contexts/credentials/v2
@@ -26,7 +26,6 @@
       "@id": "https://www.iana.org/assignments/jose#x5u",
       "@type": "@id"
     },
-    
 
     "VerifiableCredential": {
       "@id": "https://www.w3.org/2018/credentials#VerifiableCredential",
--- a/contexts/credentials/v2
+++ b/contexts/credentials/v2
@@ -4,7 +4,14 @@
     "@vocab": "https://www.w3.org/ns/credentials/issuer-dependent#",
 
     "id": "@id",
+    "kid": "@id",
+    "iss": "@id",
+    "sub": "@id",
+    "jku": "@id",
+    "x5u": "@id",
+    
     "type": "@type",
+    
 
     "VerifiableCredential": {
       "@id": "https://www.w3.org/2018/credentials#VerifiableCredential",
--- a/contexts/credentials/v2
+++ b/contexts/credentials/v2
@@ -11,7 +11,7 @@
       "@type": "@id"
     },
     "iss": {
-      "@id": "https://www.iana.org/assignments/jwt#iss",
+      "@id": "https://www.iana.org/assignments/jose#iss",
       "@type": "@id"
     },
     "sub": {
--- a/contexts/credentials/v2
+++ b/contexts/credentials/v2
@@ -7,7 +7,7 @@
     "type": "@type",
     
     "kid": {
-      "@id": "https://www.iana.org/assignments/jwt#kid",
+      "@id": "https://www.iana.org/assignments/jose#kid",
       "@type": "@id"
     },
     "iss": {
--- a/contexts/credentials/v2
+++ b/contexts/credentials/v2
@@ -23,7 +23,7 @@
       "@type": "@id"
     },
     "x5u": {
-      "@id": "https://www.iana.org/assignments/jwt#x5u",
+      "@id": "https://www.iana.org/assignments/jose#x5u",
       "@type": "@id"
     },
     
--- a/contexts/credentials/v2
+++ b/contexts/credentials/v2
@@ -5,7 +5,6 @@
 
     "id": "@id",
     "type": "@type",
-    
     "kid": {
       "@id": "https://www.iana.org/assignments/jose#kid",
       "@type": "@id"
--- a/contexts/credentials/v2
+++ b/contexts/credentials/v2
@@ -19,7 +19,7 @@
       "@type": "@id"
     },
     "jku": {
-      "@id": "https://www.iana.org/assignments/jwt#jku",
+      "@id": "https://www.iana.org/assignments/jose#jku",
       "@type": "@id"
     },
     "x5u": {
--- a/contexts/credentials/v2
+++ b/contexts/credentials/v2
@@ -15,7 +15,7 @@
       "@type": "@id"
     },
     "sub": {
-      "@id": "https://www.iana.org/assignments/jwt#sub",
+      "@id": "https://www.iana.org/assignments/jose#sub",
       "@type": "@id"
     },
     "jku": {
--- a/contexts/credentials/v2
+++ b/contexts/credentials/v2
@@ -4,13 +4,28 @@
     "@vocab": "https://www.w3.org/ns/credentials/issuer-dependent#",
 
     "id": "@id",
-    "kid": "@id",
-    "iss": "@id",
-    "sub": "@id",
-    "jku": "@id",
-    "x5u": "@id",
-    
     "type": "@type",
+    
+    "kid": {
+      "@id": "https://www.iana.org/assignments/jwt#kid",
+      "@type": "@id"
+    },
+    "iss": {
+      "@id": "https://www.iana.org/assignments/jwt#iss",
+      "@type": "@id"
+    },
+    "sub": {
+      "@id": "https://www.iana.org/assignments/jwt#sub",
+      "@type": "@id"
+    },
+    "jku": {
+      "@id": "https://www.iana.org/assignments/jwt#jku",
+      "@type": "@id"
+    },
+    "x5u": {
+      "@id": "https://www.iana.org/assignments/jwt#x5u",
+      "@type": "@id"
+    },
     
 
     "VerifiableCredential": {


OR13:
See related PR here:

- https://github.com/w3c/did-spec-registries/pull/517

OR13:
Note the data integrity spec has the same problem as did core currently, and it repeats normative definitions from did core:

https://w3c.github.io/vc-data-integrity/#verification-material

Screen Shot 2023-06-11 at 12 14 52 PM

OR13:
related PR updating the vc data model https://github.com/w3c/vc-data-integrity/pull/99

msporny:
> It stands to reason we will need to add terms related to all proof formats to the v2 context, in the spirit of making things easier for developers.

I'm not following the logic here. Where are JWT terms used inside of VCs? VC-JWT secures VCs by using an external proof. The terms you're adding to the context are never used in a VC-JWT payload. That is, the specific values you're adding are never used /inside/ of a VC, they're always used /outside/ of a VC.

What am I missing?

TallTed:
I'm with @msporny and @dlongley. This doesn't map to my mental model.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-06-14-vcwg#section1-4) on 2023-06-14

- no resolutions were taken

#### 1.4. Add graph node identifiers for registered claims (pr vc-data-model#1149)

_See github pull request [vc-data-model#1149](https://github.com/w3c/vc-data-model/pull/1149)._




**Manu Sporny:** pull 1149 has questions.  
… vc data integrity.  

</details>

contexts/credentials/v2
```diff
@@ -4,7 +4,14 @@
     "@vocab": "https://www.w3.org/ns/credentials/issuer-dependent#",
 
     "id": "@id",
+    "kid": "@id",
+    "iss": "@id",
+    "sub": "@id",
+    "jku": "@id",
+    "x5u": "@id",
```
msporny, line 11:
```suggestion
    "kid": "@id",
    "iss": "@id",
    "sub": "@id",
    "jku": "@id",
    "x5u": "@id",
```

The code above aliases everything to the `@id` JSON-LD keyword, which will lead to every JWT property listed above being mapped back to the `id` property... which is probably not what you wanted.

msporny (changes requested):
The PR accidentally introduces an invalid JSON-LD Context that would corrupt any JWT properties used in a VC. It is also not clear why a VC would need to include these properties since a JWT is an external proof format (thus the properties never appear in a VC).

Requesting clarification on what this PR is attempting to do.

contexts/credentials/v2
OR13: https://github.com/w3c/vc-data-model/pull/1149/files#diff-267b7e47789daf9a1d312b42826aa937a1bb5af4312ee90df6b8a64cd6426f4fR21

Ahh I meant to do this:

```
 "issuer": {
          "@id": "https://www.w3.org/2018/credentials#issuer",
          "@type": "@id"
        },
```

contexts/credentials/v2
```diff
@@ -5,6 +5,28 @@
...
     "type": "@type",
+    
+    "kid": {
+      "@id": "https://www.iana.org/assignments/jwt#kid",
+      "@type": "@id"
+    },
+    "iss": {
+      "@id": "https://www.iana.org/assignments/jwt#iss",
+      "@type": "@id"
+    },
```
OR13, line 16: Compare to how we define `issuer` today, this JSON-LD maps behavior for `iss` the same way.

contexts/credentials/v2
msporny: @OR13 wrote:
> Ahh I meant to do this

Ok, thanks for fixing that... the PR makes a lot more sense now.

msporny (changes requested):
What is the use case for defining these terms in the core context? Do you have an example? We might want to put that example in the core specification. I'm asking for the same sorts of demonstrations of specification and usage that were requested for `renderMethod` and `confidenceMethod`.

The thing that is still confusing me about this PR is that these properties don't show up inside VCs (AFAIK). They're properties used in an external proof. An example or two from a use case would be helpful.

I'll also note that the fragment identifiers do not resolve. When are the fragment identifiers going to be fixed in the IANA registry so that the fragments resolve?

For the avoidance of doubt, if we have a legitimate use case that the VCWG agrees to that uses these properties, and as long as we figure out what core spec text is going to exist to legitimize the inclusion of these properties, I won't object to the inclusion of these properties.

OR13:
@msporny same use case as defining `proof` and `DataIntegrityProof` in the normative v2 context.

mprorock (approved):

brentzundel:
Orie is asserting that including these properties will make life easier for developers of VC-JWT in the same way that the inclusion of `DataIntegrityProof` makes life easier for Data Integrity developers. I am trusting his assertion about VC-JWT just as I am trusting others' assertions about `DataIntegrityProof`.

Those opposed to including `DataIntegrityProof` expressed a lack of understanding about precisely how a developer's life is made easier by including it.
This lack of understanding was not sufficient to remove `DataIntegrityProof`, why should similar lack of understanding around VC-JWT terms be sufficient to prevent them from being added?

I also want to say that the bar for inclusion in the `@context` hasn't been "core spec text . . . to legitimize the inclusion of these properties" (unless there is core spec text about Data Integrity that I'm unaware of), so it seems inappropriate to make that a requirement here.

Is there a harm that merging this PR would cause that I'm not seeing? Would merging this PR prevent you from doing what you need with the `@context`? 

tl;dr I don't have deep opinions about the `@context` file, but I don't understand the pushback here, it seems inconsistent.

msporny:
> tl;dr I don't have deep opinions about the `@context` file, but I don't understand the pushback here, it seems inconsistent.

For the `proof` and `DataIntegrityProof` entries in the core VCDM context, we had to demonstrate that developers would benefit. The argument to date has been "We want to minimize the JSON-LD Contexts used to ease the developer experience." That is what led to the default vocabulary decision. Given that we have demonstration of over 20 implementers utilizing `proof`, and given that all the Data Integrity cryptosuites use `DataIntegrityProof` type now, we could reduce the number of required contexts when implementers used `proof` down from two contexts to one context. IOW, the main way you use data integrity is this:

```
// This is a VC that is secured using a data integrity proof
{
  "@context": [
  ],
  "type": ["VerifiableCredential", "ExampleDegreeCredential"],
  "issuer": "https://university.example",
  "credentialSubject": {
    "degree": {
    }
  },
  "proof": {
    "type": "DataIntegrityProof",
    "cryptosuite": "ecdsa-2019",
    "created": "2022-11-13T18:19:39Z",
    "verificationMethod": "https://university.example/issuers/14#key-1",
    "proofValue": "z58DAdFfa9SkqZMV...fFPP2oumHKtz"
  }
}```

If we didn't have `proof` and `DataIntegrityProof` in the core context, you'd have to add another JSON-LD Context to make it work, which people had complained about in the early days of the VCWG, which is what led to the inclusion in the VCDM core context.

This PR adds things like `iss`, `sub`, `jku`, and `kid` to the core context, but those values are not used in VCs. For example, I don't know of any use case that looks like this at present:

```
{
  "@context": [
  ],
  "kid": "https://university.example/issuers/14#key-1", // JOSE property
  "iss": "https://university.example", // JOSE property
  "sub": "did:example:ebfeb1f712ebc6f1c276e12ec21", // JOSE property
  "jku": "https://university.example/issuers/14.json", // JOSE property
  // rest of VC is below
  "type": ["VerifiableCredential", "ExampleDegreeCredential"],
  "credentialSubject": {
    "degree": {
    }
  }
}```

That doesn't mean a use case might not exist, which is why there is a request to understand what use case this PR addresses where those values would be expressed in embedded form. We had to provide examples for `proof` and `DataIntegrityProof`. Examples have not been provided for iss/sub/jku/kid *as embedded values* yet.

dlongley:
@brentzundel,

I don't have any objections to this based on the grounds that others used to argue a while back against adding various things to the context. I said those grounds didn't make sense back then -- and I continue to think that. And, as I said back then, adding things to the context that make things easier for developers is generally a good idea, and if this PR were doing that as well, I'd be fine with it.

My concern here is that this is making things more confusing for developers. We had successfully established a clean separation between JWT claims expressed in the external securing mechanism (VC-JWT) and its payload (`application/vc+ld+json`). This seems to co-mingle those things again.

It also seems to potentially bring back the confusion from JWT usage with VCDM 1.1: e.g., are we expecting people to use `iss` "instead of" or "in addition to" `issuer`? Clearly they must include `issuer` to be VCDM 2.0 compliant, so only "in addition to" would be valid. Did we want to bring that back? And will we see compliance in the wild with that approach or will some people use "instead of" again? Will that create additional standardization pressures? Won't using both fields encourage more confusion? What if the values don't agree in the duplicated fields? Wasn't it a victory to have eliminated all these possibilities?

This seems to reintroduce problems that were railed against in the 1.1 work. In fact, it adds another dimension to the possible (erroneous / non-compliant) ways to secure VCs using JWTs, with implementers potentially being confused over whether the payload will, for example, have only `iss`, have `iss` and `issuer`, whether those fields will agree, whether the payload will have `vc` (VCDM 1.1 JWT style) or whether `iss` will be co-mingled directly with the payload vs. using `vc` -- and so on.

All of that being said, I'm not going to formally object to adding these as scoped properties to a VC via the JSON-LD context -- I just think it's a really bad idea that's asking for trouble. This is not an academic or ideological purity argument in the slightest, but a practical one. I've said similar things in this WG when trying to help people who want to use JWTs to secure the data model in a number of places. Sometimes my advice is used, other times it is dismissed. IMO, it's too infrequently responded to in collaborative conversation.

I get the awful impression that a number of PR discussions are mentally processed like this: "You said some stuff about my PR, but are you going to block it or not?". I suspect that's not the best way to design technology standards. Neither is going around blocking people's efforts, of course, but speaking for myself, I don't go around "just trying to block things", but rather I seek to improve our work here generally. Anyway, this PR, to me, sounds like a bad idea and that it just reintroduces a number of things that were fixed during the 2.0 work. I'd love to hear how I'm wrong about that and why. I'm happy to be wrong if it means things get better.

So, to be clear, the arguments against this PR are not the same arguments that were used elsewhere -- and this PR doesn't do the same things that those other PRs did either. This is a wholly different situation in both reasoning for / against and effect. Regardless, if my advice is ignored that this is a bad idea and no arguments are put forth to help convince me why it's not a bad idea, I continue to think VC-JWT users will suffer from it.

My only other concern would be that people will make more mistakes with this in the wild after 2.0 is released -- and then a future 3.0 WG will have to repeat the same long discussions we had during 2.0 and a lot of progress in other areas will be sacrificed again because of it. I'd rather see us embrace the clean approach and avoid repeating the 1.1 mistakes.

OR13:
@msporny @dlongley since the VC Data Model is open world, if we don't reserve these words like this, we will be creating harmful security issues... nothing is more confusing than using reserved words, but given them different meaning.

dlongley:
@OR13,

> @msporny @dlongley since the VC Data Model is open world, if we don't reserve these words like this, we will be creating harmful security issues... nothing is more confusing than using reserved words, but given them different meaning.

Is that the reason for this PR -- reserving terms to avoid confusion? If this PR is about reserving terms, can you make it say so in the title or in your remarks in the opening post? I had no idea. 

This PR also only makes changes to the context -- I expect we'll want to say more in the spec. We may want to tell developers specifically NOT to ever use these terms in VCs at all and point them at VC-JWT for how to cleanly separate things. This "reservation" is a bit different from how we've been doing the other term reservations because we intend people to use and experiment with the others, but here we want to recommend they never be used, IMO. Please clarify if that's the goal here, because I suspect it changes how people will think about this PR in significant ways.

OR13:
Yes, perhaps this PR is premature, given we still have not reached consensus on (which were opened after it):

- https://github.com/w3c/vc-data-model/pull/1159
- https://github.com/w3c/vc-data-model/pull/1158

I don't see the reason to reinvent vocabularies that have already been established by IANA, do think it makes sense to use existing standards when possible, and not redo work, unless there is some fundamental improvement.

I find most of the term definitions we are relying on normatively that are hosted by https://w3id.org to be inferior to IANA, and, in general, to be less well adopted, and less used.

The current vocabulary file remains broken, and does not display terms properly, I have opted not to touch it, since it's not currently normative, and therefore does not need to be understood by implementers.

Certainly if implementers MUST understand certain terms, the working group should explicitly address all points of confusion, if this PR is not merged, a second context could overwrite these term definitions and that would have the same effect as not protecting `VerifiableCredential` or `DataIntegrityProof`.

dlongley:
@OR13,

> I don't see the reason to reinvent vocabularies that have already been established by IANA.

The point is that we don't want people to think that these terms are being properly used, but rather, the opposite. So if that's the aim, we shouldn't map them to IANA at all, we should map them to some "do not use" vocabulary so they cannot be mistaken for proper use.

mprorock:
> @OR13 —
> 
> > hosted by https://w3id.org/
> 
> _NOTHING_ is hosted by w3id.org except `.htaccess` files that redirect from `w3id.org/*` URLs to wherever the relevant document actually lives.

can we drop it then and point out to authoritative sources directly? (genuinely want to understand)


> 
> > Add graph node identifiers for registered claims
> 
> I think consideration and discussion of this PR would be much more accurate if its title was changed to "Add mappings from JWT claim names to IANA URLs, to default `@context`", and then I could say "no, don't do that, unless you also add the same claim names to the list of reserved attribute names"

those are already reserved in RFCs, so we should double reserve in your mind?

TallTed:
@OR13 said in IRC on today's call, 
> if w3id.org is compromised, proof could be defined and redirected to anything else... that is my point... previously proof pointed to the W3C CCG github... we should change what it redirects too, in this file: https://github.com/perma-id/w3id.org/blob/master/security/.htaccess

_Maybe_ the redirect _should_ be changed.  But to what?

I do notice that YOU have the power to change that redirect (at line 5 of the `.htaccess`), as a maintainer listed in the [`README.md`](https://github.com/perma-id/w3id.org/blob/master/security/README.md).

TallTed:
@mprorock — 

> can we drop [w3id.org] then and point out to authoritative sources directly? (genuinely want to understand)

We used the w3id.org-based URLs instead of CCG-based or VCWG-based because they could be kept stable while their targets move around, e.g., from CCG to VCWG to CCG... until stability eventually (maybe) happens. This was to prevent calcification of any temporary URL (`http://xmlns.com/foaf/0.1/`, anyone?) that escaped into the wild.

> those are already reserved in RFCs, so we should double reserve in your mind?

It's not about double-reservation.

For these terms, in our CR/PR/TRs, it's about communicating to users of *our* specs that those terms are reserved _elsewhere_ (e.g., RFC, other W3TR, etc.) which they may not have read as carefully, if at all, while working on *this* spec -- so they shouldn't use those terms as if they were unused/unreserved.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-06-27-vcwg-special#section3) on 2023-06-27

- no resolutions were taken

### 3. Add graph node identifiers for registered claims (pr vc-data-model#1149)

_See github pull request [vc-data-model#1149](https://github.com/w3c/vc-data-model/pull/1149)._




**Kristina Yasuda:** PR 1149. what do we need to resolve?  

**Michael Prorock:** find it weird that we are up to 22 comments when all this PR does it point over to the actual definition in the RFC.  

**Dave Longley:** there has been discussion in the PR around exactly as what the PR is doing. Don't think its as simple as what Mike is saying.  
… would this cause people to use 1.1 or cause confusing with 2.0. There is a number of comments here and another case of.  
… a PR that was intended to be a simple thing, but need discussion to figure out what we need to do.  

**Kristina Yasuda:** hopefully we get a bit more clarity. we will not resolve it here.  

**Michael Prorock:** note that based on previous comments, this PR (1149) is effectively doing the same as including the proof in the context.  

> *Dave Longley:* mprorock, please read the comments in the PR, this was discussed.

**Michael Prorock:** to me this is an either or. It only has core data model items, or has additional helpers. We can't have it both ways.  

> *Orie Steele:* +1 mprorock.

> *Manu Sporny:* It's not doing the same thing as proof in the context, that's the core of the disagreement -- you don't use iss/kid/etc like you use proof.

> *Manu Sporny:* Specifically, we have discussed this in the PR: [https://github.com/w3c/vc-data-model/pull/1149#issuecomment-1605659418](https://github.com/w3c/vc-data-model/pull/1149#issuecomment-1605659418).

> *Dave Longley:* mprorock: some of the things you're saying in the call here were discussed in the PR, please read the PR.

> *Orie Steele:* [https://w3id.org](https://w3id.org) hosts this html file.

**Ted Thibodeau Jr.:** Lots of comments and they have substance to them. Good to review them. w3.org is a redirect service. Its not hosted. You cannot point a finger here.  

> *Orie Steele:* it also redirects to [https://w3id.org/citizenship](https://w3id.org/citizenship).  
> *Orie Steele:* which defines [https://w3id.org/citizenship#PermanentResidentCard](https://w3id.org/citizenship#PermanentResidentCard).  
> *Orie Steele:* w3id.org also defines our normative concept of [https://w3id.org/security#proof](https://w3id.org/security#proof).

> *Ted Thibodeau Jr.:* [https://w3id.org/security#proof](https://w3id.org/security#proof) redirects to [https://w3c.github.io/vc-data-integrity/vocab/security/vocabulary.html#proof](https://w3c.github.io/vc-data-integrity/vocab/security/vocabulary.html#proof).  
> *Ted Thibodeau Jr.:* w3id.org is NOT defining anything.

> *Orie Steele:* TallTed, if w3id.org is compromised, proof could be defined and redirected to anything else... that is my point... previously proof pointed to the W3C CCG github... we should change what it redirects too, in this file: [https://github.com/perma-id/w3id.org/blob/master/security/.htaccess](https://github.com/perma-id/w3id.org/blob/master/security/.htaccess).

**Ted Thibodeau Jr.:** this PR has a title that is not reflective of what is happening in the PR. Please update.  

> *Manu Sporny:* Agree that the PR is not titled appropriately.

**Ted Thibodeau Jr.:** I cant provide an alternative because I'm not writing this. If orie cannot rephrase it...  

</details>

dlongley:
@OR13,

Can you provide more clarity here? Is the intent for people to actually use the terms you are trying to add to the context in the payload for a VC-JWT or are they only to be used in a header (as I had previously thought)?

If we *really* want people to use these terms in their VCs (in the JSON-LD), then it makes sense to define them in the context and I can approve the PR. But it sounds like that's a mistake (and I've said why above) and what we actually want is for people to NOT *ever* use these in the JSON-LD and instead only put them in the VC JWT header for a clean separation.

Which of these is it? I think it's technically fine to put them into the JSON-LD context either way (ignoring the consequences of VC JWT confusion I expect to occur). If it's the former it makes sense to use the URLs you've provided (I think). If it's the latter, however, it makes sense to map them to a "do not use" vocabulary.

Can you please say what your intentions were here so we can move forward in the appropriate direction? And, ideally, could you respond to my concern about making VC JWT worse by encouraging confusion over where to use the terms?

OR13:
> If we really want people to use these terms in their VCs (in the JSON-LD), then it makes sense to define them in the context and I can approve the PR. 

It's not that we "want them too"... it's that they are already allowed to... and that without this PR being merged, the behavior would be just as ambiguous as if they could define `DataIntegrityProof` or `proof` to mean different things.

Default behavior today would have them be issuer dependent... but that is not consistent with the RFCs... and its not safe.

If you think `DataIntegrityProof` should be made issuer dependent as well by default, we should remove it from the v2 context, that will accomplish that behavior, `proof` and `DataIntegrityProof` would also get "issuer dependent definitions".

> If it's the latter, however, it makes sense to map them to a "do not use" vocabulary.

^ this is not a thing that makes any sense in the context of the terms, or this PR.

> Can you please say what your intentions were here so we can move forward in the appropriate direction? And, ideally, could you respond to my concern about making VC JWT worse by encouraging confusion over where to use the terms?

My intentions are to align the v2 core context (which is now normative), with the terms related to the "securing specifications", to make things easier for developers.

dlongley:
@OR13,

> > If we really want people to use these terms in their VCs (in the JSON-LD), then it makes sense to define them in the context and I can approve the PR.
> 
> It's not that we "want them too"... it's that they are already allowed to... and that without this PR being merged, the behavior would be just as ambiguous as if they could define `DataIntegrityProof` or `proof` to mean different things.

I agree. We *should* say *something*, I just want to make sure we're encouraging the right behavior. While anything is allowed in an open world, some of those things are quite specific and we are actively telling people to use them (to get certain standard features). Of course, we can, in the core spec, specifically tell people that they MUST NOT do certain things as well, if those things are easy enough to test and if we think it's worthwhile to put in the effort to do all that.

Do we want to make it so that if people try to misuse the properties that they will at least get flagged as such? Or do we want to put links to real definitions for those things in our context ... but not say anything about how it's a bad idea to use them? That seems like exactly the kind of thing @OR13 would say is a bad idea :).

Let me be extra clear -- I'm not going to block the definitions going in either way. I just want clarity that you know what you're doing and that we're encouraging what we want to see happen. I'm trying to help prevent harm when people use  what is currently our only defined external security approach.

> 
> Default behavior today would have them be issuer dependent... but that is not consistent with the RFCs... and its not safe.
> 
> If you think `DataIntegrityProof` should be made issuer dependent as well by default, we should remove it from the v2 context, that will accomplish that behavior, `proof` and `DataIntegrityProof` would also get "issuer dependent definitions".

As I've said before, putting `@vocab` in the core v2 context is a strong driver to ensure that the core context also contains any definitions for terms we are publishing in other contexts in v2 to avoid precisely this confusion. I don't think there's another context for the VC-JWT terms from which we'd be copying term definitions -- so it's a bit out on its own, but I understand that `@vocab` still creates a possible issue because of the nature of JWT / JWT Claimset usage. The other different bit is that we actually don't want anyone to use those terms (outside of the protected header which is not JSON-LD). So, again, it's an odd duck.

Perhaps the best thing to do is to define the terms using IANA URLs but tell people they should be putting them in the protected header NOT in the VC itself. A statement that says that could alleviate confusion.

OR13:
> I'm trying to help prevent harm when people use what is currently our only defined external security approach.

Can you elaborate on how you are helping address this?

> Perhaps the best thing to do is to define the terms using IANA URLs but tell people they should be putting them in the protected header NOT in the VC itself. A statement that says that could alleviate confusion.

This would in my opinion, go against the comment above, and have the opposite effect, since many of these terms are already found in the `payload` / `claimset` and only this language, allows them to be promoted to the header:

https://datatracker.ietf.org/doc/html/rfc7519#section-5.3

In my opinion, the W3C VCWG idiomatic solution to this is to ensure terms get the best definition possible, by default.

Which means:

1. using `@vocab` ( counter to your advice, which I continue to disagree with )
2. in cases where a securing mechanism relies on a term (such a `proof` or `DataIntegrityProof`, or `cnf` ), ensuring the correct term is assigned by default in the v2 context.
3. using second contexts to extend the claims and meta data, not the security mechanism terms.

dlongley:
@OR13,

> > I'm trying to help prevent harm when people use what is currently our only defined external security approach.
> 
> Can you elaborate on how you are helping address this?

Sure. I started out by pointing out that what you wanted to do originally was clearly a mistake as it would define several `@id` aliases in the context. Rushing to approve and merge that would have been a problem.

This led me to question what your intent here even was. The issue title is still "Add graph node identifiers for registered claims", which continues to make no sense to me, but I've set it aside.

After many prose and questions here, I've started to figure out what it is you want to do so that I can make sure that this PR actually accomplishes it. I'm also trying to reconcile what it is you want to do with all of your (great) work in cleaning up ambiguities in the VC JWT 1.1 work with the new cleanly separated approach with the VC JWT 2.0 work. It would be a real shame to undo that with a mistake here, yes?

If you, as the person who more or less led the effort that cleaned up ambiguity in the "instead of"/"in addition to"/"what happens when JWT claim values don't agree" work think the best solution here is to define the terms using IANA URLs, I'm ok with that.

I also wanted to offer you an option to do something else that might better discourage the kind of ambiguity you helped convince the group should be avoided. If you think using IANA URLs is the best way forward to do that, I'm a bit dubious of it (at least without additional text somewhere explaining what's going on), but willing to go along regardless because I really don't know -- I only suspect a bad outcome.

I've also helped by suggesting that we say more (as you've put it "offer developers good advice") on the fact that they should be cleanly separating the JWT terms from the JSON-LD payload, putting the JWT terms (that are presently defined here against IANA URLs) only in the protected header.

> 
> > Perhaps the best thing to do is to define the terms using IANA URLs but tell people they should be putting them in the protected header NOT in the VC itself. A statement that says that could alleviate confusion.
> 
> This would in my opinion, go against the comment above, and have the opposite effect, since many of these terms are already found in the `payload` / `claimset` and only this language, allows them to be promoted to the header:
> 
> https://datatracker.ietf.org/doc/html/rfc7519#section-5.3
> 
> In my opinion, the W3C VCWG idiomatic solution to this is to ensure terms get the best definition possible, by default.
> 
> Which means:
> 
> 1. using `@vocab` ( counter to your advice, which I continue to disagree with )
> 2. in cases where a securing mechanism relies on a term (such a `proof` or `DataIntegrityProof`, or `cnf` ), ensuring the correct term is assigned by default in the v2 context.
> 3. using second contexts to extend the claims and meta data, not the security mechanism terms.

Well, I do think this reads like `iss` will be in the protected header, also in the payload, and there will also be `issuer` from the VCDM. Each of these values could be in disagreement. This is just one case -- I presume other JWT terms could have similar issues. I do not look forward to implementing VC-JWT to reconcile the ambiguities, so perhaps VC-JWT could at least say more on what to do.

I have to say that I'm honestly confused by the logic of your position given how vociferously you fought to remove the same ambiguities in the past within the same work item -- and that you simultaneously don't seem to have changed your mind. If you've changed you mind, that at least makes sense ... but otherwise it seems you're holding two positions that are at odds with one another. It could also be the case that you're doing something approximating this because this is standards work and you think this is the best possible outcome. I know that feeling well myself.

For that latter possibility and for everything else I've said, I'm ok with proceeding with this PR as it is. I don't seem to be able to effectively communicate any other disagreements or help push things forward to mutual understanding.

dlongley (approved):

OR13:
@msporny can you re-review?

msporny:
> @msporny can you re-review?

I requested a use case up in this comment: https://github.com/w3c/vc-data-model/pull/1149#issuecomment-1586351534

Your response was that it was the same use case as `DataIntegrityProof`, which is confusing because JWTs are an external proof format, whereas  `DataIntegrityProof`s are an internal proof format. So, it doesn't seem to be the same use case.

Except, [this PR](https://github.com/w3c/vc-jwt/pull/114/files#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051R1127-R1129) (which I don't fully understand, but am presuming others that know about SD-JWTs direction do) seems to indicate that there is a usage that mixes the `iss` and `issuer` field in the same VC (which is asking for trouble, IMHO)... but at least I can maybe see a use case there? except, the example uses the "https://www.w3.org/2018/credentials/v1" context... and `cnf` isn't in this PR, but is used in that example?

IOW, it's still not clear how these properties are expected to be used and my request for a clean example has still not been provided. My review on this is still a -1 to merge until the WG can be shown how these properties are going to be used in VCs.

OR13:
@msporny are you familiar with how key discovery works for OIDC?

Are you asserting that a property of W3C Verifiable Credentials is that they are never possibly compatible with existing token formats, and when registered claims from an existing token format are present in the claimset, credential, or verifiable credential, they should be assigned issuer dependent IRIs by `@vocab` if processed as JSON-LD, although such processing is never required by our spec today?

This means W3C vocabulary is redefining terms that IANA and IETF already defined... In a way that is... Not helpful to developers.

I can't throw a `@protected` term redefined if you refuse to include these terms in the v2 `@context`, is this the goal?

Are you asserting that a minimum of 2 contexts is needed to define `iss`?

OR13:
Related comments here:

https://github.com/w3c/vc-data-model/issues/1185#issuecomment-1629879100

I agree with what @dlongley is saying.

I also feel there is tremendous bias and reinventing of wheels happening that is harmful to implementers and is manufacturing confusion.

I don't see the benefit of going out of our way to break compatibility.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-07-11-vcwg-special#section1-4) on 2023-07-11

- no resolutions were taken

#### 1.4. Add graph node identifiers for registered claims (pr vc-data-model#1149)

_See github pull request [vc-data-model#1149](https://github.com/w3c/vc-data-model/pull/1149)._




**Brent Zundel:** graph node identifiers for registered claims.  

**Orie Steele:** P.R adds term defs for registered claim names consistent with IANA registries.  
… This will enable terms to be expanded with IANA term definitions.  
… moved status list into v2 context, moved data integrity to make things easier. Follows that we would define these registered claim names in the v2 as well.  

**Dave Longley:** I have approved the P.R.  
… my understanding is current blocking is an example for the use of this.  
… Help make sure we provide appropriate guidance to folks to avoid 1.1 mistakes.  

> *Orie Steele:* Do we have guidance on the DataIntegrity and StatusList parts, or is a different bar being applied here?

> *Dave Longley:* orie, totally different situations :).

> *Orie Steele:* disagree.

> *Dave Longley:* there are no conflicting properties there.

**Brent Zundel:** TallTed can you speak to the change requests you would like.  

**Ted Thibodeau Jr.:** still reviewing the P.R.  

**Brent Zundel:** my opinion as chair that if a P.R requesting changes to the spec. Can't just say you don't like it. If it doesn't prohibit you from doing what you need and doesnt break the internet. Then we should lean towards trusting intentions of those who raised P.R and let the work move forward.  
… TallTed if you can take some time in the next few days to see if your request for changes have been satisfied. Same for manu, alhough not present today.  

</details>

contexts/credentials/v2
```diff
@@ -5,6 +5,28 @@
...
+    "sub": {
+      "@id": "https://www.iana.org/assignments/jwt#sub",
+      "@type": "@id"
+    },
+    "jku": {
+      "@id": "https://www.iana.org/assignments/jwt#jku",
+      "@type": "@id"
+    },
+    "x5u": {
+      "@id": "https://www.iana.org/assignments/jwt#x5u",
```
davidlehn, line 26: This doesn't exist.  Perhaps it should be a jose url?  I see the jose page duplicates lots of params?  And in both cases the HTML has no anchors at all.  Are these the canonical form that community uses for params?

msporny:
> @msporny are you familiar with how key discovery works for OIDC?

Are you talking about the `jwks_uri` and the  `/.well-known/openid-configuration` file? If so, could you please elaborate on why that is relevant?
 
> Are you asserting that a property of W3C Verifiable Credentials is that they are never possibly compatible with existing token formats

No, I'm not asserting that.

> and when registered claims from an existing token format are present in the claimset, credential, or verifiable credential, they should be assigned issuer dependent IRIs by `@vocab` if processed as JSON-LD

This is exactly why I was a -1 to use of `@vocab` in the base context... it has all sorts of unintended consequences, as evidenced by this PR. So, now we have to hack stuff into the base context because of the existence of `@vocab` in the base context.

> This means W3C vocabulary is redefining terms that IANA and IETF already defined... In a way that is... Not helpful to developers.

Yes, `@vocab` in the base context is not helpful to developers. It causes problems such as the one you are describing.

> I can't throw a `@protected` term redefined if you refuse to include these terms in the v2 `@context`, is this the goal?

No, that's not a goal. 

> Are you asserting that a minimum of 2 contexts is needed to define `iss`?

That would probably have been best given the problems that `@vocab` in the base context is creating.

> Related comments here:
> 
> [#1185 (comment)](https://github.com/w3c/vc-data-model/issues/1185#issuecomment-1629879100)
> 
> I agree with what @dlongley is saying.

Sure, I agree with what @dlongley is saying there as well. How does it apply here?

> I also feel there is tremendous bias and reinventing of wheels happening that is harmful to implementers and is manufacturing confusion.
> I don't see the benefit of going out of our way to break compatibility.

The two statements above are hyperbolic, so I'm going to ignore them.

Given that you continue to not directly answer what you're trying to accomplish with this PR, I'm going to try and guess what you're trying to do (given that you have still not provided examples, which have been requested multiple times in this thread).

I /think/ you want to mix JWT claims into a VC *and* process it as JSON-LD. 

I don't know what use case there is that does that (other than some hand-waving at OIDC key discovery). The reason you want to do this is because a feature that you argued for, which was injecting `@vocab` into the base context, is now causing havok with the use case where one might mix JWT claims into a VC and process it as JSON-LD -- as was predicted would happen when adding `@vocab` into the base context was being debated in the group.

So, now we're going to have to play whack-a-mole with "other token formats", when what we should be doing is reversing the decision on `@vocab`... but I expect that would cause chaos in the group at this point, so the only option we have is to cherry pick which token formats we want to support VCs living "at the same level" at... which is what led to this PR.

Do I have that right so far? Because I'm guessing, given that you're not providing the information that has been repeatedly requested to date.

If so, have you passed those URL registrations by the JOSE/COSE WGs? Are those the URLs they want to use? There are no anchors that go to the registered properties. You have objected to PRs being merged because the URLs don't resolve correctly, are you going to object to this PR being merged until the URLs resolve correctly?

Are we missing any claims? Is this all of them, or are there more? Again, a concrete use case would really help here so that we know if we're addressing the use case with this PR, or only addressing a subset of it (which would cause incompatibilities once we're out to REC).

Here's what could get me to a +1 to merge this PR:

* Answers to the questions above so we have some way of evaluating the effectiveness of this PR.
* A concrete use case that an implementer needs this PR for, with a concrete example of how the entries in this PR are going to be used.

At this point I'm convinced that we should address the "JWT claims / VC mixing" problem... but in doing so, I expect we'll uncover even more issues with the proposal in this PR, such as having two "iss" and "issuer" properties at the same level and having to then provide guidance on how one should interpret each property.

msporny:
@brentzundel wrote:
> my opinion as chair that if a P.R requesting changes to the spec. Can't just say you don't like it. If it doesn't prohibit you from doing what you need and doesnt break the internet. 

To be clear, I'm concerned that the use case being proposed here will break the ecosystem.

Introducing `@vocab` in the base context created this problem, so the ecosystem is already broken (if what we're trying to do is mix VCs and JWT claims at the same level *and* interpret that as JSON-LD).

We can "fix" that in a hacky way by adopting this PR, but then that creates a new problem because "iss" and "issuer" will be expressed at the same level in a JWT, so the WG will have to give guidance on which one takes precedence, if it's either/or, etc. In short, we're going to end up re-visiting the "either/or VC-JWT v1.1" discussion again, but this time, instead of "iss" vs. "vc.issuer", it's going to be "iss" vs. "issuer" (at the same level).

This goes beyond "don't like it" -- these are technical concerns about how this PR is going to negatively impact the VC ecosystem. 

To be clear, I'm convinced that we need to do /something/ about this. That might be merging this PR as-is (and then providing guidance on use of "iss" and "issuer" at the same level... possibly other JWT claims), it might be merging this PR plus a number of other JWT claims (and we can't know which ones those are w/o a clear use case identified), or it might be reverting the `@vocab` value in the base context (though, I don't expect that to get consensus).

contexts/credentials/v2
iherman: Actually, it seems that none of the proposed ID-s exist, not only "x5u".

Adding URL-s with broken fragments is a no-no in the case of normative W3C specifications. If we consider the context file's _content_ to be normative (which seems to be the case) then this applies to these URL-s as well. To make it clear: the publication process may reject this publication when it comes to CR/PR/Rec publishing.

msporny: In order to fix the fragment identifiers that do exist, I expect that the JOSE WG would need to agree to update the JOSE registry to add the fragment identifiers. An alternative that might work is pointing directly to the specification, such as this one for "iss": https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.1 -- although, that's definitely not ideal.

Changing the fragment identifiers in the JOSE registry is probably the best path forward.

OR13:
> This goes beyond "don't like it" -- these are technical concerns about how this PR is going to negatively impact the VC ecosystem.

I share those concerns regarding including DataIntegrityProofs in the v2 context.

I think the cleanest approach is to not include "securing mechanism" JSON-LD in the v2 context.

However, if we are going to include terms in the v2 context related to securing mechanism, then we need to do that consistently.

> To be clear, I'm convinced that we need to do /something/ about this. That might be merging this PR as-is (and then providing guidance on use of "iss" and "issuer" at the same level... 

The problem you will face is the same one that vc-data-integrity addresses today with this section:

- https://w3c.github.io/vc-data-integrity/#retrieve-verification-method

^ this section of the securing mechanism, relies on `proof` and `DataIntegrityProof` in the v2 context.

- https://openid.net/specs/openid-connect-discovery-1_0.html
- https://auth0.com/blog/id-token-access-token-what-is-the-difference/

^ this section of a securing mechanism, relies on `iss` in the v2 context (assuming you want the JSON-LD term definition to reflect reality)

If we don't take this change, `iss` still works as expected in OIDC... but... the JSON-LD graph is `@vocab` expanded...

At least an error is not thrown.

This is my goal:

1. Terms that are used today, are documented correctly (applies to `proof` and `iss`).
2. v2 context is only context you need to have a good experience without being a JSON-LD expert.

We accomplished 2 when we included `@vocab`... we accomplish 1 when we merge this PR.

If we don't merge this PR, 2 is still true, and OIDC Key discovery still works, it's just the JSON-LD graph that is built is embarrassingly bad, and needlessly hostile to OIDC, or any other securing mechanism that depends on `iss`.

If JSON-LD is indeed, providing some value, by adding term definitions for JSON members, preventing that value from applying to "terms that are used in the wild and very popular, but which you don't like"... seems... not great.

OR13:
Possibly a better example here: https://github.com/w3c/vc-data-integrity/issues/77

there are relationships between the core data model and securing specifications that can be exploited by an attacker, even when terms are defined correctly... defining things better is only the first step towards addressing them.

brentzundel (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-07-11-vcwg-special#section1-4) on 2023-07-11

- no resolutions were taken

#### 1.4. Add graph node identifiers for registered claims (pr vc-data-model#1149)

_See github pull request [vc-data-model#1149](https://github.com/w3c/vc-data-model/pull/1149)._




**Brent Zundel:** graph node identifiers for registered claims.  

**Orie Steele:** P.R adds term defs for registered claim names consistent with IANA registries.  
… This will enable terms to be expanded with IANA term definitions.  
… moved status list into v2 context, moved data integrity to make things easier. Follows that we would define these registered claim names in the v2 as well.  

**Dave Longley:** I have approved the P.R.  
… my understanding is current blocking is an example for the use of this.  
… Help make sure we provide appropriate guidance to folks to avoid 1.1 mistakes.  

> *Orie Steele:* Do we have guidance on the DataIntegrity and StatusList parts, or is a different bar being applied here?

> *Dave Longley:* orie, totally different situations :).

> *Orie Steele:* disagree.

> *Dave Longley:* there are no conflicting properties there.

**Brent Zundel:** TallTed can you speak to the change requests you would like.  

**Ted Thibodeau Jr.:** still reviewing the P.R.  

**Brent Zundel:** my opinion as chair that if a P.R requesting changes to the spec. Can't just say you don't like it. If it doesn't prohibit you from doing what you need and doesnt break the internet. Then we should lean towards trusting intentions of those who raised P.R and let the work move forward.  
… TallTed if you can take some time in the next few days to see if your request for changes have been satisfied. Same for manu, alhough not present today.  

</details>

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-07-12-vcwg#section4-1) on 2023-07-12

- no resolutions were taken

#### 4.1. Add graph node identifiers for registered claims (pr vc-data-model#1149)

_See github pull request [vc-data-model#1149](https://github.com/w3c/vc-data-model/pull/1149)._




**Manu Sporny:** 1142 we are still waiting for someone to pick up confidenceMethod.  
… there has been discussion on this PR.  
… I have the same concerns I have had for a while.  
… I plan to respond to Orie some time later.  
… I am expecting reviews from Ted, Dave and Ivan.  
… I think we need more people to review that PR.  

</details>

nissimsan (approved):
This will shape really nice graph!

msporny (approved):
I am /very reluctantly/ not blocking this PR because of this comment:

@OR13 wrote:
> If we don't take this change, `iss` still works as expected in OIDC... but... the JSON-LD graph is `@vocab` expanded...

This presumes that we're going to mix top-level VCDM values with top-level JWT values. I don't think the WG has talked about doing this, and the WG has been given multiple opportunities to weigh in on this topic (and many have not done so). I continue to believe that mixing JWT values w/ VCDM values at the same level is a really bad idea and will work against JWT usage, but given that those that want to use JWTs are pushing for this feature, I'm not going to try to talk them out of it.

If some WG members are planning on mixing JWT values and VCDM values at the same level, we need to make sure the terms aren't corrupted by the default `@vocab`. To be clear, it is the default `@vocab` that is creating this problem. We could fix that issue by pulling the default vocab into a secondary context, but that has been proposed and rejected as a path forward.

I believe that both @iherman, @davidlehn, and @TallTed still have blocking concerns, so as long as those are addressed in some way, I'll be a +0.1 to this PR.

OR13:
> This presumes that we're going to mix top-level VCDM values with top-level JWT values. 

In an open world data model, to presume otherwise would be more problematic.

> I don't think the WG has talked about doing this, and the WG has been given multiple opportunities to weigh in on this topic (and many have not done so).

The working group has indeed talked about this at great length, and while some folks might prefer for it to be recommended, and others that it be discouraged, that is better addressed in the securing specifications, which rely on these parameters.

> I continue to believe that mixing JWT values w/ VCDM values at the same level is a really bad idea and will work against JWT usage, but given that those that want to use JWTs are pushing for this feature, I'm not going to try to talk them out of it.

I think it is worth considering what should `vc-jose-cose` say about this... but that is for that work item, which already partially addresses this topic... PRs remain welcome there.

> If some WG members are planning on mixing JWT values and VCDM values at the same level, we need to make sure the terms aren't corrupted by the default `@vocab`. To be clear, it is the default `@vocab` that is creating this problem. We could fix that issue by pulling the default vocab into a secondary context, but that has been proposed and rejected as a path forward.

I dispute that it is the default vocab causing this problem, and I have proposed that a second context be used to host all the securing specification related terms previously, which was also rejected.

I do still believe this would be a cleaner breakdown, and it would encourage implementers to follow better practices:

```
"@context": [
    (core data model) "https://www.w3.org/ns/credentials/v2",
    (securing data model) "https://www.w3.org/ns/credentials/vc-jose-cose/v2",
    (securing data model) "https://www.w3.org/ns/credentials/vc-data-integrity/v2",
    (industry terms) "https://www.w3.org/ns/credentials/examples/v2" (or other industry specific contexts)
  ],
```

StatusList and Credential  Schema would stay in `v2`.

`proof` and `DataIntegrityProof` would move to `vc-data-integrity`
`iss` and `kid` would move to `vc-jose-cose`

Perhaps another option not discussed would be to move the `@vocab` to `vc-jose-cose`, and bind it to IANA by default instead of W3C.

This would remove it from the v2 core context, and it would automatically map JOSE / COSE terms that are consistently registered correctly.

Since the thread is already very long, please ping me directly with any BLOCKING change requests that I have not yet addressed.

contexts/credentials/v2
OR13: Before I make this change, @davidlehn @iherman if I change this from:

https://www.iana.org/assignments/jwt/#iss (for all related fragments)

to

https://www.iana.org/assignments/jose/#iss (for all related fragments)

Will it resolve your concerns?

@iherman how would URLs in the normative v2 context block publication? 

Would the same apply to w3id.org URLs changing the underlying anchors?

OR13:
Tracking this as the only blocker:

- https://github.com/w3c/vc-data-model/pull/1149#discussion_r1264022299

selfissued (approved):

mprorock:
> StatusList and Credential Schema would stay in `v2`.
> 
> `proof` and `DataIntegrityProof` would move to `vc-data-integrity` `iss` and `kid` would move to `vc-jose-cose`
> 
> Perhaps another option not discussed would be to move the `@vocab` to `vc-jose-cose`, and bind it to IANA by default instead of W3C.

that feels like the right locations - additionally I think binding to IANA by default seems like something that merits discussion

gannan08:
@OR13

It's great to see your continued interest in Linked Data and trying to pave a path forward so that it can be used with JOSE terms. Unfortunately, I'm failing to see how exactly these terms will be used inside a Verifiable Credential.

Can you provide examples of VCs that will use these terms? If there aren't multiple examples, then one should suffice.

OR13:
@gannan08 sure, here are 2 examples, one from the IETF, the other from OpenID Connect:

- https://datatracker.ietf.org/doc/html/draft-ietf-oauth-selective-disclosure-jwt-05#name-example-4b-w3c-verifiable-c
- https://openid.net/specs/openid-connect-userinfo-vc-1_0.html#name-userinfo-vc-verification

And regardless of if this PR is merged, we might see lots of people doing this... 

This PR just makes the RDF graph less bad, which seems worth it to me.

Adding terms or removing terms from a context does not block or control what people do... 

It determines if what they have done will be understandable using the context we define (normatively) in this WG.

andresuribe87:
> StatusList and Credential Schema would stay in v2.
>
> proof and DataIntegrityProof would move to vc-data-integrity
iss and kid would move to vc-jose-cose

I support this. It's a good separation of concerns IMO. It's better than what we currently have.

gannan08:
> @gannan08 sure, here are 2 examples, one from the IETF, the other from OpenID Connect:
> 
> * https://datatracker.ietf.org/doc/html/draft-ietf-oauth-selective-disclosure-jwt-05#name-example-4b-w3c-verifiable-c

[Selective Disclosure for JWTs (SD-JWT)](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-selective-disclosure-jwt-05)

In the spec, there exists the following input claim set:

```json
{
  "@context": [
    "https://w3id.org/vaccination/v1"
  ],
  "type": [
    "VerifiableCredential",
    "VaccinationCertificate"
  ],
  "issuer": "https://example.com/issuer",
  "issuanceDate": "2023-02-09T11:01:59Z",
  "expirationDate": "2028-02-08T11:01:59Z",
  "name": "COVID-19 Vaccination Certificate",
  "description": "COVID-19 Vaccination Certificate",
  "credentialSubject": {
    "vaccine": {
      "type": "Vaccine",
      "atcCode": "J07BX03",
      "medicinalProductName": "COVID-19 Vaccine Moderna",
      "marketingAuthorizationHolder": "Moderna Biotech"
    },
    "nextVaccinationDate": "2021-08-16T13:40:12Z",
    "countryOfVaccination": "GE",
    "dateOfVaccination": "2021-06-23T13:40:12Z",
    "order": "3/3",
    "recipient": {
      "type": "VaccineRecipient",
      "gender": "Female",
      "birthDate": "1961-08-17",
      "givenName": "Marion",
      "familyName": "Mustermann"
    },
    "type": "VaccinationEvent",
    "administeringCentre": "Praxis Sommergarten",
    "batchNumber": "1626382736",
    "healthProfessional": "883110000015376"
  }
}
```
An SD-JWT is issued using the above claim set as an input. The payload of the SD-JWT is the following:

```json
{
  "iss": "https://example.com/issuer",
  "iat": 1683000000,
  "exp": 1883000000,
  "@context": [
    "https://w3id.org/vaccination/v1"
  ],
  "type": [
    "VerifiableCredential",
    "VaccinationCertificate"
  ],
  "issuer": "https://example.com/issuer",
  "issuanceDate": "2023-02-09T11:01:59Z",
  "expirationDate": "2028-02-08T11:01:59Z",
  "name": "COVID-19 Vaccination Certificate",
  "description": "COVID-19 Vaccination Certificate",
  "credentialSubject": {
    "_sd": [
      "1V_K-8lDQ8iFXBFXbZY9ehqR4HabWCi5T0ybIzZPeww",
      "JzjLgtP29dP-B3td12P674gFmK2zy81HMtBgf6CJNWg",
      "R2fGbfA07Z_YlkqmNZyma1xyyx1XstIiS6B1Ybl2JZ4",
      "TCmzrl7K2gev_du7pcMIyzRLHp-Yeg-Fl_cxtrUvPxg",
      "V7kJBLK78TmVDOmrfJ7ZuUPHuK_2cc7yZRa4qV1txwM",
      "b0eUsvGP-ODDdFoY4NlzlXc3tDslWJtCJF75Nw8Oj_g",
      "zJK_eSMXjwM8dXmMZLnI8FGM08zJ3_ubGeEMJ-5TBy0"
    ],
    "vaccine": {
      "_sd": [
        "1cF5hLwkhMNIaqfWJrXI7NMWedL-9f6Y2PA52yPjSZI",
        "Hiy6WWueLD5bn16298tPv7GXhmldMDOTnBi-CZbphNo",
        "Lb027q691jXXl-jC73vi8ebOj9smx3C-_og7gA4TBQE"
      ],
      "type": "Vaccine"
    },
    "recipient": {
      "_sd": [
        "1lSQBNY24q0Th6OGzthq-7-4l6cAaxrYXOGZpeW_lnA",
        "3nzLq81M2oN06wdv1shHvOEJVxZ5KLmdDkHEDJABWEI",
        "Pn1sWi06G4LJrnn-_RT0RbM_HTdxnPJQuX2fzWv_JOU",
        "lF9uzdsw7HplGLc714Tr4WO7MGJza7tt7QFleCX4Itw"
      ],
      "type": "VaccineRecipient"
    },
    "type": "VaccinationEvent"
  },
  "_sd_alg": "sha-256"
}
```

I see that the terms `iss`, `iat`, `exp` are not defined. Additionally, `_sd` and `_sd_alg` are not defined.

Since the original VC is being transformed into a different VC (credentialSubject is not preserved) the SD-JWT specification could go one step further and add an additional context that defines those terms.

IMO, this is the cleanest approach. It preserves the Spec Authors' desire to reuse the `iss`, `iat` and `exp` at the top level of the payload. It gives semantic meaning to the terms (`iss`, `iat`, `exp`, `_sd`, `_sd_alg` ) and most importantly gives them control over that meaning. This control will give the SD-JWT specification maximum flexibility. If any additional terms need to be added, then they can do so without needing to coordinate with the VCWG. 

My understanding is that there will be a point in time where we as the VCWG will freeze the context. If the SD-JWT Authors introduce a new term or rename an existing one after we publish our frozen context, then the same issue reappears: a term with unclear semantics.

```json
{
  "iss": "https://example.com/issuer",
  "iat": 1683000000,
  "exp": 1883000000,
  "@context": [
    "https://w3id.org/vaccination/v1",
    "https://sd-jwt-context-url.tld/v1" // new context url
  ],
  "type": [
    "VerifiableCredential",
    "VaccinationCertificate"
  ],
  "issuer": "https://example.com/issuer",
  "issuanceDate": "2023-02-09T11:01:59Z",
  "expirationDate": "2028-02-08T11:01:59Z",
  "name": "COVID-19 Vaccination Certificate",
  "description": "COVID-19 Vaccination Certificate",
  "credentialSubject": {
    "_sd": [
      "1V_K-8lDQ8iFXBFXbZY9ehqR4HabWCi5T0ybIzZPeww",
      "JzjLgtP29dP-B3td12P674gFmK2zy81HMtBgf6CJNWg",
      "R2fGbfA07Z_YlkqmNZyma1xyyx1XstIiS6B1Ybl2JZ4",
      "TCmzrl7K2gev_du7pcMIyzRLHp-Yeg-Fl_cxtrUvPxg",
      "V7kJBLK78TmVDOmrfJ7ZuUPHuK_2cc7yZRa4qV1txwM",
      "b0eUsvGP-ODDdFoY4NlzlXc3tDslWJtCJF75Nw8Oj_g",
      "zJK_eSMXjwM8dXmMZLnI8FGM08zJ3_ubGeEMJ-5TBy0"
    ],
    "vaccine": {
      "_sd": [
        "1cF5hLwkhMNIaqfWJrXI7NMWedL-9f6Y2PA52yPjSZI",
        "Hiy6WWueLD5bn16298tPv7GXhmldMDOTnBi-CZbphNo",
        "Lb027q691jXXl-jC73vi8ebOj9smx3C-_og7gA4TBQE"
      ],
      "type": "Vaccine"
    },
    "recipient": {
      "_sd": [
        "1lSQBNY24q0Th6OGzthq-7-4l6cAaxrYXOGZpeW_lnA",
        "3nzLq81M2oN06wdv1shHvOEJVxZ5KLmdDkHEDJABWEI",
        "Pn1sWi06G4LJrnn-_RT0RbM_HTdxnPJQuX2fzWv_JOU",
        "lF9uzdsw7HplGLc714Tr4WO7MGJza7tt7QFleCX4Itw"
      ],
      "type": "VaccineRecipient"
    },
    "type": "VaccinationEvent"
  },
  "_sd_alg": "sha-256"
}
```


> * https://openid.net/specs/openid-connect-userinfo-vc-1_0.html#name-userinfo-vc-verification

[OpenID Connect UserInfo Verifiable Credentials - Draft 00](https://openid.net/specs/openid-connect-userinfo-vc-1_0.html)

If the example I'm using is incorrect, please correct me. The link you provided took me to processing steps for UserInfo VC Verification. I scrolled further up to get the UserInfo VC.

(Note: I modified the example by adding a "," after `credentialSubject` and `credentialStatus.id`. Not doing so resulted in invalid JSON.)
```json
JWT Header:
{
  "alg": "ES256",
  "typ": "JWT",
  "kid": "XTSGmh734_J6fOWUbI7BNim7wyvj5LWx8GzuIH7WHw8"
}

JWT Payload:
{
  "vc": {
    "@context": [
      "https://www.w3.org/2018/credentials/v1"
    ],
    "type": [
      "VerifiableCredential",
      "UserInfoCredential"
    ],
    "credentialSubject": {
      "id": "did:jwk:eyJrdHkiOiJFQyIsInVzZSI6InNpZyIsImNydiI6IlAtMjU2IiwieCI6InFpR0tMd1hSSm1KUl9BT1FwV09IWExYNXVZSWZ6dlB3RHVyV3ZtWkJ3dnciLCJ5IjoiaXA4bnl1THBKNU5wcmlaekNWS2lHMFR0ZXFQTWtyemZOT1VROFl6ZUdkayIsImFsZyI6IkVTMjU2In0",
      "sub": "248289761001",
      "name": "Jane Doe",
      "given_name": "Jane",
      "family_name": "Doe",
      "preferred_username": "j.doe",
      "email": "janedoe@example.com",
      "picture": "http://example.com/janedoe/me.jpg",
      "phone_number": "+1 202 555 1212"
    },
    "credentialStatus": {
      "id": "https://server.example.com/credentials/status/3#94567",
      "type": "StatusList2021Entry",
      "statusPurpose": "revocation",
      "statusListIndex": "94567",
      "statusListCredential": "https://server.example.com/credentials/status/3"
    }
  },
  "iat": 1667575982,
  "exp": 1668180782,
  "iss": "https://server.example.com"
}
```

In this example `iat`, `exp`, and `iss` are siblings to the `vc` property. Any properties added to the VCWG context won't define those terms.

The undefined terms that I see are `UserInfoCredential`, `sub`, `name`, `given_name`, `family_name`,  `preferred_username`, `email`, `picture` and `phone_number`. Term definitions could be added via an additional context or one could choose not define those terms entirely. I don't believe the VCWG should define these terms in its context.

I believe `credentialStatus` and its properties are fine.

> And regardless of if this PR is merged, we might see lots of people doing this...

I agree, that's a possibility. 

> This PR just makes the RDF graph less bad, which seems worth it to me.

+1 For improving the RDF graph.

> Adding terms or removing terms from a context does not block or control what people do...

If there was a retweet button for this... I'd push a 1,000 times. In general, people are going to do what they want. I agree.

OR13:
In v2 none of these terms will be undefined, and all will be assigned a default URL that is issuer dependent.

The issuer assigned URL will say that the issuer did not define the term, even if it is registered with IANA and it is defined in an RFC... While this might be the convention at W3C, it's not the convention at IETF, or for JOSE or COSE... (These terms have meaning, that's not issuer dependent).

The graph produced will be inaccurate, and W3C will be mislabeling terms for 1 security format, while labeling terms correctly for another... This bias will probably lead to objections.

Perhaps the best solution is to use separate contexts for the securing formats.

If there is an objection to using `iss` or `sub` as defined by IANA or `DataIntegrityProof` as defined by w3id.org, the normative core data model context wont block publication of the core data model and it's terms, which are defined by W3C.

This was my original proposal.

Another alternative is to create a vc-jose-cose vocab that mirrors IANA, but publishes anchor tags (seems like a lot of work, that might not be necessary, hopefully we can get a confirmation from @iherman if this is truly required or not).

We could also use the existing vocabulary to define `iss` as being defined by IANA.

... Then repeat that for all the other terms IANA defines that are relevant identifiers... Again, a lot of work for no real gain, since `@id` does not need to be resolvable, in order to be valuable in the construction of the graph.

We would not be having this conversation at all, if we were not trying to bundle securing formats, credential schemas, and status list terms all into the V2 context.

Which could cause our whole spec to be blocked if any of those specs receive objections.

How should we approach resolving this?

contexts/credentials/v2
iherman: > Before I make this change, @davidlehn @iherman if I change this from:
>
> https://www.iana.org/assignments/jwt/#iss (for all related fragments)
>
> to
>
> https://www.iana.org/assignments/jose/#iss (for all related fragments)
>
> Will it resolve your concerns?

As far as i can see, it won't. The problem is identical: the target page does not have an `@id` value corresponding to the fragment in the URL.

> @iherman how would URLs in the normative v2 context block publication?

Because the context's _**content**_ is now normative, the publication rule applies: all URLs must resolve, and unresolved fragment ID-s are not o.k. This means that whoever would check these rules, whether it is a machine or one of the persons in the team checking the specification, would flag this.

> Would the same apply to w3id.org URLs changing the underlying anchors?

Yes. The general review of the security vocabulary is still waiting for us and yes, we are dreadfully behind schedule in my view...

iherman:
> Another alternative is to create a vc-jose-cose vocab that mirrors IANA, but publishes anchor tags (seems like a lot of work, that might not be necessary, hopefully we can get a confirmation from @iherman if this is truly required or not).

That would of course work. I do not know whether that is the solution or not but, as I said in https://github.com/w3c/vc-data-model/pull/1149/files#r1264368563, _something_ is to be done.

Another possible approach (I just came up with this right now, haven't given too much thought) is as follows. As far as I can remember (and that is where I may be wrong), the [current security vocabulary](https://w3c.github.io/vc-data-integrity/vocab/security/vocabulary.html) might include, eventually, terms that are _not_ defined in the [data integrity specification](https://www.w3.org/TR/vc-data-integrity/) but may be defined in one of the "satellite" specs like eddsa. Which is perfectly fine. But, if so, there may not be a reason to define a separate `vc-jose-cose` vocab; we could push these terms into the security vocabulary as well, and call it a day. (The "definition" of those terms may refer to the IANA specifications of course by just referring to the top level XHTML file.)

Yet another possible approach, if these terms are some sort of "mirrors" of existing vcdm terms, then we can just as well bite the bullet and introduce new vocabulary items in the vcdm vocabulary. 

Any of these approaches are viable, relatively easy to do in practice, and do not require any long existential debate. We should just define, swiftly, which way we want to go, and do it.

> We would not be having this conversation at all, if we were not trying to bundle securing formats, credential schemas, and status list terms all into the V2 context.

I disagree. If we define _any_ vocabulary term for our own JSON-LD environment, we are supposed to use URLs that are properly dereferencable including the fragment ID. This is a generic rule for W3C specifications. This is totally orthogonal to the question of _where_ these URL mappings are, i.e., in which `@context` file they appear. Let us not re-start this discussion.

contexts/credentials/v2
OR13: @iherman what happens if a context contains a URN? 

For example:  https://datatracker.ietf.org/doc/rfc9162/

`urn:ietf:params:trans:error:endBeforeStart`

Would publication be blocked?

msporny:
@OR13 wrote:
> Which could cause our whole spec to be blocked if any of those specs receive objections.

No, it wouldn't -- we always have the option of removing them IF we get FO's, at which point the FO's would need to be upheld (which would be very difficult to do given the number of implementations we already have). Additionally, we shouldn't let the process control architectural decisions. The entire contents of the context file is marked as at risk, we can change the contents at any point during the CR process, including removing/modifying terms defined in that context that are blocking PR publication.

IOW, we resolved this blocking concern a while ago. We also have consensus to make developers lives easier by defining useful things in the base context -- one of the arguments for `@vocab` in the base context was to reduce the number of contexts necessary to one. Given that we got consensus on that, breaking everything out into separate contexts will re-open that debate... because if now we're saying that everyone needs at least two contexts anyway to secure a VC, then we don't need `@vocab` in the base context.

OR13:
@msporny I agree with everything you wrote above.

Seems like the best path forward is as follows:

- Add all the terms we want to assign URLs or URNs to... to the https://www.w3.org/2018/credentials.
- Add URLs for them to the v2 context, for example:  https://www.w3.org/ns/credentials/iana#iss (maybe no need to subpath here)...
- Add links from the W3C Vocabulary to the associated IANA registries.

As an aside, I would be happy to do all this work, but I struggle to contribute to the current vocabulary, because I am unfamiliar with the custom tooling we use to build it: 

- https://www.w3.org/2018/credentials/ 
- its not publishes under TR space, but it is normative... is that ok?
- there is no link to github at the top of it (probably an easy fix).
- The instructions here are minimal: https://github.com/w3c/vc-data-model/tree/main/vocab/credentials

Given we currently have:

- https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v2#L4

This means the change would look like this:

- https://www.w3.org/ns/credentials/issuer-dependent#iss
- https://www.w3.org/ns/credentials/issuer-dependent#sub

(please click these links and read the current text, and assess if this is acceptable for a normative reference).

@Sakurann @selfissued @David-Chadwick @andresuribe87 

I think it would be nice to avoid the `2018` in the vc-jose-cose associated term definitions.

That also aligns better with how IANA works.

I suggest the following URLS which are shorter / more environmentally friendly :

- https://www.w3.org/ns/credentials#iss
- https://www.w3.org/ns/credentials#sub

Since these URLs have never be used, there will be no breaking changes associated with this improvement.

msporny:
@OR13 wrote:
> https://www.w3.org/ns/credentials#iss

That would look like the W3C VCWG is taking over the JOSE namespace -- which would be a very bad idea, IMHO. No, we should use URLs that the JOSE/COSE group approves of, and they should be the change controller on. Doing anything other than that would be overreaching (and would lead to all sorts of political strife that is easily avoided).

Just get the JOSE group to say which URLs (with fragment identifiers) they want to use for the definitions for `iss`, `sub`, etc.

OR13:
> No, we should use URLs that the JOSE/COSE group approves of, and they should be the change controller on. Doing anything other than that would be overreaching (and would lead to all sorts of political strife that is easily avoided).

What if the URLs they use don't care about W3C fragment processing rules? Can we still use them?

> Just get the JOSE group to say which URLs (with fragment identifiers) they want to use for the definitions for `iss`, `sub`, etc. We refer to those in our context files. Done.

Seems like thats setting us up for a blocking dependency that we can't resolve that will lead to the terms being removed.

As you said above, our process is sorta guiding our architecture here... But I don't see how to fix that, without W3C mirroring IANA...

This is what GS1 did for schema.org, see https://www.gs1.org/voc/Organization which points to https://schema.org/Organization

Or IANA could add fragments, or W3C could consider that URLS don't always resolve, but are still valid graph node identifiers...

https://www.w3.org/TR/json-ld/#node-identifiers

Given the definition of an IRI... it seems the arguments about resolution are not correct... Because IRIs have no defined resolution (only URLs do).

msporny:
> > No, we should use URLs that the JOSE/COSE group approves of, and they should be the change controller on. Doing anything other than that would be overreaching (and would lead to all sorts of political strife that is easily avoided).
> 
> What if the URLs they use don't care about W3C fragment processing rules? Can we still use them?

Let's not buy trouble from the future. Let's just wait for them to answer the question and go from there. There are a variety of ways through this maze... let's try to do the right thing first, and then when it looks like that's not going to work out, formulate a plan B (based on what we learn by asking the question of JOSE / IANA in the first place).

IOW, I'd be surprised if they refuse to add fragment identifiers to that IANA registry.

iherman:
> As you said above, our process is sorta guiding our architecture here... But I don't see how to fix that, without W3C mirroring IANA...

I think that "mirroring IANA" is a little bit of a strong terminology here, @msporny. IANA is oblivious to the usage of these ID-s in a Linked Data setting, and what we would do is to define an RDF related ID for those terms. We would not define the semantics, we would make a bridge between IANA and Linked Data. 

> 
> This is what GS1 did for schema.org, see https://www.gs1.org/voc/Organization which points to https://schema.org/Organization

@OR13, I do not know why GS1 did that, but the comparison is wrong. `https://schema.org/Organization` is a perfectly valid URL that does dereference to a page of its own. I.e., a processing engine (human or digital) can "follow its nose", i.e., dereference a URL to find out more of the term. This isn't the case if the fragid is wrong.

> Let's not buy trouble from the future. Let's just wait for them to answer the question and go from there. There are a variety of ways through this maze... let's try to do the right thing first, and then when it looks like that's not going to work out, formulate a plan B (based on what we learn by asking the question of JOSE / IANA in the first place).

+1 there. Let us ask IANA asap, and see if we are, in fact, chasing ghosts here because they would add those fragid in a blink of an eye.

> IOW, I'd be surprised if they refuse to add fragment identifiers to that IANA registry.

Hopefully you are right.

---

(Just for the fun of it, a crazy and probably nonsensical idea would be to make use of the fact that the target page is XHTML: maybe an XPath wizard could come up with an xpointer URL of the form `https://www.iana.org/assignments/jose#xpath(…)`. Alas! Xpointer and all the other XML tools have been buried and forgotten, so this would not have any practical sense... But it may be a nice thought experiment!😀)

OR13:
I am interested in this part:

>  a processing engine (human or digital) can "follow its nose", i.e., dereference a URL to find out more of the term. This isn't the case if the fragid is wrong.

Why is it not the case? Because a browser does not scroll to the fragment? isn't that making the web browser the "processing engine" ?

For example, if I search the page for the text `iss` using a browser, I immediately find the entry I am looking for... How is this not good enough for the "follow your nose" criteria, and what actual W3C process asserts this requirement?

contexts/credentials/v2
```diff
@@ -5,6 +5,28 @@
 
     "id": "@id",
     "type": "@type",
+    
+    "kid": {
+      "@id": "https://www.iana.org/assignments/jwt#kid",
```
OR13, line 10:
```suggestion
      "@id": "https://www.iana.org/assignments/jose#kid",
```

contexts/credentials/v2
```diff
@@ -5,6 +5,28 @@
...
+      "@type": "@id"
+    },
+    "iss": {
+      "@id": "https://www.iana.org/assignments/jwt#iss",
```
OR13, line 14:
```suggestion
      "@id": "https://www.iana.org/assignments/jose#iss",
```

contexts/credentials/v2
```diff
@@ -5,6 +5,28 @@
...
+      "@type": "@id"
+    },
+    "sub": {
+      "@id": "https://www.iana.org/assignments/jwt#sub",
```
OR13, line 18:
```suggestion
      "@id": "https://www.iana.org/assignments/jose#sub",
```

contexts/credentials/v2
```diff
@@ -5,6 +5,28 @@
...
+      "@type": "@id"
+    },
+    "jku": {
+      "@id": "https://www.iana.org/assignments/jwt#jku",
```
OR13, line 22:
```suggestion
      "@id": "https://www.iana.org/assignments/jose#jku",
```

contexts/credentials/v2
OR13, line 26:
```suggestion
      "@id": "https://www.iana.org/assignments/jose#x5u",
```

OR13:
@brentzundel @Sakurann A few proposals for the next opportunity:

PROPOSAL: The normative v2 context will not define any terms related to vc-jose-cose.

PROPOSAL: URLs to the IANA JOSE registry with fragments will be used to identify the terms related to vc-jose-cose in the normative v2 context.

PROPOSAL: URLs to the W3C Credentials Vocabulary will be used to identify the terms related to vc-jose-cose in the normative v2 context.

brentzundel:
@TallTed It looks like your requests for changes were all addressed, can you confirm?

contexts/credentials/v2
```diff
@@ -5,6 +5,26 @@
 
     "id": "@id",
     "type": "@type",
+    "kid": {
```
alenhorvat, line 8: @OR13 kid always appears in the header.
This means that the context should be used outside of the payload where it appears?

Same comment for all JWS claims.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-07-19-vcwg#section1-1) on 2023-07-19

- no resolutions were taken

#### 1.1. Add graph node identifiers for registered claims (pr vc-data-model#1149)

_See github pull request [vc-data-model#1149](https://github.com/w3c/vc-data-model/pull/1149)._




**Manu Sporny:** 1149 is waiting on TallTed to provide a review.  
… The PR maps certain terms to a registry.  

</details>

alenhorvat:
> Since the v2 context contains terms specific to data integrity proofs.
> 
> And I have closed the PR attempting to remove those terms: #1091
> 
> It stands to reason we will need to add terms related to all proof formats to the v2 context, in the spirit of making things easier for developers.

This is an important design decision. IMO the core vocabulary should not define securing-related vocabulary (except for the top level proof property). 

Why?
There are 3 signature "topologies"
- detached (e.g., compact serialised JWS)
- enveloping (JSON serialised JWS)
- enveloped (Data Integrity Proofs)

JWS (not JWT) and DIP are not adding any additional claims to the payload itself. DIP is defining and managing the claims inside of the proof property - that's a property of enveloped signatures.

Proposal is to keep the securing independent of the content to decouple the two. It is a design decision.
Vocabulary also contains some other default claims, such as statuslist2021, ...

One option would be to define a minimal vocabulary for the core claim set and a separate vocabulary for a "profile" that would contain definitions for that profile (JsonSchema2023, VerifiableCredentialSchema2023, StatusList2021Credential, StatusList2021, StatusList2021Entry, DataIntegrityProof)

Whether or not this is a viable proposal depends on the purpose and **design decision** of the core vocabulary. 

This is partially reopening the https://github.com/w3c/vc-data-model/pull/1091 discussion (partially)

If the decision of the WG is to keep the data integrity vocabulary in the VCDM specs, it would only be appropriate for the enveloped signatures, since for the detached and enveloped the securing-related claims reside outside of the payload.

TallTed (approved):

contexts/credentials/v2
OR13: `@context` is also allowed in JOSE headers, when its present, should the terms have definitions like this or not?

The core data model states a "VerifiableCredential" has "proof" and "metadata"... headers and signatures seem like proof and metadata to me... also... `kid` appears in payloads... see:

https://datatracker.ietf.org/doc/html/draft-ietf-oauth-selective-disclosure-jwt-05#appendix-A.3

https://www.rfc-editor.org/rfc/rfc7800.html#section-3.4

```
     {
      "iss": "https://server.example.com",
      "sub": "17760704",
      "aud": "https://client.example.org",
      "exp": 1440804813,
      "cnf":{
        "jku": "https://keys.example.net/pop-keys.json",
        "kid": "2015-08-28"
       }
     }
```

Omitting `kid` from this PR would cause `cnf` "RDF graph" to be build improperly for key bound SD-JWT Verifiable Credentials that are trying to leverage JSON-LD.

OR13:
> This is an important design decision. IMO the core vocabulary should not define securing-related vocabulary (except for the top level proof property).

I agree with you, but the working group could not establish consensus to split them up, and has defaulted to bundling them all together (as we did in v1 and v1.1).

I guess the question remains, do we reopen that debate and try to split things up, especially in light of:

https://github.com/w3c/vc-data-model/issues/1175

Which was opened after I capitulated to the group regarding this design choice last time?

OR13:
@davidlehn I believe your suggestions have been addressed, can you re-review?

@iherman Are you objecting to merging this PR, can you address my question on https://github.com/w3c/vc-data-model/pull/1149#discussion_r1264396199

@alenhorvat Lets move the debate you are reopening to #1175 , are you willing to approve the PR while we continue to discuss how we might factor these terms into separate contexts in the future?

alenhorvat:
> @alenhorvat Lets move the debate you are reopening to #1175 , are you willing to approve the PR while we continue to discuss how we might factor these terms into separate contexts in the future?

Yes, we can continue the discussion under #1175. As said, it is a design decision and my purpose is not to block any PR.

contexts/credentials/v2
alenhorvat: Ah, I see. Fine. Let's continue under 1175 -- it's just a matter of keeping everything in 1 file or splitting it per securing mechanism.

Note that SD-JWT is defining only _sd, _sd_alg, and is not requiring cnf or any jwt claims; hence these should be defined in a separate document. (which was not the case in the previous version of sd-jwt specs)

wrt context in the jose headers -- let's discuss this in the context of JAdES that introduced custom header claims

msporny:

PR #1140, opened 2023-06-02
add context integrity capabilities to the core data model

mprorock:
per conversation and feedback [on this pr on vc-jwt](https://github.com/w3c/vc-jwt/pull/90) this is probably better suited to be available in the core data model for all securing formats

***
<a href="https://pr-preview.s3.amazonaws.com/mesur-io/vc-data-model/pull/1140.html" title="Last updated on Jun 27, 2023, 4:05 PM UTC (a4ef5eb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1140/cff30d3...mesur-io:a4ef5eb.html" title="Last updated on Jun 27, 2023, 4:05 PM UTC (a4ef5eb)">Diff</a>

Issue #831, opened 2021-11-02
Proposal: Anchored Resources and Hashlinks for VCs


dmitrizagidulin:
Sidenote: I debated for a while where to open this issue (whether on the overall CCG repo, or the VC Dev Guide, or here), and decided to go with the VC Data Model spec repo, only because the proposal involves adding two new terms to the VC `@context`. This issue is meant to be a start of the discussion, and please let me know if this proposal should go somewhere else.

### Motivation
1. Provide a mechanism with which to link multiple VCs together (as peers or as parent/child relationships).
2. Provide examples of a general purpose [hashlink](https://datatracker.ietf.org/doc/html/draft-sporny-hashlink) mechanism (useful for linking VCs, but also for linking to external images, PDFs, and other files).

### Background / Use Cases
There are many use cases that involve binding multiple Verifiable Credentials together. For example, a simple Student ID credential can consist of an overall container credential, which links to several individual credentials (such as a student Photo credential, a proof of enrollment at a particular university, a proof of age, etc). 

Note that this is different than binding multiple credentials together in a Verifiable Presentation (and having the presenter sign the VP). In the VP case, the binding just means "this presenter is authenticating the handing over of these unrelated credentials". Whereas in the linked VC case, the credentials are aware of each other, and the peer or hierarchical relationship is built into the VC itself.

### Proposal
This proposal introduces two new related properties: `anchoredResource` and `contentHash`.

#### Anchored Resource
An anchored resource, points to one or more linked or bound resources. It can appear in the `credentialSubject` section (which implies that the resource is linked to the subject), or in the top level credential (same level as issuer, issuanceDate, etc), which implies that the resource is linked to the credential itself.

#### Content Hash
A `contentHash` property provides a way to refer to an external resource via a Hashlink.

#### Example
Here's what this would look like, using a mock Student ID credential.
```js
{
  "@context": [
    "https://www.w3.org/2018/credentials/v1",
    "https://w3id.org/security/suites/ed25519-2020/v1"
  ],

  "id": "urn:uuid:123445",
  "type": ["VerifiableCredential", "StudentIdCredential"],
  "issuer": {
    "id": "did:key:....",
    "image": {      "id": "https://example-university.edu/logo.png",
     "contentHash": "hl:<hashlink to the logo url above>"
    },
    "name": "Example University"
  },
  "issuanceDate": "2020-04-03T00:00:00.000Z",
  "expirationDate": "2020-12-15T00:00:00.000Z",
[...]

PR #1158, opened 2023-06-19
Make the value of the base context normative.

msporny:
This PR implements two resolutions made during the last special topic call:

1. https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-06-13-vcwg-special#resolution1
2. https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-06-13-vcwg-special#resolution2

Namely:

[Resolution #1](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-06-13-vcwg-special#resolution1): The v2 context URL will remain normative (https://www.w3.org/ns/credentials/v2), its value will be made normative through the use of a hash digest.

[Resolution #2](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-06-13-vcwg-special#resolution2): Add issue markers saying that the value of the hash digest for the v2 context may change before PR and that’s expected.

One editorial fix the Editor's might make after this PR is to move the Base Context section from the appendix into the main specification (given that having a normative appendix section is a bit strange).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1158.html" title="Last updated on Jun 30, 2023, 9:15 AM UTC (b84134c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1158/9fea3cc...b84134c.html" title="Last updated on Jun 30, 2023, 9:15 AM UTC (b84134c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4928,13 +4928,21 @@ franchise. Policy information expressed by the <a>issuer</a> in the
     <section class="appendix informative">
       <h2>Contexts, Types, and Credential Schemas</h2>
 
-      <section class="informative">
+      <section class="normative">
         <h3>Base Context</h3>
+
+        <p class="issue" title="(AT RISK) Hash values might change during Candidate Recommendation">
+This section lists cryptographic hash values that might change during the
+Candidate Recommendation phase based on implementer feedback that requires
+the referenced files to be modified.
+        </p>
           <p>
-The base context, located at <code>https://www.w3.org/ns/credentials/v2</code>
-with a SHA-256 digest of
-<strong><code>ff1fdfa8d3f07547ec149b6652fe3f5a4276b9f0c48344dc29400763ba7b44fa</code></strong>,
-can be used to implement a local cached copy. It is possible to confirm the
+Implementations MUST ensure that the base context value, located at
+<code>https://www.w3.org/ns/credentials/v2</code>, matches the following
+SHA-256 digest of the value:
+<strong><code>dc28cf86d8a3199beb9d14e29d1da662346b8660abf2edca873098c88ca2a457</code></strong>.
+The base context value matching the digest previously stated can be used to
+implement a local cached copy. It is possible to confirm the
 SHA-256 digest by running the following command from a modern Unix command
 interface line:
 `curl -s https://www.w3.org/ns/credentials/v2 | openssl dgst -sha256 -hex -r`.
--- a/index.html
+++ b/index.html
@@ -4947,7 +4947,7 @@ SHA-384 digest by running the following command from a modern Unix command
 interface line:
 `curl -s https://www.w3.org/ns/credentials/v2 | openssl dgst -sha384 -binary | openssl base64 -A`.
 
-More details regarding this hash encoding and methods may be found in the <a href="https://www.w3.org/TR/SRI/#integrity-metadata">integrity metadata</a> 
+More details regarding this hash encoding method may be found in the <a href="https://www.w3.org/TR/SRI/#integrity-metadata">integrity metadata</a> 
 section of [[SRI]].
 It is strongly advised that all JSON-LD Contexts used in an application utilize
 a similar mechanism to ensure end-to-end security.
--- a/index.html
+++ b/index.html
@@ -4946,7 +4946,10 @@ implement a local cached copy. It is possible to confirm the
 SHA-384 digest by running the following command from a modern Unix command
 interface line:
 `curl -s https://www.w3.org/ns/credentials/v2 | openssl dgst -sha384 -binary | openssl base64 -A`.
-
+                </p>
+                <p class="issue" data-number="1177">
+The Working Group is currently discussing what a processor should do if a hash value differs from one that is listed in the specification. 
+                </p>
 More details regarding this hash encoding method may be found in the <a href="https://www.w3.org/TR/SRI/#integrity-metadata">integrity metadata</a> 
 section of [[SRI]].
 It is strongly advised that all JSON-LD Contexts used in an application utilize
--- a/index.html
+++ b/index.html
@@ -4939,13 +4939,16 @@ the referenced files to be modified.
           <p>
 Implementations MUST ensure that the base context value, located at
 <code>https://www.w3.org/ns/credentials/v2</code>, matches the following
-SHA-256 digest of the value:
-<strong><code>dc28cf86d8a3199beb9d14e29d1da662346b8660abf2edca873098c88ca2a457</code></strong>.
+SHA-384 digest of the value computed and encoded according to the [[SRI]] definition of `digest`:
+<strong><code>lHKDHh0msc6pRx8PhDOMkNtSI8bOfsp4giNbUrw71nXXLf13nTqNJoRp3Nx+ArVK</code></strong>.
 The base context value matching the digest previously stated can be used to
 implement a local cached copy. It is possible to confirm the
-SHA-256 digest by running the following command from a modern Unix command
+SHA-384 digest by running the following command from a modern Unix command
 interface line:
-`curl -s https://www.w3.org/ns/credentials/v2 | openssl dgst -sha256 -hex -r`.
+`curl -s https://www.w3.org/ns/credentials/v2 | openssl dgst -sha384 -binary | openssl base64 -A`.
+
+More details regarding this hash encoding and methods may be found in the <a href="https://www.w3.org/TR/SRI/#integrity-metadata">integrity metadata</a> 
+section of [[SRI]].
 It is strongly advised that all JSON-LD Contexts used in an application utilize
 a similar mechanism to ensure end-to-end security.
           </p>


iherman (approved):
I am fine with the changes proposed in this PR, hence my approval.

However, with regard to your comment:

> One editorial fix the Editor's might make after this PR is to move the Base Context section from the appendix into the main specification (given that having a normative appendix section is a bit strange).

I am a -1 to that. There is a difference between saying that the URL of the context file is "normative" (also reinforced through the hash value) and listing the content of the context file in the specification, which makes every line of the context normative as well. This leads to the sort of confusion that I have described, for example, in https://github.com/w3c/vc-data-model/issues/1103#issuecomment-1590873676: 

> If we relied on `@context` […] as the source of that normative information, then we hit a bunch of additional questions like: what is the status of the mapping of `nonce` to `https://w3id.org/security#nonce` v.a.v. the security specification? How do we account for mappings like `name` to `https://schema.org/name`? In essence: which part of the `@context` is normative and which one is not? Etc.

My proposal is _not_ to include the context file in the specification _at all_. The URL of the context file is de-referencable, and we can also be sure that one will download the right file if the hash value is checked. By keeping only the reference in the spec, i.e., by being silent about the normativeness of the _content_ of the `@context` file, we avoid all these issues, we avoid the set of open PR-s that resulted from that confusion, etc.

My 2 cents...

TallTed (approved):
lgtm

index.html
```diff
@@ -4928,13 +4928,21 @@ <h3>Fitness for Purpose</h3>
     <section class="appendix informative">
       <h2>Contexts, Types, and Credential Schemas</h2>
 
-      <section class="informative">
+      <section class="normative">
         <h3>Base Context</h3>
+
+        <p class="issue" title="(AT RISK) Hash values might change during Candidate Recommendation">
```
decentralgabe, line 4934: is there an issue we can tag to tack this?

OR13 (approved):

mprorock (approved):

index.html
msporny: I don't think it's necessary to create an issue to track this statement. It creates more work for everyone w/o really having any upside. What's the issue going to say other than "the hash values might change" -- there's nothing to be discussed/debated there -- it's just a statement of fact.

Not every issue marker needs to map to an issue. This is just guidance to readers and implementers.

brentzundel (approved):

brentzundel:
I agree that having a normative appendix is weird. Could this language be added instead to the `@context` section?
Just a thought, I approve either way.

msporny:
> I agree that having a normative appendix is weird. Could this language be added instead to the `@context` section?

Yes, excellent idea... that's probably the ideal location for this text. I will update the PR to move the text to that location.

index.html
```diff
@@ -4928,13 +4928,21 @@ <h3>Fitness for Purpose</h3>
...
+        </p>
           <p>
-The base context, located at <code>https://www.w3.org/ns/credentials/v2</code>
-with a SHA-256 digest of
-<strong><code>ff1fdfa8d3f07547ec149b6652fe3f5a4276b9f0c48344dc29400763ba7b44fa</code></strong>,
-can be used to implement a local cached copy. It is possible to confirm the
+Implementations MUST ensure that the base context value, located at
+<code>https://www.w3.org/ns/credentials/v2</code>, matches the following
+SHA-256 digest of the value:
+<strong><code>dc28cf86d8a3199beb9d14e29d1da662346b8660abf2edca873098c88ca2a457</code></strong>.
```
OR13, line 4943: Use the same approach for data integrity and context integrity see https://github.com/w3c/vc-data-model/pull/1140

OR13: https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v2#L199

mprorock: > Rational for "why" alignment matters.
> 
> Implementers are required to load dependencies to support normative statements, when implementing the specification.
> 
> It should be a design goal of this working group to reduce the amount of vendor specific libraries (non standard implementations) that an implementation must use to implement this standard.
> 
> If implementers are already required to load a library that supports multibase (even a trivial library implemented by a startup), then they should be able to use that library to implement support for normative requirements the same way, everywhere they show up.
> 
> Here is a counter example (showing what not to do):
> 
> * v2 core context secured with sha512 hex encoded.
> * resourceIntegrity feature secured with sha384 (base64url encoded)
> * proofValue in data integrity secured with multibase (base58btc encoded)

i agree in principle and think that there is a desire to have this pr (or a follow on to it) to align with #1140 and vice versa.  This would mean that our primary normative doc (the spec itself) is consistent.


regarding the reference to proof value (i believe only a reference from core data perspective in the context) i think that this should be handled separately (e.g. what goes in the context since we agreed it is normative?) 

Broadly, i think the questions of encodings for data integrity and what is done under the data integrity spec belong there.

OR13: > (e.g. what goes in the context since we agreed it is normative?)

I agree, but unfortunately, we are *making the context normative with logs of things already in it*... so your solution does not work.

> Broadly, i think the questions of encodings for data integrity and what is done under the data integrity spec belong there.

I also agree with this, but the working group disagreed, when I attempted to separate things.

So now we have to make them consistent, or someone else needs to try to separate them in a way so that they can diverge without causing further implementer burden.

mprorock, line 4943:
```suggestion
SHA-384 digest of the value computed and encoded according to the [[SRI]] definition of `digest`:
<strong><code>lHKDHh0msc6pRx8PhDOMkNtSI8bOfsp4giNbUrw71nXXLf13nTqNJoRp3Nx+ArVK</code></strong>.
```

index.html
```diff
@@ -4928,13 +4928,21 @@ <h3>Fitness for Purpose</h3>
...
+The base context value matching the digest previously stated can be used to
+implement a local cached copy. It is possible to confirm the
 SHA-256 digest by running the following command from a modern Unix command
 interface line:
 `curl -s https://www.w3.org/ns/credentials/v2 | openssl dgst -sha256 -hex -r`.
```
mprorock, line 4948:
```suggestion
The base context value matching the digest previously stated can be used to
implement a local cached copy. It is possible to confirm the
SHA-384 digest by running the following command from a modern Unix command
interface line:
`curl -s https://www.w3.org/ns/credentials/v2 | openssl dgst -sha384 -binary | openssl base64 -A`.

More details regarding this hash encoding and methods may be found in the <a href="https://www.w3.org/TR/SRI/#integrity-metadata">integrity metadata</a> sectoion of [[SRI]].
```

mprorock (commented):
@msporny minor adjustments suggested here to bring things in line with SRI encoding of digest values and examples in [integrity metadata](https://www.w3.org/TR/SRI/#integrity-metadata) - also brings this in line with latest updates ([3ecd1b8](https://github.com/w3c/vc-data-model/pull/1140/commits/3ecd1b840f67fbec2a706131b6e2e0193023f256)) on #1140

msporny:
> @msporny minor adjustments suggested here to bring things in line with SRI encoding of digest values and examples in [integrity metadata](https://www.w3.org/TR/SRI/#integrity-metadata) - also brings this in line with latest updates ([3ecd1b8](https://github.com/w3c/vc-data-model/pull/1140/commits/3ecd1b840f67fbec2a706131b6e2e0193023f256)) on #1140

I'm fine w/ the changes and will approve when @OR13 accepts these changes.

mprorock:
> @msporny minor adjustments suggested here to bring things in line with SRI encoding of digest values and examples in [integrity metadata](https://www.w3.org/TR/SRI/#integrity-metadata) - also brings this in line with latest updates ([3ecd1b8](https://github.com/w3c/vc-data-model/pull/1140/commits/3ecd1b840f67fbec2a706131b6e2e0193023f256)) on #1140

OR13:
I assume the base64 vs base64 url is correct?

index.html
TallTed: @mprorock — `s/sectoion/section/` in line 4950 (which should also become 3 lines, to conform with other line lengths)

index.html
TallTed: @OR13 — `s/logs of things already in it/lots of things already in it/` (that is, `logs`->`lots`)?

index.html
```diff
@@ -4928,13 +4928,21 @@ <h3>Fitness for Purpose</h3>
...
+the referenced files to be modified.
+        </p>
           <p>
-The base context, located at <code>https://www.w3.org/ns/credentials/v2</code>
-with a SHA-256 digest of
-<strong><code>ff1fdfa8d3f07547ec149b6652fe3f5a4276b9f0c48344dc29400763ba7b44fa</code></strong>,
-can be used to implement a local cached copy. It is possible to confirm the
+Implementations MUST ensure that the base context value, located at
+<code>https://www.w3.org/ns/credentials/v2</code>, matches the following
+SHA-256 digest of the value:
```
jyasskin, line 4942: What are implementations supposed to do if they don't match?

msporny: For the base context... almost certainly throw an error (if you are not using a cached copy, which all VC processors should be using).

@jyasskin, we did have approval to merge this on a call today, would you prefer that we don't and address this particular concern?

jyasskin: I don't have any opinion about the timing of fixing this (i.e. feel free to merge and just queue the issue), but I think the spec will need to eventually tell the processors what to do instead of just saying they need to check. This might need a change to 5.2.5 in https://www.w3.org/TR/json-ld-api/#algorithm to make it possible to check contexts against a hash value when fetching them, or you could possibly require that all implementations treat this URL as "previously dereferenced" (step 5.2.4) with a particular value, in which case you don't need the hash at all.

index.html
```diff
@@ -4939,13 +4939,16 @@ <h3>Base Context</h3>
...
+<strong><code>lHKDHh0msc6pRx8PhDOMkNtSI8bOfsp4giNbUrw71nXXLf13nTqNJoRp3Nx+ArVK</code></strong>.
 The base context value matching the digest previously stated can be used to
 implement a local cached copy. It is possible to confirm the
-SHA-256 digest by running the following command from a modern Unix command
+SHA-384 digest by running the following command from a modern Unix command
 interface line:
-`curl -s https://www.w3.org/ns/credentials/v2 | openssl dgst -sha256 -hex -r`.
+`curl -s https://www.w3.org/ns/credentials/v2 | openssl dgst -sha384 -binary | openssl base64 -A`.
+
+More details regarding this hash encoding and methods may be found in the <a href="https://www.w3.org/TR/SRI/#integrity-metadata">integrity metadata</a>
```
TallTed, line 4950: This sentence puzzles me. `this hash encoding and methods` might should be `these hash encoding and methods` or `this hash encoding method` or something else. I'm guessing the latter, but could be convinced otherwise.
```suggestion
More details regarding this hash encoding method may be found in the <a href="https://www.w3.org/TR/SRI/#integrity-metadata">integrity metadata</a> 
```

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-06-27-vcwg-special#section1-1) on 2023-06-27

- no resolutions were taken

#### 1.1. Make the value of the base context normative. (pr vc-data-model#1158)

_See github pull request [vc-data-model#1158](https://github.com/w3c/vc-data-model/pull/1158)._




**Michael Prorock:** Suggest a quick diversion to 1158 since its related to 1140. There are a couple change I suggested to bring the language into alignment. If we are talking about hashing.  
… we should use the same mechanism with the core data model.  
… 1158 as is is inconsistent with 1140. I provided text to make those two consistent.  

**Kristina Yasuda:** lets talk about 1158.  

**Manu Sporny:** fine with these changes. Concern that it sounds like an argument is constructed based on these changes that puts us on a slippery slope.  
… This changes argues over how we express the hash as base64 or hexidecimal. We can make this change. +1, but I have a feeling its not the last we will hear about this.  
… Is this part of a larger suggest to change the way we represent hashes and other values across our specifications.  

**Kristina Yasuda:** How is this a slippery slope.  

**Manu Sporny:** The next argument could make that all DI specification need to follow the SRI encoding format.  

> *Orie Steele:* CEPC lets assume positive intent.

> *Dave Longley:* -1 to doing anything that would prohibit or align the use of digestMultibase.

**Kristina Yasuda:** clarify that the scope is only the VC DM. Would not apply to data integrity.  

> *Dave Longley:* +1 to Orie, let's assume positive intent, there's no goal here to prohibit or malign digestMultibase through this action.

**Michael Prorock:** This PR is solely to bring it up to 384 and encode the digest according to the SRI spec. this is focused just on the core data model.  

> *Manu Sporny:* As long as what MikeP is saying holds, we're all good here. :).

> *Kristina Yasuda:* thanks Orie and dave re positive intent comments.

> *Orie Steele:* I am still unsure if we can normatively depend on multiformats / multibase in a W3C TR... but that is not what resource integrity or securing the core context is about.

> *Dave Longley:* Orie: we can, and it's already been done in the past (e.g., DID core).

**Orie Steele:** agree with mike prorock. When we look at resource integrity for context, context is a resource. we found a way to make the context values normative.  
… by specifying a cryptographic hash function. many ways to represent hashes. Data integrity proof relies on multiformat. totally separate from describing how to secure resource securely.  
… in the core data model.  

> *Manu Sporny:* Good, I'm glad we have confirmation from both Orie and MikeP on those points.

> *Manu Sporny:* We're good to merge.

> *Manu Sporny:* (afaict).

> *Dave Longley:* What Orie said is why this was so related to #831 -- and does some of what it was trying to do.

> *Orie Steele:* I suspect we may see formal objections in the "hash" is used to make "down refs" normative, but we will see.

**Kristina Yasuda:** PR 1158 we have agreement to align with 1140.  
… we have 15 minutes left. Any other PRs related to this topic?  

> *Orie Steele:* Can we close the PRs that are not getting consensus?

**Kristina Yasuda:** are we OK to bring up 1100 and 1101?  

**Brent Zundel:** Can we get a sense of where each of the PRs are? we have 15 open PRs. Can we go through each of the PRs briefly?  

**Kristina Yasuda:** WIll take up PRs where we have authors on the call.  

</details>

index.html
mprorock: totally fine with your suggestion here

OR13 (approved):
Assuming we will resolve what having a normative context with the current values in it means, for processors.

Assuming we have issue markers for addressing @jyasskin comments.

index.html
OR13: If the URL is previously dereferenced, it doesn't need to be a URL, it could be a URN.

Context needs to dereference to the same bytes... Context identifier does not actually matter... 

Using a URN, or some other IRI would also protect W3C from being forced to serve the JSON-LD file that is always assumed to be previously dereferenced over and over again.

Making the value a URL encourages traffic, and implies it will be resolvable forever... It's not just a security thing, there is a cost / sustainability component to how this is made normative as well.

dlongley: @OR13,

> If the URL is previously dereferenced, it doesn't need to be a URL, it could be a URN.

We've had this discussion before (setting aside also that a "URN" is now also a "URL" per WHATWG). To increase interoperability in the linked data ecosystem, it should be a URL. Only the people who come, read, and implement the VCDM spec will be able to (and should be required to, IMO) treat the context as already dereferenced. We should not block other processors written by people who haven't read the spec from being able to consume VC data as general linked data.

index.html
TallTed: This needs to be committed before the PR is merged...

index.html
```diff
@@ -4928,16 +4928,27 @@ <h3>Fitness for Purpose</h3>
...
+<code>https://www.w3.org/ns/credentials/v2</code>, matches the following
+SHA-384 digest of the value computed and encoded according to the [[SRI]] definition of `digest`:
+<strong><code>lHKDHh0msc6pRx8PhDOMkNtSI8bOfsp4giNbUrw71nXXLf13nTqNJoRp3Nx+ArVK</code></strong>.
+The base context value matching the digest previously stated can be used to
+implement a local cached copy. It is possible to confirm the
+SHA-384 digest by running the following command from a modern Unix command
 interface line:
-`curl -s https://www.w3.org/ns/credentials/v2 | openssl dgst -sha256 -hex -r`.
+`curl -s https://www.w3.org/ns/credentials/v2 | openssl dgst -sha384 -binary | openssl base64 -A`.
+
```
msporny, line 4949:
```suggestion
                </p>
                <p class="issue" data-number="1177">
The Working Group is currently discussing what a processor should do if a hash value differs from one that is listed in the specification. 
                </p>
```

msporny:
Normative, multiple reviews, changes requested and made (issues raised for unresolved issues), no objections, merging.

Issue #1027, opened 2023-02-03
Add guidance on ignoring optional fields


OR13:
Originally reported on: https://github.com/w3c/vc-data-model/pull/987#issuecomment-1413789794

@David-Chadwick noted that there are many "optional" fields of a W3C Credential, which can safely be ignored by a verifier.

I would like to explicitly state this in normative language, to remove confusion of ambiguous processing.

`credentialStatus`, `ToU`, `Evidence`, `validity period`, `Status`... MUST NOT result in a verification or validation failure if not understood by the issuer or verifier.

^ Probably not correct, but hopefully triggering enough to bikeshed some better text on the issue.

David-Chadwick:
This is where X.509 has an edge over VCs. It has the critical  extensions flag, set by the issuer, which effectively says "if you do not understand this field then you MUST fail the verification" OTOH, if the flag is not set, if you do understand this field then you must obey it, but if you do not understand it you may safely ignore it.
@OR13 Please note the difference between verification and validation. The verification procedure should always act on all the optional fields and report the result to the application, which will then decide, during validation, whether to ignore the result of the verification or not. So an unverifiable VC may be accepted at validation time, whilst a verified VC may be rejected at validation time. But the application layer needs to know that the verification has been performed correctly and deterministically and not randomly.

David-Chadwick:
CONCRETE PROPOSAL
we should add a Critical boolean to every optional metadata field so that the verifier knows whether it can ignore fields it does not understand.

OR13:
JOSE / COSE have the `crit` header... 

- https://www.rfc-editor.org/rfc/rfc7515#section-4.1.11
- https://www.rfc-editor.org/rfc/rfc8152.html#section-3.1

David-Chadwick:
@OR13 Thanks for these pointers. The deficiency (or maybe not) in these definitions is that it does not explicitly say what a recipient must do with an extension that is not marked critical. Maybe the specs are relying on each extension to explicitly state this, whereas X.509 made a blanket statement that applied to all extensions.

Note also that the proposal I am making is to the VCDM and not to the proof method. This new Critical boolean should be independent of the proofing mechanism and should apply to all of them.

dlongley:
Linking in the discussion from the first VCWG on "critical" fields: https://github.com/w3c/vc-data-model/issues/158

Sakurann:
+1 to adding a language that says "optional field do not have to be present for the credential to be valid".

[...]

PR #1168, opened 2023-06-26
made validFrom optional, fixes #1133

### LINKED ISSUE #1133: Make `validFrom` optional
I cannot see why `validFrom` has to be mandatory in the VCDM. Especially since securing mechanism take care of the validity period of the signature. IMO, we should make it OPTIONAL.

Not every VC will have a `validFrom` property. For example, it is even questionable what would be the `validFrom` of a digital driver's license since most DMVs just define the issuance date which has a different semantic meaning than `validFrom`. 

Note that ISO 18013-5 for instance uses `issue_date` (and `expiry_date`) to express validity periods. `issue_date` is the date when the mDL was issued, not when it stated to become valid.

awoie:
closes #1133, made `validFrom` optional (also fixed JSON schema)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1168.html" title="Last updated on Jul 5, 2023, 4:00 PM UTC (c75a122)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1168/7c44300...c75a122.html" title="Last updated on Jul 5, 2023, 4:00 PM UTC (c75a122)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1724,6 +1724,14 @@ at which the information associated with the <code>credentialSubject</code>
 }
         </pre>
 
+        <p class="note">
+If <code>validFrom</code> and <code>validUntil</code> are not present, the
+<a>verifiable credential</a> validity period is considered valid
+indefinitely. In such cases, the <a>verifiable credential</a> is assumed to be
+valid from the time the <code>proof</code> was created or activated until
+further notice or until it is explicitly revoked or invalidated.
+        </p>
+
       </section>
 
       <section>
--- a/index.html
+++ b/index.html
@@ -1682,10 +1682,9 @@ when a <a>credential</a> ceases to be valid.
         <dl>
           <dt><var>validFrom</var></dt>
           <dd>
-A <a>credential</a> MUST have an <code>validFrom</code> <a>property</a>. The
-value of the <code>validFrom</code> <a>property</a> MUST be a string value of
-an [<a data-cite="XMLSCHEMA11-2#dateTime">XMLSCHEMA11-2</a>] combined
-<code>date-time</code> string representing the date and time the
+If present, the value of the <code>validFrom</code> <a>property</a> MUST be
+a string value of an [<a data-cite="XMLSCHEMA11-2#dateTime">XMLSCHEMA11-2</a>]
+combined <code>date-time</code> string representing the date and time the
 <a>credential</a> becomes valid, which could be a date and time in the future.
 Note that this value represents the earliest point in time
 at which the information associated with the <code>credentialSubject</code>
--- a/index.html
+++ b/index.html
@@ -1728,8 +1728,7 @@ at which the information associated with the <code>credentialSubject</code>
 If <code>validFrom</code> and <code>validUntil</code> are not present, the
 <a>verifiable credential</a> validity period is considered valid
 indefinitely. In such cases, the <a>verifiable credential</a> is assumed to be
-valid from the time the <code>proof</code> was created or activated until
-further notice or until it is explicitly revoked or invalidated.
+valid from the time the <code>verifiable credential</code> was created.
         </p>
 
       </section>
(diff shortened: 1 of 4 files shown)


brentzundel (approved):

dlongley (approved):

decentralgabe (approved):

msporny (commented):
I'm a bit unsure about this PR, but not going to block it. Curious to understand the use case where one can't specify a `validFrom` date? I'm not sure how we're supposed to interpret a VC that doesn't have a `validFrom` date.

If `validFrom` is not present, but the proof clearly has a date on which it was created, can the verifier presume that as the `validFrom` date?  If they can't presume that, then is the VC valid for all time leading up to the `validUntil` date? If there is no `validFrom` or `validUntil` date, then the VC is valid for all time?

If we have answers to the questions above, we should write them down in this PR. If we don't, then I'm a bit more concerned about the ecosystem issuing VCs that don't have any validity periods, or how those validity periods are going to be interpreted differently by different verifiers. Feels like an interop problem?

OR13 (approved):

msporny:
@msporny wrote:
> I'm a bit unsure about this PR, but not going to block it if we have a good use case. 

@awoie, ping. Any response on the use case request and questions asked above? 

Also, there are now merge conflicts that need to be resolved.

awoie:
> I'm a bit unsure about this PR, but not going to block it if we have a good use case. Curious to understand the use case where one can't specify a `validFrom` date? I'm not sure how we're supposed to interpret a VC that doesn't have a `validFrom` date.

IMO, it is unnecessary in a lot of cases. IMO, we have the chance to remove bloat and `validFrom` is bloat in a lot of cases. The question should be, what are the use cases that really need `validFrom` (as defined in the current VCDM definition)? In a lot of cases, credentials are valid immediately after they have been issued. If somebody really wants to know "when" the VC was issued, they could look at `proof.created` or `iat`.

> 
> If `validFrom` is not present, but the proof clearly has a date on which it was created, can the verifier presume that as the `validFrom` date? If they can't presume that, then is the VC valid for all time leading up to the `validUntil` date? If there is no `validFrom` or `validUntil` date, then the VC is valid for all time?

My interpretation would be if there is no `validFrom` and no `validUntil`, it is valid for all time.

Unless (and that is not specific to making `validFrom` OPTIONAL), an issuer decides to extend VCDM and include their own extra validity period info that is out-of-scope of this specification. For that reason, I like the approach that was proposed by @OR13 here: https://github.com/w3c/vc-data-model/issues/1027#issuecomment-1617615860, where the issuer defines critical properties that a verifier is required to validate to consider the credential valid. 

> 
> If we have answers to the questions above, we should write them down in this PR. If we don't, then I'm a bit more concerned about the ecosystem issuing VCs that don't have any validity periods, or how those validity periods are going to be interpreted differently by different verifiers. Feels like an interop problem?

awoie:
@msporny also updated the PR to reflect the case when valid from and valid until are both not present.

TallTed:
> IMO, we have the chance to remove bloat and `validFrom` is bloat in a lot of cases

But it's vital in other cases, e.g., coupons. `validFrom` should certainly be optional, but it should not be dropped, as the above suggests.

It is important, when considering features of this global standard, not to force one's own use cases onto the world. Just because *you* don't have any use for `validFrom` (for example) does not mean *no-one* has a use for it.

awoie:
> > IMO, we have the chance to remove bloat and `validFrom` is bloat in a lot of cases
> 
> But it's vital in other cases, e.g., coupons. `validFrom` should certainly be optional, but it should not be dropped, as the above suggests.
> 
> It is important, when considering features of this global standard, not to force one's own use cases onto the world. Just because _you_ don't have any use for `validFrom` (for example) does not mean _no-one_ has a use for it.

I was not suggesting to drop `validFrom`. If it was understood this way then I was not clear enough. My apologies for that. As the title of the PR and the linked issue suggests, this PR makes `validFrom` optional, and therefore removes "bloat" for use cases that don't require it while allowing other use cases to use this feature.

msporny:

Issue #1169, opened 2023-06-26
Consider removing Use Cases summary from core DM


brentzundel:
When the VC Data Model was first published, the Use Cases document wasn't in the state it currently is. So a section of use cases was added to the core data model.

Now that we have a published and (soon to be new and improved) Use Cases document, I propose we rely on that document and remove the use cases summary from the core data model.

Concretely, I think we should remove [section 1.3](https://w3c.github.io/vc-data-model/#use-cases-and-requirements) and instead point to [Use Cases and Requirements](https://www.w3.org/TR/vc-use-cases/)

OR13:
I agree.

msporny:
> Concretely, I think we should remove [section 1.3](https://w3c.github.io/vc-data-model/#use-cases-and-requirements) and instead point to [Use Cases and Requirements](https://www.w3.org/TR/vc-use-cases/)

I'd be ok with doing that if someone cross-references all of the items in this list and ensures they are captured in the use cases.

The reason we had this section initially was to ensure that the WG wrote down exactly what sort of ecosystem they wanted to create; that is, the three-party ecosystem (since we had organizations claiming that the three-party ecosystem was unnecessary at the time, or were confused about what such an ecosystem actually looked like).

I'd feel more comfortable if this section was moved to the Use Cases document in its entirety, and then whittled away over time as the statements were incorporated into that document.

OR13:
I'd be fine moving the section to use cases as a first step.

brentzundel:
@jandrieu and @KDean-GS1 there is a suggestion to either entirely remove, or move the [VC Data Model Uses Cases section](https://w3c.github.io/vc-data-model/#use-cases-and-requirements) into [VC Use Cases](https://www.w3.org/TR/vc-use-cases/). I would appreciate your thoughts here.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-07-19-vcwg#section2-1) on 2023-07-19

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 2.1. Consider removing Use Cases summary from core DM (issue vc-data-model#1169)

_See github issue [vc-data-model#1169](https://github.com/w3c/vc-data-model/issues/1169)._




**Brent Zundel:** 1169 - since the Use Cases say what we want to say, I suggest removing the Use Case summary from the data model.  

[...]

PR #1164, opened 2023-06-21
Add language clarifying VP is short-lived

brentzundel:
Fixes #937 

This PR uses language suggested in #937 to clarify that VPs are intended to be short-lived.

***
<a href="https://pr-preview.s3.amazonaws.com/brentzundel/vc-data-model/pull/1164.html" title="Last updated on Jun 23, 2023, 9:46 PM UTC (116a7e0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1164/329d6f8...brentzundel:116a7e0.html" title="Last updated on Jun 23, 2023, 9:46 PM UTC (116a7e0)">Diff</a>

Issue #1105, opened 2023-04-27
How should we refer to the Securing specifications?


msporny:
In [PR#1086](https://github.com/w3c/vc-data-model/pull/1086/files#r1169263477), @OR13 wrote:
> Feels like we should either omit [[VC-JWT]] and [[VC-DATA-INTEGRITY]] and just point to the [[VC-SPECS]] registry, or we should omit the registry pointer and just point to them.

decentralgabe:
I would prefer pointing to both, and add additional as they become normatively referencable.

OR13:
Seems like it might be better to refer to a registry that has "specification required" instead, otherwise the reference is a form of bias.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-05-17-vcwg#section2-4) on 2023-05-17

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 2.4. How should we refer to the Securing specifications? (issue vc-data-model#1105)

_See github issue [vc-data-model#1105](https://github.com/w3c/vc-data-model/issues/1105)._




**Brent Zundel:** one more 1105.  
… How should we refer to the securing specifications.  
… Who is willing to be assigned to move it forward?  
… Or should it be marked pending closed?  

**Manu Sporny:** I think we already point to the specs normatively.  
… Doesn't that address the issue?  

**Brent Zundel:** sounds like a request to mark pending close. Any objections?  

**Orie Steele:** The original issue was regarding attempting to frame the core data model to be more inclusive of more securing mechanisms.  
… pointing to DI seems to favor DI over others.  
… If we do both, we're creating a two-tier system. Those that are in the VCDM and those that aren't.  
… The original issue is how we frame securing formats. Not just those we work on here.  

**Brent Zundel:** It doesn't bother me to point to our own groups specifications.  
… but I would not be opposed to adding a link to the VC Specs Directory so that as new mechanisms are added there, they can be found.  
… Anyone willing to be assigned?  

> *Ivan Herman:* +1 to Brent's.

**Kevin Griffin:** I'll take a stab at it.  

> *Orie Steele:* thanks kevin, this seem like a good one for you.

**Brent Zundel:** Thank you.  
… with that, we didn't quite get through all of them.  

</details>

brentzundel:
I'm fine pointing directly to specs we define as examples of a normative means of accomplishing something. 
My preference is to close this issue without action.

[...]

Issue #937, opened 2022-09-24
Presentations SHOULD expire


OR13:
In general guidance on presentations is poor.

My understanding of them is that they are meant to be extremely short lived, and bound to a challenge provided by a verifier.

Therefore, they should expire in ... minutes?

If they don't expire quickly, the value of the subject / holder arguments seems less useful.

The core data model should define requirements for presentations better, and the "securing formats" should enforce them.

RieksJ:
I don't think VCDM should do much, if anything at all, with `expiration`. Here are some reasons

First, the meaning of `expiration` is what a verifier chooses to do with its value. Consider a (paper) passport: for many situations, it's ok to present a passport that hasn't expired for more than 5 years. In other situations, it may not have expired. For requesting a visa to China it must be valid 6 more months after the date at which the requester states (s)he will enter the country.

Second, the contents of the field means whatever the issuing party decides it to mean (it is a business/information-level issue that shouldn't be answered in a tech spec).

Third, its hard to decide what it should be as there are use-cases for
- very short lived VCs (e.g. a VC that attests to the result of a remote integrity test of a wallet);
- very long lived VCs (e.g. a claim that one is 18+ years old).

Similar arguments hold for VPs. Apart from the short lived example you give, long lived ones can be very beneficial (assuming a privacy friendly revocation mechanism is in place such as the one [proposed by Christiaan van Bruggen](https://repository.tudelft.nl/islandora/object/uuid%3A871e17aa-a023-461f-88b1-80d83c21de92), which allows to privacy-friendly check for revocation of credentials in a presentation an arbitrary time after the presentation was created).  The case is that a person can request a parking permit, an attestation of citizenship of a specific city, etc., that are all valid, not until such a cred expires, but until at least one of the VCs that was used to decide to issue the cred was revoked. This mechanism allows issuers of such creds to test every day/week/month/year whether or not the decision to issue the cred is still valid, enabling them to manage their resources (e.g. parking spaces) in an agile fashion, while unburdening people in that they no longer need to do all these re-applications.

David-Chadwick:
@Sakurann I have already raised an issue on data integrity that it should add an expiry time to the proof. see https://github.com/w3c/vc-data-integrity/issues/78

Sakurann:
Personally, I don't think the guidance should go beyond 

> they are meant to be extremely short lived, and bound to a challenge provided by a verifier

Details are up to verifier policy and transport protocol.

[...]

PR #1181, opened 2023-06-30
Presentations

### LINKED ISSUE #996: Question about verifiable presentations

brentzundel:
The goal of this PR is to make more explicit what a VP is.

This PR is based on feedback received on PR #1167 as well as the discussion in issue #996.
This PR fixes #996.

It changes the non-normative description (Section 3.3) of verifiable presentations to remove ambiguity.
It also adds and strengthens normative statements in section 4.10.

***
<a href="https://pr-preview.s3.amazonaws.com/brentzundel/vc-data-model/pull/1181.html" title="Last updated on Jul 7, 2023, 2:16 PM UTC (b1119cb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1181/7c44300...brentzundel:b1119cb.html" title="Last updated on Jul 7, 2023, 2:16 PM UTC (b1119cb)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1939,10 +1939,10 @@ MUST be a <a>URL</a> for the entity that is generating the <a>verifiable present
           <dt><var>proof</var></dt>
           <dd>
 The <a>verifiable presentation</a> MAY include a <code>proof</code> 
-<a>property</a>. If present, the value SHOULD be used to express a securing mechanism such as [[?VC-DATA-INTEGRITY]]. A
-<a>verifiable presentation</a> MAY be secured using an external proof such as
-[[?VC-JWT]]. For details related to the use of this property, see Section
-<a href="#securing-verifiable-credentials"></a>.
+<a>property</a>. If present, the value SHOULD be used to express a securing 
+mechanism such as [[?VC-DATA-INTEGRITY]]. A <a>verifiable presentation</a> MAY
+be secured using an external proof such as [[?VC-JWT]]. For details related to 
+the use of this property, see Section <a href="#securing-verifiable-credentials"></a>.
           </dd>
         </dl>
 
--- a/index.html
+++ b/index.html
@@ -1926,16 +1926,15 @@ details related to the use of this property, see Section <a href="#types"></a>.
           </dd>
           <dt><var>verifiableCredential</var></dt>
           <dd>
-The <code>verifiableCredential</code> <a>property</a> MUST be present. The value
+The <code>verifiableCredential</code> <a>property</a> MAY be present. The value
 MUST be an array of one or more <a>verifiable credentials</a>, or of data
 derived from <a>verifiable credentials</a> in a cryptographically
 <a>verifiable</a> format.
           </dd>
           <dt><var>holder</var></dt>
           <dd>
-The <code>holder</code> <a>property</a> is optional. If present, the value
-MUST be a <a>URL</a> for the entity that is generating the
-<a>verifiable presentation</a>.
+The <a>verifiable presentation</a> MAY include a <code>holder</code> <a>property</a>. If present, the value
+MUST be a <a>URL</a> for the entity that is generating the <a>verifiable presentation</a>.
           </dd>
           <dt><var>proof</var></dt>
           <dd>
@@ -1986,7 +1985,7 @@ cryptographically asserted such that a <a>verifier</a> can trust the value if
 they trust the <a>issuer</a>.
           </p>
           <p>
-Some selective disclosure schemes can share only a subset of <a>claims</a>
+Some selective disclosure schemes can share a subset of <a>claims</a>
 derived from a <a>verifiable credential</a>.
           </p>
 
--- a/index.html
+++ b/index.html
@@ -1943,6 +1943,9 @@ containing an <code>id</code> <a>property</a>. It is RECOMMENDED that the
 dereferenced, results in a document containing machine-readable information
 about the <a>holder</a> that can be used to <a>verify</a> the information
 expressed in the <a>verifiable presentation</a>.
+If the <code>holder</code> <a>property> is absent, information about the
+<a>holder</a> is expected to either be obtained via the securing mechanism, or
+to not pertain to the <a>validation</a> of the <a>verifiable presentation</a>.
           </dd>
           <dt><var>proof</var></dt>
           <dd>
--- a/index.html
+++ b/index.html
@@ -1938,8 +1938,8 @@ MUST be a <a>URL</a> for the entity that is generating the <a>verifiable present
           </dd>
           <dt><var>proof</var></dt>
           <dd>
-The <code>proof</code> <a>property</a> is optional. If present, the value MAY be
-be used to express a securing mechanism such as [[?VC-DATA-INTEGRITY]]. A
+The <a>verifiable presentation</a> MAY include a <code>proof</code> 
+<a>property</a>. If present, the value SHOULD be used to express a securing mechanism such as [[?VC-DATA-INTEGRITY]]. A
 <a>verifiable presentation</a> MAY be secured using an external proof such as
 [[?VC-JWT]]. For details related to the use of this property, see Section
 <a href="#securing-verifiable-credentials"></a>.
--- a/index.html
+++ b/index.html
@@ -1923,7 +1923,7 @@ normative guidance in Section <a href="#identifiers"></a> MUST be followed.
           <dt><var>type</var></dt>
           <dd>
 The <code>type</code> <a>property</a> MUST be present. It is used to express the
-type of <a>verifiable presentation</a>. The value of this property MUST be
+type of <a>verifiable presentation</a>. One value of this property MUST be
 <code>VerifiablePresentation</code>, but additional types MAY be included. The
 related normative guidance in Section <a href="#types"></a> MUST be followed.
           </dd>
@@ -1987,8 +1987,8 @@ provided in the Securing Verifiable Credentials using JSON Web Tokens
           <p>
 Some zero-knowledge cryptography schemes might enable <a>holders</a> to
 indirectly prove they hold <a>claims</a> from a <a>verifiable credential</a>
-without revealing all claims in a <a>verifiable credential</a>. In these schemes, a
-<a>verifiable credential</a> might be used to derive presentable data, which is
+without revealing all claims in that <a>verifiable credential</a>. In these schemes,
+a <a>verifiable credential</a> might be used to derive presentable data, which is
 cryptographically asserted such that a <a>verifier</a> can trust the value if
 they trust the <a>issuer</a>.
           </p>
--- a/index.html
+++ b/index.html
@@ -1936,8 +1936,13 @@ derived from <a>verifiable credentials</a> in a cryptographically
           </dd>
           <dt><var>holder</var></dt>
           <dd>
-The <a>verifiable presentation</a> MAY include a <code>holder</code> <a>property</a>. If present, the value
-MUST be a <a>URL</a> for the entity that is generating the <a>verifiable presentation</a>.
+The <a>verifiable presentation</a> MAY include a <code>holder</code>
+<a>property</a>. If present, the value MUST be either a <a>URL</a> or an object
+containing an <code>id</code> <a>property</a>. It is RECOMMENDED that the
+<a>URL</a> in the <code>holder</code> or its <code>id</code> be one which, if
+dereferenced, results in a document containing machine-readable information
+about the <a>holder</a> that can be used to <a>verify</a> the information
+expressed in the <a>verifiable presentation</a>.
           </dd>
           <dt><var>proof</var></dt>
           <dd>
--- a/index.html
+++ b/index.html
@@ -789,8 +789,11 @@ family persona, or an incognito persona.
         </p>
 
         <p>
-A <a>verifiable presentation</a> expresses data from one or more
-<a>verifiable credentials</a>.
+A <a>verifiable presentation</a> can express data from multiple
+<a>verifiable credentials</a> and contain arbitrary additional data encoded as
+JSON-LD. They are used by a <a>holder</a> to present <a>claims</a> to a
+<a>verifier</a>. It is also possible to present <a>verifiable credentials</a>
+directly.
         </p>
 
         <p>
--- a/index.html
+++ b/index.html
@@ -790,12 +790,7 @@ family persona, or an incognito persona.
 
         <p>
 A <a>verifiable presentation</a> expresses data from one or more
-<a>verifiable credentials</a>, and is packaged in such a way that the
-authorship of the data is <a>verifiable</a>. If <a>verifiable credentials</a>
-are presented directly, they become <a>verifiable presentations</a>. Data
-formats derived from <a>verifiable credentials</a> that are cryptographically
-<a>verifiable</a>, but do not of themselves contain
-<a>verifiable credentials</a>, might also be <a>verifiable presentations</a>.
+<a>verifiable credentials</a>.
         </p>
 
         <p>
--- a/index.html
+++ b/index.html
@@ -1979,7 +1979,7 @@ provided in the Securing Verifiable Credentials using JSON Web Tokens
           <p>
 Some zero-knowledge cryptography schemes might enable <a>holders</a> to
 indirectly prove they hold <a>claims</a> from a <a>verifiable credential</a>
-without revealing the entire <a>verifiable credential</a>. In these schemes, a
+without revealing all claims in a <a>verifiable credential</a>. In these schemes, a
 <a>verifiable credential</a> might be used to derive presentable data, which is
 cryptographically asserted such that a <a>verifier</a> can trust the value if
 they trust the <a>issuer</a>.
--- a/index.html
+++ b/index.html
@@ -1899,65 +1899,61 @@ status checking.
         <h3>Presentations</h3>
 
         <p>
-<a>Presentations</a> MAY be used to combine and present <a>credentials</a>.
-They can be packaged in such a way that the authorship of the data is
-<a>verifiable</a>. The data in a <a>presentation</a> is often all about the same
-<a>subject</a>, but there is no limit to the number of <a>subjects</a> or
-<a>issuers</a> in the data. The aggregation of information from multiple
-<a>verifiable credentials</a> is a typical use of
-<a>verifiable presentations</a>.
+<a>Verifiable presentations</a> MAY be used to aggregate information from
+multiple <a>verifiable credentials</a>.
         </p>
         <p>
-<a>Verifiable presentations</a> SHOULD be extremely short-lived, and
-bound to a challenge provided by a <a>verifier</a>. Details for accomplishing
-this depend on the securing mechanism, the transport protocol, and
-<a>verifier</a> policies. Unless additional requirements are defined by the
-particular securing mechanism or embedding protocol, a <a>verifier</a> cannot
-generally assume that the <a>verifiable presentation</a> has any correlation
-with the presented <a>verifiable credentials</a>.
+<a>Verifiable presentations</a> SHOULD be extremely short-lived, and bound to a
+challenge provided by a <a>verifier</a>. Details for accomplishing this depend
+on the securing mechanism, the transport protocol, and <a>verifier</a> policies.
+Unless additional requirements are defined by the particular securing mechanism
+or embedding protocol, a <a>verifier</a> cannot generally assume that the
+<a>verifiable presentation</a> has any correlation with the presented
+<a>verifiable credentials</a>.
         </p>
         <p>
-A <a>verifiable presentation</a> is typically composed of the following
-properties:
+The following properties are defined for a <a>verifiable presentation</a>:
         </p>
 
         <dl>
           <dt><var>id</var></dt>
           <dd>
-The <code>id</code> <a>property</a> is optional and MAY be used to provide a
-unique identifier for the <a>presentation</a>. For details related to the use of
-this property, see Section <a href="#identifiers"></a>.
+The <code>id</code> <a>property</a> is optional. It MAY be used to provide a
+unique identifier for the <a>verifiable presentation</a>. For details related to
+the use of this property, see Section <a href="#identifiers"></a>.
           </dd>
           <dt><var>type</var></dt>
           <dd>
-The <code>type</code> <a>property</a> is required and expresses the
-type of <a>presentation</a>, such as <code>VerifiablePresentation</code>. For
+The <code>type</code> <a>property</a> MUST be present. It is used to express the
+type of <a>verifiable presentation</a>. The value of this property MUST be
+<code>VerifiablePresentation</code>, but additional types MAY be included. For
 details related to the use of this property, see Section <a href="#types"></a>.
           </dd>
           <dt><var>verifiableCredential</var></dt>
           <dd>
-If present, the value of the <code>verifiableCredential</code> <a>property</a>
-MUST be constructed from one or more <a>verifiable credentials</a>, or of data
+The <code>verifiableCredential</code> <a>property</a> MUST be present. The value
+MUST be an array of one or more <a>verifiable credentials</a>, or of data
 derived from <a>verifiable credentials</a> in a cryptographically
 <a>verifiable</a> format.
           </dd>
           <dt><var>holder</var></dt>
           <dd>
-If present, the value of the <code>holder</code> <a>property</a>
-is expected to be a <a>URL</a> for the entity that is generating the
-<a>presentation</a>.
+The <code>holder</code> <a>property</a> is optional. If present, the value
+MUST be a <a>URL</a> for the entity that is generating the
+<a>verifiable presentation</a>.
           </dd>
           <dt><var>proof</var></dt>
           <dd>
-If present, the value of the <code>proof</code> <a>property</a> ensures that
-the <a>presentation</a> is <a>verifiable</a>. For details related to the use of
-this property, see Section <a href="#proofs-signatures"></a>.
+The <code>proof</code> <a>property</a> is optional. If present, the value MAY be
+be used to express a securing mechanism such as [[?VC-DATA-INTEGRITY]]. A
+<a>verifiable presentation</a> MAY be secured using an external proof such as
+[[?VC-JWT]]. For details related to the use of this property, see Section
+<a href="#securing-verifiable-credentials"></a>.
           </dd>
         </dl>
 
         <p>
-The example below shows a <a>verifiable presentation</a> that embeds
-<a>verifiable credentials</a>.
+The example below shows a <a>verifiable presentation</a>:
         </p>
 
         <pre class="example nohighlight" title="Basic structure of a presentation">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1989,18 +1980,14 @@ <h4>Presentations Using Derived Credentials</h4>
...
+they trust the <a>issuer</a>.
           </p>
-
           <p>
-For example, a <a>verifiable credential</a> containing the <a>claim</a>
-<code>date of birth</code> might be used to derive the presented value
-<code>over the age of 15</code> in a manner that is cryptographically
-<a>verifiable</a>. That is, a <a>verifier</a> can still trust the derived value
-if they trust the <a>issuer</a>.
+Some selective disclosure schemes can share only a subset of <a>claims</a>
```
msporny, line 1989:
```suggestion
Some selective disclosure schemes can share a subset of <a>claims</a>
```

"only" feels a bit strange in the sentence to me... "only" could imply that it's a limitation instead of an advantage (which it is).

msporny (approved):

index.html
```diff
@@ -1899,65 +1894,61 @@ <h3>Status</h3>
...
+type of <a>verifiable presentation</a>. The value of this property MUST be
+<code>VerifiablePresentation</code>, but additional types MAY be included. For
 details related to the use of this property, see Section <a href="#types"></a>.
           </dd>
           <dt><var>verifiableCredential</var></dt>
           <dd>
-If present, the value of the <code>verifiableCredential</code> <a>property</a>
-MUST be constructed from one or more <a>verifiable credentials</a>, or of data
+The <code>verifiableCredential</code> <a>property</a> MUST be present. The value
+MUST be an array of one or more <a>verifiable credentials</a>, or of data
```
awoie, line 1930:
```suggestion
The <code>verifiableCredential</code> <a>property</a> MAY be present. The value
MUST be an array of one or more <a>verifiable credentials</a>, or of data
```

awoie: Some people are using VPs without VCs in it. I believe CHAPI used such an approach for DID Auth? (cc @dmitrizagidulin)

index.html
```diff
@@ -1899,65 +1894,61 @@ <h3>Status</h3>
...
 <a>verifiable</a> format.
           </dd>
           <dt><var>holder</var></dt>
           <dd>
-If present, the value of the <code>holder</code> <a>property</a>
-is expected to be a <a>URL</a> for the entity that is generating the
-<a>presentation</a>.
+The <code>holder</code> <a>property</a> is optional. If present, the value
+MUST be a <a>URL</a> for the entity that is generating the
+<a>verifiable presentation</a>.
```
awoie, line 1938:
```suggestion
The <a>verifiable presentation</a> MAY include a <code>holder</code> <a>property</a>. If present, the value
MUST be a <a>URL</a> for the entity that is generating the <a>verifiable presentation</a>.
```

index.html
```diff
@@ -1899,65 +1894,61 @@ <h3>Status</h3>
...
           </dd>
           <dt><var>proof</var></dt>
           <dd>
-If present, the value of the <code>proof</code> <a>property</a> ensures that
-the <a>presentation</a> is <a>verifiable</a>. For details related to the use of
-this property, see Section <a href="#proofs-signatures"></a>.
+The <code>proof</code> <a>property</a> is optional. If present, the value MAY be
+be used to express a securing mechanism such as [[?VC-DATA-INTEGRITY]]. A
```
awoie, line 1943:
```suggestion
The <a>verifiable presentation</a> MAY include a <code>proof</code> <a>property</a>. If present, the value MAY be be used to express a securing mechanism such as [[?VC-DATA-INTEGRITY]]. A
```

awoie (changes requested):
LGTM after suggestions applied. "optional" is not a normative verb according to 1.4 which is why I rewrote those sections.

index.html
msporny: Agree with this change. What we've done in the past is using phrasing like: "The FOO property is OPTIONAL. If present, it MUST..."

That's purely editorial, though, it's effectively saying the same thing as the suggestion made by @awoie .

msporny:
> LGTM after suggestions applied. "optional" is not a normative verb according to 1.4 which is why I rewrote those sections.

OPTIONAL is a normative word, but we were using the lowercase version of it (on purpose). We could also fix things by just using "OPTIONAL", as that's equivalent to "MAY".

I'll note that having "SHOULD" and "MAY" statements are often not tested because people complain when their implementations fail normative statements.

In any case, I agree with the normative changes @awoie is suggesting to the contents of the VerifiablePresentation object.

index.html
```diff
@@ -790,12 +790,7 @@ <h3>Presentations</h3>
 
         <p>
 A <a>verifiable presentation</a> expresses data from one or more
-<a>verifiable credentials</a>, and is packaged in such a way that the
-authorship of the data is <a>verifiable</a>. If <a>verifiable credentials</a>
-are presented directly, they become <a>verifiable presentations</a>. Data
```
decentralgabe, line 795: I like some of this text, don't think it should all be removed.

index.html
```diff
@@ -1899,65 +1894,61 @@ <h3>Status</h3>
...
         </p>
         <p>
-<a>Verifiable presentations</a> SHOULD be extremely short-lived, and
-bound to a challenge provided by a <a>verifier</a>. Details for accomplishing
-this depend on the securing mechanism, the transport protocol, and
-<a>verifier</a> policies. Unless additional requirements are defined by the
-particular securing mechanism or embedding protocol, a <a>verifier</a> cannot
-generally assume that the <a>verifiable presentation</a> has any correlation
-with the presented <a>verifiable credentials</a>.
+<a>Verifiable presentations</a> SHOULD be extremely short-lived, and bound to a
```
decentralgabe, line 1901: Why? What's the risk in a holder creating a presentation that's long lived? Let's say I want to host a presentation on my website, or LinkedIn, display my employment history. That can be a use can for a long-lived presentation.

index.html
brentzundel: I'm open to suggestions. My attempt to adjust this section did not find consensus.

index.html
brentzundel: This PR only changes the line length of this paragraph, it isn't introducing any new text.
The text about short lived presentations was added when PR #1164 was merged (which was based on Issue #937

index.html
brentzundel, line 1930:
```suggestion
The <code>verifiableCredential</code> <a>property</a> is optional. If present, the value
MUST be an array of one or more <a>verifiable credentials</a>, or of data
```

index.html
brentzundel: I'm wondering if this one should be SHOULD

TallTed:
> I'll note that having "SHOULD" and "MAY" statements are often not tested because people complain when their implementations fail normative statements.

Then we SHOULD be clear about what SHOULD be tested, and how the results SHOULD be interpreted. Failure of a MAY is only problematic if the implementation claims that it implemented that feature. Failure of a SHOULD is only problematic if the implementation claims that it implemented that feature according to the SHOULD. Both of these failures are helpful to evaluators who SHOULD be able to discern implementations which DID implement those features from implementations which did NOT.

People who complain at that point will need to make a very strong case for why we ought to pay attention to their complaints.

awoie:
> OPTIONAL is a normative word, but we were using the lowercase version of it (on purpose). We could also fix things by just using "OPTIONAL", as that's equivalent to "MAY".

@msporny According to the [Conformance](https://w3c.github.io/vc-data-model/#conformance) section, we only have the following:

> The key words MAY, MUST, MUST NOT, RECOMMENDED, SHOULD, and SHOULD NOT in this document are to be interpreted as described in [BCP 14](https://datatracker.ietf.org/doc/html/bcp14) [[RFC2119](https://w3c.github.io/vc-data-model/#bib-rfc2119)] [[RFC8174](https://w3c.github.io/vc-data-model/#bib-rfc8174)] when, and only when, they appear in all capitals, as shown here.

I know that OPTIONAL is used in IETF RFCs but I couldn't find it in the paragraph above. Should we add OPTIONAL there too? Or what is the reason it was not added?

msporny:
> I know that OPTIONAL is used in IETF RFCs but I couldn't find it in the paragraph above. Should we add OPTIONAL there too? Or what is the reason it was not added?

That paragraph is auto-generated by ReSpec based on the normative RFC language used throughout the document. So, if you don't use a certain RFC term, it's not added to that paragraph. If we use "OPTIONAL" (in all caps) in the document, ReSpec will modify that paragraph I pointed to and will add the word automatically to that paragraph.

So, no reason to modify that particular paragraph, it'll auto-update if we use the OPTIONAL language in the places where we use lowercase "optional" today.

index.html
decentralgabe: can we keep 

```
<a>verifiable credentials</a>, and is packaged in such a way that the
authorship of the data is <a>verifiable</a>. If <a>verifiable credentials</a>
are presented directly, they become <a>verifiable presentations</a>.
```

index.html
brentzundel: The second MAY in the suggestion should be a SHOULD

index.html
```diff
@@ -1899,65 +1894,60 @@ <h3>Status</h3>
...
+The <a>verifiable presentation</a> MAY include a <code>holder</code> <a>property</a>. If present, the value
+MUST be a <a>URL</a> for the entity that is generating the <a>verifiable presentation</a>.
           </dd>
           <dt><var>proof</var></dt>
           <dd>
-If present, the value of the <code>proof</code> <a>property</a> ensures that
-the <a>presentation</a> is <a>verifiable</a>. For details related to the use of
-this property, see Section <a href="#proofs-signatures"></a>.
+The <code>proof</code> <a>property</a> is optional. If present, the value MAY be
+be used to express a securing mechanism such as [[?VC-DATA-INTEGRITY]]. A
```
brentzundel, line 1942:
```suggestion
The <a>verifiable presentation</a> MAY include a <code>proof</code> 
<a>property</a>. If present, the value SHOULD be used to express a securing mechanism such as [[?VC-DATA-INTEGRITY]]. A
```

index.html
```diff
@@ -1899,65 +1894,60 @@ <h3>Status</h3>
...
           <dt><var>proof</var></dt>
           <dd>
-If present, the value of the <code>proof</code> <a>property</a> ensures that
-the <a>presentation</a> is <a>verifiable</a>. For details related to the use of
-this property, see Section <a href="#proofs-signatures"></a>.
+The <a>verifiable presentation</a> MAY include a <code>proof</code> 
+<a>property</a>. If present, the value SHOULD be used to express a securing mechanism such as [[?VC-DATA-INTEGRITY]]. A
+<a>verifiable presentation</a> MAY be secured using an external proof such as
+[[?VC-JWT]]. For details related to the use of this property, see Section
+<a href="#securing-verifiable-credentials"></a>.
```
brentzundel, line 1945:
```suggestion
The <a>verifiable presentation</a> MAY include a <code>proof</code> 
<a>property</a>. If present, the value SHOULD be used to express a securing 
mechanism such as [[?VC-DATA-INTEGRITY]]. A <a>verifiable presentation</a> MAY
be secured using an external proof such as [[?VC-JWT]]. For details related to 
the use of this property, see Section <a href="#securing-verifiable-credentials"></a>.
```

brentzundel:
@awoie I believe I have addressed your concerns, could you re-review?

index.html
```diff
@@ -790,12 +790,7 @@ <h3>Presentations</h3>
 
         <p>
 A <a>verifiable presentation</a> expresses data from one or more
-<a>verifiable credentials</a>, and is packaged in such a way that the
-authorship of the data is <a>verifiable</a>. If <a>verifiable credentials</a>
-are presented directly, they become <a>verifiable presentations</a>. Data
-formats derived from <a>verifiable credentials</a> that are cryptographically
-<a>verifiable</a>, but do not of themselves contain
-<a>verifiable credentials</a>, might also be <a>verifiable presentations</a>.
+<a>verifiable credentials</a>.
```
OR13, line 793: 0 or more credentials AND arbitrary additional data encoded in JSON-LD

index.html
```diff
@@ -1899,65 +1894,60 @@ <h3>Status</h3>
...
+The following properties are defined for a <a>verifiable presentation</a>:
         </p>
 
         <dl>
           <dt><var>id</var></dt>
           <dd>
-The <code>id</code> <a>property</a> is optional and MAY be used to provide a
-unique identifier for the <a>presentation</a>. For details related to the use of
-this property, see Section <a href="#identifiers"></a>.
+The <code>id</code> <a>property</a> is optional. It MAY be used to provide a
```
OR13, line 1916: You need to comment on the `type` that is legal here... It needs to be consistent with `@id` / IRIs.

index.html
```diff
@@ -1899,65 +1894,60 @@ <h3>Status</h3>
...
+unique identifier for the <a>verifiable presentation</a>. For details related to
+the use of this property, see Section <a href="#identifiers"></a>.
           </dd>
           <dt><var>type</var></dt>
           <dd>
-The <code>type</code> <a>property</a> is required and expresses the
-type of <a>presentation</a>, such as <code>VerifiablePresentation</code>. For
+The <code>type</code> <a>property</a> MUST be present. It is used to express the
+type of <a>verifiable presentation</a>. The value of this property MUST be
+<code>VerifiablePresentation</code>, but additional types MAY be included. For
```
OR13, line 1924: You need to comment on if its an array, what values are legal, etc...

index.html
```diff
@@ -1899,65 +1894,60 @@ <h3>Status</h3>
...
 details related to the use of this property, see Section <a href="#types"></a>.
           </dd>
           <dt><var>verifiableCredential</var></dt>
           <dd>
-If present, the value of the <code>verifiableCredential</code> <a>property</a>
-MUST be constructed from one or more <a>verifiable credentials</a>, or of data
+The <code>verifiableCredential</code> <a>property</a> MAY be present. The value
```
OR13, line 1929: if its present, can it be empty?

index.html
```diff
@@ -1899,65 +1894,60 @@ <h3>Status</h3>
...
 derived from <a>verifiable credentials</a> in a cryptographically
 <a>verifiable</a> format.
           </dd>
           <dt><var>holder</var></dt>
           <dd>
-If present, the value of the <code>holder</code> <a>property</a>
-is expected to be a <a>URL</a> for the entity that is generating the
-<a>presentation</a>.
+The <a>verifiable presentation</a> MAY include a <code>holder</code> <a>property</a>. If present, the value
+MUST be a <a>URL</a> for the entity that is generating the <a>verifiable presentation</a>.
```
OR13, line 1937: we probably need this URL reference everywhere we see `id`... and we probably want to allow for this as well:

```
holder: { id: ...., name: alice }
```

index.html
```diff
@@ -1899,65 +1894,60 @@ <h3>Status</h3>
...
           </dd>
           <dt><var>proof</var></dt>
           <dd>
-If present, the value of the <code>proof</code> <a>property</a> ensures that
-the <a>presentation</a> is <a>verifiable</a>. For details related to the use of
-this property, see Section <a href="#proofs-signatures"></a>.
+The <a>verifiable presentation</a> MAY include a <code>proof</code> 
+<a>property</a>. If present, the value SHOULD be used to express a securing
```
OR13, line 1942: what happens when this not the case, how does a verifier tell if a proof is for "vc-data-integrity" or "something else" ?

index.html
```diff
@@ -1989,18 +1979,14 @@ <h4>Presentations Using Derived Credentials</h4>
           <p>
 Some zero-knowledge cryptography schemes might enable <a>holders</a> to
 indirectly prove they hold <a>claims</a> from a <a>verifiable credential</a>
-without revealing the <a>verifiable credential</a> itself. In these schemes, a
-<a>claim</a> from a <a>verifiable credential</a> might be used to derive a
-presented value, which is cryptographically asserted such that a <a>verifier</a>
-can trust the value if they trust the <a>issuer</a>.
+without revealing the entire <a>verifiable credential</a>. In these schemes, a
```
OR13, line 1982:
```suggestion
without revealing all claims in a <a>verifiable credential</a>. In these schemes, a
```

OR13 (changes requested):
Only blocking feedback is this:

https://github.com/w3c/vc-data-model/pull/1181/files#r1253523816

It might be better to just more directly state the normative requirements of `vp+ld+json`.

`@context` and `type` what is required.

index.html
brentzundel: Excellent question, I think that the PR addressing #1105 could address this.

index.html
brentzundel: Now that we've introduced media types, I don't think it is true that  a verifiable credential (vc+ld+json) becomes a verifiable presentation (vp+ld+json).

What about a sentence like: "It is also possible to present verifiable credentials directly."

index.html
brentzundel: I thought this was covered by the reference to the `identifiers` section.

index.html
brentzundel: Maybe (as above) the links to other sections where this is already described normatively aren't sufficient? Should I just be explicit and say if present, it MUST follow the guidance in the types section?

index.html
brentzundel: What is better:
1. required and can be empty, or
2. not required but has to have at least one value?

index.html
brentzundel: I'll modify the PR to mimic the language in the issuer property. I think that will address this.

index.html
brentzundel: I've changed the language to match Oliver's suggestion, but with SHOULD instead of MAY.

index.html
decentralgabe: sure I think that's a good sentence

index.html
brentzundel: Re-worded in [12f46a4](https://github.com/w3c/vc-data-model/pull/1181/commits/12f46a47beeeeb949b9ed15b6553c06338a6ec87)

index.html
brentzundel: Re-worded in [12f46a4](https://github.com/w3c/vc-data-model/pull/1181/commits/12f46a47beeeeb949b9ed15b6553c06338a6ec87)

index.html
brentzundel: I've re-worded the language in [816f606](https://github.com/w3c/vc-data-model/pull/1181/commits/816f606c4890577297e84d3dda63d0a9d678f0a1) to normatively point to the guidance you suggest.

index.html
brentzundel: I've re-worded the language in [816f606](https://github.com/w3c/vc-data-model/pull/1181/commits/816f606c4890577297e84d3dda63d0a9d678f0a1) to normatively point to the guidance you suggest.

index.html
brentzundel: modified the language in [a8a03e0](https://github.com/w3c/vc-data-model/pull/1181/commits/a8a03e0484537831023fd597f527185b0fa56f96)

brentzundel:
@OR13 and @awoie I have made changes in reference to your suggestions. Please re-review.

index.html
brentzundel: modified the language in [e5b8dd0](https://github.com/w3c/vc-data-model/pull/1181/commits/e5b8dd056271f06b19b8e36890ce573853d273df)

index.html
brentzundel: Re-worded in [29f5cc9](https://github.com/w3c/vc-data-model/pull/1181/commits/29f5cc974498b19b1f4062f4ddc05bdeb5e954f5)

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-07-05-vcwg#section2-6) on 2023-07-05

- no resolutions were taken

#### 2.6. Determine interoperable way for Holder to make claims directly in VPs (issue vc-data-model#860)

_See github issue [vc-data-model#860](https://github.com/w3c/vc-data-model/issues/860)._




**Kristina Yasuda:** This is about self-signed VCs, right?  

**Orie Steele:** Brent has a raised a PR that addresses a lot of this issue.  
… Pull 1181.  
… PR 1181 doesn't currently reflect what you can do with a VP.  
… Brent, can you please revise per the comments.  

_See github pull request [vc-data-model#1181](https://github.com/w3c/vc-data-model/pull/1181)._




> *Orie Steele:* it could address that issue.

**Brent Zundel:** I don't remember if I tried to address the comments or not yet, but I definitely can.  

**Kristina Yasuda:** I've assigned this to you, Brent.  

> *Joe Andrieu:* present?

</details>

index.html
```diff
@@ -1899,65 +1897,65 @@ <h3>Status</h3>
...
+unique identifier for the <a>verifiable presentation</a>. If present, the
+normative guidance in Section <a href="#identifiers"></a> MUST be followed.
           </dd>
           <dt><var>type</var></dt>
           <dd>
-The <code>type</code> <a>property</a> is required and expresses the
-type of <a>presentation</a>, such as <code>VerifiablePresentation</code>. For
-details related to the use of this property, see Section <a href="#types"></a>.
+The <code>type</code> <a>property</a> MUST be present. It is used to express the
+type of <a>verifiable presentation</a>. The value of this property MUST be
```
TallTed, line 1926:
```suggestion
type of <a>verifiable presentation</a>. One value of this property MUST be
```
This is necessary to allow for the "additional types" a/k/a "additional type values" in the next line.

index.html
```diff
@@ -1989,18 +1987,14 @@ <h4>Presentations Using Derived Credentials</h4>
           <p>
 Some zero-knowledge cryptography schemes might enable <a>holders</a> to
 indirectly prove they hold <a>claims</a> from a <a>verifiable credential</a>
-without revealing the <a>verifiable credential</a> itself. In these schemes, a
-<a>claim</a> from a <a>verifiable credential</a> might be used to derive a
-presented value, which is cryptographically asserted such that a <a>verifier</a>
-can trust the value if they trust the <a>issuer</a>.
+without revealing all claims in a <a>verifiable credential</a>. In these schemes, a
+<a>verifiable credential</a> might be used to derive presentable data, which is
```
TallTed, line 1991:
```suggestion
without revealing all claims in that <a>verifiable credential</a>. In these schemes,
a <a>verifiable credential</a> might be used to derive presentable data, which is
```

awoie (approved):
lgtm

index.html
```diff
@@ -789,13 +789,11 @@ <h3>Presentations</h3>
...
-authorship of the data is <a>verifiable</a>. If <a>verifiable credentials</a>
-are presented directly, they become <a>verifiable presentations</a>. Data
-formats derived from <a>verifiable credentials</a> that are cryptographically
-<a>verifiable</a>, but do not of themselves contain
-<a>verifiable credentials</a>, might also be <a>verifiable presentations</a>.
+A <a>verifiable presentation</a> can express data from multiple
+<a>verifiable credentials</a> and contain arbitrary additional data encoded as
+JSON-LD. They are used by a <a>holder</a> to present <a>claims</a> to a
+<a>verifier</a>. It is also possible to present <a>verifiable credentials</a>
+directly.
```
awoie, line 796: What does presenting VCs "directly" mean?

awoie (changes requested):

index.html
```diff
@@ -1899,65 +1897,65 @@ <h3>Status</h3>
...
-If present, the value of the <code>holder</code> <a>property</a>
-is expected to be a <a>URL</a> for the entity that is generating the
-<a>presentation</a>.
+The <a>verifiable presentation</a> MAY include a <code>holder</code>
+<a>property</a>. If present, the value MUST be either a <a>URL</a> or an object
+containing an <code>id</code> <a>property</a>. It is RECOMMENDED that the
+<a>URL</a> in the <code>holder</code> or its <code>id</code> be one which, if
+dereferenced, results in a document containing machine-readable information
+about the <a>holder</a> that can be used to <a>verify</a> the information
+expressed in the <a>verifiable presentation</a>.
```
awoie, line 1945: We have to say something about who is the holder in case there is no holder property provided.

awoie (changes requested):

index.html
awoie: If "directly" means that a VC is a VP, then we need an example.

If we just say that one does not need to use a VP as an envelope for a VC (which I think is what the sentence means), then we have to define under which circumstances VPs are useful at all. IMO, presenting VCs should be the default case, and only if x, y and z matter to the implementer, then you can use a VP to express x, y, z. Does this make sense?

awoie: In the present VC directly case, we should also define what the holder is. The holder might be the credential subject, or just unknown, or inferred by the proof and respective securing mechanism.

brentzundel: I don't disagree with much of what you say, I'm wondering if this needs to happen in this PR.

index.html
```diff
@@ -1899,65 +1897,65 @@ <h3>Status</h3>
...
```
brentzundel, line 1945:
```suggestion
expressed in the <a>verifiable presentation</a>.
If the <code>holder</code> <a>property> is absent, information about the
<a>holder</a> is expected to either be obtained via the securing mechanism, or
to not pertain to the <a>validation</a> of the <a>verifiable presentation</a>.
```

index.html
brentzundel: I've added a sentence, let me know what you think.

index.html
brentzundel: n.b., the section in which this language is found is a non-normative use cases summary that I believe should simply be removed #1169

awoie: Got it, that makes sense.

index.html
awoie: That sounds good to me!

awoie (approved):

index.html
OR13: perfect

index.html
OR13: 2 is better... this is a perfect case of "many different JSONs" produce the same "graph"... and proof mechanisms might not preserve all of them the same way...

We have the exact same problem come up with `credentialSubject`.

we said it MUST be present and it MUST NOT be empty.

Doing that for Verifiable Presentations would destroy "DID Auth" which use data integrity proofs to prove possession of a private key for a holder, but does not communicate any credentials.

So my reasoning for 2 is to allow for that behavior to conintue to be legal, while removing equivalent JSON syntax for the same RDF Graph.

OR13 (approved):
Approved, assuming an issue marker or some attempt to address https://github.com/w3c/vc-data-model/pull/1181/files#r1256018885

index.html
brentzundel: excellent. 2 is how it's currently worded, so I won't make any changes.

brentzundel:
> Approved, assuming an issue marker or some attempt to address https://github.com/w3c/vc-data-model/pull/1181/files#r1256018885

Can you say more what this issue would say? I think the current PR text supports what you're looking for here.

index.html
```diff
@@ -1899,65 +1897,68 @@ <h3>Status</h3>
...
 
         <dl>
           <dt><var>id</var></dt>
           <dd>
-The <code>id</code> <a>property</a> is optional and MAY be used to provide a
-unique identifier for the <a>presentation</a>. For details related to the use of
-this property, see Section <a href="#identifiers"></a>.
+The <code>id</code> <a>property</a> is optional. It MAY be used to provide a
+unique identifier for the <a>verifiable presentation</a>. If present, the
+normative guidance in Section <a href="#identifiers"></a> MUST be followed.
```
msporny, line 1921: Just noting that we don't need the normative statement on identifiers, as all identifiers must follow that guidance anyway. This normative statement doesn't make the implementation do anything above and beyond what's stated in the `#identifiers` section. I think we can safely remove this statement (in a future PR).

index.html
```diff
@@ -1899,65 +1897,68 @@ <h3>Status</h3>
...
           </dd>
           <dt><var>type</var></dt>
           <dd>
-The <code>type</code> <a>property</a> is required and expresses the
-type of <a>presentation</a>, such as <code>VerifiablePresentation</code>. For
-details related to the use of this property, see Section <a href="#types"></a>.
+The <code>type</code> <a>property</a> MUST be present. It is used to express the
+type of <a>verifiable presentation</a>. One value of this property MUST be
+<code>VerifiablePresentation</code>, but additional types MAY be included. The
+related normative guidance in Section <a href="#types"></a> MUST be followed.
```
msporny, line 1928: Just noting that we don't need the normative statement on types, as all types must follow that guidance anyway. This normative statement doesn't make the implementation do anything above and beyond what's stated in the #types section. I think we can safely remove this statement (in a future PR).

msporny:

index.html
msporny: PR #1212 has been raised to address issue #1105.

PR #1182, opened 2023-07-01
removing remaining references to JSON

Sakurann:
Clarifying that "There is no such thing as `vc+json`", addressing #1044 and preventing future confusion like in #1088 after it was merged. As part of that clarification, it references JSON-LD spec directly wrt the processing of `@context`, since if the base media type is `vc+ld+json`, it needs to be a valid JSON-LD.

***
<a href="https://pr-preview.s3.amazonaws.com/Sakurann/vc-data-model/pull/1182.html" title="Last updated on Jul 1, 2023, 9:49 PM UTC (8278dc6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1182/562b9f3...Sakurann:8278dc6.html" title="Last updated on Jul 1, 2023, 9:49 PM UTC (8278dc6)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1124,8 +1124,7 @@ in a document containing machine-readable information about the
         </dl>
         <p class="note">
 This specification requires for a <code>@context</code> <a>property</a>
-to be present, and it is required that the value of the <code>@context</code>
-<a>property</a> is processed using JSON-LD.
+to be present; this property is defined by [[JSON-LD]].
         </p>
         <pre class="example nohighlight" title="Usage of the @context property">
 {
--- a/index.html
+++ b/index.html
@@ -608,12 +608,11 @@ proof mechanism; it requires clearly identifying the mechanism a
         </p>
 
         <p>
-This document also contains examples that contain JSON and JSON-LD content.
-Some of these examples contain characters that are invalid JSON, such as
+This document also contains examples that contain characters that are invalid JSON, such as
 inline comments (<code>//</code>) and the use of ellipsis (<code>...</code>)
 to denote information that adds little value to the example. Implementers are
 cautioned to remove this content if they desire to use the information as
-valid JSON or JSON-LD.
+valid.
         </p>
       </section>
 
@@ -1124,15 +1123,9 @@ in a document containing machine-readable information about the
           </dd>
         </dl>
         <p class="note">
-Though this specification requires that a <code>@context</code> <a>property</a>
-be present, it is not required that the value of the <code>@context</code>
-<a>property</a> be processed using JSON-LD. This is to support processing using
-plain JSON libraries, such as those that might be used when the
-<a>verifiable credential</a> is encoded as a JWT. All libraries or processors
-MUST ensure that the order of the values in the <code>@context</code>
-<a>property</a> is what is expected for the specific application. Libraries or
-processors that support JSON-LD can process the <code>@context</code>
-<a>property</a> using full JSON-LD processing as expected.
+This specification requires for a <code>@context</code> <a>property</a>
+to be present, and it is required that the value of the <code>@context</code>
+<a>property</a> is processed using JSON-LD.
         </p>
         <pre class="example nohighlight" title="Usage of the @context property">
 {
@@ -2411,22 +2404,7 @@ these extension points.
         <section>
           <h4>Semantic Interoperability</h4>
 
-          <p>
-This specification ensures that "plain" JSON and JSON-LD syntaxes are
-semantically compatible without requiring JSON implementations to use a JSON-LD
-processor. To achieve this, the specification imposes the following additional
-requirements on both syntaxes:
-          </p>
-
           <ul>
-            <li>
-JSON-based processors MUST process the <code>@context</code> key, ensuring the
-expected values exist in the expected order for the <a>credential</a> type being
-processed. The order is important because keys used in a <a>credential</a>,
-which are defined using the values associated with <code>@context</code>, are
-defined using a "first defined wins" mechanism and changing the order might
-result in a different key definition "winning".
-            </li>
             <li>
 JSON-LD-based processors MUST produce an error when a JSON-LD context redefines
 any term in the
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1124,15 +1123,9 @@ <h3>Contexts</h3>
...
-<a>property</a> be processed using JSON-LD. This is to support processing using
-plain JSON libraries, such as those that might be used when the
-<a>verifiable credential</a> is encoded as a JWT. All libraries or processors
-MUST ensure that the order of the values in the <code>@context</code>
-<a>property</a> is what is expected for the specific application. Libraries or
-processors that support JSON-LD can process the <code>@context</code>
-<a>property</a> using full JSON-LD processing as expected.
+This specification requires for a <code>@context</code> <a>property</a>
+to be present, and it is required that the value of the <code>@context</code>
+<a>property</a> is processed using JSON-LD.
```
dlongley, line 1128: Well, we should avoid possibly deviating from the JSON-LD spec itself:

> JSON-LD is used in many specifications as the specified data format. However, there is also a desire to allow some JSON-LD contents to be processed as plain JSON, without using any of the JSON-LD algorithms. Because JSON-LD is very flexible, some terms from the original format may be locally overridden through the use of [embedded contexts](https://www.w3.org/TR/json-ld11/#dfn-embedded-context), and take a different meaning for JSON-LD based implementations. On the other hand, "plain JSON" implementations may not be able to interpret these [embedded contexts](https://www.w3.org/TR/json-ld11/#dfn-embedded-context), and hence will still interpret those terms with their original meaning. To prevent this divergence of interpretation, JSON-LD 1.1 allows term definitions to be protected.

https://www.w3.org/TR/json-ld11/#protected-term-definitions

We can instead just reference JSON-LD directly to make it very clear that that spec is what is to be used to understand `@context`. Perhaps like this:

```suggestion
This specification requires for a <code>@context</code> <a>property</a>
to be present; this property is defined by [[JSON-LD]].
```

dlongley (changes requested):

index.html
Sakurann: I really like the suggestion! Thanks.

index.html
```diff
@@ -608,12 +608,11 @@ <h3>Use Cases and Requirements</h3>
...
 
         <p>
-This document also contains examples that contain JSON and JSON-LD content.
-Some of these examples contain characters that are invalid JSON, such as
+This document also contains examples that contain characters that are invalid JSON, such as
 inline comments (<code>//</code>) and the use of ellipsis (<code>...</code>)
 to denote information that adds little value to the example. Implementers are
 cautioned to remove this content if they desire to use the information as
-valid JSON or JSON-LD.
+valid.
```
msporny, line 615:
```suggestion
a valid document.
```

msporny (approved):
I'm a bit on the fence about removing this language, as that is how some implementations (including some of Digital Bazaar's processing pipelines), handle the VC as it is coming into our systems. We do, when it matters, do full JSON-LD processing, but that is not always necessary to perform. As long as we don't suggest that full JSON-LD processing is always required, this PR is fine. We might want to move a variation of this text to the implementation guide to ensure that developers know that it's ok to 1) ensure the order of the `@context` entries, and then 2) process as ordinary JSON, in some circumstances.

dlongley (approved):

OR13 (approved):

decentralgabe (approved):

brentzundel (approved):

OR13:
Keep in mind that if `+ld+json` is not registered... none of what we are doing will matter... because there will be no general guidance on processing the media types we are requesting registration for.

https://github.com/w3c/json-ld-syntax/pull/415

OR13 (changes requested):
I'm requesting changes until the structured suffixes details are resolved sufficiently to protect implementers from the confusion that would arise from removing these references while not defining how the suffixes are processed.

Sakurann:
@OR13 what concrete details on the structures suffixes do you want to be resolved? multiple suffixes draft in IETF and how `+ld+jwt` is being processed is orthogonal to the changes in this PR and should block it.

OR13:
@Sakurann I guess you are right, the problem will be... if `+ld+json` is never registered, and its the only media type suffix we rely on, then our document should not advance.

I am approving, with the hope that does not happen.

OR13 (approved):

msporny:
@Sakurann, waiting on a typo fix before merging: https://github.com/w3c/vc-data-model/pull/1182/files#r1249374162

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-07-11-vcwg-special#section1-7) on 2023-07-11

- no resolutions were taken

#### 1.7. removing remaining references to JSON (pr vc-data-model#1182)

_See github pull request [vc-data-model#1182](https://github.com/w3c/vc-data-model/pull/1182)._




**Brent Zundel:** this P.R removes differentiation between JSONLD and JSON. Has relatively broad approval.  
… minor changes from manu, that I agree with.  
… not sure this P.R calls for much discussion, but open to any if there are some.  

</details>

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-07-11-vcwg-special#section1-7) on 2023-07-11

- no resolutions were taken

#### 1.7. removing remaining references to JSON (pr vc-data-model#1182)

_See github pull request [vc-data-model#1182](https://github.com/w3c/vc-data-model/pull/1182)._




**Brent Zundel:** this P.R removes differentiation between JSONLD and JSON. Has relatively broad approval.  
… minor changes from manu, that I agree with.  
… not sure this P.R calls for much discussion, but open to any if there are some.  

</details>

msporny:
> @Sakurann, waiting on a typo fix before merging: https://github.com/w3c/vc-data-model/pull/1182/files#r1249374162

Actually, the suggestion is Editorial, and made by me, so I can just merge at this point and make the change, which is what I'll do.

msporny:

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-07-11-vcwg-special#section1-7) on 2023-07-11

- no resolutions were taken

#### 1.7. removing remaining references to JSON (pr vc-data-model#1182)

_See github pull request [vc-data-model#1182](https://github.com/w3c/vc-data-model/pull/1182)._




**Brent Zundel:** this P.R removes differentiation between JSONLD and JSON. Has relatively broad approval.  
… minor changes from manu, that I agree with.  
… not sure this P.R calls for much discussion, but open to any if there are some.  

</details>

Issue #846, opened 2021-11-29
Include a regex for XML Date Time in the next version of the standard.


OR13:
If we can't agree to relax XML Data Time to the broader category or ISO 8601 date times, we should include a regex that defines the allows string representations, and we should document all known issues with string representations of date times in the spec, so that implementers don't have to guess about "leap seconds" or other time related issues.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2021-12-01-vcwg#section3-3) on 2021-12-01

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 3.3. Include a regex for XML Date Time in the next version of the standard. (issue vc-data-model#846)

_See github issue [vc-data-model#846](https://github.com/w3c/vc-data-model/issues/846)._




**Brent Zundel:** This seems like it's v2.0, anyone disagree?.  

</details>

OR13:
Related https://us-cert.cisa.gov/sites/default/files/documents/Best_Practices_for_Leap_Second_Event_Occurring_on_31_December_2016_S508C.pdf

OR13:
See also: https://en.wikipedia.org/wiki/Year_2038_problem

mprorock:
something like `(\d{4}-\d{2}-\d{2})[A-Z]+(\d{2}:\d{2}:\d{2}).([0-9+-:]+)` ?

OR13:
possibly useful: https://github.com/transmute-industries/verifiable-data/blob/43b2dd1f91dc200927bd8c859ece31eae465e4f1/packages/vc.js/src/datetime/datetime.test.ts

brentzundel:
possible regex from @mprorock: 
> mprorock spits (\d{4}-\d{2}-\d{2})[A-Z]+(\d{2}:\d{2}:\d{2}).([0-9+-:]+) out of the magic "i hate regex" scratchpad

mprorock:
and a more brutal, but probably more correct one:
[...]

PR #1183, opened 2023-07-04
Add dateTimeStamp regex and explanation of representing time.

msporny:
This PR attempts to fix issue #846 by doing the following:

* Makes the specification of timezone mandatory.
* Explains why not all syntactically legal timestamps are sometimes not valid across different cultures.
* Utilizes a regex that has survived the test of time (XML Schema 1.1, Part 2: dateTimeStamp), even if not every time stamp that passes the regex is recognized as a valid point in time (scientifically and/or culturally).

This is probably the best we can do, unless we want to try writing our own regex, which I strongly advise us against doing because it will deviate from previous WGs that have [dumped way more time than we have into this topic](http://lists.xml.org/archives/xml-dev/201808/msg00024.html).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1183.html" title="Last updated on Jul 14, 2023, 7:48 PM UTC (2b674b7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1183/3a0e423...2b674b7.html" title="Last updated on Jul 14, 2023, 7:48 PM UTC (2b674b7)">Diff</a>


iherman (approved):
***If*** we wanted to make a change on the regex, we may want to restrict it by only 'Z' at the end, because we do say that the time must be in UTC.

But I would agree with the caution of not changing a regex, because it can be incredibly fragile to do so...

index.html
```diff
@@ -1725,6 +1727,69 @@ <h3>Validity Period</h3>
...
+and times.
+        </p>
+
+        <pre class="example nohighlight"
+          title="Regular expression to detect a valid XML Schema 1.1: Part 2 dateTimeStamp">
+-?([1-9][0-9]{3,}|0[0-9]{3})
+-(0[1-9]|1[0-2])
+-(0[1-9]|[12][0-9]|3[01])
+T(([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9](\.[0-9]+)?|(24:00:00(\.0+)?))
+(Z|(\+|-)((0[0-9]|1[0-3]):[0-5][0-9]|14:00))?
```
pchampin, line 1790:
```suggestion
(Z|(\+|-)((0[0-9]|1[0-3]):[0-5][0-9]|14:00))
```
The question mark at the end of this line contradicts the fact that the timezone is mandatory, which is the whole point of replacing dateTime with dateTimeStamp.

pchampin (changes requested):

index.html
msporny: Excellent catch, @pchampin -- the regex was for `dateTime` and not `dateTimeStamp`, which modifies the `dateTime` regex via prose (and not by re-stating the proper regex).

Fixed.

msporny: Hmm, on second inspection, looks like the regexes are designed to stack... so, the narrower one for `dateTime` would apply. IOW, I believe the regex as it is now is equivalent to applying the `dateTime` regex and then the `dateTimeStamp` regex.

index.html
```diff
@@ -1682,23 +1682,25 @@ <h3>Validity Period</h3>
         <dl>
           <dt><var>validFrom</var></dt>
           <dd>
-A <a>credential</a> MUST have an <code>validFrom</code> <a>property</a>. The
+A <a>credential</a> MUST have a <code>validFrom</code> <a>property</a>. The
 value of the <code>validFrom</code> <a>property</a> MUST be a string value of
```
TallTed, line 1686:
```suggestion
value of the <code>validFrom</code> <a>property</a> MUST be
```

index.html
```diff
@@ -1682,23 +1682,25 @@ <h3>Validity Period</h3>
...
-an [<a data-cite="XMLSCHEMA11-2#dateTime">XMLSCHEMA11-2</a>] combined
-<code>date-time</code> string representing the date and time the
+an [<a data-cite="XMLSCHEMA11-2#dateTimeStamp">XMLSCHEMA11-2</a>] combined
+<code>dateTimeStamp</code> string representing the date and time the
```
TallTed, line 1688:
```suggestion
<code>dateTimeStamp</code> string value representing the date and time the
```

index.html
```diff
@@ -1682,23 +1682,25 @@ <h3>Validity Period</h3>
...
-  If present, the value of the <code>validUntil</code> <a>property</a> MUST be
-  a string value of an [<a data-cite="XMLSCHEMA11-2#dateTime">XMLSCHEMA11-2</a>]
-  combined <code>date-time</code> string representing the date and time the
-  <a>credential</a> ceases to be valid, which could be a date and time in the past.
-  Note that this value represents the latest point in time at which the information
-  associated with the <code>credentialSubject</code> <a>property</a> is valid.
+If present, the value of the <code>validUntil</code> <a>property</a> MUST be a
+string value of an
+[<a data-cite="XMLSCHEMA11-2#dateTimeStamp">XMLSCHEMA11-2</a>] combined
+<code>dateTimeStamp</code> string representing the date and time the
```
TallTed, line 1699:
```suggestion
If present, the value of the <code>validUntil</code> <a>property</a> MUST be an
[<a data-cite="XMLSCHEMA11-2#dateTimeStamp">XMLSCHEMA11-2</a>] combined
<code>dateTimeStamp</code> string value representing the date and time the
```

index.html
```diff
@@ -1725,6 +1727,69 @@ <h3>Validity Period</h3>
...
+are not implemented in a uniform manner throughout the world. For example:
+        </p>
+
+        <ul>
+          <li>
+The observation of daylight savings time does not start or end on the same day
+and time, if at all, between various regions.
+          </li>
+          <li>
+The adoption of the Gregorian calendar did not happen in a single year, it was
```
TallTed, line 1743:
```suggestion
The adoption of the Gregorian calendar did not happen in a single year; it was
```

index.html
```diff
@@ -1725,6 +1727,69 @@ <h3>Validity Period</h3>
...
+
+        <p>
+These are just a few examples that illustrate that whether or not a time is seen
+as valid can be based on cultural norms. In other words, some dates and times
+exist for some cultures, while they do not exist for other cultures.
+        </p>
+
+        <p>
+This specification attempts to increase the number of universally recognized
+combinations of dates and times, and reduce the potentials for
```
TallTed, line 1760:
```suggestion
combinations of dates and times, and reduce the potential for
```

index.html
```diff
@@ -1725,6 +1727,69 @@ <h3>Validity Period</h3>
...
+mis-interpretation of time values, by utilizing the
+<code>dateTimeStamp</code> construction first established by the [<a
+data-cite="XMLSCHEMA11-2#dateTimeStamp">XMLSCHEMA11-2</a>] specification. In
+order to reduce mis-interpretations between different cultures, all times are
```
TallTed, line 1764:
```suggestion
misinterpretation of time values, by utilizing the
<code>dateTimeStamp</code> construction first established by the [<a
data-cite="XMLSCHEMA11-2#dateTimeStamp">XMLSCHEMA11-2</a>] specification. In
order to reduce misinterpretations between different cultures, all times are
```

index.html
```diff
@@ -1725,6 +1727,69 @@ <h3>Validity Period</h3>
...
+specified as offsets against Universal Coordinated Time (UTC).
+        </p>
+
+        <p>
+Implementers that desire to check <code>dateTimeStamp</code> values for validity
+can use the regular expression provided below, which is reproduced from the [<a
+data-cite="XMLSCHEMA11-2#dateTimeStamp">XMLSCHEMA11-2</a>] specification for the
+purposes of convenience. For the avoidance of doubt, the regular expression in
```
TallTed, line 1772:
```suggestion
data-cite="XMLSCHEMA11-2#dateTimeStamp">XMLSCHEMA11-2</a>] specification for
convenience. To avoid doubt, the regular expression in
```

TallTed (changes requested):
Some minor cleanup

index.html
```diff
@@ -1757,19 +1756,19 @@ <h4>Representing Time</h4>
...
 This specification attempts to increase the number of universally recognized
-combinations of dates and times, and reduce the potentials for
-mis-interpretation of time values, by utilizing the
+combinations of dates and times, and reduce the potential for
+misinterpretation of time values, by utilizing the
 <code>dateTimeStamp</code> construction first established by the [<a
 data-cite="XMLSCHEMA11-2#dateTimeStamp">XMLSCHEMA11-2</a>] specification. In
-order to reduce mis-interpretations between different cultures, all times are
+order to reduce misinterpretations between different cultures, all times are
 specified as offsets against Universal Coordinated Time (UTC).
```
aphillips: A key problem when using offsets is that they do not represent the set of rules used to compute the offset. The rules, which are bound up in a "time zone", are subject to change and, as they are controlled by various national and sometimes sub-national governments and subject to local conditions, they _do_ change on a not-infrequent basis. Generally the change is small (by an hour here or there), so the impact is not perhaps critical to most applications. However, it does mean that a `validFrom` or `validTo` computed according to a local wall time under one set of time zone rules might start/expire earlier/later than expected due to a subsequent change to time zone rules. In particular, it could leave a gap of an hour or two between one credential's expiration and a new credential's validity if the new credential was computed with new time zone rules.

Standards for encoding the time zone into a time stamp are not yet "de jurre" and it seems unlikely that XML Schema will be revised to accommodate newer serializations incorporating zone IDs. For a peek into serialization types, see [this work in progress](https://w3c.github.io/timezone/#serializations) by I18N.

---

I don't see guidance to normalize dates to use UTC. That's a best practice which the specification can call out to users to avoid problems with time zone and it allows easier 1:1 comparison between separately generated values (even if it obscures what the local wall time was/will be)

---

> In order to reduce misinterpretations between different cultures, all times are
> specified as offsets against Universal Coordinated Time (UTC).

I think "cultures" is inappropriate here. Locale and culture are really orthogonal to the representation of time. The use of ISO8601-based serialization (and the preleptic Gregorian calendar inherent in that serialization) is what deals with cultural variation regarding the representation of time. The use of UTC and UTC offsets reduces the misinterpretation between _regions_.

aphillips (changes requested):
Generally good improvements, however the discussion of the Gregorian calendar still needs a bit of adjustment. The discussion of the use of offsets also could be improved: implementers should know where the "sharp edges" are to avoid unintentional gaps between two credentials due to time zone changes.

index.html
aphillips: Note also that we're doing a review of this spec and there is a comment related to this PR [here](https://github.com/w3c/i18n-activity/issues/1729). This comment has not yet been reviewed by the I18N WG.

decentralgabe (approved):

decentralgabe (approved):

index.html
```diff
@@ -1740,7 +1739,7 @@ <h4>Representing Time</h4>
 and time, if at all, between various regions.
           </li>
           <li>
-The adoption of the Gregorian calendar did not happen in a single year, it was
+The adoption of the Gregorian calendar did not happen in a single year; it was
```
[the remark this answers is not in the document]
msporny: Ok, then I'll strike the text about the Gregorian calendar and replace it with something more relevant along the lines of: "other calendaring systems exist and are use by various cultures..."

index.html
```diff
@@ -1682,23 +1682,24 @@ <h3>Validity Period</h3>
         <dl>
           <dt><var>validFrom</var></dt>
           <dd>
-A <a>credential</a> MUST have an <code>validFrom</code> <a>property</a>. The
-value of the <code>validFrom</code> <a>property</a> MUST be a string value of
-an [<a data-cite="XMLSCHEMA11-2#dateTime">XMLSCHEMA11-2</a>] combined
-<code>date-time</code> string representing the date and time the
+A <a>credential</a> MUST have a <code>validFrom</code> <a>property</a>. The
```
OR13, line 1685: What happens when this MUST is not enforced by an issuer?

What error does the verify raise?

index.html
```diff
@@ -1725,6 +1726,69 @@ <h3>Validity Period</h3>
...
+          <li>
+The observation of daylight savings time does not start or end on the same day
+and time, if at all, between various regions.
+          </li>
+          <li>
+The adoption of the Gregorian calendar did not happen in a single year; it was
+phased in over many years across various societies.
+          </li>
+          <li>
+Leap seconds might not be taken into account in all software systems, especially
```
OR13, line 1746: excellent to see this comment regarding leap seconds make its way into the spec.

index.html
```diff
@@ -1725,6 +1726,69 @@ <h3>Validity Period</h3>
...
+days in every month, which allows for leap years, and leap seconds, as well as
+days in places where they do not exist. That said, modern system libraries that
+generate <code>dateTimeStamp</code> are often error-free in their generation of
+valid <code>dateTimeStamp</code> values and thus the regular expression shown
+below is often adequate when processing modern system library-generated dates
+and times.
+        </p>
+
+        <pre class="example nohighlight"
+          title="Regular expression to detect a valid XML Schema 1.1: Part 2 dateTimeStamp">
```
OR13, line 1784: thank you!

OR13 (changes requested):
The only blocking feedback I have is regarding to what happens when bad times are serialized in JSON... maybe we just file an issue for that?

Just not sure what a processor will do with the normative statements.

index.html
```diff
@@ -1725,6 +1726,69 @@ <h3>Validity Period</h3>
...
+combinations of dates and times, and reduce the potential for
+misinterpretation of time values, by utilizing the
+<code>dateTimeStamp</code> construction first established by the [<a
+data-cite="XMLSCHEMA11-2#dateTimeStamp">XMLSCHEMA11-2</a>] specification. In
+order to reduce misinterpretations between different cultures, all times are
+specified as offsets against Universal Coordinated Time (UTC).
+        </p>
+
+        <p>
+Implementers that desire to check <code>dateTimeStamp</code> values for validity
```
davidlehn, line 1768: Why copy the regex here instead of [linking](https://www.w3.org/TR/xmlschema11-2/#dateTime-lexical-mapping) to it?  How to interpret time correctly seems out of place in a spec like this.

index.html
```diff
@@ -1725,6 +1726,69 @@ <h3>Validity Period</h3>
 }
         </pre>
 
+        <h4>Representing Time</h4>
+
+        <p>
+When representing time values, implementers are urged to understand that times
+are not implemented in a uniform manner throughout the world. For example:
```
davidlehn, line 1733: There are [many lists of date/time falsehoods](https://github.com/kdeldycke/awesome-falsehood#dates-and-time) that all programmers should familiarize themselves with.  And date and time handling in general is a large topic.  Listing only a few random examples of time handling issues here seems out of place.  Can this text be more general rather than listing examples?  If anything, I'd say the best place for non-trivial examples would be a test suite.

index.html
msporny: > What happens when this MUST is not enforced by an issuer?

We answer that question here:

https://w3c.github.io/vc-data-model/#conformance

with this text:

> A conforming processor is any algorithm realized as software and/or hardware that generates or consumes a [conforming document](https://w3c.github.io/vc-data-model/#dfn-conforming-document). Conforming processors MUST produce errors when non-conforming documents are consumed.

> What error does the verify raise?

I'm saying this w/o a value judgement... at present, we don't say, on purpose:

> This specification makes no normative statements with regard to the conformance of roles in the ecosystem, such as [issuers](https://w3c.github.io/vc-data-model/#dfn-issuers), [holders](https://w3c.github.io/vc-data-model/#dfn-holders), or [verifiers](https://w3c.github.io/vc-data-model/#dfn-verifier), because the conformance of ecosystem roles are highly application, use case, and market vertical specific.

The group tried to define the algorithm for VC verification and WG members disagreed on when what types of errors should be raised. A lot of time has passed since we had that previous discussion, and I would expect that (if we had the time), we could get to some sort of algorithm with some sort of set of errors defined. That said, given the short time we have until the CR phase, having that discussion now could eat up a ton of time.

So, for right now, we just say that a processor has to raise an error and we're silent on what that error should be.

index.html
```diff
@@ -1725,6 +1726,69 @@ <h3>Validity Period</h3>
 }
         </pre>
 
+        <h4>Representing Time</h4>
```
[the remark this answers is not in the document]
msporny: No, this is not just about XML or RDF. XML Schema 1.1: Part 2 `dateTimeStamp` is a valid subset of ISO-8601 (which has multiple release years itself), which is used across a broad swath of programming languages and operating environments.

index.html
msporny: > Can this text be more general rather than listing examples?

It could, but I'd be guessing at what sort of "more general" text you're looking for. If you can suggest something concrete, I might be able to incorporate it.

index.html
msporny: Because the issue this PR is trying to address specifically requested that the Regex be copied into the spec: https://github.com/w3c/vc-data-model/issues/846 

I agree that linking to it is probably better, but that would mean that implementers would have to interpret the way XML Schema 1.1 Part 2 layers regexes to achieve the regex that's provided in this PR.

If there is somewhere else that we could point to that highlights the issues regarding interpretation of times, let me know. I couldn't find a good place to point to that had a summary like this.

index.html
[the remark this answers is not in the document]
msporny: Fixed in 299a72968b41417a51dc14472877e9fbf6658acf.

index.html
msporny: Ok, focused on "regions" vs. "cultures" wrt. language around why this stuff is more complicated than it seems at first. Fixed in 01368c36ed23ef9e5d8421f8b350b728e6198d0d.

I also added a normative requirement to encode against Zulu time, but allowed for offsets. That seems to go against the guidance you gave above, primarily because I don't understand why encoding to an offset from Zulu is any different than encoding to Zulu. That is, whatever `dateTimeStamp` you create is going to be based off of your local time, isn't it? That is, you always lose the set of rules used to construct the time?

I'm happy to change it to something else, but I'm struggling w/ what the XML Schema 1.1: Part 2 spec says and what you're conveying (and the normative regex). I'm trying very hard to not modify the regex from what's used in  XML Schema 1.1: Part 2.

msporny:
@OR13 wrote:
> The only blocking feedback I have is regarding to what happens when bad times are serialized in JSON... maybe we just file an issue for that?

We answer that question in the spec today in this section:

https://w3c.github.io/vc-data-model/#conformance

with this text:

> A conforming processor is any algorithm realized as software and/or hardware that generates or consumes a [conforming document](https://w3c.github.io/vc-data-model/#dfn-conforming-document). Conforming processors MUST produce errors when non-conforming documents are consumed.

> Just not sure what a processor will do with the normative statements.

A processor MUST throw an error, the spec has been pretty clear on that for a while. 

As to what sort of error, and I say this w/o a value judgement on whether this is good or bad... at present, we don't say anything, on purpose. See this text in the "conformance" section of the spec:

> This specification makes no normative statements with regard to the conformance of roles in the ecosystem, such as [issuers](https://w3c.github.io/vc-data-model/#dfn-issuers), [holders](https://w3c.github.io/vc-data-model/#dfn-holders), or [verifiers](https://w3c.github.io/vc-data-model/#dfn-verifier), because the conformance of ecosystem roles are highly application, use case, and market vertical specific.

The group tried to define the algorithm for VC verification and WG members disagreed on when what types of errors should be raised (and what the values for those errors should be). A lot of time has passed since we had that previous discussion, and I would expect that (if we had the time), we could get to some sort of algorithm with some sort of set of errors defined. That said, given the short time we have until the CR phase, having that discussion now could eat up a ton of time.

So, for right now, we just say that a processor has to raise an error and we're silent on what that error should be. Based on the above, what we have in the spec is probably the best we can do for now, unless you can think of some text that would gain consensus (ideally, quickly)?

index.html
```diff
@@ -1705,6 +1706,70 @@ <h3>Validity Period</h3>
...
+
+        <p>
+When representing time values, implementers are urged to understand that times
+are not implemented in a uniform manner throughout the world. For example:
+        </p>
+
+        <ul>
+          <li>
+The observation of daylight savings time does not start or end on the same day
```
dlongley, line 1718:
```suggestion
The observation of daylight saving time does not start or end on the same day
```

dlongley (approved):

OR13 (approved):

brentzundel (approved):

index.html
davidlehn: I think copying other spec details is unneeded, error prone, and bloats this spec.  If you do insist on including it, at least add the RFC note about removing whitespace in the regex.

index.html
davidlehn: I'm not sure where to take this.  It seems unnecessary to remind implementers to process dates and times correctly.  Beyond representation requirements, why is any processing advice needed?

msporny:
Waiting on @aphillips' response to some responses to his feedback to merge this PR. I want to make sure it addresses his concerns before merging.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-07-11-vcwg-special#section1-8) on 2023-07-11

- no resolutions were taken

#### 1.8. Add dateTimeStamp regex and explanation of representing time. (pr vc-data-model#1183)

_See github pull request [vc-data-model#1183](https://github.com/w3c/vc-data-model/pull/1183)._




**Brent Zundel:** this clarifies the date time requirements for things like valid from and till. Provides a regex for validation. P.R has broad approval, awaiting a final thumbs up.  
… from chair of the internationalization working group.  
… open to comments.  

**Ivan Herman:** isn't it better to make this one and previous P.R. to say that as soon as submitter is finished they can merge the p.r.  

**Brent Zundel:** not in the habit of doing this in the group.  
… happy to find a label/make a comment to that affect.  

</details>

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-07-11-vcwg-special#section1-8) on 2023-07-11

- no resolutions were taken

#### 1.8. Add dateTimeStamp regex and explanation of representing time. (pr vc-data-model#1183)

_See github pull request [vc-data-model#1183](https://github.com/w3c/vc-data-model/pull/1183)._




**Brent Zundel:** this clarifies the date time requirements for things like valid from and till. Provides a regex for validation. P.R has broad approval, awaiting a final thumbs up.  
… from chair of the internationalization working group.  
… open to comments.  

**Ivan Herman:** isn't it better to make this one and previous P.R. to say that as soon as submitter is finished they can merge the p.r.  

**Brent Zundel:** not in the habit of doing this in the group.  
… happy to find a label/make a comment to that affect.  

</details>

index.html
aphillips: I'll also add: the "tripping hazard" related to time zones and DST/ST for your specification's users is that governments change time zone rules. These changes can result in a given wall time being expressed differently. For example, consider if the zone `America/Los_Angeles`, which has a raw offset of UTC-8 and observes DST decided to stop observing DST in 2024.

A given certificate might have been generated to be:

```json
"validUntil": "2024-07-12T12:00:00-07:00"  // noon on July 12
```

The replacement certificate, created after the change to the time zone might have:

```json
"validFrom": "2024-07-12T12:00:00-08:00" // noon on July 12
```

This leaves a gap of an hour...

Consider adding a note or comment such as:

```suggestion
<p class="note">Time zone definitions are occasionally changed by their governing body. When replacing or issuing new credentials, care should be exercised to ensure that changes to local time zone rules do not result in unexpected gaps in validity.</p>
```

(You could then adapt the example above. I intend to add this example to the "past and future" use case in the timezone note.)

index.html
aphillips: Leap seconds can affects highly sensitive systems that depend on the exact millisecond offset from the epoch. 

However, note that for most applications using ISO8601 serializations the only moment in time that is affected is the one second period of the leap second itself. That is, the moment *after* the most recent leap second can always be represented as `2017-01-01T00:00:00Z`, regardless of whether the system in question understands leap seconds.

index.html
```diff
@@ -1705,6 +1706,70 @@ <h3>Validity Period</h3>
...
+Leap seconds might not be taken into account in all software systems, especially
+for dates and times that precede the introduction of the leap second.
+          </li>
+        </ul>
+
+        <p>
+These are just a few examples that illustrate that whether or not a time is seen
+as valid can be based on the region in which one resides. In other words, some
+dates and times exist in some regions, while they do not exist for other
+regions.
```
aphillips, line 1735: This is misleading. Some **_wall time_** values do not exist in certain regions, but the timeline is continuous. Further, for `dateTimeStamp` values, it is always possible to normalize any valid value to an incremental (think `Date`) value, even if the digits in the time stamp never show on local clocks (generally because of DST/ST rules).

For example, in `America/Los_Angeles`, DST started at `2023-03-12T02:00:00-08:00` (the next second was `2023-03-12T03:00:01-07:00` in local time). However, it is perfectly possible to interpret `2023-03-12T02:00:01-08:00` to be the same moment in time because an offset is not a time zone.

aphillips (changes requested):
This is nearly there, although I have some lengthy comments. I think you could remove a lot of the text about calendars and such and focus on the problem of checking validity (the regex) and on implementer/user advice regarding zone offsets.

TallTed (commented):
Making good progress. Needs action on the latest comments from @aphillips and @davidlehn, after which I will re-review and almost certainly approve.

index.html
```diff
@@ -1657,21 +1657,22 @@ <h3>Validity Period</h3>
...
           </dd>
           <dt><var>validUntil</var></dt>
           <dd>
-  If present, the value of the <code>validUntil</code> <a>property</a> MUST be
-  a string value of an [<a data-cite="XMLSCHEMA11-2#dateTime">XMLSCHEMA11-2</a>]
-  combined <code>date-time</code> string representing the date and time the
-  <a>credential</a> ceases to be valid, which could be a date and time in the past.
-  Note that this value represents the latest point in time at which the information
-  associated with the <code>credentialSubject</code> <a>property</a> is valid.
+If present, the value of the <code>validUntil</code> <a>property</a> MUST be an
```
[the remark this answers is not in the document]
msporny: Fixed in 1ccdc3c54bce27c888554ffafd8852f65abbdb34.

index.html
msporny: I'll raise another PR to try and remove the text and see if it gets traction. At this point, people are grumbling to the opening paragraph, but are not objecting. I'm going to merge it as is, and then try to remove it and see if people are supportive. I'll tag both of you on that PR.

index.html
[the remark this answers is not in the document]
msporny: Also added your DST example here: 9bb9ffe0a8d2dcb46a972214e845f66bf87f3c04

index.html
msporny: Fixed in e555089c2c41dd63b613b2aabbda8f73b12c9881.

index.html
msporny: Fixed in 00ee176d6df204e879e02e36c360b817d0a4d1ee.

index.html
msporny: Fixed in 555d69fcb5f7289f9474a7bc541d59f59cfbf9ac.

msporny:
Alright, I've integrated all of @davidlehn and @aphillips latest suggestions. This section is getting much larger than I had intended, but I can't cut anything w/o rejecting input from folks. I'll raise a separate PR to see what we can cut (and if we can get consensus around cutting some of the verbiage).

@TallTed, please re-review.

aphillips (approved):
LGTM.

Thank you for persevering through this. I hope we (I18N) can provide both our note on working with time and updated guidelines in the best practices docs to help others not have to repeat this.

msporny:
> Thank you for persevering through this. I hope we (I18N) can provide both our note on working with time and updated guidelines in the best practices docs to help others not have to repeat this.

Yes, that would be great. Also note that, if that document is done in time, we could (and should) reference it from this section of the spec.

msporny:
> Making good progress. Needs action on the latest comments from @aphillips and @davidlehn, after which I will re-review and almost certainly approve.

Your re-review is the last thing we're waiting on for this PR @TallTed... merging after we hear from you.

index.html
```diff
@@ -1772,9 +1772,9 @@ <h4>Representing Time</h4>
 days in every month, which allows for leap years, and leap seconds, as well as
 days in places where they do not exist. That said, modern system libraries that
 generate <code>dateTimeStamp</code> values are often error-free in their
-generation of valid <code>dateTimeStamp</code> values and thus the regular
-expression shown below is often adequate when processing modern system
-library-generated dates and times.
+generation of valid <code>dateTimeStamp</code> values. The regular
+expression shown below, with whitespace added for readability, is often
+adequate when processing modern system library-generated dates and times.
```
TallTed, line 1782:
```suggestion
expression shown below (minus the whitespace included here for readability),
is often adequate when processing library-generated dates and times on
modern systems.
```

msporny:

Issue #860, opened 2022-01-12
Determine interoperable way for Holder to make claims directly in VPs


jandrieu:
VPs include a "verifiableCredential" for embedding VCs in the VP.

It's also possible to add properties to the VP that can represent claims made by the holder about the claims presented, thanks to the open world data model.

For example, in the VC Use Cases,  [5.1 Citizenship by Parentage
](https://www.w3.org/TR/vc-use-cases/#citizenship-by-parentage) the holder claims that he is the person born in the birth certificate, his mother is the parent in that birth certificate, that the person in the US passport is his mother, and that one of the subjects in the marriage license is his mother, all using the identifiers in those VCs to describe the relationship between the VCs.

You could of course, have the holder self-issue a VC and include in as any other VC. However, in practice, that is an unfortunate overhead, needing two signatures (the VC and the VP). It's easier just to add a property to the VP and get rid of the extra proof.

I'd like to see this standardized in some way (a candidate for v2.0). I initially considered "holderSubject" for symmetry with "credentialSubject" in the VCs. However, "holderClaims" might be more appropriate as statements about his mother have a different subject.

In a very pseudocode example:
```json
"holderClaims": [{ 
  "id":"did:example.com:mom",
  "sameAs": [
    "did:example.com:momInBirthCertificate",
    "did:example.com:momInMarriageLicense",
    "did:example.com:momInPassport",
   ]
  "parentOf": "did:example.com:holder"
}]
```

jandrieu:
Discussed on WG call 2022.08.24. @msporny and @dlongley advocated for sticking with the use of VCs rather than supporting a different way to make claims.

@jandrieu Suggested that unless we change the open-world data model of the VP itself, this behavior "holderClaims" as a property in the VP is allowed, and as such, is likely be used by developers who find it to be a more convenient solution. I could support a debate about moving away from an open world model (and considering restricting this practice), but if we stay with the current spec, which allows arbitrary properties in the VP, then standardizing the use would increase interoperability.

Note that if a developer wants to stick with the VC-based approach, a standard property for holder asserted claims, such as "holderClaims" would not prevent that approach. However, today, the only interoperable option requires duplicate signatures and duplicate verifications. In some use cases, the additional cryptography is more of a burden than justified, e.g., in IOT devices where computation and data size are both at a premium.

Kristina mentioned there is support for this use case and recommended we leave the conversation open for people to consider.

No resolutions made.

agropper:
[...]

PR #1186, opened 2023-07-05
Add interoperable way for holder-asserted claims in a VP

brentzundel:
This PR addresses #860 by introducing normative guidance for a holder to include self-asserted claims inside of a VP.
It is based on the conversation in #860 where it was suggested that the holder should construct a VC with the self-asserted claims and include that in the VP.

***
<a href="https://pr-preview.s3.amazonaws.com/brentzundel/vc-data-model/pull/1186.html" title="Last updated on Jul 19, 2023, 4:55 PM UTC (9559c36)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1186/7c44300...brentzundel:9559c36.html" title="Last updated on Jul 19, 2023, 4:55 PM UTC (9559c36)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2084,7 +2084,7 @@ mechanism as the <a>verifiable presentation</a>.
   "holder": "did:example:12345678",
   "verifiableCredential": [{
     "@context": "https://www.w3.org/ns/credentials/v2",
-    "type": ["VerifiableCredential", "SelfAssertedCredential", "ExampleFoodPreferenceCredential"],
+    "type": ["VerifiableCredential", "ExampleFoodPreferenceCredential"],
     "issuer": "did:example:12345678",
     "credentialSubject": {
       "favoriteCheese": "Gouda"
@@ -2108,14 +2108,14 @@ self-asserted <a>verifiable credential</a> that holds <a>claims</a> about the
     "https://www.w3.org/ns/credentials/examples/v2"
   ],
   "type": ["VerifiablePresentation", "ExamplePresentation"],
-  <span class="highlight">"id": "http://example.com/presentation/5223"</span>,
+  <span class="highlight">"id": "urn:uuid:313801ba-24b7-11ee-be02-ff560265cf9b"</span>,
   "holder": "did:example:12345678",
   "verifiableCredential": [{
     "@context": "https://www.w3.org/ns/credentials/v2",
     "type": ["VerifiableCredential", "ExampleTimeLimitCredential"],
     "issuer": "did:example:12345678",
     "credentialSubject": {
-      <span class="highlight">"id": "http://example.com/presentation/5223"</span>,
+      <span class="highlight">"id": "urn:uuid:313801ba-24b7-11ee-be02-ff560265cf9b"</span>,
       "validUntil": "2024-01-01T19:23:24Z"
     },
     { <span class="comment">...</span> }
--- a/index.html
+++ b/index.html
@@ -2033,15 +2033,24 @@ encoding would be determined by the schema.
         <section>
           <h4>Presentations Including Holder Claims</h4>
           <p>
-A <a>holder</a> MAY use the <code>verifiableCredential</code>
-<a>property</a> in a <a>verifiable presentation</a> to include
-<a>verifiable credentials</a> from any <a>issuer</a>, including
-themselves. When the <a>issuer</a> of a <a>verifiable credential</a>
-is the <a>holder</a>, its <a>claims</a> are considered to be
-self-asserted. These self-asserted claims can be secured by their
-own mechanism like any other <a>verifiable credential</a> or by
-the same mechanism that secures the <a>verifiable presentation</a>
-in which they are included.
+A <a>holder</a> MAY use the <code>verifiableCredential</code> <a>property</a> in
+a <a>verifiable presentation</a> to include <a>verifiable credentials</a> from
+any <a>issuer</a>, including themselves. When the <a>issuer</a> of a
+<a>verifiable credential</a> is the <a>holder</a>, its <a>claims</a> are
+considered to be self-asserted. These self-asserted claims can be secured by
+their own mechanism like any other <a>verifiable credential</a> or by the same
+mechanism that secures the <a>verifiable presentation</a> in which they are
+included.
+          </p>
+          <p>
+The <a>subject</a> of these self-asserted <a>claims</a> is not limited and can
+include statements about the <a>holder</a>, one of the other included
+<a>verifiable credentials</a>, or even the <a>verifiable presentation</a> in
+which the self-asserted <a>verifiable credential</a> is included. In each of
+these cases the <code>CredentialSubject</code> would have an <code>id</code>
+<a>property</a> that matches the <code>id</code> of the <a>subject</a>. This is
+no different than how it is done in <a>verifiable credentials</a> that are not
+self-asserted.
           </p>
           <p>
 A <a>verifiable presentation</a> that includes a self-asserted
--- a/index.html
+++ b/index.html
@@ -2045,28 +2045,41 @@ secured by the same mechanism that secures the <a>verifiable presentation</a> in
 which they are included.
           </p>
           <p>
+A <a>verifiable presentation</a> that includes a self-asserted
+<a>verifiable credential</a> that is only secured using the same mechanism as
+the <a>verifiable presentation</a> MUST include a <code>holder</code>
+<a>property</a>.
+          </p>
+          <p>
 A self-asserted <a>verifiable credential</a> that has been included in the value
 of the <code>verifiableCredential</code> <a>property</a> of a
 <a>verifiable presentation</a> MUST include the following properties:
-        </p>
+          </p>
 
         <dl>
           <dt><var>type</var></dt>
           <dd>
 The <code>type</code> <a>property</a> is required. The value MUST be
-<code>VerifiableCredential</code> and IT IS RECOMMENDED that an additional value
-<code>SelfAssertedCredential</code> be included. For example,<br>
-<code>"type": ["VerifiableCredential", "SelfAssertedCredential"]</code>
+<code>VerifiableCredential</code> and IT IS RECOMMENDED that the additional
+value <code>SelfAssertedCredential</code> be included. Additional types are also
+acceptable. For example,<br>
+<code>"type": ["VerifiableCredential", "SelfAssertedCredential", "ExampleTypeCredential"]</code>
           </dd>
           <dt><var>issuer</var></dt>
           <dd>
-The <code>issuer</code> <a>property</a> is required. The value MUST be identical
-to the <code>holder</code> <a>property</a> of the <a>verifiable presentation</a>.
+The <code>issuer</code> <a>property</a> is required. For self-asserted
+<a>verifiable credentials</a> that are secured using the same mechanism as the
+<a>verifiable presentation</a>, the value MUST be identical to the
+<code>holder</code> <a>property</a> of the <a>verifiable presentation</a>.
+Self-asserted <a>verifiable credentials</a> that use their own securing
+mechanism MUST follow the normative guidance for the <code>issuer</code>
+<a>property</a> in Section <a href="#issuer"></a>.
           </dd>
         </dl>
         <p>
 The example below shows a <a>verifiable presentation</a> that embeds a
-self-asserted <a>verifiable credential</a>.
+self-asserted <a>verifiable credential</a> that is secured using the same
+mechanism as the <a>verifiable presentation</a>.
         </p>
 
         <pre class="example nohighlight" title="Verifiable presentation with a self-asserted verifiable credential">
--- a/index.html
+++ b/index.html
@@ -2039,7 +2039,8 @@ In addition to <a>claims</a> from an <a>issuer</a> in the form of
 are considered to be self-asserted. If present, these self-asserted
 <a>claims</a> MUST be formatted as a <a>verifiable credential</a> and included
 in the value of the <code>verifiableCredential</code> <a>property</a> of the
-<a>verifiable presentation</a>. These self-asserted claims are considered to be
+<a>verifiable presentation</a>. These self-asserted claims can either use their
+own securing mechanism like any other <a>verifiable credential</a> or be
 secured by the same mechanism that secures the <a>verifiable presentation</a> in
 which they are included.
           </p>
--- a/index.html
+++ b/index.html
@@ -2033,16 +2033,15 @@ encoding would be determined by the schema.
         <section>
           <h4>Presentations Including Holder Claims</h4>
           <p>
-In addition to <a>claims</a> from an <a>issuer</a> in the form of
-<a>verifiable credentials</a>, a <a>verifiable presentation</a> MAY include
-<a>claims</a> directly from the <a>holder</a>. These <a>holder</a> <a>claims</a>
-are considered to be self-asserted. If present, these self-asserted
-<a>claims</a> MUST be formatted as a <a>verifiable credential</a> and included
-in the value of the <code>verifiableCredential</code> <a>property</a> of the
-<a>verifiable presentation</a>. These self-asserted claims can be secured
-by their own mechanism like any other <a>verifiable credential</a> or by
-the same mechanism that secures the <a>verifiable presentation</a> in which
-they are included.
+A <a>holder</a> MAY use the <code>verifiableCredential</code>
+<a>property</a> in a <a>verifiable presentation</a> to include
+<a>verifiable credentials</a> from any <a>issuer</a>, including
+themselves. When the <a>issuer</a> of a <a>verifiable credential</a>
+is the <a>holder</a>, its <a>claims</a> are considered to be
+self-asserted. These self-asserted claims can be secured by their
+own mechanism like any other <a>verifiable credential</a> or by
+the same mechanism that secures the <a>verifiable presentation</a>
+in which they are included.
           </p>
           <p>
 A <a>verifiable presentation</a> that includes a self-asserted
--- a/index.html
+++ b/index.html
@@ -2053,8 +2053,8 @@ of the <code>verifiableCredential</code> <a>property</a> of a
           <dt><var>type</var></dt>
           <dd>
 The <code>type</code> <a>property</a> is required. The value MUST be
-<code>VerifiableCredential</code> and <code>SelfAssertedCredential</code>.For
-example,<br>
+<code>VerifiableCredential</code> and IT IS RECOMMENDED that an additional value
+<code>SelfAssertedCredential</code> be included. For example,<br>
 <code>"type": ["VerifiableCredential", "SelfAssertedCredential"]</code>
           </dd>
           <dt><var>issuer</var></dt>
@@ -2078,7 +2078,7 @@ self-asserted <a>verifiable credential</a>.
   "holder": "did:example:12345678",
   "verifiableCredential": [{
     "@context": "https://www.w3.org/ns/credentials/v2",
-    "type": ["VerifiableCredential", "SelfAssertedCredential"],
+    "type": ["VerifiableCredential", "SelfAssertedCredential", "ExampleFoodPreferenceCredential"],
     "issuer": "did:example:12345678",
     "credentialSubject": {
       "favoriteCheese": "Gouda"
--- a/index.html
+++ b/index.html
@@ -2030,6 +2030,65 @@ encoding would be determined by the schema.
             </figcaption>
           </figure>
         </section>
+        <section>
+          <h4>Presentations Including Holder Claims</h4>
+          <p>
+In addition to <a>claims</a> from an <a>issuer</a> in the form of
+<a>verifiable credentials</a>, a <a>verifiable presentation</a> MAY include
+<a>claims</a> directly from the <a>holder</a>. These <a>holder</a> <a>claims</a>
+are considered to be self-asserted. If present, these self-asserted
+<a>claims</a> MUST be formatted as a <a>verifiable credential</a> and included
+in the value of the <code>verifiableCredential</code> <a>property</a> of the
+<a>verifiable presentation</a>. These self-asserted claims are considered to be
+secured by the same mechanism that secures the <a>verifiable presentation</a> in
+which they are included.
+          </p>
+          <p>
+A self-asserted <a>verifiable credential</a> that has been included in the value
+of the <code>verifiableCredential</code> <a>property</a> of a
+<a>verifiable presentation</a> MUST include the following properties:
+        </p>
+
+        <dl>
+          <dt><var>type</var></dt>
+          <dd>
+The <code>type</code> <a>property</a> is required. The value MUST be
+<code>VerifiableCredential</code> and <code>SelfAssertedCredential</code>.For
+example,<br>
+<code>"type": ["VerifiableCredential", "SelfAssertedCredential"]</code>
+          </dd>
+          <dt><var>issuer</var></dt>
+          <dd>
+The <code>issuer</code> <a>property</a> is required. The value MUST be identical
+to the <code>holder</code> <a>property</a> of the <a>verifiable presentation</a>.
+          </dd>
+        </dl>
+        <p>
+The example below shows a <a>verifiable presentation</a> that embeds a
+self-asserted <a>verifiable credential</a>.
+        </p>
+
+        <pre class="example nohighlight" title="Verifiable presentation with a self-asserted verifiable credential">
+{
+  "@context": [
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
+  ],
+  "type": ["VerifiablePresentation", "ExamplePresentation"],
+  "holder": "did:example:12345678",
+  "verifiableCredential": [{
+    "@context": "https://www.w3.org/ns/credentials/v2",
+    "type": ["VerifiableCredential", "SelfAssertedCredential"],
+    "issuer": "did:example:12345678",
+    "credentialSubject": {
+      "favoriteCheese": "Gouda"
+    },
+    { <span class="comment">...</span> }
+  }],
+  "proof": [{ <span class="comment">...</span> }]</span>
+}
+        </pre>
+        </section>
       </section>
     </section>
 
--- a/index.html
+++ b/index.html
@@ -2047,7 +2047,7 @@ The <a>subject(s)</a> of these self-asserted <a>claims</a> are not limited, so
 these <a>claims</a> can include statements about the <a>holder</a>, one of the
 other included <a>verifiable credentials</a>, or even the <a>verifiable
 presentation</a> in which the self-asserted <a>verifiable credential</a> is
-included. In each of these cases, the <code>id</code> <a>property</a>
+included. In each case, the <code>id</code> <a>property</a>
 is used to identify the specific <a>subject</a>, in the object where the
 <a>claims</a> about it are made, just as it is done in
 <a>verifiable credentials</a> that are not self-asserted.
@@ -2063,10 +2063,11 @@ All of the normative requirements defined for <a>verifiable credentials</a>
 apply to self-asserted <a>verifiable credentials</a>.
           </p>
           <p>
-For self-asserted <a>verifiable credentials</a> that are secured using the same
+When a self-asserted <a>verifiable credential</a> is secured using the same
 mechanism as the <a>verifiable presentation</a>, the value of the
-<code>issuer</code> <a>property</a> MUST be identical to the <code>holder</code>
-<a>property</a> of the <a>verifiable presentation</a>.
+<code>issuer</code> <a>property</a> of the <a>verifiable credential</a>
+MUST be identical to the <code>holder</code> <a>property</a> of the
+<a>verifiable presentation</a>.
           </p>
           <p>
 The example below shows a <a>verifiable presentation</a> that embeds a
@@ -2074,7 +2075,7 @@ self-asserted <a>verifiable credential</a> that is secured using the same
 mechanism as the <a>verifiable presentation</a>.
           </p>
 
-          <pre class="example nohighlight" title="A secured verifiable presentation relying on an embedded proof, including a self-asserted verifiable credential">
+          <pre class="example nohighlight" title="A verifiable presentation, secured with an embedded proof, with a self-asserted verifiable credential">
 {
   "@context": [
     "https://www.w3.org/ns/credentials/v2",
@@ -2101,7 +2102,7 @@ self-asserted <a>verifiable credential</a> that holds <a>claims</a> about the
 <a>verifiable presentation</a>.
           </p>
 
-          <pre class="example nohighlight" title="A secured verifiable presentation, relying on an embedded proof, with a self-asserted verifiable credential about the verifiable presentation">
+          <pre class="example nohighlight" title="A verifiable presentation, secured with an embedded proof, with a self-asserted verifiable credential about the verifiable presentation">
 {
   "@context": [
     "https://www.w3.org/ns/credentials/v2",
--- a/index.html
+++ b/index.html
@@ -2039,10 +2039,10 @@ In addition to <a>claims</a> from an <a>issuer</a> in the form of
 are considered to be self-asserted. If present, these self-asserted
 <a>claims</a> MUST be formatted as a <a>verifiable credential</a> and included
 in the value of the <code>verifiableCredential</code> <a>property</a> of the
-<a>verifiable presentation</a>. These self-asserted claims can either use their
-own securing mechanism like any other <a>verifiable credential</a> or be
-secured by the same mechanism that secures the <a>verifiable presentation</a> in
-which they are included.
+<a>verifiable presentation</a>. These self-asserted claims can be secured
+by their own mechanism like any other <a>verifiable credential</a> or by
+the same mechanism that secures the <a>verifiable presentation</a> in which
+they are included.
           </p>
           <p>
 A <a>verifiable presentation</a> that includes a self-asserted
@@ -2059,10 +2059,10 @@ of the <code>verifiableCredential</code> <a>property</a> of a
         <dl>
           <dt><var>type</var></dt>
           <dd>
-The <code>type</code> <a>property</a> is required. The value MUST be
-<code>VerifiableCredential</code> and IT IS RECOMMENDED that the additional
-value <code>SelfAssertedCredential</code> be included. Additional types are also
-acceptable. For example,<br>
+The <code>type</code> <a>property</a> is required. One value MUST be
+<code>VerifiableCredential</code>, and it is RECOMMENDED that another
+value be <code>SelfAssertedCredential</code>. Additional types MAY also
+be included. For example,<br>
 <code>"type": ["VerifiableCredential", "SelfAssertedCredential", "ExampleTypeCredential"]</code>
           </dd>
           <dt><var>issuer</var></dt>
--- a/index.html
+++ b/index.html
@@ -2060,32 +2060,21 @@ the <a>verifiable presentation</a> MUST include a <code>holder</code>
           </p>
           <p>
 All of the normative requirements defined for <a>verifiable credentials</a>
-apply to self-asserted <a>verifiable credentials</a>. Some additional
-considerations follow:
+apply to self-asserted <a>verifiable credentials</a>.
           </p>
-
-        <dl>
-          <dt><var>type</var></dt>
-          <dd>
-It is RECOMMENDED that <code>SelfAssertedCredential</code> be included as an
-additional <code>type</code> value. For example,<br>
-<code>"type": ["VerifiableCredential", "SelfAssertedCredential", "ExampleTypeCredential"]</code>
-          </dd>
-          <dt><var>issuer</var></dt>
-          <dd>
+          <p>
 For self-asserted <a>verifiable credentials</a> that are secured using the same
 mechanism as the <a>verifiable presentation</a>, the value of the
 <code>issuer</code> <a>property</a> MUST be identical to the <code>holder</code>
 <a>property</a> of the <a>verifiable presentation</a>.
-          </dd>
-        </dl>
-        <p>
+          </p>
+          <p>
 The example below shows a <a>verifiable presentation</a> that embeds a
 self-asserted <a>verifiable credential</a> that is secured using the same
 mechanism as the <a>verifiable presentation</a>.
-        </p>
+          </p>
 
-        <pre class="example nohighlight" title="A secured verifiable presentation relying on an embedded proof, including a self-asserted verifiable credential">
+          <pre class="example nohighlight" title="A secured verifiable presentation relying on an embedded proof, including a self-asserted verifiable credential">
 {
   "@context": [
     "https://www.w3.org/ns/credentials/v2",
@@ -2104,15 +2093,15 @@ mechanism as the <a>verifiable presentation</a>.
   }],
   "proof": [{ <span class="comment">...</span> }]</span>
 }
-        </pre>
-                <p>
+          </pre>
+          <p>
 The example below shows a <a>verifiable presentation</a> that embeds a
 self-asserted <a>verifiable credential</a> that holds <a>claims</a> about the
 <a>verifiable presentation</a>. It is secured using the same mechanism as the
 <a>verifiable presentation</a>.
-        </p>
+          </p>
 
-        <pre class="example nohighlight" title="Verifiable presentation with a self-asserted verifiable credential about the verifiable presentation">
+          <pre class="example nohighlight" title="Verifiable presentation with a self-asserted verifiable credential about the verifiable presentation">
 {
   "@context": [
     "https://www.w3.org/ns/credentials/v2",
@@ -2133,7 +2122,7 @@ self-asserted <a>verifiable credential</a> that holds <a>claims</a> about the
   }],
   "proof": [{ <span class="comment">...</span> }]</span>
 }
-        </pre>
+          </pre>
         </section>
       </section>
     </section>
--- a/index.html
+++ b/index.html
@@ -2085,7 +2085,7 @@ self-asserted <a>verifiable credential</a> that is secured using the same
 mechanism as the <a>verifiable presentation</a>.
         </p>
 
-        <pre class="example nohighlight" title="Verifiable presentation with a self-asserted verifiable credential">
+        <pre class="example nohighlight" title="A secured verifiable presentation relying on an embedded proof, including a self-asserted verifiable credential">
 {
   "@context": [
     "https://www.w3.org/ns/credentials/v2",
--- a/index.html
+++ b/index.html
@@ -2813,7 +2813,7 @@ the <a>verifier</a> to <code>https://university.example/refresh/3732</code>.
       <section>
         <h3>Terms of Use</h3>
 
-        <p class="issue" title="(AT RISK) Feature depends on demonstration of independent implementations">
+        <p class="issue" data-number="1010" title="(AT RISK) Feature depends on demonstration of independent implementations">
 This feature is at risk and will be removed from the specification if at least
 two independent, interoperable implementations are not demonstrated for a
 single extension type by the end of the Candidate Recommendation Phase.
@@ -2985,7 +2985,7 @@ appropriate section of the Verifiable Credentials Implementation Guidelines
       <section>
         <h3>Evidence</h3>
 
-        <p class="issue" title="(AT RISK) Feature depends on demonstration of independent implementations">
+        <p class="issue" data-number="870" title="(AT RISK) Feature depends on demonstration of independent implementations">
 This feature is at risk and will be removed from the specification if at least
 two independent, interoperable implementations are not demonstrated for a
 single extension type by the end of the Candidate Recommendation Phase. If
--- a/index.html
+++ b/index.html
@@ -2113,11 +2113,11 @@ self-asserted <a>verifiable credential</a> that holds <a>claims</a> about the
   "holder": "did:example:12345678",
   "verifiableCredential": [{
     "@context": "https://www.w3.org/ns/credentials/v2",
-    "type": ["VerifiableCredential", "ExampleTimeLimitCredential"],
+    "type": ["VerifiableCredential", "ExampleAssertCredential"],
     "issuer": "did:example:12345678",
     "credentialSubject": {
       <span class="highlight">"id": "urn:uuid:313801ba-24b7-11ee-be02-ff560265cf9b"</span>,
-      "validUntil": "2024-01-01T19:23:24Z"
+      "assertion": "This VP is submitted by the subject as evidence of a legal right to drive"
     },
     { <span class="comment">...</span> }
   }],
--- a/index.html
+++ b/index.html
@@ -2101,7 +2101,7 @@ self-asserted <a>verifiable credential</a> that holds <a>claims</a> about the
 <a>verifiable presentation</a>.
           </p>
 
-          <pre class="example nohighlight" title="Verifiable presentation with a self-asserted verifiable credential about the verifiable presentation">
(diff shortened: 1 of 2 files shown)


decentralgabe (approved):

David-Chadwick:
Surely `@context` is also mandatory to include in the self-asserted VC

brentzundel:
> Surely `@context` is also mandatory to include in the self-asserted VC

Yes, of course.
I should probably add a line like: "The following are required for self-attested VCs _in addition to the normative requirements in Sections 4 and 5 _

andresuribe87 (commented):
I would favor removing the special casing the `SelfAssertedCredential`. Specifically, removing the text that reads 

> These self-asserted claims are considered to be
> secured by the same mechanism that secures the <a>verifiable presentation</a> in
> which they are included.

and the one that says

> The <code>type</code> <a>property</a> is required. The value MUST be
<code>VerifiableCredential</code> and <code>SelfAssertedCredential</code>.

Removing them would simplify implementations of the VCDM. Curious why this is the approach suggested?

brentzundel:
@andresuribe87 The PR proposes a type for a self-asserted credential so that every implementer can use the same mechanism for indicating that a VC in a VP is self-asserted. I'm open to suggestions for another mechanism that would accomplish the same thing.

On the securing language: I could probably add a line indicating that a self-asserted VC shouldn't have a proof to make it more clear that the proof for the self-asserted VC is the same as the proof for the VP.

andresuribe87:
> I'm open to suggestions for another mechanism that would accomplish the same thing.

I'm arguing for not introducing such mechanism. Verifiers could determine this by comparing the issuer of the proof for the VP vs. the issuer of the proof of a VC. When they match, then the VC can be considered self-asserted. Does that make sense?

TallTed:
> [@brentzundel] On the securing language: I could probably add a line indicating that a self-asserted VC shouldn't have a proof to make it more clear that the proof for the self-asserted VC is the same as the proof for the VP.

Saying that "a self-asserted VC shouldn't have a proof" is saying that a self-asserted VC is not V.

On the other hand, saying that "the proof for [a] self-asserted VC is the same as the proof for the VP [containing that VC]" seems clear, and leaves the V intact.

brentzundel:
> Saying that "a self-asserted VC shouldn't have a proof" is saying that a self-asserted VC is not V.

I was not clear (and have edited the comment for clarity). I meant to say "a self-asserted VC shouldn't have a proof _property_, because the proof for a self-asserted VC is the same as the proof for the VP"

andresuribe87:
> > I'm arguing for not introducing such mechanism. Verifiers could determine this by comparing the issuer of the proof for the VP vs. the issuer of the proof of a VC. When they match, then the VC can be considered self-asserted. Does that make sense?
> 
> Currently, the `holder` property is optional, so it can't be relied on to be present in the verifiable presentation for comparison to the issuer field of the self-asserted VC (which means I need to tweak the language in this PR).
> 
> Maybe a VP that contains a self-asserted VC MUST have a`holder` property?

I think that's better. Yet I still feel that we'd be better off by not adding a the `SelfAssertedCredential` type. Another reason why I would like to not add it is because Verifiers should define the way in which they decide whether a VC is self-asserted. That is very much business logic, and should be up to them to define how it is done.

brentzundel:
> Verifiers should define the way in which they decide whether a VC is self-asserted. That is very much business logic, and should be up to them to define how it is done.

I agree that a verifier should be able to determine what business logic they apply. It is not clear to me what that has to do with a holder being given the ability to make clear that their claims are self-asserted. A verifier remains free to accept additional ways, or to ignore the `SelfAssertedCredential` type.
The whole point of the PR is to provide a single common way that this can be done. That doesn't prohibit the myriad other ways a verifier might prefer to do it.

I believe the interop story is much stronger (and much more testable) with the inclusion of a new type than with any combination of other properties from which a verifier can infer that a VC is self-attested.
If having the new type plus matching the holder/issuer fields is too complex, I'd much rather remove the requirement that the holder/issuer fields match and rely solely on the type, especially considering that some securing methods for VPs won't require the holder property be present in the VP at all.

andresuribe87:
> I believe the interop story is much stronger (and much more testable) with the inclusion of a new type than with any combination of other properties from which a verifier can infer that a VC is self-attested.

I agree that it's much more testable, but there are also drawbacks. Imagine a verifier that relies on multiple services for validation. When the verifier sends over the `SelfAssertedCredential` VC to the validator, then the validator would need the whole VP in order to be able to independently verify the contents of such VC. Such a scenario makes interoperability difficult, as both will need to understand and implement support for the `SelfAssertedCredential` type.

brentzundel:
> When the verifier sends over the `SelfAssertedCredential` VC to the validator, then the validator would need the whole VP in order to be able to independently verify the contents of such VC.

Wouldn't this already be the case, as the self-asserted VC relies on the VP's securing mechanism for verifiablility.

andresuribe87:
> > When the verifier sends over the `SelfAssertedCredential` VC to the validator, then the validator would need the whole VP in order to be able to independently verify the contents of such VC.
> 
> Wouldn't this already be the case, as the self-asserted VC relies on the VP's securing mechanism for verifiablility.

My understanding is that the self-asserted VC is a VC thats secured independently from the VP.

brentzundel:
> My understanding is that the self-asserted VC is a VC thats secured independently from the VP.

I understand that that's what you want this PR to say. That's not what it says and not what I believe reflects the previous conversation on this topic held in Issue #860
(particularly https://github.com/w3c/vc-data-model/issues/860#issuecomment-1592069531).

I believe that this PR is the simplest mechanism for normative inclusion of a self-asserted VC inside of a VP.

As it is clear that you and I do not agree, I ask @Sakurann @jandrieu and @David-Chadwick to indicate whether this PR is acceptable to them so we can get a better sense of where consensus may lie.

Sakurann:
I honestly think both options should be mentioned: a) self-attested claims being secured by a VP and b) self-attested claims included in a holder-signed VC.

For option a), what is the benefit of formatting self-attested claims in an unsigned verifiable credential? it can be pretty confusing. Why not include self-attested claims as top level claims alongside VP's `type` etc?

I think option b) above is beneficial because there is a possibility that self-signed VC is sent by itself without any other VC/VP, or sent alongside other VP(s) (that contain non-Holder signed VC(s)), in which case I am not sure what is the benefit of mandating nesting VC in a VP.

I do not think there are enough implementation experience around holder-signed VCs to choose a) vs b) and I think VCDM is extensible enough to support both without major interop issues.

index.html
```diff
@@ -2030,6 +2030,65 @@ <h4>Presentations Using Derived Credentials</h4>
...
+<a>verifiable presentation</a> MUST include the following properties:
+        </p>
+
+        <dl>
+          <dt><var>type</var></dt>
+          <dd>
+The <code>type</code> <a>property</a> is required. The value MUST be
+<code>VerifiableCredential</code> and <code>SelfAssertedCredential</code>.For
+example,<br>
+<code>"type": ["VerifiableCredential", "SelfAssertedCredential"]</code>
```
msporny, line 2058: I think I agree with @andresuribe87 here -- it /could/ be `SelfAssertedCredential`, or it could be just about any other type of VC. IOW, the thing that makes a VC "self-asserted" is that it's issued by the holder.

So, there is probably some sort of interplay here with `VP.holder` and the issuer for the self-asserted VC.

index.html
```diff
@@ -2030,6 +2030,65 @@ <h4>Presentations Using Derived Credentials</h4>
...
+          </dd>
+          <dt><var>issuer</var></dt>
+          <dd>
+The <code>issuer</code> <a>property</a> is required. The value MUST be identical
+to the <code>holder</code> <a>property</a> of the <a>verifiable presentation</a>.
```
msporny, line 2063: Yes to this.

index.html
```diff
@@ -2030,6 +2030,65 @@ <h4>Presentations Using Derived Credentials</h4>
...
+{
+  "@context": [
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
+  ],
+  "type": ["VerifiablePresentation", "ExamplePresentation"],
+  "holder": "did:example:12345678",
+  "verifiableCredential": [{
+    "@context": "https://www.w3.org/ns/credentials/v2",
+    "type": ["VerifiableCredential", "SelfAssertedCredential"],
```
msporny, line 2081:
```suggestion
    "type": ["VerifiableCredential", "ExampleFoodPreferenceCredential"],
```

I'm unsure about this suggestion. I think it's the required type of `SelfAssertedCredential` that's throwing me off.

Given that we want to create a level playing field (anyone can issue VCs about anything, holders included), it feels like we shouldn't create a distinction between "self-asserted" vs. "officially-asserted". These are all just VCs and you either trust the issuer to make statements, or you don't. One of those issuers that you choose to trust might be the holder, like when you ask them about their favorite cheese... and you need to have them on record that they (the holder) expressed what their favorite cheese was to you... perhaps due to some sort of really important cheese census taking effort (which I expect they take seriously in countries where a significant portion of their commerce is related to cheese).

So, in this case, they'd probably be issuing a VC with a type of `FoodPreferenceCredential`, not `SelfAssertedCredential`.

msporny:
@andresuribe87 wrote:
> I still feel that we'd be better off by not adding a the SelfAssertedCredential type. 

This resonates with me. The VC type could be anything, given that the VC architecture allows anyone to say anything about anything... so best to not establish a VC type for self-asserted claims. I tried to elaborate more about this concern here: https://github.com/w3c/vc-data-model/pull/1186#pullrequestreview-1520770336

@Sakurann wrote:
> I honestly think both options should be mentioned: a) self-attested claims being secured by a VP and b) self-attested claims included in a holder-signed VC.

Yes, agree with @Sakurann. Here are some use cases we might base each mechanism on:

>  a) self-attested claims being secured by a VP

Expressing the terms of use for a particular presentation (such as, "Use only for the purposes of this transaction", or "Do not cache", or "Do not monetize"). The unfortunate truth here is that we don't have any good terms of use expression languages for VCs (beyond ODRL, which doesn't seem to be getting much uptake).

> b) self-attested claims included in a holder-signed VC.

This is what this PR does, so I think we have this use case covered.

To be clear, I'm supportive of this PR and there is a variation of it that is useful and I think can gain consensus. Waiting on feedback from @jandrieu and @David-Chadwick to read their thoughts and then will try to provide some alterations that might align and get us to consensus.

dlongley:
> @andresuribe87 wrote:
> 
> > I still feel that we'd be better off by not adding a the SelfAssertedCredential type.
> 
> This resonates with me. The VC type could be anything, given that the VC architecture allows anyone to say anything about anything... so best to not establish a VC type for self-asserted claims. I tried to elaborate more about this concern here: [#1186 (review)](https://github.com/w3c/vc-data-model/pull/1186#pullrequestreview-1520770336)

+1 ... I agree with the other things said about this as well.

> @Sakurann wrote:
> 
> > I honestly think both options should be mentioned: a) self-attested claims being secured by a VP and b) self-attested claims included in a holder-signed VC.
> > 
> > a) self-attested claims being secured by a VP

I think this is really just a VC without an assertion method proof, i.e., it is only secured using the authentication proof associated with the VP. Whether or not it's "self-attested" doesn't seem like a requirement for putting the components together this way, though it may be the most common use case for doing this sort of thing.

So, it seems we should just be clear that a VP may have a VC in it that doesn't carry its own proof (e.g., no proof clearly marked as signed for the purpose of making an assertion... if using data integrity proofs). Really, it could be more than just one, allowing for the holder to reuse the full breadth of credential vocabularies out there just like any other issuer. Such a VC can't be separated from the VP without losing the ability to verify any cryptography associated with it -- which has advantages and disadvantages.

It's not clear how easy it will be to build general purpose verification services (that are separable from the business logic component in a system) to handle this case -- except in the cases that `holder` matches `issuer` in the VCs that lack such proofs. So, I expect, in practice, those will be the only sorts of VCs that could pass verification by such software -- but I don't think we need to add any language to the data model spec restricting other possibilities.

We should just say that a "self-attested" VC in a VP is one where the VP has a `holder` property and its value matches the `issuer` value in the VC. I think that keeps it simple and probably allows for general purpose verification services to be easily implemented.

>> b) self-attested claims included in a holder-signed VC.

I think this is already covered generally and is just a case where the issuer of a VC happens to also be the party presenting.

OR13 (approved):

index.html
```diff
@@ -2030,6 +2030,65 @@ <h4>Presentations Using Derived Credentials</h4>
...
+          <p>
+In addition to <a>claims</a> from an <a>issuer</a> in the form of
+<a>verifiable credentials</a>, a <a>verifiable presentation</a> MAY include
+<a>claims</a> directly from the <a>holder</a>. These <a>holder</a> <a>claims</a>
+are considered to be self-asserted. If present, these self-asserted
+<a>claims</a> MUST be formatted as a <a>verifiable credential</a> and included
+in the value of the <code>verifiableCredential</code> <a>property</a> of the
+<a>verifiable presentation</a>. These self-asserted claims are considered to be
+secured by the same mechanism that secures the <a>verifiable presentation</a> in
+which they are included.
```
dlongley, line 2044:
```suggestion
<a>verifiable presentation</a>. These self-asserted claims can either use their
own securing mechanism like any other <a>verifiable credential</a> or be
secured by the same mechanism that secures the <a>verifiable presentation</a> in
which they are included.
```

index.html
```diff
@@ -2030,6 +2030,65 @@ <h4>Presentations Using Derived Credentials</h4>
...
+<code>VerifiableCredential</code> and <code>SelfAssertedCredential</code>.For
+example,<br>
+<code>"type": ["VerifiableCredential", "SelfAssertedCredential"]</code>
+          </dd>
+          <dt><var>issuer</var></dt>
+          <dd>
+The <code>issuer</code> <a>property</a> is required. The value MUST be identical
+to the <code>holder</code> <a>property</a> of the <a>verifiable presentation</a>.
+          </dd>
+        </dl>
```
dlongley, line 2065:
```suggestion
A <a>verifiable presentation</a> that includes a self-asserted
<a>verifiable credential</a> that is only secured using the same mechanism
as the <a>verifiable presentation</a> MUST include a <code>holder</code>
<a>property</a>. The value of the <code>holder</code> <a>property</a>
MUST be identical to the value of the <code>issuer</code> <a>property</a>
for every such self-asserted <a>verifiable credential</a>.
        </p>
```

dlongley (changes requested):
I've left some suggestions that match my earlier comments and that seem like they would be inline with what others have suggested.

index.html
brentzundel, line 2058:
```suggestion
The <code>type</code> <a>property</a> is required. The value MUST be
<code>VerifiableCredential</code> and IT IS RECOMMENDED that an additional value
<code>SelfAssertedCredential</code> be included. For example,<br>
<code>"type": ["VerifiableCredential", "SelfAssertedCredential"]</code>
```

index.html
brentzundel, line 2081:
```suggestion
    "type": ["VerifiableCredential", "SelfAssertedCredential", "ExampleFoodPreferenceCredential"],
```

index.html
brentzundel: I did not accept this suggestion because, in addition to clarifying requirements for a self-asserted VC that is secured using the VP mechanism, it removes the `type` requirements. I believe that there is real value in having a unified type for self-asserted credentials, so rather than removing it entirely have changed the requirement from a MUST to RECOMMENDED.

brentzundel:
I have made changes in response to feedback:

- I accepted suggestions and otherwise added language to support both ways of securing self-asserted VCs.
- I changed the `type` requirement from MUST to RECOMMENDED.

I'm not well-pleased with the move away from a single simple mechanism for including a self-asserted VC in a VP to a bifurcated mechanism, it seems that this is an unnecessary complication to me, but it also seems to be where consensus lies.

I did not entirely remove the `type` requirement because I believe there is real value in providing a single normative, interoperable `type` value that indicates a VC is self-asserted.

index.html
```diff
@@ -2030,6 +2030,79 @@ <h4>Presentations Using Derived Credentials</h4>
...
+In addition to <a>claims</a> from an <a>issuer</a> in the form of
+<a>verifiable credentials</a>, a <a>verifiable presentation</a> MAY include
+<a>claims</a> directly from the <a>holder</a>. These <a>holder</a> <a>claims</a>
+are considered to be self-asserted. If present, these self-asserted
+<a>claims</a> MUST be formatted as a <a>verifiable credential</a> and included
+in the value of the <code>verifiableCredential</code> <a>property</a> of the
+<a>verifiable presentation</a>. These self-asserted claims can either use their
+own securing mechanism like any other <a>verifiable credential</a> or be
+secured by the same mechanism that secures the <a>verifiable presentation</a> in
+which they are included.
```
TallTed, line 2045:
```suggestion
<a>verifiable presentation</a>. These self-asserted claims can be secured
by their own mechanism like any other <a>verifiable credential</a> or by
the same mechanism that secures the <a>verifiable presentation</a> in which
they are included.
```

index.html
```diff
@@ -2030,6 +2030,79 @@ <h4>Presentations Using Derived Credentials</h4>
...
+<a>verifiable presentation</a> MUST include the following properties:
+          </p>
+
+        <dl>
+          <dt><var>type</var></dt>
+          <dd>
+The <code>type</code> <a>property</a> is required. The value MUST be
+<code>VerifiableCredential</code> and IT IS RECOMMENDED that the additional
+value <code>SelfAssertedCredential</code> be included. Additional types are also
+acceptable. For example,<br>
```
TallTed, line 2065:
```suggestion
The <code>type</code> <a>property</a> is required. One value MUST be
<code>VerifiableCredential</code>, and it is RECOMMENDED that another
value be <code>SelfAssertedCredential</code>. Additional types MAY also
be included. For example,<br>
```

index.html
```diff
@@ -2030,6 +2030,79 @@ <h4>Presentations Using Derived Credentials</h4>
...
+In addition to <a>claims</a> from an <a>issuer</a> in the form of
+<a>verifiable credentials</a>, a <a>verifiable presentation</a> MAY include
+<a>claims</a> directly from the <a>holder</a>. These <a>holder</a> <a>claims</a>
+are considered to be self-asserted. If present, these self-asserted
+<a>claims</a> MUST be formatted as a <a>verifiable credential</a> and included
+in the value of the <code>verifiableCredential</code> <a>property</a> of the
+<a>verifiable presentation</a>. These self-asserted claims can be secured
+by their own mechanism like any other <a>verifiable credential</a> or by
+the same mechanism that secures the <a>verifiable presentation</a> in which
+they are included.
```
dlongley, line 2045: Hmm, I think the MUST here may not be testable -- or it's suggesting that nothing else can ever be said in a VP without using a VC, which I think would be a mistake. Even adding another custom property that references an encrypted VC or a JWT-secured verifiable credential, for example, would be a "claim" and therefore disallowed.

Here's a tweak to address that and to hopefully make it more clear that the holder can act like any other issuer -- rather than treating them as a special case. I think what's special here is not that the holder is not an issuer (they are), but that because they are the one presenting they may optionally choose to reuse the securing mechanism on the presentation to secure the VCs where they are the issuer. Of course, this is only permitted if the requirements laid out in text following this paragraph are met (`holder` property exists in the VP and has the same value as the `issuer` property in the self-asserted VC).

```suggestion
A <a>holder</a> MAY use the <code>verifiableCredential</code>
<a>property</a> in a <a>verifiable presentation</a> to include
<a>verifiable credentials</a> from any <a>issuer</a>, including
themselves. When the <a>issuer</a> of a <a>verifiable credential</a>
is the <a>holder</a>, its <a>claims</a> are considered to be
self-asserted. These self-asserted claims can be secured by their
own mechanism like any other <a>verifiable credential</a> or by
the same mechanism that secures the <a>verifiable presentation</a>
in which they are included.
```

dlongley (changes requested):

jandrieu:
@dlongley The point of issue #860 is to have a standard way to do it. Hence the request for documenting that standard way.

index.html
brentzundel: I have modified the example to include `FoodPreferenceCredential`, but I am not yet convinced that removing `SelfAssertedCredential` is the best path forward.

index.html
longpd: +1 to this as well.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-07-11-vcwg-special#section1-9) on 2023-07-11

- no resolutions were taken

#### 1.9. Add interoperable way for holder-asserted claims in a VP (pr vc-data-model#1186)

_See github pull request [vc-data-model#1186](https://github.com/w3c/vc-data-model/pull/1186)._




**Brent Zundel:** P.R seeks to address issue #860.  
… provides normative guidance for holder claims in a presentation.  
… a holder can elect to either produce a VC secured on its own and insert into vc array of the presentation.  

_See github issue [vc-data-model#860](https://github.com/w3c/vc-data-model/issues/860)._




**Brent Zundel:** alternatively they can include an unsecured VC in the presented vcs.  
… as long as issuer field of the VC and the holder field of the presentation match. Then the issuer can have confidence that this is self attested claim from the holder.  
… all comments have been responded to. Awaiting final review from TallTed.  
… open to comments.  

> *Phillip Long:* +1 to that PR pertains to issue 860.

**Andres Uribe:** made a comment about enabling forwarding of credentials.  
… not sure how that would be possible using the self asserted type.  
… maybe this isn't a use case that is as important to support.  

**Oliver Terbu:** can see interop issues with this approach in the future.  
… many other ways to achieve the same. Are we saying this is the only approach to achieve this.  
… not sure this is the right approach. Not objecting, just a comment.  

**Dave Longley:** think forwarding use case can be solved by attaching a signature to the VC.  
… using a SelfAsserted type is just a recommendation.  
… think P.R is really around enabling the VP to be the securing for self attested claims.  
… otherwise holder can just act in the role of issuer.  
… this adds a mechanism to secure using proof of the VP itself.  

**Joe Andrieu:** part of this shift has missed some of initial point from issue #860.  
… this was about how you say these claims are about this specific thing right now.  

> *Dave Longley:* perhaps give the VP an ID and include that in your VC claims.

**Joe Andrieu:** not sure how we address that.  

> *Will Abramson:* PL-ASU: this is in relation to the self asserted. The fact that there may be a circumstance where self-asserted claims do not want them to be forwarded.

**Oliver Terbu:** I heard that we have two different issues.  
… JoeAndrieu wanted author of credentials about the VP and the ability to include those in a presentation.  
… then there is the one about self asserted with proof from the VP used to secure.  
… maybe we need separate P.Rs for this.  

**Brent Zundel:** any concrete proposals or changes to this P.R is appreciated. Not sure how to act on the general guidance.  
… responding to JoeAndrieu, should be easy to add a line to the spec that addresses your use case.  

**Joe Andrieu:** think that is pretty good. E.g. if it says if you are making a VP, you could have the ID of a VP that you then refer to.  

**Brent Zundel:** okay I can make that change.  

</details>

brentzundel:
@jandrieu I have added text that I think addresses the additional language you were seeking in (14fb906)[https://github.com/w3c/vc-data-model/pull/1186/commits/14fb906378a53d37f6246a7f412394a5be6dad3c]

Please review.

brentzundel:
@TallTed the PR is still showing that you have changes requested. I believe I have made the modifications you were seeking. Could you please re-review?

index.html
```diff
@@ -2033,15 +2033,24 @@ <h4>Presentations Using Derived Credentials</h4>
...
-the same mechanism that secures the <a>verifiable presentation</a>
-in which they are included.
+A <a>holder</a> MAY use the <code>verifiableCredential</code> <a>property</a> in
+a <a>verifiable presentation</a> to include <a>verifiable credentials</a> from
+any <a>issuer</a>, including themselves. When the <a>issuer</a> of a
+<a>verifiable credential</a> is the <a>holder</a>, its <a>claims</a> are
+considered to be self-asserted. These self-asserted claims can be secured by
+their own mechanism like any other <a>verifiable credential</a> or by the same
+mechanism that secures the <a>verifiable presentation</a> in which they are
+included.
```
TallTed, line 2043:
```suggestion
<a>verifiable credential</a> is the <a>holder</a>, the <a>claims</a> in that
<a>verifiable credential</a> are considered to be <em>self-asserted</em>.
Such self-asserted claims can be secured by the same mechanism that secures
the <a>verifiable presentation</a> in which they are included or by any
mechanism usable for other <a>verifiable credentials</a>.
```

index.html
```diff
@@ -2033,15 +2033,24 @@ <h4>Presentations Using Derived Credentials</h4>
...
+          </p>
+          <p>
+The <a>subject</a> of these self-asserted <a>claims</a> is not limited and can
+include statements about the <a>holder</a>, one of the other included
+<a>verifiable credentials</a>, or even the <a>verifiable presentation</a> in
+which the self-asserted <a>verifiable credential</a> is included. In each of
+these cases the <code>CredentialSubject</code> would have an <code>id</code>
+<a>property</a> that matches the <code>id</code> of the <a>subject</a>. This is
+no different than how it is done in <a>verifiable credentials</a> that are not
+self-asserted.
```
TallTed, line 2053:
```suggestion
The <a>subject(s)</a> of these self-asserted <a>claims</a> are not limited, so
these <a>claims</a> can include statements about the <a>holder</a>, one of the
other included <a>verifiable credentials</a>, or even the <a>verifiable
presentation</a> in which the self-asserted <a>verifiable credential</a> is
included. In each of these cases, the <code>CredentialSubject</code> would
have an <code>id</code> <a>property</a> that matches the <code>id</code> of
the <a>subject</a>, just as it is done in <a>verifiable credentials</a> that
are not self-asserted.
```

index.html
```diff
@@ -2030,6 +2030,87 @@ <h4>Presentations Using Derived Credentials</h4>
...
+          </dd>
+          <dt><var>issuer</var></dt>
+          <dd>
+The <code>issuer</code> <a>property</a> is required. For self-asserted
+<a>verifiable credentials</a> that are secured using the same mechanism as the
+<a>verifiable presentation</a>, the value MUST be identical to the
+<code>holder</code> <a>property</a> of the <a>verifiable presentation</a>.
+Self-asserted <a>verifiable credentials</a> that use their own securing
+mechanism MUST follow the normative guidance for the <code>issuer</code>
+<a>property</a> in Section <a href="#issuer"></a>.
```
TallTed, line 2084: This `MUST` + these `required`s seem redundant, especially when considering VCs in general (i.e., not self-asserted), which already require `type` and `issuer`.... If I'm not confused about this (please confirm!), I'll draft a rephrasing of lines 2062-2084.

index.html
dlongley: On top of @TallTed's edits, I'd think we'd just want to say that the subjects (little 's') in the VCs would have an `id` property instead of saying it only applies to the top-level credential subjects. Any subject will do in any claim made throughout the VC.

index.html
brentzundel: I would much appreciate a suggested rephrasing. The clarification was added in response to previous comments, and I am open to other possible ways of wording this.

index.html
```diff
@@ -2036,21 +2036,21 @@ <h4>Presentations Including Holder Claims</h4>
...
           <p>
-The <a>subject</a> of these self-asserted <a>claims</a> is not limited and can
-include statements about the <a>holder</a>, one of the other included
-<a>verifiable credentials</a>, or even the <a>verifiable presentation</a> in
-which the self-asserted <a>verifiable credential</a> is included. In each of
-these cases the <code>CredentialSubject</code> would have an <code>id</code>
-<a>property</a> that matches the <code>id</code> of the <a>subject</a>. This is
-no different than how it is done in <a>verifiable credentials</a> that are not
-self-asserted.
+The <a>subject(s)</a> of these self-asserted <a>claims</a> are not limited, so
```
OR13, line 2046: this is complicated enough that it really ought to have an example.

brentzundel: added an example in [bef11ba](https://github.com/w3c/vc-data-model/pull/1186/commits/bef11bab1f6d8dc111f10d669b529b99a50451ea)

index.html
brentzundel: @TallTed in a fit of productivity, I have rephrased and attempted to eliminate the redundancies in this section. [bef11ba](https://github.com/w3c/vc-data-model/pull/1186/commits/bef11bab1f6d8dc111f10d669b529b99a50451ea)

index.html
OR13: imo this can be resolved.

index.html
```diff
@@ -2030,6 +2030,111 @@ <h4>Presentations Using Derived Credentials</h4>
...
+  "verifiableCredential": [{
+    "@context": "https://www.w3.org/ns/credentials/v2",
+    "type": ["VerifiableCredential", "SelfAssertedCredential", "ExampleFoodPreferenceCredential"],
+    "issuer": "did:example:12345678",
+    "credentialSubject": {
+      "favoriteCheese": "Gouda"
+    },
+    { <span class="comment">...</span> }
+  }],
+  "proof": [{ <span class="comment">...</span> }]</span>
```
OR13, line 2105: technically optional, perhaps update the title for 

"A secured Verifiable presentation relying on an embedded proof, with a self-asserted verifiable credential"

index.html
```diff
@@ -2030,6 +2030,111 @@ <h4>Presentations Using Derived Credentials</h4>
...
+<a>property</a> of the <a>verifiable presentation</a>.
+          </dd>
+        </dl>
+        <p>
+The example below shows a <a>verifiable presentation</a> that embeds a
+self-asserted <a>verifiable credential</a> that is secured using the same
+mechanism as the <a>verifiable presentation</a>.
+        </p>
+
+        <pre class="example nohighlight" title="Verifiable presentation with a self-asserted verifiable credential">
```
msporny, line 2088:
```suggestion
        <pre class="example nohighlight" title="A secured verifiable presentation relying on an embedded proof, including a self-asserted verifiable credential">
```

index.html
msporny: Done in 8eca073f12e5bd6a81f1341d18fcf3f2ab6f3497.

msporny:
Just trying to clarify before I merge... 

@jandrieu would you object if this PR was merged? 

@TallTed I believe your concerns were addressed, can you confirm that they were?

@dlongley some of your changes were not taken, are you ok with merging this PR as-is?

index.html
```diff
@@ -2030,6 +2030,111 @@ <h4>Presentations Using Derived Credentials</h4>
...
+}
+        </pre>
+                <p>
+The example below shows a <a>verifiable presentation</a> that embeds a
+self-asserted <a>verifiable credential</a> that holds <a>claims</a> about the
+<a>verifiable presentation</a>. It is secured using the same mechanism as the
+<a>verifiable presentation</a>.
+        </p>
+
+        <pre class="example nohighlight" title="Verifiable presentation with a self-asserted verifiable credential about the verifiable presentation">
```
OR13, line 2115:
```suggestion
        <pre class="example nohighlight" title="A secured verifiable presentation, relying on an embedded proof including a self-asserted verifiable credential about the verifiable presentation">
```

OR13: for consistency with https://github.com/w3c/vc-data-model/pull/1186/files#r1264100438

OR13: @msporny as an aside, I suspect your PR to fix `credential` vs `verifiable credential` will be some version of these changes... over and over again... until you go mad.... : )

Basically replacing `credential` with `unsecured verifiable credential` or `secured verifiable credential` depending on the context... looking forward to that one : )

Maybe we can be clever and do that once in terms, and not have to be so verbose everywhere....

selfissued (changes requested):
It's not clear to me that `SelfAssertedCredential` needs to exist.  The difference is only who controls the key - not in the data structure itself.

index.html
msporny: > until you go mad....

That happened long ago... during v1.0...

I'll see if we can be clever wrt. terminology, as you mention above... but at this point, I'm leaning toward NOT making the change (as it sounds like it's going to be a lot of trouble, and result in a confusing outcome for readers).

OR13: I don't see how we can avoid making fixing it, given what it currently normatively states:

- https://w3c.github.io/vc-data-model/#securing-verifiable-credentials

It makes the interpretation of `<a>verifiable credential<a>` and `VerifiableCredential` inconsistent.

The reader learns that embedded or external proof is required for `<a>verifiable credential<a>`.... and optional for `VerifiableCredential`... Seems like we will have to fix that... The same applies to presentations... This PR is digging the whole deeper.... but sometimes a tunnel is necessary to escape.

index.html
```diff
@@ -2030,6 +2030,111 @@ <h4>Presentations Using Derived Credentials</h4>
...
+          </p>
+          <p>
+The <a>subject(s)</a> of these self-asserted <a>claims</a> are not limited, so
+these <a>claims</a> can include statements about the <a>holder</a>, one of the
+other included <a>verifiable credentials</a>, or even the <a>verifiable
+presentation</a> in which the self-asserted <a>verifiable credential</a> is
+included. In each of these cases, the <code>CredentialSubject</code> would
+have an <code>id</code> <a>property</a> that matches the <code>id</code> of
+the <a>subject</a>, just as it is done in <a>verifiable credentials</a> that
+are not self-asserted.
```
dlongley, line 2053: To ensure we're covering all subjects (not just top-level credential subjects):

```suggestion
included. In each of these cases, the <code>id</code> <a>property</a>
is used to identify the specific <a>subject</a> in the object where the
<a>claims</a> about it are made, just as it is done in
<a>verifiable credentials</a> that are not self-asserted.
```

dlongley (commented):
I've added a specific suggestion based on the comment I made earlier.

index.html
```diff
@@ -2030,6 +2030,111 @@ <h4>Presentations Using Derived Credentials</h4>
...
+Such self-asserted claims can be secured by the same mechanism that secures
+the <a>verifiable presentation</a> in which they are included or by any
+mechanism usable for other <a>verifiable credentials</a>.
+          </p>
+          <p>
+The <a>subject(s)</a> of these self-asserted <a>claims</a> are not limited, so
+these <a>claims</a> can include statements about the <a>holder</a>, one of the
+other included <a>verifiable credentials</a>, or even the <a>verifiable
+presentation</a> in which the self-asserted <a>verifiable credential</a> is
+included. In each of these cases, the <code>CredentialSubject</code> would
```
msporny, line 2050:
```suggestion
included. In each of these cases, the <code>credentialSubject</code> value would
```

index.html
```diff
@@ -2030,6 +2030,111 @@ <h4>Presentations Using Derived Credentials</h4>
...
+          <p>
+All of the normative requirements defined for <a>verifiable credentials</a>
+apply to self-asserted <a>verifiable credentials</a>. Some additional
+considerations follow:
+          </p>
+
+        <dl>
+          <dt><var>type</var></dt>
+          <dd>
+It is RECOMMENDED that <code>SelfAssertedCredential</code> be included as an
```
msporny, line 2070: I agree with @selfissued and @dlongley that this is not necessary. You can just check to see if the key used to sign a VC belongs to the holder. For Data Integrity, you check to see if the `proof.verificationMethod` is associated with `holder` using the algorithm defined in https://w3c.github.io/vc-data-integrity/#retrieve-verification-method. I expect VC COSE JOSE would define the same sort of mapping. IOW, just because someone asserts this, doesn't mean it's true, you need to verify it... and if only a subset of the ecosystem is doing it, you can't really count on it as a dependable mechansim.

This feedback is non-blocking, just thinking out loud.

msporny (approved):

brentzundel:
I still believe that having a `type` would be useful, but have removed it as it does not seem to be possible to merge this PR otherwise.

index.html
OR13: > I expect VC COSE JOSE would define the same sort of mapping. IOW, just because someone asserts this, doesn't mean it's true, you need to verify it... and if only a subset of the ecosystem is doing it, you can't really count on it as a dependable mechanism.

This describes "Verifiable Presentations" as a whole... And I agree... its not very helpful.

We seem to have to do a lot of work for "holder" and "VerifiablePresentation"... when they are really just special cases of "issuer" and "VerifiableCredential".... 

I continue to assert there is a better solution to this, but I don't think we can turn this boat at this point.

index.html
```diff
@@ -2030,6 +2030,100 @@ <h4>Presentations Using Derived Credentials</h4>
...
+  "type": ["VerifiablePresentation", "ExamplePresentation"],
+  <span class="highlight">"id": "http://example.com/presentation/5223"</span>,
+  "holder": "did:example:12345678",
+  "verifiableCredential": [{
+    "@context": "https://www.w3.org/ns/credentials/v2",
+    "type": ["VerifiableCredential", "ExampleTimeLimitCredential"],
+    "issuer": "did:example:12345678",
+    "credentialSubject": {
+      <span class="highlight">"id": "http://example.com/presentation/5223"</span>,
+      "validUntil": "2024-01-01T19:23:24Z"
```
msporny, line 2119:
```suggestion
      "validUntil": "2024-01-01T19:23:24Z"
```

Huh... this is valid, and begs the question... why not just put it on the presentation itself? We might want to have an example of someone putting a self-asserted claim on the VP itself... possibly in another PR, we don't want to complicate this one further.

IOW, do we want to enable `validFrom` and `validUntil` onto the Presentation too? I think we do want to enable that. At present, it's only allowed on a `VerifiableCredential`. Historically, this was because presentations were expected to always be short lived... but the short-lived-ness being determined by the verifier and the `created/iat` date on the proof.

index.html
```diff
@@ -2030,6 +2030,100 @@ <h4>Presentations Using Derived Credentials</h4>
...
+          </p>
+
+          <pre class="example nohighlight" title="A secured verifiable presentation, relying on an embedded proof, with a self-asserted verifiable credential about the verifiable presentation">
+{
+  "@context": [
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
+  ],
+  "type": ["VerifiablePresentation", "ExamplePresentation"],
+  <span class="highlight">"id": "http://example.com/presentation/5223"</span>,
```
msporny, line 2111:
```suggestion
  <span class="highlight">"id": "urn:uuid:313801ba-24b7-11ee-be02-ff560265cf9b"</span>,
```

index.html
```diff
@@ -2030,6 +2030,100 @@ <h4>Presentations Using Derived Credentials</h4>
...
+  "holder": "did:example:12345678",
+  "verifiableCredential": [{
+    "@context": "https://www.w3.org/ns/credentials/v2",
+    "type": ["VerifiableCredential", "ExampleTimeLimitCredential"],
+    "issuer": "did:example:12345678",
+    "credentialSubject": {
+      <span class="highlight">"id": "http://example.com/presentation/5223"</span>,
```
msporny, line 2118:
```suggestion
      <span class="highlight">"id": "urn:uuid:313801ba-24b7-11ee-be02-ff560265cf9b"</span>,
```

index.html
```diff
@@ -2030,6 +2030,100 @@ <h4>Presentations Using Derived Credentials</h4>
...
+{
+  "@context": [
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
+  ],
+  "type": ["VerifiablePresentation", "ExamplePresentation"],
+  "holder": "did:example:12345678",
+  "verifiableCredential": [{
+    "@context": "https://www.w3.org/ns/credentials/v2",
+    "type": ["VerifiableCredential", "SelfAssertedCredential", "ExampleFoodPreferenceCredential"],
```
msporny, line 2087:
```suggestion
    "type": ["VerifiableCredential", "ExampleFoodPreferenceCredential"],
```

I /think/ you intended to remove this in your response to @selfissued?

index.html
msporny: Agreed, we don't want to derail this PR with trying to find the perfect example. I defer to @jandrieu to suggest a use case. I'll try to think of a use case (with markup) as well.

andresuribe87 (approved):

index.html
```diff
@@ -2059,29 +2059,24 @@ <h4>Presentations Including Holder Claims</h4>
...
           <dd>
-The <code>issuer</code> <a>property</a> is required. For self-asserted
-<a>verifiable credentials</a> that are secured using the same mechanism as the
-<a>verifiable presentation</a>, the value MUST be identical to the
-<code>holder</code> <a>property</a> of the <a>verifiable presentation</a>.
-Self-asserted <a>verifiable credentials</a> that use their own securing
-mechanism MUST follow the normative guidance for the <code>issuer</code>
-<a>property</a> in Section <a href="#issuer"></a>.
+For self-asserted <a>verifiable credentials</a> that are secured using the same
+mechanism as the <a>verifiable presentation</a>, the value of the
```
TallTed, line 2067:
```suggestion
When a self-asserted <a>verifiable credentials</a> is secured using the same
mechanism as the <a>verifiable presentation</a>, the value of the
```

index.html
```diff
@@ -2059,29 +2059,24 @@ <h4>Presentations Including Holder Claims</h4>
...
+<code>issuer</code> <a>property</a> MUST be identical to the <code>holder</code>
+<a>property</a> of the <a>verifiable presentation</a>.
```
TallTed, line 2069:
```suggestion
When a self-asserted <a>verifiable credential</a> is secured using the same
mechanism as the <a>verifiable presentation</a>, the value of the
<code>issuer</code> <a>property</a> of the <a>verifiable credential</a>
MUST be identical to the <code>holder</code> <a>property</a> of the
<a>verifiable presentation</a>.
```

index.html
```diff
@@ -2101,7 +2101,7 @@ <h4>Presentations Including Holder Claims</h4>
 <a>verifiable presentation</a>.
           </p>
 
-          <pre class="example nohighlight" title="Verifiable presentation with a self-asserted verifiable credential about the verifiable presentation">
+          <pre class="example nohighlight" title="A secured verifiable presentation, relying on an embedded proof, with a self-asserted verifiable credential about the verifiable presentation">
```
TallTed, line 2104:
```suggestion
          <pre class="example nohighlight" title="A verifiable presentation, secured with an embedded proof, with a self-asserted verifiable credential about the verifiable presentation">
```

index.html
```diff
@@ -2030,6 +2030,100 @@ <h4>Presentations Using Derived Credentials</h4>
...
+Such self-asserted claims can be secured by the same mechanism that secures
+the <a>verifiable presentation</a> in which they are included or by any
+mechanism usable for other <a>verifiable credentials</a>.
+          </p>
+          <p>
+The <a>subject(s)</a> of these self-asserted <a>claims</a> are not limited, so
+these <a>claims</a> can include statements about the <a>holder</a>, one of the
+other included <a>verifiable credentials</a>, or even the <a>verifiable
+presentation</a> in which the self-asserted <a>verifiable credential</a> is
+included. In each of these cases, the <code>id</code> <a>property</a>
```
TallTed, line 2050:
```suggestion
included. In each case, the <code>id</code> <a>property</a>
```

index.html
```diff
@@ -2030,6 +2030,100 @@ <h4>Presentations Using Derived Credentials</h4>
...
+<code>issuer</code> <a>property</a> MUST be identical to the <code>holder</code>
+<a>property</a> of the <a>verifiable presentation</a>.
+          </p>
+          <p>
+The example below shows a <a>verifiable presentation</a> that embeds a
+self-asserted <a>verifiable credential</a> that is secured using the same
+mechanism as the <a>verifiable presentation</a>.
+          </p>
+
+          <pre class="example nohighlight" title="A secured verifiable presentation relying on an embedded proof, including a self-asserted verifiable credential">
```
TallTed, line 2077:
```suggestion
          <pre class="example nohighlight" title="A verifiable presentation, secured with an embedded proof, with a self-asserted verifiable credential">
```

index.html
```diff
@@ -2030,6 +2030,101 @@ <h4>Presentations Using Derived Credentials</h4>
...
+          <p>
+All of the normative requirements defined for <a>verifiable credentials</a>
+apply to self-asserted <a>verifiable credentials</a>.
+          </p>
+          <p>
+When a self-asserted <a>verifiable credential</a> is secured using the same
+mechanism as the <a>verifiable presentation</a>, the value of the
+<code>issuer</code> <a>property</a> of the <a>verifiable credential</a>
+MUST be identical to the <code>holder</code> <a>property</a> of the
+<a>verifiable presentation</a>.
```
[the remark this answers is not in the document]
OR13: Seems like we can perhaps remove that text from validation, if this section repeats it.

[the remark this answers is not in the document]
brentzundel: I think that the language already present is simpler and is sufficient.
The holder is the entity constructing both the VP and the self-asserted VC, so they can make sure those values are identical, whether they choose to use an object with an `id` or just a URL.

Plus, we already have this line above, which was introduced specifically to avoid pointing back to each normative section of the spec that also applies: "All of the normative requirements defined for verifiable credentials apply to self-asserted verifiable credentials."

I would prefer to not make this change.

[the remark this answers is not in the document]
dlongley: I'm not going to block this PR if the change isn't made either.

brentzundel:
@selfissued and @TallTed I believe I have made the changes you've requested. Could you please re-review?

index.html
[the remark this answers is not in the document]
OR13: as an aside, presentations and credentials don't require `id`, but when one is present, the claims are bound to the `id`.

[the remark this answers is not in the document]
OR13: I'm this suggestion should be applied, we can clean up when the dust settles.

index.html
brentzundel: I've updated the example based on Joe's comment here: https://github.com/w3c/vc-data-model/pull/1186/files#r1268305835

David-Chadwick:
"As it is clear that you and I do not agree, I ask @Sakurann @jandrieu and @David-Chadwick to indicate whether this PR is acceptable to them so we can get a better sense of where consensus may lie."
This is not the approach I would have taken. As I said earlier, a simpler approach is for the holder to put the assertions in the VP itself, rather than one level down in an embedded VC. Then it is obvious that the holder is making these assertions about this VP. But given that the majority of folk seem to prefer a self signed VC embedded in the VP, then I will not strongly object. I just think it is technically a worse solution for the reasons I stated earlier, and for some raised here by others

OR13:
+1 to this:

> a simpler approach is for the holder to put the assertions in the VP itself, rather than one level down in an embedded VC.

However the reason people seem not to be doing this is that issuer and holder are treated differently than credentialSubject.

Despite all being RDF graph nodes... This seems to be a convention not a normative requirement.

In a different universe we could have had this:

```
{
type: VC,
issuer: Alice,
credentials: [{
type: VC
issuer: Bob,
subject: { id: Charlie }
}]
}
```

No need for VP or holder.

Because they are both just special cases of VC and issuer.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-07-19-vcwg#section1-2) on 2023-07-19

- no resolutions were taken

#### 1.2. Add interoperable way for holder-asserted claims in a VP (pr vc-data-model#1186)

_See github pull request [vc-data-model#1186](https://github.com/w3c/vc-data-model/pull/1186)._




**Manu Sporny:** 1186 is about 'Holder Asserted Claims'. Brent has been processing feedback about it; JoeAndrieu, TallTed are you OK with this PR?  

**Joe Andrieu:** The language doesn't address what I was referring to. I have some language which would fit, but the PR is rather tangential now.  

**Brent Zundel:** I think I addressed it in a part of the PR you were not tagged for.  

> *Ted Thibodeau Jr.:* I'll re-review 1186 and 1149 (and any others currently awaiting my input, if I can have a list of them) after this call.

</details>

TallTed (approved):
Go with it

selfissued (approved):

OR13 (approved):

msporny:

jandrieu (commented):
This looks good. My last comments are effectively editorial. I think we're trying to say something correct, but saying it badly.

Issue #1134, opened 2023-05-18
Should we bundle contexts for credentialSchema and credentialStatus int he v2 core context?


OR13:
The argument for doing so is that they are used by data integrity proofs, and this would make things easier for developers.

If we don't do this, we should make sure to include normative guidance on adding context values to the context array when these features are used, to avoid confusion regarding term definitions.

iherman:
(Meta comment.) 

This issue seems to refer to the same underlying question than, for example, issues #1091 or #1090 (I did not do an exhaustive search). That question can be formulated as:

> Should the V2 Core context include ***only*** terms that are strictly defined in the VCDM specification, or should it also include references to other terms, constructions, etc.?

To make it clear, the current context is based on the answer "no". If the answer is "yes", then the v2 context should be significantly pruned, removing the terms listed in the issue title, the terms that are currently defined in other specs or vocabulary, like DI, ODRL, or schema.org specs, and would also lead to the removal of the `@vocab` statement.

Note that the question is also closely related to #1103.

I would propose to close this issue and open an issue with that question instead. That would also makes it easy to handle #1091 and #1090 (and possibly others).

OR13:
Noting the current vc json schema does not require context bundling, see https://w3c.github.io/vc-json-schema/

OR13:
@decentralgabe you are volunteered for this.

decentralgabe:
Does bundling context meaning adding term definitions for `JsonSchema2023` and `CredentialStatusList2021` etc in the core context? If so I'm willing to make that change.

OR13:
@decentralgabe yes, assume W3C will define the URLs for them, same as it did for "VerifiableCredential".

PR #1194, opened 2023-07-11
Add vocab terms for status list 2021

decentralgabe:
fixes the second part of #1134

### DIFF
--- a/vocab/credentials/v2/vocabulary.yml
+++ b/vocab/credentials/v2/vocabulary.yml
@@ -52,11 +52,11 @@ class:
 
   - id: StatusList2021Entry
     label: Status List 2021 Entry
-    comment: A Status List 2021 Entry provides issuers a mechanism to provide status information for verifiable credentials.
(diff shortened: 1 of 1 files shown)


decentralgabe:
already in the context, but not the vocab https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v2#L117

brentzundel (approved):

msporny (approved):
Approving, noting that due to the change suggested by @mprorock that changes the index calculation, that we might need to rename StatusList2021 to StatusList2023. I know that Troy from Gen Digital had concerns about any breaking changes to StatusList2021 and requested that we not use the same type identifier there.

That, however, can be dealt with in a future PR, during CR if necessary (because of the at risk marker we have).

iherman (approved):

TallTed (changes requested):
Editorial, trivial

brentzundel (approved):

dlongley (approved):

decentralgabe:
Over a week, multiple approvals, changes addressed. Merging.

PR #1195, opened 2023-07-12
Remove outdated v1.0 JWT claim registrations.

msporny:
This PR removes the outdated v1.0 IANA registrations for "vc" and "vp" (since that happened a while ago and they now exist in the JWT Claims registry here):

https://www.iana.org/assignments/jwt/jwt.xhtml

We don't need to keep these registration requests around, as they've been processed by IANA, and the request exists here (in perpetuity): https://www.w3.org/TR/vc-data-model/#iana-considerations

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1195.html" title="Last updated on Jul 12, 2023, 2:06 AM UTC (64fa61f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1195/eb58625...64fa61f.html" title="Last updated on Jul 12, 2023, 2:06 AM UTC (64fa61f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5300,64 +5300,11 @@ and <a>verifiable presentation</a>.
 
     <section class="appendix informative">
       <h2>IANA Considerations</h2>
-      <section id="json-web-token-claim-registrations">
-        <h3>JSON Web Token Claim Registrations</h3>
-        <p>
-          This section will be submitted to the Internet Engineering Steering Group
-          (IESG) for review, approval, and registration with IANA in the
-          "JSON Web Token Claims Registry".
-        </p>
 
-          <section id="vc-json-web-token-claim">
-            <h4>`vc`</h4>
-            <table>
-              <tr>
-                <th>Claim Name: </th>
-                <td>"vc" </td>
-              </tr>
-              <tr>
-                <th>Claim Description: </th>
-                <td>Verifiable Credential</td>
-              </tr>
-              <tr>
-                <th>Change Controller:  </th>
-                <td>W3C</td>
-              </tr>
-              <tr>
-                <th>Specification Document(s): </th>
-                <td>
-                <a href="https://www.w3.org/TR/vc-data-model/">Section 6.3.1.2: JSON Web Token
-                  Extensions of Verifiable Credentials Data Model 1.0</a>
-                </td>
-              </tr>
-            </table>
-          </section>
-
-          <section id="vp-json-web-token-claim">
-            <h4>`vp`</h4>
-            <table>
-              <tr>
-                <th>Claim Name: </th>
-                <td>"vp"</td>
-              </tr>
-              <tr>
-                <th>Claim Description: </th>
-                <td>Verifiable Presentation</td>
-              </tr>
-              <tr>
-                <th>Change Controller: </th>
-                <td>W3C</td>
-              </tr>
-              <tr>
-                <th>Specification Document(s):</th>
-                <td>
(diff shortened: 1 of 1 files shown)


iherman (commented):
I am not 100% sure. 

Regardless of this PR there seems to be already a slight issue with the current JWT registration: it refers to a dated URL for the vc Rec. Which means that if someone clicks on that URL, he/she is directed to an outdated version of the spec, which is clearly indicated by the ugly red thingy popping up in the face. It is not a bug, but it looks ugly, to say the least. I do not know the IANA/JWT procedures, but it would look better if that was amended.

However... even if that change is done, wouldn't we get into the same situation, eventually, if this PR is accepted? Wouldn't V1 become outdated and, therefore, create similar problems again?

The introductory text in the spec should clearly be changed (it is not a "will be submitted" any more),  but it may be cleaner to keep the text in our spec and try to get the IANA to change the reference when the time comes... 

This is not a strong objection, so if others are o.k. with the PR, I won't object. But it may be food for thought.

msporny:
> it may be food for thought.

I wouldn't object to keeping it in there, but then the IANA registration will have to be updated (which is fine, there is a process for doing that to the JWT claims registry). 

The problem w/ keeping that in the vc-data-model is that this spec no longer defines JWT usage, it's the VC-JWT spec that does that, and that spec has deprecated the usage of the "vc" and "vp" properties. 

Other options include:

1. Deprecate usage of "vc" and "vp" properties in JWT claims registry (either request their removal, or deprecation).
2. Move the IANA registration of the "vc" and "vp" properties to the VC-JWT specification and update the JWT claims registry to point there, noting that the properties are deprecated.
3. Keep the registration of the "vc" and "vp" properties in this specification and update the JWT claims registry to point here, noting that the properties are deprecated, or covered by the VC-JWT specification.

Maybe option 2 is the best option?

OR13 (approved):

OR13:
@msporny thanks for raising this PR, It think it is the correct approach.

I think 2 is probably the way to go.

OR13:
See https://github.com/w3c/vc-jwt/issues/119

OR13:
Here is a companion PR, that moves them to vc-jwt, and adds a note about asking the IESG to fix the registry:

https://github.com/w3c/vc-jwt/pull/124

iherman:
> Other options include:
> 
> 1. Deprecate usage of "vc" and "vp" properties in JWT claims registry (either request their removal, or deprecation).
> 2. Move the IANA registration of the "vc" and "vp" properties to the VC-JWT specification and update the JWT claims registry to point there, noting that the properties are deprecated.
> 3. Keep the registration of the "vc" and "vp" properties in this specification and update the JWT claims registry to point here, noting that the properties are deprecated, or covered by the VC-JWT specification.

Option (2) is probably the cleanest, provided that the JWT registry can be changed easily (I am not familiar with its process)

decentralgabe (approved):

brentzundel (approved):

awoie (approved):

dlongley (approved):

iherman (approved):

msporny:
Substantive (but already processed by IANA), multiple reviews, PR opened on VC-JWT to adopt text until JWT registry is fixed https://github.com/w3c/vc-jose-cose/pull/124/files, no changes requested, no objections, merging.

PR #1172, opened 2023-06-27
Add "author" and "party" to terminology, rewrite "claim" terminology

RieksJ:
Signed-off-by: Rieks <RieksJ@users.noreply.github.com>

PR #1198, opened 2023-07-12
Add language around multiple schemas; update example

### LINKED ISSUE #950: Verification of multiple credential schemas

decentralgabe:
Fix #950
- Updated reference name to include the short name (`vc-json-schema`)
- Revised language to note processing of multiple schemas should be handled according to each specific type's rules
- Updated example to show multiple schemas (there are now examples with one and more than one schema)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1198.html" title="Last updated on Jul 17, 2023, 6:25 PM UTC (a40c3ce)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1198/3a0e423...a40c3ce.html" title="Last updated on Jul 17, 2023, 6:25 PM UTC (a40c3ce)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2489,7 +2489,7 @@ integrity protection mechanism. The <code>credentialSchema</code>
     "type": "VerifiableCredentialSchema2023"
   },
   {
-    "id": "https://example.org/examples/name.json",
+    "id": "https://example.org/examples/person.json",
     "type": "JsonSchema2023"
   }]</span>
 }
--- a/index.html
+++ b/index.html
@@ -2445,7 +2445,7 @@ that MUST be a <a>URL</a> identifying the schema file. The precise contents of
 each data schema is determined by the specific type definition.
             </p>
             <p>
-If multiple schemas are present, validity MUST be determined according to the
+If multiple schemas are present, validity is determined according to the
 processing rules outlined by each associated <code>credentialSchema</code> 
 <code>type</code> property.
             </p>
--- a/index.html
+++ b/index.html
@@ -2435,13 +2435,20 @@ the <a>verifiable credentials</a> that it issues:
         <dl>
           <dt><var>credentialSchema</var></dt>
           <dd>
-The value of the <code>credentialSchema</code> <a>property</a> MUST be one or
+            <p>
+The value of the <code>credentialSchema</code> <a>property</a> SHOULD be one or
 more data schemas that provide <a>verifiers</a> with enough information to
-determine if the provided data conforms to the provided schema. Each
+determine if the provided data conforms to the provided schema(s). Each
 <code>credentialSchema</code> MUST specify its <code>type</code> (for example,
-<code>CredentialSchema2022</code>), and an <code>id</code> <a>property</a>
+<code>VerifiableCredentialSchema2023</code>), and an <code>id</code> <a>property</a>
 that MUST be a <a>URL</a> identifying the schema file. The precise contents of
 each data schema is determined by the specific type definition.
+            </p>
+            <p>
+If multiple schemas are present, validity MUST be determined according to the
+processing rules outlined by each associated <code>credentialSchema</code> 
+<code>type</code> property.
+            </p>
           </dd>
         </dl>
 
@@ -2453,7 +2460,7 @@ vocabulary using <code>credentialSchema</code> that is locked to some content
 integrity protection mechanism. The <code>credentialSchema</code>
 <a>property</a> also makes it possible to perform syntactic checking on the
 <a>credential</a> and to use <a>verification</a> mechanisms such as JSON Schema
-[[?CREDENTIAL-SCHEMA-2023]] validation.
+[[?VC-JSON-SCHEMA-2023]] validation.
         </p>
 
         <pre class="example nohighlight"
@@ -2472,24 +2479,32 @@ integrity protection mechanism. The <code>credentialSchema</code>
     "degree": {
       "type": "ExampleBachelorDegree",
       "name": "Bachelor of Science and Arts"
+    },
+    "university": {
+      "type": "ExampleUniversity",
+      "name": "University of Examples, W3C"
     }
   },
-  <span class="highlight">"credentialSchema": {
+  <span class="highlight">"credentialSchema": [{
     "id": "https://example.org/examples/degree.json",
-    "type": "CredentialSchema2023"
-  }</span>
+    "type": "VerifiableCredentialSchema2023"
+  },
+  {
+    "id": "https://example.org/examples/university.json",
+    "type": "JsonSchema2023"
+  }]</span>
 }
         </pre>
 
         <p>
 In the example above, the <a>issuer</a> is specifying a
-<code>credentialSchema</code>, which points to a [[?CREDENTIAL-SCHEMA-2023]] file that
+<code>credentialSchema</code>, which points to a [[?VC-JSON-SCHEMA-2023]] file that
 can be used by a <a>verifier</a> to determine if the
 <a>verifiable credential</a> is well formed.
         </p>
 
         <p class="note" >
-For information about linkages to JSON Schema [[?CREDENTIAL-SCHEMA-2023]] or other
+For information about linkages to JSON Schema [[?VC-JSON-SCHEMA-2023]] or other
 optional <a>verification</a> mechanisms, see the Verifiable Credentials
 Implementation Guidelines [[VC-IMP-GUIDE]] document.
         </p>
@@ -2532,6 +2547,7 @@ In the example above, the <a>issuer</a> is specifying a
 data into a format which can then be used by a <a>verifier</a> to determine if
 the proof provided with the <a>verifiable credential</a> is valid.
         </p>
+
 
       </section>
 
@@ -4960,7 +4976,7 @@ If the <code>credentialSchema</code> property is available, the schema of a
 according to the <code>credentialSchema</code> <a>type</a> definition for the
 <a>verifiable credential</a> and the <a>verifier's</a> own schema evaluation
 criteria. For example, if the <code>credentialSchema</code>'s <code>type</code> value is
-[[?CREDENTIAL-SCHEMA-2023]], then a <a>verifier</a> can ensure a credential's
+[[?VC-JSON-SCHEMA-2023]], then a <a>verifier</a> can ensure a credential's
 data is valid against the given JSON Schema.
         </p>
       </section>
@@ -5168,7 +5184,7 @@ cryptographic hash.
         <p>
 The <a>verifiable credential</a> and <a>verifiable presentation</a> data models
 leverage a variety of underlying technologies including [[JSON-LD]] and
-[[?CREDENTIAL-SCHEMA-2023]]. This section will provide a comparison of the
+[[?VC-JSON-SCHEMA-2023]]. This section will provide a comparison of the
 <code>@context</code>, <code>type</code>, and <code>credentialSchema</code>
 properties, and cover some of the more specific use cases where it is possible
 to use these features of the data model.
--- a/index.html
+++ b/index.html
@@ -2471,7 +2471,7 @@ integrity protection mechanism. The <code>credentialSchema</code>
     "https://www.w3.org/ns/credentials/examples/v2"
   ],
   "id": "http://university.example/credentials/3732",
-  "type": ["VerifiableCredential", "ExampleDegreeCredential", "ExampleNameCredential"],
+  "type": ["VerifiableCredential", "ExampleDegreeCredential", "ExamplePersonCredential"],
   "issuer": "https://university.example/issuers/14",
   "validFrom": "2010-01-01T19:23:24Z",
   "credentialSubject": {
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -2435,13 +2435,20 @@ <h3>Data Schemas</h3>
...
-determine if the provided data conforms to the provided schema. Each
+determine if the provided data conforms to the provided schema(s). Each
 <code>credentialSchema</code> MUST specify its <code>type</code> (for example,
-<code>CredentialSchema2022</code>), and an <code>id</code> <a>property</a>
+<code>VerifiableCredentialSchema2023</code>), and an <code>id</code> <a>property</a>
 that MUST be a <a>URL</a> identifying the schema file. The precise contents of
 each data schema is determined by the specific type definition.
+            </p>
+            <p>
+If multiple schemas are present, validity MUST be determined according to the
```
msporny, line 2448:
```suggestion
If multiple schemas are present, validity is determined according to the
```

If we put a MUST here, we will have to test it in the core data model test suite... it's do-able, but puts a pretty high burden on this feature... and if we don't get enough implementations for this feature, it'll be struck from the specification. Are we sure that we'll have enough implementers that implement two types of JSON Schema checking for this to survive CR?

index.html
```diff
@@ -2472,24 +2479,32 @@ <h3>Data Schemas</h3>
...
-  <span class="highlight">"credentialSchema": {
+  <span class="highlight">"credentialSchema": [{
     "id": "https://example.org/examples/degree.json",
-    "type": "CredentialSchema2023"
-  }</span>
+    "type": "VerifiableCredentialSchema2023"
+  },
+  {
+    "id": "https://example.org/examples/university.json",
+    "type": "JsonSchema2023"
```
msporny, line 2494: I know I owe you a response in the other repo on the difference between these two things. I'm still concerned that we don't need both of them, but we'd need to hash this out over there. So, I won't block this PR because of this... just noting that I'm still super uneasy about having two different types of schema.

index.html
```diff
@@ -2472,24 +2479,32 @@ <h3>Data Schemas</h3>
     "degree": {
       "type": "ExampleBachelorDegree",
       "name": "Bachelor of Science and Arts"
+    },
+    "university": {
+      "type": "ExampleUniversity",
+      "name": "University of Examples, W3C"
```
msporny, line 2485: Move this "university" block inside of the "degree" credential, otherwise you're not tying the degree to the university (and vice-versa).

msporny (changes requested):
Minor changes requested, some concerns noted, but none of it is blocking.

TallTed (approved):
lgtm, given @msporny's suggestions

index.html
decentralgabe: I'm fine to change these examples to just use `JsonSchema2023`, though I was hoping to demonstrate that the types do not need to be the same. Is there another second type I could use in place?

index.html
decentralgabe: yes, we have the implementers. I'm fine with removing the MUST

andresuribe87 (approved):

index.html
decentralgabe: I believe that was my intention--I want to show that there can be related data in the same credential, covered by two distinct JSON schemas.

The tying happens when the issuer signs over the entire subject IMO. 

For example, the first schema for the degree may be....
```json
{
    "$id": "https://example.com/schemas/degree-credential-schema.json",
    "$schema": "https://json-schema.org/draft/2020-12/schema",
    "name": "Degree VC Schema",
    "type": "object",
    "properties": {
      "credentialSubject": {
        "type": "object",
        "properties": {
          "degree": {
            "type": "object",
            "properties": {
              "type": { "type": "string"},
              "name": { "type": "string"}
            },
             "required": ["type", "name"]
          }
        },
        "required": ["degree"]
      }
    }
```

And the second, for the university...
```json
{
    "$id": "https://example.com/schemas/degree-credential-schema.json",
    "$schema": "https://json-schema.org/draft/2020-12/schema",
    "name": "University VC Schema",
    "type": "object",
    "properties": {
      "credentialSubject": {
        "type": "object",
        "properties": {
          "university": {
            "type": "object",
            "properties": {
              "type": { "type": "string"},
              "name": { "type": "string"}
            },
             "required": ["type", "name"]
          }
        },
        "required": ["university"]
      }
    }
```

However, if the university block was inside degree it would be awkward to have this represented by two schemas, though possible.

OR13: I agree with @msporny ... however... this kinda highlights why just doing JSON-LD is not enough... 

You need to *think* about the RDF graph the JSON-LD becomes... in order to ensure that claims attribute to the correct `id`.... out spec says a bunch of informative stuff about this, which implementers ignore (because they are not required to understand the RDF graph)... but then we see a lot of comments like this... all over the place on implementations and profiles of the spec... it is a hint at a missing normative requirement.

You current example says:

`did:example:ebfeb1f712ebc6f1c276e12ec21` has a `degree`
`did:example:ebfeb1f712ebc6f1c276e12ec21` has a `university`

I guess if `did:example:ebfeb1f712ebc6f1c276e12ec21` is some identifier for a digital diploma this makes sense... so the example is valid... the comments on architecture of claims still hint at potentially providing better guidance, normatively... without some normative reason to follow them I think its safe to ignore them.

OR13: Screen Shot 2023-07-12 at 5 08 29 PM

decentralgabe: Yeah, after reflection, I better understand the point @msporny is making, but also agree with you @OR13 it seems like there needs to be normative guidance around this. Without it, I can see a credential being interpreted in multiple different ways.

As I have it written, it doesn't make much sense when viewed as an RDF graph. I'm wondering if there's another example I can use with separate top-level claims. Maybe both a degree and a name credential would be less controversial.

OR13: Another version of the same confusion can be found here: https://github.com/w3c/vc-data-model/pull/1172#discussion_r1261786118

dlongley: I think the main difference is that the VCDM requires the connections between things to be explicit in how the modeling occurs in the data vs. just being in the head of the author. You don't have to guess what the author intended for how things "could be interpreted", but rather, it's in the data itself based on the required patterns. And, it's more of a consequence of that -- that we can build tools that will show the relationships between the data as a graph of information. You can't do that when the "interpretation" is in the author's head and not the structure of the data itself.

index.html
```diff
@@ -2464,32 +2471,40 @@ <h3>Data Schemas</h3>
...
   "validFrom": "2010-01-01T19:23:24Z",
   "credentialSubject": {
     "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
     "degree": {
       "type": "ExampleBachelorDegree",
       "name": "Bachelor of Science and Arts"
+    },
+    "name": {
+      "firstName": "John",
+      "lastName": "Doe"
```
msporny, line 2485:
```suggestion
    "name": {
      "firstName": "John",
      "lastName": "Doe"
```

Example above doesn't follow best practices: https://schema.org/name

See the example here (JSON-LD tab, for now `name` is typically used): https://schema.org/Person#eg-0001

... when in doubt, look at schema.org for examples. You could do `alumniOf` (or `memberOf`):

https://schema.org/alumniOf

```
  "alumniOf": {
    "name": "Example University"
   }
```

decentralgabe: thanks, updated to `alumniOf`

index.html
```diff
@@ -2464,32 +2471,39 @@ <h3>Data Schemas</h3>
...
   },
-  <span class="highlight">"credentialSchema": {
+  <span class="highlight">"credentialSchema": [{
     "id": "https://example.org/examples/degree.json",
-    "type": "CredentialSchema2023"
-  }</span>
+    "type": "VerifiableCredentialSchema2023"
+  },
+  {
+    "id": "https://example.org/examples/name.json",
```
decentralgabe, line 2492:
```suggestion
    "id": "https://example.org/examples/person.json",
```

index.html
msporny: > Is there another second type I could use in place?

At this point in time, no. :)

Having both be JsonSchema2023 would be fine, IMHO... as long as the files referenced are different (which they would be).

index.html
```diff
@@ -2435,13 +2435,20 @@ <h3>Data Schemas</h3>
...
           <dd>
-The value of the <code>credentialSchema</code> <a>property</a> MUST be one or
+            <p>
+The value of the <code>credentialSchema</code> <a>property</a> SHOULD be one or
 more data schemas that provide <a>verifiers</a> with enough information to
-determine if the provided data conforms to the provided schema. Each
+determine if the provided data conforms to the provided schema(s). Each
 <code>credentialSchema</code> MUST specify its <code>type</code> (for example,
-<code>CredentialSchema2022</code>), and an <code>id</code> <a>property</a>
+<code>VerifiableCredentialSchema2023</code>), and an <code>id</code> <a>property</a>
```
msporny, line 2443:
```suggestion
<code>JsonSchema2023</code>), and an <code>id</code> <a>property</a>
```

msporny, line 2443:
```suggestion
<code>JsonSchema2023</code>), and an <code>id</code> <a>property</a>
```

index.html
```diff
@@ -2464,32 +2471,39 @@ <h3>Data Schemas</h3>
...
+    "alumniOf": {
+      "name": "Example University"
     }
   },
-  <span class="highlight">"credentialSchema": {
+  <span class="highlight">"credentialSchema": [{
     "id": "https://example.org/examples/degree.json",
-    "type": "CredentialSchema2023"
-  }</span>
+    "type": "VerifiableCredentialSchema2023"
```
msporny, line 2489:
```suggestion
    "type": "JsonSchema2023"
```

msporny (approved):
Two more minor nits, take 'em or leave 'em, otherwise, LGTM.

index.html
```diff
@@ -2435,13 +2435,20 @@ <h3>Data Schemas</h3>
         <dl>
           <dt><var>credentialSchema</var></dt>
           <dd>
-The value of the <code>credentialSchema</code> <a>property</a> MUST be one or
+            <p>
+The value of the <code>credentialSchema</code> <a>property</a> SHOULD be one or
```
brentzundel, line 2439: curious why the normative downgrade. Is there a use case that can't happen if the value of `credentiSchema` property MUST be one or more data schemas?

decentralgabe: this was a suggestion from @TallTed's comment here https://github.com/w3c/vc-data-model/issues/950#issuecomment-1608008708


> I also note that the current VCDM spec says that `credentialSchema` "_can_ be included by an [issuer](https://www.w3.org/TR/vc-data-model-2.0/#dfn-issuers) in the [verifiable credentials](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verifiable-credential) that it issues". This optionality should be reflected in the definition that follows, i.e., —
> 
> _If present_, the value of the _optional_ `credentialSchema` [property](https://www.w3.org/TR/vc-data-model-2.0/#dfn-property) MUST be one or more data schemas that provide [verifiers](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verifier) with enough information to determine if the provided data conforms to the provided schema.
> 
> All things considered, it would probably also be better to change that MUST to a SHOULD.


tl;dr if including a `credentialSchema` is a _can_ then the normative guidance is a _should_.

I personally do not see a problem with leaving the language as it was, but I defer to those more experienced than me with normative language.

brentzundel: with other optional properties we followed the pattern of saying the property MAY be present, but if it is present, it MUST follow the guidance. See also [validity period](https://w3c.github.io/vc-data-model/#validity-period), [status](https://w3c.github.io/vc-data-model/#status), and `id`, `verifiableCredential`, and `proof` in [presentations](https://w3c.github.io/vc-data-model/#presentations-0)

Softening the MUST to a SHOULD makes the presence of the property not mean the same thing in all uses, which is substantially different than the previous wording, and it no longer matches the language in the rest of the spec. If there is no use case driving this change, I am opposed to it.

OR13: It also means we don't need to test it... MUST requires us to test, SHOULD is optional to test.

decentralgabe: OK, reading this I am in favor of reverting my change. Reverted.

decentralgabe, line 2439:
```suggestion
The value of the <code>credentialSchema</code> <a>property</a> MUST be one or
```

brentzundel (approved):

dlongley (approved):

decentralgabe:
All comments addressed. Been over a week. Multiple editor approvals. Merging

Issue #959, opened 2022-10-26
Clarification needed on the ease of deducing that a subject is the holder.


RieksJ:
[Section C.1 of the VCDM](https://www.w3.org/TR/vc-data-model/#subject-is-the-holder) says: "The most common relationship is when a [subject](https://www.w3.org/TR/vc-data-model/#dfn-subjects) is the [holder](https://www.w3.org/TR/vc-data-model/#dfn-holders). In this case, a [verifier](https://www.w3.org/TR/vc-data-model/#dfn-verifier) can easily deduce that a [subject](https://www.w3.org/TR/vc-data-model/#dfn-subjects) is the [holder](https://www.w3.org/TR/vc-data-model/#dfn-holders) if the [verifiable presentation](https://www.w3.org/TR/vc-data-model/#dfn-verifiable-presentations) is digitally signed by the [holder](https://www.w3.org/TR/vc-data-model/#dfn-holders) and all contained [verifiable credentials](https://www.w3.org/TR/vc-data-model/#dfn-verifiable-credentials) are about a [subject](https://www.w3.org/TR/vc-data-model/#dfn-subjects) that can be identified to be the same as the [holder](https://www.w3.org/TR/vc-data-model/#dfn-holders)."

The fact that VC's (as such) are not about a subject (claims in VCs are) is already being discussed in w3c/vc-imp-guide#70, so no need to do that here as well.

The last sentence states "it is easy to deduce that a subject is the holder if [...] a subject [...] can be identified to be the same as the holder." While that is obviously true (an inference of the form X --> X is always true), I do not see how this would imply that a subject can be identified to be the same as the holder. I would like to see some clarification of how this would then be easy.

Note that using the `nonTransferable` property doesn't help here, because there is no general way by which a verifier can determine whether or not a VP proof/signature was made by the subject of (supposedly the only claim in) the credential (see w3c/vc-data-model#960). That is because the subject identifier can be *any* identifier - it need not be a DID, public key or similar.

TallTed:
fwiw, I also take issue with the first sentence of the quoted segment.  I believe that the assertion that "The most common relationship is when a [subject](https://www.w3.org/TR/vc-data-model/#dfn-subjects) is the [holder](https://www.w3.org/TR/vc-data-model/#dfn-holders)" is entirely baseless (beyond the belief of the writer), and should be changed to *at least* start with "A common relationship..." instead of "The most common relationship...".

David-Chadwick:
[...]

PR #1199, opened 2023-07-12
Add validation section regarding holder

OR13:
Addresses https://github.com/w3c/vc-data-model/issues/959

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1199.html" title="Last updated on Sep 1, 2023, 1:49 PM UTC (de684ba)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1199/3a0e423...de684ba.html" title="Last updated on Sep 1, 2023, 1:49 PM UTC (de684ba)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4857,7 +4857,7 @@ checking the proofs on the <a>verifiable credentials</a>.
         <h4>Holder</h4>
         <p>
 The value associated with the <code>holder</code> <a>property</a> is expected
-to identify a <a>holder</a> that is known and trusted by the
+to identify the <a>holder</a> to the verifier.
 <a>verifier</a>.
         </p>
         <p>
--- a/index.html
+++ b/index.html
@@ -4880,7 +4880,7 @@ See the <a data-cite="VC-IMP-GUIDE/#subject-holder-relationships"></a> and
 `Issuer`, `subject`, and `holder` are graph nodes which support multiple representations, 
 potentially making it complex to evaluate whether these roles are being filled
 by appropriate entities.  Validation is the process by which verifiers apply business rules to 
-evaluate the appropriateness of a particular use of a <a>verifiable credential</a>.
+evaluate the propriety of a particular use of a <a>verifiable credential</a>.
         </p>
         <ul>
           <li>
--- a/index.html
+++ b/index.html
@@ -4889,7 +4889,7 @@ making validating that these roles are performed by the same entity potentially 
         </ul>
         <p>
 A <a>verifier</a> might believe that a <a>verifiable presentation</a> <a>holder</a> 
-is the same entity as a <a>verifiable credential</a> <a>subject</a>, when the following happens:
+is the same entity as a <a>verifiable credential</a> <a>subject</a>, when the following conditions apply:
         </p>
         <ul>
           <li>
--- a/index.html
+++ b/index.html
@@ -4885,7 +4885,7 @@ evaluate the appropriateness of a particular use of a Verifiable Credential.
 <a href="#presentations-0">Presentations</a> define expressions of a `holder` in <a>presentations</a>
           </li>
           <li>
-<a href="#credential-subject">Credential subjects</a> define expressions of `credentialSubject` in <a>credentials</a>
+<a href="#credential-subject">Credential subjects</a> define expressions of a `credentialSubject` in <a>credentials</a>
           </li>
         </ul>
         <p>
--- a/index.html
+++ b/index.html
@@ -4857,7 +4857,7 @@ checking the proofs on the <a>verifiable credentials</a>.
         <h4>Holder</h4>
         <p>
 The value associated with the <code>holder</code> <a>property</a> is expected
-to identify the <a>holder</a> to the <a>verifier</a>.
+to be usable to identify the <a>holder</a> to the <a>verifier</a>.
         </p>
         <p>
 Relevant metadata about the <code>holder</code> <a>property</a> is expected
--- a/index.html
+++ b/index.html
@@ -4908,15 +4908,11 @@ The <a>verifiable presentation</a> includes one or more <a>verifiable credential
 using a mechanism the <a>verifier</a> trusts to protect the integrity of the content.
           </li>
           <li>
-            <p>
 When the identifiers for `holder` and `subject` are the same.
-            </p>
           </li>
           <li>
-            <p>
 When the verification material used to secure a <a>verifiable presentation</a>
 is also present in the claims about the credential <a>subject</a>, either by value or by reference.
-            </p>
           </li>
         </ul>
       </section>
--- a/index.html
+++ b/index.html
@@ -4892,7 +4892,7 @@ evaluate the appropriateness of a particular use of a Verifiable Credential.
 A <a>verifier</a> might need to validate a given <a>verifiable presentation</a>
 against complex business rules; for example, the verifier might need confidence
 that the <a>holder</a> is the same entity as a <a>subject</a> of a <a>verifiable
-credential</a>. In that situation, the following factors should provide
+credential</a>. In that situation, the following factors can provide
 reasonable confidence that the holder presenting a given VC is, in fact, a
 subject of that VC:
         </p>
--- a/index.html
+++ b/index.html
@@ -4907,7 +4907,7 @@ When the identifiers for `holder` and `subject` are the same.
           </li>
           <li>
             <p>
-When the verification material used to provide a <code>proof</code> on a <a>verifiable presentation</a>,
+When the verification material used to secure a <a>verifiable presentation</a>,
 is also present in the claims about the credential <a>subject</a>, either by value or by reference.
             </p>
           </li>
--- a/index.html
+++ b/index.html
@@ -4890,7 +4890,7 @@ evaluate the appropriateness of a particular use of a Verifiable Credential.
         </ul>
         <p>
 A <a>verifier</a> might need to validate a given <a>verifiable presentation</a>
-against complex business rules; for example, the verifier may need confidence
+against complex business rules; for example, the verifier might need confidence
 that the <a>holder</a> is the same entity as a <a>subject</a> of a <a>verifiable
 credential</a>. In that situation, the following factors should provide
 reasonable confidence that the holder presenting a given VC is, in fact, a
--- a/index.html
+++ b/index.html
@@ -4857,8 +4857,7 @@ checking the proofs on the <a>verifiable credentials</a>.
         <h4>Holder</h4>
         <p>
 The value associated with the <code>holder</code> <a>property</a> is expected
-to identify the <a>holder</a> to the verifier.
-<a>verifier</a>.
+to identify the <a>holder</a> to the <a>verifier</a>.
         </p>
         <p>
 Relevant metadata about the <code>holder</code> <a>property</a> is expected
--- a/index.html
+++ b/index.html
@@ -4860,8 +4860,9 @@ The value associated with the <code>holder</code> <a>property</a> is expected
 to be usable to identify the <a>holder</a> to the <a>verifier</a>.
         </p>
         <p>
-Relevant metadata about the <code>holder</code> <a>property</a> is expected
-to be available to the <a>verifier</a>. For example, a <a>holder</a> can
+Relevant metadata about the entity identified by the value of the
+<code>holder</code> <a>property</a> is expected to be available to or
+retrievable by the <a>verifier</a>. For example, a <a>holder</a> can
 publish information containing the verification material used to secure
 <a>verifiable presentations</a>. This metadata is relevant when
 checking proofs on <a>verifiable presentations</a>.
--- a/index.html
+++ b/index.html
@@ -4894,7 +4894,7 @@ is the same entity as a <a>verifiable credential</a> <a>subject</a>, when the fo
         <ul>
           <li>
 The <a>verifiable presentation</a> is secured, 
-using a mechanism the <a>verifier</a> trusts to protect the integrity of `vp+ld+json`.
+using a mechanism the <a>verifier</a> trusts to protect the integrity of <code>vp+ld+json</code>.
           </li>
           <li>
 The <a>verifiable presentation</a> includes one or more <a>verifiable credentials</a> that are secured, 
--- a/index.html
+++ b/index.html
@@ -4873,7 +4873,7 @@ See the <a data-cite="VC-IMP-GUIDE/#subject-holder-relationships"></a> and
         </p>
 
         <p class="note">
-`issuer`, `subject` and `holder` are graph nodes, which support multiple representations, 
+`issuer`, `subject`, and `holder` are graph nodes, which support multiple representations, 
 making validating that these roles are performed by the same entity potentially complex.
         </p>
         <ul>
--- a/index.html
+++ b/index.html
@@ -4884,13 +4884,6 @@ evaluate the propriety of a particular use of a <a>verifiable credential</a>.
         </p>
         <ul>
           <li>
-<a href="#issuer">Issuers</a> define expressions of an `issuer` in <a>credentials</a>
-          </li>
-          <li>
-<a href="#presentations-0">Presentations</a> define expressions of a `holder` in <a>presentations</a>
-          </li>
-          <li>
-<a href="#credential-subject">Credential subjects</a> define expressions of a `credentialSubject` in <a>credentials</a>
           </li>
         </ul>
         <p>
--- a/index.html
+++ b/index.html
@@ -4874,7 +4874,8 @@ See the <a data-cite="VC-IMP-GUIDE/#subject-holder-relationships"></a> and
         <p class="note">
 `Issuer`, `subject`, and `holder` are graph nodes which support multiple representations, 
 potentially making it complex to evaluate whether these roles are being filled
-by the same entity.
+by appropriate entities.  Validation is the process by which verifiers apply business rules to 
+evaluate the appropriateness of a particular use of a Verifiable Credential.
         </p>
         <ul>
           <li>
--- a/index.html
+++ b/index.html
@@ -4890,10 +4890,10 @@ evaluate the propriety of a particular use of a <a>verifiable credential</a>.
 A <a>verifier</a> might need to validate a given <a>verifiable presentation</a>
 against complex business rules; for example, the verifier might need confidence
 that the <a>holder</a> is the same entity as a <a>subject</a> of a <a>verifiable
-credential</a>. In that situation, the following factors can provide
-reasonable confidence that the holder presenting a given 
-<a>verifiable credential</a> is, in fact, a subject of that 
-<a>verifiable credential</a>:
+credential</a>. In such a situation, the following factors can provide a
+<a>verifier</a> with reasonable confidence that the claims expressed regarding
+that identifier, in included <a>verifiable credentials</a>, are, in fact, about
+the current presenter:
         </p>
         <ul>
           <li>
--- a/index.html
+++ b/index.html
@@ -4877,9 +4877,7 @@ See the <a data-cite="VC-IMP-GUIDE/#subject-holder-relationships"></a> and
         </p>
 
         <p class="note">
-`Issuer`, `subject`, and `holder` are graph nodes which support multiple representations, 
-potentially making it complex to evaluate whether these roles are being filled
-by appropriate entities.  Validation is the process by which verifiers apply business rules to 
+Validation is the process by which verifiers apply business rules to 
 evaluate the propriety of a particular use of a <a>verifiable credential</a>.
         </p>
         <ul>
--- a/index.html
+++ b/index.html
@@ -4908,8 +4908,8 @@ using a mechanism the <a>verifier</a> trusts to protect the integrity of the con
 When the identifiers for `holder` and `subject` are the same.
           </li>
           <li>
-When the verification material used to secure a <a>verifiable presentation</a>
-is also present in the claims about the credential <a>subject</a>, either by value or by reference.
+That common identifier can be used to discover or derive the verification
+material used to verify the integrity of that <a>verifiable presentation</a>.
           </li>
         </ul>
       </section>
--- a/index.html
+++ b/index.html
@@ -4907,7 +4907,7 @@ When the identifiers for `holder` and `subject` are the same.
           </li>
           <li>
             <p>
-When the verification material used to secure a <a>verifiable presentation</a>,
+When the verification material used to secure a <a>verifiable presentation</a>
 is also present in the claims about the credential <a>subject</a>, either by value or by reference.
             </p>
           </li>
--- a/index.html
+++ b/index.html
@@ -4888,7 +4888,7 @@ making validating that these roles are performed by the same entity potentially 
           </li>
         </ul>
         <p>
-A <a>verifier</a> might believe that a <a>verifiable presentation</a> <a>holder</a> 
+A <a>verifier</a> might conjecture that a <a>verifiable presentation</a> <a>holder</a> 
 is the same entity as a <a>verifiable credential</a> <a>subject</a>, when the following conditions apply:
         </p>
         <ul>
--- a/index.html
+++ b/index.html
@@ -4904,7 +4904,7 @@ using a mechanism the <a>verifier</a> trusts to protect the integrity of the con
           </li>
           <li>
 The <a>verifiable presentation</a> includes one or more <a>verifiable credentials</a> that are secured, 
-using a mechanism the <a>verifier</a> trusts to protect the integrity of `vc+ld+json`.
+using a mechanism the <a>verifier</a> trusts to protect the integrity of the content.
           </li>
           <li>
             <p>
--- a/index.html
+++ b/index.html
@@ -4893,8 +4893,9 @@ A <a>verifier</a> might need to validate a given <a>verifiable presentation</a>
 against complex business rules; for example, the verifier might need confidence
 that the <a>holder</a> is the same entity as a <a>subject</a> of a <a>verifiable
 credential</a>. In that situation, the following factors can provide
-reasonable confidence that the holder presenting a given VC is, in fact, a
-subject of that VC:
+reasonable confidence that the holder presenting a given 
+<a>verifiable credential</a> is, in fact, a subject of that 
+<a>verifiable credential</a>:
         </p>
         <ul>
           <li>
--- a/index.html
+++ b/index.html
@@ -4873,7 +4873,8 @@ See the <a data-cite="VC-IMP-GUIDE/#subject-holder-relationships"></a> and
 
         <p class="note">
 `issuer`, `subject`, and `holder` are graph nodes, which support multiple representations, 
-making validating that these roles are performed by the same entity potentially complex.
+potentially making it complex to evaluate whether these roles are being filled
+by the same entity.
         </p>
         <ul>
           <li>
--- a/index.html
+++ b/index.html
@@ -4884,7 +4884,7 @@ making validating that these roles are performed by the same entity potentially 
 <a href="#presentations-0">Presentations</a> defines expressions of `holder` in <a>presentations</a>
           </li>
           <li>
-<a href="#credential-subject">credential subject</a> defines expressions of `credentialSubject` in <a>credentials</a>
+<a href="#credential-subject">Credential subject</a> defines expressions of `credentialSubject` in <a>credentials</a>
           </li>
         </ul>
         <p>
--- a/index.html
+++ b/index.html
@@ -4907,7 +4907,7 @@ When the identifiers for `holder` and `subject` are the same.
           </li>
           <li>
             <p>
-When the verification material used to provide a `proof` on a <a>verifiable presentation</a>,
+When the verification material used to provide a <code>proof</code> on a <a>verifiable presentation</a>,
 is also present in the claims about the credential <a>subject</a>, either by value or by reference.
             </p>
           </li>
--- a/index.html
+++ b/index.html
@@ -4857,7 +4857,7 @@ checking the proofs on the <a>verifiable credentials</a>.
         <h4>Holder</h4>
         <p>
 The value associated with the <code>holder</code> <a>property</a> is expected
-to identify an <a>holder</a> that is known to and trusted by the
+to identify a <a>holder</a> that is known and trusted by the
 <a>verifier</a>.
         </p>
         <p>
--- a/index.html
+++ b/index.html
@@ -4884,7 +4884,7 @@ by the same entity.
 <a href="#presentations-0">Presentations</a> define expressions of `holder` in <a>presentations</a>
           </li>
           <li>
-<a href="#credential-subject">Credential subject</a> defines expressions of `credentialSubject` in <a>credentials</a>
+<a href="#credential-subject">Credential subjects</a> define expressions of `credentialSubject` in <a>credentials</a>
           </li>
         </ul>
         <p>
--- a/index.html
+++ b/index.html
@@ -4879,10 +4879,10 @@ evaluate the appropriateness of a particular use of a Verifiable Credential.
         </p>
         <ul>
           <li>
-<a href="#issuer">Issuers</a> define expressions of `issuer` in <a>credentials</a>
+<a href="#issuer">Issuers</a> define expressions of an `issuer` in <a>credentials</a>
           </li>
           <li>
-<a href="#presentations-0">Presentations</a> define expressions of `holder` in <a>presentations</a>
+<a href="#presentations-0">Presentations</a> define expressions of a `holder` in <a>presentations</a>
           </li>
           <li>
 <a href="#credential-subject">Credential subjects</a> define expressions of `credentialSubject` in <a>credentials</a>
--- a/index.html
+++ b/index.html
@@ -4889,8 +4889,12 @@ evaluate the appropriateness of a particular use of a Verifiable Credential.
           </li>
         </ul>
         <p>
-A <a>verifier</a> might conjecture that a <a>verifiable presentation</a> <a>holder</a> 
-is the same entity as a <a>verifiable credential</a> <a>subject</a>, when the following conditions apply:
+A <a>verifier</a> might need to validate a given <a>verifiable presentation</a>
+against complex business rules; for example, the verifier may need confidence
+that the <a>holder</a> is the same entity as a <a>subject</a> of a <a>verifiable
+credential</a>. In that situation, the following factors should provide
+reasonable confidence that the holder presenting a given VC is, in fact, a
+subject of that VC:
         </p>
         <ul>
           <li>
--- a/index.html
+++ b/index.html
@@ -4905,7 +4905,9 @@ The <a>verifiable presentation</a> includes one or more <a>verifiable credential
 using a mechanism the <a>verifier</a> trusts to protect the integrity of the content.
           </li>
           <li>
-When the identifiers for `holder` and `subject` are the same.
+The identifier in the `holder` property of the <a>verifiable presentation</a> and
+at least one identifier property of at least one object in the `credentialSubject` array
+are the same.
           </li>
           <li>
 That common identifier can be used to discover or derive the verification
--- a/index.html
+++ b/index.html
@@ -4878,7 +4878,7 @@ by the same entity.
         </p>
         <ul>
           <li>
-<a href="#issuer">Issuer</a> defines expressions of `issuer` in <a>credentials</a>
+<a href="#issuer">Issuers</a> define expressions of `issuer` in <a>credentials</a>
           </li>
           <li>
 <a href="#presentations-0">Presentations</a> defines expressions of `holder` in <a>presentations</a>
--- a/index.html
+++ b/index.html
@@ -4862,10 +4862,10 @@ to identify a <a>holder</a> that is known and trusted by the
         </p>
         <p>
 Relevant metadata about the <code>holder</code> <a>property</a> is expected
-to be available to the <a>verifier</a>. For example, an <a>holder</a> can
+to be available to the <a>verifier</a>. For example, a <a>holder</a> can
 publish information containing the verification material used to secure
 <a>verifiable presentations</a>. This metadata is relevant when
-checking the proofs on the <a>verifiable presentations</a>.
+checking proofs on <a>verifiable presentations</a>.
         </p>
         <p>
 See the <a href="https://www.w3.org/TR/vc-data-model/#subject-holder-relationships">Subject-Holder Relationships</a> and
--- a/index.html
+++ b/index.html
@@ -4880,7 +4880,7 @@ See the <a data-cite="VC-IMP-GUIDE/#subject-holder-relationships"></a> and
 `Issuer`, `subject`, and `holder` are graph nodes which support multiple representations, 
 potentially making it complex to evaluate whether these roles are being filled
 by appropriate entities.  Validation is the process by which verifiers apply business rules to 
-evaluate the appropriateness of a particular use of a Verifiable Credential.
+evaluate the appropriateness of a particular use of a <a>verifiable credential</a>.
         </p>
         <ul>
           <li>
--- a/index.html
+++ b/index.html
@@ -4881,7 +4881,7 @@ by the same entity.
 <a href="#issuer">Issuers</a> define expressions of `issuer` in <a>credentials</a>
           </li>
           <li>
-<a href="#presentations-0">Presentations</a> defines expressions of `holder` in <a>presentations</a>
+<a href="#presentations-0">Presentations</a> define expressions of `holder` in <a>presentations</a>
           </li>
           <li>
 <a href="#credential-subject">Credential subject</a> defines expressions of `credentialSubject` in <a>credentials</a>
--- a/index.html
+++ b/index.html
@@ -4860,12 +4860,16 @@ The value associated with the <code>holder</code> <a>property</a> is expected
 to be usable to identify the <a>holder</a> to the <a>verifier</a>.
         </p>
         <p>
-Relevant metadata about the entity identified by the value of the
-<code>holder</code> <a>property</a> is expected to be available to or
-retrievable by the <a>verifier</a>. For example, a <a>holder</a> can
+Often relevant metadata about the <a>holder</a>, as identified by the value of the
+<code>holder</code> <a>property</a>, is available to, or
+retrievable by, the <a>verifier</a>. For example, a <a>holder</a> can
 publish information containing the verification material used to secure
-<a>verifiable presentations</a>. This metadata is relevant when
-checking proofs on <a>verifiable presentations</a>.
+<a>verifiable presentations</a>. This metadata is expected to be used when
+checking proofs on <a>verifiable presentations</a>. Some cryptographic
+identifiers contain all necessary metadata in the identifier itself. In those 
+cases, no additional metadata is required. Other identifiers use verifiable
+data registries where such metadata is automatically published for use
+by <a>verifiers</a>, without any additional action by the <a>holder</a>.
         </p>
         <p>
 See the <a data-cite="VC-IMP-GUIDE/#subject-holder-relationships"></a> and
--- a/index.html
+++ b/index.html
@@ -4854,6 +4854,67 @@ checking the proofs on the <a>verifiable credentials</a>.
       </section>
 
       <section class="informative">
+        <h4>Holder</h4>
+        <p>
+The value associated with the <code>holder</code> <a>property</a> is expected
+to identify an <a>holder</a> that is known to and trusted by the
+<a>verifier</a>.
+        </p>
+        <p>
+Relevant metadata about the <code>holder</code> <a>property</a> is expected
+to be available to the <a>verifier</a>. For example, an <a>holder</a> can
+publish information containing the verification material used to secure
+<a>verifiable presentations</a>. This metadata is relevant when
+checking the proofs on the <a>verifiable presentations</a>.
+        </p>
+        <p>
+See the <a href="https://www.w3.org/TR/vc-data-model/#subject-holder-relationships">Subject-Holder Relationships</a> and
+<a data-cite="VC-USE-CASES#user-tasks"></a> for additional examples related to <a>subject</a> and <a>holder</a>.
+        </p>
+
+        <p class="note">
+`issuer`, `subject` and `holder` are graph nodes, which support multiple representations, 
+making validating that these roles are performed by the same entity potentially complex.
+        </p>
+        <ul>
+          <li>
+<a href="#issuer">Issuer</a> defines expressions of `issuer` in <a>credentials</a>
+          </li>
+          <li>
+<a href="#presentations-0">Presentations</a> defines expressions of `holder` in <a>presentations</a>
+          </li>
+          <li>
+<a href="#credential-subject">credential subject</a> defines expressions of `credentialSubject` in <a>credentials</a>
+          </li>
+        </ul>
+        <p>
+A <a>verifier</a> might believe that a <a>verifiable presentation</a> <a>holder</a> 
+is the same entity as a <a>verifiable credential</a> <a>subject</a>, when the following happens:
+        </p>
+        <ul>
+          <li>
+The <a>verifiable presentation</a> is secured, 
+using a mechanism the <a>verifier</a> trusts to protect the integrity of `vp+ld+json`.
+          </li>
+          <li>
+The <a>verifiable presentation</a> includes one or more <a>verifiable credentials</a> that are secured, 
+using a mechanism the <a>verifier</a> trusts to protect the integrity of `vc+ld+json`.
+          </li>
+          <li>
+            <p>
+When the identifiers for `holder` and `subject` are the same.
+            </p>
+          </li>
+          <li>
+            <p>
+When the verification material used to provide a `proof` on a <a>verifiable presentation</a>,
+is also present in the claims about the credential <a>subject</a>, either by value or by reference.
+            </p>
+          </li>
+        </ul>
+      </section>
+
+      <section class="informative">
         <h3>Issuance Date</h3>
 
         <p>
--- a/index.html
+++ b/index.html
@@ -4872,7 +4872,7 @@ See the <a data-cite="VC-IMP-GUIDE/#subject-holder-relationships"></a> and
         </p>
 
         <p class="note">
-`issuer`, `subject`, and `holder` are graph nodes, which support multiple representations, 
+`Issuer`, `subject`, and `holder` are graph nodes which support multiple representations, 
 potentially making it complex to evaluate whether these roles are being filled
 by the same entity.
         </p>
--- a/index.html
+++ b/index.html
@@ -4900,7 +4900,7 @@ reasonable confidence that the holder presenting a given
         <ul>
           <li>
 The <a>verifiable presentation</a> is secured, 
-using a mechanism the <a>verifier</a> trusts to protect the integrity of <code>vp+ld+json</code>.
+using a mechanism the <a>verifier</a> trusts to protect the integrity of the content.
           </li>
           <li>
 The <a>verifiable presentation</a> includes one or more <a>verifiable credentials</a> that are secured, 
--- a/index.html
+++ b/index.html
@@ -4868,7 +4868,7 @@ publish information containing the verification material used to secure
 checking proofs on <a>verifiable presentations</a>.
         </p>
         <p>
-See the <a href="https://www.w3.org/TR/vc-data-model/#subject-holder-relationships">Subject-Holder Relationships</a> and
+See the <a data-cite="VC-IMP-GUIDE/#subject-holder-relationships"></a> and
 <a data-cite="VC-USE-CASES#user-tasks"></a> for additional examples related to <a>subject</a> and <a>holder</a>.
         </p>
 


index.html
```diff
@@ -4853,6 +4853,67 @@ <h3>Issuer</h3>
...
+          <li>
+<a href="#presentations-0">Presentations</a> defines expressions of `holder` in <a>presentations</a>
+          </li>
+          <li>
+<a href="#credential-subject">credential subject</a> defines expressions of `credentialSubject` in <a>credentials</a>
+          </li>
+        </ul>
+        <p>
+A <a>verifier</a> might believe that a <a>verifiable presentation</a> <a>holder</a> 
+is the same entity as a <a>verifiable credential</a> <a>subject</a>, when the following happens:
```
OR13, line 4892: @TallTed this reads terribly, can you help?

index.html
```diff
@@ -4853,6 +4853,67 @@ <h3>Issuer</h3>
...
+        </p>
+        <p>
+Relevant metadata about the <code>holder</code> <a>property</a> is expected
+to be available to the <a>verifier</a>. For example, an <a>holder</a> can
+publish information containing the verification material used to secure
+<a>verifiable presentations</a>. This metadata is relevant when
+checking the proofs on the <a>verifiable presentations</a>.
+        </p>
+        <p>
+See the <a href="https://www.w3.org/TR/vc-data-model/#subject-holder-relationships">Subject-Holder Relationships</a> and
```
OR13, line 4871: Not sure how to refer to this better... its no longer in our spec, but it was in the previous version... a better link here would be good.

index.html
```diff
@@ -4853,6 +4853,67 @@ <h3>Issuer</h3>
         </p>
       </section>
 
+      <section class="informative">
+        <h4>Holder</h4>
+        <p>
+The value associated with the <code>holder</code> <a>property</a> is expected
+to identify an <a>holder</a> that is known to and trusted by the
```
decentralgabe, line 4860:
```suggestion
The value associated with the <code>holder</code> <a>property</a> is expected
to identify a <a>holder</a> that is known and trusted by the
```

index.html
```diff
@@ -4853,6 +4853,67 @@ <h3>Issuer</h3>
...
+<a>verifier</a>.
+        </p>
+        <p>
+Relevant metadata about the <code>holder</code> <a>property</a> is expected
+to be available to the <a>verifier</a>. For example, an <a>holder</a> can
+publish information containing the verification material used to secure
+<a>verifiable presentations</a>. This metadata is relevant when
+checking the proofs on the <a>verifiable presentations</a>.
```
decentralgabe, line 4868:
```suggestion
to be available to the <a>verifier</a>. For example, a <a>holder</a> can
publish information containing the verification material used to secure
<a>verifiable presentations</a>. This metadata is relevant when
checking proofs on <a>verifiable presentations</a>.
```

index.html
```diff
@@ -4853,6 +4853,67 @@ <h3>Issuer</h3>
...
+          </li>
+          <li>
+<a href="#presentations-0">Presentations</a> defines expressions of `holder` in <a>presentations</a>
+          </li>
+          <li>
+<a href="#credential-subject">credential subject</a> defines expressions of `credentialSubject` in <a>credentials</a>
+          </li>
+        </ul>
+        <p>
+A <a>verifier</a> might believe that a <a>verifiable presentation</a> <a>holder</a>
```
decentralgabe, line 4891: "might believe" seems too weak. maybe something like "can reason"

msporny (commented):
+1 overall to the concept of this PR. I've been wanting to have stronger statements around how to verify subject == holder. We haven't put that text in to date because people insisted that had to do with protocol, but this PR demonstrates that we can describe expectations w/o getting into protocol. I'll do a full review of this after the text has settled a bit... just noting preliminary support for "saying something" about this topic in the spec. My preference is we get normative about the statements so it's clear how one can do DIDAuth (or something equivalent) with the data in the VP and VCs (which reflects reality today).

index.html
awoie, line 4860: I'm having issues with the statement that the `holder` is supposed to be known and trusted by the verifier`. It is certainly a possibility but two questions:

1) why do you assume such a strong statement about the holder/verifier relationship? The PR suggests that this statement is universal an does not explain why. I also want to say that the securing mechanisms have nothing to do with whether a holder is trusted or known by the verifier. They just protect the integrity of the payload.
2) why is this even important in the definition of the holder property?

awoie (changes requested):
pls address my comment.

index.html
```diff
@@ -4853,6 +4853,67 @@ <h3>Issuer</h3>
...
+The <a>verifiable presentation</a> is secured, 
+using a mechanism the <a>verifier</a> trusts to protect the integrity of `vp+ld+json`.
+          </li>
+          <li>
+The <a>verifiable presentation</a> includes one or more <a>verifiable credentials</a> that are secured, 
+using a mechanism the <a>verifier</a> trusts to protect the integrity of `vc+ld+json`.
+          </li>
+          <li>
+            <p>
+When the identifiers for `holder` and `subject` are the same.
```
awoie, line 4905: Are we referring to the SUBJECT as `subject`  or as `credentialSubject`? And what happens if a VP has one VC with multiple `credentialSubject`s with different identifiers and/or VCs with different identifiers per `credentialSubject`?

index.html
awoie, line 4892:
```suggestion
is the same entity as a <a>verifiable credential</a> <a>subject</a>, when all of the following applies:
```

index.html
```diff
@@ -4853,6 +4853,67 @@ <h3>Issuer</h3>
...
+          </li>
+        </ul>
+        <p>
+A <a>verifier</a> might believe that a <a>verifiable presentation</a> <a>holder</a> 
+is the same entity as a <a>verifiable credential</a> <a>subject</a>, when the following happens:
+        </p>
+        <ul>
+          <li>
+The <a>verifiable presentation</a> is secured, 
+using a mechanism the <a>verifier</a> trusts to protect the integrity of `vp+ld+json`.
```
awoie, line 4897:
```suggestion
using a mechanism the <a>verifier</a> trusts to protect the integrity of <code>vp+ld+json</code>.
```

index.html
```diff
@@ -4853,6 +4853,67 @@ <h3>Issuer</h3>
...
+          </li>
+          <li>
+            <p>
+When the identifiers for `holder` and `subject` are the same.
+            </p>
+          </li>
+          <li>
+            <p>
+When the verification material used to provide a `proof` on a <a>verifiable presentation</a>,
+is also present in the claims about the credential <a>subject</a>, either by value or by reference.
```
awoie, line 4911:
```suggestion
When the verification material used to provide a <code>proof</code> on a <a>verifiable presentation</a>,
is also present in the claims about the credential <a>subject</a>, either by value or by reference.
```

index.html
awoie: My suggestion would be to remove that sentence entirely.

awoie (changes requested):
direction looks good, pls address comments. i'm also curious how this relates to https://w3c.github.io/vc-jwt.

index.html
OR13: yes, thank you... trying to tap the value of "knowledge representations like RDF"... which are used to build beliefs (like that a subject has a name Alice, or that a drivers license expires in 2 days)...

index.html
OR13: Not my fault, we refer to `subject` constantly throughout the document, despite using the JSON member `credentialSubject` in JSON-LD and an RDF term that ends with that string in application n-quads.

OR13:
> i'm also curious how this relates to https://w3c.github.io/vc-jwt, or is this something that the jose/cose-specific media types need to take care of which might be also fine.

I don't think holder or issuer are restricted to specific media types in the core data model.

Some securing mechanisms (like JWT) already have registered claims that support existing verification systems.

See https://github.com/w3c/vc-data-model/pull/1149 for more details

awoie (approved):
lgtm. perhaps accept my latest suggestion.

index.html
dlongley: > So am I, but its unavoidable, since the verifier needs to resolve keys that represent them... and those keys might be in a country the verifier does not trust, or or a blockchain they don't trust, or from an identifier who has previously lied, been convicted, or sanctioned...
> 
> Maybe the word trust is the problem here, any suggestions?

Hmm, the trouble here perhaps stems from a conflation of the holder's identity with its identifier. I agree that the verifier must be able to understand and process the identifier used to refer to the holder. This, however, does not mean that the verifier has a pre-existing relationship or a trust relationship with the holder itself or that the verifier recognizes the holder's "identity" in any way (other than by an identifier that references 'the same holder').

I think this is what is troubling @awoie and I agree.

It's worth noting (since it was mentioned above) that the issuer is fundamentally different in this respect -- issuer claims cannot be accepted (assumed) "as true" without some kind of external understanding or relationship around who the issuer is (the issuer's actual identity), which is more than just what identifier is being used to refer to the issuer. Of course, a verifier similarly needs to also be able to understand, process, and have confidence that an identifier does, in fact, refer to the issuer they think it does, but that's not the same thing.

I'm not sure yet how we resolve this, but the problem seems to stem from what I read as @OR13 desire to clearly indicate that the holder *identifier* (the "value" of the holder property) needs to be of a sort that the verifier can understand and process (this is certainly true or it won't be accepted) and @awoie's desire to make clear that that doesn't mean the verifier necessarily has any other knowledge about the identity of who the identifier refers to. In fact, a completely new relationship (or "introduction") could be getting established at the very moment.

awoie: IMO, if we would commit my latest suggestion below, this would address my concern: https://github.com/w3c/vc-data-model/pull/1199#discussion_r1262640725

index.html
```diff
@@ -4853,6 +4853,67 @@ <h3>Issuer</h3>
         </p>
       </section>
 
+      <section class="informative">
+        <h4>Holder</h4>
+        <p>
+The value associated with the <code>holder</code> <a>property</a> is expected
+to identify a <a>holder</a> that is known and trusted by the
```
dlongley, line 4860: Another option that might be simpler:

```suggestion
The value associated with the <code>holder</code> <a>property</a> is expected
to identify the <a>holder</a> to the verifier.
```

awoie: +1, that would resolve my concern.

index.html
brentzundel: The original text was moved to the implementation guide: https://w3c.github.io/vc-imp-guide/#subject-holder-relationships

index.html
```diff
@@ -4853,6 +4853,67 @@ <h3>Issuer</h3>
...
+          </li>
+          <li>
+            <p>
+When the identifiers for `holder` and `subject` are the same.
+            </p>
+          </li>
+          <li>
+            <p>
+When the verification material used to provide a <code>proof</code> on a <a>verifiable presentation</a>,
+is also present in the claims about the credential <a>subject</a>, either by value or by reference.
```
brentzundel, line 4911: This makes it sounds like a `proof` property needs to be present, what about 
```suggestion
When the verification material used to secure a <a>verifiable presentation</a>,
is also present in the claims about the credential <a>subject</a>, either by value or by reference.
```

brentzundel: or you could change `<code>proof</code>` to `<a>proof</a>`, but that is more subtle.

awoie (approved):
lgtm

TallTed:
This PR is hugely problematic. I need to give it a close re-read, including all the prior comments, but just touching on a snippet of one comment --

> the verifier must be able to understand and process the identifier used to refer to the holder

There may be use cases where this is so, but an anonymous holder (presenter) is entirely plausible in many cases.

Example: I present a VC coupon to the cashier at my local supermarket. If that coupon is not restricted to one use per customer — such as when it is instead restricted to one use per transaction, or to one use per matching item purchased, etc. — then there is no need to identify myself (as is true for most paper coupons, today) — so there is no need for the verifier to understand nor process any identifier used to refer to me.

index.html
```diff
@@ -4853,6 +4853,67 @@ <h3>Issuer</h3>
         </p>
       </section>
 
+      <section class="informative">
+        <h4>Holder</h4>
+        <p>
+The value associated with the <code>holder</code> <a>property</a> is expected
+to identify the <a>holder</a> to the verifier.
+<a>verifier</a>.
```
dlongley, line 4861: Typo got in here (probably from suggestion mode somewhere):

```suggestion
to identify the <a>holder</a> to the <a>verifier</a>.
```

index.html
```diff
@@ -4853,6 +4853,67 @@ <h3>Issuer</h3>
...
+<a>verifiable presentations</a>. This metadata is relevant when
+checking proofs on <a>verifiable presentations</a>.
+        </p>
+        <p>
+See the <a href="https://www.w3.org/TR/vc-data-model/#subject-holder-relationships">Subject-Holder Relationships</a> and
+<a data-cite="VC-USE-CASES#user-tasks"></a> for additional examples related to <a>subject</a> and <a>holder</a>.
+        </p>
+
+        <p class="note">
+`issuer`, `subject` and `holder` are graph nodes, which support multiple representations,
```
dlongley, line 4876: oxford, ftw:

```suggestion
`issuer`, `subject`, and `holder` are graph nodes, which support multiple representations, 
```

index.html
```diff
@@ -4853,6 +4853,67 @@ <h3>Issuer</h3>
...
+        </p>
+        <ul>
+          <li>
+<a href="#issuer">Issuer</a> defines expressions of `issuer` in <a>credentials</a>
+          </li>
+          <li>
+<a href="#presentations-0">Presentations</a> defines expressions of `holder` in <a>presentations</a>
+          </li>
+          <li>
+<a href="#credential-subject">credential subject</a> defines expressions of `credentialSubject` in <a>credentials</a>
```
dlongley, line 4887: For consistency with the other capitalized bullet points:

```suggestion
<a href="#credential-subject">Credential subject</a> defines expressions of `credentialSubject` in <a>credentials</a>
```

index.html
```diff
@@ -4853,6 +4853,67 @@ <h3>Issuer</h3>
...
+          </li>
+        </ul>
+        <p>
+A <a>verifier</a> might believe that a <a>verifiable presentation</a> <a>holder</a>
```
dlongley, line 4891: If you want another word for "believe" that is still based on reasoning but not "complete":

```suggestion
A <a>verifier</a> might conjecture that a <a>verifiable presentation</a> <a>holder</a> 
```

index.html
```diff
@@ -4853,6 +4853,67 @@ <h3>Issuer</h3>
...
+using a mechanism the <a>verifier</a> trusts to protect the integrity of `vc+ld+json`.
+          </li>
+          <li>
+            <p>
+When the identifiers for `holder` and `subject` are the same.
+            </p>
+          </li>
+          <li>
+            <p>
+When the verification material used to secure a <a>verifiable presentation</a>,
```
dlongley, line 4910:
```suggestion
When the verification material used to secure a <a>verifiable presentation</a>
```

dlongley (approved):
A few minor nits in suggestions. Thanks!

index.html
```diff
@@ -4853,6 +4853,67 @@ <h3>Issuer</h3>
...
+        </p>
+        <p>
+Relevant metadata about the <code>holder</code> <a>property</a> is expected
+to be available to the <a>verifier</a>. For example, a <a>holder</a> can
+publish information containing the verification material used to secure
+<a>verifiable presentations</a>. This metadata is relevant when
+checking proofs on <a>verifiable presentations</a>.
+        </p>
+        <p>
+See the <a href="https://www.w3.org/TR/vc-data-model/#subject-holder-relationships">Subject-Holder Relationships</a> and
```
brentzundel, line 4871: perhaps
```suggestion
See the <a data-cite="VC-IMP-GUIDE/#subject-holder-relationships"></a> and
```

dlongley:
@TallTed's comments also make sense and we should adjust the language a bit to ensure we're covering the use case he mentions. I think we want to make it clear that it's both possible for the holder to provide information the verifier needs for one set of use cases and that it's also acceptable that this doesn't need to happen for another set of use cases. Maybe some text like "when the verifier requires it" or something (somewhere) could be all that's needed. I'm not sure.

@TallTed is there a simple enough suggestion you could make like that to ensure we're covering both sets of use cases?

OR13:
> Maybe some text like "when the verifier requires it" or something (somewhere) could be all that's needed. I'm not sure.

Related thread from sd-jwt....  https://github.com/oauth-wg/oauth-selective-disclosure-jwt/pull/246

This eventually landed as this: https://datatracker.ietf.org/doc/html/draft-ietf-oauth-selective-disclosure-jwt-05#name-forwarding-credentials

> When Key Binding is not enforced, any entity in possession of an SD-JWT Presentation can forward the contents to third parties. When doing so, that entity may remove Disclosures such that the receiver learns only a subset of the claims contained in the original SD-JWT.

> For example, a device manufacturer might produce an SD-JWT containing information about upstream and downstream supply chain contributors. Each supply chain party can verify only the claims that were selectively disclosed to them by an upstream party, and they can choose to further reduce the disclosed claims when presenting to a downstream party.

> In some scenarios this behavior could be desirable, but if it is not, Issuers need to support and Verifiers need to enforce Key Binding.

index.html
```diff
@@ -4853,6 +4853,66 @@ <h3>Issuer</h3>
...
+        </p>
+        <ul>
+          <li>
+<a href="#issuer">Issuer</a> defines expressions of `issuer` in <a>credentials</a>
+          </li>
+          <li>
+<a href="#presentations-0">Presentations</a> defines expressions of `holder` in <a>presentations</a>
+          </li>
+          <li>
+<a href="#credential-subject">Credential subject</a> defines expressions of `credentialSubject` in <a>credentials</a>
```
TallTed, line 4886:
```suggestion
<a href="#credential-subject">Credential subjects</a> define expressions of `credentialSubject` in <a>credentials</a>
```

index.html
```diff
@@ -4853,6 +4853,66 @@ <h3>Issuer</h3>
...
+          </li>
+          <li>
+            <p>
+When the identifiers for `holder` and `subject` are the same.
+            </p>
+          </li>
+          <li>
+            <p>
+When the verification material used to secure a <a>verifiable presentation</a>
+is also present in the claims about the credential <a>subject</a>, either by value or by reference.
```
TallTed, line 4910:
```suggestion
A <a>verifier</a> might speculate that the <a>holder</a> of a <a>verifiable presentation</a> 
is the same entity as the <a>subject</a> of a <a>verifiable credential</a>, based on one or more of the following conditions:
        </p>
        <ul>
          <li>
The <a>verifiable presentation</a> is secured
using a mechanism the <a>verifier</a> trusts to protect the integrity of <code>application/vp+ld+json</code>.
          </li>
          <li>
The <a>verifiable presentation</a> includes one or more <a>verifiable credentials</a> that are secured 
using a mechanism the <a>verifier</a> trusts to protect the integrity of <code>application/vc+ld+json</code>.
          </li>
          <li>
            <p>
The identifiers of the <a>holder</a> of the <a>verifiable presentation</a> and the <a>subject</a> of the <a>verifiable credential</a> are the same.
            </p>
          </li>
          <li>
            <p>
The verification material used to secure the <a>verifiable presentation</a>
is also present in one or more claims about the <a>verifiable credentials</a> <a>subject</a>, either by value or by reference.
```

index.html
```diff
@@ -4853,6 +4853,66 @@ <h3>Issuer</h3>
...
+<a>verifiable presentations</a>. This metadata is relevant when
+checking proofs on <a>verifiable presentations</a>.
+        </p>
+        <p>
+See the <a data-cite="VC-IMP-GUIDE/#subject-holder-relationships"></a> and
+<a data-cite="VC-USE-CASES#user-tasks"></a> for additional examples related to <a>subject</a> and <a>holder</a>.
+        </p>
+
+        <p class="note">
+`issuer`, `subject`, and `holder` are graph nodes, which support multiple representations,
```
TallTed, line 4875:
```suggestion
`Issuer`, `subject`, and `holder` are graph nodes which support multiple representations, 
```

TallTed (changes requested):
My edits are primarily focused on the text. I believe the concerns [I expressed earlier](https://github.com/w3c/vc-data-model/pull/1199#issuecomment-1636025876) have been largely resolved by more recent edits, especially in combination with the changes I request here.

Tagging of `issuer`, `subject`, and `holder` is inconsistent — sometimes they're wrapped in `<code>`, sometimes in `<a>`. It seems we should settle on one tag or the other.

index.html
OR13: @dlongley since I merged suggestions from you on this section before, mind commenting on this? I think we should take it as is.

OR13:
I think this is the last major pending suggestion:

- https://github.com/w3c/vc-data-model/pull/1199#discussion_r1265445465

index.html
```diff
@@ -4853,6 +4853,67 @@ <h3>Issuer</h3>
...
+<a>verifiable presentations</a>. This metadata is relevant when
+checking proofs on <a>verifiable presentations</a>.
+        </p>
+        <p>
+See the <a data-cite="VC-IMP-GUIDE/#subject-holder-relationships"></a> and
+<a data-cite="VC-USE-CASES#user-tasks"></a> for additional examples related to <a>subject</a> and <a>holder</a>.
+        </p>
+
+        <p class="note">
+`Issuer`, `subject`, and `holder` are graph nodes which support multiple representations,
```
decentralgabe, line 4875: is there a href that can be associated with `graph nodes`? which I assume the average reader may be confused by

index.html
```diff
@@ -4853,6 +4853,67 @@ <h3>Issuer</h3>
...
+`Issuer`, `subject`, and `holder` are graph nodes which support multiple representations, 
+potentially making it complex to evaluate whether these roles are being filled
+by the same entity.
+        </p>
+        <ul>
+          <li>
+<a href="#issuer">Issuers</a> define expressions of `issuer` in <a>credentials</a>
+          </li>
+          <li>
+<a href="#presentations-0">Presentations</a> define expressions of `holder` in <a>presentations</a>
```
decentralgabe, line 4884:
```suggestion
<a href="#issuer">Issuers</a> define expressions of an `issuer` in <a>credentials</a>
          </li>
          <li>
<a href="#presentations-0">Presentations</a> define expressions of a `holder` in <a>presentations</a>
```

decentralgabe (approved):

index.html
OR13: Yes, the JSON-LD spec defines graph nodes and how to identify their id an type.

index.html
```diff
@@ -4853,6 +4853,67 @@ <h3>Issuer</h3>
...
+        <p>
+The value associated with the <code>holder</code> <a>property</a> is expected
+to identify the <a>holder</a> to the <a>verifier</a>.
+        </p>
+        <p>
+Relevant metadata about the <code>holder</code> <a>property</a> is expected
+to be available to the <a>verifier</a>. For example, a <a>holder</a> can
+publish information containing the verification material used to secure
+<a>verifiable presentations</a>. This metadata is relevant when
+checking proofs on <a>verifiable presentations</a>.
```
jandrieu, line 4867: This section presumes far too much on behalf of the verifier.

VCs are most useful when, in fact, the verifier knows nothing about holder except what is presented by VCs and VPs.

It may be that you are thinking in terms of a specific securing mechanism, but to be clear, if I have a DID from a method I'm willing to accept, there is no meta-data required to verify. The DID provides the data I need.

index.html
```diff
@@ -4853,6 +4853,67 @@ <h3>Issuer</h3>
         </p>
       </section>
 
+      <section class="informative">
+        <h4>Holder</h4>
+        <p>
+The value associated with the <code>holder</code> <a>property</a> is expected
+to identify the <a>holder</a> to the <a>verifier</a>.
```
jandrieu, line 4860:
```suggestion
to be usable to identify the <a>holder</a> to the <a>verifier</a>.
```

The identifier used may not be known to the verifier prior to the creation of this VP. In that case it does not yet identify anything for the verifier.

index.html
```diff
@@ -4853,6 +4853,67 @@ <h3>Issuer</h3>
...
+        </p>
+        <p>
+See the <a data-cite="VC-IMP-GUIDE/#subject-holder-relationships"></a> and
+<a data-cite="VC-USE-CASES#user-tasks"></a> for additional examples related to <a>subject</a> and <a>holder</a>.
+        </p>
+
+        <p class="note">
+`Issuer`, `subject`, and `holder` are graph nodes which support multiple representations, 
+potentially making it complex to evaluate whether these roles are being filled
+by the same entity.
```
jandrieu, line 4877:
```suggestion
potentially making it complex to evaluate whether these roles are being filled
by appropriate entities.  Validation is the process by which verifiers apply business rules to 
evaluate the appropriateness of a particular use of a Verifiable Credential.
```

Subject==Holder is not the only acceptable validation rule. When my administrator gives my driver's license VC to my travel agent, the agent doesn't need to use cryptography to prove anything about the creator of the presentation. Rather, the verifier knows the business rules that apply, which find it acceptable to use a VC with a matching legal name, even if provided by someone other than the subject.

index.html
```diff
@@ -4853,6 +4853,67 @@ <h3>Issuer</h3>
...
+        </p>
+        <ul>
+          <li>
+<a href="#issuer">Issuers</a> define expressions of `issuer` in <a>credentials</a>
+          </li>
+          <li>
+<a href="#presentations-0">Presentations</a> define expressions of `holder` in <a>presentations</a>
+          </li>
+          <li>
+<a href="#credential-subject">Credential subjects</a> define expressions of `credentialSubject` in <a>credentials</a>
```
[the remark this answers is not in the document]
jandrieu: I don't understand what this is trying to say. On its surface, it's not correct. I'm guessing there's another meaning I'm not picking up.

The credential subjects don't "define expressions of `credentialSubject` in credentials.

The issuer defines the properties of `credentialSubject`s in credentials. Those values are claims represented as key-value properties where the key is taken as an RDF predicate and the value is the RDF object for an RDF statement with the credentialSubject as subject.

What are you trying to say here? Are you defining the property or talking about "Credential Subjects"?

index.html
```diff
@@ -4853,6 +4853,67 @@ <h3>Issuer</h3>
...
+          </li>
+        </ul>
+        <p>
+A <a>verifier</a> might conjecture that a <a>verifiable presentation</a> <a>holder</a> 
+is the same entity as a <a>verifiable credential</a> <a>subject</a>, when the following conditions apply:
```
jandrieu, line 4892: This is not particularly rigorous language and probably doesn't belong in a specification. We shouldn't say anything about "what an [entity] might conjecture". Rather, we should describe the problem and possible approaches concretely.

```suggestion
A <a>verifier</a> might need to validate a given <a>verifiable presentation</a> against complex business rules. For example, when the verifier needs confidence that the <a>holder</a> is the same entity as a the  <a>subject</a> of <a>verifiable credential</a>. In that situation, the following factors should provide reasonable confidence that the holder presenting a given VC is, in fact, the subject of that VC:
```

jandrieu (changes requested):
I'd like to see the distinctions between validation and verification clarified here, as that nuance continues to be a significant point of confusion.

Ironically, although the PR talks about a 'validation' section, neither 'valid' nor 'validation' appear in the actual text. 

I made some specific requests for changes that I think would help. I expect the flow of the whole section could also be revised, since I'm not sure my piecemeal suggestions end up with the most coherent narrative.

OR13:
It's trying to say that the data model treats these roles as graph nodes... And therefore that comparing id values has some meaning from JSON-LD... Of course, if the id values can change, this doesn't work. Hence the current complexity, regarding believing that the data has somehow been made "verifiable"... I agree with a lot of your comments.

index.html
dlongley: @OR13 it's fine with me.

dlongley: The one change I recommend... I wish I could easily suggest it, would be in this part of the opening statement in the above change request:

```
is the same entity as the <a>subject</a> of a <a>verifiable credential</a>,
```

Should probably be:

```
is the same entity as a <a>subject</a> of a <a>verifiable credential</a>,
```

Allowing for multiple statements ("the" => "a").

jandrieu:
> It's trying to say that the data model treats these roles as graph nodes... And therefore that comparing id values has some meaning from JSON-LD... Of course, if the id values can change, this doesn't work. Hence the current complexity, regarding believing that the data has somehow been made "verifiable"... I agree with a lot of your comments.

I take "id values" here to mean attestations about a subject and your concern is that a statement that is factual at one point in time, might not be factual at a later point in time?

That's just how language works. The things we say are often true (or false) just in that moment.

I think you're repeating a common mistake. 

VCs don't verify data. 

VCs only verify, if crypto secrets remain secrets, the authenticity of an attestation made using a known cryptographic identifier. That is, we can check that the issuer made a statement (and with the credentialStatus property we can check if it is still timely). However, we cannot verify that the statement itself is a fact.

The understanding is that verifiers would find a surfeit of authorities from whom it could gather data to ascertain its own confidence in a set of operational facts. Sometimes a single authority is fine. Sometimes you need many. Sometimes a "fact" may be true forever (like my birthday) other times, its transient (like my address). Whatever the nature of the fact, its up to the verifier to evaluate its own business rules to decide if a given attestation from a given authority is appropriate to rely on. Which actually has nothing to do with **facts** per se, and everything to do with satisfying the economics of the verifier. Verifiers will only invest so much as makes sense in determining "facts".

You can't use VCs to verify facts. You use them to convince yourself that certain attestations are asserted by knowable authorities. Whether or not you treat such attestations as "facts" is up to you and your validation rules.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-07-19-vcwg#section1-3) on 2023-07-19

- no resolutions were taken

#### 1.3. Add validation section regarding holder (pr vc-data-model#1199)

_See github pull request [vc-data-model#1199](https://github.com/w3c/vc-data-model/pull/1199)._




**Manu Sporny:** We'll wait for re-reviews but there doesn't appear to be any objection.  

</details>

index.html
TallTed:
```suggestion
A <a>verifier</a> might need to validate a given <a>verifiable presentation</a> against complex business rules; for example, the verifier may need confidence that the <a>holder</a> is the same entity as a <a>subject</a> of a <a>verifiable credential</a>. In that situation, the following factors should provide reasonable confidence that the holder presenting a given VC is, in fact, a subject of that VC:
```

OR13:
@msporny since you hold the primary action on the "credential" vs "verifiable credential" clean up... Perhaps using a word other than "evaluate" might help here... perhaps "verify" and "validate" are relevant.

index.html
```diff
@@ -4853,6 +4853,72 @@ <h3>Issuer</h3>
...
+          </li>
+          <li>
+<a href="#presentations-0">Presentations</a> define expressions of `holder` in <a>presentations</a>
+          </li>
+          <li>
+<a href="#credential-subject">Credential subjects</a> define expressions of `credentialSubject` in <a>credentials</a>
+          </li>
+        </ul>
+        <p>
+A <a>verifier</a> might need to validate a given <a>verifiable presentation</a>
```
msporny, line 4892:
```suggestion
A <a>verifier</a> might need to evaluate a given <a>verifiable presentation</a>
```

index.html
```diff
@@ -4853,6 +4853,72 @@ <h3>Issuer</h3>
...
+against complex business rules; for example, the verifier may need confidence
```
msporny, line 4893:
```suggestion
against complex business rules; for example, the verifier might need confidence
```

index.html
```diff
@@ -4853,6 +4853,72 @@ <h3>Issuer</h3>
...
+that the <a>holder</a> is the same entity as a <a>subject</a> of a <a>verifiable
+credential</a>. In that situation, the following factors should provide
```
msporny, line 4895:
```suggestion
credential</a>. In that situation, the following factors can provide
```

index.html
```diff
@@ -4853,6 +4853,72 @@ <h3>Issuer</h3>
...
+reasonable confidence that the holder presenting a given VC is, in fact, a
+subject of that VC:
```
msporny, line 4897:
```suggestion
reasonable confidence that the holder presenting a given 
<a>verifiable credential</a> is, in fact, a subject of that 
<a>verifiable credential</a>:
```

index.html
```diff
@@ -4853,6 +4853,72 @@ <h3>Issuer</h3>
...
+        </p>
+        <ul>
+          <li>
+The <a>verifiable presentation</a> is secured, 
+using a mechanism the <a>verifier</a> trusts to protect the integrity of <code>vp+ld+json</code>.
```
msporny, line 4902:
```suggestion
using a mechanism the <a>verifier</a> trusts to protect the integrity of the content.
```

Note that there could be many types of integrity protection in play, not just on the VP... but for `relatedResource` as well.

index.html
```diff
@@ -4853,6 +4853,72 @@ <h3>Issuer</h3>
...
+          </li>
+          <li>
+The <a>verifiable presentation</a> includes one or more <a>verifiable credentials</a> that are secured, 
+using a mechanism the <a>verifier</a> trusts to protect the integrity of `vc+ld+json`.
```
msporny, line 4906:
```suggestion
using a mechanism the <a>verifier</a> trusts to protect the integrity of the content.
```

msporny (changes requested):
I'm a +1 to this PR with the changes requested by @decentralgabe applied and questions asked by @jandrieu answered. My change requests are largely editorial in nature. I'd rather not specify the media type, but if you disagree, you should use the complete media type (because that is what is at the bottom of the spec and what will be registered w/ IANA).

index.html
OR13: Are these 2 words actually the same, or does validate have specific meaning in the data model, that is different than evaluate?

OR13: Are these 2 words actually the same, or does validate have specific meaning in the data model, that is different than evaluate?

msporny: I suggested this change based off of your suggestion here: https://github.com/w3c/vc-data-model/pull/1199#issuecomment-1644407950

If you didn't mean to imply that we should use the word "evaluate", then you don't need to take this change.

The word "validate" does have a specific meaning in the data model: 

https://w3c.github.io/vc-data-model/#dfn-credential-validation

We have a whole section on it in the spec:

https://w3c.github.io/vc-data-model/#validation

OR13: Sorry if my comment there was not clear, I was trying to say, we should use the word validate if it's meaning in our spec matches it's use here, and avoid using a new term.

If the words are essentially synonyms we should update our terminology to express that, especially to assist with internationalization.

Until that happens I prefer to keep it as.

msporny: @OR13 wrote:
> Until that happens I prefer to keep it as.

Works for me, resolving conversation.

msporny (approved):
Pre-emptively approving as long as @jandrieu and @decentralgabe's concerns are addressed.

index.html
TallTed: > if I have a DID from a method I'm willing to accept, there is no meta-data required to verify. The DID provides the data I need.

...which implies that there is no *relevant metadata* to be available, which satisfies the first sentence.

What would you change?

index.html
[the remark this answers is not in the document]
TallTed: @OR13 — I tweaked the language on these single-sentence paragraphs, but I confess to not fully understanding what they are trying to do. As @jandrieu is in the same boat, it seems some additional text is needed from you...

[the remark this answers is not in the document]
decentralgabe: agree @jandrieu maybe something like

```
The `credentialSubject` property is used to define expressions applied to the subject of a credential
```

index.html
jandrieu: If there is no metadata, then I would not expect it to be available. 

Perhaps more importantly, this notion that a holder is going to publish anything is confusing. Holders don't "publish".

I think this is *trying* to explain that if the holder property is not a DID, you can do magic, but if it's not, you need a bunch of extra metadata. But its trying to say this without mentioning DIDs.

However, this is not correct. It could be a simple URL with a type someone knows how to use to do the verification.

TallTed:
@OR13 — Can you answer the open questions above?

index.html
```diff
@@ -4853,6 +4853,73 @@ <h3>Issuer</h3>
...
 
+      <section class="informative">
+        <h4>Holder</h4>
+        <p>
+The value associated with the <code>holder</code> <a>property</a> is expected
+to be usable to identify the <a>holder</a> to the <a>verifier</a>.
+        </p>
+        <p>
+Relevant metadata about the <code>holder</code> <a>property</a> is expected
+to be available to the <a>verifier</a>. For example, a <a>holder</a> can
```
jandrieu, line 4864:
```suggestion
Relevant metadata about the <code>holder</code> <a>property</a> is expected
to be available to the <a>verifier</a>. For example, a <a>holder</a> can
```

This sentence doesn't improve anything, and implies that relevant metadata is expected. Since there may be no other metadata published anywhere, we should probably just delete it.

Is there something that gets lost if we remove it?

To my mind, if the property value is usable to identify the holder, then the holder understands enough to do so, including any implicit metadata such as how to interpret a particular identifier format.

OR13: it won't let me accept this change suggestion

OR13: it does seem like verifier needs to discover holder keys in order to verify things.

index.html
OR13: This is copy pasted from the issuer section on security... IMO it either applies to issuers and holders consistently (meta data related to keys is required to check proofs), or holders don't have keys / presentations don't have security.

index.html
OR13: https://www.w3.org/TR/json-ld11/#node-identifiers

I don't see a need to link JSON-LD, since its covered in other sections, but here is the direct link.

OR13 (commented):
I have applied some feedback, please re-review.

index.html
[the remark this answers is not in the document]
OR13: I can't commit this, its outdated.

index.html
OR13: I can't commit this its outdated.

index.html
TallTed: > [@OR13] it won't let me accept this change suggestion

That might be because there doesn't appear to be any change in the suggested new text from @jandrieu, which I think was meant to delete the quoted existing sentence?

That said... I think this sentence _is_ a problem. Maybe this will help?

```suggestion
Relevant metadata about the entity identified by the value of the
<code>holder</code> <a>property</a> is expected to be available to or
retrievable by the <a>verifier</a>. For example, a <a>holder</a> can
```

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-08-01-vcwg-special#section1-2) on 2023-08-01

- no resolutions were taken

#### 1.2. Add validation section regarding holder (pr vc-data-model#1199)

_See github pull request [vc-data-model#1199](https://github.com/w3c/vc-data-model/pull/1199)._




**Brent Zundel:** issue 1199, Add Validation Section regarding Holder.  
… I see TallTed and JoeAndrieu specifically requested changes. Have they been addressed?  
… if not, what would you like to still see?  
… the last note from Orie, couple weeks ago saying "Joe I think I got all your suggestions".  

**Ted Thibodeau Jr.:** I think I still have open questions to a few people.  
… at least to Orie, and to Joe at different points, basically saying "what change would you want this to have?".  
… and also - "what was this supposed to mean?".  
… there are several lines of defined expressions that I tweaked, but I wasn't clear what they meant to say.  

**Joe Andrieu:** I have to echo that; I didn't quite understand what Orie was trying to say.  

**Brent Zundel:** I recommend you view the text as it exists now, see if there's any greater clarity there.  
… is that something you can do?  

**Joe Andrieu:** will review again.  

**Ted Thibodeau Jr.:** the pieces that I referred to haven't changed (the "define expressions of"). I pinged Orie at the end.  

**Brent Zundel:** ok, moving to next PR.  

</details>

index.html
TallTed: > [@jandrieu] if the holder property is not a DID, you can do magic, but if it's not, you need a bunch of extra metadata

I'm confused by `if ... not, but if ... not`. It seems that one of these should not have the `not`.

> [@OR13] This is copy pasted from the issuer section on security

Why is anything being said identically in two sections?

> [@OR13] applies to issuers and holders consistently

I think issuers (who will be issuing _MANY_ VCs) are expected to have substantially more VC-related infrastructure than holders (who will be issuing relatively small numbers of VCs/VPs, probably via wallets in most if not all cases), so I don't think much will `[apply] consistently` between them. I'm not yet convinced that this should be one of those things that `[apply] consistently`.

index.html
TallTed: @OR13 — Any comment on the above?

TallTed (changes requested):

OR13:
@TallTed in general, issuers and holders produce proofs for credentials and presentations.

Hence they should have similar text that is relevant to verifiers.

TallTed:
> [@OR13] they should have similar text ... holder validation and issuer validation considerations are similar

Similar, not same. So the copied-and-pasted segment should also have been edited, to reflect the (apparently) subtle differences.

-----

There are several open questions above, which I've now edited to include your handle. Please review, and respond to all.

index.html
OR13: I'm accepting this change, and noting that it applies to every instance of `issuer` as well... the `issuer` identifies an `entity` that performs issuance? the `issuer` is not the `entity` ? This seems like a version of http range 14.

index.html
```diff
@@ -4853,6 +4853,73 @@ <h3>Issuer</h3>
...
+        </p>
+        <ul>
+          <li>
+<a href="#issuer">Issuers</a> define expressions of an `issuer` in <a>credentials</a>
+          </li>
+          <li>
+<a href="#presentations-0">Presentations</a> define expressions of a `holder` in <a>presentations</a>
+          </li>
+          <li>
+<a href="#credential-subject">Credential subjects</a> define expressions of a `credentialSubject` in <a>credentials</a>
```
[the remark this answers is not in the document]
OR13: Essentially the same thing you proposed, and I accepted here: https://github.com/w3c/vc-data-model/pull/1199#discussion_r1283406163

For clarity:

JSON Members -> RDF Graph -> Knowledge Representation

"holder" -> "graph node" -> entity that possesses bytes.
"issuer" -> "graph node" -> entity that signs bytes.
"verifier" -> "graph node" -> entity that verifies bytes.
"subject" -> "graph node" -> entity about which bytes representation of claims are made.

OR13:
@jandrieu have your concerns been addressed?

index.html
TallTed: `issuer`, `subject`, `holder`, and `verifier` all have range `entity`.

* Yes, `issuer` is an explicitly identified `entity` that issues s VC.
* Yes, `subject` is an explicitly or implicitly identified `entity`, referenced within a VC as the subject of one or more claims.
* Yes, `holder` is an implicitly and usually explicitly identifiable `entity`, to which a VC is issued or transferred.
* Yes, `verifier` is an explicitly identifiable `entity`, which tests whether a VC has been tampered with.

OR13: This seems potentially much clearer that what we have in the current spec... and it also seems to apply directly to: https://github.com/w3c/vc-data-model/pull/1172

brentzundel: @TallTed I agree with all of the ranges you have outlined except for `subject`, which does not need to be an `entity` as the `subject` of a verifiable credential can be anything.

OR13:
@jandrieu per the call, I opened https://github.com/w3c/vc-data-model/issues/1232

index.html
RieksJ: I read the current definition of `entity` in the terminology section as something with an independent existence that performs one or more roles in the ecosystem, which would typically be 'holder', 'issuer', 'verifier'. That would mean that a something with an independent existence that is not (capable of) performing any such role, doesn't qualify. Examples include books, dogs, passports, and lots of other stuff. If this interpretation is correct, then any such example cannot be used as a subject in a claim, because they do n't qualify as an entity.

I suggest to have a term (name to be decided, let's call it `X` for now) that refers to something (or someone) that has an independent existence. We don't need to use `entity` for that. It's another concept than what `entity` refers to; `entity` is a specialization of `X`.

`issuer`, `holder`, `verifier` then keep having range `entity`, and `subject` would have range `X`, and since `entity` is a specialization of `X`, any `entity` can still be the subject of a claim, but can then also make claims about an `X` that is not an `entity`.

TallTed: As noted elsewhere, I refer all and sundry to [RDF Concepts, Introduction](https://www.w3.org/TR/rdf11-concepts/#resources-and-statements), where `resource` are clearly stated to be synonymous with `entity`, and both/either may be *anything*, and may be the `subject` of a VC.

In order to perform a role, the given entity/resource must be capable of such actions as are required for that role; for instance, a document probably cannot be an issuer or verifier.

TallTed:
As noted [above](https://github.com/w3c/vc-data-model/pull/1199#discussion_r1291580028) and elsewhere, I refer all and sundry to [RDF Concepts, Introduction](https://www.w3.org/TR/rdf11-concepts/#resources-and-statements), where resource are clearly stated to be synonymous with entity, and both/either may be anything, and may be the subject of a VC.

In order to perform a role, the given entity/resource must be capable of such actions as are required for that role; for instance, a document probably cannot be an issuer or verifier.

The *ranges* of `holder`, `issuer`, `verifier`, and `subject` all remain `entity` (a/k/a `resource`). The capacity of an individual entity/resource to fulfill a role is a distinct question, which might best fall into the business logic of a `verifier`.

jandrieu (changes requested):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-08-15-vcwg-special#section1-2) on 2023-08-15

- no resolutions were taken

#### 1.2. Add validation section regarding holder (pr vc-data-model#1199)

_See github pull request [vc-data-model#1199](https://github.com/w3c/vc-data-model/pull/1199)._




**Brent Zundel:** next 1199 add validation section regarding holder; a few approvals. outstanding change requests from Joe. believe they've been addressed. please re-review.  
… Ted also has some outstanding changes.  

**Joe Andrieu:** I have reviewed recently.  
… have pending comments. forgot to hit button to send comments.  

**Brent Zundel:** moving in a direction where consensus is still possible.  

</details>

index.html
```diff
@@ -4853,6 +4853,70 @@ <h3>Issuer</h3>
...
+The value associated with the <code>holder</code> <a>property</a> is expected
+to be usable to identify the <a>holder</a> to the <a>verifier</a>.
+        </p>
+        <p>
+Relevant metadata about the entity identified by the value of the
+<code>holder</code> <a>property</a> is expected to be available to or
+retrievable by the <a>verifier</a>. For example, a <a>holder</a> can
+publish information containing the verification material used to secure
+<a>verifiable presentations</a>. This metadata is relevant when
+checking proofs on <a>verifiable presentations</a>.
```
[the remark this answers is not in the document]
TallTed: * Line 4867 needs a `when`, to fall between `used` and `checking` (the latter starting line 4868).
* Line 4872 needs a terminating `.` following `<a>holder</a>`.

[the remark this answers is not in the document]
TallTed:
```suggestion
Often relevant metadata about the <a>holder</a>, as identified by the value of the
<code>holder</code> <a>property</a>, is available to, or
retrievable by, the <a>verifier</a>. For example, a <a>holder</a> can
publish information containing the verification material used to secure
<a>verifiable presentations</a>. This metadata is expected to be used when
checking proofs on <a>verifiable presentations</a>. Some cryptographic
identifiers contain all necessary metadata in the identifier itself. In those 
cases, no additional metadata is required. Other identifiers use verifiable
data registries where such metadata is automatically published for use
by <a>verifiers</a>, without any additional action by the <a>holder</a>.
```

OR13:
@jandrieu I took @TallTed 's suggestion... have your concerns been addressed, or are there others?

index.html
```diff
@@ -4853,6 +4853,74 @@ <h3>Issuer</h3>
...
+        <p>
+See the <a data-cite="VC-IMP-GUIDE/#subject-holder-relationships"></a> and
+<a data-cite="VC-USE-CASES#user-tasks"></a> for additional examples related to <a>subject</a> and <a>holder</a>.
+        </p>
+
+        <p class="note">
+`Issuer`, `subject`, and `holder` are graph nodes which support multiple representations, 
+potentially making it complex to evaluate whether these roles are being filled
+by appropriate entities.  Validation is the process by which verifiers apply business rules to 
+evaluate the appropriateness of a particular use of a Verifiable Credential.
```
brentzundel, line 4883:
```suggestion
evaluate the appropriateness of a particular use of a <a>verifiable credential</a>.
```

brentzundel (approved):
minor nit

OR13:
AFAIK, this is still waiting on a re-review from @jandrieu .

index.html
```diff
@@ -4880,7 +4880,7 @@ <h4>Holder</h4>
 `Issuer`, `subject`, and `holder` are graph nodes which support multiple representations, 
 potentially making it complex to evaluate whether these roles are being filled
 by appropriate entities.  Validation is the process by which verifiers apply business rules to 
-evaluate the appropriateness of a particular use of a Verifiable Credential.
+evaluate the appropriateness of a particular use of a <a>verifiable credential</a>.
```
TallTed, line 4883:
```suggestion
evaluate the propriety of a particular use of a <a>verifiable credential</a>.
```

index.html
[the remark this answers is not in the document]
TallTed: Unfortunately, https://github.com/w3c/vc-data-model/pull/1199#discussion_r1283406163 no longer goes there, because that discussion was resolved, and your "for clarity" segment brings me (and perhaps others) no closer to understanding what is meant by the three "define expressions" sentences.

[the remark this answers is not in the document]
dlongley: I agree these three sentences are confusing ... I don't really know what they mean. Some of that confusion may come from what seems to be a conflation between roles and the expression of information about the entities that play those roles. I don't believe "Presentations" or "Credential subjects" "define" anything either.

index.html
```diff
@@ -4853,6 +4853,74 @@ <h3>Issuer</h3>
...
+          <li>
+The <a>verifiable presentation</a> includes one or more <a>verifiable credentials</a> that are secured, 
+using a mechanism the <a>verifier</a> trusts to protect the integrity of the content.
+          </li>
+          <li>
+When the identifiers for `holder` and `subject` are the same.
+          </li>
+          <li>
+When the verification material used to secure a <a>verifiable presentation</a>
+is also present in the claims about the credential <a>subject</a>, either by value or by reference.
```
jandrieu, line 4919: This seems incorrect.

As long as the verifier can retrieve the appropriate verification material, it need not be present in the claims. Although you can infer that "by reference" with the right logical gymnastics, it's more clear to say that the claims in a credentialSubject value are about the subject referred to by the identifier of that entry in the credentialSubject array. The "id" is not a claim, nor is it "present in the claims".

```suggestion
When that common identifier can be used to discover or derive the verification material used to verify the integrity of that<a>verifiable presentation</a>.

When these are true, then <a>verifiers</a> can reasonably infer that the claims expressed regarding that identifier, in included VCs, are about the current presenter.
```

It's important that we respect the meaning of this in the case of multiple subjects. In a marriage license, the officiant would be identified and they are as legitimate a holder/presenter as the witness or the spouses.

jandrieu (changes requested):
It's getting there, but we still have some sections that need work.

index.html
```diff
@@ -4853,6 +4853,74 @@ <h3>Issuer</h3>
...
+          <li>
+The <a>verifiable presentation</a> is secured, 
+using a mechanism the <a>verifier</a> trusts to protect the integrity of the content.
+          </li>
+          <li>
+The <a>verifiable presentation</a> includes one or more <a>verifiable credentials</a> that are secured, 
+using a mechanism the <a>verifier</a> trusts to protect the integrity of the content.
+          </li>
+          <li>
+When the identifiers for `holder` and `subject` are the same.
```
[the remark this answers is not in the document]
TallTed: Small tweak of @jandrieu's suggestion
```suggestion
When the identifier in the `holder` property of the <a>verifiable presentation</a> and
at least one identifier property of at least one object in the `credentialSubject` array
are the same.
```

index.html
```diff
@@ -4853,6 +4853,74 @@ <h3>Issuer</h3>
...
+          </li>
+        </ul>
+        <p>
+A <a>verifier</a> might need to validate a given <a>verifiable presentation</a>
+against complex business rules; for example, the verifier might need confidence
+that the <a>holder</a> is the same entity as a <a>subject</a> of a <a>verifiable
+credential</a>. In that situation, the following factors can provide
+reasonable confidence that the holder presenting a given 
+<a>verifiable credential</a> is, in fact, a subject of that 
+<a>verifiable credential</a>:
```
TallTed, line 4903: Working in @jandrieu's [suggestion, seen in full below](https://github.com/w3c/vc-data-model/pull/1199/files#r1310473764)
```suggestion
against complex business rules; for example, the verifier might need confidence
that the <a>holder</a> is the same entity as a <a>subject</a> of a <a>verifiable
credential</a>. In such a situation, the following factors can provide a
<a>verifier</a> with reasonable confidence that the claims expressed regarding
that identifier, in included <a>verifiable credentials</a>, are, in fact, about
the current presenter:
```

index.html
TallTed: Tweaks to @jandrieu's suggestion. Note that @jandrieu's second paragraph here is redundant against lines 4900-4903 above, which introduced this `<ul>`; 
```suggestion
When that common identifier can be used to discover or derive the verification
material used to verify the integrity of that <a>verifiable presentation</a>.
```

OR13: @jandrieu I took @TallTed 's suggestion feel free to suggest further corrections now that its been applied.

OR13:
@jandrieu I mostly took @TallTed 's suggestions which were on top of yours to try to save on review time, please re-review.

index.html
```diff
@@ -4853,6 +4853,69 @@ <h3>Issuer</h3>
...
+        </p>
+        <p>
+See the <a data-cite="VC-IMP-GUIDE/#subject-holder-relationships"></a> and
+<a data-cite="VC-USE-CASES#user-tasks"></a> for additional examples related to <a>subject</a> and <a>holder</a>.
+        </p>
+
+        <p class="note">
+`Issuer`, `subject`, and `holder` are graph nodes which support multiple representations, 
+potentially making it complex to evaluate whether these roles are being filled
+by appropriate entities.  Validation is the process by which verifiers apply business rules to
```
dlongley, line 4882: This sentence has confusion, conflation, and / or contradiction in it. The end of the sentence says three things (`Issuer`, `subject`, and `holder`) are roles. The sentence also says that these roles are graph nodes, which conflates roles with the expression of information about the entities playing the roles. Additionally, a `subject` is neither a graph node nor a role in the VCDM, but rather it's a thing about which claims are made. An `issuer` is also not a graph node, it is a role, one an entity can perform by asserting claims about one or more subjects, etc.

I recommend we strike this sentence; I don't know what's trying to be said, it's quite mixed up, IMO, and I don't have a recommendation to fix it.

```suggestion
Validation is the process by which verifiers apply business rules to 
```

dlongley (changes requested):

index.html
OR13: I agree, however, we will probably need to do a second pass after this, because "validation" applies to expression of information, not to "roles".... I think its probably better to try to land this PR first, before tackling that.

dlongley (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-08-29-vcwg-special#section1-1) on 2023-08-29

- no resolutions were taken

#### 1.1. Add validation section regarding holder (pr vc-data-model#1199)

_See github pull request [vc-data-model#1199](https://github.com/w3c/vc-data-model/pull/1199)._



**Manu Sporny:** waiting on JoeAndrieu re-review.  

**Joe Andrieu:** will look at again.  

</details>

jandrieu (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-09-05-vcwg-special#section3-1) on 2023-09-05

- no resolutions were taken

#### 3.1. Add validation section regarding holder (pr vc-data-model#1199)

_See github pull request [vc-data-model#1199](https://github.com/w3c/vc-data-model/pull/1199)._




**Manu Sporny:** +1199 joe said he would review.  

**Joe Andrieu:** there has been some discussions on this already.  

</details>

TallTed (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-09-14-vcwg#section1-1) on 2023-09-14

- no resolutions were taken

#### 1.1.  (pr vc-data-model#1199)

_See github pull request [vc-data-model#1199](https://github.com/w3c/vc-data-model/pull/1199)._




**Brent Zundel:** Raised by orie. Good review and a number of comments. 1 open request for changes from TallTed. Are you're changes unaddressed?  

**Ted Thibodeau Jr.:** Haven't had the chance to review.  

**Brent Zundel:** If it meets your approval, we can merge.  

</details>

PR #1200, opened 2023-07-14
Better not set §B as non-normative (minor editorial)

iherman:
At the moment §B declares itself as non-normative, and this is, sort of, "inherited" by subsections unless stated otherwise. Ie, §B.1 and §B.2 may be considered as non-normative, which is not the intention. Better leave §B to normative (the default) and §B.3 explicitly declares itself as non-normative.

Just to avoid misunderstandings...

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1200.html" title="Last updated on Jul 14, 2023, 10:15 AM UTC (b536d68)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1200/3a0e423...b536d68.html" title="Last updated on Jul 14, 2023, 10:15 AM UTC (b536d68)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4988,7 +4988,7 @@ franchise. Policy information expressed by the <a>issuer</a> in the
       </section>
     </section>
 
-    <section class="appendix informative">
+    <section class="appendix">
       <h2>Contexts, Vocabularies, Types, and Credential Schemas</h2>
 
       <section class="normative">


OR13:
I prefer the explicit over implicit tagging personally...

Same goes for IETF standards including IANA considerations even when there are none.

But if the convention is different at W3C, better to conform, even if it's less clear for readers.

iherman:
> I prefer the explicit over implicit tagging personally...
> 

Me too. But tagging a class (at least in respec) as 'normative' does _not_ automatically add a notice on the section being normative, only the 'informative' does. I do not think we should lose our time on dealing with this, hence this PR.

OR13:
Tagging something explicitly as informative accomplishes the same effect...

I guess the first question is:

Is the appendix supposed to be informative?

Is so the informative tag should stay imo.

iherman:
> Tagging something explicitly as informative accomplishes the same effect...
> 
> I guess the first question is:
> 
> Is the appendix supposed to be informative?
> 
> Is so the informative tag should stay imo.

I do not really understand what you are asking, or what your objection is, @OR13.

_Currently,_  that appendix has three subsections. Two out of those are (meant to be) normative, one informative. The top level section does not include any top level text on its own. The top level section is explicitly made informative, and there is no notice in the text about the to-be-normative sections being indeed normative, so my claim is that it can be misunderstood as being informative by virtue of inheritance.

_The PR_ simply removes the 'informative' mention at the top level of the appendix which, therefore, becomes automatically normative per the W3C publication rules (See the start of the [conformance section](https://www.w3.org/TR/vc-data-model-2.0/#conformance) in the specification).

The subsection that is meant to be informative is explicitly marked as such, so no change there.

_There is no mechanism in the tool to automatically mark a section to be **normative**_. So if we wanted to keep the top level informative (for whatever reason) then the editors would have to add such a notice manually. I do not see why that would be a better approach.

brentzundel (approved):

msporny (approved):

decentralgabe (approved):

msporny:
It looks like @OR13 is objecting to this PR? I can't tell, can you clarify, please, @OR13?

seabass-labrax (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-08-01-vcwg-special#section1-1) on 2023-08-01

- no resolutions were taken

#### 1.1. Better not set §8 as non-normative (minor editorial) (pr vc-data-model#1200)

_See github pull request [vc-data-model#1200](https://github.com/w3c/vc-data-model/pull/1200)._




**Manu Sporny:** this has a number of agreements, but I think Orie is objecting.  
… I asked to clarify, waiting to hear back.  
… feels like a fairly straightforward change. what Ivan is doing here is, the other sections of this thing are normative,.  
… and he's setting this one as normative as well.  
… Orie is saying "I would rather that we explicitly declare it as normative".  

**Ted Thibodeau Jr.:** lacking Orie (though he might not agree regardless), the issue appears to be partly how ReSpec handles things.  
… so, we can't explicitly declare appendixes to be normative.  
… given the rest of the document, like the top says "Everything is normative, unless marked as informative".  
… so, I don't think there's a change to be made.  

**Brent Zundel:** my read of Orie's comments is not that he's requesting changes, just giving clarifications.  
… he's had plenty of time since then to continue to review. I'm comfortable with the approvals moving forward & merging this.  
… I don't want to get too formal, but we could do resolutions if folks are more comfortable with that.  

> *Manu Sporny:* +1 to merge as-is.

> *Joe Andrieu:* +1 to merge.

**Brent Zundel:** not hearing any objections, queue is empty.  
… let's move on to next issue.  

</details>

msporny:
Editorial, multiple reviews, no changes requested, no objections, merging.

Issue #1201, opened 2023-07-14
Document how VCs are processed as JSON today


msporny:
There are a variety of deployed systems that process VCs as JSON today. 

The v1.0 and v1.1 specification [provided clear guidance](https://www.w3.org/TR/vc-data-model/#contexts) on how to process a VC using only JSON tooling and a little extra logic (from the latest published VCDM v1.1 global standard):

> NOTE: Though this specification requires that a @context [property](https://www.w3.org/TR/vc-data-model/#dfn-property) be present, it is not required that the value of the @context [property](https://www.w3.org/TR/vc-data-model/#dfn-property) be processed using JSON-LD. This is to support processing using plain JSON libraries, such as those that might be used when the [verifiable credential](https://www.w3.org/TR/vc-data-model/#dfn-verifiable-credentials) is encoded as a JWT. All libraries or processors MUST ensure that the order of the values in the @context [property](https://www.w3.org/TR/vc-data-model/#dfn-property) is what is expected for the specific application. Libraries or processors that support JSON-LD can process the @context [property](https://www.w3.org/TR/vc-data-model/#dfn-property) using full JSON-LD processing as expected.

This guidance was removed in PR #1182 and it is [leading to confusion in the ecosystem](https://github.com/w3c/vc-data-model/pull/1088#issuecomment-1591275654); people are starting to believe that you can't process a VC in "pure JSON" mode, and that does not reflect the reality that organizations are doing just that today. Here are the use cases where VCs are processed in "pure JSON" mode today:

* When using the VC-JOSE-COSE specification to sign, processing performed is purely JSON-based.
* When performing JSON Schema validation (via vc-json-schema), processing performed is purely JSON-based.
* When serializing/deserializing VCs into document-oriented databases (or databases in general), processing performed is purely JSON-based (or text-based).
* When operating on the data in a software application, after verification is performed, processing performed is often JSON-based.

The specification used to be clear about this, and after PR #1182 was merged, is no longer clear about this reality. What was in the specification before should be improved. We need to document that the ecosystem is performing this sort of processing today on VCs and that it's an entirely legitimate way to process this data. IOW, JSON-LD processing is optional in a number of cases, and is only mandatory when you want to ensure that the semantics are well formed (and there are a number of use cases where you don't need to do that).

msporny:
PR #1202 has been raised to address this issue. This issue will be closed if PR #1202 is merged.

OR13:
[...]

PR #1202, opened 2023-07-14
Add section on JSON Processing.

msporny:
This PR attempts to address issue #1201 by documenting how systems process the VCDM using pure JSON processing today.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1202.html" title="Last updated on Aug 9, 2023, 10:18 PM UTC (f007a19)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1202/6c5cebb...f007a19.html" title="Last updated on Aug 9, 2023, 10:18 PM UTC (f007a19)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4000,7 +4000,8 @@ implementer is advised to follow the following rule:
             <li>
 Ensure that all values associated with a `@context` property are in the
 expected order, the contents of the context files match known good
-cryptographic hashes for each file, and the contents are known to be valid.
+cryptographic hashes for each file, and domain experts have deemed that the
+contents are appropriate for the intended use case.
             </li>
           </ul>
 
--- a/index.html
+++ b/index.html
@@ -3967,7 +3967,8 @@ Prior to verification when using the JOSE or COSE specifications to secure a
 <a>verifiable credential</a> or <a>verifiable presentation</a>.
             </li>
             <li>
-When performing JSON Schema validation.
+When performing JSON Schema validation, as described in Section
+<a href="#data-schemas"></a>.
             </li>
             <li>
 When serializing or deserializing <a>verifiable credentials</a> or
--- a/index.html
+++ b/index.html
@@ -3963,7 +3963,7 @@ scenarios include, but are not limited to:
 
           <ul>
             <li>
-When using the JOSE or COSE specifications to secure a
+Prior to verification when using the JOSE or COSE specifications to secure a
 <a>verifiable credential</a> or <a>verifiable presentation</a>.
             </li>
             <li>
@@ -3976,7 +3976,8 @@ When serializing or deserializing <a>verifiable credentials</a> or
             <li>
 When operating on <a>verifiable credentials</a> or <a>verifiable
 presentations</a> in a software application after verification or validation
-is performed.
+is performed for securing mechanisms that require an understanding of
+and processing of JSON-LD.
             </li>
             <li>
 When an application chooses to process the media type using the `+json`
@@ -3999,13 +4000,14 @@ cryptographic hashes for each file, and the contents are known to be valid.
           </ul>
 
           <p>
-Using static context files and a JSON Schema to ensure proper URL values and
-order is one acceptable approach when implementing the rule above.
+Using static context files with a JSON Schema is one acceptable approach to
+implementing the rule above. This can ensure proper term identification,
+typing, and order, when a JSON document is processed as JSON-LD.
           </p>
 
           <p>
 The rule above guarantees semantic interoperability between JSON and JSON-LD for
-terms defined by the `@context` mechanism. While JSON-LD processors will use the
+literal JSON keys mapped to URIs by the `@context` mechanism. While JSON-LD processors will use the
 specific mechanism provided and can verify that all terms are correctly
 specified, JSON-based processors implicitly accept the same set of terms without
 testing that they are correct. In other words, the context in which the data
--- a/index.html
+++ b/index.html
@@ -3956,9 +3956,10 @@ system.
           <h2>JSON Processing</h2>
 
           <p>
-While the base media types for this specification are JSON-LD documents, JSON-LD
-processing is not required to be performed in a variety of scenarios. These
-scenarios include, but are not limited to:
+While the base media type used by this specification relies on JSON-LD, JSON-LD
+processing is not required to be performed, since JSON-LD is JSON. Some
+scenarios where processing a <a>verifiable credential</a> or a <a>verifiable
+presentation</a> as JSON is desirable include, but are not limited to:
           </p>
 
           <ul>
--- a/index.html
+++ b/index.html
@@ -3952,6 +3952,69 @@ system.
           </p>
         </section>
 
+        <section>
+          <h2>JSON Processing</h2>
+
+          <p>
+While the base media types for this specification are JSON-LD documents, JSON-LD
+processing is not required to be performed in a variety of scenarios. These
+scenarios include, but are not limited to:
+          </p>
+
+          <ul>
+            <li>
+When using the JOSE or COSE specifications to secure a
+<a>verifiable credential</a> or <a>verifiable presentation</a>.
+            </li>
+            <li>
+When performing JSON Schema validation.
+            </li>
+            <li>
+When serializing or deserializing <a>verifiable credentials</a> or
+<a>verifiable presentations</a> into databases that index their contents.
+            </li>
+            <li>
+When operating on <a>verifiable credentials</a> or <a>verifiable
+presentations</a> in a software application after verification or validation
+is performed.
+            </li>
+            <li>
+When an application chooses to process the media type using the `+json`
+structured media type suffix.
+            </li>
+          </ul>
+
+          <p>
+That is, pure JSON processing is allowed as long as the document being
+consumed or produced is a <a>conforming document</a>. If pure JSON processing
+is desired, an implementer is advised to follow the following rule:
+          </p>
+
+          <ul>
+            <li>
+Ensure that all values associated with a `@context` property are in the
+expected order, and the contents of the context files match known good
+cryptographic hashes for each file.
+            </li>
+          </ul>
+
+          <p>
+Using static context files and a JSON Schema to ensure proper URL values and
+order is one acceptable approach when implementing the rule above.
+          </p>
+
+          <p>
+The rule above guarantees semantic interoperability between JSON and JSON-LD for
+terms defined by the `@context` mechanism. While JSON-LD processors will use the
+specific mechanism provided and can verify that all terms are correctly
+specified, JSON-based processors implicitly accept the same set of terms without
+testing that they are correct. In other words, the context in which the data
+exchange happens is explicitly stated for both JSON and JSON-LD by using the
+same mechanism. With respect to JSON-based processors, this is achieved in a
+lightweight manner, without having to use JSON-LD processing libraries.
+          </p>
+
+        </section>
       </section>
 
       <section>
--- a/index.html
+++ b/index.html
@@ -3967,7 +3967,7 @@ desirable include, but are not limited to:
             <li>
 Before securing or after verifying content
 that requires <a href="https://csrc.nist.gov/glossary/term/data_integrity">data
-integrity</a> such as a
+integrity</a>, such as a
 <a>verifiable credential</a> or <a>verifiable presentation</a>.
             </li>
             <li>
@@ -3980,9 +3980,9 @@ When serializing or deserializing <a>verifiable credentials</a> or
             </li>
             <li>
 When operating on <a>verifiable credentials</a> or <a>verifiable
-presentations</a> in a software application after verification or validation
+presentations</a> in a software application, after verification or validation
 is performed for securing mechanisms that require an understanding of
-and processing of JSON-LD.
+and/or processing of JSON-LD.
             </li>
             <li>
 When an application chooses to process the media type using the `+json`
--- a/index.html
+++ b/index.html
@@ -4015,11 +4015,11 @@ The rule above guarantees semantic interoperability between JSON and JSON-LD for
 literal JSON keys mapped to URIs by the `@context` mechanism. While JSON-LD
 processors will use the specific mechanism provided and can verify that all
 terms are correctly specified, JSON-based processors implicitly accept the same
-set of terms without testing that they are correct. In other words, the context
-in which the data exchange happens is explicitly stated for both JSON and
-JSON-LD by using the same mechanism. With respect to JSON-based processors, this
-is achieved in a lightweight manner, without having to use JSON-LD processing
-libraries.
+semantics without performing any JSON-LD transformations, but instead by
+applying the above rules. In other words, the context in which the data exchange
+happens is explicitly stated for both JSON and JSON-LD by using the same
+mechanism. With respect to JSON-based processors, this is achieved in a
+lightweight manner, without having to use JSON-LD processing libraries.
           </p>
 
         </section>
--- a/index.html
+++ b/index.html
@@ -3964,7 +3964,7 @@ presentation</a> as JSON is desirable include, but are not limited to:
 
           <ul>
             <li>
-Prior to verification when using the JOSE or COSE specifications to secure a
+Before securing, or after verifying content leveraging <a>external proofs</a> that require <a href="https://csrc.nist.gov/glossary/term/data_integrity">data integrity</a> such as a
 <a>verifiable credential</a> or <a>verifiable presentation</a>.
             </li>
             <li>
--- a/index.html
+++ b/index.html
@@ -3558,7 +3558,7 @@ have at least a draft specification that is being incubated in a community group
             <tr>
               <td>`confidenceMethod`</td>
               <td>
-A property used for specifying one or more methods that a verifier 
+A property used for specifying one or more methods that a verifier
 might use to increase their confidence that the value of an attribute in or of
 a verifiable credential or verifiable presentation is accurate, including but not
 limited to attributes such as `initialRecipient` (a/k/a `issuee`), `presenter`,
@@ -3952,7 +3952,7 @@ system.
           </p>
         </section>
 
-        <section>
+        <section class="informative">
           <h2>JSON Processing</h2>
 
           <p>
--- a/index.html
+++ b/index.html
@@ -3965,8 +3965,8 @@ desirable include, but are not limited to:
 
           <ul>
             <li>
-Before securing, or after verifying content leveraging <a>external proofs</a>
-that require <a href="https://csrc.nist.gov/glossary/term/data_integrity">data
+Before securing or after verifying content
+that requires <a href="https://csrc.nist.gov/glossary/term/data_integrity">data
 integrity</a> such as a
 <a>verifiable credential</a> or <a>verifiable presentation</a>.
             </li>
--- a/index.html
+++ b/index.html
@@ -3972,7 +3972,7 @@ When performing JSON Schema validation, as described in Section
             </li>
             <li>
 When serializing or deserializing <a>verifiable credentials</a> or
-<a>verifiable presentations</a> into databases that index their contents.
+<a>verifiable presentations</a> into systems that store or index their contents.
             </li>
             <li>
 When operating on <a>verifiable credentials</a> or <a>verifiable
--- a/index.html
+++ b/index.html
@@ -3993,8 +3993,8 @@ is desired, an implementer is advised to follow the following rule:
           <ul>
             <li>
 Ensure that all values associated with a `@context` property are in the
-expected order, and the contents of the context files match known good
-cryptographic hashes for each file.
+expected order, the contents of the context files match known good
+cryptographic hashes for each file, and the contents are known to be valid.
             </li>
           </ul>
 
--- a/index.html
+++ b/index.html
@@ -3956,7 +3956,7 @@ system.
           <h2>JSON Processing</h2>
 
           <p>
-While the base media type used by this specification relies on JSON-LD, JSON-LD
+While the media types describing conforming documents defined in this specification always express JSON-LD, JSON-LD
 processing is not required to be performed, since JSON-LD is JSON. Some
 scenarios where processing a <a>verifiable credential</a> or a <a>verifiable
 presentation</a> as JSON is desirable include, but are not limited to:
--- a/index.html
+++ b/index.html
@@ -3956,15 +3956,18 @@ system.
           <h2>JSON Processing</h2>
 
           <p>
-While the media types describing conforming documents defined in this specification always express JSON-LD, JSON-LD
-processing is not required to be performed, since JSON-LD is JSON. Some
-scenarios where processing a <a>verifiable credential</a> or a <a>verifiable
-presentation</a> as JSON is desirable include, but are not limited to:
+While the media types describing conforming documents defined in this
+specification always express JSON-LD, JSON-LD processing is not required to be
+performed, since JSON-LD is JSON. Some scenarios where processing a
+<a>verifiable credential</a> or a <a>verifiable presentation</a> as JSON is
+desirable include, but are not limited to:
           </p>
 
           <ul>
             <li>
-Before securing, or after verifying content leveraging <a>external proofs</a> that require <a href="https://csrc.nist.gov/glossary/term/data_integrity">data integrity</a> such as a
+Before securing, or after verifying content leveraging <a>external proofs</a>
+that require <a href="https://csrc.nist.gov/glossary/term/data_integrity">data
+integrity</a> such as a
 <a>verifiable credential</a> or <a>verifiable presentation</a>.
             </li>
             <li>
@@ -3988,9 +3991,9 @@ structured media type suffix.
           </ul>
 
           <p>
-That is, pure JSON processing is allowed as long as the document being
-consumed or produced is a <a>conforming document</a>. If pure JSON processing
-is desired, an implementer is advised to follow the following rule:
+That is, JSON processing is allowed as long as the document being consumed or
+produced is a <a>conforming document</a>. If JSON processing is desired, an
+implementer is advised to follow the following rule:
           </p>
 
           <ul>
@@ -4009,13 +4012,14 @@ typing, and order, when a JSON document is processed as JSON-LD.
 
           <p>
 The rule above guarantees semantic interoperability between JSON and JSON-LD for
-literal JSON keys mapped to URIs by the `@context` mechanism. While JSON-LD processors will use the
-specific mechanism provided and can verify that all terms are correctly
-specified, JSON-based processors implicitly accept the same set of terms without
-testing that they are correct. In other words, the context in which the data
-exchange happens is explicitly stated for both JSON and JSON-LD by using the
-same mechanism. With respect to JSON-based processors, this is achieved in a
-lightweight manner, without having to use JSON-LD processing libraries.
+literal JSON keys mapped to URIs by the `@context` mechanism. While JSON-LD
+processors will use the specific mechanism provided and can verify that all
+terms are correctly specified, JSON-based processors implicitly accept the same
+set of terms without testing that they are correct. In other words, the context
+in which the data exchange happens is explicitly stated for both JSON and
+JSON-LD by using the same mechanism. With respect to JSON-based processors, this
+is achieved in a lightweight manner, without having to use JSON-LD processing
+libraries.
           </p>
 
         </section>


dlongley (approved):

index.html
```diff
@@ -3752,6 +3752,69 @@ <h3>Media Type Precision</h3>
...
+That is, pure JSON processing is allowed as long as the document being
+consumed or produced is a <a>conforming document</a>. If pure JSON processing
+is desired, an implementer is advised to follow the following rule:
+          </p>
+
+          <ul>
+            <li>
+Ensure that all values associated with a `@context` property are in the
+expected order, and the contents of the context files match known good
+cryptographic hashes for each file.
```
dlongley, line 3797:
```suggestion
expected order, the contents of the context files match known good
cryptographic hashes for each file, and the contents are known to be valid.
```

index.html
```diff
@@ -3752,6 +3752,69 @@ <h3>Media Type Precision</h3>
...
+
+          <p>
+That is, pure JSON processing is allowed as long as the document being
+consumed or produced is a <a>conforming document</a>. If pure JSON processing
+is desired, an implementer is advised to follow the following rule:
+          </p>
+
+          <ul>
+            <li>
+Ensure that all values associated with a `@context` property are in the
```
iherman, line 3795:
```suggestion
consumed or produced is a <a>conforming document</a>. If pure JSON processing
is desired, an implementer MUST follow the following rule:
          </p>

          <ul>
            <li>
Ensure that all values associated with a `@context` property are in the
```

This section is normative. If this is indeed the intention, then this should be made by the usage of _MUST_.

If this is _not_ the case, then of course this text is o.k. as is, but the section must be marked as informative.

msporny: That's fine, I wanted to start the PR out as purely informative and see if folks still desire normative statements on this. This statement was normative in v1.0 and v1.1. We'll see what the other reviewers say.

mprorock:
Isn't this more of just a JSON-LD issue than a VC issue?  I am not sure I see the value of this PR as it is just saying that JSON-LD is JSON

mtaimela:
Thank you @msporny  for the clarification PR, I'm in favour of including this PR as is.

@mprorock wrote:
> Isn't this more of just a JSON-LD issue than a VC issue? I am not sure I see the value of this PR as it is just saying that JSON-LD is JSON

I would say that it is beneficial for the community that is less technical than average contributor. The clarification is very clearly specifying the duality and the differences of json-ld/json in context of VC/VP, while promoting the use of @context. This "promoted" use of the mandatory @context is against pure json mindset, thus this added section is beneficial for the community.

index.html
```diff
@@ -3752,6 +3752,69 @@ <h3>Media Type Precision</h3>
...
+          <p>
+While the base media types for this specification are JSON-LD documents, JSON-LD
+processing is not required to be performed in a variety of scenarios. These
+scenarios include, but are not limited to:
+          </p>
+
+          <ul>
+            <li>
+When using the JOSE or COSE specifications to secure a
+<a>verifiable credential</a> or <a>verifiable presentation</a>.
```
mprorock, line 3767:
```suggestion
Prior to verification when using the JOSE or COSE specifications to secure a
<a>verifiable credential</a> or <a>verifiable presentation</a>.
```

index.html
```diff
@@ -3752,6 +3752,69 @@ <h3>Media Type Precision</h3>
...
+When performing JSON Schema validation.
+            </li>
+            <li>
+When serializing or deserializing <a>verifiable credentials</a> or
+<a>verifiable presentations</a> into databases that index their contents.
+            </li>
+            <li>
+When operating on <a>verifiable credentials</a> or <a>verifiable
+presentations</a> in a software application after verification or validation
+is performed.
```
mprorock, line 3779:
```suggestion
When operating on <a>verifiable credentials</a> or <a>verifiable
presentations</a> in a software application after verification or validation
is performed for securing mechanisms that require an understanding of
and processing of JSON-LD.
```

index.html
```diff
@@ -3752,6 +3752,69 @@ <h3>Media Type Precision</h3>
...
+            <li>
+Ensure that all values associated with a `@context` property are in the
+expected order, the contents of the context files match known good
+cryptographic hashes for each file, and the contents are known to be valid.
+            </li>
+          </ul>
+
+          <p>
+Using static context files and a JSON Schema to ensure proper URL values and
+order is one acceptable approach when implementing the rule above.
```
mprorock, line 3803:
```suggestion
Using static context files with a JSON Schema is one acceptable approach when implementing the rule above. This ensures proper term identification, typing, and order when a JSON document is processed as JSON-LD.
```
or something like that

mprorock (changes requested):

index.html
mprorock: basically just want some language that calls out that for certain securing mechanisms, the JSON-LD is already processed and validated, therefore after those steps have happened, you are free to go on to just json.

msporny: Yes, agreed, we'll get some variation of this in there, it's important to point it out.

index.html
mprorock: There is going to be better language here - I would acually go so far as to say that if you are using JOSE/COSE you don't look at the payload as json or json-ld until verification has taken place.  this means of course that you don't process as json-ld prior to verification.

mprorock: then if you go further with the protected header post verification, and `cty` tells you this thing you just verified is JSON-LD, e.g. via `vc+ld+json`, then you go through and have an option to process according to the content type, so you can process as `ld+json` (and get the glorious semantics` or just as `json` for some other reason (as you noted below re media types).

index.html
```diff
@@ -3752,6 +3752,69 @@ <h3>Media Type Precision</h3>
...
+scenarios include, but are not limited to:
+          </p>
+
+          <ul>
+            <li>
+When using the JOSE or COSE specifications to secure a
+<a>verifiable credential</a> or <a>verifiable presentation</a>.
+            </li>
+            <li>
+When performing JSON Schema validation.
```
decentralgabe, line 3770: could we link to the section that describes this?

decentralgabe (approved):
preemptively approving given a small number of tweaks. great addition!

index.html
TallTed:
```suggestion
Using static context files with a JSON Schema is one acceptable approach to
implementing the rule above. This can ensure proper term identification,
typing, and order, when a JSON document is processed as JSON-LD.
```

index.html
```diff
@@ -3752,6 +3752,69 @@ <h3>Media Type Precision</h3>
...
+          </ul>
+
+          <p>
+Using static context files and a JSON Schema to ensure proper URL values and
+order is one acceptable approach when implementing the rule above.
+          </p>
+
+          <p>
+The rule above guarantees semantic interoperability between JSON and JSON-LD for
+terms defined by the `@context` mechanism. While JSON-LD processors will use the
```
TallTed, line 3808: We continue to say in various places that terms are "defined" by use of `@context`.

This is not accurate.

Rather, the `@context` maps literal JSON keys (which are not properly considered "terms") to URI terms (which are), which may be dereferenced to get their published definitions, so that all readers/consumers/users of those terms can work with the same meaning/definition as intended by their originator/publisher.

index.html
Sakurann: There is a confusion between payload and securing mechanism processing. JSON-LD is needed for the payload, for securing mechanism, JWS library might be enough.

Sakurann (changes requested):
This PR creates more confusion than clarification. One is confusion between payload and securing mechanism processing. Even for the the payload processing, since the base data model is +ld+json, JSON-LD is needed and for example it's misleading to talk about only +json.

index.html
```diff
@@ -3752,6 +3752,69 @@ <h3>Media Type Precision</h3>
...
+While the base media types for this specification are JSON-LD documents, JSON-LD
+processing is not required to be performed in a variety of scenarios. These
+scenarios include, but are not limited to:
+          </p>
+
+          <ul>
+            <li>
+When using the JOSE or COSE specifications to secure a
+<a>verifiable credential</a> or <a>verifiable presentation</a>.
+            </li>
```
Sakurann, line 3768:
```suggestion
```

This is a confusion between payload and securing mechanism processing. this should talk only about the payload which is JSON-LD.

msporny: I have accepted @mprorock's suggestions, does that work for you, @Sakurann? If not, please suggest some language.

index.html
msporny: @iherman wrote:
>  the section must be marked as informative.

Done in d4574f8da5c820c186ff2f87c5dd194ebffcab55.

iherman (approved):

index.html
msporny: > Would it help to add a note here around the inclusion of `@vocab` into the core context making this a bit simpler?

Hmm, good point.

There are a number of things that we've done to make it easier to work w/ the data in a JSON-only mode. Use of `@protected`, use of `@vocab`, aliasing `@id` and `@type`, etc. We do talk about some of these mechanisms: 

* https://w3c.github.io/vc-data-model/#syntactic-sugar
* https://w3c.github.io/vc-data-model/#getting-started

... but as you note, we don't mention "the issuer defined vocabulary" anywhere in the spec, and we probably should. I've raised https://github.com/w3c/vc-data-model/issues/1210 to address this concern. We should probably add it in the "Syntactic Sugar" section.

index.html
msporny: Fixed in 91cb4e6bcc438de69c8d7cebb9e9f8cc96f8276c.

msporny:
@mprorock I took all of your suggestions on this PR, please re-review.

@Sakurann, @mprorock's suggestions (which have been incorporated), addressed the same concerns you raised (AFAICT).

@TallTed requesting a re-review from you just to make sure your changes were incorporated in a way that you felt addressed your concerns.

alenhorvat (approved):

index.html
```diff
@@ -3857,6 +3857,72 @@ <h3>Media Type Precision</h3>
...
+
+          <p>
+The rule above guarantees semantic interoperability between JSON and JSON-LD for
+literal JSON keys mapped to URIs by the `@context` mechanism. While JSON-LD processors will use the
+specific mechanism provided and can verify that all terms are correctly
+specified, JSON-based processors implicitly accept the same set of terms without
+testing that they are correct. In other words, the context in which the data
+exchange happens is explicitly stated for both JSON and JSON-LD by using the
+same mechanism. With respect to JSON-based processors, this is achieved in a
+lightweight manner, without having to use JSON-LD processing libraries.
```
awoie, line 3922: What does semantic interoperability mean in terms of concrete example? IMO, this term is quite overloaded. IMO, there is no semantic interoperability if JSON-only and JSON-LD implementers if they use the default `@vocab` only.

index.html
```diff
@@ -3857,6 +3857,72 @@ <h3>Media Type Precision</h3>
           </p>
         </section>
 
+        <section class="informative">
+          <h2>JSON Processing</h2>
+
+          <p>
+While the base media types for this specification are JSON-LD documents, JSON-LD
+processing is not required to be performed in a variety of scenarios. These
+scenarios include, but are not limited to:
```
awoie, line 3866: IMO, this section is confusing to readers and implementers. I don't think it adds a lot to list steps where JSON-processing is used.

awoie (changes requested):
Some comments

TallTed (approved):
More forward motion

index.html
msporny: Are you saying the list of examples is confusing? The reason it was added was because people wanted to know about scenarios where implementers didn't do JSON-LD processing as highlighted in https://github.com/w3c/vc-data-model/pull/1088#issuecomment-1591275654 

The list are specific examples of when implementers don't do JSON-LD processing today (and just process as JSON). 

What concrete suggestions do you have to improve this text given the above?

index.html
msporny: Yes, good point. It might be best to not use the word and just describe what is achieved.

"semantic interoperability" means that the terms have one meaning whether they are interpreted as JSON-LD or if they're interpreted as JSON (because they use the same mechanism for how the terms are defined). An analogy is using a single source file and executing it via runtime interpretation vs. JIT compiling and running the compiled form. The output of the program in both cases is exactly the same, but how each system interprets the semantics of the source file (runtime interpretation vs. compiled executable) is different. It's the output of the program that matters, and that's entirely dependent on the semantics expressed by the source file being the same between both execution environments. IOW, semantic interoperability is proven when the output of the program, for the same inputs, is indistinguishable between the runtime vs. compiled execution of the program.

I'd argue that there is semantic interoperability between  JSON-only and JSON-LD implementers if they use the default @vocab only... both interpretations will lead to the same semantics... that undefined terms are issuer defined. So, I believe the "semantic interop" statement holds given the definition above... but I'll try re-writing the section to avoid the use of the term given that it has to be explained.

mtaimela (approved):

index.html
mprorock: I keep struggling with this a bit.  As an author of some extensive crawling / search and other functions that work with JSON-LD a lot we have never felt the need to process JSON-LD as other than JSON except in cases where we needed for some reason to operate on that data as RDF.  Could we rephrase this as something like:

```
While the base media type for data types described in this specification
is JSON-LD, JSON-LD processing is not required to be performed, 
since JSON-LD is JSON.  Some scenarios where processing a 
verifiable credential as JSON only is desirable include, but are not limited to:
```

Then, perhaps we could add a corresponding section that talks about when you would want to leverage the graph properties of JSON-LD to get things into RDF.

index.html
```diff
@@ -3857,6 +3857,72 @@ <h3>Media Type Precision</h3>
...
+Prior to verification when using the JOSE or COSE specifications to secure a
+<a>verifiable credential</a> or <a>verifiable presentation</a>.
+            </li>
+            <li>
+When performing JSON Schema validation, as described in Section
+<a href="#data-schemas"></a>.
+            </li>
+            <li>
+When serializing or deserializing <a>verifiable credentials</a> or
+<a>verifiable presentations</a> into databases that index their contents.
```
mprorock, line 3880:
```suggestion
<a>verifiable presentations</a> into systems that store or index their contents.
```

mprorock (changes requested):
definitely heading in the right direction

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-07-25-vcwg-special#section1-4) on 2023-07-25

- no resolutions were taken

#### 1.4. Add section on JSON Processing. (pr vc-data-model#1202)

_See github pull request [vc-data-model#1202](https://github.com/w3c/vc-data-model/pull/1202)._




**Manu Sporny:** 1202 has multiple positive reviews, but has pending requested changes. kristina, will you be able to re-review this week?  

**Kristina Yasuda:** I will try.  

</details>

index.html
```diff
@@ -3857,6 +3857,72 @@ <h3>Media Type Precision</h3>
...
+
+          <p>
+While the base media types for this specification are JSON-LD documents, JSON-LD
+processing is not required to be performed in a variety of scenarios. These
+scenarios include, but are not limited to:
+          </p>
+
+          <ul>
+            <li>
+Prior to verification when using the JOSE or COSE specifications to secure a
```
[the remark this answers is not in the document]
msporny: This sounds like a normative statement (in a non-normative section). The statement above should probably go in the VC-COSE-JOSE specification. The list here is a list of examples of when one might not do full JSON-LD processing. The reasoning/normative statements belong elsewhere.

Can you reword this in some way to address the concern above?

index.html
```diff
@@ -3857,6 +3857,72 @@ <h3>Media Type Precision</h3>
           </p>
         </section>
 
+        <section class="informative">
+          <h2>JSON Processing</h2>
+
+          <p>
+While the base media types for this specification are JSON-LD documents, JSON-LD
```
jandrieu, line 3864:
```suggestion
While the base media type for this specification is a JSON-LD document, JSON-LD
```

We only have one base media type.

brentzundel: we currently have both `vc+ld+json` and `vp+ld+json`

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-07-26-vcwg#section1-3) on 2023-07-26

- no resolutions were taken

#### 1.3. Add section on JSON Processing. (pr vc-data-model#1202)

_See github pull request [vc-data-model#1202](https://github.com/w3c/vc-data-model/pull/1202)._




**Manu Sporny:** For the section on JSON processing, there are a few remaining points. oliver, are you objecting to the proposal or something else?  

**Oliver Terbu:** I have to re-read the pull request. To me it seems rather random and written in a very condensed style.  

**Manu Sporny:** There was a PR by kristina that removed all the statements about JSON processing. There was consensus for this, but others misinterpreted it as us removing support for JSON-only processing.  
… This new PR is to add a condensed list to compensate for the removed sections on JSON processing.  

**Oliver Terbu:** I was just reading 1203 where it was proposed that the transformed document has to follow linked data best practices. This means that VCs have to follow linked data practices, not just plain JSON.  

**Manu Sporny:** The two different PRs address different things. 1203 does not introduce a normative requirement; it is guidance for those transforming data to JSON-LD.  

</details>

msporny:
@mprorock I implemented your request in https://github.com/w3c/vc-data-model/pull/1202#discussion_r1273762313 in commit d4e3d74312e6f480958111ee8f03d6fbd860cf0a. Requesting a re-review from you and @awoie.

index.html
msporny: This has been implemented in https://github.com/w3c/vc-data-model/commit/d4e3d74312e6f480958111ee8f03d6fbd860cf0a .

> perhaps we could add a corresponding section that talks about when you would want to leverage the graph properties of JSON-LD to get things into RDF.

I suggest we handle this in another PR. Please feel free to open a new issue if you'd like a section on that.

index.html
```diff
@@ -3857,6 +3857,73 @@ <h3>Media Type Precision</h3>
           </p>
         </section>
 
+        <section class="informative">
+          <h2>JSON Processing</h2>
+
+          <p>
+While the base media type used by this specification relies on JSON-LD, JSON-LD
+processing is not required to be performed, since JSON-LD is JSON. Some
```
Sakurann, line 3865:
```suggestion
While the base media type used by this specification relies on JSON-LD, and
understanding of JSON-LD is a prerequisite, JSON-LD
processing is not required to be performed in certain cases, since JSON-LD is JSON. Some
```

index.html
```diff
@@ -3857,6 +3857,73 @@ <h3>Media Type Precision</h3>
...
+The rule above guarantees semantic interoperability between JSON and JSON-LD for
+literal JSON keys mapped to URIs by the `@context` mechanism. While JSON-LD processors will use the
+specific mechanism provided and can verify that all terms are correctly
+specified, JSON-based processors implicitly accept the same set of terms without
+testing that they are correct. In other words, the context in which the data
+exchange happens is explicitly stated for both JSON and JSON-LD by using the
+same mechanism. With respect to JSON-based processors, this is achieved in a
+lightweight manner, without having to use JSON-LD processing libraries.
+          </p>
+
```
Sakurann, line 3925:
```suggestion
```

I elaborated in me previous comment above, but saying that `the terms "in the valid JSON-LD document" have one meaning whether they are interpreted as JSON-LD or if they're interpreted as JSON "when they follow certain rules"` is very misleading and I am against us saying so.

and if "pure JSON processing" of a conforming document is allowed (even if under the certain conditions), `@context` should rather be made optional.

Sakurann (changes requested):
the part about "pure JSON processing" is very misleading.

If there is no reason to do JSON-LD processing, and "pure JSON processing" is ok, the data model should be just JSON.

this is like saying "you are using JSON-LD without knowing you are using it but that's ok."

index.html
TallTed: Building off @Sakurann's suggestion —
```suggestion
While the base media type used by this specification relies on JSON-LD, and
understanding of JSON-LD is a prerequisite, JSON-LD processing is not always
required; sometimes processing the JSON-LD as JSON &mdash; of which JSON-LD
is a subtype &mdash; is sufficient. Some
```

index.html
mtaimela: I would see that "pure json" and "required `@context`" can be split into two different topics.

The first, "pure json", can be seen as inherited property/feature of JSON-LD. If you would like to change this, please submit amendment into JSON-LD.

> JSON-LD is designed to be usable directly as JSON, with no knowledge of RDF [[RDF11-CONCEPTS](https://www.w3.org/TR/json-ld11/#bib-rdf11-concepts)].

> A JSON-LD document is always a valid JSON document. This ensures that all of the standard JSON libraries work seamlessly with JSON-LD documents.

The second, "required `@context`", can be seen as VCDM2.0 data structure enforcement on top of underlying data format. This is normal use and compliant with the data formats.

I agree that "pure" is a new description and "vanilla" has also be seen as replacement, but neither work as good as RFC reference.

index.html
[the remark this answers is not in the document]
selfissued: I agree with @mprorock's suggestion.

index.html
```diff
@@ -3857,6 +3857,73 @@ <h3>Media Type Precision</h3>
...
+implementing the rule above. This can ensure proper term identification,
+typing, and order, when a JSON document is processed as JSON-LD.
+          </p>
+
+          <p>
+The rule above guarantees semantic interoperability between JSON and JSON-LD for
+literal JSON keys mapped to URIs by the `@context` mechanism. While JSON-LD processors will use the
+specific mechanism provided and can verify that all terms are correctly
+specified, JSON-based processors implicitly accept the same set of terms without
+testing that they are correct. In other words, the context in which the data
```
dlongley, line 3920: We should clarify that this testing doesn't happen at runtime. The assumption is that someone, at some point, has ensured that the contexts (with well-known hash values) are already valid (this someone could be the same issuer that a verifier trusts, for example). Then anyone can confidently write interoperable software that does not need to perform JSON-LD transformations at runtime.

```suggestion
specified, JSON-based processors implicitly accept the same set of terms without
testing that they are correct at runtime. In other words, the context in which the data
```

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-08-01-vcwg-special#section1-7) on 2023-08-01

- no resolutions were taken

#### 1.7. Add section on JSON Processing. (pr vc-data-model#1202)

_See github pull request [vc-data-model#1202](https://github.com/w3c/vc-data-model/pull/1202)._




**Brent Zundel:** moving to next PR, 1202.  
… Add Section on JSON Processing, all approvals except for one change req from Kristina.  

**Kristina Yasuda:** [.. garbled ] bottom line is, you can achieve the same result via JSON-LD processing.  
… that whole section, the 3-4 paragraphs, makes me nervous.  

**Brent Zundel:** I do see a smaller change suggestion, looks like Ted made a modification to.  

**Manu Sporny:** so there's a number of positive reviews on this. Implementations are doing this today. I believe the things that are stated here are true, since that's the whole point behind JSON-LD, that you can process things this way.  
… it was a very conscious design choice. So, we're just reminding that it's a valid operating mode.  
… that if you do these things, the semantic interpretation is the same, between JSON-LD and JSON.  
… not because of anything in this WG, just because this was how JSON-LD was designed to work in the beginning.  
… so my concern is that your change request just deletes all the statements of how you can process JSON-LD (like current implementations are doing). If we delete that stuff, it guts the section.  
… so I'm a -1 on taking those changes.  
… the changes above that Ted modified seem like it can work, but the wording ends up being a bit more complicated than it needs to be.  
… if you are concerned about this, or it makes you uneasy, concrete change suggestions would be helpful. but deleting core contents of the section is not something that'd go over well.  

**Kristina Yasuda:** is that an option, to move that chunk to the Implementers Doc?  
… the Guidelines doc?  
… and add a sentence below, something like "this is when you can do JSON processing..." with bullet points.  
… and point to the implementers guide.  

**Brent Zundel:** would it work to point to the JSON-LD spec -- I believe it talks about the processing, would it be sufficient to point there, in addition to impl guide?  

**Manu Sporny:** both of these things are possibilities, the reason this PR was raised was because a number of people, when they saw us remove the JSON processing parts of the spec, came to the conclusion that you cannot do JSON processing at all.  
… that the WG was anti-JSON, which led to a whole lot of confusion.  
… I don't think moving this stuff to Implementer Guide, or elsewhere, will help, because people want to know specifically how they can operate.  
… so it should be explicit.  
… we /do/ talk about JSON-only processing in the JSON-LD 1.1 spec, but pointing to a section in the giant spec that somebody has to extract, won't really help.  
… the biggest issue is to clarify - how JSON only docs are processed.  

**Kristina Yasuda:** I don't think anyone should read another giant spec to understand what to do for JSON processing for the VC Data Model. on the other hand, these couple of paragraphs feel like they're a tip of the iceberg too.  
… maybe we should elaborate on the bullet points above this section.  
… I think just saying, under the bullet points, "for details, go see here in the Impl Guide", would be a good compromise.  

**Brent Zundel:** manu, you said this PR was raised to address the confusion folks had, that now the spec couldn't be processed as JSON.  
… would it be sufficient to address that confusion, to just add a brief note, "Just a reminder, JSON-LD can be processed as JSON, many use cases we're aware of do this, for details, see implementation guide".  

**Manu Sporny:** that's not what Alan and Mattheas were asking for; they said it needs to be stated in the core spec.  
… I think moving it to the guide would not address their concerns.  
… others in the group, please chime in. But we should check in with Alan and Mattheas.  
… it's clear that this is something that has confused people for a long time. I don't think the way to clarify is to talk about it for a sentence and then point elsewhere.  
… we can just very clearly say "this is what you do".  
… as far as tip of the iceberg -- I don't know what more we can write. There's just one simple rule.  
… not sure how much more we can elaborate.  

**Kristina Yasuda:** let's take it offline, and let's ask Alan and folks re implementer guide.  

> *Manu Sporny:* +1, thank you Kristina -- if we can add text, that would be good -- deleting text makes me concerned.

**Joe Andrieu:** this sounds good. i wanted to say - we do need to make it clearer / simpler. But I think moving it to implementation guide would be a mistake.  

> *Shigeya Suzuki:* +1, clear and simpler, easy to understand.

**Joe Andrieu:** but I look forward to seeing what Kristina comes up with.  

**Manu Sporny:** +1 Kristina, if we can add text to address your concerns, and still keep it simple and clear, I'd much rather prefer that.  
… we've had enough people get confused.  
… so +1 if you can propose additions/clarifications, happy to process those.  

**Brent Zundel:** thank you all, moving on to another PR.  
… I believe the path forward for this one is - Kristina is going to propose adding a sentence.  

</details>

index.html
```diff
@@ -3857,6 +3857,73 @@ <h3>Media Type Precision</h3>
...
+
+          <p>
+That is, pure JSON processing is allowed as long as the document being
+consumed or produced is a <a>conforming document</a>. If pure JSON processing
+is desired, an implementer is advised to follow the following rule:
+          </p>
+
+          <ul>
+            <li>
+Ensure that all values associated with a `@context` property are in the
```
awoie, line 3903: I really don't understand this section in general. What is the purpose of adding this? How does this help a developer implementing the VCDM 2.0?

awoie (changes requested):

index.html
decentralgabe: `understanding of JSON-LD is a prerequisite` --> `an understanding of JSON-LD is a prerequisite`

mprorock (approved):
Approving so we can move forward

index.html
```diff
@@ -3857,6 +3857,73 @@ <h3>Media Type Precision</h3>
...
+            <li>
+When an application chooses to process the media type using the `+json`
+structured media type suffix.
+            </li>
+          </ul>
+
+          <p>
+That is, pure JSON processing is allowed as long as the document being
+consumed or produced is a <a>conforming document</a>. If pure JSON processing
+is desired, an implementer is advised to follow the following rule:
```
mprorock, line 3898:
```suggestion
That is,  processing as vanilla JSON is allowed as long as the document being
consumed or produced is a <a>conforming document</a>. If vanilla JSON processing
is desired, an implementer is advised to follow the following rule:
```

alenhorvat:
@msporny @Sakurann, thank you for the great discussion. Experts reading this document of course understand JSON and JSON-LD and know that every JSON-LD can be processed as JSON. We learned that this is not true for readers who encounter VCs for the first time.

Text that's been added has been shared with several stakeholders and implementers and they're welcoming the changes since it helps to clarify the JSON processing.

I would kindly ask, if the text can be kept in the document. 

@Sakurann are you proposing to remove most of the text because it is incorrect or misleading?
If so, can we help improving it? We believe community will benefit from such clarification.

index.html
alenhorvat: I believe community would greatly benefit from such explanation.

awoie: I read through the PR many times and I still don't understand why it is actually there. What does this PR really want to express or explain? Because I don't understand what it is.

What I'm worried about is the following, the base media type is based on JSON-LD, and there are probably some situations where a specific constellation of @context entries can make a difference between JSON and JSON-LD processing and in those cases JSON processing is probably not enough.

On the other hand, issuers should implement linked data best practices if they are issuing VCs in the base media type. That includes authoring proper vocabs and context definitions. In both scenarios, verifier and issuer, need to have a decent understanding of how JSON-LD works.

awoie: The base media type is JSON-LD and this implies JSON-LD processing in my opinion. Everything else feels it has the potential to cause a lot of troubles.

awoie: To me, this the JSON processing section is a big foot gun especially as it contradicts what is written in the media types (precision) section(s).

mtaimela: @awoie Could it be possible to be more specific, maybe include the RFC reference?

[JSON-LD 1.1](https://www.w3.org/TR/json-ld11/), the data format you are referring into, declares the following:
> JSON-LD is designed to be usable directly as JSON, with no knowledge of RDF [[RDF11-CONCEPTS](https://www.w3.org/TR/json-ld11/#bib-rdf11-concepts)].

> A JSON-LD document is always a valid JSON document. This ensures that all of the standard JSON libraries work seamlessly with JSON-LD documents.

Information content from IETF [rfc6839](https://www.w3.org/TR/json-ld11/#bib-rfc6839) also defines how the `+json` suffix is used in the `ld+json`.

In my opinion it is a quite a big burden for adoption to require immediately authoring the vocabs and context definitions. These anyway have no meaning if everyone defines their own, the power comes from unification, and that takes time.

dlongley: @awoie,

> That includes authoring proper vocabs and context definitions. In both scenarios, verifier and issuer, need to have a decent understanding of how JSON-LD works.

If an issuer (or creator of a shared vocab, such as the 1EdTech community that works on Open Badges) provides a spec that says: "This is what all these properties mean and what values they can have", then a consuming application can write code against that just like they would with any other spec. What they do need to do, to ensure interoperability with other people in the ecosystem that do want to perform JSON-LD transformations, is follow the rules being expressed here. These rules might be explicitly provided (or now more easily linked to) in that spec or that spec can be designed such that they will be implicitly followed through the requirement of readers of their spec to use JSON schema and hash-checking and / or static context files.

None of this requires "a decent understanding of how JSON-LD works" for these consumers.

index.html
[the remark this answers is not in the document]
OR13: @msporny accept or dismiss

index.html
```diff
@@ -3857,6 +3857,73 @@ <h3>Media Type Precision</h3>
...
+processing is not required to be performed, since JSON-LD is JSON. Some
+scenarios where processing a <a>verifiable credential</a> or a <a>verifiable
+presentation</a> as JSON is desirable include, but are not limited to:
+          </p>
+
+          <ul>
+            <li>
+Prior to verification when using the JOSE or COSE specifications to secure a
+<a>verifiable credential</a> or <a>verifiable presentation</a>.
+            </li>
```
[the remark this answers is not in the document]
OR13: @msporny accept or dismiss

OR13 (changes requested):
Feedback on the PR needs to be accepted to dismissed, so we can see if any items actually remain to be addressed.

OR13:
I filed https://github.com/w3c/vc-data-model/issues/1227 to track documenting the value of "processing the core data model as JSON-LD" per the call yesterday.

My change requests on this PR are limited to accepting or dismissing the open suggestions, and once those are resolved, I intend to approve... We should start the related discussions on the value of processing the core data model as JSON-LD on the issue above, and not continue them on this PR... as its getting very difficult to review.

index.html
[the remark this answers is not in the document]
msporny: Yes, correct.

index.html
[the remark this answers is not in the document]
msporny: I applied @OR13's suggestion in https://github.com/w3c/vc-data-model/pull/1202/commits/3927a47f1fd78d9216a9dda193306fdc2bf2034a, which addresses the concern in a more targeted manner (not using RFC language, but still achieving what @mprorock was requesting).

index.html
msporny: a VP has a structure that is different from VC. Namely, allows the inclusion of `verifiableCredential` ... and allows one to express properties about the VP. Collapsing both things to one concept would ignore that they are different concepts that are to be used for different purposes.

index.html
```diff
@@ -3752,6 +3752,69 @@ <h3>Media Type Precision</h3>
...
+          <p>
+That is, pure JSON processing is allowed as long as the document being
+consumed or produced is a <a>conforming document</a>. If pure JSON processing
+is desired, an implementer is advised to follow the following rule:
+          </p>
+
+          <ul>
+            <li>
+Ensure that all values associated with a `@context` property are in the
+expected order, and the contents of the context files match known good
```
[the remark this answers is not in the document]
msporny: There could be multiple ways this is done, which are application-specific and out of scope for the spec:

* Someone writes the expected order down in a specification somewhere.
* A Verifier only accepts VCs/VPs of a certain structure (based on the `@context` value)
* An issuer provides a JSON Schema that is well known to the Verifier

... and so on. It's done just like any other JSON processing toolchain -- you just have to know what to do through some human readable documentation or tribal knowledge.

index.html
```diff
@@ -3752,6 +3752,69 @@ <h3>Media Type Precision</h3>
...
+cryptographic hashes for each file.
+            </li>
+          </ul>
+
+          <p>
+Using static context files and a JSON Schema to ensure proper URL values and
```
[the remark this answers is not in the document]
msporny: Using JSON Schema alone doesn't save you if the context file values are changing. Granted, if that's happening, you're building on jello, but that's what the "Using static context files" is attempting to convey.

index.html
```diff
@@ -3752,6 +3752,69 @@ <h3>Media Type Precision</h3>
...
+order is one acceptable approach when implementing the rule above.
+          </p>
+
+          <p>
+The rule above guarantees semantic interoperability between JSON and JSON-LD for
+terms defined by the `@context` mechanism. While JSON-LD processors will use the
+specific mechanism provided and can verify that all terms are correctly
+specified, JSON-based processors implicitly accept the same set of terms without
```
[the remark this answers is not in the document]
msporny: JSON doesn't process the contents of `@context` and thus the semantics are implicit (assumed), not explicit (processed).

index.html
```diff
@@ -3752,6 +3752,69 @@ <h3>Media Type Precision</h3>
...
+            </li>
+            <li>
+When an application chooses to process the media type using the `+json`
+structured media type suffix.
+            </li>
+          </ul>
+
+          <p>
+That is, pure JSON processing is allowed as long as the document being
+consumed or produced is a <a>conforming document</a>. If pure JSON processing
```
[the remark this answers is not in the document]
msporny: I have implemented your suggestion in 0e572793cefd18a51bdcb66ca439dba033846855.

We no longer use the word "pure" and now just refer to "JSON" and "JSON-LD".

index.html
msporny: I have implemented @OR13's suggestion in 0e572793cefd18a51bdcb66ca439dba033846855.

We no longer use the word "pure" or "vanilla" and now just refer to "JSON" and "JSON-LD".

dmitrizagidulin (approved):
:+1:

index.html
msporny: Fixed in 02fa6bf19378478aca22ba05fc705fece35b4076.

index.html
msporny: The "pure JSON" language has been removed and replaced with @OR13's suggestion to just speak about "JSON" and "JSON-LD".

msporny: @Sakurann I  am not seeing support to remove the examples (see multiple thumbs ups on @mtaimela's comment) and multiple change suggestions improving the content (not removing it).

index.html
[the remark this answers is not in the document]
msporny: > conforming document here only applies to the registered media types we have and normative statements constraining their members, right?

Yes.

> I feel like this might actually be describing vc+ld+json, since a +json processor would not care about these members being present....

Well, by that definition a +json processor doesn't care about any member being present, 'cause it's JSON and anything that is syntactically valid goes. I would expect that people don't think "any JSON goes" since we do talk about and link to "conforming document".

What we are trying to say here is that "Application-specific JSON processing". If you have an idea of how to say that better, please suggest some language.

index.html
msporny: I tried an alternative here: 698f08efe9db23276424d386d300488cb0342b04.

index.html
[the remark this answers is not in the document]
msporny: Accepted.

index.html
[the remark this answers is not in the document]
msporny: Accepted.

index.html
```diff
@@ -3857,6 +3857,78 @@ <h3>Media Type Precision</h3>
...
+specification always express JSON-LD, JSON-LD processing is not required to be
+performed, since JSON-LD is JSON. Some scenarios where processing a
+<a>verifiable credential</a> or a <a>verifiable presentation</a> as JSON is
+desirable include, but are not limited to:
+          </p>
+
+          <ul>
+            <li>
+Before securing, or after verifying content leveraging <a>external proofs</a>
+that require <a href="https://csrc.nist.gov/glossary/term/data_integrity">data
```
dlongley, line 3874: I think we can simplify this because it's also true for internal proofs -- as the JSON-LD transformations are only needed for some internal proof formats *during* securing or verification, but not before or after:

```suggestion
Before securing or after verifying content
that requires <a href="https://csrc.nist.gov/glossary/term/data_integrity">data
```

msporny:
@OR13 wrote:
> My change requests on this PR are limited to accepting or dismissing the open suggestions, and once those are resolved, I intend to approve... 

I have applied your change requests, and made additional commits to implement your suggestion to remove the "pure JSON" language and clarify what it means to have a "valid context". I believe some of these changes also address issues raised by @mprorock, @decentralgabe, @selfissued, @Sakurann, and @awoie. 

Multiple people have responded to @Sakurann and @awoie to explain sections of this PR and provide support for the current language and opposition to remove language that has been suggested for removal. 

I'm requesting re-reviews from each of you.

OR13 (approved):

index.html
awoie: I didn't disagree with JSON-LD being valid JSON. I was just arguing about the meaningfulness and the justification of JSON-LD in this standard. I'm wondering what is the value of JSON-LD processing and having context normative in that case. It would really help to get answers on that https://github.com/w3c/vc-data-model/issues/1227.

awoie:
> I filed #1227 to track documenting the value of "processing the core data model as JSON-LD" per the call yesterday.
> 
> My change requests on this PR are limited to accepting or dismissing the open suggestions, and once those are resolved, I intend to approve... We should start the related discussions on the value of processing the core data model as JSON-LD on the issue above, and not continue them on this PR... as its getting very difficult to review.

I will no longer block this PR if this is well documented in #1227. If we don't do that, the JSON section just leads to developer confusion.

index.html
msporny: Yes, we will document the benefits of JSON-LD in #1227. 

What we have found is that governments, education institutions, and large enterprises DO define interop vocabularies and have done so for many years. A number of developers creating simple/single systems are rarely exposed to this stuff because they're not in the business of integrating large, complex systems that are used by these large enterprises and large governments, nor are they exposed to having to connect systems between different countries. The goal of the VCDM is to ensure that both simple systems and complex systems are possible to support. To support simple systems, we provide this section (which has always been supported in the VCDM spec) and the base vocabulary so that if people don't want well defined semantics, they just use the base context and stop there. However, that only solves a small part of the problem space -- you can do identity cards, or simple JWT-style token use cases, but you can't scale beyond that level of simplicity. To address the larger enterprise, government, cross-border use cases, you need to interop at a much larger scale with more complex data. This is where formal semantics come in -- again, you don't have to use those with VCDM, but VCDM has to support that mechanism in some way.

In order to do the more complex tasks, where a lot of value can be unlocked, there is a need for formally defined vocabularies so that you know where you can and cannot map data between systems. Some examples are below... starting with education:

Credential Transparency Initiative, which is broadly used in education:

https://credreg.net/ctdl/terms#classes

The 1EdTech (standards setting body for Education) uses JSON-LD vocabularies heavily in their global standards:

https://1edtech.github.io/openbadges-specification/ob_v3p0.html

The Rich Skills Descriptions is another initiative that is used by education:

https://www.openskillsnetwork.org/rsd

In the Supply Chain space, the GS1 Web Vocabulary is used for global trade (using JSON-LD and vocabularies):

https://www.gs1.org/voc/?show=properties

The Traceability Vocabulary is yet another:

https://w3id.org/traceability

... and the list goes on.

awoie: > Yes, we will document the benefits of JSON-LD in #1227.
> 
> What we have found is that governments, education institutions, and large enterprises DO define interop vocabularies and have done so for many years. A number of developers creating simple/single systems are rarely exposed to this stuff because they're not in the business of integrating large, complex systems that are used by these large enterprises and large governments, nor are they exposed to having to connect systems between different countries. The goal of the VCDM is to ensure that both simple systems and complex systems are possible to support. To support simple systems, we provide this section (which has always been supported in the VCDM spec) and the base vocabulary so that if people don't want well defined semantics, they just use the base context and stop there. However, that only solves a small part of the problem space -- you can do identity cards, or simple JWT-style token use cases, but you can't scale beyond that level of simplicity. To address the larger enterprise, government, cross-border use cases, you need to interop at a much larger scale with more complex data. This is where formal semantics come in -- again, you don't have to use those with VCDM, but VCDM has to support that mechanism in some way.
> 
> In order to do the more complex tasks, where a lot of value can be unlocked, there is a need for formally defined vocabularies so that you know where you can and cannot map data between systems. Some examples are below... starting with education:
> 
> Credential Transparency Initiative, which is broadly used in education:
> 
> https://credreg.net/ctdl/terms#classes
> 
> The 1EdTech (standards setting body for Education) uses JSON-LD vocabularies heavily in their global standards:
> 
> https://1edtech.github.io/openbadges-specification/ob_v3p0.html
> 
> The Rich Skills Descriptions is another initiative that is used by education:
> 
> https://www.openskillsnetwork.org/rsd
> 
> In the Supply Chain space, the GS1 Web Vocabulary is used for global trade (using JSON-LD and vocabularies):
> 
> https://www.gs1.org/voc/?show=properties
> 
> The Traceability Vocabulary is yet another:
> 
> https://w3id.org/traceability
> 
> ... and the list goes on.

I'd argue that in cross-border and cross-jurisdictional use cases JSON-LD is not required either. There is evidence that ICAO and ISO did a pretty good job at that in the past. We have Digital Traveler Credentials (DTC) and mobile driver's license (mDL) and also mobile identity documents (mID) all defined in ISO. Also AAMVA extends their non-JSON-LD-based vocab by including additional terms. What I'm asking for is the underlying reasons why "cross-border use cases require JSON-LD and context" require that because apparently those use cases can be solved without JSON-LD and context as well. Also note that I'm totally fine with context and JSON-LD being included but developers need to understand why they are doing what and what are the benefits and implications of the approaches. Let's have that discussion in #1227.

awoie (commented):
I still doubt the usefulness of this section but I'm not blocking it either.

index.html
```diff
@@ -3857,6 +3857,78 @@ <h3>Media Type Precision</h3>
...
+performed, since JSON-LD is JSON. Some scenarios where processing a
+<a>verifiable credential</a> or a <a>verifiable presentation</a> as JSON is
+desirable include, but are not limited to:
+          </p>
+
+          <ul>
+            <li>
+Before securing or after verifying content
+that requires <a href="https://csrc.nist.gov/glossary/term/data_integrity">data
+integrity</a> such as a
```
TallTed, line 3875:
```suggestion
integrity</a>, such as a
```

index.html
```diff
@@ -3857,6 +3857,78 @@ <h3>Media Type Precision</h3>
...
+When performing JSON Schema validation, as described in Section
+<a href="#data-schemas"></a>.
+            </li>
+            <li>
+When serializing or deserializing <a>verifiable credentials</a> or
+<a>verifiable presentations</a> into systems that store or index their contents.
+            </li>
+            <li>
+When operating on <a>verifiable credentials</a> or <a>verifiable
+presentations</a> in a software application after verification or validation
```
TallTed, line 3888:
```suggestion
presentations</a> in a software application, after verification or validation
```

index.html
```diff
@@ -3857,6 +3857,78 @@ <h3>Media Type Precision</h3>
...
+is performed for securing mechanisms that require an understanding of
+and processing of JSON-LD.
```
TallTed, line 3890:
```suggestion
and/or processing of JSON-LD.
```

TallTed (changes requested):

OR13:
Noting conflicts exist, there are new change suggestions, and despite @awoie 's comment here: https://github.com/w3c/vc-data-model/pull/1202#issuecomment-1665464006

He is still marked as blocking the PR... as is @Sakurann .

TallTed (approved):

brentzundel:
@awoie since you have indicated you are no longer blocking, I have dismissed your review.

@Sakurann there have been many changes since your request, please re-review.

brentzundel (approved):
I agree with others that this information will be much better once accompanied by the changes discussed in issue #1227

Sakurann (approved):
Appreciate the work that has gone into this PR. Not blocking. Looking forward to a similar section on JSON-LD processing.

msporny:
Editorial, substantive, multiple reviews, changes requested and made, no remaining objections, merging.

Issue #1009, opened 2023-01-14
Clarifying credential from verifiable credential


David-Chadwick:
The credential exists regardless of whether it has a cryptographic verification mechanism attached to it or not. Various (incompatible) verification methods exist which are being separately standardised. Nevertheless the underlying credential should be identical in all cases. Consequently many of the properties in the VCDM actually refer to the credential and not to the verifiable credential. Examples include: valid from, valid to, id, type, `@context`, issuer.
A PR is needed to ensure that all the properties that describe the credential refer to the credential and not to the verifiableCredential.
By way of example, the current (v1.1) text for identifier states
"The first identifier is for the `verifiable credential` and uses an HTTP-based URL. The second identifier is for the `subject` of the `verifiable credential` (the thing the [claims] are about)"
This should be more correctly stated as
"The first identifier is for the `credential` and uses an HTTP-based URL. The second identifier is for the `subject` of the `credential` (the thing the [claims] are about)"

Note that different proofing mechanisms may add their own `id` for the proof (i.e. for the verifiableCredential), akin to the serialNumber in X.509 public key certificates, and this should not be confused with the `id` of the `credential`.

awoie:
I do agree with @David-Chadwick . I remember that the VCWG made this distinction between credentials and verifiable credentials in the past. A PR to fix these references make sense from my perspective.

Sakurann:
no objection with the direction to clarify.

just a nuance..
> Nevertheless the underlying credential should be identical in all cases.

As defined in v1.1, the actual body of what is signed is different between JWS and LDP - hence transformation section of mapping to JWT claims in section 6.3.1. 
I would agree that `the data model used by a credential should be identical in all cases", though I honestly want a term "data model" to be defined explicitly. right now it is implicit, which I think have contributed to this confusion around data-model -> credential -> verifiable credential transformation..
not sure how much this will impact replacing `verifiable credential` with `credential` - concrete text would be helpful

David-Chadwick:
@Sakurann Since we are now making changes to v2, and not v1.1, then I believe/hope that the credential proofed by either a JWT or JSON-LD will be identical in all cases. i.e. we will adopt the "as well as" algorithm for JWT instead of "instead of". 

I will produce some concrete text for you.

TallTed:
@David-Chadwick -- 

Please edit your initial comment on this issue, and put code fences around `@context`. 

[...]

PR #1211, opened 2023-07-23
Clarify the difference between a "credential" and a "verifiable credential"

msporny:
This PR is an attempt to address #1009 and #1126 by clarifying the difference between a "credential" (an abstract concept) and a "verifiable credential" (a concrete concept). Adjustments are made to normative statements to not talk about "credentials" or "presentations", but rather only "verifiable credentials" and "verifiable presentations" (which MUST be secured).

I expect other adjustments need to be made throughout the spec, and if this PR goes through, we can make those adjustments more easily given consensus around this text being the basis for removing mentions of "credential" or "presentation", when what we meant was "verifiable credential" or "verifiable presentation". Namely, no normative statement in the specification should talk about "credential" or "presentation" if this PR goes through.

What this PR is trying to say is: "If the thing you're talking about isn't secured, it isn't a verifiable credential. A credential is an abstract concept. A verifiable credential is a concrete concept. This specification normatively defines verifiable credential, and that thing MUST be secured."

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1211.html" title="Last updated on Aug 15, 2023, 8:57 PM UTC (01060d7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1211/59a9ca0...01060d7.html" title="Last updated on Aug 15, 2023, 8:57 PM UTC (01060d7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -641,15 +641,15 @@ specification.
       <p class="note"
         title="Abstract concepts vs. concrete concepts">
 Readers might note that some abstract concepts described in this section
-do not have normative requirements or associated media types defined by this
-specification. For example, the abstract concepts of a <a>credential</a> and a
+do not have normative requirements nor associated media types defined by this
+specification. For example, the abstract concepts of <a>credential</a> and
 <a>presentation</a> do not have normative requirements nor associated media
 types. However, the concrete concepts of a <a>verifiable credential</a> or a
 <a>verifiable presentation</a> are defined as <a>conforming documents</a> and
-do have associated media types. Furthermore, what differentiates the
-abstract concepts of <a>credential</a> and <a>presentation</a> from the
-concrete concepts of a <a>verifiable credential</a> and a
-<a>verifiable presentation</a> is the fact that only the "verifiable" objects
+do have associated media types. Furthermore, the only difference between the
+abstract concepts — of <a>credential</a> and <a>presentation</a> — and the
+concrete concepts — of <a>verifiable credential</a> and
+<a>verifiable presentation</a> — is the fact that the "verifiable" objects
 are secured in a way that is cryptographically verifiable. For more details,
 see Section <a href="#securing-verifiable-credentials"></a>.
       </p>
@@ -1824,7 +1824,7 @@ This specification recognizes two classes of securing mechanisms: those that use
 external proofs and those that use embedded proofs. An
 <dfn class="lint-ignore">external proof</dfn> is one that wraps an expression of
 this data model, such as via a JSON Web Token, which is elaborated on in the
-<em>Securing Verifiable Credentials using JSON Web Tokens</em> [[VC-JWT]]
+<em>Securing Verifiable Credentials using JOSE and COSE</em> [[VC-JOSE-COSE]]
 specification. An <dfn>embedded proof</dfn> is a mechanism where the proof is
 included in the data model, such as a Data Integrity Proof, which is elaborated
 on in <em>Verifiable Credential Data Integrity</em> [[VC-DATA-INTEGRITY]].
@@ -1898,8 +1898,8 @@ proof mechanisms, and this specification does not standardize nor recommend any
 single proof mechanisms for use with <a>verifiable credentials</a> or
 <a>verifiable presentations</a>. For more information about the
 <code>proof</code> mechanisms, see the following specifications: Data Integrity
-[[VC-DATA-INTEGRITY]], Securing Verifiable Credentials using JSON Web Tokens
-[VC-JWT], and the "Proofs" section of the Verifiable Credential Specifications
+[[VC-DATA-INTEGRITY]], Securing Verifiable Credentials using JOSE and COSE
+[VC-JOSE-COSE], and the "Proofs" section of the Verifiable Credential Specifications
 Directory [[VC-SPECS]].
         </p>
 
--- a/index.html
+++ b/index.html
@@ -1812,7 +1812,7 @@ T(([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9](\.[0-9]+)?|(24:00:00(\.0+)?))
 
         <p>
 At least one securing mechanism, and the details necessary to evaluate it, MUST
-be expressed for a <a>verifiable credential</a> or a
+be expressed to construct a <a>verifiable credential</a> or a
 <a>verifiable presentation</a>.
         </p>
         <p>
@@ -1845,7 +1845,7 @@ presentations</a> that use an external proof MAY use the <code>proof</code>
 One or more cryptographic proofs that can be used to detect tampering and verify
 the authorship of a <a>verifiable credential</a> or <a>verifiable
 presentation</a>. The specific method used for an <a>embedded proof</a> MUST be
-included using the <code>type</code> <a>property</a>.
+identified using the <code>type</code> <a>property</a>.
           </dd>
         </dl>
 
@@ -1891,9 +1891,9 @@ the signing date. The example below uses Ed25519 digital signatures.
         <p class="note">
 As discussed in Section <a href="#conformance"></a>, there are multiple viable
 proof mechanisms, and this specification does not standardize nor recommend any
-single proof mechanism for use with <a>verifiable credentials</a> or
+single proof mechanisms for use with <a>verifiable credentials</a> or
 <a>verifiable presentations</a>. For more information about the
-<code>proof</code> mechanism, see the following specifications: Data Integrity
+<code>proof</code> mechanisms, see the following specifications: Data Integrity
 [[VC-DATA-INTEGRITY]], Securing Verifiable Credentials using JSON Web Tokens
 [VC-JWT], and the "Proofs" section of the Verifiable Credential Specifications
 Directory [[VC-SPECS]].
--- a/index.html
+++ b/index.html
@@ -1799,39 +1799,40 @@ T(([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9](\.[0-9]+)?|(24:00:00(\.0+)?))
 
         <p>
 At least one securing mechanism, and the details necessary to evaluate it, MUST
-be expressed for a <a>credential</a> or <a>presentation</a> to be a
-<a>verifiable credential</a> or <a>verifiable presentation</a>; that is, to be
-<a>verifiable</a>.
+be expressed for a <a>verifiable credential</a> or a
+<a>verifiable presentation</a>.
         </p>
         <p>
 This specification recognizes two classes of securing mechanisms: those that use
 external proofs and those that use embedded proofs. An
 <dfn class="lint-ignore">external proof</dfn> is one that wraps an expression of
 this data model, such as via a JSON Web Token, which is elaborated on in the
-<em>Securing Verifiable Credentials using JSON Web Tokens</em> [[VC-JWT]] specification.
-An <dfn>embedded proof</dfn> is a mechanism where the proof is included in the
-data model, such as a Data Integrity Proof, which is elaborated on in <em>Verifiable
-Credential Data Integrity</em> [[VC-DATA-INTEGRITY]].
+<em>Securing Verifiable Credentials using JSON Web Tokens</em> [[VC-JWT]]
+specification. An <dfn>embedded proof</dfn> is a mechanism where the proof is
+included in the data model, such as a Data Integrity Proof, which is elaborated
+on in <em>Verifiable Credential Data Integrity</em> [[VC-DATA-INTEGRITY]].
         </p>
         <p>
 It should be noted that these two classes of securing mechanisms are not
 mutually exclusive.
         </p>
         <p>
-Methods of securing <a>credentials</a> or <a>presentations</a> that embed a
-proof in the data model MUST use the <code>proof</code> <a>property</a>.
+Methods of securing <a>verifiable credentials</a> or
+<a>verifiable presentations</a> that embed a proof in the data model MUST use
+the <code>proof</code> <a>property</a>.
         </p>
         <p>
-Methods of securing <a>credentials</a> or <a>presentations</a> that use an
-external proof MAY use the <code>proof</code> <a>property</a>.
+Methods of securing <a>verifiable credentials</a> or <a>verifiable
+presentations</a> that use an external proof MAY use the <code>proof</code>
+<a>property</a>.
         </p>
         <dl>
           <dt><var>proof</var></dt>
           <dd>
-One or more cryptographic proofs that can be used to detect tampering and
-verify the authorship of a <a>credential</a> or <a>presentation</a>. The
-specific method used for an <a>embedded proof</a> MUST be included using the
-<code>type</code> <a>property</a>.
+One or more cryptographic proofs that can be used to detect tampering and verify
+the authorship of a <a>verifiable credential</a> or <a>verifiable
+presentation</a>. The specific method used for an <a>embedded proof</a> MUST be
+included using the <code>type</code> <a>property</a>.
           </dd>
         </dl>
 
@@ -1877,11 +1878,12 @@ the signing date. The example below uses Ed25519 digital signatures.
         <p class="note">
 As discussed in Section <a href="#conformance"></a>, there are multiple viable
 proof mechanisms, and this specification does not standardize nor recommend any
-single proof mechanism for use with <a>verifiable credentials</a>. For more
-information about the <code>proof</code> mechanism, see the following
-specifications: Data Integrity [[VC-DATA-INTEGRITY]], Securing Verifiable
-Credentials using JSON Web Tokens [VC-JWT], and the "Proofs" section of
-the Verifiable Credential Specifications Directory [[VC-SPECS]].
+single proof mechanism for use with <a>verifiable credentials</a> or
+<a>verifiable presentations</a>. For more information about the
+<code>proof</code> mechanism, see the following specifications: Data Integrity
+[[VC-DATA-INTEGRITY]], Securing Verifiable Credentials using JSON Web Tokens
+[VC-JWT], and the "Proofs" section of the Verifiable Credential Specifications
+Directory [[VC-SPECS]].
         </p>
 
       </section>
--- a/index.html
+++ b/index.html
@@ -639,15 +639,19 @@ specification.
       </p>
 
       <p class="note"
-        title="Not every abstract concept is realized in implementations">
-Implementers might note that while there are a number of abstract concepts
-described in this section, such as <a>credential</a> and <a>presentation</a>,
-those concepts are not always concretely realized in implementations. These
-abstract concepts are provided as a mechanism to document a helpful mental model
-to readers. Normative language in this specification is focused on
-<a>verifiable credentials</a> and <a>verifiable presentations</a> and thus not
-every abstract concept described in this section is expected to be realized in
-an implementation.
+        title="Abstract concepts vs. concrete concepts">
+Readers might note that some abstract concepts described in this section
+do not have normative requirements or associated media types defined by this
+specification. For example, the abstract concepts of a <a>credential</a> and a
+<a>presentation</a> do not have normative requirements nor associated media
+types. However, the concrete concepts of a <a>verifiable credential</a> or a
+<a>verifiable presentation</a> are defined as <a>conforming documents</a> and
+do have associated media types. Furthermore, what differentiates the
+abstract concepts of <a>credential</a> and <a>presentation</a> from the
+concrete concepts of a <a>verifiable credential</a> and a
+<a>verifiable presentation</a> is the fact that only the "verifiable" objects
+are secured in a way that is cryptographically verifiable. For more details,
+see Section <a href="#securing-verifiable-credentials"></a>.
       </p>
 
       <section class="informative">
--- a/index.html
+++ b/index.html
@@ -639,19 +639,17 @@ specification.
       </p>
 
       <p class="note"
-        title="Abstract concepts vs. concrete concepts">
-Readers might note that some abstract concepts described in this section
-do not have normative requirements nor associated media types defined by this
-specification. For example, the abstract concepts of <a>credential</a> and
-<a>presentation</a> do not have normative requirements nor associated media
-types. However, the concrete concepts of a <a>verifiable credential</a> or a
-<a>verifiable presentation</a> are defined as <a>conforming documents</a> and
-do have associated media types. Furthermore, the only difference between the
-abstract concepts — of <a>credential</a> and <a>presentation</a> — and the
-concrete concepts — of <a>verifiable credential</a> and
-<a>verifiable presentation</a> — is the fact that the "verifiable" objects
-are secured in a way that is cryptographically verifiable. For more details,
-see Section <a href="#securing-verifiable-credentials"></a>.
+        title="The difference between a credential and a verifiable credential">
+Readers might note that some concepts described in this section, such as
+<a>credential</a> and <a>presentation</a>, do not have media types defined by
+this specification. However, the concepts of a <a>verifiable credential</a> or a
+<a>verifiable presentation</a> are defined as <a>conforming documents</a> and do
+have associated media types. The concrete difference between the
+concepts &mdash; of <a>credential</a> and <a>presentation</a> &mdash; and the
+concepts &mdash; of <a>verifiable credential</a> and <a>verifiable
+presentation</a> &mdash; is the fact that the "verifiable" objects are secured
+in a way that is cryptographically verifiable. For more details, see Section
+<a href="#securing-verifiable-credentials"></a>.
       </p>
 
       <section class="informative">
--- a/index.html
+++ b/index.html
@@ -645,10 +645,10 @@ Readers might note that some concepts described in this section, such as
 this specification. However, the concepts of a <a>verifiable credential</a> or a
 <a>verifiable presentation</a> are defined as <a>conforming documents</a> and do
 have associated media types. The concrete difference between these concepts
-&mdash; between <a>credential</a> and <a>presentation</a> Vs. <a>verifiable
+&mdash; between <a>credential</a> and <a>presentation</a> vs. <a>verifiable
 credential</a> and <a>verifiable presentation</a> &mdash; is simply the fact
 that the "verifiable" objects are secured in a way that is cryptographically
-verifiable, which the others are not. For more details, see Section
+verifiable, and the others are not. For more details, see Section
 <a href="#securing-verifiable-credentials"></a>.
       </p>
 
@@ -1813,9 +1813,9 @@ T(([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9](\.[0-9]+)?|(24:00:00(\.0+)?))
         <h3>Securing Verifiable Credentials</h3>
 
         <p>
-At least one securing mechanism, and the details necessary to evaluate it, MUST
-be expressed to construct a <a>verifiable credential</a> or a
-<a>verifiable presentation</a>.
+A securing mechanism MUST express the details necessary to evaluate whether a
+<a>verifiable credential</a> or <a>verifiable presentation</a> has been secured
+using it.
         </p>
         <p>
 This specification recognizes two classes of securing mechanisms: those that use
--- a/index.html
+++ b/index.html
@@ -633,8 +633,21 @@ information as a valid document.
 
       <p>
 The following sections outline core data model concepts, such as <a>claims</a>,
-<a>credentials</a>, and <a>presentations</a>, which form the foundation of this
+<a>credentials</a>, <a>presentations</a>, <a>verifiable credentials</a>, and
+<a>verifiable presentations</a>, which form the foundation of this
 specification.
+      </p>
+
+      <p class="note"
+        title="Not every abstract concept is realized in implementations">
+Implementers might note that while there are a number of abstract concepts
+described in this section, such as <a>credential</a> and <a>presentation</a>,
+those concepts are not always concretely realized in implementations. These
+abstract concepts are provided as a mechanism to document a helpful mental model
+to readers. Normative language in this specification is focused on
+<a>verifiable credentials</a> and <a>verifiable presentations</a> and thus not
+every abstract concept described in this section is expected to be realized in
+an implementation.
       </p>
 
       <section class="informative">
--- a/index.html
+++ b/index.html
@@ -644,11 +644,11 @@ Readers might note that some concepts described in this section, such as
 <a>credential</a> and <a>presentation</a>, do not have media types defined by
 this specification. However, the concepts of a <a>verifiable credential</a> or a
 <a>verifiable presentation</a> are defined as <a>conforming documents</a> and do
-have associated media types. The concrete difference between the
-concepts &mdash; of <a>credential</a> and <a>presentation</a> &mdash; and the
-concepts &mdash; of <a>verifiable credential</a> and <a>verifiable
-presentation</a> &mdash; is the fact that the "verifiable" objects are secured
-in a way that is cryptographically verifiable. For more details, see Section
+have associated media types. The concrete difference between these concepts
+&mdash; between <a>credential</a> and <a>presentation</a> Vs. <a>verifiable
+credential</a> and <a>verifiable presentation</a> &mdash; is simply the fact
+that the "verifiable" objects are secured in a way that is cryptographically
+verifiable, which the others are not. For more details, see Section
 <a href="#securing-verifiable-credentials"></a>.
       </p>
 


dlongley (approved):

index.html
```diff
@@ -633,10 +633,23 @@ <h2>Core Data Model</h2>
...
+<a>verifiable presentations</a>, which form the foundation of this
 specification.
       </p>
 
+      <p class="note"
+        title="Not every abstract concept is realized in implementations">
+Implementers might note that while there are a number of abstract concepts
+described in this section, such as <a>credential</a> and <a>presentation</a>,
+those concepts are not always concretely realized in implementations. These
+abstract concepts are provided as a mechanism to document a helpful mental model
```
David-Chadwick, line 646: Sorry but I do not see how the abstract concepts of credential and presentation provide a helpful mental model. Precisely what are these abstract concepts, and why are they different from the concept of a verifiable credential or verifiable presentation. 
OTOH, I can see that the abstract concepts of credential and presentation are imaginary tokens that may or may not be verifiable, but by adding the adjective verifiable to them, we can see that they are verifiable and are therefore a subclass of the more general token.

David-Chadwick: Perhaps it would be better to simply purge all mention of the terms credential and presentation and use verifiableCredential and verifiablePresentation throughout?

TallTed:
> only "verifiable credentials" and "verifiable presentations" (which MUST be secured)

This is problematic. I submit, again, that "verifiable credentials" and "verifiable presentations" MUST be secur*able* but not necessarily secur*ed*.

The verifiability is the ability to perform a test, that being the verification process.

It does not matter whether this process succeeds or not, nor whether a proof or related mechanism _is_ attached. It matters whether a proof or related mechanism *can be* attached and used, in a standardized manner.

I will read the PR a bit later, and respond to the actual content, rather than focusing on the introductory meta-content. :-)

index.html
```diff
@@ -1799,39 +1812,40 @@ <h3>Securing Verifiable Credentials</h3>
 
         <p>
 At least one securing mechanism, and the details necessary to evaluate it, MUST
-be expressed for a <a>credential</a> or <a>presentation</a> to be a
-<a>verifiable credential</a> or <a>verifiable presentation</a>; that is, to be
-<a>verifiable</a>.
+be expressed for a <a>verifiable credential</a> or a
```
TallTed, line 1815:
```suggestion
be expressed to construct a <a>verifiable credential</a> or a
```

index.html
```diff
@@ -1799,39 +1812,40 @@ <h3>Securing Verifiable Credentials</h3>
...
           <dt><var>proof</var></dt>
           <dd>
-One or more cryptographic proofs that can be used to detect tampering and
-verify the authorship of a <a>credential</a> or <a>presentation</a>. The
-specific method used for an <a>embedded proof</a> MUST be included using the
-<code>type</code> <a>property</a>.
+One or more cryptographic proofs that can be used to detect tampering and verify
+the authorship of a <a>verifiable credential</a> or <a>verifiable
+presentation</a>. The specific method used for an <a>embedded proof</a> MUST be
+included using the <code>type</code> <a>property</a>.
```
TallTed, line 1848:
```suggestion
identified using the <code>type</code> <a>property</a>.
```

index.html
```diff
@@ -1877,11 +1891,12 @@ <h3>Securing Verifiable Credentials</h3>
         <p class="note">
 As discussed in Section <a href="#conformance"></a>, there are multiple viable
 proof mechanisms, and this specification does not standardize nor recommend any
-single proof mechanism for use with <a>verifiable credentials</a>. For more
-information about the <code>proof</code> mechanism, see the following
-specifications: Data Integrity [[VC-DATA-INTEGRITY]], Securing Verifiable
-Credentials using JSON Web Tokens [VC-JWT], and the "Proofs" section of
-the Verifiable Credential Specifications Directory [[VC-SPECS]].
+single proof mechanism for use with <a>verifiable credentials</a> or
```
TallTed, line 1894:
```suggestion
single proof mechanisms for use with <a>verifiable credentials</a> or
```

index.html
```diff
@@ -1877,11 +1891,12 @@ <h3>Securing Verifiable Credentials</h3>
...
+<a>verifiable presentations</a>. For more information about the
+<code>proof</code> mechanism, see the following specifications: Data Integrity
```
TallTed, line 1896:
```suggestion
<code>proof</code> mechanisms, see the following specifications: Data Integrity
```

index.html
```diff
@@ -1799,39 +1812,40 @@ <h3>Securing Verifiable Credentials</h3>
 
         <p>
 At least one securing mechanism, and the details necessary to evaluate it, MUST
-be expressed for a <a>credential</a> or <a>presentation</a> to be a
-<a>verifiable credential</a> or <a>verifiable presentation</a>; that is, to be
-<a>verifiable</a>.
+be expressed for a <a>verifiable credential</a> or a
+<a>verifiable presentation</a>.
         </p>
         <p>
```
TallTed, line 1818:
```suggestion
      <p class="note"
        title="Being verifiable does not require successful verification">
"Verifiable credentials" and "verifiable presentations" MUST be securable, but MAY be unsecured. The point of this verifiability is the ability to test, via the verification process, whether the "verifiable credential" or "verifiable presentation" was secured, and, if so, whether its content has been tampered with. It does not matter whether this process succeeds or not, nor whether a proof or related mechanism is actually attached; it only matters whether the document is structured such that a proof or related mechanism _can be_ attached and used, as described by this specification.
        </p>
        <p>
```

TallTed (changes requested):
Some small stuff, and a key additional NOTE.

index.html
msporny: Hrm, I think this interpretation is what's tripping us up. IOW, this interpretation seems to be what @OR13 keeps objecting to. What I'm trying to do in this PR is make it clear that if you want to call something a VC/VP, it MUST be secured. 

What about something like:

```suggestion
      <p class="note"
        title="Being verifiable does not require successful verification">
"Verifiable credentials" and "verifiable presentations" MUST be secured. The point of this verifiability is the ability to test, via the verification process, whether the "verifiable credential" or "verifiable presentation" was secured, and, if so, whether its content has been tampered with. It does not matter whether this process succeeds or not; it only matters that the document was secured in some way where the verification process can be executed.
        </p>
        <p>
```

index.html
msporny: > Sorry but I do not see how the abstract concepts of credential and presentation provide a helpful mental model. Precisely what are these abstract concepts, and why are they different from the concept of a verifiable credential or verifiable presentation.

This section attempts to explain that. They are abstract in that the thing isn't realized in an implementation... just like the concept of an "entity" or "repository" or a "subject" is not realized in an implementation. They are abstract (in the computer science sense of the word) because while they provide a way to think about the data model... they're not directly realized as instances. 

So, the concept of a "credential" is something that's not secured... and we don't have any normative statements about those things in this specification. The only normative statements we have are around "verifiable credentials"... because making normative statements around "credentials" doesn't really buy us anything useful.

> Perhaps it would be better to simply purge all mention of the terms credential and presentation and use verifiableCredential and verifiablePresentation throughout?

There are 504 occurrences of the word "credential" in the specification, and 79 instances of the word "presentation" in the specification. I tried going through and adding "verifiable" to the beginning of the 583 occurrences and stopped at about 100+ into the effort. Many of the sentences that we have written about just "credentials" really are about just "credentials"... not "verifiable credentials". We distinguish between the two carefully in the specification. IOW, I tried doing what you said above and it didn't go well. 

There is a difference between a credential (not secured) and a verifiable credential (secured)... and it's useful to make that distinction in the specification (and we do, in lots of places).

David-Chadwick: I agree that there is a difference a credential (not secured) and a verifiable credential (secured) so we should describe what this difference is in the specification. You said "we don't have any normative statements about those things (i.e. credentials) in this specification". Maybe this is where we are going wrong? Because we can say that a credential MUST have a subject object and an issuer object and an `@context` and a verifiableCredential MUST have a type, an ID, an internal or external proof and MAY have validFrom, validUntil, ToU, etc. We can easily make the concept of credential concrete if we want to, and I think this will strengthen the model and the description.

index.html
David-Chadwick: "Verifiable credentials" and "verifiable presentations" MUST be secured. 
+1 to this. We have already said that this securing mechanism differentiates the credential from the verifiable credential.

TallTed: >> "Verifiable credentials" and "verifiable presentations" MUST be secured.

No. Very much, no.

> We have already said that this securing mechanism differentiates the credential from the verifiable credential.

Yes, the securing mechanism, which works out to the ability to attach a proof (internally or externally) and relevant details/data, which can then be used to test whether the payload (the collection of claims) is as issued or has been tampered with.

But that test does *not* need to succeed.

Real world example: US Birth and Death Certificates *may* be certified but this is not a requirement of all copies thereof. Many uses of these documents require a certified copy but many other uses do *not* require the certification. The difference between them typically amounts to the certified (a/k/a secured) a stamp (akin to if not actually a notary stamp), usually a signature, sometimes some special stationery. Photocopies of the certified have *exactly* the same information, but do not have an original signature, nor the stamp, nor the stationery.

msporny: I think we're getting tripped up on what it means to be "secured"... I'm finding myself agreeing w/ both @David-Chadwick and @TallTed -- which might mean that this is a term/definition problem (which is true for many disagreements). :)

I think we agree? maybe? That in order for a "credential" to be a "verifiable credential", it MUST be secured using a securing mechanism. Simultaneously, we are NOT saying that verification of the securing mechanism has to succeed... we're just saying: "It has to be there in a way that a known verification process can be run against it."

I think some in the group are objecting to the notion that we can call something a "verifiable credential" when it isn't secured. So, those folks are drawing a distinction between "to be able to be secured" and "is secured, but might not verify" -- I think the group is pushing for the latter. I don't know if you, @TallTed, are pushing for the former?

In any case, just an observation that it's the "secured" language that might be tripping up consensus on this item?

TallTed: > I don't know if you, @TallTed, are pushing for ["to be able to be secured"]?

Yes, I am.

A door with a lock that is not locked is still a lockable door. Once locked, the lock might require a code on a keypad (internal) or a physical key/card (external) to be unlocked. It is typically obvious upon cursory inspection which of these is needed, and so the entity desiring entry knows what next-steps to take.

According to the VCDM, a verifiable {credential|presentation} must conform to a defined structure (Compacted JSON-LD), which must include two objects: a `@context` and a `@type`. Everything else is optional — including `proof`, because an external mechanism is allowed; an internal mechanism is not required.

Now, if you leave out `proof` because you intend to use an external mechanism, and then you fail — for any reason — in your use of that external mechanism, verification *will* fail — as intended and designed!  Verifiability is not lost; rather, this failure is confirmation that verifiability is working!

I *might* be convinced to require declaration of securing mechanism — at least when `proof` is not used for an internal mechanism, i.e., when an external mechanism (and proof material) is intended — as this would narrow the testing which a verifier needs to perform.

It seems to me that failure to attach/declare an external securing mechanism *and* failure to fill in `proof` serves as an obvious and quick verification failure.  But the other data in the VP/VC *might* still be considered *valid* in some scenarios.... This is up to the verifier and their business rules.

msporny: > I might be convinced to require declaration of securing mechanism — at least when proof is not used for an internal mechanism, i.e., when an external mechanism (and proof material) is intended — as this would narrow the testing which a verifier needs to perform.

Can you elaborate on how you would see that working?

TallTed: From the current draft:
`At least one securing mechanism, and the details necessary to evaluate it, MUST
be expressed to construct a <a>verifiable credential</a> or a
<a>verifiable presentation</a>.`

It seems it's already expected to be working... for what is "expressing" a securing mechanism other than declaring it? `alg=none` might be the only detail the verifier needs to evaluate the mechanism according to *their* business logic.

But I stand by my assertion that, for instance, `ROT13` (or heck, `ROT0`) is a securing mechanism which can be expressed with no additional details necessary for its evaluation, and can be used to construct a VC or VP.

dlongley: Perhaps we can just make a statement like this: "A <a>verifiable credential</a> without a securing mechanism (neither internal nor external), MUST (will?) fail verification."?

TallTed: I could probably be convinced to something like that (reserving some language tweaks).

What does it do with `alg=none` (or whatever the closest parallel would be)? Is a mechanism with a weak (or `NULL`) key (or password) considered to be present or absent?

msporny: > I could probably be convinced to something like that (reserving some language tweaks).
> 
> What does it do with `alg=none` (or whatever the closest parallel would be)? Is a mechanism with a weak (or `NULL`) key (or password) considered to be present or absent?

Absent. `alg=none` has no place in the VC Ecosystem.

TallTed: That would seem to cover a NULL key or password.

What about a weak key, password, mechanism, etc? How is this (particularly mechanism, where they may be broken after VC issuance and possibly after demise of VC issuer -- so reissue/update is not possible) to be monitored/maintained?

msporny: > What about a weak key, password, mechanism, etc? How is this (particularly mechanism, where they may be broken after VC issuance and possibly after demise of VC issuer -- so reissue/update is not possible) to be monitored/maintained?

The topic above is important, and is probably something for the Security Considerations section. We kinda-sorta talk about it here, but clearly (given the above) the section could be improved:

https://w3c.github.io/vc-data-model/#cryptography-suites-and-libraries

... but "solving it" in this PR is deviating from the purpose of the PR, IMHO. Would you be ok w/ raising an issue on this topic? It requires discussion before we try to write anything into the spec.

David-Chadwick:
"is secured, but might not verify" - this is certainly the semantic that I support. If it is not secured then it cannot be a verifiable credential, but only an unverifiable credential. 
 Note that a verifiable credential might not verify, but that makes it an unverifiable verifiable credential, which is different to a an unverifiable credential.

index.html
David-Chadwick: One of the purposes of this PR was to differentiate between a credential and a verifiableCredential, so agreeing on what a verifiableCredential is, is critical to this PR.

brentzundel (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-08-01-vcwg-special#section1-6) on 2023-08-01

- no resolutions were taken

#### 1.6. Clarify that some abstract concepts are not realized in implementations (pr vc-data-model#1211)

_See github pull request [vc-data-model#1211](https://github.com/w3c/vc-data-model/pull/1211)._




**Brent Zundel:** next PR is 1211.  
… Clarify that some abstract concepts are not realized in implementations.  
… this has a couple approvals, decent conversations, one request for changes from Ted.  
… Ted, has that been addressed?  

**Ted Thibodeau Jr.:** I have to read it over again.  

**Brent Zundel:** I reviewed and approved, and support Manu's suggestion that perhaps Ted, perhaps your recommended change could be discussed in a separate issue.  

**Ted Thibodeau Jr.:** yeah, I think that's the direction we're going. I'll open the new issue.  

**Brent Zundel:** thanks. any other comments?  

**Manu Sporny:** I hesitated to merge this, I didn't know what David Chadwick was going for.  
… waiting for specific change requests.  

**Brent Zundel:** I understood his comment to mean he was agreeing with things above.  

**Manu Sporny:** he also said "one of the purposes of this PR was to differentiate between a VC and a VP", so agreeing on the difference is crucial.  
… does that mean he's saying we haven't decided, or..?  

**Brent Zundel:** that's fine, if David objects, we'll handle those cases in our workflow.  

</details>

David-Chadwick:
I am happy with definition of verifiableCredential as a secured credential, but it might not verify. The problem I continue to have is with the definition of credential. I still believe that it is a different object to a verifiableCredential, and should be defined as such i.e. a verifiableCredential without the verification proof.

index.html
```diff
@@ -1877,11 +1891,12 @@ <h3>Securing Verifiable Credentials</h3>
...
-information about the <code>proof</code> mechanism, see the following
-specifications: Data Integrity [[VC-DATA-INTEGRITY]], Securing Verifiable
-Credentials using JSON Web Tokens [VC-JWT], and the "Proofs" section of
-the Verifiable Credential Specifications Directory [[VC-SPECS]].
+single proof mechanisms for use with <a>verifiable credentials</a> or
+<a>verifiable presentations</a>. For more information about the
+<code>proof</code> mechanisms, see the following specifications: Data Integrity
+[[VC-DATA-INTEGRITY]], Securing Verifiable Credentials using JSON Web Tokens
+[VC-JWT], and the "Proofs" section of the Verifiable Credential Specifications
+Directory [[VC-SPECS]].
```
decentralgabe, line 1899: I'd like to note that vc-jose-cose will be added to [a different section](https://github.com/w3c/vc-specs-dir/pull/14) in the specs dir, so this language should probably be more generic....or the specs dir should be reorganized to include all embedded and non embedded securing mechanisms in a similar section cc: @OR13

TallTed:
> [@David-Chadwick] If it is not secured then it cannot be a verifiableCredential, but only an unverifiable credential.

On the contrary, if it is not secured then it cannot be _successfully verified_.

If it conforms to the structure we have defined for a verifiableCredential, including either having an internal `proof` property (which value might be gibberish or even `NULL`) or being encapsulated in a securable envelope (e.g., JWK, which might have `alg=none`), then it is a verifiableCredential which will never pass verification. That is, it can *fail* verification, which failure is a function of its verifiability.

There is nothing wrong with this!

I can have a credential, such as an uncertified birth certificate (lacking original signature, clerk's imprint, special stationery, etc.) which contains all the same information as another credential, such as a certified birth certificate (including original signature, clerk's imprint, special stationery, etc.). The first will never *verify* but may still be sufficient to be considered *valid* by the verifier, and by conforming to our data model, the data is more useful and usable than if it did not conform to our model.

index.html
```diff
@@ -633,10 +633,23 @@ <h2>Core Data Model</h2>
...
       <p>
 The following sections outline core data model concepts, such as <a>claims</a>,
-<a>credentials</a>, and <a>presentations</a>, which form the foundation of this
+<a>credentials</a>, <a>presentations</a>, <a>verifiable credentials</a>, and
+<a>verifiable presentations</a>, which form the foundation of this
 specification.
       </p>
 
+      <p class="note"
+        title="Not every abstract concept is realized in implementations">
```
OR13, line 642: I don't feel this language captures consensus, or is accessible / comprehensible enough... but It think it might be possible to get to where we need to be, if we refer to media types, and normative requirements as they related to the vocabulary and context... which are normative.

index.html
OR13: imo, sorta blocked by https://github.com/w3c/vc-specs-dir/pull/14

OR13 (changes requested):
I think a version of this PR could be successful:

1. reference media types
2. reference normative requirements and explicitly address "securing / proofs" and align with what RDF and vocabulary say.

index.html
msporny: Please provide some concrete language.

I will note that this section explicitly talks to a conceptual data model and not a serialization, which is where media types come in. Keep that in mind when proposing concrete text.

OR13: Possible language:

This specification does not define `credential` or `presentation` as conforming documents, or provide media types for expressing them. 

This specification does define `verifiable credential` and `verifiable presentation` as conforming documents, and provides media types for expressing them. 

There is no testable difference between a `credential` and `verifiable credential`, however we assert that at least one securing mechanism and the details necessary to evaluate it MUST be present to call a conforming document a verifiable credential.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-08-09-vcwg#section2-4) on 2023-08-09

- no resolutions were taken

#### 2.4. Clarify that some abstract concepts are not realized in implementations (pr vc-data-model#1211)

_See github pull request [vc-data-model#1211](https://github.com/w3c/vc-data-model/pull/1211)._




**Manu Sporny:** PR 1211 "abstract concepts" trying to get agreement with Orie and David Chadwick. Media types...  

**Orie Steele:** "testable difference" between the two concepts... Feeling mushy.  
… media type distinguishable, RDF, english sentence, securing mechanism, duration of proof vs information.  
… not sure how to fix.  

**Manu Sporny:** Will propose concrete text. One is secure, one is not, or ability to be secured; Can door be locked, does it have a door...  
… Three interpretations floating around and they don't line up. Depending on the securing mechanism may has "proof" on it. JWT and SD-JWT don't have proof on them.  
… okay to have prose that says VC must be secured. We can test it... Concrete suggestion/testable. Can Ted/Orie/DavidC agree?  

> *Orie Steele:* I'm not sure its testable, but I hope it can be made testable.

> *Orie Steele:* perhaps it will become more obvious with better text.

**Manu Sporny:** its absolutely testable. Run a verification algorithm. Know which algorithm to run.  

> *Andres Uribe:* You can run algorithms on anything, which is why that test is problematic IMHO.

**Sebastian Crane:** Analogy -- Safe is secure, sledge hammer breaks safe, transfer to VC world, if alg is later insecure it still was a VC.  

**Michael Jones:** agree with Manu. If crypto signed verifiable if not no. That is it.  

> *Manu Sporny:* yes, +1000 selfissued !

> *Joe Andrieu:* +1 to that language.

**Manu Sporny:** plus 1000 to Mike Jones.  

> *Phillip Long:* +1 to selfissued.

> *Greg Bernstein:* +1.

> *Dave Longley:* +1 if this doesn't trigger a bunch of other changes, i'm happy with that solution.

> *Kristina Yasuda:* why can't we clarify it's "cryptographicaly verifiable" credential.

> *Manu Sporny:* that might work as well, kristina.

> *Dave Longley:* maybe we can add "cryptographically" before "verifiable credential" to solve this like kristina said.

> *Manu Sporny:* depends on how far down the rabbit hole we want to go.

**TalTed:** if it remains that loose, I can live with it. If the crypto needs to be at a certain level that is different.  

**Orie Steele:** agree, the abstract concept of "proof", versus the terminology "proof" in our defs.  
… in summary keeping it higher level and not using word proof...  

**Michael Jones:** agree, keep it simple, omit use of word proof.  

> *Sebastian Crane:* +1 JoeAndrieu (IRC).

**Joe Andrieu:** on vetting securing mechanisms, that is us (W3C).  

> *Ted Thibodeau Jr.:* -1 we're not sufficiently crypto expert to perform this vetting, especially not for the indeterminate futurr.

> *Brent Zundel:* we are absolutely not qualified to vet all possible VC securing mechanisms.

> *Orie Steele:* in my experience we are *not* careful enough to not confuse "abstract proof" with "data integrity proof".

> *Orie Steele:* and this issue is a result of that.

**Manu Sporny:** on same path until we had to remove the word proof, concern that we need to replace every use of the word proof...  

> *Dave Longley:* Orie: i'm actually not sure if the outcome of this discussion changes anything practical.

> *Brent Zundel:* we were on the right path until the suggestion was made that the VCWG needs to bless every possible securing mechanism.

**Manu Sporny:** we have been very careful on the use of the language and the word "proof".  

> *Orie Steele:* dlongley I am in the same boat... that is why its so concerning.

> *Michael Jones:* +1.

> *Dave Longley:* Orie: I'm actually *not* concerned for that reason :).

> *Orie Steele:* It's ok for me to be.

**Sebastian Crane:** agree with Manu's direction. This WG *is* qualified to say something about cryptography...  

> *Orie Steele:* "in the rough" on the proof language...

> *Kristina Yasuda:* we have a section titled "Proofs (signature)" so I think we can safely add clarification that "VC is cryptographically verifiable/secured" without affecting proof terms.

**Ted Thibodeau Jr.:** Sebastian this WG has an end date. Cannot make any guarantees on on going crypto...  

> *Kristina Yasuda:* this sentence is why no need to change anything 'proof' while adding 'cryptographically verifiable': "The cryptographic mechanism used to prove that the information in a verifiable credential or verifiable presentation was not tampered with is called a proof. ".

> *Manu Sporny:* Yes, to some variation of what Kristina said above ^.

> *Orie Steele:* most of my comments regarding use of the "word proof" are based on the numerous conversations tallted and I have had on many issues regarding this.

**Ted Thibodeau Jr.:** on lockable/locked door.  

> *Manu Sporny:* ... and I feel like we already say that in the spec.

**Ted Thibodeau Jr.:** key left hanging on the door...  

> *Dmitri Zagidulin:* @TallTed - in your analogy, the question is "is it the door with a lock?" and YES, it still is, even if key is hanging outside.

> *Brent Zundel:* I'm now officially lost on the locked door analogies.

> *Orie Steele:* FWIW I agree there is a way to make it clear when we are being abstract, and when we are being concrete... my only concern is that when we are concrete we are consistent in both RDF and english.

**Joe Andrieu:** yes, to Manu's direction. Let's get to specifics... To Ted the door is securable.  

> *Ted Thibodeau Jr.:* YES, securABLE, not necessarily securED!

> *Orie Steele:* This is the primary place where we are being messy: [https://w3c.github.io/vc-data-model/#securing-verifiable-credentials](https://w3c.github.io/vc-data-model/#securing-verifiable-credentials).

> *Dmitri Zagidulin:* @Orie - interesting. any particular place in that section? Looking through it, it seems pretty clear..

</details>

index.html
msporny: I have reworked the language to specify that the thing that differentiates a "credential" from a "verifiable credential" is the fact that the latter is secured in a way that is cryptographically verifiable, and the former is not. Requesting re-reviews from @TallTed @David-Chadwick and @OR13.

msporny:
@OR13 wrote:
> I think a version of this PR could be successful:
> 
> 1. reference media types
> 2. reference normative requirements and explicitly address "securing / proofs"

I believe I have applied a variation of the wording you provided in a way that I hope you (@OR13), @TallTed, and @David-Chadwick  find acceptable. I expect it'll still need some word smithing, but the hope is that it strikes the right balance.

Re-requesting reviews now.

index.html
```diff
@@ -1799,39 +1816,40 @@ <h3>Securing Verifiable Credentials</h3>
...
         <p>
 This specification recognizes two classes of securing mechanisms: those that use
 external proofs and those that use embedded proofs. An
 <dfn class="lint-ignore">external proof</dfn> is one that wraps an expression of
 this data model, such as via a JSON Web Token, which is elaborated on in the
-<em>Securing Verifiable Credentials using JSON Web Tokens</em> [[VC-JWT]] specification.
-An <dfn>embedded proof</dfn> is a mechanism where the proof is included in the
-data model, such as a Data Integrity Proof, which is elaborated on in <em>Verifiable
-Credential Data Integrity</em> [[VC-DATA-INTEGRITY]].
+<em>Securing Verifiable Credentials using JSON Web Tokens</em> [[VC-JWT]]
```
OR13, line 1827: For another PR, but perhaps we can make these also references to the spec dir... Anticipating additional embedded or external proofs in the future.

OR13 (approved):
It will still need some cleanup, but I think the explicit language around media types helped a lot... We might also be able to use the conforming document language instead, in a future editorial pass.

index.html
TallTed: If your rework is lines 641-654 above, I'm OK with it.

index.html
```diff
@@ -633,10 +633,27 @@ <h2>Core Data Model</h2>
...
+<a>verifiable presentations</a>, which form the foundation of this
 specification.
       </p>
 
+      <p class="note"
+        title="Abstract concepts vs. concrete concepts">
+Readers might note that some abstract concepts described in this section
+do not have normative requirements or associated media types defined by this
+specification. For example, the abstract concepts of a <a>credential</a> and a
+<a>presentation</a> do not have normative requirements nor associated media
```
TallTed, line 646:
```suggestion
specification. For example, the abstract concepts of <a>credential</a> and
<a>presentation</a> do not have normative requirements nor associated media
```

index.html
```diff
@@ -633,10 +633,27 @@ <h2>Core Data Model</h2>
...
-<a>credentials</a>, and <a>presentations</a>, which form the foundation of this
+<a>credentials</a>, <a>presentations</a>, <a>verifiable credentials</a>, and
+<a>verifiable presentations</a>, which form the foundation of this
 specification.
       </p>
 
+      <p class="note"
+        title="Abstract concepts vs. concrete concepts">
+Readers might note that some abstract concepts described in this section
+do not have normative requirements or associated media types defined by this
```
TallTed, line 644:
```suggestion
do not have normative requirements nor associated media types defined by this
```

index.html
```diff
@@ -633,10 +633,27 @@ <h2>Core Data Model</h2>
...
+specification. For example, the abstract concepts of a <a>credential</a> and a
+<a>presentation</a> do not have normative requirements nor associated media
+types. However, the concrete concepts of a <a>verifiable credential</a> or a
+<a>verifiable presentation</a> are defined as <a>conforming documents</a> and
+do have associated media types. Furthermore, what differentiates the
+abstract concepts of <a>credential</a> and <a>presentation</a> from the
+concrete concepts of a <a>verifiable credential</a> and a
+<a>verifiable presentation</a> is the fact that only the "verifiable" objects
+are secured in a way that is cryptographically verifiable. For more details,
+see Section <a href="#securing-verifiable-credentials"></a>.
```
TallTed, line 654:
```suggestion
do have associated media types. Furthermore, the only difference between the
abstract concepts — of <a>credential</a> and <a>presentation</a> — and the
concrete concepts — of <a>verifiable credential</a> and
<a>verifiable presentation</a> — is the fact that the "verifiable" objects
are secured in a way that is cryptographically verifiable. For more details,
see Section <a href="#securing-verifiable-credentials"></a>.
```

index.html
TallTed, line 1827:
```suggestion
<em>Securing Verifiable Credentials using JOSE and COSE</em> [[VC-JOSE-COSE]]
```

index.html
```diff
@@ -1877,11 +1895,12 @@ <h3>Securing Verifiable Credentials</h3>
...
-single proof mechanism for use with <a>verifiable credentials</a>. For more
-information about the <code>proof</code> mechanism, see the following
-specifications: Data Integrity [[VC-DATA-INTEGRITY]], Securing Verifiable
-Credentials using JSON Web Tokens [VC-JWT], and the "Proofs" section of
-the Verifiable Credential Specifications Directory [[VC-SPECS]].
+single proof mechanisms for use with <a>verifiable credentials</a> or
+<a>verifiable presentations</a>. For more information about the
+<code>proof</code> mechanisms, see the following specifications: Data Integrity
+[[VC-DATA-INTEGRITY]], Securing Verifiable Credentials using JSON Web Tokens
+[VC-JWT], and the "Proofs" section of the Verifiable Credential Specifications
```
TallTed, line 1902:
```suggestion
[[VC-DATA-INTEGRITY]], Securing Verifiable Credentials using JOSE and COSE
[VC-JOSE-COSE], and the "Proofs" section of the Verifiable Credential Specifications
```

TallTed (changes requested):
Editorial polish. I don't I'm asking to change anything normative.

index.html
```diff
@@ -641,15 +641,15 @@ <h2>Core Data Model</h2>
...
+do not have normative requirements nor associated media types defined by this
+specification. For example, the abstract concepts of <a>credential</a> and
 <a>presentation</a> do not have normative requirements nor associated media
 types. However, the concrete concepts of a <a>verifiable credential</a> or a
 <a>verifiable presentation</a> are defined as <a>conforming documents</a> and
-do have associated media types. Furthermore, what differentiates the
-abstract concepts of <a>credential</a> and <a>presentation</a> from the
-concrete concepts of a <a>verifiable credential</a> and a
-<a>verifiable presentation</a> is the fact that only the "verifiable" objects
+do have associated media types. Furthermore, the only difference between the
```
OR13, line 649: I read this again, and the word only is misleading here

The list of differences: `credentials` are abstract, have no serialization, and no media type, `verifiable credentials` are concrete, have serializations, and have media types. `verifiable credentials` have "abstracts proofs / cryptographic assurances"... `credentials` to do not.

I feel we are already saying this, but I also feel it could be made clearer, without leaning so heavily on "abstracts proofs / cryptographic assurances" as the dividing line.

msporny: Yes, I agree with the language above. I'll try another reword in a future pass that tries to simplify the language. I specifically like the language you provided here: "credentials are abstract, have no serialization, and no media type, verifiable credentials are concrete, have serializations, and have media types. verifiable credentials have "abstracts proofs / cryptographic assurances"... credentialsdo not." ... I will try to use a variation of that without losing other things that previous other reviewers wanted in the text as well.

dlongley: > You have to produce a serialization of a credential to feed it to a mechanism that will output a verifiable credential including internal or external proof.

Yes to this. We need to be careful with the language such that we don't make this impossible or "questionably wrong" in some way. This is why I alluded to "verifiable credential" just being "the name" we happen to have chosen to refer to a document that conforms to the VCDM.

It's important that we say that there are mechanisms that can be used to secure a VC such that it becomes "cryptographically verifiable", but we should not create some language that makes it impossible / incorrect to, for example, send a VC (with `application/vc+ld+json` as the media type) to a service that will secure it in some way.

index.html
```diff
@@ -633,10 +633,27 @@ <h2>Core Data Model</h2>
...
-<a>credentials</a>, and <a>presentations</a>, which form the foundation of this
+<a>credentials</a>, <a>presentations</a>, <a>verifiable credentials</a>, and
+<a>verifiable presentations</a>, which form the foundation of this
 specification.
       </p>
 
+      <p class="note"
+        title="Abstract concepts vs. concrete concepts">
+Readers might note that some abstract concepts described in this section
+do not have normative requirements nor associated media types defined by this
```
David-Chadwick, line 644: The normative requirements for a verifiableCredential also apply to the credential ie. the VC without the proof. So I do not agree with the sentence above. The difference is that the credential does not have a standardised media type whilst the verifiableCredential does have. (As @TallTed  points out, someone could specify a media type for a credential as well if they wanted to).

David-Chadwick (changes requested):

index.html
OR13: > This is why I alluded to "verifiable credential" just being "the name" we happen to have chosen to refer to a document that conforms to the VCDM.

The problem with this, is we know there are several competing concrete syntaxes for the name "verifiable credential"...  so we are choosing a name, and an approach that is making this worse, not better, when we create further confusion regarding `credential` vs `verifiable credential`.

> but we should not create some language that makes it impossible / incorrect to, for example, send a VC (with application/vc+ld+json as the media type) to a service that will secure it in some way.

This works fine, if you assume `verifiable credential` is a document format, not a securing format.

If you assert that a `verifiable credential` has at least one layer of cryptography applied to it... but you also assert that the media type works for serializations that have 0 layers of cryptography around it.... we have just reinvented "alg: none" for JSON.

The problem appears to be media type confusion born of the fact that data integrity proofs do not have a separate media type.

We don't have this problem for JWT / SD-JWT etc... because they have media types and that are secured representations, that secure JSON payloads (ignoring the fact that "alg: none" still exists there).

They treat the content / payload as unsecured (explicitly in the case of "alg:none")... they wrap it with an external proof, and you end with 2 media types:

1. `application/vc+ld+json` a JSON media type with possibly no security
1. `application/jwt` / `application/sd-jwt`  / `application/vc+ld+json+sd-jwt` a string encoding of a media type that secured a JSON claimset, (possibly with no security, aka "alg: none").

With Data integrity Proofs, we have the following:

1. `application/vc+ld+json` a JSON media type with possibly no security (proof is optional)
1. `application/vc+ld+json` a JSON media type with security (proof is mandatory)

"alg: none" happens when you use the same representation for both secured and unsecured content... data integrity proofs are creating this issue, by riding along inside the core data model's media type, based on the presence of a JSON member.... that is exactly what happens when you decode the "protected header" and find "alg:none" instead of "alg:ES384".... its presence of a member, determining if the media type is actually secured.

IMO, the best solution to this is to actually drop the "security language" from the core data model entirely.. just treat the core data model as a JSON-LD Document model, and don't have ANY langauge that implies anything about security (remove the `proof` concept from the core data model completely).

Then is the securing specifications, explain the concept of proof, is specific to the securing specification, and comment on its unique properties...

For example: 
- JWT has "alg:none", is this allowed for VC?
- Data Integrity Proofs have "proof sets and proof chains", are these allowed for VCs?

etc... the problem remains that we are doing too much security work in the core data model, and that leads to bias towards specific security formats, like data integrity proofs... or weak language, caused by trying to leave the door open for each security specification to handle things differently...  (embedded vs external proofs).

index.html
OR13: @dlongley just asserted above that the media type for a "credential" could be `application/vc+ld+json`... so I don't know if we have consensus yet.

TallTed: > The media type for a credential could be `application/json`

I would prefer `application/ld+json`, as `application/json` is likely to encourage credentials which require securing as arbitrary JSON, as that is what they are...

index.html
```diff
@@ -633,10 +633,27 @@ <h2>Core Data Model</h2>
...
       <p>
 The following sections outline core data model concepts, such as <a>claims</a>,
-<a>credentials</a>, and <a>presentations</a>, which form the foundation of this
+<a>credentials</a>, <a>presentations</a>, <a>verifiable credentials</a>, and
+<a>verifiable presentations</a>, which form the foundation of this
 specification.
       </p>
 
+      <p class="note"
+        title="Abstract concepts vs. concrete concepts">
```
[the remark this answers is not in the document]
TallTed: It seems there might be good reasons we went there with DIDv1, which reasons might also mean we should go there with VCDMv2. I have not come to that conclusion, but I do not think it should be discarded without consideration and justification beyond "it was hard and took time".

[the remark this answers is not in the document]
msporny: I agree that we don't want to go down the abstract data model rathole that the DID spec fell into, which is why this group picked JSON-LD as the data model and base serialization type. 

This PR does not suggest that the spec uses an abstract data model (and, like you, I would argue against it if it did). This PR speaks to abstract concepts... not abstract data models.... if you'd like to use a different word for something that's not concretely realized in a serialization, please provide an alternative word... or an alternative fix for issues https://github.com/w3c/vc-data-model/issues/1009 and https://github.com/w3c/vc-data-model/issues/1126.

index.html
msporny: > @dlongley just asserted above that the media type for a "credential" could be `application/vc+ld+json`... so I don't know if we have consensus yet.

No, I don't think he did... I think he was saying (in the very least): "Don't make it impossible to add a proof to a chain of proofs via a issuer service." ... in that situation, you would send an `application/vc+ld+json` ... but I would expect a number of issuer implementations to accept `application/json`, `application/ld+json`, or `application/vc+ld+json` as long as it could process it correctly. Developers often tend to be lazy/wrong about media types... which is why it's difficult to build any software system that depends on them.

Sakurann (changes requested):
why is there a need to introduce abstract vs. concrete notion? Can we simply add a note that says something like ` the difference between concepts of verifiable credential/verifiable presentation and credential/presentation is that the former are secured in a way that is cryptographically verifiable.`?

index.html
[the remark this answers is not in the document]
msporny: @selfissued I have removed all language referring to "abstract" in this note in commit 770c8e340eed1d19a789e64f8d0c89b52306debc.

OR13:
@Sakurann For historical context...  we have this MUST statement, which requires us to test...

> At least one securing mechanism, and the details necessary to evaluate it, MUST be expressed for a [credential](https://w3c.github.io/vc-data-model/#dfn-credential) or [presentation](https://w3c.github.io/vc-data-model/#dfn-presentation) to be a [verifiable credential](https://w3c.github.io/vc-data-model/#dfn-verifiable-credential) or [verifiable presentation](https://w3c.github.io/vc-data-model/#dfn-verifiable-presentation); that is, to be [verifiable](https://w3c.github.io/vc-data-model/#dfn-verify).

Then we have the word "credential" used all over the place, including a few where "verifiable credential" might be more acceptable.

Early on the working group considered a media type for "credential" and decided to use a media type for "verifiable credential" instead... The problem is that there are cases where you want to consume a "credential" as  `application/vc+ld+json` to produce "verifiable credential" as  `application/vc+ld+json` ... which leads to a scenario like when an "alg: none" JWT is consumed to produce an "alg: ES256" JWT.

A while ago I accepted this, and I treat the core data model has being a document model with no security guarantees... but that mindset is not compatible with data integrity proofs, which embed the proofs alongside the content.

It all boils down to using the same media type `application/vc+ld+json` to describe `verifiable credentials` with and without a `proof`.... a "verifiable credential" without a "proof" was formerly called a "credential"... and was distinguishable from "verifiable credential" by the absence of a "proof".

TLDR

data integrity proofs don't have media types, and are needing for `application/vc+ld+json` to work for both secured and unsecured representations.

jwt / sd-jwt etc... do have media types, and can clearly be distinguished from `application/vc+ld+json`, making their secured and unsecured representations easier to work with.... at the cost of having separate media types.... they have the same problem with the naming, since "verifiable credential" could refer to either a JWT payload (unsecured json object), or a JWT (secured JSON object).

index.html
```diff
@@ -639,19 +639,17 @@ <h2>Core Data Model</h2>
...
+Readers might note that some concepts described in this section, such as
+<a>credential</a> and <a>presentation</a>, do not have media types defined by
+this specification. However, the concepts of a <a>verifiable credential</a> or a
+<a>verifiable presentation</a> are defined as <a>conforming documents</a> and do
+have associated media types. The concrete difference between the
+concepts &mdash; of <a>credential</a> and <a>presentation</a> &mdash; and the
+concepts &mdash; of <a>verifiable credential</a> and <a>verifiable
+presentation</a> &mdash; is the fact that the "verifiable" objects are secured
+in a way that is cryptographically verifiable. For more details, see Section
+<a href="#securing-verifiable-credentials"></a>.
```
TallTed, line 652: Further massage may be needed. This is better than what it would replace, anyway.
```suggestion
have associated media types. The concrete difference between these concepts
&mdash; between <a>credential</a> and <a>presentation</a> Vs. <a>verifiable
credential</a> and <a>verifiable presentation</a> &mdash; is simply the fact
that the "verifiable" objects are secured in a way that is cryptographically
verifiable, which the others are not. For more details, see Section
<a href="#securing-verifiable-credentials"></a>.
```

TallTed (changes requested):

index.html
```diff
@@ -1799,39 +1814,40 @@ <h3>Securing Verifiable Credentials</h3>
 
         <p>
 At least one securing mechanism, and the details necessary to evaluate it, MUST
-be expressed for a <a>credential</a> or <a>presentation</a> to be a
-<a>verifiable credential</a> or <a>verifiable presentation</a>; that is, to be
-<a>verifiable</a>.
+be expressed to construct a <a>verifiable credential</a> or a
+<a>verifiable presentation</a>.
```
dlongley, line 1818: The above text doesn't really make sense. Expressed where and how? It just says that someone must say that a securing mechanism exists somewhere in the wild ... and if they do that, then someone can construct a VC.

Perhaps if we say this instead it will have more force and (I hope) we can get to consensus:

"It MUST be possible to test whether a <a>verifiable credential</a> or <a>verifiable presentation</a> has been secured by at least one securing mechanism."

longpd (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-08-15-vcwg-special#section1-3) on 2023-08-15

- no resolutions were taken

#### 1.3. Clarify the difference between a "credential" and a "verifiable credential" (pr vc-data-model#1211)

_See github pull request [vc-data-model#1211](https://github.com/w3c/vc-data-model/pull/1211)._




**Brent Zundel:** next PR is 1211: Clarify the difference between a "credential" and a "verifiable credential".  
… multiple approvals, multiple requests for changes. believe that Mike & Kristina's changes have been made. please re-review.  

**Manu Sporny:** have applied change suggestions from Mike and Kristina. David Chadwick approved. Integrated Ted's changes. Dave has a couple changes to make, will integrate those. Only review left will be Orie.  
… Orie left an explanation of his concern, but left out concrete suggestions except to not refer to securing which would be unworkable.  

**Brent Zundel:** noting that Orie has an approval on this.  
… kristina please review 1211.  

**Kristina Yasuda:** will review!  

</details>

TallTed:
@dlongley -- Should [your latest revision](https://github.com/w3c/vc-data-model/pull/1211#pullrequestreview-1579412796) be made into a one-click-merge suggestion?

dlongley:
@TallTed,

It can't be (as there's a line that is outside of this PR so far) -- but Manu is already looking to incorporate the couple of sentences I put in my suggestion.

awoie (approved):

index.html
msporny: Fixed in 073c722ae6d0f75113f89afec809de3874ee281e.

msporny:

PR #1209, opened 2023-07-21
Vocabulary improvement Phase 1

iherman:
This is the first of several steps to clean up the Credentials Vocabulary. The main changes are:

- Anchors have been added to all term definition to allow for an external reference. That is the only change on the spec text itself, no content has been changed. Note that this change takes care of issue #1080
- For all properties, and also for some of the classes, a `defined by` tag has been added in the (YAML) vocabulary definition. This is translated into the `rdfs:isDefinedBy` statement in the formal vocabularies, and a textual translation thereof in the HTML text. Any other text in the property definitions had been removed. Bottom line: the official term definition is the in VCWG specifications, the vocabulary just "refers" to that. Some additional comments:
  - There is one term in the vocabulary, namely `serviceEndpoint` that has ***no*** formal definition. The vocabulary has a comment on this, and this must be solved asap
  - All terms are actually defined in the VCDM spec. The only exception is the `JsonSchema2023` class is defined by the [schema spec](https://www.w3.org/TR/vc-json-schema/#jsonschema2023). 
  - The deprecated terms refer back to the v1.1 of the VCDM specification.
  - Some classes have no direct reference in the spec, their roles are, sort of, "scaffolding" in the vocabulary. This is perfectly fine, they are intermediary classes and carry no other role.
  - The definition of deprecated terms is kept to the bare minimum. In particular, no range or domain statement on other elements on the vocabulary are kept. After all, these are mostly placeholder items, to avoid 404-s on their relevant identifiers.
  - This part of the PR takes care of issue #1061

### For reviewers

The PR does not include the visible files, i.e., the vocabulary in HTML/JSON-LD/Turtle. This is because the generation of these files happen automatically by a GitHub action. To make the reviews possible:

- The files have been generated separately, and can be reached from [the relevant index file](https://w3c.github.io/yml2vocab/previews/vcdm/) in the tool repository for review.
- The textual changes are to be made on the `vocab/credentials/v2/vocabulary.yml` and, for the text framework in the generated HTML, on the `vocab/credentials/v2/template.html` files. (These are the vocabulary generation input files.)

***
[...]

PR #1215, opened 2023-07-26
Update term definition for vc json schemas as creds

decentralgabe:
As per https://github.com/w3c/vc-json-schema/pull/186#issuecomment-1650620431

### DIFF
--- a/contexts/credentials/v2
+++ b/contexts/credentials/v2
@@ -80,6 +80,9 @@
         "termsOfUse": {
           "@id": "https://www.w3.org/2018/credentials#termsOfUse",
           "@type": "@id"
+        },
+        "jsonSchema": {
+          "@id": ""
         }
       }
     },
@@ -123,15 +126,29 @@
       }
     },
 
-    "VerifiableCredentialSchema2023": {
+    "JsonSchemaCredential2023": {
       "@id":
-        "https://w3.org/2018/credentials#VerifiableCredentialSchema2023",
+        "https://w3.org/2018/credentials#JsonSchemaCredential2023",
       "@context": {
         "@protected": true,
 
         "id": "@id",
         "type": "@type"
       }
+    },
+
+    "JsonSchema": {
+      "@id": "https://www.w3.org/TR/vc-json-schema#JsonSchema",
+      "@context": {
+        "@protected": true,
+
+        "id": "@id",
+        "type": "@type",
+
+        "jsonSchema": {
+          "id": "https://www.w3.org/TR/vc-json-schema#JsonSchema",
+          "type": "@json"
+       }
     },
 
     "StatusList2021Credential": {


decentralgabe:
This should be merged (and will need to fix conflicts) after #1209

contexts/credentials/v2
```diff
@@ -123,9 +123,9 @@
       }
     },
 
-    "VerifiableCredentialSchema2023": {
+    "JsonSchemaCredential2023": {
```
msporny, line 126:
```suggestion
    "JsonSchemaCredential": {
```

Are you keeping the date or removing it?

contexts/credentials/v2
```diff
@@ -123,9 +123,9 @@
...
       "@id":
-        "https://w3.org/2018/credentials#VerifiableCredentialSchema2023",
+        "https://w3.org/2018/credentials#JsonSchemaCredential2023",
```
msporny, line 128:
```suggestion
        "https://w3.org/2018/credentials#JsonSchemaCredential",
```

Same question as above?

contexts/credentials/v2
```diff
@@ -134,6 +134,20 @@
       }
     },
 
+    "JsonSchema": {
+      "@id": "https://www.w3.org/TR/vc-json-schema#JsonSchema",
```
msporny, line 138:
```suggestion
      "@id": "https://www.w3.org/TR/vc-json-schema#JsonSchema",
```

VCDM Vocabulary or a different one? I'm fine either way.

contexts/credentials/v2
```diff
@@ -134,6 +134,20 @@
...
+      "@context": {
+        "@protected": true,
+
+        "id": "@id",
+        "type": "@type",
+
+        "jsonSchema": {
+          "id": "https://www.w3.org/TR/vc-json-schema#JsonSchema",
```
msporny, line 146: VCDM Vocabulary or a different one? I'm fine either way.

msporny (changes requested):
Concrete change request: date-stamped or not?

contexts/credentials/v2
decentralgabe: updated

contexts/credentials/v2
decentralgabe: update

contexts/credentials/v2
decentralgabe: removed

contexts/credentials/v2
decentralgabe: removed

decentralgabe:
removing versioning here https://github.com/w3c/vc-json-schema/pull/192

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-08-01-vcwg-special#section1-5) on 2023-08-01

- no resolutions were taken

#### 1.5. Update term definition for vc json schemas as creds (pr vc-data-model#1215)

_See github pull request [vc-data-model#1215](https://github.com/w3c/vc-data-model/pull/1215)._




**Brent Zundel:** Update Term Definition for VC JSON Schema as creds, raised by Gabe, so far no reviews.  
… Gabe, can you talk us through this PR?  

**Gabe Cohen:** there is a change to the naming of the credential representation of the JSON Schema VC, which this PR addresses.  
… there's also docs on how to use the `@json` property in the JSON Schema credential.  
… so, two small changes - to use a type, and to represent a json schema in the credentialSubject property of the VC.  

**Brent Zundel:** thank you, seem straightforward.  

**Manu Sporny:** Gabe, I was waiting for you to update before reviewing, I thought you'd get rid of the datestamp stuff. (I just added a request).  
… in StatusList, looks like we're getting rid of the datetime stamp.  
… you're putting the vocab definition for JSON Schema into the spec, but not any vocabulary file.  
… not sure if you meant to put it into the VC DM vocab, or create a separate one.  
… the current state - the URL might not resolve over time, that would be bad.  
… anyways, I left comments.  

**Gabe Cohen:** not intentional. I know Ivan has another PR opened, 1218, that does add it to vocab.  

**Manu Sporny:** it's sufficient if you just use the vocab term from that PR.  

**Gabe Cohen:** happy to delay the merge, till the VC JSON Spec removes the year; I didn't want to be the first one to do that.  

**Manu Sporny:** you don't have to wait, if you're gonna update the JSON Schema spec, you can do both in parallel.  
… so this PR can go in. (remove the date, update to the vocab url Ivan created in the other PR, and you're good to go).  

**Gabe Cohen:** thanks, will do.  

**Brent Zundel:** Gabe is quick to respond, so changes will be in very soon, so folks, review those.  
… any more comments on this PR?  

</details>

contexts/credentials/v2
```diff
@@ -112,9 +112,9 @@
       }
     },
 
-    "JsonSchema2023": {
+    "JsonSchema": {
       "@id":
-        "https://w3.org/2018/credentials#JsonSchema2023",
+        "https://www.w3.org/TR/vc-json-schema/#JsonSchema",
```
msporny, line 117:
```suggestion
        "https://w3.org/2018/credentials#JsonSchema",
```

Fixed to align with this:

https://github.com/w3c/vc-data-model/pull/1218/files

contexts/credentials/v2
```diff
@@ -123,9 +123,9 @@
       }
     },
 
-    "VerifiableCredentialSchema2023": {
+    "JsonSchemaCredential": {
       "@id":
-        "https://w3.org/2018/credentials#VerifiableCredentialSchema2023",
+        "https://www.w3.org/TR/vc-json-schema/#JsonSchemaCredential",
```
msporny, line 128:
```suggestion
        "https://w3.org/2018/credentials#JsonSchemaCredential",
```

Fixed to align with this:

https://github.com/w3c/vc-data-model/pull/1218/files

contexts/credentials/v2
msporny:
```suggestion
      "@id": "https://w3.org/2018/credentials#JsonSchema",
```

No, still wrong, I believe

Fixed to align with this:

https://github.com/w3c/vc-data-model/pull/1218/files

contexts/credentials/v2
msporny:
```suggestion
          "id": "https://w3.org/2018/credentials#JsonSchema",
```

msporny (changes requested):
Still requires fixes to use the `https://w3.org/2018/credentials#` base vocabulary URL... see PR https://github.com/w3c/vc-data-model/pull/1218/files for details.

contexts/credentials/v2
```diff
@@ -112,9 +112,9 @@
       }
     },
 
-    "JsonSchema2023": {
+    "JsonSchema": {
```
dlongley, line 115: This looks like a duplicate of what's below. This one should be removed as the one below looks more correct (has `jsonSchema` defined in its scoped context).

dlongley (changes requested):
See my suggestion and +1 to @msporny's suggestions.

contexts/credentials/v2
decentralgabe: updated

decentralgabe: updated

contexts/credentials/v2
```diff
@@ -112,26 +112,29 @@
       }
     },
 
-    "JsonSchema2023": {
+
+    "JsonSchemaCredential": {
       "@id":
-        "https://w3.org/2018/credentials#JsonSchema2023",
+        "https://w3.org/2018/credentials#JsonSchemaCredential",
       "@context": {
```
dlongley, line 119: Hmm -- this scoped context has now moved into `JsonSchemaCredential`. It should be under `JsonSchema` instead. That way, when an object of "type" `JsonSchema` is used, it will have the `jsonSchema` property (where the actual JSON schema goes) defined. It's that object that has that property, not the `JsonSchemaCredential`.

dlongley (changes requested):

contexts/credentials/v2
OR13: @dlongley that seems revered to me. Perhaps examples make this clearer:

## Case 1
`JsonSchema` ... a type that refers to a JSON Object that validates JSON... (we intend to implement this one).

```json
{
  "@context": [
  ],
  "id": "https://example.com/credentials/3732",
  "type": ["VerifiableCredential", "EmailCredential"],
  "issuer": "https://example.com/issuers/14",
  "credentialSubject": {
    "emailAddress": "subject@example.com"
  },
  "credentialSchema": {
    "id": "https://example.com/schemas/email.json",
    "type": "JsonSchema"
  }
}
```

## Case 2
`JsonSchemaCredential` ... a type that refers to a VerifiableCredential, that contains a json object, that validates JSON... (we do not plan to implement this one).

```json
{
  "@context": [
  ],
  "id": "https://example.com/credentials/3734",
  "type": ["VerifiableCredential", "JsonSchemaCredential"],
  "issuer": "https://example.com/issuers/14",
  "credentialSubject": {
    "id": "https://example.com/schemas/email-credential-schema.json",
    "type": "JsonSchema",
    "jsonSchema": {
       "$id": "https://example.com/schemas/email-credential-schema.json",
       "$schema": "https://json-schema.org/draft/2020-12/schema",
       "name": "EmailCredential",
       "description": "EmailCredential using JsonSchemaCredential",
       "type": "object",
       "properties": {
         "credentialSubject": {
           "type": "object",
           "properties": {
             "emailAddress": {
               "type": "string",
               "format": "email"
             }
           },
           "required": ["emailAddress"]
         }
       }
    }
  }
}
```

OR13: Perhaps you are suggesting this?

```json
{
  "@context": [
  ],
  "id": "https://example.com/credentials/3734",
  "type": ["VerifiableCredential", "JsonSchemaCredential"], // type
  "issuer": "https://example.com/issuers/14",
  "credentialSubject": {
       "$id": "https://example.com/schemas/email-credential-schema.json",
       "$schema": "https://json-schema.org/draft/2020-12/schema",
       "name": "EmailCredential",
       "description": "EmailCredential using JsonSchemaCredential",
       "type": "object",
       "properties": {
         "credentialSubject": {
           "type": "object",
           "properties": {
             "emailAddress": {
               "type": "string",
               "format": "email"
             }
           },
           "required": ["emailAddress"]
         }
       }
    }
  }
}
```

dlongley: @OR13,

My suggestions will make Case 1 and Case 2 work. The "Perhaps you were suggesting this" is not something I was suggesting and don't believe we should do.

dlongley: @OR13,

Right now, this PR will create these term mappings:

```json
    "JsonSchemaCredential": {
      "@id":
        "https://w3.org/2018/credentials#JsonSchemaCredential",
      "@context": {
        "@protected": true,

        "id": "@id",
        "type": "@type",

       "jsonSchema": {
          "id": "https://w3.org/2018/credentials#JsonSchema",
          "type": "@json"
       }
    },

    "JsonSchema": {
      "@id": "https://w3.org/2018/credentials#JsonSchema",
      "@context": {
        "@protected": true,

        "id": "@id",
        "type": "@type"
    },
```

This would mean that `jsonSchema` would only be defined at the top-level of a VC that used the type `JsonSchemaCredential`. Something like this (which we do not want):

```json
{
  "@context": [
  ],
  "id": "https://example.com/credentials/3734",
  "type": ["VerifiableCredential", "JsonSchemaCredential"], // type which
  // defines the scoped-context with "jsonSchema" for use on this object
  // which we do not want
  "issuer": "https://example.com/issuers/14",
  // we do not want this
  "jsonSchema": {
     "$id": "https://example.com/schemas/email-credential-schema.json",
     "$schema": "https://json-schema.org/draft/2020-12/schema",
     "name": "EmailCredential",
     "description": "EmailCredential using JsonSchemaCredential",
     "type": "object",
     "properties": {
       "credentialSubject": {
         "type": "object",
         "properties": {
           "emailAddress": {
             "type": "string",
             "format": "email"
           }
         },
         "required": ["emailAddress"]
       }
     }
  },
  "credentialSubject": {
    "id": "https://example.com/schemas/email-credential-schema.json",
    "type": "JsonSchema"
  }
}
```

Instead, it should do this:

```json
    "JsonSchemaCredential": "https://w3.org/2018/credentials#JsonSchemaCredential",

    "JsonSchema": {
      "@id": "https://w3.org/2018/credentials#JsonSchema",
      "@context": {
        "@protected": true,

        "id": "@id",
        "type": "@type",

       "jsonSchema": {
          "id": "https://w3.org/2018/credentials#JsonSchema",
          "type": "@json"
       }
    },
```

Which will enable Case 2:

```json
{
  "@context": [
  ],
  "id": "https://example.com/credentials/3734",
  "type": ["VerifiableCredential", "JsonSchemaCredential"], // type
  "issuer": "https://example.com/issuers/14",
  "credentialSubject": {
    "id": "https://example.com/schemas/email-credential-schema.json",
    "type": "JsonSchema", // this type has the scoped-context that
    // defines "jsonSchema" for use on this object
    "jsonSchema": {
       "$id": "https://example.com/schemas/email-credential-schema.json",
       "$schema": "https://json-schema.org/draft/2020-12/schema",
       "name": "EmailCredential",
       "description": "EmailCredential using JsonSchemaCredential",
       "type": "object",
       "properties": {
         "credentialSubject": {
           "type": "object",
           "properties": {
             "emailAddress": {
               "type": "string",
               "format": "email"
             }
           },
           "required": ["emailAddress"]
         }
       }
    }
  }
}
```

It also enables Case 1 by simply defining `JsonSchema` as a usable type at all.

OR13: Strong disagreement.... I think.

I don't want to make every instance of "JsonSchema" a verifiable credential.

I don't even want to use "JsonSchemaCredential" at all.

My goal is to just use JSON Schema with `credentialSchema` property.

All I need in the v2 context to achieve this is this:

```json
"JsonSchema": {
      "@id": "https://w3.org/2018/credentials#JsonSchema",
}
```
All I need in a credential to achieve this is:

```json
 "credentialSchema": {
    "id": "https://example.com/schemas/email-credential-schema.json",
    "type": "JsonSchema" // type
  }
```

mprorock: > 

huge +1 to this - pretty hard objection to this pr as it stands

decentralgabe: @dlongley if your suggestion makes case 1 and case 2 true then I'll update the PR

dlongley: @decentralgabe,

> @dlongley if your suggestion makes case 1 and case 2 true then I'll update the PR

Yup, that's what it does. You'll have to figure out whether @OR13 and @mprorock will take issue with you supporting both cases. They appear to only want to support Case 1, but I'm not convinced that there isn't some confusion here.

dlongley: @OR13,

> Strong disagreement.... I think.
> 
> I don't want to make every instance of "JsonSchema" a verifiable credential.

Me either. This doesn't do that.

> I don't even want to use "JsonSchemaCredential" at all.

Yes, but @decentralgabe wants to.

> My goal is to just use normal OpenJS Foundation spec conformant JSON Schemas with `credentialSchema` property.

That's fine with what I've proposed.

> All I need in the v2 context to achieve this is this:
> 
> ```json
> "JsonSchema": {
>       "@id": "https://w3.org/2018/credentials#JsonSchema",
> }
> ```

True. But if you add a scoped context, you get to do what you want and @decentralgabe gets to do what he wants too.

> All I need in a credential to achieve this is:
> 
> ```json
>  "credentialSchema": {
>     "id": "https://example.com/schemas/email-credential-schema.json",
>     "type": "JsonSchema" // type
>   }
> ```

True. You still get to do this with my suggestion.

> I agree regarding not wanting the "JsonSchemaCredential" and its dependence on the "jsonSchema" predicate... but please don't break "JsonSchema" when fixing that problem : )

I don't think something is breaking. What specifically do you think is breaking and how?

decentralgabe: @dlongley I've updated the PR to reflect your changes ( I think )

contexts/credentials/v2
```diff
@@ -123,14 +123,18 @@
...
+      "@id": "https://w3.org/2018/credentials#JsonSchema",
       "@context": {
         "@protected": true,
 
         "id": "@id",
-        "type": "@type"
+        "type": "@type",
+
+        "jsonSchema": {
+           "id": "https://w3.org/2018/credentials#JsonSchema",
```
dlongley, line 135:
```suggestion
           "id": "https://w3.org/2018/credentials#jsonSchema",
```

contexts/credentials/v2
```diff
@@ -112,9 +112,9 @@
       }
     },
 
-    "JsonSchema2023": {
+    "JsonSchemaCredential": {
       "@id":
-        "https://w3.org/2018/credentials#JsonSchema2023",
+        "https://w3.org/2018/credentials#JsonSchemaCredential",
```
dlongley, line 117: Since no scoped context is needed here, this can be simplified to:

```
    "JsonSchemaCredential": "https://w3.org/2018/credentials#JsonSchemaCredential",
```

dlongley (changes requested):
@decentralgabe, very close -- a couple more suggestions.

The latter suggestion is important since the property `jsonSchema` should be mapped to the URL with the matching lower case URL fragment. The class / type name `JsonSchema` (upper case) is mapped to the URL with the upper case URL fragment.

contexts/credentials/v2
decentralgabe: thanks, updated

contexts/credentials/v2
decentralgabe: @dlongley  one question -- I would assume since we want to expose the term `jsonSchema` inside the `credentialSubject` of a `JsonSchemaCredential` then the term would need to be defined under `JsonSchemaCredential` why is it that there's no need to define `jsonSchema` within the `JsonSchemaCredential` type?

andresuribe87: There's a situation I don't quite understand. Imagine the following scenario

The VC has the property below.
```json
{
  "credentialSchema": {"id": "https://example.com/my-json-schema.json", "type": "JsonSchema"}
}
```

And the contents that result from dereferencing `https://example.com/my-json-schema.json` are the following:
```json
    {
       "$id": "https://example.com/my-json-schema.json",
       "$schema": "https://json-schema.org/draft/2020-12/schema",
       "name": "EmailCredential",
       "type": "object",
       "properties": {}
    }
```

Would the `VC.credentialSchema.type` conflict with `dereferenceURL(VC.credentialSchema.id).type`? Note that the result of `dereferenceURL` does not contain any `jsonSchema` property within it.

dlongley: @decentralgabe,

> @dlongley one question -- I would assume since we want to expose the term `jsonSchema` inside the `credentialSubject` of a `JsonSchemaCredential` then the term would need to be defined under `JsonSchemaCredential` why is it that there's no need to define `jsonSchema` within the `JsonSchemaCredential` type?

Because `jsonSchema` is not a property of an object of type `JsonSchemaCredential`. It's a property of an object of type `JsonSchema`.

Type-scoped contexts are applied to the objects where the `type` is declared. It's similar to OOP where you declare the properties that can appear on an object of a certain type. An object of type `JsonSchema` can use the `jsonSchema` property. We want that property to be defined on an object of that type, not on `JsonSchemaCredential`.

As an example:

```
{
  "@context": {
    "pet": "example:pet",
    "id": "@id",
    "type": "@type",
    "Dog": {
      "@id": "example:Dog",
      "@context": {
        "id": "@id",
        "type": "@type",
        "name": "example:name",
        "breed": "example:breed",
        "favoriteToy": "example:favoriteToy"
        // etc.
      }
    }
  },
  "pet": {
    "id": "example:some_id",
    "type": "Dog",
    "name": "Bark Hamill",
    "breed": "Chihuahua"
    // "favoriteToy" you don't have to use every defined property
  }
}
```

dlongley (approved):

contexts/credentials/v2
dlongley: @andresuribe87,

> There's a situation I don't quite understand. Imagine the following scenario
> 
> The VC has the property below.
> 
> ```json
> {
>   "credentialSchema": {"id": "https://example.com/my-json-schema.json", "type": "JsonSchema"}
> }
> ```
> 
> And the contents that result from dereferencing `https://example.com/my-json-schema.json` are the following:
> 
> ```json
>     {
>        "$id": "https://example.com/my-json-schema.json",
>        "$schema": "https://json-schema.org/draft/2020-12/schema",
>        "name": "EmailCredential",
>        "type": "object",
>        "properties": {}
>     }
> ```
> 
> Would the `VC.credentialSchema.type` conflict with `dereferenceURL(VC.credentialSchema.id).type`? Note that the result of `dereferenceURL` does not contain any `jsonSchema` property within it.

Just because a term mapping exists in a JSON-LD context does not mean it has to be used. Also, since the content at that URL is not `application/ld+json`, there's no reason to think (nor any automated way) its contents should be merged into a knowledge graph. If you wanted to enable some kind of automated way to merge, you could also serve `application/ld+json` from that URL with a different representation (when requested via content negotiation), but that doesn't sound like it's really even a use case here.

OR13 (changes requested):
Based on the discussion on the PR, I don't feel comfortable approving this until additional examples are added along with the context changes.

OR13:
We have the following test case today:

https://github.com/transmute-industries/vc-jwt-test-suite/blob/main/data/inputs/claimsets/alumni-credential-with-schema.json#L12

## Case A.1

```json
{
  "@context": [
  ],
  "id": "http://university.example/credentials/1872",
  "type": ["VerifiableCredential", "ExampleAlumniCredential"],
  "credentialSchema": {
    "id": "https://university.example/schemas/ExampleAlumniCredential",
    "type": "JsonSchema"
  },
  "credentialSubject": {
    "alumniOf": {
      "id": "did:example:c276e12ec21ebfeb1f712ebc6f1",
      "name": "Example University"
    }
  }
}
```

After this PR is merged, this test case still be valid? 

Can we include an example to make this clearer?

Can we also add an example based on the thread above?

## Case A.2
```json
{
  "@context": [
  ],
  "id": "http://university.example/credentials/1872",
  "type": ["VerifiableCredential", "ExampleAlumniCredential"],
  "credentialSchema": {
    "id": "https://university.example/schemas/ExampleAlumniCredential",
    "type": "JsonSchema",
    "jsonSchema": {
      "$id": "https://university.example/schemas/ExampleAlumniCredential",
      "$schema": "https://json-schema.org/draft/2020-12/schema",
      "name": "ExampleAlumniCredential",
      "type": "object",
      "properties": { ... }
    }
  },
  "credentialSubject": {
    "alumniOf": {
      "id": "did:example:c276e12ec21ebfeb1f712ebc6f1",
      "name": "Example University"
    }
  }
}
```

To be clear, current blocking feedback:

1. add an example for A.1 and A.2
2. confirm my understanding is correct regarding both being legal after the PR is merged.

msporny:
> confirm my understanding is correct regarding both being legal after the PR is merged.

FWIW, @OR13's understanding of the PR is aligned with my understanding of the PR, but I think there is another use case that needs to be explained...

Non-blocking thought: 

I wouldn't expect `credentialSchema.id` to be used in the second example... or if it's used, I would expect the contents of `credentialSchema.jsonSchema` to match the `credentialSchema.id` file when dereferenced. That is, I would expect the contents of `jsonSchema` to match the contents of `https://university.example/schemas/ExampleAlumniCredential`.

In addition to the examples above, I'd also expect an example that would be something along these lines

## Case A.2
```json
{
  "@context": [
  ],
  "id": "http://university.example/credentials/schemas/9288",
  "type": ["VerifiableCredential", "JsonSchemaCredential"],
  "issuer": "https://university.example/issuers/827",
  "validFrom": "2023-08-01T19:23:24Z",
  "credentialSubject": {
    "id": "https://university.example/schemas/ExampleAlumniCredential",
    "type": "JsonSchema",
    "jsonSchema": {
      "$id": "https://university.example/schemas/ExampleAlumniCredential",
      "$schema": "https://json-schema.org/draft/2020-12/schema",
      "name": "ExampleAlumniCredential",
      "type": "object",
      "properties": { ... }
    }
  }
}
```

... and then, again, `jsonSchema` is expected to match the contents of `https://university.example/schemas/ExampleAlumniCredential` ... it feels like using a digest would be simpler, not duplicate content, and might possibly lead to a better developer experience. That would be an argument to remove `jsonSchema` from the specification. 

Again, all of the above are non-blocking thoughts on this PR.

msporny (approved):

decentralgabe:
comments addressed. multiple approvals from editors. open for >1 week. merging.

PR #1225, opened 2023-08-03
Editorial update index.html: use JsonSchema and not JsonSchema2023

iherman:
This is just a leftover of the naming/versioning change around JSON Schemas.

(Editorial change, only in the examples).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1225.html" title="Last updated on Aug 4, 2023, 4:11 AM UTC (8b2e277)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1225/892eeac...8b2e277.html" title="Last updated on Aug 4, 2023, 4:11 AM UTC (8b2e277)">Diff</a>

PR #1223, opened 2023-08-02
Add aud registered claim to v2 context

OR13:
```json
{
  "@context": {
      "aud": {
      "@id": "https://www.iana.org/assignments/jwt#aud",
      "@type": "@id"
    }
  },
  "@id": "urn:uuid:123",
  "aud": ["urn:uuid:456", "https://api.example/verifier"]
}
```

```
<urn:uuid:123> <https://www.iana.org/assignments/jwt#aud> <https://api.example/verifier> .
<urn:uuid:123> <https://www.iana.org/assignments/jwt#aud> <urn:uuid:456> .

```

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1223.html" title="Last updated on Aug 4, 2023, 12:57 PM UTC (ca8de49)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1223/892eeac...ca8de49.html" title="Last updated on Aug 4, 2023, 12:57 PM UTC (ca8de49)">Diff</a>

PR #1222, opened 2023-08-02
Add exp registered claim to v2 context 

OR13:
Basically same as: https://github.com/w3c/vc-data-model/pull/1221

PR #1221, opened 2023-08-02
Add nbf registered claim to v2 context 

OR13:
This PR adds `nbf` to the v2 context.

```json
{
  "@context": [
    "https://w3id.org/traceability/v1", {
    "nbf": {
      "@id": "https://www.iana.org/assignments/jwt#iat",
      "@type": "https://www.w3.org/2001/XMLSchema#nonNegativeInteger"
    }
    }
  ],
  "type": "Organization",
  "name": "Bartell Inc 🔥",
  "description": "Realigned maximized alliance",
  "nbf": 123
}
```

...

```json
{
  "@type": "https://schema.org/Organization",
  "https://schema.org/description": "Realigned maximized alliance",
  "https://schema.org/name": "Bartell Inc 🔥",
  "https://www.iana.org/assignments/jwt#iat": {
    "@type": "https://www.w3.org/2001/XMLSchema#nonNegativeInteger",
    "@value": 123
  }
}
```

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1221.html" title="Last updated on Aug 4, 2023, 1:03 PM UTC (8ddfc8d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1221/7e6e996...8ddfc8d.html" title="Last updated on Aug 4, 2023, 1:03 PM UTC (8ddfc8d)">Diff</a>

PR #1226, opened 2023-08-03
Added reserved confidenceMethod property to the vocabulary

iherman:
This follows the latest update of the VCDM spec. Note that I have also added a ConfidenceMethod class, serving as a common superclass for specific instances.

It was not clear in the defintions whether there is any extra domain information to be added. So I left that open.

### DIFF
--- a/index.html
+++ b/index.html
@@ -3568,7 +3568,7 @@ in the specification if at least one specification with two independent
 implementations are demonstrated by the end of the Candidate Recommendation
 Phase. If that does not occur, this reservation will remain, but the existing
 section in the specification will be removed.
-See <a href="https://w3c-ccg.github.io/vc-confidence-method/">Verifiable Credential Confidence Methods</a>.
+See <a href="https://w3c-ccg.github.io/confidence-method-spec/">Verifiable Credential Confidence Methods</a>.
                 </p>
               </td>
             </tr>
--- a/index.html
+++ b/index.html
@@ -3556,6 +3556,23 @@ have at least a draft specification that is being incubated in a community group
 
           <tbody>
             <tr>
+              <td>`confidenceMethod`</td>
+              <td>
+A property used for specifying one or more methods that a verifier 
+might rely on in order to authenticate the holder of a presentation. 
+The associated vocabulary URL MUST be
+`https://www.w3.org/2018/credentials#confidenceMethod`.
+                <p class="issue" title="(AT RISK) Reservation depends on implementations">
+This property reservation might be deleted in favor of an existing section
+in the specification if at least one specification with two independent
+implementations are demonstrated by the end of the Candidate Recommendation
+Phase. If that does not occur, this reservation will remain, but the existing
+section in the specification will be removed.
+See <a href="https://w3c-ccg.github.io/vc-confidence-method/">Verifiable Credential Confidence Methods</a>.
+                </p>
+              </td>
+            </tr>
+            <tr>
               <td>`evidence`</td>
               <td>
 A property used for specifying the evidence that was presented in order to
@@ -3588,13 +3605,14 @@ section in the specification will be removed.
             <tr>
               <td>`renderMethod`</td>
               <td>
-A property used for specifying how to render a credential into a visual,
+A property used for specifying one or more methods to render a credential into a visual,
 auditory, or haptic format. The associated vocabulary URL MUST be
 `https://www.w3.org/2018/credentials#renderMethod`.
                 <p class="issue" title="(AT RISK) Reservation depends on implementations">
 This reserved property is at risk and will be removed from the
 specification if at least one specification with two independent implementations
 are not demonstrated by the end of the Candidate Recommendation Phase.
+See <a href="https://w3c-ccg.github.io/vc-render-method/">Verifiable Credential Rendering Methods</a>.
                 </p>
               </td>
             </tr>
--- a/index.html
+++ b/index.html
@@ -2230,7 +2230,7 @@ each data schema is determined by the specific type definition.
             </p>
             <p>
 If multiple schemas are present, validity is determined according to the
-processing rules outlined by each associated <code>credentialSchema</code> 
+processing rules outlined by each associated <code>credentialSchema</code>
 <code>type</code> property.
             </p>
           </dd>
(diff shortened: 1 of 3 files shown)


vocab/credentials/v2/vocabulary.yml
```diff
@@ -79,6 +85,12 @@ property:
     range: IRI
     comment: An entity about which claims are made.
 
+  - id: confidenceMethod
+    label: Confidence method
+    defined_by: https://w3c-ccg.github.io/confidence-method-spec/
```
OR13, line 90: We reserved this in a table... do we expect that term definition to show up here? 

https://www.w3.org/2018/credentials#confidenceMethod

https://w3c.github.io/vc-data-model/#reserved-extension-points

vocab/credentials/v2/vocabulary.yml
```diff
@@ -29,6 +29,12 @@ class:
     label: Credential status
     comment: A Credential Status provides enough information to determine the current status of the credential (for example, suspended or revoked). This class serves as a supertype for specific status types.
 
+  - id: ConfidenceMethod
```
OR13, line 32: This appears to be reserving a "class" not a "predicate"... I guess we really need to add `Credential` or we will have a weird mismatch between how to handle classes in the spec.

decentralgabe (approved):

msporny (changes requested):
PR includes commits on main, needs a rebase.

iherman:
> PR includes commits on main, needs a rebase.

@msporny I do not understand this remark. Rebase on what?

vocab/credentials/v2/vocabulary.yml
iherman: The current vocabulary includes entries for the other reserved terms already, and the definition in the spec explicitly refers to a URL in the credentials vocabulary. Ie., consistency requires to put this entry into the table, too.

vocab/credentials/v2/vocabulary.yml
iherman: I do not understand what you are saying, @OR13. The vocabulary includes a number of "Classes" that serve as some sort of an "abstract" superclass. I fail to see how this has any relation to the presence or not of a separate class for credentials.

msporny:
> @msporny I do not understand this remark. Rebase on what?

Look at the file diff... it contains your changes and a bunch of changes on an invalid branch on main. There was a PR that was pulled in a few days ago (the Ecosystem Compatability PR) that destroyed history, so I had to re-create history on main and force push to main, which then caused this PR (and many others) to have issues. To rebase:

```
git checkout main
git pull
git checkout iherman-patch-1
git rebase main
```

^^ That might do the trick, or you might have conflicts you need to resolve.

I can also squash-merge, but the current diff is probably going to confuse people that are reviewing, and they'll start suggesting changes to things that were not in this PR.

msporny:
> @msporny I do not understand this remark. Rebase on what?

Look at the file diff... it contains your changes and a bunch of changes on an invalid branch on main. There was a PR that was pulled in a few days ago (the Ecosystem Compatability PR) that destroyed history, so I had to re-create history on main and force push to main, which then caused this PR (and many others) to have issues. To rebase:

```
git checkout main
git pull
git checkout iherman-add-confidenceMethod-vocabulary
git rebase main
```

^^ That might do the trick, or you might have conflicts you need to resolve.

I can also squash-merge, but the current diff is probably going to confuse people that are reviewing, and they'll start suggesting changes to things that were not in this PR.

iherman:
@msporny that set of instructions completely screwed up my local copy :-( Git asks me to push a bunch of things to main (?) which I presume I should not do, and the branch `iherman-add-...` got also garbled beyond recognition... And, to be honest, I am not sure what I am doing. I should certainly _not_ push anything onto main, I presume...

The only thing I can propose to do is a hard reset (the changes for this PR and for #1225 are minimal, so it may be simpler). Ie, when you think everything is fine on the github side, then I remove my local copies, do a fresh cloning, and then submit new PR with the changes. Otherwise this might become way too complicated.

vocab/credentials/v2/vocabulary.yml
OR13: I think you are saying, we don't expect to see these URLs in the vocubulary. 

And defined by URL can change in the future if the working group decides to define reserved terms?

OR13:
@iherman I recently addressed a similar concern with:

- https://github.com/w3c/vc-data-model/pull/1223
- https://github.com/w3c/vc-data-model/pull/1222
- https://github.com/w3c/vc-data-model/pull/1221
- https://github.com/w3c/vc-data-model/pull/1220
- https://github.com/w3c/vc-data-model/pull/1219

You can use rebase, if you are comfortable with it... Or you can look at the diff, and remove the changes that 'teleported' into your PR thanks to the previous merges to main.

I prefer the latter, an extra commit is easier for me to manage that caring about rebase... the resulting diff will be the same at the end of it.

Its tedius, but you can do all this from the github UI if you are patient enough.

index.html
```diff
@@ -2230,7 +2230,7 @@ <h3>Data Schemas</h3>
             </p>
             <p>
 If multiple schemas are present, validity is determined according to the
-processing rules outlined by each associated <code>credentialSchema</code> 
+processing rules outlined by each associated <code>credentialSchema</code>
```
OR13, line 2233:
```suggestion
processing rules outlined by each associated <code>credentialSchema</code> 
```

index.html
```diff
@@ -3588,13 +3608,14 @@ <h3>Reserved Extension Points</h3>
             <tr>
               <td>`renderMethod`</td>
               <td>
-A property used for specifying how to render a credential into a visual,
+A property used for specifying one or more methods to render a credential into a visual,
```
OR13, line 3611:
```suggestion
A property used for specifying how to render a credential into a visual,
```

index.html
```diff
@@ -3588,13 +3608,14 @@ <h3>Reserved Extension Points</h3>
...
 auditory, or haptic format. The associated vocabulary URL MUST be
 `https://www.w3.org/2018/credentials#renderMethod`.
                 <p class="issue" title="(AT RISK) Reservation depends on implementations">
 This reserved property is at risk and will be removed from the
 specification if at least one specification with two independent implementations
 are not demonstrated by the end of the Candidate Recommendation Phase.
+See <a href="https://w3c-ccg.github.io/vc-render-method/">Verifiable Credential Rendering Methods</a>.
```
OR13, line 3618:
```suggestion
```

OR13:
@iherman I suggested the changes I think you need to apply to fix this, in case you just want to accept them instead of muck with rebase.

vocab/credentials/v2/vocabulary.yml
iherman: I am saying there is no definition as of today. We are not yet at Rec and the reserved terms are somehow at risk. If they become "regular" than a proper URL must come to the fore.

vocab/credentials/v2/vocabulary.yml
iherman: I would prefer not to discuss this issue here. It is a valid question, but it would turn this PR into an endless discussion thread, going way beyond a simple addition to the vocabulary. Let us avoid that; this is what may lead to problems like https://github.com/w3c/vc-data-model/pull/1226#issuecomment-1665495765

iherman:
> @iherman I suggested the changes I think you need to apply to fix this, in case you just want to accept them instead of muck with rebase.

Thank you @OR13 but I am uneasy getting into something I do not understand... let alone the fact that, as I said in https://github.com/w3c/vc-data-model/pull/1226#issuecomment-1665559996, the rebase instructions seem to have completely messed up my local clone, so I would probably be better off starting with a fresh clone anyway.

My only question is (@msporny ?) is the github content in order _now_, ie, can I clone and submit the two super simple PR-s now, or should I wait?

iherman:
> 
>  let alone the fact that, as I said in [#1226 (comment)](https://github.com/w3c/vc-data-model/pull/1226#issuecomment-1665559996), the rebase instructions seem to have completely messed up my local clone, so I would probably be better off starting with a fresh clone anyway.

I usually try to keep the local clones up-to-date by pulling all changes almost every morning. I presume that I pulled down the mess to my local copy at some point...

msporny:
> I usually try to keep the local clones up-to-date by pulling all changes almost every morning. I presume that I pulled down the mess to my local copy at some point...

Yes, correct, that's what happened. :(

> My only question is (@msporny ?) is the github content in order now, ie, can I clone and submit the two super simple PR-s now, or should I wait?

Yes, the main branch is back to a stable state. Feel free to proceed.

iherman:
Move to #1228

TallTed:
> Move to https://github.com/w3c/vc-data-model/pull/1228, this PR should be closed without merge.

Maybe change the title, as most of the discussion here is about the glitch, not the intended meat of the PR?

iherman:
> > Move to #1228, this PR should be closed without merge.
> 
> Maybe change the title, as most of the discussion here is about the glitch, not the intended meat of the PR?

I'd prefer to close this PR right away. But I am not a code owner

msporny:
@iherman wrote:
> I'd prefer to close this PR right away. But I am not a code owner

Closing.

PR #1242, opened 2023-08-15
Add sd-jwt registered claim names to v2 context

OR13:
This PR adds registered claim names for sd-jwt to the v2 context.

Like the other PRs, it assumes IANA registries are navigable.

Unlike the other PRs, these terms are not yet present in the JWT registries... so the OAuth WG should be consulted on the best location for implementers to understand these terms.

[Example SD-JWT that would use these definitions](https://jwt.io/#debugger-io?token=eyJhbGciOiJFUzM4NCJ9.eyJAY29udGV4dCI6WyJodHRwczovL3d3dy53My5vcmcvbnMvY3JlZGVudGlhbHMvdjIiLCJodHRwczovL3czaWQub3JnL3RyYWNlYWJpbGl0eS92MSJdLCJpZCI6Imh0dHA6Ly9zdXBwbHktY2hhaW4uZXhhbXBsZS9jcmVkZW50aWFscy9kZDBjNmY5YS01ZGY2LTQwYTMtYmIzNC04NjNjZDFmZGE2MDYiLCJ0eXBlIjpbIlZlcmlmaWFibGVDcmVkZW50aWFsIiwiRW50cnlOdW1iZXJDcmVkZW50aWFsIl0sInZhbGlkRnJvbSI6IjIwMjMtMDgtMTVUMTU6NTg6NTUuNzMyWiIsInZhbGlkVW50aWwiOiIyMDIzLTA5LTE1VDE1OjU4OjU1LjczMloiLCJpc3N1ZXIiOnsidHlwZSI6WyJPcmdhbml6YXRpb24iXSwiaWQiOiJkaWQ6d2ViOmlzc3Vlci5leGFtcGxlIiwibmFtZSI6IkFDTUUgQ3VzdG9tcyBCcm9rZXIiLCJfc2QiOlsiX1lqMkM3akhIWGozRHNtV19QQTVHc1NTZFNTX0RHN2s2ZV9hNFpPdUJURSJdfSwiY3JlZGVudGlhbFN1YmplY3QiOnsidHlwZSI6WyJFbnRyeU51bWJlciJdLCJfc2QiOlsiNW1wOG1iejBsQWtfd0txTzcwbWtfZzJLRkh1MFRBQzk2SVpSaXhja3pSUSJdfSwiX3NkX2FsZyI6InNoYS0yNTYiLCJpc3MiOiJkaWQ6d2ViOmlzc3Vlci5leGFtcGxlIiwiaWF0IjoxNjkyMTE1MTM1LCJleHAiOjE2OTQ3OTM1MzUsImNuZiI6eyJqd2siOnsia3R5IjoiRUMiLCJjcnYiOiJQLTM4NCIsIngiOiJEX2pIbVZ3TVRsNnF3LTQ5U2lYSjU5OVI2bnB0WTN2Yk96ckNEQjhUSFlGZklvVFVJeXZkaXBMR2FNcFRYblBIIiwieSI6ImhwQmx1RHB1VDY4QzE3cnVYMGtkOU5OT0g5V016bFE5V2lHWWQteEhDZ0ZnYjg4d2pQNTlaVkdjYm1mMUdJdWIifX19.l-l0HIRuyxbpxiisQdcS260Fz5j0yJgWdLyk9rgS1UZP6yoq2Y6BVGoROQwfw7t6dxzUrSxqzan7qwCB2_Q56LMaFb-JdkOQAPlVEZZlOT1Zfw9gpzrV38_0n3rZd_7z~WyJmaGhwQ1pvMlR1VmNJekJBNDUxZWNBIiwgImxvY2F0aW9uIiwgeyJ0eXBlIjogWyJQbGFjZSJdLCAiYWRkcmVzcyI6IHsidHlwZSI6IFsiUG9zdGFsQWRkcmVzcyJdLCAic3RyZWV0QWRkcmVzcyI6ICIxMjMgRXhhbXBsZSBTdHJlZXQiLCAiYWRkcmVzc0xvY2FsaXR5IjogIlRvcm9udG8iLCAiYWRkcmVzc1JlZ2lvbiI6ICJPTiIsICJhZGRyZXNzQ291bnRyeSI6ICJDQSIsICJwb3N0YWxDb2RlIjogIk0zQiAxQTIifX1d~WyIzSU9kakNWLWl2Tml6YXVtLUt3YW53IiwgImVudHJ5TnVtYmVyIiwgIjEyMzQ1MTIzNDU2Il0)

Example n-quads produced from the current v2 context:

```
<did:web:issuer.example> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://schema.org/Organization> .
<did:web:issuer.example> <https://schema.org/name> "ACME Customs Broker" .
🔥 <did:web:issuer.example> <https://www.w3.org/ns/credentials/issuer-dependent#_sd> "_Yj2C7jHHXj3DsmW_PA5GsSSdSS_DG7k6e_a4ZOuBTE" .
<http://supply-chain.example/credentials/dd0c6f9a-5df6-40a3-bb34-863cd1fda606> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://w3id.org/traceability#EntryNumberCredential> .
<http://supply-chain.example/credentials/dd0c6f9a-5df6-40a3-bb34-863cd1fda606> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://www.w3.org/2018/credentials#VerifiableCredential> .
<http://supply-chain.example/credentials/dd0c6f9a-5df6-40a3-bb34-863cd1fda606> <https://www.iana.org/assignments/jose#iss> <did:web:issuer.example> .
<http://supply-chain.example/credentials/dd0c6f9a-5df6-40a3-bb34-863cd1fda606> <https://www.iana.org/assignments/jwt#cnf> _:c14n0 .
<http://supply-chain.example/credentials/dd0c6f9a-5df6-40a3-bb34-863cd1fda606> <https://www.iana.org/assignments/jwt#exp> "1694793535"^^<https://www.w3.org/2001/XMLSchema#nonNegativeInteger> .
<http://supply-chain.example/credentials/dd0c6f9a-5df6-40a3-bb34-863cd1fda606> <https://www.iana.org/assignments/jwt#iat> "1692115135"^^<https://www.w3.org/2001/XMLSchema#nonNegativeInteger> .
<http://supply-chain.example/credentials/dd0c6f9a-5df6-40a3-bb34-863cd1fda606> <https://www.w3.org/2018/credentials#credentialSubject> _:c14n1 .
<http://supply-chain.example/credentials/dd0c6f9a-5df6-40a3-bb34-863cd1fda606> <https://www.w3.org/2018/credentials#issuer> <did:web:issuer.example> .
<http://supply-chain.example/credentials/dd0c6f9a-5df6-40a3-bb34-863cd1fda606> <https://www.w3.org/2018/credentials#validFrom> "2023-08-15T15:58:55.732Z"^^<http://www.w3.org/2001/XMLSchema#dateTime> .
<http://supply-chain.example/credentials/dd0c6f9a-5df6-40a3-bb34-863cd1fda606> <https://www.w3.org/2018/credentials#validUntil> "2023-09-15T15:58:55.732Z"^^<http://www.w3.org/2001/XMLSchema#dateTime> .
🔥 <http://supply-chain.example/credentials/dd0c6f9a-5df6-40a3-bb34-863cd1fda606> <https://www.w3.org/ns/credentials/issuer-dependent#_sd_alg> "sha-256" .
_:c14n0 <https://www.iana.org/assignments/jwt#jwk> "{\"crv\":\"P-384\",\"kty\":\"EC\",\"x\":\"D_jHmVwMTl6qw-49SiXJ599R6nptY3vbOzrCDB8THYFfIoTUIyvdipLGaMpTXnPH\",\"y\":\"hpBluDpuT68C17ruX0kd9NNOH9WMzlQ9WiGYd-xHCgFgb88wjP59ZVGcbmf1GIub\"}"^^<http://www.w3.org/1999/02/22-rdf-syntax-ns#JSON> .
_:c14n1 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://w3id.org/traceability#EntryNumber> .
🔥 _:c14n1 <https://www.w3.org/ns/credentials/issuer-dependent#_sd> "5mp8mbz0lAk_wKqO70mk_g2KFHu0TAC96IZRixckzRQ" .

```

### DIFF
--- a/contexts/credentials/v2
+++ b/contexts/credentials/v2
@@ -55,6 +55,15 @@
         }
       }
     },
+    "_sd_alg": {
+      "@id": "https://www.iana.org/assignments/jwt#_sd_alg"
+    },
+    "_sd": {
+      "@id": "https://www.iana.org/assignments/jwt#_sd"
+    },
+    "...": {
+      "@id": "https://www.iana.org/assignments/jwt#..."
+    },
 
     "VerifiableCredential": {
       "@id": "https://www.w3.org/2018/credentials#VerifiableCredential",


contexts/credentials/v2
```diff
@@ -55,6 +55,15 @@
         }
       }
     },
+    "_sd_alg": {
+      "@id": "https://www.iana.org/assignments/jwt#_sd_alg"
+    },
+    "_sd": {
+      "@id": "https://www.iana.org/assignments/jwt#_sd"
+    },
+    "...": {
```
msporny, line 64: "..." is a registered claim name?

msporny: answering my own questions... yep: https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-05.html#section-5.1

OR13: haha yes : ) its how they annotate "array" values.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-08-15-vcwg-special#section1-9) on 2023-08-15

- no resolutions were taken

#### 1.9. Add sd-jwt registered claim names to v2 context (pr vc-data-model#1242)

_See github pull request [vc-data-model#1242](https://github.com/w3c/vc-data-model/pull/1242)._




**Brent Zundel:** 7 hours old. happy to hear consideration [none]. moving to last PR 1242. add sd-jwt claims to the v2 context. Orie can you introduce?  

**Orie Steele:** similar to the other context changes which added reserved claim names by the securing mechanisms that are not data integrity, this adds reserved claim names from SD-JWT. a little different since it's not an RFC and could change. extra caution warranted.  
… would be great to hear from Kristina/Oliver if they don't intend to add these claim names. would give claim names consistent application in the v2 context.  

**Brent Zundel:** comments?  

**Manu Sporny:** only comment is the "..." claim used in SD-JWT which was surprising. dont think it creates any type of issue. seems like a valid fragment identifier. any plans to change it?  

**Manu Sporny:** first two seem stable, "..." is ???  

**Orie Steele:** the "..." is used to annotate elements of an array which are disclosed individually; don't know if it will change. up to the OAuth WG, since it's their adopted work item.  

**Brent Zundel:** we already have the "this may change during CR" caveat in place. we are OK there. if not registered in time, or registered differently, we can adjust.  
… any other comments on PRs before issue triage?  

</details>

Sakurann (approved):

peacekeeper:
No objection, but just curious, why would one ever need this? Why would you ever serialize a VC with "_sd_jwt" claims to N-Quads?

Sakurann:
you would serialize to N-quads a payload that you obtain after processing the disclosed claims, which does not contain _sd hashes.

peacekeeper:
> You would see this, when processing a partially disclosed payload as JSON-LD / RDF.

I think mixing JSON-LD with SD-JWT is a really bad idea. This will cause similar problems as we saw with the JWT "vc" claim in VCDM1.1. JSON-LD is NOT "just JSON".

- If you want to use the JWT data model, use JWT, JWS, SD-JWT, etc.
- If you want to use the JSON-LD VC data model, use Data Integrity.

Anyway, sorry this may be a bit off-topic.

brentzundel:
@peacekeeper The WG has already come to consensus around the `@context` file being a sort-of "catch-all" for the various additional uses that are anticipated for VCs, which is why, in addition to terms from the VC data model, it has Data Integrity, JWT, and many other terms.

A number of us suggested paring the `@context` down to just the VC Data Model terms, but that direction was rejected and instead we are following a "kitchen sink" model of inclusivity.

OR13:
Have you benchmarked this recommendation:

> If you want to use the JSON-LD VC data model, use Data Integrity.

We have... The performance is not very good.

We do use the jaon ld data model, even the rdf representations... Just not as a blocking dependency of sign or verify operations.

For context, ecdsa-sd uses both RDF canonicalization and multiple signatures, making it much more compute expensive then sd jwt... Those dependencies can also preclude use cases, like iot signers, etc...

VCDM V2 is much better about working with undefined terms than previous version or did core.

After this PR is merged, the terms will be defined properly just like they are for data integrity proofs.

The working group consensus so far seems to be to make thing easy for developers, by adding the terms they rely on for security, schema and status to the V2 context.

Only supporting data integrity would be an option, if we recommend that, I would suggest withdrawing the `vc-jose-cose` work item, since it would be clear the working group has a preference for securing formats.

We should also revise the current draft to eliminate the day 3 resolution and we can simplify a lot of the spec if it only has to focus on data integrity.

Of course all of this is also off topic and this PR just defines terms.

Related to your point about data integrity, you could use sd-jwt to reveal data integrity proofs, or multiple subjects.

The partially revealed documents won't need to have any default vocab or issuer defined terms after this PR is merged.

peacekeeper:
I see, thanks for explanations @brentzundel @OR13 .

TallTed:
> [@OR13] Have you benchmarked this recommendation:
>
>> [@peacekeeper] If you want to use the JSON-LD VC data model, use Data Integrity.
>
> [@OR13] We have... The performance is not very good.

Experience has shown that benchmarking remains at least as much of an art as a science, and that people who are unfamiliar and/or inexperienced with particular tools, data models and structures, languages, etc., often build benchmarks that are broken at the start — that are guaranteed to produce misleading results. Software that is not properly configured, data loads that are not appropriately structured, queries that are not reasonably optimized and/or written as expected to be written in live deployments, inappropriate and/or varying hardware beneath software, tests that are (un)intentionally written to (dis)favor some subset of elements being tested, etc., can all influence results, and make some things appear "good" and other things appear "bad" when neither label would be accurate for these things in appropriate deployment.

Benchmarking and other performance rating systems are worthy endeavors, but they should not be implemented in a vacuum nor without significant cross-pollination between companies, implementation teams, product experts, and others with deep understanding of what's being tested and why. It might be worthwhile to consider a new W3C CG, W3C WG, other standards development organization (SDO) focus group, or (optimally) cross-SDO group to delve into this area.

OR13:
@TallTed totally agree regarding benchmarking.

There are some obvious runtime differences, which makes benchmarking sorta pointless in this case, namely knowing how many signatures and hashes happen, and knowing that RDF canonicalization can time out.

TallTed:
> [@OR13] There are some obvious runtime differences, which makes benchmarking sorta pointless in this case, namely knowing how many signatures and hashes happen, and knowing that RDF canonicalization can time out.

"Knowing how many signatures and hashes happen" is a horrible handwave which just says "don't sign or hash; it can't be done fast enough".

"Knowing that RDF canonicalization can time out" is worse. *Anything* can time out, if the timeout is set low enough, or the processor is slow enough, or the data pipes are small enough and/or the data sets are large enough.

Formalize your test regimen and provide it along with specs of your test hardware, configuration details of your test software, the sample data sets used for your testing... And then we can have a semi-reasonable conversation about comparative performance (is this properly configured and tuned toolset faster than that properly configured and tuned toolset? how do they compare to humans performing the same taskset by hand? etc.), which is the only real reason for benchmarking.

mprorock (approved):

msporny:
@OR13 wrote:
> We have... The performance is not very good.

We've done the same and performance has always been acceptable. I'll note that we've designed systems that require upwards of 50M transactions per day, all using the stuff that is claimed to not have good performance, where the signing/verification times were dwarfed (by orders of magnitude) by the network flight times, application logic, and database access times.

The benchmarks that we've seen produced to date either mis-implemented something critical or focused on using synthetic data that was designed as a (detectable) attack rather than reflecting real world use cases. If you are having performance issues, we're happy to take a look, but you're going to have to produce some non-synthetic real world use case instead of repeating this "not very good" rhetoric.

All algorithms have design boundaries and trade-offs. We could go into the fact that SD-JWT requires one to issue an entirely new, JWT-incompatible data format... increasing the complexity of all the software in the ecosystem... you then need to pepper it with "decoys" because it has side-channel attacks, but in a way that keeps those pepperings below a statistically significant threshold. Again, some might be ok with making those trade-offs while others aren't. We could say that SD-JWTs "performance is not very good" in that respect... but that's not very helpful is it? It's important to understand the trade-offs one is making.

msporny (approved):
Usage of these properties should be explained in the VC-JOSE-COSE specification, specifically, how they are expected to be used by the VCDM. I'll note that there is no mention of the "..." property in VC-JOSE-COSE.

OR13:
Another thing to consider from a performance side is adding better RDF types to short circuit the cost of canonicalization, so for example, if you have a matrix, just declare it as `@json` in your second context, don't just assume that JSON-LD processing will handle data efficiently (it doesn't magically know how to organize your JSON), you need to look at the N-Quads produced, and the time it takes to produce them.

I wouldn't call a matrix a degenerate use case, but I would call it one of it's not declared as json in JSON-LD.

From a cryptanalysis perspective, I don't think the extra signatures on a single use public key really weaken ECDSA-SD, but as signatures and public key sizes grow, it can be nice to rely on hashes more, and signatures less, when that's possible.

Another dimension of performance to consider is how many well known vocabularies a wallet or verifier will support, and how to update them over time.

The maintenance cost for updates in those systems, is also part of the picture.

Of course all of this is off topic for this PR.

I don't think we expect any of the values in the V2 core context to be covered in tests for vc-jose-cose, since no rdf processing is required.

Since conforming documents are compact, all of the rdf processing time, is extraneous given, that it's not needed to secure a conforming document.

That's kind of the whole point of my performance comment.

Sakurann:
> the fact that SD-JWT requires one to issue an entirely new, JWT-incompatible data format... increasing the complexity of all the software in the ecosystem... 

SD-JWT is not a JWT, but if you look at the very-actively-emerging libraries, they all leverage existing JWT libraries, and succeed in keeping it simple.  

> you literally need to use a different data format for your credential based on if you secure it with JWT, SD-JWT, or JPT/JWP...

fortunately, or unfortunately, unlike SD-JWT, JWP is not stable enough to be used in production, yet.

OR13:
Since we are super off topic, one thing that bothers me about all of these systems, is they all exploit schemas differently.

Data integrity proofs canonicalize / use RDF.

SD-JWT issuers have to mark properties for selective disclosure somehow.

JWP has the transform content into lists.

Gordion solves this a different way that requires you encode your document differently.

The schema management and tooling costs are not trivial for all of these... And i am reminded that when you can, you should just sign the bytes... Everything else you do is complexity that has to pay for itself eventually

dlongley:
@OR13,

> The schema management and tooling costs are not trivial for all of these... And i am reminded that when you can, you should just sign the bytes... Everything else you do is complexity that has to pay for itself eventually

I agree these things aren't trivial, but I disagree with the "everything else" comment. The complexity is *always paid for*, it's just a question of how many times you pay and who pays.

If "you" (the signature library) just "signs the bytes", then someone else further up the stack pays the complexity cost. Maybe multiple someones. It's just choosing to not do the work that still needs to be done. I think this can be very easily misunderstood if only the perspective of a crypto library developer is considered. An approach that separates concerns without addressing sufficiently common ones simply pushes those problems to others, who may be great in number. Instead, good layering should maximize reuse and enable diversification of labor to help minimize how many times the complexity cost must be paid in aggregate in a system.

OR13:
Agree on the importance of layering.

Ideally you (the security expert) can pull the layers you need without pulling layers you don't, when building a security product.

Credential libraries should not feel like cable tv bundles.

Even if some people always want to watch news that is indistinguishable from entertainment :)

TallTed:
> Since we are super off topic, one thing that bothers me about all of these systems

It's not `one thing`, it's `another several things.`

Several of which amount to "the work of this WG is worthless; just write your data in whatever form you like, and sign the bits."

Please take that elsewhere. 

Another issue, another PR, another repository, another SDO... 

I'm sure there is an appropriate space.

But this PR isn't it.

OR13:
@TallTed which standards organization would be better to discuss securing arbitrary content consistently, without excessive schema based preprocessing?

Would you recommend the working group focus exclusively on data integrity proofs?

If the first examples in the core spec are always data integrity proofs, won't people conclude they are the recommended format anyway?

Why is schema based preprocessing an optimal way to secure a known content type like JSON-LD?

Seems like signing application/n-quads or application/ld+json directly is more efficient.

msporny:
Normative, multiple reviews, no changes requested, no objections, merging.

msporny:
@Sakurann wrote:
> SD-JWT is not a JWT, but if you look at the very-actively-emerging libraries, they all leverage existing JWT libraries, and succeed in keeping it simple.

No, they fail to keep things simple. Completely and utterly. :)

Sure, you got to re-use some library, but the complexity is pushed to the data format layer... which many more developers have to deal with. Now you've forced developers to have to deal with VC-1.1-JWTs, and VC-JOSEs, and VC-COSEs, and VC-JAdESes, and VC-SD-JWTs, and VC-JPTs or VC-JWPs... all completely different opaque encapsulating formats that will need further encapsulating formats when stored/retrieved... massive complexity that issuers, digital wallets, and verifiers will have to deal with. It's anything but simple.

Compare that w/ VC-DI protected VCs... which all have the same data format, where the complexity is shoved into the library layer where only a few developers have to mess with it... the format is JSON-based (and doesn't require extra layers of encapsulation to figure out what you're dealing with), and where the securing mechanisms can be layered/parallel... you don't need an entirely new format to support new crypto. Case in point JWTs failed to solve for what SD-JWTs solve for, which fail to solve for what JWP/JPTs will solve for... and each and every time, you will go back to the developer community and ask them to scrap the entire signature format for a new one. That is complexity incarnate. :)

> fortunately, or unfortunately, unlike SD-JWT, JWP is not stable enough to be used in production, yet.

... and when JWP /does/ become stable enough... there will be calls to abandon SD-JWT, just as there have been calls to abandon JWTs, because neither of the formats will support what JWPs will support... and once again, there will be a call to the developer community to switch up their data formats yet again.

OR13:
> I'd feel much better about all of this if the VC-JOSE-COSE spec could figure out what it's going to be about rather than a smattering of disconnected normative statements and a collection of issue markers and examples that raise more questions than provide concrete answers to implementers.

I agree.

Now that we don't have to try to support anything but JSON-LD it should be pretty easy to fix though... It's too bad that took so long.

The payloads that are secured are the media types defined in the core data model... And they are secured with RFCs / documents maintained outside this working group... Hence not a lot of normative statements... And lots of examples.

No rdf processing is required, so we don't have to write a bunch of normative statements that only apply to representations that are not conforming documents.

My favorite application of sd-jwt is selectively disclosing members of `@context`... That way verifiers see different RDF graphs based on the preferences of holders according to what the issuers have marked disclosable.

You can also do this with `@id` and `@type` and you get totally different graph shapes depending on what the holder decides to disclose.

You can't do this kind of thing with ecdsa-sd, because it operates on N-Quads, which are produced by applying the `@context`, meaning it's not a part of the data model, that is actually secured... Unless you use the resource protection mechanism, in which case there would need to be a hash for each possible disclosed context... And I don't think it works even then without some extra modification to ecdsa-sd.

JSON processing on the core data model has opened a wide door, for ignoring the rdf data model or augmenting it, in interesting ways.

It will be fun to see what implementers come up with, but my guess is most of them will use data integrity proofs since those are the examples in the core specification.

OR13:
I've implemented both ecdsa-sd and sd-jwt.

sd-jwt is much simpler, and you can use it to secure the entire conforming document, including the `@context`.

sd-jwt has also been contributed to and attacked by a much more rigorous process, and it's standards track by a working group that routinely delivers specifications used to secure consumer, corporate and government identity.

It's ok if JWP comes along and makes old less efficient stuff irrelevant... That's what happened to xhtml and URLs... It's ok to improve things, sometimes to do that you need to drop some dependencies.

msporny:
> I've implemented both ecdsa-sd and sd-jwt.

So, let me get this straight... a single developer implemented ecdsa-sd in a weekend... sounds simple enough to implement to me. :)

> sd-jwt is much simpler, and you can use it to secure the entire conforming document, including the `@context`.

The point continues to be missed. The problem isn't that SD-JWT is simpler... it is... but by being simpler in the way that it is, it pushes enormous complexity higher up the stack into the protocols, storage systems, and, in general, into the space of most application developers. This is the same mistake that JWK made by exposing things like `d` and `x` and `y` to the application layer -- the application layer has no business dealing w/ those formats -- nor do the developers working in application space. It pushed the complexity of managing that stuff into application-space and by doing so caused critical vulnerabilities that could have been avoided.

The point above is that: this is technology, and with any technology there are trade-offs, and SD-JWT continues to make the same bad trade-offs that JWTs do, but in a format that is not as generally useful as JWTs. ECDSA-SD, on the other hand, pushes that complexity DOWN the stack, not UP the stack, so that the people that have to deal with the complexity are the very small handful of developers writing cryptographic libraries and not exposing the much larger set of application developers to nuances and idiosyncrasies at the application-layer. 

Assertions that "SD-JWT is simple" continue to miss the point that the complexity SD-JWT creates is ecosystem complexity. The same is true for the VC-JOSE-COSE spec in general... there is now a proliferation of incompatible securing formats in that ecosystem and the problem keeps getting worse.

> It's ok if JWP comes along and makes old less efficient stuff irrelevant...

No, it's not ok if it requires the entire ecosystem to upgrade and change their formats yet again... VC-1.1-JWT -> VC-JWT -> VC-SD-JWT -> VC-JWP <-- every upgrade is wildly disruptive to the ecosystem because the encapsulating format changes each and every time. 

Ignoring this reality won't make it go away.

TallTed:
It would probably be good to capture and shift the discussion of the last few comments into a new Issue that may stay open a while longer than this already merged (and soon to be entirely ignored) PR.

Issue #1231, opened 2023-08-08
Minor change to validFrom requested


David-Chadwick:
validFrom is currently defined as

If present, the value of the validFrom [property ](https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1199/3a0e423...5d44e0e.html#dfn-property)MUST be a string value of an [ [XMLSCHEMA11-2 ](https://www.w3.org/TR/xmlschema11-2/#dateTime)] combined date-time string representing the date and time the credential  becomes valid, which could be a date and time in the future. Note that this value represents the earliest point in time at which the information associated with the credentialSubject [property ](https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1199/3a0e423...5d44e0e.html#dfn-property)becomes valid.

The grammatical meaning of this paragraph is that the credential becomes valid at some point in the future, but it does not cater for the past. Please replace with

If present, the value of the validFrom [property ](https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1199/3a0e423...5d44e0e.html#dfn-property)MUST be a string value of an [ [XMLSCHEMA11-2 ](https://www.w3.org/TR/xmlschema11-2/#dateTime)] combined date-time string representing the date and time the credential  either became valid, or will become valid i.e. a date and time in the future. Note that this value represents the earliest point in time at which the information associated with the credentialSubject [property ](https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1199/3a0e423...5d44e0e.html#dfn-property) is valid.

TallTed:
Tweaked replacement (NOTE: some URIs here and above point to `pr-preview` targets... this text cannot be cut-and-pasted; it must be manually merged into the existing doc):

If present, the value of the `validFrom` [property](https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1199/3a0e423...5d44e0e.html#dfn-property) MUST be an [[XMLSCHEMA11-2](https://www.w3.org/TR/xmlschema11-2/#dateTime)] combined date-time string value representing the date and time the credential became, or will become, valid. Note that this value represents the earliest point in time at which the information associated with the `credentialSubject` [property](https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1199/3a0e423...5d44e0e.html#dfn-property) is valid.

dlongley:
[...]

PR #1250, opened 2023-08-20
Apply range requirements to validFrom/validUntil.

msporny:
This PR is an attempt to address issue #1231 by applying value range requirements to `validFrom` and `validUntil` as requested by @David-Chadwick, @TallTed, and @dlongley.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1250.html" title="Last updated on Aug 27, 2023, 11:13 PM UTC (d0c9d3d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1250/f4a8ea1...d0c9d3d.html" title="Last updated on Aug 27, 2023, 11:13 PM UTC (d0c9d3d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1669,6 +1669,14 @@ the <code>validUntil</code> <a>property</a> for expressing the date and time
 when a <a>credential</a> ceases to be valid.
         </p>
 
+        <p>
+When comparing dates and times, the calculation is done "temporally", which
+means that the string value is converted to a "temporal value" which exists
+as a point on a timeline. Temporal comparisons are then performed by checking
+to see where the date and time being compared is in relation to
+a particular point on the timeline.
+        </p>
+
         <dl>
           <dt><var id="defn-validFrom">validFrom</var></dt>
           <dd>
@@ -1679,7 +1687,7 @@ an [<a data-cite="XMLSCHEMA11-2#dateTime">XMLSCHEMA11-2</a>]
 Note that this value represents the earliest point in time
 at which the information associated with the <code>credentialSubject</code>
 <a>property</a> becomes valid. If a `validUntil` value also exists, the
-`validFrom` value MUST express a datetime that is temporally the same or 
+`validFrom` value MUST express a datetime that is temporally the same or
 earlier than the datetime expressed by the `validUntil` value.
           </dd>
           <dt><var id="defn-validUntil">validUntil</var></dt>
--- a/index.html
+++ b/index.html
@@ -1679,8 +1679,8 @@ an [<a data-cite="XMLSCHEMA11-2#dateTime">XMLSCHEMA11-2</a>]
 Note that this value represents the earliest point in time
 at which the information associated with the <code>credentialSubject</code>
 <a>property</a> becomes valid. If a `validUntil` value also exists, the
-`validFrom` value MUST be temporally the same or earlier than the `validUntil`
-value.
+`validFrom` value MUST express a datetime that is temporally the same or 
+earlier than the datetime expressed by the `validUntil` value.
           </dd>
           <dt><var id="defn-validUntil">validUntil</var></dt>
           <dd>
@@ -1690,8 +1690,9 @@ If present, the value of the <code>validUntil</code> <a>property</a> MUST be an
 <a>credential</a> ceases to be valid, which could be a date and time in the
 past. Note that this value represents the latest point in time at which the
 information associated with the <code>credentialSubject</code> <a>property</a>
-is valid. If a `validFrom` value also exists, the `validUntil` value MUST be
-temporally the same or later than the `validFrom` value.
+is valid. If a `validFrom` value also exists, the `validUntil` value MUST
+express a datetime that is temporally the same or later than the datetime
+expressed by the `validFrom` value.
           </dd>
         </dl>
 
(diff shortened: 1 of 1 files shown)


dlongley (approved):

Wind4Greg (approved):

iherman (approved):

index.html
```diff
@@ -1674,7 +1674,9 @@ <h3>Validity Period</h3>
 <a>credential</a> becomes valid, which could be a date and time in the future.
 Note that this value represents the earliest point in time
 at which the information associated with the <code>credentialSubject</code>
-<a>property</a> becomes valid.
+<a>property</a> becomes valid. If a `validUntil` value also exists, the
+`validFrom` value MUST be temporally the same or earlier than the `validUntil`
```
decentralgabe, line 1678: I wonder if `temporally the same` could be confusing (both to readers and implementers) since it allows for a different format value that complies with the format restrictions, if the time value is the same, I think...?

Like -- I could write a date as `2002-04-20T05:55:55Z` or `2020-04-20T05:55:55+00:00` and they're temporally the same, though doing this may increase risk an implementation breaks for some reason. Probably not a big concern.

If you agree there is a potential risk firming up the language to "the same" would remove this

decentralgabe (approved):

OR13 (approved):

brentzundel (approved):

index.html
```diff
@@ -1674,7 +1674,9 @@ <h3>Validity Period</h3>
...
+value.
```
TallTed, line 1679: @decentralgabe @David-Chadwick @msporny — What do you think? Parallel change also applied below...
```suggestion
<a>property</a> becomes valid. If a `validUntil` value also exists, the
`validFrom` value MUST express a datetime that is temporally the same or 
earlier than the datetime expressed by the `validUntil` value.
```

index.html
```diff
@@ -1684,8 +1686,8 @@ <h3>Validity Period</h3>
 <a>credential</a> ceases to be valid, which could be a date and time in the
 past. Note that this value represents the latest point in time at which the
 information associated with the <code>credentialSubject</code> <a>property</a>
-is valid. If a `validFrom` value exists, the `validUntil` value MUST be
-temporally greater than the `validFrom` value.
+is valid. If a `validFrom` value also exists, the `validUntil` value MUST be
+temporally the same or later than the `validFrom` value.
```
TallTed, line 1690:
```suggestion
is valid. If a `validFrom` value also exists, the `validUntil` value MUST
express a datetime that is temporally the same or later than the datetime
expressed by the `validFrom` value.
```

TallTed (changes requested):
Editorial tweaks to address @decentralgabe's concern.

index.html
msporny: >  it allows for a different format value that complies with the format restrictions, if the time value is the same

Yes, exactly, we want that; that is good. :)

I believe the whole point of saying "temporally the same" is to indicate that the comparison is happening in value space, not lexical space. If an implementation breaks in value space when two lexical values should be "temporally the same", then that's an implementation bug in the time value processing code and we tend to not write advice for potential implementation bugs at that layer (unless they're common, which they're not, IMHO).

We could say "the same time value (note that different lexical time strings can represent the same temporal value)"... but that's a bit of a mouthful and raises even more questions that would require further explanation about why comparing time values is a difficult and complicated thing (when you get down into the time processing library code).

All that to say, I think we're good here, unless I missed something and you can provide some concrete text that would address your concern?

OR13 (approved):
Might be good to *briefly elaborate on "temporally"... otherwise seems fine.

index.html
decentralgabe: agree on the small risk, which is an impl bug, but my concern is non blocking

index.html
```diff
@@ -1674,7 +1674,9 @@ <h3>Validity Period</h3>
 <a>credential</a> becomes valid, which could be a date and time in the future.
```
TallTed, line 1674:
```suggestion
<a>credential</a> becomes valid, which could be a date and time in the future or past.
```

TallTed (changes requested):
to address @David-Chadwick's reiterated concern

longpd (approved):

index.html
```diff
@@ -1684,8 +1686,9 @@ <h3>Validity Period</h3>
 <a>credential</a> ceases to be valid, which could be a date and time in the
 past. Note that this value represents the latest point in time at which the
```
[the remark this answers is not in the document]
David-Chadwick: "or in the future" preferred

David-Chadwick (changes requested):

index.html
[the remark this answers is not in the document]
TallTed: Change suggestion updated.

index.html
TallTed: Change suggestion updated.

index.html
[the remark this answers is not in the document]
David-Chadwick: Thanks

shigeya (approved):

msporny:
> Might be good to *briefly elaborate on "temporally"... otherwise seems fine.

Done in d0c9d3db7d7ab061060beccb199e895b4a1cfd9a.

msporny:

Issue #1175, opened 2023-06-29
Ensure the base context doesn't constrain lower-maturity specifications


jyasskin:
https://github.com/w3c/vc-data-model/pull/1158 will freeze the content of the base context (https://www.w3.org/ns/credentials/v2) when vc-data-model goes to REC. The base context currently includes definitions for terms like `DataIntegrityProof`, which is defined in https://www.w3.org/TR/vc-data-integrity/#dataintegrityproof rather than in the spec that's freezing the context, so if vc-data-integrity goes to REC _after_ vc-data-model, there's a chance that it would want to change the definition and wouldn't be able to because it was already frozen.

I think that implies that the vc-data-model REC should be delayed until all of the terms its context defines are also defined by Recommendations (or equivalent maturity levels published by other standards bodies). This may not actually be a delay if the specs are finished in the right order.

OR13:
We discussed not including status list or data integrity terms in the v2 context.

The working group decided to include those terms to "make it easier for developers".

I don't agree that including the terms makes it easier for developers.

If we are at risk of delaying the core data model if there are objections to data integrity or status list, I prefer to not include them in the v2 context.

selfissued:
It would be much cleaner to remove all dependencies on the "securing" specs from the core data model.  I suggest that we do this before going to Candidate Recommendation.

mprorock:
I think the best approach is to only include terms in the core context that are defined in the core context - other specs can provide their own contexts if required

mprorock:
The more I dig into this @brentzundel @Sakurann I think we need to prioritize ensuring nothing in the core data model is dependent in any normative fashion (directly or indirectly) on work outside the core data model.  Otherwise an issue with another spec could prevent the core data model from advancing

OR13:
Potential solution discussed in

https://github.com/w3c/vc-data-model/pull/1149#issuecomment-1636302571

msporny:
PR #1251 has been raised to address this issue. This issue will be closed once PR #1251 has been merged.

PR #1251, opened 2023-08-20
Outline plan to mitigate normative context risk when transitioning to Proposed Recommendation

msporny:
This PR attempts to address issue #1175 by outlining the mitigation strategy for ensuring that the contents in the base context don't impact the various specifications being worked on by this WG from transitioning to Recommendation status in a timely way.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1251.html" title="Last updated on Aug 20, 2023, 9:33 PM UTC (13e8004)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1251/03b0d90...13e8004.html" title="Last updated on Aug 20, 2023, 9:33 PM UTC (13e8004)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5451,6 +5451,17 @@ franchise. Policy information expressed by the <a>issuer</a> in the
 This section lists cryptographic hash values that might change during the
 Candidate Recommendation phase based on implementer feedback that requires
 the referenced files to be modified.
+<br><br>
+The Working Group is expecting all of the terms and URLs supplied in the
+JSON-LD Context to be either stabilized, or removed, before the publication of
+this specification as a Proposed Recommendation. While that means that this
+specification could be delayed if dependencies such as [[?VC-DATA-INTEGRITY]],
+[[?VC-JOSE-COSE]], SD-JWT, [[?VC-JSON-SCHEMA-2023]], or status list
+do not enter the Proposed Recommendation phase around the same time frame, the
+Working Group is prepared to remove the dependencies if an undue burden is
+placed on transitioning to the Recommendation phase. This is a calculated
+risk that the Working Group is taking and has a mitigation strategy in place
+to ensure the timely transition of this specification to a Recommendation.
         </p>
         <p>
 Implementations MUST ensure that the base context value, located at


Wind4Greg (approved):

index.html
```diff
@@ -5451,6 +5451,17 @@ <h3>Base Context</h3>
 This section lists cryptographic hash values that might change during the
 Candidate Recommendation phase based on implementer feedback that requires
 the referenced files to be modified.
+<br><br>
+The Working Group is expecting all of the terms and URLs supplied in the
+JSON-LD Context to be either stabilized, or removed, before the publication of
+this specification as a Proposed Recommendation. While that means that this
+specification could be delayed if dependencies such as [[?VC-DATA-INTEGRITY]],
+[[?VC-JOSE-COSE]], SD-JWT, [[?VC-JSON-SCHEMA-2023]], or status list
```
iherman, line 5459:
```suggestion
[[?VC-JOSE-COSE]], [[?VC-JSON-SCHEMA-2023]], or status list
```

At this moment at least, SD-JWT is not even on the list of the [official publications of the WG](https://www.w3.org/groups/wg/vc/publications/), i.e., talking about the PR phase of that document is meaningless.

msporny: PR #1242 adds SD-JWT terms to a normative document (the base context) that the group is publishing. It highlights the problem that @jyasskin raised... we now have a normative dependency on a document at IETF that is almost certainly not going to be finished on the timeline that the VCWG is on. While it's not on the official publication of the WG, the VC-JOSE-COSE specification, and the base context have a normative dependency on it.

Again, the base context having a dependency on it isn't a big deal... by the time we hit REC (Summer 2024), SD-JWT should be stable enough (in the properties that it uses)... and if it's not, the group can remove references to it from the base context (which is what this PR says for those properties, and any other unstable properties).

index.html
```diff
@@ -5451,6 +5451,17 @@ <h3>Base Context</h3>
 This section lists cryptographic hash values that might change during the
 Candidate Recommendation phase based on implementer feedback that requires
 the referenced files to be modified.
+<br><br>
+The Working Group is expecting all of the terms and URLs supplied in the
+JSON-LD Context to be either stabilized, or removed, before the publication of
```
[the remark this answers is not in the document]
msporny: We could do the same thing that we did here:

https://w3c.github.io/vc-data-integrity/#contexts-and-vocabularies

Looking for feedback from reviewers if they want to do that.

index.html
iherman: Ok. 

B.t.w., let us be precise. By believe that in this context (sic!) our dependency is _on the URL-s_ and not on the details of the respective specifications. To take an example, we have a dependency on `http://schema.org/name` but not on the content of the HTML page that describes that particular term. Maybe it is worth emphasizing this, because the text reads as we had a normative dependency on the respective specification as a whole. Is that correct?

And yes, the SD-JWT specification might be in flux, but the URLs we use may not...

decentralgabe (approved):

brentzundel (approved):

dlongley (approved):

index.html
msporny: Yes, correct, our dependency is on the URLs, primarily. 

@jyasskin's point was that we could accidentally create a bi-directional dependency. Not only do we lock in the terms and URLs from our standpoint, but we could make it such that the spec that we reference then couldn't change the terms/URLs if they wanted to (because we'd ship something that they weren't ready to ship). 

For example, if SD-JWT decides to change the `...` term to `o_O`, because we VCWG shipped with `...`, that might cause them to need to support both properties in their spec for backwards compatibility reasons, which is something they wouldn't do if we didn't include those properties in our base context.

The mitigation to all of that is just keeping all of the groups in sync w/ one another and being ready to pull the terms that are not stable enough as we go into Proposed Rec. We have 9-ish more months for all of this to stabilize, so making the decision now is premature... we'll reserve the right to make the decision later (which is what this PR attempts to say).

index.html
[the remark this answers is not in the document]
OR13: I would recommend only focusing on the JSON-LD / Turtle files.

index.html
OR13: yes, the v2 context has terms from several related documents, if any of them stall / fail or are blocked indefinitely, the bundling strategy could bite us, the list includes:

- vc-data-integrity
- vc-status-list
- vc-json-schema
- sd-jwt (IETF)

Possible there are other documents that could cause problems beyond these.

vc-jose-cose does not define any terms, hence it has nothing appear in the v2 context.

OR13 (approved):

David-Chadwick (approved):

index.html
msporny: > vc-jose-cose does not define any terms, hence it has nothing appear in the v2 context, but normative dependency issues would still exist.

Debatable... One of our specs should speak to the fact of all the ".../jose#" and ".../jwt#" entries in the core context now... same for SD-JWT properties. If the JOSE group decides they don't want to use those URLs, or SD-JWT changes (or vice versa), it creates the same issue... and the resolution is the same as all the other specs -- features will be cut/removed if they get in the way for REC.

index.html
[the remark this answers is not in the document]
msporny: I raised https://github.com/w3c/vc-data-model/issues/1261 to track @iherman's request.

msporny:
Editorial, multiple reviews, changes requested (and tracked in another issue), no objections, merging.

Issue #1214, opened 2023-07-25
Spec does not contain "name" and "description" terms in the context


aljones15:
```js
{
  "@context": {
    "@protected": true,
    "@vocab": "https://www.w3.org/ns/credentials/issuer-dependent#",

    "id": "@id",
    "type": "@type",
    "kid": {
      "@id": "https://www.iana.org/assignments/jose#kid",
      "@type": "@id"
    },
    "iss": {
      "@id": "https://www.iana.org/assignments/jose#iss",
      "@type": "@id"
    },
    "sub": {
      "@id": "https://www.iana.org/assignments/jose#sub",
      "@type": "@id"
    },
    "jku": {
      "@id": "https://www.iana.org/assignments/jose#jku",
      "@type": "@id"
    },
    "x5u": {
      "@id": "https://www.iana.org/assignments/jose#x5u",
      "@type": "@id"
    },

    "VerifiableCredential": {
      "@id": "https://www.w3.org/2018/credentials#VerifiableCredential",
      "@context": {
        "@protected": true,

        "id": "@id",
        "type": "@type",

        "credentialSchema": {
          "@id": "https://www.w3.org/2018/credentials#credentialSchema",
          "@type": "@id"
        },
        "credentialStatus": {
          "@id": "https://www.w3.org/2018/credentials#credentialStatus",
          "@type": "@id"
        },
        "credentialSubject": {
          "@id": "https://www.w3.org/2018/credentials#credentialSubject",
          "@type": "@id"
        },
        "description": {
          "@id": "https://schema.org/description"
        },
        "evidence": {
          "@id": "https://www.w3.org/2018/credentials#evidence",
          "@type": "@id"
        },
        "validFrom": {
          "@id": "https://www.w3.org/2018/credentials#validFrom",
          "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
        },
        "validUntil": {
          "@id": "https://www.w3.org/2018/credentials#validUntil",
          "@type": "http://www.w3.org/2001/XMLSchema#dateTime"
        },
        "issuer": {
          "@id": "https://www.w3.org/2018/credentials#issuer",
          "@type": "@id"
        },
        "name": {
          "@id": "https://schema.org/name"
        },
        "proof": {
          "@id": "https://w3id.org/security#proof",
          "@type": "@id",
          "@container": "@graph"
        },
        "refreshService": {
          "@id": "https://www.w3.org/2018/credentials#refreshService",
          "@type": "@id"
[...]

PR #1252, opened 2023-08-20
Add section about Names and Descriptions of credentials.

msporny:
This PR is an attempt to address issue #1214 by adding a section that defines the `name` and `description` properties.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1252.html" title="Last updated on Aug 29, 2023, 8:09 PM UTC (1974cd2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1252/dbd67ab...1974cd2.html" title="Last updated on Aug 29, 2023, 8:09 PM UTC (1974cd2)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1514,11 +1514,11 @@ An OPTIONAL property that expresses the name of the <a>credential</a>. If
 present, the value of the <code>name</code> <a>property</a> MUST be a string or
 an object. If the value is an object, the object MUST contain a `value` property
 whose value is a string, and SHOULD contain a `lang` property whose value is a
-`langtag` defined in [[BCP47]], and MAY contain a `dir` property whose value is
-a text direction code defined by the `dir` attribute in [[HTML5]]. Ideally, the
-name of a <a>credential</a> is concise, human-readable, and could be used to
-enable an individual to quickly differentiate the <a>credential</a> from other
-<a>credentials</a> that they might hold.
+`langtag` string defined in [[BCP47]], and MAY contain a `dir` property whose
+value is a text direction string defined by the `dir` attribute in [[HTML5]].
+Ideally, the name of a <a>credential</a> is concise, human-readable, and could
+be used to enable an individual to quickly differentiate the <a>credential</a>
+from other <a>credentials</a> that they might hold.
           </dd>
           <dt><dfn class="lint-ignore">description</dfn></dt>
           <dd>
@@ -1526,10 +1526,10 @@ An OPTIONAL property that conveys specific details about a <a>credential</a>. If
 present, the value of the <code>description</code> <a>property</a> MUST be a
 string or an object. If the value is an object, the object MUST contain a
 `value` property whose value is a string, and SHOULD contain a `lang` property
-whose value is a `langtag` defined in [[BCP47]], and MAY contain a `dir`
-property whose value is a text direction code defined by the `dir` attribute in
-[[HTML5]]. Ideally, the description of a <a>credential</a> is no longer than a
-few sentences in length and conveys enough information about the
+whose value is a `langtag` string defined in [[BCP47]], and MAY contain a `dir`
+property whose value is a text direction string defined by the `dir` attribute
+in [[HTML5]]. Ideally, the description of a <a>credential</a> is no longer than
+a few sentences in length and conveys enough information about the
 <a>credential</a> to remind an individual as to its contents without having to
 look through the entirety of the <a>claims</a>.
           </dd>
--- a/index.html
+++ b/index.html
@@ -1520,8 +1520,8 @@ be used to enable an individual to quickly differentiate the
           <dd>
 An OPTIONAL property that conveys specific details about a <a>credential</a>.
 If present, the value of the <code>description</code> <a>property</a> MUST be a
-string. Ideally, the description of a <a>credential</a> is a few sentences
-in length and conveys enough information about the
+string. Ideally, the description of a <a>credential</a> is no longer than a
+few sentences in length and conveys enough information about the
 <a>credential</a> to remind an individual as to its contents without having
 to look through the entirety of the <a>claims</a>.
           </dd>
--- a/index.html
+++ b/index.html
@@ -1511,19 +1511,27 @@ are provided for these purposes.
           <dt><dfn class="lint-ignore">name</dfn></dt>
           <dd>
 An OPTIONAL property that expresses the name of the <a>credential</a>. If
-present, the value of the <code>name</code> <a>property</a> MUST be a string.
-Ideally, the name of a <a>credential</a> is concise, human-readable, and could
-be used to enable an individual to quickly differentiate the
-<a>credential</a> from other <a>credentials</a> that they might hold.
+present, the value of the <code>name</code> <a>property</a> MUST be a string or
+an object. If the value is an object, the object MUST contain a `value` property
+whose value is a string, and SHOULD contain a `lang` property whose value is a
+`langtag` defined in [[BCP47]], and MAY contain a `dir` property whose value is
+a text direction code defined by the `dir` attribute in [[HTML5]]. Ideally, the
+name of a <a>credential</a> is concise, human-readable, and could be used to
+enable an individual to quickly differentiate the <a>credential</a> from other
+<a>credentials</a> that they might hold.
           </dd>
           <dt><dfn class="lint-ignore">description</dfn></dt>
           <dd>
-An OPTIONAL property that conveys specific details about a <a>credential</a>.
-If present, the value of the <code>description</code> <a>property</a> MUST be a
-string. Ideally, the description of a <a>credential</a> is no longer than a
+An OPTIONAL property that conveys specific details about a <a>credential</a>. If
+present, the value of the <code>description</code> <a>property</a> MUST be a
+string or an object. If the value is an object, the object MUST contain a
+`value` property whose value is a string, and SHOULD contain a `lang` property
+whose value is a `langtag` defined in [[BCP47]], and MAY contain a `dir`
+property whose value is a text direction code defined by the `dir` attribute in
+[[HTML5]]. Ideally, the description of a <a>credential</a> is no longer than a
 few sentences in length and conveys enough information about the
-<a>credential</a> to remind an individual as to its contents without having
-to look through the entirety of the <a>claims</a>.
+<a>credential</a> to remind an individual as to its contents without having to
+look through the entirety of the <a>claims</a>.
           </dd>
         </dl>
 
--- a/index.html
+++ b/index.html
@@ -1501,10 +1501,10 @@ Implementation Guidelines [[?VC-IMP-GUIDE]].
 
         <p>
 When displaying a <a>credential</a>, it can be useful to have
-text provided by the <a>issuer</a> that provides the
-<a>credential</a> with a name as well as a short description of the
-purpose of the <a>credential</a>. The `name` and `description` <a>properties</a>
-are provided for these purposes.
+text provided by the <a>issuer</a> that furnishes the
+<a>credential</a> with a name as well as a short description of its
+purpose. The `name` and `description` <a>properties</a>
+are meant to serve these purposes.
         </p>
 
         <dl>
@@ -1514,11 +1514,12 @@ An OPTIONAL property that expresses the name of the <a>credential</a>. If
 present, the value of the <code>name</code> <a>property</a> MUST be a string or
 an object. If the value is an object, the object MUST contain a `value` property
 whose value is a string, and SHOULD contain a `lang` property whose value is a
-`langtag` string defined in [[BCP47]], and MAY contain a `dir` property whose
+string containing a well-formed `Language-Tag` as defined by [[BCP47]],
+and MAY contain a `dir` property whose
 value is a text direction string defined by the `dir` attribute in [[JSON-LD]].
 Ideally, the name of a <a>credential</a> is concise, human-readable, and could
-be used to enable an individual to quickly differentiate the <a>credential</a>
-from other <a>credentials</a> that they might hold.
+enable an individual to quickly differentiate one <a>credential</a>
+from any other <a>credentials</a> that they might hold.
           </dd>
           <dt><dfn class="lint-ignore">description</dfn></dt>
           <dd>
@@ -1526,11 +1527,12 @@ An OPTIONAL property that conveys specific details about a <a>credential</a>. If
 present, the value of the <code>description</code> <a>property</a> MUST be a
 string or an object. If the value is an object, the object MUST contain a
 `value` property whose value is a string, and SHOULD contain a `lang` property
-whose value is a `langtag` string defined in [[BCP47]], and MAY contain a `dir`
+whose value is a string containing a well-formed language tag as defined
+by [[BCP47]], and MAY contain a `dir`
 property whose value is a text direction string defined by the `dir` attribute
-in [[JSON-LD]]. Ideally, the description of a <a>credential</a> is no longer than
+in [[JSON-LD]]. Ideally, the description of a <a>credential</a> is no more than
 a few sentences in length and conveys enough information about the
-<a>credential</a> to remind an individual as to its contents without having to
+<a>credential</a> to remind an individual of its contents without their having to
 look through the entirety of the <a>claims</a>.
           </dd>
         </dl>
@@ -1567,7 +1569,7 @@ look through the entirety of the <a>claims</a>.
 Names and descriptions also support expressing content in different languages.
 To express a string with language and text direction information, one can use
 an object that contains the `value`, `lang`, and `dir` properties to express
-the text value, language code, and text direction, respectively.
+the text value, language tag, and text direction, respectively.
 See <a href="#language-and-base-direction"></a> for further information.
         </p>
 
--- a/index.html
+++ b/index.html
@@ -1515,7 +1515,7 @@ present, the value of the <code>name</code> <a>property</a> MUST be a string or
 an object. If the value is an object, the object MUST contain a `value` property
 whose value is a string, and SHOULD contain a `lang` property whose value is a
 `langtag` string defined in [[BCP47]], and MAY contain a `dir` property whose
-value is a text direction string defined by the `dir` attribute in [[HTML5]].
+value is a text direction string defined by the `dir` attribute in [[JSON-LD]].
 Ideally, the name of a <a>credential</a> is concise, human-readable, and could
 be used to enable an individual to quickly differentiate the <a>credential</a>
 from other <a>credentials</a> that they might hold.
@@ -1528,7 +1528,7 @@ string or an object. If the value is an object, the object MUST contain a
 `value` property whose value is a string, and SHOULD contain a `lang` property
 whose value is a `langtag` string defined in [[BCP47]], and MAY contain a `dir`
 property whose value is a text direction string defined by the `dir` attribute
-in [[HTML5]]. Ideally, the description of a <a>credential</a> is no longer than
+in [[JSON-LD]]. Ideally, the description of a <a>credential</a> is no longer than
 a few sentences in length and conveys enough information about the
 <a>credential</a> to remind an individual as to its contents without having to
 look through the entirety of the <a>claims</a>.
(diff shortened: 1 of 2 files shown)


Wind4Greg (approved):
Nice and clear.

index.html
```diff
@@ -1492,6 +1492,62 @@ <h3>Types</h3>
...
+Ideally, the name of a <a>credential</a> is concise, human-readable, and could
+be used to enable an individual to quickly differentiate the
+<a>credential</a> from other <a>credentials</a> that they might hold.
+          </dd>
+          <dt><dfn class="lint-ignore">description</dfn></dt>
+          <dd>
+An OPTIONAL property that conveys specific details about a <a>credential</a>.
+If present, the value of the <code>description</code> <a>property</a> MUST be a
+string. Ideally, the description of a <a>credential</a> is a few sentences
+in length and conveys enough information about the
```
dlongley, line 1520:
```suggestion
string. Ideally, the description of a <a>credential</a> is no longer than a
few sentences in length and conveys enough information about the
```

dlongley (approved):

iherman (approved):
I do not have problems with the text there. However, we know that these properties are, in fact, schema.org properties that we adopt (which is perfectly fine, no reason to reinvent the wheel!). I wonder whether we should keep it hidden (as it is now, nobody will see it unless he/she looks into the context file) or whether we should add an informative note making that clear. I am in favor of the latter. Otherwise, the question will pop up later (just as it did in #1214, but for the vocabulary file).

peacekeeper (approved):

msporny:
> we should add an informative note making that clear. I am in favor of the latter.

If we do that, should we do that for every vocabulary URL in the base context? What do we say about the JWT-derived vocabulary URLs (that don't resolve today)... what about the SD-JWT ones... the Data Integrity ones? It feels like it's a detail that might be distracting... or perhaps we can have a section in the appendix that notes all of the vocabularies that the base context relies on?

iherman:
> perhaps we can have a section in the appendix that notes all of the vocabularies that the base context relies on?

that sounds like a good compromise. And we should probably which of those are _not_ under the control of this WG...

aphillips:
I won't comment on the text, which looks fine for what it is trying to do. I am slightly concerned that the example does not include an \@context indicating the language (or direction) of the name and direction field.

An I18N concern about natural language fields like `name` and `description` is the implicit assumption that a credential issued by some authority in one language will match the needs of all of its consumers or that, since a given credential can only be in one language, we might see goofy multilingual descriptions or issuers generating multiple credentials that differ only in language.

msporny:
> I won't comment on the text, which looks fine for what it is trying to do. I am slightly concerned that the example does not include an @context indicating the language (or direction) of the name and direction field.

Good point, I'll try to work such an example in...

> An I18N concern about natural language fields like `name` and `description` is the implicit assumption that a credential issued by some authority in one language will match the needs of all of its consumers or that, since a given credential can only be in one language, we might see goofy multilingual descriptions or issuers generating multiple credentials that differ only in language.

Sadly, it looks like some in the ecosystem are going to do the "issuers generating multiple credentials that differ only in language." approach given that they don't want to go to the trouble of defining their JSON-LD Contexts. Given that JSON doesn't have a general solution for a number of i18n use cases, and that some organizations are unaware and unwilling to take i18n into account (because they are only deploying in one homogeneous region, there is only so much we can do to nudge organizations to do the right thing. I'm afraid that, we're going to be forced into a "lowest common denominator" position no matter what we do here. We're just dealing w/ companies that don't have adequate i18n experience.

Perhaps what we should do is assert this design pattern for the `name` and `description` properties in the spec:

https://w3c.github.io/vc-data-model/#language-and-base-direction

at least doing that would provide a solution for applications performing JSON-LD processing and applications that do JSON processing can rely on the design pattern: https://w3c.github.io/vc-data-model/#json-processing

That pattern is vital for countries and regions that are legally compelled to publish content in multiple languages. 

Thanks for the nudge @aphillips -- I'll rework the PR to not only include a multi-lingual example, but ensure that the base VC context supports it as well.

decentralgabe (approved):

OR13 (changes requested):
Based on the thread on the PR, I think we are expecting to see a comment on i18n ?

OR13:
Concrete paths forward:

1. add an issue / marker / advisement / note
1. figure out some kind of "strings file" recommendation (its been discussed in other issues)
1. use the JSON-LD `@lang` feature (this seems to not be a thing, based on previous discussions).

index.html
```diff
@@ -1492,6 +1492,62 @@ <h3>Types</h3>
...
+        <p>
+When displaying a <a>credential</a>, it can be useful to have
+text provided by the <a>issuer</a> that provides the
+<a>credential</a> with a name as well as a short description of the
+purpose of the <a>credential</a>. The `name` and `description` <a>properties</a>
+are provided for these purposes.
+        </p>
+
+        <dl>
+          <dt><dfn class="lint-ignore">name</dfn></dt>
```
OR13, line 1507: why lint ignore?

index.html
```diff
@@ -1492,6 +1492,62 @@ <h3>Types</h3>
...
+          title="Usage of the name and description property"
+          data-vc-vm="https://university.example/issuers/565049#key-1">
+{
+  "@context": [
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
+  ],
+  "id": "http://university.example/credentials/3732",
+  <span class="highlight">"type": ["VerifiableCredential", "ExampleDegreeCredential"]</span>,
+  "issuer": "https://university.example/issuers/565049",
```
OR13, line 1536:
```suggestion
  "issuer": {
    "id": "https://university.example/issuers/565049",
    "name": "Example University",
    "description": "🧑‍🎓 Dàxué"
  },
```

index.html
```diff
@@ -1492,6 +1492,62 @@ <h3>Types</h3>
...
+  "validFrom": "2015-05-10T12:30:00Z",
+  "name": "Example University Degree",
+  "description": "2015 Bachelor of Science and Arts Degree from Example University."
```
OR13, line 1539:
```suggestion
  "description": "2015 Bachelor of Science and Arts Degree"
```

index.html
msporny: The term is not /internally/ referenced anywhere else in the specification, and if ReSpec detects that, it throws warnings, which triggers Echidna to not publish the specification (because there are warnings). 

We do reference this term externally, from the vocabulary, so we want to keep the DFN and we use `lint-ignore` to silence the ReSpec warning that's thrown (by telling ReSpec that we really do mean to define the term).

index.html
msporny: What does "🧑‍🎓 Dàxué" mean? I'm fine w/ expanding the issuer to include name and description, but the description should be something that's not going to confuse readers.

index.html
```diff
@@ -1492,6 +1492,62 @@ <h3>Types</h3>
...
+<a>credential</a> to remind an individual as to its contents without having
+to look through the entirety of the <a>claims</a>.
+          </dd>
+        </dl>
+
+        <pre class="example nohighlight vc"
+          title="Usage of the name and description property"
+          data-vc-vm="https://university.example/issuers/565049#key-1">
+{
+  "@context": [
```
aphillips, line 1530: Add `"@language": "en"` ?

aphillips (changes requested):
Still no example of language tagging the values and no mention of providing language metadata, localization, or the like. These definitions probably shouldn't dwell on the topic, but since they are the main source of natural language text in a credential, some mention close to the definition is probably appropriate.

msporny:
@aphillips wrote:
> Still no example of language tagging the values and no mention of providing language metadata, localization, or the like. 

Correct, I haven't had the chance to add this language yet... I plan on doing that today/tomorrow.

index.html
msporny: Fixed in 390360d0f3a2cf5f8f3d4510e8e96f5d8650ca51.

index.html
msporny: Fixed in 390360d0f3a2cf5f8f3d4510e8e96f5d8650ca51.

msporny:
@aphillips wrote:
> Still no example of language tagging the values and no mention of providing language metadata, localization, or the like.
@OR13 wrote:
>  I think we are expecting to see a comment on i18n ?

I've updated the context to support i18n for `name`/`description`, updated the normative text to allow the new pattern, and added a new example showing i18n for the issuer name and description as well as the VC name and descripiton.

Requesting a re-review from both of you now.

index.html
OR13: would it be better to just add a single internal reference? (off topic, and not blocking)

OR13 (approved):

index.html
msporny: I guess we could add an internal reference (in a future PR). I usually make a pass right before Proposed Rec to clean up the "lint-ignore" concerns. There is also the possibility of exporting the term (so it's added to the xref glossary, though I imagine adding `name` and `description` to the xref glossary would be frowned upon since they're so generic).

index.html
```diff
@@ -1496,6 +1496,149 @@ <h3>Types</h3>
...
+        </p>
+
+        <dl>
+          <dt><dfn class="lint-ignore">name</dfn></dt>
+          <dd>
+An OPTIONAL property that expresses the name of the <a>credential</a>. If
+present, the value of the <code>name</code> <a>property</a> MUST be a string or
+an object. If the value is an object, the object MUST contain a `value` property
+whose value is a string, and SHOULD contain a `lang` property whose value is a
+`langtag` string defined in [[BCP47]], and MAY contain a `dir` property whose
```
aphillips, line 1517: Is `langtag` meant to reference the ABNF in RFC5646 (the language tag part of BCP47)? If so, it should be `Language-Tag`, as the `langtag` production leaves out private use and grandfathered productions.

I would probably be more generic here and _also_ ensure that the level of conformance with BCP47 is called out:

```suggestion
string containing a well-formed language tag as defined by [[BCP47]], and MAY
```

index.html
```diff
@@ -1496,6 +1496,149 @@ <h3>Types</h3>
...
+    }
+  }
+}
+        </pre>
+
+        <p>
+Names and descriptions also support expressing content in different languages.
+To express a string with language and text direction information, one can use
+an object that contains the `value`, `lang`, and `dir` properties to express
+the text value, language code, and text direction, respectively.
```
aphillips, line 1570:
```suggestion
the text value, language, and text direction,
```

or perhaps:

```suggestion
the text value, language tag, and text direction,
```

aphillips (changes requested):
Thanks for this! Some minor comments/questions.

index.html
```diff
@@ -1496,6 +1496,149 @@ <h3>Types</h3>
 
       </section>
 
+      <section>
+        <h3>Names and Descriptions</h3>
+
+        <p>
+When displaying a <a>credential</a>, it can be useful to have
+text provided by the <a>issuer</a> that provides the
```
TallTed, line 1504:
```suggestion
text provided by the <a>issuer</a> that furnishes the
```

index.html
```diff
@@ -1496,6 +1496,149 @@ <h3>Types</h3>
...
+<a>credential</a> with a name as well as a short description of the
+purpose of the <a>credential</a>. The `name` and `description` <a>properties</a>
```
TallTed, line 1506:
```suggestion
<a>credential</a> with a name as well as a short description of its
purpose. The `name` and `description` <a>properties</a>
```

index.html
```diff
@@ -1496,6 +1496,149 @@ <h3>Types</h3>
...
+are provided for these purposes.
```
TallTed, line 1507:
```suggestion
are meant to serve these purposes.
```

index.html
TallTed: @aphillips suggestion is good, but the text following it was truncated, so I make this suggestion, fixing that truncation.
```suggestion
string containing a well-formed language tag as defined by [[BCP47]],
and MAY contain a `dir` property whose
```

index.html
```diff
@@ -1496,6 +1496,149 @@ <h3>Types</h3>
...
+          <dd>
+An OPTIONAL property that expresses the name of the <a>credential</a>. If
+present, the value of the <code>name</code> <a>property</a> MUST be a string or
+an object. If the value is an object, the object MUST contain a `value` property
+whose value is a string, and SHOULD contain a `lang` property whose value is a
+`langtag` string defined in [[BCP47]], and MAY contain a `dir` property whose
+value is a text direction string defined by the `dir` attribute in [[JSON-LD]].
+Ideally, the name of a <a>credential</a> is concise, human-readable, and could
+be used to enable an individual to quickly differentiate the <a>credential</a>
+from other <a>credentials</a> that they might hold.
```
TallTed, line 1521:
```suggestion
enable an individual to quickly differentiate one <a>credential</a>
from any other <a>credentials</a> that they might hold.
```

index.html
```diff
@@ -1496,6 +1496,149 @@ <h3>Types</h3>
...
+          </dd>
+          <dt><dfn class="lint-ignore">description</dfn></dt>
+          <dd>
+An OPTIONAL property that conveys specific details about a <a>credential</a>. If
+present, the value of the <code>description</code> <a>property</a> MUST be a
+string or an object. If the value is an object, the object MUST contain a
+`value` property whose value is a string, and SHOULD contain a `lang` property
+whose value is a `langtag` string defined in [[BCP47]], and MAY contain a `dir`
```
TallTed, line 1529:
```suggestion
whose value is a string containing a well-formed language tag as defined
by [[BCP47]], and MAY contain a `dir`
```

index.html
```diff
@@ -1496,6 +1496,149 @@ <h3>Types</h3>
...
+property whose value is a text direction string defined by the `dir` attribute
+in [[JSON-LD]]. Ideally, the description of a <a>credential</a> is no longer than
```
TallTed, line 1531:
```suggestion
in [[JSON-LD]]. Ideally, the description of a <a>credential</a> is no more than
```

index.html
```diff
@@ -1496,6 +1496,149 @@ <h3>Types</h3>
...
+a few sentences in length and conveys enough information about the
+<a>credential</a> to remind an individual as to its contents without having to
```
TallTed, line 1533:
```suggestion
<a>credential</a> to remind an individual of its contents without them having to
```

index.html
TallTed: @aphillips -- Note that you can make it easier for PR authors to accept any suggestion from you by making each such suggestion in its own comment. GitHub provides a 1-click merge button for single-suggestion comments, while any suggestion from a multi-suggestion comment requires the PR author to manually apply whichever (if any) suggestion with which they agree. (My own preference here is the latter, `language tag`.)

TallTed (changes requested):

index.html
dlongley: @aphillips,

The targeted approach that is currently shown in this PR is a better practice for security and simpler processing than the above (where a language is set context-wide). This is because the term definitions can be protected such that consuming the data without performing JSON-LD transformations would produce the same results as doing so. So both simple consumption and more complex consumption cases will have the same results.

Much of the VCDM is designed around expressing in simple JSON terms what JSON-LD already does, so that simpler processors can just code against the VCDM rules, consuming a subset of JSON-LD syntax without needing additional knowledge or transformations (and this also avoids having to reinvent the wheel that JSON-LD already provides).

aphillips: I don't understand. I thought `@language` in the `@context` was provided by JSON-LD and that, when one has a document in which all of the natural language text fields (such as `name`  and `description`) happen to be in the same language, it would be simpler to declare it in the `@context`. This is likely to be the high-runner case, I suspect, for most credentials. Does it impact the security and processing to show it in this example?

dlongley: @aphillips,

Not everyone wants to have to consume the value of `@language` in `@context` -- so there's better interop if it isn't done that way, but is instead done in a more targeted fashion. Additionally, a number of fields that hold string values that are not natural language text fields would have the `@language` value applied to them, which would create mistakes, for example, base-encoded status lists, any JOSE / SD-JWT fields that are expected to hold string identifiers or base-encoded values, and so on.

msporny: @aphillips, agree with @dlongley... but I'm going to wait for your confirmation before merging this PR. I think we're doing the most precise, correct thing wrt. i18n for both `name` and `description`... which is all this PR is focusing on. Doing blanket `@language` tags would be problematic for VCs, which we'll probably want to provide guidance on in the i18n section. I've raised issue #1264 to make sure we add this text.

Given that, @aphillips -- what are your thoughts on the above and this PR -- ok to merge, or needs more work?

aphillips: I thought that the `@context`/`@language` mechanism was a common pattern and a decent compromised between no language/direction metadata at all and having it on every natural language string. I think it would be appropriate to specify that the `@language` value does not apply to non-linguistic string data values (just as it doesn't really apply to, say, numbers). Using this mechanism is easy for producers who generate a credential in only a single language and doesn't clutter up the resulting file with lots of repetitive language values. So it's attractive that way. But it doesn't solve the problem of having multiple languages, of course, and it introduces processing that your group (appropriately) cares about. 

I do not have your group's domain expertise here, so would be glad to know if I shouldn't suggest this mechanism.

Other solutions include:

There is `rdf:langString` datatype, which has the downside of not including direction metadata.

JSON-LD 1.1 introduced the [`i18n` namespace](https://www.w3.org/TR/json-ld/#the-i18n-namespace), which allows single-line RDF serialization (which you may not care about) of a string with both language and direction metadata. This fixes some of the serialization issues. I think this is called an `rdf:compoundLiteral` and your PR is (basically) using it in the object representation of `name` and `description` defined here. The downside of this mechanism is that the consumer has to deserialize all of the compound literals to find the best matching one for the consumer's language preference (because the language tag is a field in the object). And both producers and consumers need to expect these (effectively custom) objects.

To solve the searching/language negotiation feature, there are [language maps](https://www.w3.org/TR/json-ld/#language-maps), but they only work with plain strings (to create `langStrings` via a backdoor). You don't show the use of a language map and I didn't suggest it because it doesn't solve the direction problem.

So what am I saying?

Basically, I think you could merge this. But I expect a lot of users will then ignore language and direction and just emit values that look like around line 1550 in the PR. The `compoundLiteral`-like solution exists, but we'll perpetuate a lack of lang/dir metadata:

```javascript
  /* no language, no direction == no rendering information */
  "issuer": {
    "id": "https://university.example/issuers/565049",
    "name": "Example University",
    "description": "A public university focusing on teaching examples."
  },
```

If the goal is to satisfy the _possibility_ of localization by providing the compoundLiteral-like solution, I think that this PR accomplishes that. I am not sure how the string representation of `name` or `description` get lang/dir metadata, though. What am I missing?

I'll look at #1264 and we can go from there.

TallTed (approved):

brentzundel (approved):

seabass-labrax (approved):

index.html
msporny: > I'll look at #1264 and we can go from there.

Ok, I'll start engaging on this issue over there.

> If the goal is to satisfy the possibility of localization by providing the compoundLiteral-like solution, I think that this PR accomplishes that.

Yes, that's all this PR is attempting to do -- it's the one thing I know that we can do safely... more in #1264.

> I am not sure how the string representation of name or description get lang/dir metadata, though. What am I missing?

They don't... at least, not automatically. There is more of an explanation in https://github.com/w3c/vc-data-model/issues/1264#issuecomment-1701830685

msporny:
Normative, multiple reviews, changes requested and made, remaining issues tracked, no objections, merging.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-08-29-vcwg-special#section1-2) on 2023-08-29

- no resolutions were taken

#### 1.2.  (pr vc-data-model#1252)

_See github pull request [vc-data-model#1252](https://github.com/w3c/vc-data-model/pull/1252)._




**Manu Sporny:** changes requested by aphillips, waiting on review of changes.  
… can add description to the issuer or to the credential itself. Issue raised were internationalization based, now allows for language tags and has examples. Still raised issue of setting a default language for entire credential. That raises issues on side effects on languages being applied to other strings such as encoded images.  

</details>

Issue #1150, opened 2023-06-12
Protected term errors when supporting v1 and v2


OR13:
Consider:

```json
{
  "@context": ["https://www.w3.org/ns/credentials/v2"],
  "type": ["VerifiablePresentation"],
  "verifiableCredential": [
    {
      "@context": ["https://www.w3.org/2018/credentials/v1"],
      "type": ["VerifiableCredential"]
    }
  ]
}
```
> jsonld.SyntaxError: Invalid JSON-LD syntax; tried to redefine a protected term.

mprorock:
excellent catch @OR13

OR13:
I was able to work around this issue, by removing the v2 context from the outer object:

```ts
const multiPresentationSupport = {
        '@context': {'@vocab': 'https://www.w3.org/2018/credentials#'},
        type: ['VerifiablePresentation', 'VerifiablePresentations'],
        id: `urn:uuid:${uuid.v4()}`,
        holder: process.env.API_BASE_URL as string,
        verifiablePresentation: verifiablePresentations,
      }
```

As you can see, most of the terms that get defined related to a presentation of a presentation of a mill test report are controlled by origins other than W3C...

<img width="2702" alt="Screen Shot 2023-06-12 at 12 17 45 PM" src="https://github.com/w3c/vc-data-model/assets/8295856/771b216c-b010-4eaa-b3e8-329c50eeb679">

OR13:
Probably a better resolution to the above error

```ts
const multiPresentationSupport = {
        '@context': {'@vocab': 'https://www.w3.org/2018/credentials#'},
        type: ['VerifiablePresentation'],
        id: `urn:uuid:${uuid.v4()}`,
        holder: holder,
        verifiableCredential: verifiablePresentations,
      }
```

<img width="2716" alt="Screen Shot 2023-06-12 at 12 28 03 PM" src="https://github.com/w3c/vc-data-model/assets/8295856/cbb1fbfc-a900-4b1d-ab70-37b527d94df2">

Notice this time the graphs are joined properly, because `https://www.w3.org/2018/credentials#verifiableCredential` is a known property of a `VerifiablePresentation` where as `verifiablePresentation` IS NOT.

TallTed:
@OR13 — `don't run the above code`

[...]

Issue #1240, opened 2023-08-15
Address normative concept of VerifiableCredentialGraph


OR13:
See the original comment here: https://github.com/w3c/vc-data-model/pull/1236#issuecomment-1678561871

The word "envelop" is being used here, in a way that aligns with "VerifiablePresentation".

I find the concept of "VerifiableCredentialGraph" confusing, and note that it does not occur even once in the current TR.

https://github.com/search?q=repo%3Aw3c%2Fvc-data-model%20VerifiableCredentialGraph&type=code

I prefer to remove the term, since it can only lead to more confusion.

If the term is providing value, I prefer to unify that value with "VerifiablePresentation", but would need to know more about how it actually works, to even propose a path to that.

OR13:
In the original comment, @iherman  wrote:

> How is it different from: the former is the "envelope" that contains the latter. Just as the envelope is not the same as what it contains, the VerifiableCredentialGraph is different from what it contains, i.e., a VerifiableCredential. This differentiation is crucial for, e.g., the usage of the RDFC 1.0 algorithm.

Can someone elaborate further on this?.. seems like changes here would potentially be breaking, I'd like to understand why that is a good thing / bad thing.

iherman:
> I find the concept of "VerifiableCredentialGraph" confusing, and note that it does not occur even once in the current TR.
> 
> https://github.com/search?q=repo%3Aw3c%2Fvc-data-model%20VerifiableCredentialGraph&type=code
> 

Indeed, it is not explicitly in the spec. But the concept is already there in the context file:

```json
"verifiableCredential": {
          "@id": "https://www.w3.org/2018/credentials#verifiableCredential",
          "@type": "@id",
          "@container": "@graph"
        },
```

Ie, the fact that the value of a `verifiableCredential` is a _graph_ and not a `VerifiableCredential`. That aspect must be expressed in the vocabulary. Speaking bluntly, the vocabulary specification in the [1.1 version](https://github.com/w3c/vc-data-model/blob/b63f65b7353f296416a2bfc96411cde81473c4c3/vocab/credentials/credentials.ttl#L102) for that term was a bug.

Maybe it is worth referring to the concept of RDF graphs (a.k.a. datasets) in the [RDF 1.1 spec](https://www.w3.org/TR/rdf11-concepts/#section-dataset).

> I prefer to remove the term, since it can only lead to more confusion.

I would -1 to remove the "concept" from the vocabulary. I am fine trying to find a different name.

[...]

PR #1259, opened 2023-08-25
Sync and fix bugs in v2 context.

msporny:
This PR is an attempt to address issue #1150 by ensuring that redefinition doesn't fail in `verifiableCredential` when holding a VC v1.x compliant VC. It also fixes two bugs in the v2 context: 1) ensures that redefinition doesn't fail for `cnf`, and 2) synchronizes the DataIntegrity scoped context terms to now include `cryptosuiteString`.

### DIFF
--- a/contexts/credentials/v2
+++ b/contexts/credentials/v2
@@ -44,20 +44,17 @@
     },
     "cnf": {
       "@id": "https://www.iana.org/assignments/jwt#cnf",
-      "@context": [
-        null,
-        {
-          "@protected": true,
-          "kid": {
-            "@id": "https://www.iana.org/assignments/jwt#kid",
-            "@type": "@id"
-          },
-          "jwk": {
-            "@id": "https://www.iana.org/assignments/jwt#jwk",
-            "@type": "@json"
-          }
+      "@context": {
+        "@protected": true,
+        "kid": {
+          "@id": "https://www.iana.org/assignments/jwt#kid",
+          "@type": "@id"
+        },
+        "jwk": {
+          "@id": "https://www.iana.org/assignments/jwt#jwk",
+          "@type": "@json"
         }
-      ]
+      }
     },
     "_sd_alg": {
       "@id": "https://www.iana.org/assignments/jwt#_sd_alg"
--- a/contexts/credentials/v2
+++ b/contexts/credentials/v2
@@ -5,6 +5,7 @@
 
     "id": "@id",
     "type": "@type",
+
     "kid": {
       "@id": "https://www.iana.org/assignments/jose#kid",
       "@type": "@id"
@@ -43,17 +44,20 @@
     },
     "cnf": {
       "@id": "https://www.iana.org/assignments/jwt#cnf",
-      "@context": {
-        "@protected": true,
-        "kid": {
-          "@id": "https://www.iana.org/assignments/jwt#kid",
-          "@type": "@id"
-        },
-        "jwk": {
-          "@id": "https://www.iana.org/assignments/jwt#jwk",
-          "@type": "@json"
+      "@context": [
+        null,
+        {
+          "@protected": true,
+          "kid": {
+            "@id": "https://www.iana.org/assignments/jwt#kid",
+            "@type": "@id"
+          },
+          "jwk": {
+            "@id": "https://www.iana.org/assignments/jwt#jwk",
+            "@type": "@json"
+          }
         }
-      }
+      ]
     },
     "_sd_alg": {
       "@id": "https://www.iana.org/assignments/jwt#_sd_alg"
@@ -146,7 +150,8 @@
         "verifiableCredential": {
           "@id": "https://www.w3.org/2018/credentials#verifiableCredential",
           "@type": "@id",
-          "@container": "@graph"
+          "@container": "@graph",
+          "@context": null
         },
         "termsOfUse": {
           "@id": "https://www.w3.org/2018/credentials#termsOfUse",
@@ -273,7 +278,10 @@
             }
           }
         },
-        "cryptosuite": "https://w3id.org/security#cryptosuite",
+        "cryptosuite": {
+          "@id": "https://w3id.org/security#cryptosuite",
+          "@type": "https://w3id.org/security#cryptosuiteString"
+        },
         "proofValue": {
           "@id": "https://w3id.org/security#proofValue",
           "@type": "https://w3id.org/security#multibase"


dlongley (approved):

contexts/credentials/v2
```diff
@@ -146,7 +150,8 @@
         "verifiableCredential": {
           "@id": "https://www.w3.org/2018/credentials#verifiableCredential",
           "@type": "@id",
-          "@container": "@graph"
+          "@container": "@graph",
+          "@context": null
```
iherman, line 154: Just trying to test my own understanding, @dlongley @msporny. My JSON-LD knowledge is rusty.

Setting `@context` to `null` means that, at this point, that the active context is reduced to nothing, except some basic mapping specified by the JSON-LD spec (like `@type` mapping to a URL). Per our spec, the value of the `verifiableCredential` property in JSON-LD is supposed to be a `VerifiableCredential` (which is, in the RDF side, enclosed in a separate graph). The effect of this is new setting that _unless the `@type` is explicitly set to `VerifiableCredential` in the value object_ things will, and should, go wrong. Put it another way, it reinforces the requirement that the value object must explicitly declare itself to be a Verifiable Credential (and relates to the fact that the vocabulary explicitly says that a VerifiableCredentialGraph should contain a single Verifiable Credential).

Is this indeed the goal of this extra setting? If so, shouldn't there be a similar setting in the security `@context` for the `proof` property?

contexts/credentials/v2
```diff
@@ -43,17 +44,20 @@
...
-        "@protected": true,
-        "kid": {
-          "@id": "https://www.iana.org/assignments/jwt#kid",
-          "@type": "@id"
-        },
-        "jwk": {
-          "@id": "https://www.iana.org/assignments/jwt#jwk",
-          "@type": "@json"
+      "@context": [
+        null,
```
OR13, line 48: can you explain what this is "fixing" ?

contexts/credentials/v2
```diff
@@ -43,17 +44,20 @@
...
+        {
+          "@protected": true,
```
OR13, line 50: I assume this is just preventing someone from changing term definitions... what would be a use case for doing that?

Why is protected required for this object?

The other part that this section does seems to fix the way the graph looks by default, which seems great.

contexts/credentials/v2
```diff
@@ -273,7 +278,10 @@
             }
           }
         },
-        "cryptosuite": "https://w3id.org/security#cryptosuite",
+        "cryptosuite": {
+          "@id": "https://w3id.org/security#cryptosuite",
+          "@type": "https://w3id.org/security#cryptosuiteString"
```
OR13, line 283: This name makes me sad, I get what its doing, but it feels like the RDF version of:

```ts
type MyString = string;
const myString: MyString = "string";
```

instead of:

```ts
type CryptoString = "CRYDI2" | "CRYDI3" | "CRYDI5";
const cryptosuite: CryptoString = "CRYDI5";
```

Is there really no better option for this `@type` ?

contexts/credentials/v2
OR13: It seems like it would make some of these examples illegal:

https://w3c.github.io/vc-jose-cose/#presentations

```json
{
  "@context": ["https://www.w3.org/ns/credentials/v2"],
  "holder": "urn:ietf:params:oauth:jwk-thumbprint:sha-256:_Fpfe27AuGmEljZE9s2lw2UH-qrZLRFNrWbJrWIe4SI",
  "verifiableCredential": [{
      "@context": [
        "https://www.w3.org/ns/credentials/v2"
      ],
      "type": [
        "VerifiableCredential"
      ],
      "issuer": "https://issuer.example/issuers/68",
      "validFrom": "2023-06-07T21:14:14.148Z",
      "credentialSubject": {
        "id": "https://subject.vendor.example"
      }
    },
    "https://vendor.example/credentials/42", "did:example:123",
    "urn:uuid:01ec9426-c175-4e39-a006-d30050e28214",
    "urn:ietf:params:oauth:jwk-thumbprint:sha-256:_Fpfe27AuGmEljZE9s2lw2UH-qrZLRFNrWbJrWIe4SI",
    "data:application/vc+ld+json;base64,eyJAY29udGV4dCI6WyJodHRwczovL3d3dy53My5vcmcvbnMvY3JlZGVudGlhbHMvdjIiXSwidHlwZSI6WyJWZXJpZmlhYmxlQ3JlZGVudGlhbCJdLCJpc3N1ZXIiOiJkaWQ6andrOmV5SnJhV1FpT2lKMWNtNDZhV1YwWmpwd1lYSmhiWE02YjJGMWRHZzZhbmRyTFhSb2RXMWljSEpwYm5RNmMyaGhMVEkxTmpwdlFtUm1kbVpET1hoNk1GOUJVWFpSTjNZMU1YbERXbDl6ZUdwNU56VkNUSEpJZWsxT1Jqa3lPV1U0SWl3aWEzUjVJam9pVDB0UUlpd2lZM0oySWpvaVJXUXlOVFV4T1NJc0ltRnNaeUk2SWtWa1JGTkJJaXdpZUNJNklqTmljbU5zYjBJNGFEUk5XbFZJYms5UVVHbGtTbXd0U2pkdVVsRkpXSFJUYUZwM1oyNW1jbHAxVDI4aWZRIiwidmFsaWRGcm9tIjoiMjAyMy0wNi0wN1QyMToxNDoxNC4xNDhaIiwiY3JlZGVudGlhbFN1YmplY3QiOnsiaWQiOiJodHRwczovL3N1YmplY3QudmVuZG9yLmV4YW1wbGUifX0="
  ]
}
```

Or are these ok, as long as they "dereference" to a compact JSON-LD object representing a conforming document for a `verifiable credential` ?

OR13 (changes requested):
I have question about the "null" context behavior, how well supported is that, will this break existing JSON-LD processors, etc... assuming we don't think this artificially limits the software processor space, these seem like great additions, and we should take them.

contexts/credentials/v2
msporny: > can you explain what this is "fixing" ? (in more detail than the PR description has).

You defined `kid` as `https://www.iana.org/assignments/jose#kid"` above... and then re-define it to `https://www.iana.org/assignments/jwt#kid` below -- don't know why you are switching between `.../jose#`  and `.../jwt#`, but presuming you meant to do that, this is ensuring that your re-definition when used in `cnf` works as intended.

If you didn't mean to do that, we're going to have to rework all the JWT/JOSE terms to align them with either `.../jose#`  or `.../jwt#`.

contexts/credentials/v2
msporny: > I assume this is just preventing someone from changing term definitions... what would be a use case for doing that?

Yes. The use case is making sure that people don't re-define the JWT terms to use some other URL value (that doesn't point to the IANA registries).

> Why is protected required for this object?

For the reason above.

> The other part that this section does seems to fix the way the graph looks by default, which seems great.

Yep.

contexts/credentials/v2
msporny: > Or are these ok, as long as they "dereference" to a compact JSON-LD object representing a conforming document for a verifiable credential?

In theory, yes, but in reality, you couldn't guarantee the security of that sort of reference unless you were to also integrate some sort of content hash for the target object. 

> It seems like it would make some of these examples illegal:

Those examples were never legal in v1.0, v1.1, and v2.0... more below.

The addition is the null'ing of the context, which is required to address issue #1150. So, this change is purely additive and doesn't impact the examples you provided (the answers below would be the same w/o this PR).

That said:

> "https://vendor.example/credentials/42"

If this is an identifier for a graph, you could potentially express it, but then you couldn't guarantee the security around the object unless you were to provide a content hash.

> "did:example:123",

A DID Document isn't a VC, so, no, that wouldn't work, AFAICT.

> "urn:uuid:01ec9426-c175-4e39-a006-d30050e28214",

Presuming this thing is an identifier for a graph, you could potentially express it, but then it's not clear how you'd dereference it or guarantee the security around the object unless you were to provide a content hash. This couldn't work unless one were to presume that there was some sort of "graph dereferencing" function in play. 

>  "urn:ietf:params:oauth:jwk-thumbprint:sha-256:_Fpfe27AuGmEljZE9s2lw2UH-qrZLRFNrWbJrWIe4SI",

Same answer as the above.

>  "data:application/vc+ld+json;base64,eyJAY29udG...ifX0=

This is not a valid VC per the VCDM. If you pre-processed it and put this object into the array, it might be a valid VC.

contexts/credentials/v2
msporny: > Is there really no better option for this @type ?

At present, no, there isn't... because the range of cryptosuiteString over time can't be known (just like all future entries in the JOSE registry can't be known today).

contexts/credentials/v2
```diff
@@ -43,17 +44,20 @@
     },
     "cnf": {
       "@id": "https://www.iana.org/assignments/jwt#cnf",
-      "@context": {
-        "@protected": true,
-        "kid": {
-          "@id": "https://www.iana.org/assignments/jwt#kid",
```
msporny, line 49: @OR13 Just to be clear, all values you've declared as `"@type": "@id"` MUST be URIs... you cannot have non-URL values in there... are you sure you want to do this?

msporny:
> I have question about the "null" context behavior, how well supported is that, will this break existing JSON-LD processors, etc... assuming we don't think this artificially limits the software processor space, these seem like great additions, and we should take them.

"null" context behaviour is supported by all modern JSON-LD Processors, AFAIK. There are explicit tests for the behavior: https://w3c.github.io/json-ld-api/reports/ , namely Test tpr14: Clear protection with null context. (new in JSON-LD 1.1), and 
Test tpr15: Clear protection with array with null context (new in JSON-LD 1.1).

It won't break existing JSON-LD processors (see above).

contexts/credentials/v2
OR13: So what this is fixing is:

Using the more specific registry to define the terms underneath `cnf`, ensuring that `kid` and `jwk` are defined according to the JWT registry, and not "generic jose".

This seems like an excellent improvement.

OR13 (approved):

contexts/credentials/v2
OR13: JSON-LD also says that some text fields are datetime strings... my experience is that eventually you process a string that is supposed to be a URL or a datetime, and it turns out to be an emoji.... Also, google uses `@id` values that are not URLs in their enterprise knowledge graph search API, so i'm not sure I would assume that those values are URLs.

contexts/credentials/v2
dlongley: @OR13,

If anything else needs to be defined under `cnf`, you'll want to do so as well. Right now only `kid` and `jwk` will be defined.

contexts/credentials/v2
dlongley: @OR13,

What's being said here is that processors will raise errors if you use `"@type": "@id"` in the term's context definition if something other than a URL or graph node is used in the data. This is not the same case as using a datetime string that is malformed, etc., as those will not be processed by default. So, if you want to allow any kind of string in the field, you should remove `"@type": "@id"`, otherwise default processing will try to enforce the use of URLs or graph nodes by raising errors when something else is present.

brentzundel (approved):

contexts/credentials/v2
selfissued: A JOSE `kid` is a string and need not be a URI.  The typing should reflect that.

selfissued (changes requested):
A JOSE `kid` is a string and need not be a URI.  The typing should reflect that.

contexts/credentials/v2
seabass-labrax: I agree with you, @OR13, that this is sub-optimal. However, I think it is unlikely to amount to much work for implementers, as they can hard-code the string values that their current or future implementation of the VC data model supports. The only other issue is the inability to shorten the string values into a vocabulary that can, say, be compressed with CBOR.

seabass-labrax (approved):

contexts/credentials/v2
OR13: @dlongley that seems like breaking changes, `cnf` should get `issuer dependent` terms when a none "registered claim name" is used imo.

OR13 (changes requested):
Based on other reviews:

- remove changes that cause `cnf` to have undefined terms... they should get the default issuer defined terms, they should not error... if new json members are added under it.

contexts/credentials/v2
OR13: I don't think we should remove the `@type: @id` from `kid` or `iss` which can both be strings or URIs because.... this causes the RDF graph to be reliant on `issuer` and `verificationMethod` which are only defined in the context of JSON-LD.... and are always URIs.... 

Removing the ability to use `iss` and `kid` and have them treated similar to `issuer` and `verificationMethod` would undermine our ability to use the core data model as JSON-LD, this is related to the "benefits of treating the data model as JSON-LD vs JSON".

OR13: If its possible for the `@type: @id` and String to both be legal, that would be preferred... but in absence of that, we prefer the processing of these fields with DIDs or other URLs or URNs to build useful graphs.

msporny: > @selfissued not sure it matters since URIs are strings

URIs have a structure, they're not just arbitrary strings (which, IIUC, @selfissued is saying `kid` can be any arbitrary string).

If you make `kid` and `iss` an `type: @id`, a JSON-LD processor will throw an error when processing `kid` and `iss` if it isn't a URI... I don't think we want that behavior, it'll lead to interop issues. 

That said, if you want to enforce that `kid` and `iss` be a URI when used in VCs, then using `type: @id` is one way to do that... but again, we'd be narrowing the scope of it in that case.

msporny:
> * remove changes that cause `cnf` to have undefined terms... they should get the default issuer defined terms, they should not error... if new json members are added under it.

hmm, right, yes, will fix.

contexts/credentials/v2
dlongley: @OR13,

If you want to keep all of the other terms we can just remove `null` and just let these other specific overrides happen. If you want to clear all the other terms we should keep `null` and add back in whatever else you specifically want.

contexts/credentials/v2
OR13: > If you pre-processed this URL and put this object into the array, it might be a conforming document. By itself, it's not a conforming document.

I'd like to see spec text addressing this, its not obvious that what you are saying is true, and these n-quads look fine to me:

```
_:c14n0 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://www.w3.org/2018/credentials#VerifiablePresentation> .
_:c14n0 <https://www.w3.org/2018/credentials#holder> <urn:ietf:params:oauth:jwk-thumbprint:sha-256:_Fpfe27AuGmEljZE9s2lw2UH-qrZLRFNrWbJrWIe4SI> .
_:c14n0 <https://www.w3.org/2018/credentials#verifiableCredential> _:c14n1 .
_:c14n0 <https://www.w3.org/2018/credentials#verifiableCredential> _:c14n3 .
_:c14n0 <https://www.w3.org/2018/credentials#verifiableCredential> _:c14n4 .
_:c14n0 <https://www.w3.org/2018/credentials#verifiableCredential> _:c14n5 .
_:c14n0 <https://www.w3.org/2018/credentials#verifiableCredential> _:c14n6 .
_:c14n0 <https://www.w3.org/2018/credentials#verifiableCredential> _:c14n7 .
_:c14n2 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://www.w3.org/2018/credentials#VerifiableCredential> _:c14n1 .
_:c14n2 <https://www.w3.org/2018/credentials#credentialSubject> <https://subject.vendor.example> _:c14n1 .
_:c14n2 <https://www.w3.org/2018/credentials#issuer> <https://issuer.example/issuers/68> _:c14n1 .
_:c14n2 <https://www.w3.org/2018/credentials#validFrom> "2023-06-07T21:14:14.148Z"^^<http://www.w3.org/2001/XMLSchema#dateTime> _:c14n1 .
```

> A DID Document isn't a VC, so, no, that wouldn't work, AFAICT. It's not a conforming document.

This can't be true, the only normative requirement for a DID Document is that it have an `id` that is a DID... that makes any VC that has an `id` that is a DID a DID Document... especially if its returned with content type `application/did+json` (where context can be present and is ignored).... there are a lot of use cases where a controller document might need to be a VC. Consider also https://openid.net/specs/openid-connect-userinfo-vc-1_0-00.html

This is one step off of a signed did resolution result... we should not be designing incompatibility here, we don't want to preclude use cases we see developing in the wild.

> Presuming this thing is an identifier for a graph, you could potentially express it, but then it's not clear how you'd dereference it or guarantee the security around the object unless you were to provide a content hash. This couldn't work unless one were to presume that there was some sort of "graph dereferencing" function in play. Again, not a conforming document.

Similar to DID Documents, you can't presume that DIDs can't be dereferenced, even though it's not defined normatively.

I didn't share the resolved document, but if I shared one that had an `id` that matched this, and was a status list credential, that would work... the verifier may have previously dereferenced the identifier, the holder does not need to send them a massive status list when thats the case.

> Same answer as the above.

Same answer as above.

> If you pre-processed this URL and put the resulting object into the array, it might be a conforming document. By itself, it's not a conforming document.

Its clear we don't have agreement on the domain / range of `verifiableCredential`... I see it as having the same structure as `issuer` and `holder`, or `proof`... it can be an object, or an array of strings and objects.... similar to verification relationships.

The v1 context seems to support this assertion.

```json
"issuer": {
"@id": "cred:issuer",
"@type": "@id"
},
```

```json
"proof": {
"@id": "sec:proof",
"@type": "@id",
"@container": "@graph"
},
```

```json
"verifiableCredential": {
"@id": "cred:verifiableCredential",
"@type": "@id",
"@container": "@graph"
}
```

This is the definition in the v2 context today:

```json
"verifiableCredential": {
"@id": "https://www.w3.org/2018/credentials#verifiableCredential",
"@type": "@id",
"@container": "@graph"
},
```

For comparison, here are the context definitions for verification relationships:

```json
"assertionMethod": {
"@id": "https://w3id.org/security#assertionMethod",
"@type": "@id",
"@container": "@set"
},
```

As you can see the only difference is `@graph` vs `@set`... this does not imply that the property must ALWAYS be an array of objects of type "VerifiableCredential".... afaik.

It does mean that the n-quads produce blank nodes, as you can see from the example above.

There are lots of cases in RDF / JSON-LD where you may have previously dereferenced an identifier to a graph node / object... and you don't need to send a "full copy" of the same object over and over again... this is especially common in supply chain use cases cc @mprorock @mkhraisha ... we use this pattern when communicating references to previous "credentials" and "presentations" and "workflows" which are just another large JSON object that is referred to be "id".

TLDR, I am opposed to breaking changes to this experience in the v2 context.

I am fine leaving the behavior as it is above (where you can always replace an `id` with an `object` that contains that `id`), for places where you see `"@type": "@id",`... This would preserve the experience we created with DIDs, where the same "verificationMethod" object is referred to by several "verification relationships", for example, a P-256 Public Key might be used for both authentication and encryption, or authentication and credential issuance.

contexts/credentials/v2
OR13: im going to resolve this, since we don't seem to be requesting changes, but feel free to open again if you disagree.

contexts/credentials/v2
OR13: > That said, if you want to enforce that `kid` and `iss` be a URI when used in VCs, then using `type: @id` is one way to do that

As shown in the example above, the industry does not agree that `@id` must be a URI... we don't need to do anything here, here is digital bazaar's json-ld processor handling the above example without erroring:

https://v.jsld.org/FwjjEErfzWY5ogNK7HM6ZFCRsfCBtfGxfpp64361qdd2bu7zFeykxD3e4gTXmcpNrztDGpfaa4zn5BNd8Uy5znajBTBBbFpwZK8QEGKSAgaU3aF7rq3DsLmuvJdGi2FFqgmQkvPiwzULpji68V6WNcHCvKmSB136ngyXwizmqP6BY8hB3zWCKAeNpt9Y5o45rdMNyUxNsdUa9JMfoecVqB94VKAHJ2z92U5KfbaDEHdAkH56JnwjVQf1EA7r2HEDVTefgiTiz81Q1KzbSrv636XupufPGKS6CwTBTZLT5paZjcQkjiiYBMRZQWxYxc8Ji4ZzUHzLngspxERKcUheDaKvemoRGrTxktsR7bwDFm6dTkStg8baTymnbPCYiX1GbHTv7mK6aJUUeykESpwedtzGRMZTmZuhdRJ2eCyZvZTyWRe8NmNMZXeek5drfFjjpCavDuHDQiZP76iuo2xEMkuL99hgrNJbGaKC6ULpYTVtXbEC6ep9oCXqebF97jibSuEkHkZMWjEKL1hF5PXKr5dReptbxExcLeUqNVbMtASXYqrQvhR9UikDTabmX19HPnua6S3KPanmMm9Mg58VGSAqiZKw1u9S8WJ1anzBy4UpS7ANWkAnTDAAZoyxKWEQrapoSgppqzEYHDqHqg1kzXJW59cQEXn1yPYW8pwNX2XUL1Kf6EjwHLjXLR

contexts/credentials/v2
dlongley: > TLDR, I am opposed to breaking changes to this experience in the v2 context.

There aren't breaking changes here -- `verifiableCredential` has always used a `@graph` container to ensure that the data expressed in the value there is put into its own graph. The VP has a relationship with other graph containers via `verifiableCredential`, where each value is its own graph containing either a `VerifiableCredential` or other data conformant with the data model that is derived from a `VerifiableCredential`. This latter bit was for the CL Signature ZKP folks that wanted to not necessarily put a VC directly in the value, but to put a "derived credential with ZKP proof data" in it. This has been the case since 1.1.

If you just put a string value in that field, then all that happens is an empty graph container (with a blank node identifying it) is created and there's a relationship established between the VP and this empty graph. A string value (ID / URL) for a subject does not establish any claims or additional relationships so the new graph is just empty. The only relationship added is between the VP and the blank node of this empty graph via the `verifiableCredential` property (loosely, the claim is: VP -> verifiableCredential -> blank node of an empty graph). This empty graph is not a valid `VerifiableCredential` nor is it data derived from one, so it's not a valid use of the VCDM.

But! There is a property that can do what you want to do in VCDM 2.0. I would recommend using `relatedResource` to add whatever additional resources you want -- and there's the added bonus of being able to include a content integrity hash to go along with it, which is something you'd want for that approach anyway. Hopefully this resolves the issue.

contexts/credentials/v2
dlongley: I'm happy to just drop the `null` and let the two overrides take place, leaving the other things there. Seems like the simplest solution to me at this time.

dlongley, line 48: Not easy to suggest this in github right now, but want it tracked, so don't apply directly:

```suggestion
```

contexts/credentials/v2
dlongley: @OR13,

That value only works when there's a base URL (like there always is with Web pages during search crawling and all that). With VCs, there is no base URL, it's presumed to be `null`, so it would result in a relative URL being used and an error as shown here: https://tinyurl.com/pys6xa4k

So I recommend we remove `"@type": "@id"` from all of the JOSE term context definitions that can take something other than a URL.

contexts/credentials/v2
dlongley: @OR13,

I've also mentioned that we need to make sure to add `relatedResource` to both VCs and VPs here: https://github.com/w3c/vc-data-model/issues/1263#issuecomment-1701349644

OR13: Staying on topic, I am opposed to adding the "null" context here: https://github.com/w3c/vc-data-model/pull/1259/files#diff-267b7e47789daf9a1d312b42826aa937a1bb5af4312ee90df6b8a64cd6426f4fR154

Until we achieve clarity / consensus on the topic above.

dlongley: > > This latter bit was for the CL Signature ZKP folks that wanted to not necessarily put a VC directly in the value, but to put a "derived credential with ZKP proof data" in it. This has been the case since 1.1.
>
> This AFAIK, never actually resulted in implementations.

Yes, if you wanted to strike that part, I'd be ok with it and would favor just requiring embedded VCs.

> > A string value (ID / URL) for a subject does not establish any claims or additional relationships so the new graph is just empty.
> 
> When you say `subject` here, you mean the identifier for a verifiable credential, (assuming it has graph node id), right?

Well, what the `subject` actually is doesn't matter, its identifier isn't associated with any statements because nothing is said about it (making the graph of statements "empty"). It's true that the `subject` could be a verifiable credential, but if nothing is claimed about that credential directly within the embedded data (as opposed to who knows what might be there by reference). In short, the `verifiableCredential` property has not been for referencing things, but instead for directly embedding data-model-compliant things in a presentation whilst preserving a wrapper around them (a "named graph") to isolate them from one another.

> > The only relationship added is between the VP and the blank node of this empty graph via the `verifiableCredential` property (loosely, the claim is: VP -> verifiableCredential -> blank node of an empty graph). This empty graph is not a valid `VerifiableCredential` nor is it data derived from one, so it's not a valid use of the VCDM.
> 
> This seems to be confusing "VerifiableCredential" with "VerifiableCredentialGraph" cc @iherman & @msporny , we discussed confusion related to this on #1240

The `verifiableCredential` (property name) relationship is between a VP and a graph (VerifiableCredentialGraph`). This is what I said above when I said: "(loosely, the claim is: VP -> verifiableCredential -> blank node of an empty graph)". Of course, this embedded graph must not be empty to be valid, but I was pointing out the invalid case in that sentence.

> It would help to understand the normative / english language value of the "VerifiableCredentialGraph" since that is what is "empty".

If I understand you, the value is to put a wrapper around a set of statements to isolate them from other statements made: to help keep track of associations. The statements made in a VC are "wrapped" by a separate graph (put "into" that graph) to keep them separate from statements made in any other VC embedded in the VP. A VP can carry N-many VCs this way, potentially coming from many different issuers, whilst tracking the association of the statements in the graph model.

> > But! There is a property that can do what you want to do in VCDM 2.0. I would recommend using `relatedResource` to add whatever additional resources you want -- and there's the added bonus of being able to include a content integrity hash to go along with it, which is something you'd want for that approach anyway. Hopefully this resolves the issue.
> 
> I like the suggestion, lets not let it stop this PR, I will file a separate issue for follow up: #1265

+1

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-08-29-vcwg-special#section1-6) on 2023-08-29

- no resolutions were taken

#### 1.6. Sync and fix bugs in v2 context. (pr vc-data-model#1259)

_See github pull request [vc-data-model#1259](https://github.com/w3c/vc-data-model/pull/1259)._




**Manu Sporny:** at the point where we need to lock down the JSON-LD context, Dave Longley and Manu have taken passes to fix issues in the context.  
… one new issue - concern about JWT properties are defined. They are currently defined as having to be URLs, may not be the right thing for all properties. For example, kid is defined as being a URL value.  

**Michael Jones:** kid has no semantics, it is a string.  

**Manu Sporny:** asks for note from selfisused on 1259 to kid, needs general review of types.  

</details>

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-09-05-vcwg-special#section3-3) on 2023-09-05

- no resolutions were taken

#### 3.3. Sync and fix bugs in v2 context. (pr vc-data-model#1259)

_See github pull request [vc-data-model#1259](https://github.com/w3c/vc-data-model/pull/1259)._




**Manu Sporny:** we might be OK to merge this, as it addresses Orie's issue.  

> *Orie Steele:* Move that discussion to another issue.

**Orie Steele:** lots of graph nodes are identified with strings that are not URLs, so Google's implementation works with both.  
… we should not make changes to `@context` that throws errors in processors.  

**Michael Jones:** in XML we have namespaces, but in JSON we don't.  

> *Orie Steele:* See this link for examples of `@id` that are strings. [https://github.com/w3c/vc-data-model/pull/1259#discussion_r1310768121](https://github.com/w3c/vc-data-model/pull/1259#discussion_r1310768121).

**Manu Sporny:** path forward seems to be to merge what there in the PR and raise issues on anything that we dont like.  

> *Orie Steele:* Not true, and not relevant to the issue.

**Manu Sporny:** web pages have a base URL for strings to be relative to, whereas VCs do not have.  

> *Manu Sporny:* yes, agreed w/ Orie's "This is what we're waiting for".

> *Dave Longley:* +1 to what Orie said, just need to remove that one `null`.

**Orie Steele:** we are waiting for a patch from `@dlongley` before we can merge.  

**Sebastian Crane:** no objections to this PR. Namespaces are valuable, so URLs are very useful.  

**Michael Jones:** namespaces are totally unnecessary.  

> *Orie Steele:* A registry is just a namespace `@selfissued`... so your comment makes very little sense to me, as a frequent user of registries.

> *Benjamin Young:* +1.

> *Dave Longley:* +1 to Orie, registries are namespaces too.

</details>

contexts/credentials/v2
msporny: Fixed in c44da90e4d0be776e8f7dc465bfee1b87fd4d0fc

contexts/credentials/v2
msporny: This concern is now tracked in https://github.com/w3c/vc-data-model/issues/1275.

msporny:
Normative, multiple reviews, changes requested and made, remaining issues tracked, no objections, merging.

Issue #1227, opened 2023-08-03
Document the value of processing as JSON-LD


OR13:
Previous discussed on:

- https://github.com/w3c/vc-data-model/pull/1202
- https://github.com/w3c/vc-data-model/issues/1201

Working group members asked for clarity on why processing the data model as JSON-LD was valuable, especially given that the context and vocabulary are now normative.

@msporny said he could take a stab at text on a recent working group call.

OR13:
On the last working group call, I said:

There is currently no text describing the value that is unlocked by leveraging the normative context and vocabulary, because we don't have any cases where the context is applied.

This causes people to see the "JSON-LD data model" as "not valuable", and resulted in a lot of confusion on the "processing the core data model as JSON PR".

To remedy this, the working group should agree that there is value in processing the core data model as "JSON-LD", we should provide examples for how that value might be obtained... this will make the current text in the "processing the data model as JSON", less confusing.

mtaimela:
Great proposal to have the JSON-LD side thoroughly described. My own personal opinion is that JSON-LD will help greatly in the integration and after it is well adopted, this can even be used in VP Exchange to request proofs for a meaning.

On the other side, I see JSON as a gateway to implement JSON-LD vocabs. The vocab is not an easy task for governmental domains, which anyway yield many benefits for users (identity certificates, taxation..). This will take time, but if the path is there, and there is demand, then it will happen.

I really like the fact that `@context` is now mandatory for the base structure of VCDM, while it allows first adopters to go with JSON and take the iteration path towards evolution of unified vocabs. If everyone creates their own vocabs, it's pretty much pointless work.

awoie:
> Great proposal to have the JSON-LD side thoroughly described. My own personal opinion is that JSON-LD will help greatly in the integration and after it is well adopted, this can even be used in VP Exchange to request proofs for a meaning.
> 
> On the other side, I see JSON as a gateway to implement JSON-LD vocabs. The vocab is not an easy task for governmental domains, which anyway yield many benefits for users (identity certificates, taxation..). This will take time, but if the path is there, and there is demand, then it will happen.
> 
> I really like the fact that `@context` is now mandatory for the base structure of VCDM, while it allows first adopters to go with JSON and take the iteration path towards evolution of unified vocabs. If everyone creates their own vocabs, it's pretty much pointless work.

Sorry, but IMO, this sounds like a tech is used for the sake of adopting a tech but I cannot see a good reason for using the tech in the first place.

[...]

PR #1270, opened 2023-09-08
Explain JSON-LD further.

BigBlueHat:
Addresses #1227.

That issue accumulated many references with many requests for it to address all maner of things. I'm confident I did not address them all. However, I hope this further explains the value of JSON-LD to the VCDM.

***
<a href="https://pr-preview.s3.amazonaws.com/BigBlueHat/vc-data-model/pull/1270.html" title="Last updated on Oct 2, 2023, 6:13 PM UTC (8416b8a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1270/929edc8...BigBlueHat:8416b8a.html" title="Last updated on Oct 2, 2023, 6:13 PM UTC (8416b8a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4009,7 +4009,8 @@ parsing approaches.
 
           <p>
 A JSON document becomes a [[!JSON-LD]] document when it includes an `@context`
-property with a URL associated with a context file. The context file provides
+property with a URL associated with a context file and is interpreted in an
+object-oriented style. The context file provides
 the mapping between the idiosyncratic in-document terminology and universally
 shareable identifiers and value definitions. This provides not only
 contextualization and meaning, but also reusability through universally
@@ -4030,7 +4031,7 @@ processing VCDM documents consistently with the context file, but without
 
           <p>
 The referenced context file provides the foundation for a consistent
-grammatical layer based on object-oriented modeling and community driven
+grammatical layer based on top of object-oriented modeling and community driven
 terminology selection which enables coexistence of various JSON tree shapes and
 colloquial terms while preventing likely terminology collisions or creating a
 need to write new tree traversal code for every new media type or JSON shape or
@@ -4047,7 +4048,9 @@ registered and specified while communities are free to build on that foundation
 through their own specification process. The terminology mappings are
 referenced from the VCDM document removing the need for an ever changing
 landscape of "private claims" or narrowly defined "public claims" housed in a
-centralized registry. The common grammar and object-oriented data modeling
+centralized registry. Instead, contexts function as decentralized registries
+that can be shared within or across communities. The common grammar and
+object-oriented data modeling
 remain the same and properly contextualized terms can commingle. Applications
 and communities can extend what is claimed and signed through the use of
 vocabularies.
@@ -4055,12 +4058,12 @@ vocabularies.
 
           <p>
 The underlying graph data model enables signing of the document's statements
-rather than just the envelope in which it's contained. When using this feature, the
-data — while signed — is accessible in the same way it had been prior to signing,
+rather than just an envelope in which it's contained. When using this feature, the
+data — while signed — is accessible in the same format it had been prior to signing,
 allowing for easier indexing, storage, and in-Wallet selection. The existence of
-the proof within the object-oriented graph model provides confidence that not
-only the wrapping structure was signed, but its actual informational contents.
-This enables selective disclosure of statements within the same JSON format
+one or more proofs within the object-oriented graph model, each fit for purpose,
+can be used to obtain confidence its actual informational contents are signed.
+This enables, for example, selective disclosure of statements within the same JSON format,
 avoiding the ever increasing cost of retooling for new disclosure formats or
 new envelope types. The proof is consistently placed alongside the claims.
           </p>
--- a/index.html
+++ b/index.html
@@ -3954,7 +3954,7 @@ or an array of values.
 The world of credentials is vast and varied. A physical wallet places
 relatively few constraints on the shape and size of those credentials–and it
 places no restriction on their contents. Likewise, Verifiable Credential Data
-Model’s use of [[!JSON-LD]] provides the consistent shape for all compliant
+Model's use of [[!JSON-LD]] provides the consistent shape for all compliant
 Verifiable Credentials, so that digital wallets can store credentials with
 widely varying content, providing value to a plethora of communities, and
 enabling many diverse use cases–from coupons to authentication.
@@ -3962,13 +3962,13 @@ enabling many diverse use cases–from coupons to authentication.
 
         <p>
 Consequently, Wallet providers can have the confidence that any Verifiable
-Credential will “fit” while Issuers and Verifiers benefit from a common
+Credential will "fit" while Issuers and Verifiers benefit from a common
 exchange representation that provides built in extensibility without the
 high costs from centralized terminology registration or JSON format churn.
         </p>
 
         <p>
-Issuers can add claims terminology unique to their communities via [[!JSON-LD]]’s
+Issuers can add claims terminology unique to their communities via [[!JSON-LD]]'s
 context files rather than be dependent on a single, centralized registry
 system. This webby decentralized approach provides infinite scalability. Future
 standardization efforts, then focus on vocabulary definitions and
@@ -3989,13 +3989,13 @@ grammatical structures and inbuilt object-oriented data modeling.
         </p>
 
         <p>
-The Verifiable Credential Data Model’s consistent object-oriented modeling
+The Verifiable Credential Data Model's consistent object-oriented modeling
 constraints are familiar to developers and enable better reuse and composition
 of data. That modeling is built from clean expressions of subject, property,
 values which comprise the shape of each atomic claim within a VCDM credential.
 This approach empowers Wallets to store any credential shaped in this way (as
 with physical wallets) and to index the contained claims made across all the
-holder’s credentials such that credentials may be quickly and easily found and
+holder's credentials such that credentials may be quickly and easily found and
 presented when needed.
         </p>
 
@@ -4044,7 +4044,7 @@ community or use case.
 Consequently, without any additional processing a JSON document gains the value
 of an in-message association with its terminology definitions. Details on
 processing VCDM documents consistently with the context file, but without
-[[!JSON-LD]] tooling can be found in the “JSON Processing” section below.
+[[!JSON-LD]] tooling can be found in the "JSON Processing" section below.
           </p>
 
           <p>
@@ -4065,7 +4065,7 @@ for every new application or community. The core data model terms are
 registered and specified while communities are free to build on that foundation
 through their own specification process. The terminology mappings are
 referenced from the VCDM document removing the need for an ever changing
-landscape of “private claims” or narrowly defined “public claims” housed in a
+landscape of "private claims" or narrowly defined "public claims" housed in a
 centralized registry. The common grammar and object-oriented data modeling
 remain the same and properly contextualized terms can commingle. Applications
 and communities can easily extend what is claimed and signed through the use of
@@ -4073,8 +4073,8 @@ vocabularies.
           </p>
 
           <p>
-The underpinning graph data model enables signing of the document’s statements
-rather than just the envelope it’s contained in. When using this feature, the
+The underpinning graph data model enables signing of the document's statements
+rather than just the envelope it's contained in. When using this feature, the
 data–while signed–is accessible in the same way it had been prior to signing
 allowing for easier indexing, storage, and in-Wallet selection.The existence of
 the proof within the object-oriented graph model provides confidence that not
--- a/index.html
+++ b/index.html
@@ -3960,7 +3960,7 @@ or an array of values.
   Credentials using a syntax that can be enforced using complementary
   technologies such as JSON Schema. This means issuers can add claim terminology
   unique to their communities without the high costs of centralized terminology
-  registration or format churn. In turn, digital wallets can index and store
+  registration or format churn. In turn, digital wallets can store and index
   credentials with widely varying content, providing value to a plethora of
   communities, and enabling many diverse use cases.
           </p>
@@ -3969,7 +3969,7 @@ or an array of values.
   consistent object-oriented modeling constraints that are familiar to
   developers and enable better reuse and composition of data. That modeling is
   built from clean expressions of <em>subjects, properties, and values</em>
-  (also referred to as <em>entities, attributes, and values</em>,
+  (<em>S-P-V</em>, also referred to as <em>entities, attributes, and values</em>,
   <em>E-A-V</em>, as well as <em>subjects, predicates, and objects</em>,
   <em>S-P-O</em>) which comprise the shape of each atomic claim within a VCDM
   credential. This approach empowers Wallets to store any credential shaped in
@@ -4031,7 +4031,7 @@ processing VCDM documents consistently with the context file, but without
 
           <p>
 The referenced context file provides the foundation for a consistent
-grammatical layer based on top of object-oriented modeling and community driven
+grammatical layer on top of object-oriented modeling and community-driven
 terminology selection which enables coexistence of various JSON tree shapes and
 colloquial terms while preventing likely terminology collisions or creating a
 need to write new tree traversal code for every new media type or JSON shape or
@@ -4051,7 +4051,7 @@ landscape of "private claims" or narrowly defined "public claims" housed in a
 centralized registry. Instead, contexts function as decentralized registries
 that can be shared within or across communities. The common grammar and
 object-oriented data modeling
-remain the same and properly contextualized terms can commingle. Applications
+remain the same, and properly contextualized terms can commingle. Applications
 and communities can extend what is claimed and signed through the use of
 vocabularies.
           </p>
@@ -4062,7 +4062,7 @@ rather than just an envelope in which it's contained. When using this feature, t
 data — while signed — is accessible in the same format it had been prior to signing,
 allowing for easier indexing, storage, and in-Wallet selection. The existence of
 one or more proofs within the object-oriented graph model, each fit for purpose,
-can be used to obtain confidence its actual informational contents are signed.
+can be used to increase confidence that its actual informational contents are signed.
 This enables, for example, selective disclosure of statements within the same JSON format,
 avoiding the ever increasing cost of retooling for new disclosure formats or
 new envelope types. The proof is consistently placed alongside the claims.
--- a/index.html
+++ b/index.html
@@ -3950,69 +3950,72 @@ or an array of values.
       <section>
         <h3>JSON-LD</h3>
 
-        <p>
-The world of credentials is vast and varied. A physical wallet places
-relatively few constraints on the shape and size of those credentials — and it
-places no restriction on their contents. Likewise, the Verifiable Credential Data
-Model's use of [[!JSON-LD]] provides the consistent shape for all compliant
-Verifiable Credentials, so that digital wallets can store credentials with
-widely varying content, providing value to a plethora of communities, and
-enabling many diverse use cases — from coupons to authentication.
-        </p>
+        <section class="informative">
+          <h4>Data Model Foundations</h4>
+          <p>
+  The world of credentials is vast and varied. A physical wallet places
+  relatively few constraints on the shape and size of those credentials — and it
+  places no restriction on their contents. Likewise, the Verifiable Credential Data
+  Model's use of [[!JSON-LD]] provides the consistent shape for all compliant
+  Verifiable Credentials, so that digital wallets can store credentials with
+  widely varying content, providing value to a plethora of communities, and
+  enabling many diverse use cases — from coupons to authentication.
+          </p>
 
-        <p>
-Consequently, Wallet providers can have the confidence that any Verifiable
-Credential will "fit", while Issuers and Verifiers benefit from a common
-exchange representation that provides built in extensibility without the
-high costs of centralized terminology registration or JSON format churn.
-        </p>
+          <p>
+  Consequently, Wallet providers can have the confidence that any Verifiable
+  Credential will "fit", while Issuers and Verifiers benefit from a common
+  exchange representation that provides built in extensibility without the
+  high costs of centralized terminology registration or JSON format churn.
+          </p>
 
-        <p>
-Issuers can add claim terminology unique to their communities via [[!JSON-LD]]'s
-context files rather than being dependent on a single, centralized registry
-system. This webby, decentralized approach provides infinite scalability. Future
-standardization efforts may then focus on vocabulary definitions and
-object-oriented modeling provided by communities of practice rather than idiosyncratic
-JSON format creation for individual use cases or scenarios.
-        </p>
+          <p>
+  Issuers can add claim terminology unique to their communities via [[!JSON-LD]]'s
+  context files rather than being dependent on a single, centralized registry
+  system. This webby, decentralized approach provides infinite scalability. Future
+  standardization efforts may then focus on vocabulary definitions and
+  object-oriented modeling provided by communities of practice rather than idiosyncratic
+  JSON format creation for individual use cases or scenarios.
+          </p>
 
-        <p>
-Furthermore, Verifiers can furthermore lean on similarly provided proof mechanisms
-without writing new document parsers for each new security mechanism.
-        </p>
+          <p>
+  Furthermore, Verifiers can furthermore lean on similarly provided proof mechanisms
+  without writing new document parsers for each new security mechanism.
+          </p>
 
-        <p>
-Lastly, Wallets can move these consistently-shaped credentials between
-Issuers and Verifiers without needing to create new storage systems for each
-format, with the confidence that each relying party can depend on consistent
-grammatical structures and in-built object-oriented data modeling.
-        </p>
+          <p>
+  Lastly, Wallets can move these consistently-shaped credentials between
+  Issuers and Verifiers without needing to create new storage systems for each
+  format, with the confidence that each relying party can depend on consistent
+  grammatical structures and in-built object-oriented data modeling.
+          </p>
 
-        <p>
-The consistent object-oriented modeling constraints of the Verifiable
-Credential Data Model (VCDM) are familiar to developers and enable better
-reuse and composition of data. That modeling is built from clean expressions
-of <em>subjects, properties, and values</em> (also referred to as <em>entities,
-attributes, and values</em>, <em>E-A-V</em>, as well as <em>subjects, predicates, and
-objects</em>, <em>S-P-O</em>) which comprise the shape of each atomic claim within
-a VCDM credential. This approach empowers Wallets to store any credential
-shaped in this way (as with physical wallets) and to index the contained
-claims across all the holder's credentials, such that credentials may be
-quickly and easily found and presented when needed.
-        </p>
+          <p>
+  The consistent object-oriented modeling constraints of the Verifiable
+  Credential Data Model (VCDM) are familiar to developers and enable better
+  reuse and composition of data. That modeling is built from clean expressions
+  of <em>subjects, properties, and values</em> (also referred to as <em>entities,
+  attributes, and values</em>, <em>E-A-V</em>, as well as <em>subjects, predicates, and
+  objects</em>, <em>S-P-O</em>) which comprise the shape of each atomic claim within
+  a VCDM credential. This approach empowers Wallets to store any credential
+  shaped in this way (as with physical wallets) and to index the contained
+  claims across all the holder's credentials, such that credentials may be
+  quickly and easily found and presented when needed.
+          </p>
 
-        <p>
-This foundational data modeling and grammatical structure is provided through
-the addition of a single `@context` key which identifies context files that map
-otherwise idiosyncratic and often ephemeral terms into universally shareable,
-collision resistant vocabularies. This context mechanism provides the
-underlying subject, property, and value graph-based data model which enables
-aggregation, collation, and even separation of the individual statements held
-within the [[!JSON-LD]] graph and brings the object-oriented structure to previously
-freeform JSON documents.
-        </p>
+          <p>
+  This foundational data modeling and grammatical structure is provided through
+  the addition of a single `@context` key which identifies context files that map
+  otherwise idiosyncratic and often ephemeral terms into universally shareable,
+  collision resistant vocabularies. This context mechanism provides the
+  underlying subject, property, and value graph-based data model which enables
+  aggregation, collation, and even separation of the individual statements held
+  within the [[!JSON-LD]] graph and brings the object-oriented structure to previously
+  freeform JSON documents.
+          </p>
+        </section>
 
-        <section>
+        <section class="informative">
           <h3>Extending JSON</h3>
 
           <p>
--- a/index.html
+++ b/index.html
@@ -3952,51 +3952,51 @@ or an array of values.
 
         <p>
 The world of credentials is vast and varied. A physical wallet places
-relatively few constraints on the shape and size of those credentials–and it
-places no restriction on their contents. Likewise, Verifiable Credential Data
+relatively few constraints on the shape and size of those credentials — and it
+places no restriction on their contents. Likewise, the Verifiable Credential Data
 Model's use of [[!JSON-LD]] provides the consistent shape for all compliant
 Verifiable Credentials, so that digital wallets can store credentials with
 widely varying content, providing value to a plethora of communities, and
-enabling many diverse use cases–from coupons to authentication.
+enabling many diverse use cases — from coupons to authentication.
         </p>
 
         <p>
 Consequently, Wallet providers can have the confidence that any Verifiable
-Credential will "fit" while Issuers and Verifiers benefit from a common
+Credential will "fit", while Issuers and Verifiers benefit from a common
 exchange representation that provides built in extensibility without the
-high costs from centralized terminology registration or JSON format churn.
+high costs of centralized terminology registration or JSON format churn.
         </p>
 
         <p>
-Issuers can add claims terminology unique to their communities via [[!JSON-LD]]'s
-context files rather than be dependent on a single, centralized registry
-system. This webby decentralized approach provides infinite scalability. Future
-standardization efforts, then focus on vocabulary definitions and
-object-oriented modeling provided by communities of practice vs. idiosyncratic
+Issuers can add claim terminology unique to their communities via [[!JSON-LD]]'s
+context files rather than being dependent on a single, centralized registry
+system. This webby, decentralized approach provides infinite scalability. Future
+standardization efforts may then focus on vocabulary definitions and
+object-oriented modeling provided by communities of practice rather than idiosyncratic
 JSON format creation for individual use cases or scenarios.
         </p>
 
         <p>
-Verifiers can furthermore lean on the similarly provided proof mechanisms
+Furthermore, Verifiers can furthermore lean on similarly provided proof mechanisms
 without writing new document parsers for each new security mechanism.
         </p>
 
         <p>
-Lastly, Wallets can then move these consistently shaped credentials between
-Issuers and Verifiers without the need to create new storage systems for each
-format with the confidence that each relying party can depend on consistent
-grammatical structures and inbuilt object-oriented data modeling.
+Lastly, Wallets can move these consistently-shaped credentials between
+Issuers and Verifiers without needing to create new storage systems for each
+format, with the confidence that each relying party can depend on consistent
+grammatical structures and in-built object-oriented data modeling.
         </p>
 
         <p>
 The Verifiable Credential Data Model's consistent object-oriented modeling
 constraints are familiar to developers and enable better reuse and composition
 of data. That modeling is built from clean expressions of subject, property,
-values which comprise the shape of each atomic claim within a VCDM credential.
-This approach empowers Wallets to store any credential shaped in this way (as
-with physical wallets) and to index the contained claims made across all the
-holder's credentials such that credentials may be quickly and easily found and
-presented when needed.
+and values which comprise the shape of each atomic claim within a Verifiable
+Credential Data Model credential. This approach empowers Wallets to store any
+credential shaped in this way (as with physical wallets) and to index the
+contained claims made across all the holder's credentials such that credentials
+may be quickly and easily found and presented when needed.
         </p>
 
         <p>
@@ -4004,7 +4004,7 @@ This foundational data modeling and grammatical structure is provided through
 the addition of a single `@context` key which identifies context files that map
 otherwise idiosyncratic and often ephemeral terms into universally shareable,
 collision resistant vocabularies. This context mechanism provides the
-underlying subject, property, value graph-based data model which enables
+underlying subject, property, and value graph-based data model which enables
 aggregation, collation, and even separation of the individual statements held
 within the [[!JSON-LD]] graph and brings the object-oriented structure to previously
 freeform JSON documents.
@@ -4044,7 +4044,7 @@ community or use case.
 Consequently, without any additional processing a JSON document gains the value
 of an in-message association with its terminology definitions. Details on
 processing VCDM documents consistently with the context file, but without
-[[!JSON-LD]] tooling can be found in the "JSON Processing" section below.
+[[!JSON-LD]] tooling, can be found in the "JSON Processing" section below.
           </p>
 
           <p>
@@ -4057,7 +4057,7 @@ parsing approach.
           </p>
 
           <p>
-Verifiable Credentials Data Model uses this consistent foundation to provide a
+The Verifiable Credentials Data Model uses this consistent foundation to provide a
 common layer across all credentials, presentations, and transaction formats
 within a single, consistently parseable media type definition. This allows for
 ecosystem extensibility without requiring a centralized registry to be updated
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3951,41 +3951,191 @@ <h2>Syntaxes</h2>
...
+
+          <p>
+[[!JSON-LD]] provides Verifiable Credentials with a consistent grammatical
+foundation, object-oriented modeling, and collision-free terminology sharing
+which allows for ecosystem growth, document format iteration (rather than
+reinvention), and an extensible framework for new cryptographic proof
+mechanisms without altering or damaging the underlying data structure. This
+creates an economies of scale which reduces development time, decreased
+retooling cost, increases time to market, and provides long term stability for
+Issuers, Verifiers, and Wallets.
```
davidlehn, line 4095:
```suggestion
mechanisms without altering or damaging the underlying data structure. This
creates economies of scale which reduces development time, decreases
retooling cost, increases time to market, and provides long term stability for
Issuers, Verifiers, and Wallets.
```

index.html
```diff
@@ -3951,41 +3951,191 @@ <h2>Syntaxes</h2>
         <h3>JSON-LD</h3>
 
         <p>
+The world of credentials is vast and varied. A physical wallet places
+relatively few constraints on the shape and size of those credentials–and it
+places no restriction on their contents. Likewise, Verifiable Credential Data
+Model’s use of [[!JSON-LD]] provides the consistent shape for all compliant
```
decentralgabe, line 3957: I'd argue JSON Schema provides a more consistent "shape" than LD does...

dlongley: > I'd argue JSON Schema provides a more consistent "shape" than LD does...

JSON schema doesn't provide a more consistent shape; it is an enforcement or validation mechanism for whatever idiosyncratic shape the author wants to use. There's no implicit constraint in how the author does their modeling, rather, JSON schema is a mechanism to enable testing whether some particular data fits their chosen shape (which can be anything). That's certainly useful and complementary technology, but not the same thing being said here.

The text in this PR is about the advantage of consistent data modeling through simple constraints such as using an object-oriented approach and the fact that this shape consistently maps to atomic statements (claims) of subject => predicate => value.

index.html
```diff
@@ -3951,41 +3951,191 @@ <h2>Syntaxes</h2>
         <h3>JSON-LD</h3>
 
         <p>
+The world of credentials is vast and varied. A physical wallet places
+relatively few constraints on the shape and size of those credentials–and it
+places no restriction on their contents. Likewise, Verifiable Credential Data
```
TallTed, line 3956:
```suggestion
relatively few constraints on the shape and size of those credentials — and it
places no restriction on their contents. Likewise, the Verifiable Credential Data
```

index.html
```diff
@@ -3951,41 +3951,191 @@ <h2>Syntaxes</h2>
...
+Model’s use of [[!JSON-LD]] provides the consistent shape for all compliant
+Verifiable Credentials, so that digital wallets can store credentials with
+widely varying content, providing value to a plethora of communities, and
+enabling many diverse use cases–from coupons to authentication.
```
TallTed, line 3960:
```suggestion
enabling many diverse use cases — from coupons to authentication.
```

index.html
```diff
@@ -3951,41 +3951,191 @@ <h2>Syntaxes</h2>
...
+        </p>
+
+        <p>
+Consequently, Wallet providers can have the confidence that any Verifiable
+Credential will “fit” while Issuers and Verifiers benefit from a common
+exchange representation that provides built in extensibility without the
+high costs from centralized terminology registration or JSON format churn.
```
TallTed, line 3967:
```suggestion
Consequently, Wallet providers can have the confidence that any Verifiable
Credential will "fit", while Issuers and Verifiers benefit from a common
exchange representation that provides built in extensibility without the
high costs of centralized terminology registration or JSON format churn.
```

index.html
```diff
@@ -3951,41 +3951,191 @@ <h2>Syntaxes</h2>
...
+        </p>
+
+        <p>
+Issuers can add claims terminology unique to their communities via [[!JSON-LD]]’s
+context files rather than be dependent on a single, centralized registry
+system. This webby decentralized approach provides infinite scalability. Future
+standardization efforts, then focus on vocabulary definitions and
+object-oriented modeling provided by communities of practice vs. idiosyncratic
+JSON format creation for individual use cases or scenarios.
```
TallTed, line 3976:
```suggestion
Issuers can add claim terminology unique to their communities via [[!JSON-LD]]’s
context files rather than being dependent on a single, centralized registry
system. This webby, decentralized approach provides infinite scalability. Future
standardization efforts may then focus on vocabulary definitions and
object-oriented modeling provided by communities of practice rather than idiosyncratic
JSON format creation for individual use cases or scenarios.
```

index.html
```diff
@@ -3951,41 +3951,191 @@ <h2>Syntaxes</h2>
...
+        </p>
+
+        <p>
+Verifiers can furthermore lean on the similarly provided proof mechanisms
+without writing new document parsers for each new security mechanism.
```
TallTed, line 3981:
```suggestion
Furthermore, Verifiers can lean on similarly provided proof mechanisms
without writing new document parsers for each new security mechanism.
```

index.html
```diff
@@ -3951,41 +3951,191 @@ <h2>Syntaxes</h2>
...
+        </p>
+
+        <p>
+Lastly, Wallets can then move these consistently shaped credentials between
+Issuers and Verifiers without the need to create new storage systems for each
+format with the confidence that each relying party can depend on consistent
+grammatical structures and inbuilt object-oriented data modeling.
```
TallTed, line 3988:
```suggestion
Lastly, Wallets can move these consistently-shaped credentials between
Issuers and Verifiers without needing to create new storage systems for each
format, with confidence that each relying party can depend on consistent
grammatical structures and built-in object-oriented data modeling.
```

index.html
```diff
@@ -3951,41 +3951,191 @@ <h2>Syntaxes</h2>
...
+
+        <p>
+The Verifiable Credential Data Model’s consistent object-oriented modeling
+constraints are familiar to developers and enable better reuse and composition
+of data. That modeling is built from clean expressions of subject, property,
+values which comprise the shape of each atomic claim within a VCDM credential.
+This approach empowers Wallets to store any credential shaped in this way (as
+with physical wallets) and to index the contained claims made across all the
+holder’s credentials such that credentials may be quickly and easily found and
+presented when needed.
```
TallTed, line 3999:
```suggestion
The consistent object-oriented modeling constraints of the Verifiable
Credential Data Model (VCDM) are familiar to developers and enable better
reuse and composition of data. That modeling is built from clean expressions
of _subjects, properties, and values_ (also referred to as _entities,
attributes, and values_, _E-A-V_, as well as _subjects, predicates, and
objects_, _S-P-O_,) which comprise the shape of each atomic claim within
a VCDM credential. This approach empowers Wallets to store any credential
shaped in this way (as with physical wallets) and to index the contained
claims across all the holder's credentials, such that credentials may be
quickly and easily found and presented when needed.
```

index.html
```diff
@@ -3951,41 +3951,191 @@ <h2>Syntaxes</h2>
...
+credential definitions will focus on the use of their terminology rather than
+force implementers to parse another unique JSON format specific to that
+community or use case.
+          </p>
+
+          <p>
+Consequently, without any additional processing a JSON document gains the value
+of an in-message association with its terminology definitions. Details on
+processing VCDM documents consistently with the context file, but without
+[[!JSON-LD]] tooling can be found in the “JSON Processing” section below.
```
TallTed, line 4047:
```suggestion
[[!JSON-LD]] tooling, can be found in the "JSON Processing" section below.
```

index.html
```diff
@@ -3951,41 +3951,191 @@ <h2>Syntaxes</h2>
...
+The referenced context file provides the foundation for a consistent
+grammatical layer based on object-oriented modeling and community driven
+terminology selection which enables coexistence of various JSON tree shapes and
+colloquial terms while preventing likely terminology collisions or creating a
+need to write new tree traversal code for every new media type or JSON shape or
+parsing approach.
+          </p>
+
+          <p>
+Verifiable Credentials Data Model uses this consistent foundation to provide a
```
TallTed, line 4060:
```suggestion
The Verifiable Credentials Data Model uses this consistent foundation to provide a
```

index.html
```diff
@@ -3951,41 +3951,191 @@ <h2>Syntaxes</h2>
...
+centralized registry. The common grammar and object-oriented data modeling
+remain the same and properly contextualized terms can commingle. Applications
+and communities can easily extend what is claimed and signed through the use of
+vocabularies.
+          </p>
+
+          <p>
+The underpinning graph data model enables signing of the document’s statements
+rather than just the envelope it’s contained in. When using this feature, the
+data–while signed–is accessible in the same way it had been prior to signing
```
TallTed, line 4078:
```suggestion
The underpinning graph data model enables signing of the document's statements
rather than just the envelope in which it's contained. When using this feature, the
data — while signed — is accessible in the same way it had been prior to signing,
```

index.html
```diff
@@ -3951,41 +3951,191 @@ <h2>Syntaxes</h2>
...
+allowing for easier indexing, storage, and in-Wallet selection.The existence of
```
TallTed, line 4079:
```suggestion
allowing for easier indexing, storage, and in-Wallet selection. The existence of
```

index.html
```diff
@@ -3951,41 +3951,191 @@ <h2>Syntaxes</h2>
...
+new envelope types. The proof is consistently placed alongside the claims.
+          </p>
+
+          <p>
+[[!JSON-LD]] provides Verifiable Credentials with a consistent grammatical
+foundation, object-oriented modeling, and collision-free terminology sharing
+which allows for ecosystem growth, document format iteration (rather than
+reinvention), and an extensible framework for new cryptographic proof
+mechanisms without altering or damaging the underlying data structure. This
+creates an economies of scale which reduces development time, decreased
```
TallTed, line 4093:
```suggestion
creates an economy of scale which reduces development time, decreases
```

index.html
```diff
@@ -3951,41 +3951,191 @@ <h2>Syntaxes</h2>
...
+retooling cost, increases time to market, and provides long term stability for
```
TallTed, line 4094:
```suggestion
retooling cost, decreases time to market, and provides long term stability for
```

TallTed (changes requested):

index.html
```diff
@@ -3951,41 +3951,191 @@ <h2>Syntaxes</h2>
...
+within a single, consistently parseable media type definition. This allows for
+ecosystem extensibility without requiring a centralized registry to be updated
+for every new application or community. The core data model terms are
+registered and specified while communities are free to build on that foundation
+through their own specification process. The terminology mappings are
+referenced from the VCDM document removing the need for an ever changing
+landscape of “private claims” or narrowly defined “public claims” housed in a
+centralized registry. The common grammar and object-oriented data modeling
+remain the same and properly contextualized terms can commingle. Applications
+and communities can easily extend what is claimed and signed through the use of
```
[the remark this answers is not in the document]
BigBlueHat: Agree that "ease" is subjective.

index.html
BigBlueHat: @decentralgabe JSON Schema (by design) only validates a "surface syntax" not any underlying data modeling. JSON-LD provided mapping tree structures into a graph-based data model that can exist across many different tree "shapes."

There can't ever be a single JSON Schema to validate all JSON structures, but there can be "universal" JSON-LD storage, construction, re-construction, aggregation, etc. because the data model remains the same regardless of the tree shape.

I realize the use of the word "shape" in the physical wallet analogy causes that confusion...I'll look into how we could resolve that, but I know of no other way to provide a universally available data structure in JSON (or even across CBOR, YAML, etc.) that can survive the surface syntax shift other than JSON-LD.

index.html
BigBlueHat: Added with credit in the commit message. 😄

selfissued (changes requested):
Most of this text is not normative requirements that affect implementations.  Please pare it down to limit what it says to normative requirements.

BigBlueHat:
> Most of this text is not normative requirements that affect implementations. Please pare it down to limit what it says to normative requirements.

Looks like I forgot the "non-normative" intro text. I'll add it. I agree it can use paring down, but given the group continues to attempt to span to ecosystems of parsing approaches--in which there are multiple "normative" ways to do most things--the connection this makes to the JSON-LD ecosystem and the benefits available within it is rather the point and focus of the text.

Happy to entertain specific reduction/simplification requests!

BigBlueHat:
> Most of this text is not normative requirements that affect implementations. Please pare it down to limit what it says to normative requirements.

The text is now marked as "informative". I'll continue to work to pare it down as more reviews come in--but given that "pare down" is not a defined target, if you could re-review or un-minus-one until there's a substantive thing I can address, I'd greatly appreciate it!

index.html
decentralgabe: @dlongley my point is that JSON LD usage commonly defines a large set of terms and then in practice a subset of those terms are used without any way for the context creator to dictate which terms are required or not -- that's not really a shape. A JSON Schema on the other hand allows specifying required and optional properties, which are constraints I view was necessary to even have a 'shape.'

@BigBlueHat you're correct re:surface syntax vs 'underlying' data modeling if you consider semantics and graphs to be a requisite for data modeling.

I'd recommend rephrasing to emphasize semantics, graph modeling, and term definition space instead of 'shape.'

index.html
```diff
@@ -3950,42 +3950,197 @@ <h2>Syntaxes</h2>
...
 
-        <p>
+        <section class="informative">
+          <h4>Data Model Foundations</h4>
+          <p>
+  The world of credentials is vast and varied. A physical wallet places
+  relatively few constraints on the shape and size of those credentials — and it
+  places no restriction on their contents. Likewise, the Verifiable Credential Data
+  Model's use of [[!JSON-LD]] provides the consistent shape for all compliant
+  Verifiable Credentials, so that digital wallets can store credentials with
```
decentralgabe, line 3960: I do not think we should use the term 'digital wallet' at all, it's confusing. I like the term agent. I don't think we'll agree on a terminology so I'd omit it entirely.

index.html
dlongley: @decentralgabe,

I think you actually want both (JSON-LD and JSON schema) -- and they apply different kinds of constraints on the "shape" of data.

JSON-LD constrains the modeling and how it will be interpreted consistently as described above, and JSON schema enforces whatever schema someone invents to describe that modeling. JSON schema alone doesn't constrain what gets modeled or how it will be interpreted; you can make whatever schema you want and JSON schema says nothing about the interpretation -- but it will enforce syntactic expression. JSON-LD alone doesn't add as much enforcement of its syntactic constraints as JSON schema does, but it ensures the modeling will be interpreted according to the constraints even if the author tries to deviate. So both are useful tools that play different roles in shaping data (model wise and syntactically). Neither has a single claim to "shape" constraining, they just work in different dimensions.

index.html
```diff
@@ -3950,42 +3950,197 @@ <h2>Syntaxes</h2>
...
+
+          <p>
+[[!JSON-LD]] provides Verifiable Credentials with a consistent grammatical
+foundation, object-oriented modeling, and collision-free terminology sharing
+which allows for ecosystem growth, document format iteration (rather than
+reinvention), and an extensible framework for new cryptographic proof
+mechanisms without altering or damaging the underlying data structure. This
+creates economies of scale which reduces development time, decreases
+retooling cost, decreases time to market, and provides long term stability for
+Issuers, Verifiers, and Wallets.
```
Sakurann, line 4100:
```suggestion
mechanisms without altering or damaging the underlying data structure.
```

nadalin:
not in favor of this PR

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-09-14-vcwg#section1-4) on 2023-09-14

- no resolutions were taken

#### 1.4. Explain JSON-LD further. (pr vc-data-model#1270)

_See github pull request [vc-data-model#1270](https://github.com/w3c/vc-data-model/pull/1270)._




**Brent Zundel:** Explain json-ld further. A section that was widely asked for. Has had comments back and forth, no approvals. More work to be done. We'll time box to 10 minutes what changes are needed from folks.  

**Michael Jones:** This PR is a lot of philosophy and not a lot that helps implementers.  

**Manu Sporny:** The group asked us to write this. They asked for "we want to see a section that explains json-ld". We have other places in the spec that explain why some choices were made.  

**Sebastian Crane:** It looks good on a first impression. I would prefer for it to be cut back. Not because it is incorrect, but it includes some things that sounds like opinions. This makes folks uncomfortable. I would suggest som word smithing. I would object to merging now, but ok in the future.  

**Brent Zundel:** Suggestion is to make concrete suggestions.  

**Michael Jones:** There are 14 new paragraphs. Please cut it down accordingly.  

**Ted Thibodeau Jr.:** The entire section is informative. It might be better as an appendix.  
� It doesn't talk about JSON, but its use. That's the way it is.  
� If Mike really thinks it could be said in a single paragraph, please suggest how to concretely do it.  

> *Kristina Yasuda:* we need something in main text as opposed to appendix.

> *Shigeya Suzuki:* +1 kristina.

**Brent Zundel:** Happy to hear thought about the appendix suggestion.  

**Manu Sporny:** +1 what TallTed said. It's really difficult to understand the changes requested without suggested changes. Please either write a PR, or make concrete github suggestions.  

**Kristina Yasuda:** Re: moving to appendix. I do believe we need a few short and sweet paragraphs in the main text. More details can go in the appendix. Doesn't need to be pre-cr. Bottom line, the goal is to give folks some background on JSON-ld before sections that work off of it.  

> *Manu Sporny:* +1 kristina, keep shorter section in the main spec... but don't feel super strongly about it.

**Brent Zundel:** Since this PR is non-normative, it's not urgent to merge this PR.  

</details>

BigBlueHat:
> not in favor of this PR, suggest that this PR not be merged

Could you provide something more concrete to review, discuss, or respond to?

index.html
```diff
@@ -3950,42 +3950,197 @@ <h2>Syntaxes</h2>
...
+          <p>
+  The world of credentials is vast and varied. A physical wallet places
+  relatively few constraints on the shape and size of those credentials — and it
+  places no restriction on their contents. Likewise, the Verifiable Credential Data
+  Model's use of [[!JSON-LD]] provides the consistent shape for all compliant
+  Verifiable Credentials, so that digital wallets can store credentials with
+  widely varying content, providing value to a plethora of communities, and
+  enabling many diverse use cases — from coupons to authentication.
+          </p>
+
```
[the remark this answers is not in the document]
BigBlueHat: This is the foundational comparison being made... Could you explain why you'd like it removed?

index.html
BigBlueHat: Shared foundational data models and consistent data processing pipelines do quantifiably reduce development time, decrease retooling cost, etc. Could you expand on why you'd like this statement removed?

index.html
decentralgabe: > The phrase "digital wallet" is already in use within the VCDM spec.

I believe this is a mistake we should not double down on.

> Also, the comparison is between "physical wallet" and "digital wallet" so introducing a non-parallel term would add more confusion than clarity.

I believe it's a false comparison, which is why it's dangerous. A digital wallet is much more feature rich, and consequently complex, with many more considerations in all directions. So I agree with @Sakurann 's comment below to remove the text entirely.

index.html
```diff
@@ -3950,42 +3950,197 @@ <h2>Syntaxes</h2>
...
+referenced from the VCDM document removing the need for an ever changing
+landscape of "private claims" or narrowly defined "public claims" housed in a
+centralized registry. The common grammar and object-oriented data modeling
+remain the same and properly contextualized terms can commingle. Applications
+and communities can extend what is claimed and signed through the use of
+vocabularies.
+          </p>
+
+          <p>
+The underpinning graph data model enables signing of the document's statements
```
peacekeeper, line 4081:
```suggestion
The underlying graph data model enables signing of the document's statements
```

index.html
```diff
@@ -4057,26 +4062,26 @@ <h3>Extending JSON</h3>
...
 vocabularies.
           </p>
 
           <p>
-The underpinning graph data model enables signing of the document’s statements
-rather than just the envelope it’s contained in. When using this feature, the
-data–while signed–is accessible in the same way it had been prior to signing
-allowing for easier indexing, storage, and in-Wallet selection.The existence of
+The underpinning graph data model enables signing of the document's statements
+rather than just the envelope in which it's contained in. When using this feature, the
```
TallTed, line 4082:
```suggestion
rather than just the envelope in which it's contained. When using this feature, the
```

index.html
```diff
@@ -3950,42 +3950,197 @@ <h2>Syntaxes</h2>
...
+  Issuers can add claim terminology unique to their communities via [[!JSON-LD]]'s
+  context files rather than being dependent on a single, centralized registry
+  system. This webby, decentralized approach provides infinite scalability. Future
+  standardization efforts may then focus on vocabulary definitions and
+  object-oriented modeling provided by communities of practice rather than idiosyncratic
+  JSON format creation for individual use cases or scenarios.
+          </p>
+
+          <p>
+  Furthermore, Verifiers can furthermore lean on similarly provided proof mechanisms
```
wip-abramson, line 3982:
```suggestion
  Furthermore, Verifiers can lean on similarly provided proof mechanisms
```

index.html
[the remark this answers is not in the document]
BigBlueHat: @Sakurann I've greatly reduced this intro section. Hopefully it removes any concerns.

index.html
BigBlueHat: @Sakurann this paragraph remains unchanged in the most recent revision, so please let me know if there are specific complaints with the above claims. Thanks!

index.html
BigBlueHat: @decentralgabe this intro section has been reduced significantly and the analogy minimized considerably. I hope it addresses your concern!

BigBlueHat:
The intro is significantly reduced--with special thanks to @dlongley for his collaboration!

dlongley (approved):

index.html
```diff
@@ -3953,64 +3953,40 @@ <h3>JSON-LD</h3>
...
-  widely varying content, providing value to a plethora of communities, and
-  enabling many diverse use cases — from coupons to authentication.
+  The world of credentials is vast and varied. A physical wallet places a few
+  constraints on the shape and size of those credentials, and no restriction on
+  their contents. Likewise, the Verifiable Credential Data Model's use of
+  [[!JSON-LD]] provides a consistent way to interpret all compliant Verifiable
+  Credentials using a syntax that can be enforced using complementary
+  technologies such as JSON Schema. This means issuers can add claim terminology
+  unique to their communities without the high costs of centralized terminology
+  registration or format churn. In turn, digital wallets can index and store
```
TallTed, line 3963: Indexing doesn't happen without storage, so storage should be mentioned first.
```suggestion
  registration or format churn. In turn, digital wallets can store and index
```

index.html
```diff
@@ -3953,64 +3953,40 @@ <h3>JSON-LD</h3>
...
-  objects</em>, <em>S-P-O</em>) which comprise the shape of each atomic claim within
-  a VCDM credential. This approach empowers Wallets to store any credential
-  shaped in this way (as with physical wallets) and to index the contained
-  claims across all the holder's credentials, such that credentials may be
-  quickly and easily found and presented when needed.
+  [[!JSON-LD]] provides the Verifiable Credential Data Model (VCDM) with
+  consistent object-oriented modeling constraints that are familiar to
+  developers and enable better reuse and composition of data. That modeling is
+  built from clean expressions of <em>subjects, properties, and values</em>
+  (also referred to as <em>entities, attributes, and values</em>,
```
TallTed, line 3972:
```suggestion
  (<em>S-P-V</em>, also referred to as <em>entities, attributes, and values</em>,
```

index.html
```diff
@@ -4054,7 +4031,7 @@ <h3>Extending JSON</h3>
 
           <p>
 The referenced context file provides the foundation for a consistent
-grammatical layer based on object-oriented modeling and community driven
+grammatical layer based on top of object-oriented modeling and community driven
```
TallTed, line 4034:
```suggestion
grammatical layer on top of object-oriented modeling and community-driven
```

index.html
```diff
@@ -4071,20 +4048,22 @@ <h3>Extending JSON</h3>
 through their own specification process. The terminology mappings are
 referenced from the VCDM document removing the need for an ever changing
 landscape of "private claims" or narrowly defined "public claims" housed in a
-centralized registry. The common grammar and object-oriented data modeling
+centralized registry. Instead, contexts function as decentralized registries
+that can be shared within or across communities. The common grammar and
+object-oriented data modeling
 remain the same and properly contextualized terms can commingle. Applications
```
TallTed, line 4054:
```suggestion
remain the same, and properly contextualized terms can commingle. Applications
```

index.html
```diff
@@ -4071,20 +4048,22 @@ <h3>Extending JSON</h3>
...
-data — while signed — is accessible in the same way it had been prior to signing,
+The underlying graph data model enables signing of the document's statements
+rather than just an envelope in which it's contained. When using this feature, the
+data — while signed — is accessible in the same format it had been prior to signing,
 allowing for easier indexing, storage, and in-Wallet selection. The existence of
-the proof within the object-oriented graph model provides confidence that not
-only the wrapping structure was signed, but its actual informational contents.
-This enables selective disclosure of statements within the same JSON format
+one or more proofs within the object-oriented graph model, each fit for purpose,
+can be used to obtain confidence its actual informational contents are signed.
```
TallTed, line 4065:
```suggestion
can be used to increase confidence that its actual informational contents are signed.
```

selfissued (changes requested):
This PR adds *a lot* of informative text that isn't actionable by implementers.  It's more philosophy than specification.

It was the job of the JSON-LD spec to explain JSON-LD.  Please simply cite it, possibly with a section reference, along the lines of "For more about JSON-LD, see [ref]" and stop there.  The current contents of this PR doesn't belong in our specification.

OR13 (changes requested):
I've like to see specific text providing guidance on processing the core data model as RDF.

> JSON-LD is a [concrete RDF syntax](https://www.w3.org/TR/rdf11-concepts/#dfn-concrete-rdf-syntax) as described in [[RDF11-CONCEPTS](https://www.w3.org/TR/json-ld11/#bib-rdf11-concepts)]. Hence, a JSON-LD document is both an RDF document and a JSON document and correspondingly represents an instance of an [RDF data model](https://www.w3.org/TR/rdf11-concepts/#data-model).

- https://www.w3.org/TR/json-ld11/#relationship-to-rdf

Some ideas to help frame what I would like to see:

- https://www.w3.org/TR/json-ld11/#advanced-context-usage
- https://www.w3.org/TR/json-ld11-framing/
- https://www.w3.org/TR/sparql11-overview/

Basically, we have a lot of text that says "you don't need to process the core data model as JSON-LD".... but we have yet to show any value for processing the core data model as something other than JSON (as JSON-LD, which by definition is also RDF).

I think it would be best to address this directly in the context of related work at W3C, for example:

https://github.com/WICG/identity-credential/blob/main/identity-credential-proposal.md#mdocs

Notice the way the above specification does not use RDF predicates... Why would it be better to use RDF predicates and JSON-LD frames?

If there is no benefit to processing the data model as RDF, we should state that explicitly, and warn implementers that RDF predicates and RDF types are not stable extension or interoperability points.

BigBlueHat:
> Basically, we have a lot of text that says "you don't need to process the core data model as JSON-LD".... but we have yet to show any value for processing the core data model as something other than JSON (as JSON-LD, which by definition is also RDF).

We've had this backwards. Developers always process the core data model as JSON-LD, they may just be using the more limited approach defined in the JSON Processing section. JSON-LD is a syntax and you can access it directly. You can do things like pre-process a context once (or have a trusted party do this for you) and then write your code against supported contexts or context hash values. This is still doing JSON-LD processing…just a more limited form of it which attempts to optimize it based on accepting only specific inputs.

The content we have in the spec today that explains how to do these optimizations in JSON parsing land is unnecessary when you use a JSON-LD library to convert from one context to another or to RDF, but not doing that does not mean you're not doing JSON-LD processing. We should probably rename that section to "Optimizing JSON-LD processing (for specific use)" instead of "JSON Processing". Again, JSON-LD is a syntax and developers should read it as long as it is being provided with a context you accept. JSON-LD tells you the context of the data as currently expressed. This is true even if one chooses to process JSON-LD in a more limited way. It’s still JSON-LD.

Additionally, these more limited JSON parsers do not do the following (at least not without nearly recreating JSON-LD processing...):
 - handle term consistency--they always assume the key names mean the same thing unless they're within very narrowly defined media type namespaces
 - they do not have constructs for advanced data types
 - nor language value objects
 - nor a consistent extensibility mechanism (other than one media type per document use)
 - nor underlying graph model which enables aggregation, faceted querying, statement signing/extraction/reuse

All of this is expressed in what I've written.

> https://github.com/WICG/identity-credential/blob/main/identity-credential-proposal.md#mdocs
> 
> Notice the way the above specification does not use RDF predicates... Why would it be better to use RDF predicates and JSON-LD frames?

This happens everywhere there's a belief that English words mean the same thing within all the communities that use them. You can read through the various specifications which they are trying to combine to find that's not the case. Each of those specs have different type systems, value definitions, etc. All of those facts could coexist more easily when mapped to non-conflicting identifiers, value definitions, etc. The [reconciliation](https://github.com/WICG/identity-credential/blob/main/identity-credential-proposal.md#reconcilation) example becomes far simpler (perhaps using JSON-LD Framing) when JSON-LD is used.

OR13:
> We've had this backwards. Developers always process the core data model as JSON-LD, they may just be using the more limited approach defined in the JSON Processing section. JSON-LD is a syntax and you can access it directly.

This is great, I think the "JSON Processing" section is confusing and should be removed.

> The content we have in the spec today that explains how to do these optimizations in JSON parsing land is unnecessary when you use a JSON-LD library to convert from one context to another or to RDF, but not doing that does not mean you're not doing JSON-LD processing.

I'd say its not necessary regardless of which libraries you use.... and the fact that you can be "doing JSON-LD processing" when "not realizing it" is a security and readability issue in the current spec, that needs to be addressed.

> Additionally, these more limited JSON parsers do not do the following (at least not without nearly recreating JSON-LD processing...):
> 
> * handle term consistency--they always assume the key names mean the same thing unless they're within very narrowly defined media type namespaces
> * they do not have constructs for advanced data types
> * nor language value objects
> * nor a consistent extensibility mechanism (other than one media type per document use)
> * nor underlying graph model which enables aggregation, faceted querying, statement signing/extraction/reuse

^ maybe it would be better to define "JSON-LD Processing first", and then define "limited / unsafe JSON processing second".

OR13:
@BigBlueHat I suggest you pull anything useful that is removed in https://github.com/w3c/vc-data-model/pull/1298 into this PR, and say in plain english essentially what you said in https://github.com/w3c/vc-data-model/pull/1270#issuecomment-1745483763

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-10-04-vcwg#section1-1) on 2023-10-04

- no resolutions were taken

#### 1.1. Explain JSON-LD further. (pr vc-data-model#1270)

_See github pull request [vc-data-model#1270](https://github.com/w3c/vc-data-model/pull/1270)._




**Manu Sporny:** so, 1270 has multiple requests for changes.  
… selfissued is blocking. Orie is blocking.  
… what changes need to be made?  

**Orie Steele:** we have this section about JSON processing...which I don't really like.  
… I'd believed we would/should add a JSON-LD section to explain applying `@context` more completely.  
… and these two things differ substantially when you just process JSON.  
… I think the text in the JSON processing section is dangerous and misleading.  
… that text is mostly information that doesn't help implementers.  

> *Dave Longley:* -1 to what Orie is saying :) ... the information is the same, JSON-LD is a syntax that expresses graph information.

**Orie Steele:** and doesn't help people use RDF or JSON-LD.  
… so the changes I want to see are to remove these differences.  
… bigbluehat's comments are excellent.  
… he says there dangers of doing just JSON processing.  
… it's a huge mistake for us to be vague.  
… and this will harm interop if these things differ too much.  
… the JSON-LD processing section needs to talk more about RDF data model.  
… `@id`, `@list`, etc.  
… basically the stuff that's in the `@context` that only an advanced processor would use.  
… I strongly disagree with the position that dlongley stated.  
… that we should avoid the JSON-LD feature to keep it compatible with JSON processing.  
… bigbluehat's PR has some of this, but not enough which is why I'm requesting changes.  

> *Kristina Yasuda:* +1 to json processing is not compatible with JSON-LD processing (without additional checks on top co json processing).

**Michael Jones:** this PR kind of runs at the mouth.  
… it says a lot of things that aren't useful to implementers.  
… we don't need to be an evangelist mouth piece for JSON-LD.  
… I will not approve this until it's down to a sentence or a paragraph.  

> *Orie Steele:* I would encourage better direct references to the JSON-LD spec, for example: [https://www.w3.org/TR/json-ld11/#relationship-to-rdf](https://www.w3.org/TR/json-ld11/#relationship-to-rdf).

**Dave Longley:** trying to figure out where to start.  
… there are many competing requests.  
… many from Orie requesting explanation of the advantages of JSON-LD.  

> *Kristina Yasuda:* chairs hat on, i don't think the request was "json-ld advantage" it was about "json-ld processing guide".

> *Orie Steele:* I disagree that the PR captures the advantages on JSON-LD, but I agree that is an objective... and I think citing the JSON-LD spec, would better accomplish the mission.

**Dave Longley:** new comers are often confused by the ability to process JSON-LD further with libraries.  
… but they miss that JSON-LD is a syntax.  
… if you understand the `@context`, then you can move ahead.  
… if you don't understand what's in `@context, then you do need to process that and understand it.  
… so this section is really talking about that difference.  
… JSON is really just a syntax.  
… and if you want to do anything with JSON, you need a spec to read and implement.  
… whereas JSON-LD is an object-oriented graph model that defines how to be processed.  
… JSON-LD imposes additional constraints.  
… so you don't have this problem of missing out on this extensibility and data model.  

> *Orie Steele:* I agree with what dave is saying, that JSON-LD and JSON claims can't be merged, without understanding RDF... that is the security issue that the group is being unclear regarding.

**Dave Longley:** the JSON processing model section basically says that if you know it's using these `@context` definitions that the spec specifies, then you can just process this in JSON--because you trust it implemented the spec.  
… on top of that Orie has been asking on elaborating on the additional value of JSON-LD.  

> *Orie Steele:* We don't have the section, we have draft text, which I don't feel is acceptable yet... the mission is still ongoing.

**Dave Longley:** but now we have the counter request to remove all of this--both JSON-LD values that were requested and the JSON processing section.  
… many of the new comers have missed the history here.  
… that this context ability was chosen on purpose.  
… that it provides the clarity of what the data is about referenced from that data.  

> *Orie Steele:* Agree with what dave is saying about JSON-LD providing value to W3C Verifiable Credentials, hence the need to communicate the value of JSON-LD and RDF processing as opposed to "JSON Processing", which does not require reading the JSON-LD spec, to do safely.

> *Orie Steele:* +1 to landing the PR, after revising it.

**Dave Longley:** it's valuable to explain this value and how it helps and has helped this community since the beginning.  

**Kristina Yasuda:** I don't think we don't want to have this section.  
… it feels like there's more discussion needed for this PR.  
… it's not like this needs huge conversation.  
… so lets entertain the queue.  
… but keep it sort and sweet to exactly what you want to see here.  
… not debate the values of JSON-LD etc.  

**Ted Thibodeau Jr.:** I started off being extremely frustrated by two voices in this conversation.  
… both are asking for vague changes with no specific changes.  

> *Orie Steele:* If the working group agrees, I am happy to author an alternative PR.

**Ted Thibodeau Jr.:** I kinda want to lock Orie and selfissued in a room and let you hash it out.  
… but I don't think that's going to be useful.  
… and others should support you if they agree.  
… Orie's been asking for this for months.  
… but without any actual contribution to the conversation.  

> *Orie Steele:* IMO its better to refer to a spec, than it is to mischaracterize the benefits of a spec, without references.

**Ted Thibodeau Jr.:** he hasn't put the concrete suggestion forth.  
… and selfissued if you want this down to a sentence, then write that sentence.  
… otherwise you're just blocking to block.  

**Kristina Yasuda:** this is an inclusive environment.  

**Ted Thibodeau Jr.:** you're spending more time on this than anyone.  
… let's move onto the queue.  

**Kristina Yasuda:** please leave the call TallTed.  

**Brent Zundel:** ivan please remove TallTed from the call.  

> *Orie Steele:* My offer to author an alternative is sincere, we may find that merging 2 different PRs is an effective path forward.

> *Manu Sporny:* yes, please write something Orie ^.

> *Orie Steele:* will do.

**Michael Jones:** strangely related to what TallTed was saying.  
… I do have a specific action.  
… put this in a note, and reference the note.  

**Ivan Herman:** clarification. you mean a separate WG note document, correct?  

**Michael Jones:** yes.  

**Kristina Yasuda:** there is a separate JSON-LD spec...so I'm not sure a separate note gets us much.  
… please offer concrete PRs with concrete suggestions Orie.  
… and we can move from there.  
… Orie is that OK?  

> *Orie Steele:* IMO, its worth highlighting the parts of the JSON-LD spec that can provide the most value to Verifiable Credentials.

> *Orie Steele:* yes, i will do a PR.

> *Dave Longley:* +1 to see something from Orie ... but also note, Orie, I think there may be more benefits to VCs than you're thinking of (including all of the ones in the PR of that subtopic).

</details>

decentralgabe (approved):

iherman:
This comment is generally on the text, regardless of where it ends up (whether in a separate note or in the VCDM spec).

The text uses the analogy/reference to object-oriented modeling as a paradigm "familiar to developers". This familiarity is what worries me, because, usually, object-oriented terminology refers to the "class" feature in languages like C++, Java, Python, TypeScript and, more recently, JavaScript. And this analogy is _wrong_ and misleading when it comes to RDF. RDF uses the term "class", but that is in the sense of ontologies. 

In, say, Java one _assigns_ the type of an "object" as a declaration, and that is bound to the object for life. Object declaration is (also) a constraint for the properties (attributes) that one can use in conjunction with that object. The type of an RDF resource can also be assigned at "declaration" (that is what `rdf:type` does) but, more importantly, it can also be _inferred_ from statements like property domains and ranges (if the RDF environment uses some inference engine for, e.g., RDFS processing). Hence, the domains and ranges in the vocabulary are _not_ constraints, they are _licenses to infer_. This is in line with the "Ontology" nature of RDFS (and OWL), which is used to categorize knowledge in classes.

As an example, if my dataset contains these two triples:

```
ex:R ex:p "something" ;
ex:p rdfs:domain ex:Cl;
```

then I can _infer_ the following statement:

```
ex:R rdf:type ex:Cl;
```


The misunderstanding on the nature of domains and ranges has come up several times in this Working Group as well, and this is mostly because people often believe that RDF's classes are analogous to classes in object-oriented languages. We should _not_ reinforce this misunderstanding. _I would therefore propose to remove all references to object-oriented modeling from the text in the PR._

cc @pchampin

OR13:
In case you missed this, https://github.com/w3ctag/design-principles/issues/239#issuecomment-1748162752

Very relevant to the guidance we are attempting on this PR, and on https://github.com/w3c/vc-data-model/pull/1302

dlongley:
@iherman,

The object-oriented references in the text above pertain to how statements about subjects are expressed in JSON (as JSON-LD). They are not about the RDF as much as your comment suggests, but the modeling *does* enable consistent processing to enable mapping to / from RDF much more simply.

What is trying to be said here, is that, in JSON, there are "objects", which have "keys" (aka "properties") and "values" associated with those keys. Someone designing a JSON document, could, if they wanted (and clearly some often do!), decouple their data model from these primitives, using them any which way.

For example, if we wanted to model a "car", which has an engine, and a number of wheels, we could express this information in JSON any number of ways. For example:

```json
{
  "thing": "car1",
  "engine": true,
  "wheels": true,
  "car1Props": [["color", "red"]],
  "car1Wheels": 4,
  "car1Engine": ["electric"]
}
```

There's a lot of JSON in the wild that evolves organically and ends up looking like that. Applications that consume it might have bespoke rules that have also evolved over time... "if you see 'thing' and it's value starts with 'car', then it's a car, and its properties will be in the rest of that value + 'Wheels' or + 'Engine', and you'll know if those are there if 'engine' or 'wheels' are true, and if the car has other properties it will be under + 'Props' as nested arrays, etc.".

An alternative expression might be:

```json
{
  "id": "car1",
  "type": "Car",
  "color": "red",
  "wheels": 4,
  "engine": {
    "type": "ElectricEngine"
  }
}
```

And the rules for processing this data can be more consistent. In fact, you can write consistent processing rules for most things modeled this way because the data model concepts map better to the primitives provided by JSON. It also adds compositional capabilities to your data. JSON-LD requires JSON to be interpreted in this way, such that authors must model their data accordingly as consumers can and will presume this basic modeling is built-in.

This is an important concept to express as an advantage of using JSON-LD (over "vanilla JSON" where anything goes) as it eliminates (or greatly reduces) situations like the first example, where complex bespoke processing rules must be written into applications to even figure out a basic interpretation of the information one is expressing. Requiring "JSON objects" to actually express "objects" in the information model and JSON keys/properties to actually express *the properties of those objects* is "object oriented" and should be highlighted as a benefit.

iherman:
@dlongley I do not question those advantages of JSON or of JSON-LD. But "object-oriented", to me, resonates primarily to the programming concepts in object-oriented programming languages. That, plus the obvious relationship to RDF, does bother me.

I won't lie down the road over this, obviously; if this works for everyone else, be it. I just wanted to express my discomfort. Maybe the reason that I never look at JSON-LD in isolation but only through the goggles of RDF... 

(And no, I do not have a good alternative term in my mind right now.)

msporny:
Proposal from the WG call: We should move this text to the VC Implementation Guide in order to "move it forward"... no objections on the call.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-10-17-vcwg-special#section2-1) on 2023-10-17

- no resolutions were taken

#### 2.1. Explain JSON-LD further. (pr vc-data-model#1270)

_See github pull request [vc-data-model#1270](https://github.com/w3c/vc-data-model/pull/1270)._




**Brent Zundel:** The chairs feel that this is prime example of a PR that is unlikely to achieve consensus.  

**Manu Sporny:** +1 to this one not making it in. It has text that some in the group feel strongly about, moving it to the VC implementation guide and linking it is something Mike Jones did not object to.  

**Brent Zundel:** Anyone want to object to that proposal?  
… I will mark this pending close and say it will be done in the implementation guide rather than the core data model.  

</details>

brentzundel:
I have raised https://github.com/w3c/vc-imp-guide/issues/71 to track adding this text to the implementation guide.

msporny:
Pending close has been marked on this PR for 2 weeks, the issue is being tracked in the vc-imp-guide repo. Closing.

Issue #1266, opened 2023-09-02
Recommend that authors don't use `@base` in `@context`


msporny:
Based on a request from @iherman in https://github.com/w3c/vc-data-integrity/pull/185#issuecomment-1694682168, he suggests that we discourage the use of `@base` in `@context` to ensure better alignment when not doing full JSON-LD processing.

msporny:
PR #1271 has been raised to address this issue. This issue will be closed once PR #1271 is merged.

msporny:
PR #1271 is merged, closing.

Issue #1264, opened 2023-08-29
Recommend that authors don't use `@language` in `@context`


msporny:
In order to maximize interoperability when doing non-JSON-LD processing, we should suggest that authors don't set `@language` in `@context` (which would set the default language for /all text fields/ in the VC). This would be problematic as some text fields in a VC carry base-encoded information and shouldn't have a language applied to it. Furthermore, processors that don't do JSON-LD Processing would be unaware of the default language being set in `@context`.

aphillips:
As noted in #1252, I think a lot of specifications just say that "only fields `name` and `description` contain natural language text and thus `@language` and `@direction` only apply to these fields". Specifying `@language` in the `@context` saves having to repeat the language metadata for every `name` and `description` when the file is in a single language (and makes the file a bit self-documenting regarding the intended language audience).

But... if this is not an appropriate mechanism to use, better to say that in your I18N Considerations.

msporny:
> As noted in #1252, I think a lot of specifications just say that "only fields `name` and `description` contain natural language text and thus `@language` and `@direction` only apply to these fields". Specifying `@language` in the `@context` saves having to repeat the language metadata for every `name` and `description` when the file is in a single language (and makes the file a bit self-documenting regarding the intended language audience).

Yes, in general using `@language` to set the default language for a general JSON-LD document is expected to be fine in many use cases. However...

> But... if this is not an appropriate mechanism to use, better to say that in your I18N Considerations.

The issue is that we're dealing w/ a subset of JSON-LD in Verifiable Credentials because there is a subset of the group that doesn't want to do full JSON-LD processing, but rather a much simpler version of it:

https://w3c.github.io/vc-data-model/#json-processing

... thus, we're not at liberty to use JSON-LD features that appear in the `@context` as much as we'd like. In addition to that, VCs tend to carry base-encoded information, and because of that, using a blanket `@language` declaration would end up adding language to base-encoded strings that shouldn't have that language attached to it.

One option here is to alias `@value`, `@language`, and `@direction` globally, but then we might be invoking something we don't want to invoke. I'll have to do more analysis on that to see if it's even feasible (I expect that it's not).

aphillips:
First off, thank you so much for putting the effort into #1252 and taking our comments seriously. I really appreciate it.

[...]

Issue #1155, opened 2023-06-15
Internationalization Review for VCDM 2.0


awoie:
The following is an Internationalization Review for "W3C Verifiable Credentials Data Model v2.0" (VCDM 2.0). The latest published version of the specification can be found [here](https://www.w3.org/TR/vc-data-model-2.0/).

The specification is a JSON-LD data model specification for Verifiable Credentials and Verifiable Presentations.

All features can be internationalized using JSON-LD features. Specifically, the validity period for a Verifiable Credential is expressed using XML Schema 1.1 where the dates can be localized and made accessible given the nature of XML Schema 1.1 date time values.

The specification contains an [Internationalization Considerations](https://w3c.github.io/vc-data-model/#internationalization-considerations) section that provides more details on how internationlization is achieved.

The following is a review based on the short i18n review checklist from [here](https://www.w3.org/International/i18n-drafts/techniques/shortchecklist):

- If the spec or its implementation contains any natural language text that will be read by a human (this includes error messages or other UI text, JSON strings, etc, etc), then ensure that there’s metadata about and support for basic things such as language and text direction. Also check the detailed guidance for [Language](https://www.w3.org/TR/international-specs/#resource) and [Text direction](https://www.w3.org/TR/international-specs/#text_direction).

    - [x] applicable
    - [ ] not applicable

The specification is a JSON-LD data model specification and can use all internationalization features that JSON-LD offers. The [Internationalization Consideration](https://w3c.github.io/vc-data-model/#internationalization-considerations) section specifically points out how to achieve supporting different languages, text direction and so on. 

---
- If the spec or its implementation allows content authors to produce typographically appealing text, either in its own right, or in association with graphics, then take into account the different typographic styles used around the world (for things such as line-breaking, text justification, emphasis or other text decorations, text selection and units, etc.). Also check the detailed guidance for [Typographic](https://www.w3.org/TR/international-specs/#typography) support.

    - [ ] applicable
    - [x] not applicable

---
- If the spec or its implementation allows the user to point into text, creates text fragments, concatenates text, allows the user to select or step through text (using a cursor or other methods), etc., then make allowances for the ways different scripts handle units of text. Also check the detailed guidance on [Text-processing](https://www.w3.org/TR/international-specs/#operations).

    - [ ] applicable
    - [x] not applicable

---
[...]

PR #1271, opened 2023-09-10
Warn authors that usage of global JSON-LD keywords might harm interoperability

msporny:
This PR is an attempt to address issues #1264 and #1266. It provides guidance that document authors shouldn't use JSON-LD keywords that affect document processing inside the `@context` array because it can impact how processors that don't do full JSON-LD processing interpret the document, leading to a degradation in interoperability.

A separate PR will be raised for guidance on how to use the language features to increase interoperability. This PR just says "don't use the `@language` sledgehammer in `@context`".

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1271.html" title="Last updated on Nov 14, 2023, 6:45 PM UTC (5c322b1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1271/922dce6...5c322b1.html" title="Last updated on Nov 14, 2023, 6:45 PM UTC (5c322b1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3989,7 +3989,7 @@ base media type, `application/vc+ld+json`.
 
         <p>
 As elaborated upon in Section <a href="#json-processing"></a>, some software
-applications might not perform full JSON-LD processing.
+applications might not perform dynamic JSON-LD processing.
 Authors of <a>conforming documents</a> are advised that using JSON-LD keywords
 in the `@context` value to globally affect values in documents, such as
 globally setting the keywords `@base` or `@language`, might reduce
--- a/index.html
+++ b/index.html
@@ -3990,9 +3990,9 @@ base media type, `application/vc+ld+json`.
         <p>
 As elaborated upon in Section <a href="#json-processing"></a>, some software
 applications might not perform full JSON-LD processing.
-<a>Conforming document</a> authors are urged to not use JSON-LD keywords that
-affect document processing in the `@context` value, such as globally setting
-`@base` and `@language`, in order to increase interoperability.
+To increase interoperability, authors of <a>conforming document</a> are urged
+to not use JSON-LD keywords in the `@context` value to globally affect values
+in documents, such as globally setting the keywords `@base` or `@language`.
         </p>
 
         <section>
--- a/index.html
+++ b/index.html
@@ -3993,8 +3993,7 @@ applications might not perform full JSON-LD processing.
 Authors of <a>conforming documents</a> are advised that using JSON-LD keywords
 in the `@context` value to globally affect values in documents, such as
 globally setting the keywords `@base` or `@language`, might reduce
-interoperability with certain processors that do not perform full JSON-LD
-processing.
+interoperability.
         </p>
 
         <section>
--- a/index.html
+++ b/index.html
@@ -3988,7 +3988,7 @@ base media type, `application/vc+ld+json`.
         </p>
 
         <p>
-As elaborated upon in Section <a href="#json-processing"></a>, some software
+As elaborated upon in Section <a href="#static-processing"></a>, some software
 applications might not perform dynamic JSON-LD processing.
 Authors of <a>conforming documents</a> are advised that using JSON-LD keywords
 in the `@context` value to globally affect values in documents, such as
--- a/index.html
+++ b/index.html
@@ -3990,10 +3990,13 @@ base media type, `application/vc+ld+json`.
         <p>
 As elaborated upon in Section <a href="#static-processing"></a>, some software
 applications might not perform dynamic JSON-LD processing.
-Authors of <a>conforming documents</a> are advised that using JSON-LD keywords
-in the `@context` value to globally affect values in documents, such as
-globally setting the keywords `@base` or `@language`, might reduce
-interoperability.
+Authors of <a>conforming documents</a> are advised that interoperability
+might be reduced if JSON-LD keywords in the `@context` value are used to
+globally affect values in `@context` or other documents, such as by
+globally setting the keywords `@base` or `@language`. Using these
+keywords with specific focus, as when aligned with one or more values
+of one or more other keywords, may broaden the utility of specific
+`@context` files and/or <a>conforming documents</a>.
         </p>
 
         <section>
--- a/index.html
+++ b/index.html
@@ -3987,6 +3987,14 @@ form</a> MUST be utilized for all representations of the data model in the
 base media type, `application/vc+ld+json`.
         </p>
 
+        <p>
+As elaborated upon in Section <a href="#json-processing"></a>, some software
+applications might not perform full JSON-LD processing.
+<a>Conforming document</a> authors are urged to not use JSON-LD keywords that
+affect document processing in the `@context` value, such as globally setting
+`@base` and `@language`, in order to increase interoperability.
+        </p>
+
         <section>
           <h3>Syntactic Sugar</h3>
 
--- a/index.html
+++ b/index.html
@@ -3990,9 +3990,11 @@ base media type, `application/vc+ld+json`.
         <p>
 As elaborated upon in Section <a href="#json-processing"></a>, some software
 applications might not perform full JSON-LD processing.
-To increase interoperability, authors of <a>conforming document</a> are urged
-to not use JSON-LD keywords in the `@context` value to globally affect values
-in documents, such as globally setting the keywords `@base` or `@language`.
+Authors of <a>conforming documents</a> are advised that using JSON-LD keywords
+in the `@context` value to globally affect values in documents, such as
+globally setting the keywords `@base` or `@language`, might reduce
+interoperability with certain processors that do not perform full JSON-LD
+processing.
         </p>
 
         <section>
--- a/index.html
+++ b/index.html
@@ -3966,7 +3966,8 @@ specification. Instances of the data model are encoded in JSON-LD compact
 form [[!JSON-LD]] and include the <code>@context</code> <a>property</a>. The
 <a href="https://www.w3.org/TR/json-ld/#the-context">JSON-LD context</a>
 is described in detail in the [[!JSON-LD]] specification and its use is
-elaborated on in Section <a href="#extensibility"></a>.
+elaborated on in Section <a href="#contexts"></a> and
+Section <a href="#extensibility"></a>.
         </p>
 
         <p>
@@ -3988,15 +3989,17 @@ base media type, `application/vc+ld+json`.
         </p>
 
         <p>
-As elaborated upon in Section <a href="#static-processing"></a>, some software
-applications might not perform dynamic JSON-LD processing.
+As elaborated upon in Section
+<a href="#credential-type-specific-processing"></a>, some software
+applications might not perform generalized JSON-LD processing.
 Authors of <a>conforming documents</a> are advised that interoperability
 might be reduced if JSON-LD keywords in the `@context` value are used to
-globally affect values in `@context` or other documents, such as by
-globally setting the keywords `@base` or `@language`. Using these
-keywords with specific focus, as when aligned with one or more values
-of one or more other keywords, may broaden the utility of specific
-`@context` files and/or <a>conforming documents</a>.
+globally affect values in a <a>verifiable credential</a> or
+<a>verifiable presentation</a>, such as by globally setting the `@base` keyword.
+For example, globally setting these values might trigger a failure in a
+mis-implemented JSON Schema check on the `@context` value in an implementation
+that is performing credential-type-specific processing and not expecting the
+`@base` value to be expressed in the `@context` value.
         </p>
 
         <section>


iherman (commented):
I am fine with the text concerning `@base`. But #1264 is still open regarding `@language`... depending on what we decide there, we have to be careful not to disallow the standalone `@language` setting:


```json
"@context": [
    "https://...somecontextfile.jsonld",
   {"@language":"en"}
],
```

which is, according to the new text, not advised.

brentzundel (approved):

dlongley (approved):

index.html
```diff
@@ -3987,6 +3987,14 @@ <h3>JSON-LD</h3>
 base media type, `application/vc+ld+json`.
         </p>
 
+        <p>
+As elaborated upon in Section <a href="#json-processing"></a>, some software
+applications might not perform full JSON-LD processing.
+<a>Conforming document</a> authors are urged to not use JSON-LD keywords that
+affect document processing in the `@context` value, such as globally setting
+`@base` and `@language`, in order to increase interoperability.
```
BigBlueHat, line 3995: The key constraint is that the handling of `@base` and `@language` are not (currently) part of the JSON Processing model, and so would result in a deviation in result (as with `@base`) or a "richer" dataset (from inclusion of `@language`) after JSON-LD Processing which would make the two approaches inequivalent.

So, how about this text instead:
> To support consistency across both JSON Processing and
JSON-LD Processing models, the `@base` and `@language` keywords should not
be used within context files nor as inline values within the `@context` array.

BigBlueHat (changes requested):
The issue seems to be that the JSON Processing model doesn't take these keywords into account, so JSON-LD authors should avoid assuming they'll be consistently processed within he wider ecosystem and, therefore, should avoid them.

index.html
dlongley: If we tweak the language here we need it to focus on global settings only -- individual context term mapping definitions in extending context files should be able to use `@language` without issue.

OR13: I'm not sure it makes sense to sacrifice JSON-LD processing benefits for JSON processors that don't care about JSON-LD.... That feels like watering down the value of JSON-LD, for preserving a value that is not recognized by JSON Processors.

I'd recommend not anchoring the JSON-LD value to the JSON processing experience.

The opposite in fact, JSON-LD processing should have a set of features that are greater than JSON Processors can offer, and we should lean into recommending them... when they make sense.

decentralgabe (approved):

seabass-labrax (approved):
The comments regarding presently open debates notwithstanding, I am in favour of this pull request both individually and in the wider context.

aphillips:
I think I might object to this proposed text, which recommends against the use of a language-support construct used by JSON-LD (and maybe ostensibly for the benefit of non-LD processors that do not and are not expected to understand it). If there are issues with `@language` being wrongly attributed to non-language fields by LD processors, we should figure out how to resolve those rather than removing or requiring the removal of something that LD-aware producers might be emitting in a standardized way.

seabass-labrax:
@aphillips, this PR doesn't preclude us from allowing `@language` in the context later depending on the resolution of #1264.

If #1264 is resolved in a way that makes `@language` in the context an option, in order to support tooling which is not aware of JSON-LD, `@language` would need to be explicitly described in the VC specification. That may require a change to the text that this PR adds, but I wouldn't say that merging of this PR constitutes a decision on the language aspect - it only describes the general principle of not changing the global context where possible.

aphillips:
@seabass-labrax Thanks!

> it only describes the general principle of not changing the global context where possible.

I read:

> ...authors are urged to not use JSON-LD keywords...

... as "do not use" guidance (presumably there is no pre-existing global context to not change 😉 when generating the credential). I appreciate that the feature has unintentional consequences in VC and thus warrants such a warning, but I would normally expect that I18N would recommend such a feature to be used or allowed.

Anyway, I withdraw any objection. Let's work the language issues elsewhere.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-09-14-vcwg#section1-5) on 2023-09-14

- no resolutions were taken

#### 1.5. Add guidance to not use JSON-LD keywords in @context value. (pr vc-data-model#1271)

_See github pull request [vc-data-model#1271](https://github.com/w3c/vc-data-model/pull/1271)._




**Brent Zundel:** Raised 4 days ago. Within window of review.  

> *jose-gataca:* jose-gataca has joined #vcwg.

**Brent Zundel:** Some approvals already, one change request.  

**Sebastian Crane:** Issue 1264, which this PR quotes, doesn't need to be merged for this to be merged.  

**Manu Sporny:** BigBlueHat just wants some rewording. Can be done and merged after that.  

> *Ted Thibodeau Jr.:* just to note -- I have not reviewed this PR yet.

**Brent Zundel:** PR is on course to be merged soon.  

</details>

index.html
msporny: @BigBlueHat not having a concrete change suggestion here is making things difficult to determine if we have consensus on this PR. I don't know if @OR13 is objecting to the original language, your language, or the PR (in general).

At this point, I need to understand if either of your input is blocking or not. Also note that the outcome of https://github.com/w3c/vc-data-model/issues/1264#issuecomment-1712807665 will probably drive more changes in the spec (beyond this PR).

dlongley, line 3995:
```suggestion
<a>Conforming document</a> authors are urged to not use JSON-LD keywords that
globally affect document processing in the `@context` value, such as setting the
global keywords `@base` or `@language`, in order to increase interoperability.
```

BigBlueHat: @msporny sorry for any confusion. @dlongley's suggestions below (https://github.com/w3c/vc-data-model/pull/1271/files#r1337759360) are great. 👍 to using those instead of my earlier suggestions.

TallTed:
```suggestion
To increase interoperability, authors of <a>conforming document</a> are urged
to not use JSON-LD keywords in the `@context` value to globally affect values
in documents, such as globally setting the keywords `@base` or `@language`.
```

brianorwhatever:
It's not clear to me why `@base` can't be used. I don't see how it would effect interoperability at the VC layer. It seems like a JSON processors and a JSON-LD processor would have the same data in the VC but the LD processor may be able to extend that data. This PR limits functionality without any significant benefit for interop.. although I might be missing something.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-10-03-vcwg-special#section1) on 2023-10-03

- no resolutions were taken

### 1. Internationalization WG review.

**Kristina Yasuda:** Special meeting due to feedback on Internationalization.  
… Existing options had not been decided on.  

**Kristina Yasuda:** Please introduce yourself Addison.  

_See github issue [vc-data-model#1155](https://github.com/w3c/vc-data-model/issues/1155)._




_See github issue [vc-data-model#1264](https://github.com/w3c/vc-data-model/issues/1264)._




_See github pull request [vc-data-model#1271](https://github.com/w3c/vc-data-model/pull/1271)._




**Addison Phillips:** I'm the chair of the I18N group at the W3C.  

> *Kristina Yasuda:* five options? [https://github.com/w3c/vc-data-model/issues/1264#issuecomment-1712807665](https://github.com/w3c/vc-data-model/issues/1264#issuecomment-1712807665).

**Manu Sporny:** The background of this: we've had guidance about supporting internationalization, with a design pattern for people to follow. In the 1.0 and 1.1 work, we haven't seen much adoption of the I18N features.  
… For 2.0, we are adding two fields expected to be multilingual.  

> *Manu Sporny:* Here are the potential options that we're considering: [https://github.com/w3c/vc-data-model/issues/1264#issuecomment-1712807665](https://github.com/w3c/vc-data-model/issues/1264#issuecomment-1712807665).

**Manu Sporny:** We had a number of options to consider for how to do it in 2.0.  
… We're looking to get to consensus on the option that we should choose, and one that will satisfy the I18N group.  

**Addison Phillips:** I had read through the summary of the discussion and will summarise here to ensure that we have a common understanding.  
… In general, the I18N Group would like to see that any natural language string field has metadata about at least A: language and B: text direction.  
… We aren't prescriptive about how that is performed. We also like to see a default language for documents.  

> *Orie Steele:* It would be good to get a stronger opinionated recommendation regarding approaches.

> *Kristina Yasuda:* we kind of have 1.90 foot in LD world.

> *Dave Longley:* I think we have both feet in LD world, we just want to use the simplest on-ramps.

> *Dave Longley:* and adding `@value`, `@language`, and `@direction` in fields locally is the easiest way to do that.

**Addison Phillips:** I think it sounds like being somewhat in the Linked Data world as well as a more general specification produces some complications.  
… We would like to understand more the concerns around global @language directives, because we are wondering whether these concerns apply to the wider LD community.  
… From what we've learnt so far, the I18N group is trying to produce best practice recommendations to other groups.  
… One of the ones that we've already been working on is quite different from your approach. I would like to share it with you.  

**Manu Sporny:** On the topic of being both in the LD world and not, it seems like a subset of our community are less likely to adopt the specification when LD features are added.  
… We've tried to reduce the LD features to a minimum up until now.  

> *Orie Steele:* -1 to asserting that "avoiding using LD is a possibility at this point".

> *Dave Longley:* -1 that we can / are avoiding it, +1 that we're choosing the easiest on-ramps.

> *Orie Steele:* -1 to being vague about understanding conforming documents (which are JSON-LD in compact form).

**Manu Sporny:** As for Option E (using a translation file), I think you mentioned that you were against it, and there seems to be agreement within this WG.  

**Addison Phillips:** I would not object to translation files per se, but I would point out technical complications about multiple requests and resources. I think that doesn't sound like the right pattern for credentials.  

**Manu Sporny:** Option E can eliminated then!  
… For option D, I don't think this option has any advantages over using the LD method of a global @language, which is effectively the same in effect.  

**Addison Phillips:** It is common for us to recommend that specifications do this. It would be better if there were generic mechanisms, but specification-specific fields are OK.  

> *Dave Longley:* -1 to option D.

> *Dave Longley:* -1 to option E.

> *Shigeya Suzuki:* For the record: option E is externalization, and it will not be possible unless we define internal way to express it IMO.

> *Dave Longley:* +1 to eliminate E.

**Kristina Yasuda:** Are there any objections to eliminating option E?  

> *Andres Uribe:* +1 to eliminate E.

> *Manu Sporny:* +1 to eliminate option E.

> *Shigeya Suzuki:* I'm fine with eliminating option E for now..

> *Phillip Long:* -1 to option E & D.

> *Joe Andrieu:* +1 to eliminate E.

> *Manu Sporny:* +1 to eliminate option D (but keep it around as a backup plan).

**Addison Phillips:** I would suggest that you could keep it for a 'backup'.  

**Andres Uribe:** I think that's the default if we can't get consensus of anything else.  

**Sebastian Crane:** I wanted to mention option E with the translation files, sometimes they look really good on paper, but in practice, lots of complications.  
… networked translations, even when installed on computer, there are still lots of issues, GNU style translation -- .pot files -- translates based on literal value of string, but as linguists say, there are cases where you can have same words which mean two semantically different things and language files as used in GNU world don't have opportunity to disambiguate those. Number of complicates here with Option E.  

> *Shigeya Suzuki:* I don't want to spend time time on this, but the way gettext/po used is studied well and in some non-english area esp. in CJK area, it's useful.

> *Sebastian Crane:* I agree with addison: they can be used correctly but I think that is unlikely to be the case in the VC world.

**Ivan Herman:** Option D means having two properties: language and text direction. We need both on the default level.  
… Is this the general view as well?  

> *Manu Sporny:* We do need to express language AND direction.

> *Manu Sporny:* yes, that's the general agreement, I believe.

**Addison Phillips:** Indeed, I agree we would need to have this.  

> *Manu Sporny:* Here is option C:.

```
    "credentialSubject": {
        "myHumanReadableProperty": [{
            "@value": "This is some human-readable text.",
            "@language": "en"
        }, {
            "@value": "هذا بعض النص الذي يمكن قراءته بواسطة الإنسان.",
            "@language": "ar",
            "@direction": "rtl"
        }]
```

> *Dave Longley:* +1 to this option (C, I believe), i think it's the simplest and will work generally for any natural language field.

**Manu Sporny:** We would express the value of the string, the language, and the text direction. I believe this meets the requirement that addison illustrated.  

> *Dmitri Zagidulin:* +1 to option C, works well for multi-language credentials in Edu land.

> *Phillip Long:* +1 to Option C, as it does indeed work well in edu-land.

**Manu Sporny:** We've mainly been discussing whether to alias the "@X" terms. Sebastian proposed options C on two occasions, and there were no objections raised. I believe it addresses all concerns except for the ability to specify a global default.  

> *Sebastian Crane:* +1 to Option C obviously :).

**Manu Sporny:** It's not easy to test this, as multiple languages are optional.  

> *Manu Sporny:* +1 to speaking to Option C in the specification.

**Addison Phillips:** I'm concerned that whilst the 'SHOULD' and 'MAY' are good supports for internationalisation, there will still be completely unlabeled strings.  
… I would like it to be possible to know a default, for when people don't want to put all the extra syntax in.  

**Manu Sporny:** Can we ask if the group is OK exploring option C further?  

**Kristina Yasuda:** If anyone is strongly in objection to option C, please speak now.  

**Sebastian Crane:** To repeat for addisons' benefit -- VCs don't inherently have a language... language on field such as name/description is for human holder of VC in a wallet application. When you have the RDF world, link things together based on ontological truth, actual meaning...  
… you don't necessarily want to apply a language to the specific credential, you want to apply language to description of credential... that's why I like option C -- translate those human-readable values, credential itself doesn't have a language.  

> *Orie Steele:* Conforming documents are represented in JSON-LD.... the philosophical concept of credentials is not helpful... JSON-LD will have text that is in a human readable language (both the term definitions, the text behind them, and their literal values).

**Addison Phillips:** Yes, important observation, locale-neutral data... when people talk about these things, name/description is how humans interact... can't look at other things and talk meaningfully about them... credential has BS of science -- those names/descriptions are of natural language pieces... want natural language to be associated with those parts, not other data.  
… challenge is that machine generates these things, people writing code may or may not be willing to generate multiple language versions, or they may not wish to obtain and serialize information on per-field basis... if you're willing to say MUST, then we're good.  
… I think that's an important distinction: one wants to have language-neutral data if at all possible. A complication is that humans can't talk meaningfully about the pure data, only about the natural language descriptions.  

**David Chadwick:** I think MUST is fine, but not sufficient. Let's say you have a degree from a Japanese university and has language metadata, that degree credential is still not readable by a typical English person.  
… I believe that C is necessary, but doesn't completely solve the internationalisation concerns.  

**Andres Uribe:** I'm definitely supportive of Option C. In addition, I would like to see aliasing. I don't really understand why aliasing will cause problems with JSON-LD, so I would appreciate an explanation here.  

> *Dave Longley:* aliasing `@value` will alias it for everything, not just language values.

**Manu Sporny:** The short answer is that `@value` is also used for non-natural-language fields. We can't just aliases it globally without making other fields have unwanted language features.  

> *Dave Longley:* so making it say `langValue` (or whatever) for non-language values will be weird / confusing.

> *Orie Steele:* The comment about "re-compacting" / "compacting" is critical for the WG to understand.

**Manu Sporny:** For that reason, I would be strongly opposed to aliasing if option C can be sufficient. It is only a single character difference.  

> *Orie Steele:* I'm not sure that there is understanding here... and we should clarify.

**Andres Uribe:** Thank you, that answered my question.  

**Manu Sporny:** We would end up getting our alias appearing in unexpected places.  

**Kristina Yasuda:** I would like to ensure that other options are considered as well and we are running out of time.  

> *Manu Sporny:* I'm afraid that we're not going to be able to get to "MUST always use `@value`/$MD_CODE$/$MD_CODE$" when expressing human-readable strings.

**Kristina Yasuda:** Let's discuss option B and A.  

> *Dmitri Zagidulin:* +1 manu.

**Manu Sporny:** The suggestion I'm hearing is to remove option A. It just allows us to use 'prettier' values, but doesn't have any advantage.  

> *Orie Steele:* If folks don't understand "compact vs non compact LD"... they don't understand what a conforming document is.... so we should be cautious requiring "non compact" processing of languages, because they spec does not require people to understand that.

> *Dave Longley:* if option B is putting `@language` as a default language in the context then -1 to that, it corrupts the data.

**Manu Sporny:** Option B provides a document-level default. The issue that we would need to flag is non-natural-language fields being classed as a specific language, such as Base64 data being marked as natural language.  
… We could make Option B a fallback to Option C, but that has downsides for the JSON-LD context architecture.  
… I believe the options here are Option B+C - OR - Option C+D.  

**Addison Phillips:** You can't prevent people from serialising `@language` globally. You could deprecate that behaviour of course.  

> *Orie Steele:* IMO, if you can't stop people from doing something, its considered best practice to give them guidance... and not be silent.

**Sebastian Crane:** I'd like to talk about Option C only.  
… This is an implementation consideration, authenticate users, use existing libraries -- if those tools made it as easy to set a default in the code and have the serialization of fields automatic, as writing the serialized language feature at the top, then people would use that feature. In contrast to HTML, people were hand-writing it... but due to cryptography involved, people aren't hand-writing VCs. Lack of global language feature could.  

> *Manu Sporny:* be side-stepped in implementations.

> *Kristina Yasuda:* I do not like the idea of let's rely on the library to implement this correctly..

> *Shigeya Suzuki:* +1 kristina.

**Ivan Herman:** We could argue the same thing about HTML, as few people write HTML by hand. What's the proportion of tools that produce linguistically undefined documents? Perhaps addison knows.  

> *Shigeya Suzuki:* It's depends on complexity of the output. for a simple VCs, it's not necessary to depends on huge library. not all people have freedom on memory and energy usage.

**Ivan Herman:** Maybe putting the language metadata in all fields is a bit naive.  
… I am not particularly partisan to the technique, but I think it's important to have something for global language.  

> *Ted Thibodeau Jr.:* can we globally say "language: undefined" or "language: various" or similar?

**Addison Phillips:** To Sebastian's point, if you say that it's a MUST and all the libraries implement it, maybe it would be moot. I'm not sure that you could expect that response without a MUST.  

**Kristina Yasuda:** can everyone put your favorite option on IRC.  

> *Kristina Yasuda:* C+D.

> *Ivan Herman:* C+B.

> *Dmitri Zagidulin:* C+D for me.

> *Sebastian Crane:* C.

> *Dave Longley:* -1 to C+B because it corrupts all string fields that are not natural language fields (which is a common thing in VCs).

> *David Chadwick:* c+d.

> *Manu Sporny:* C+B.

> *Phillip Long:* C+B or if we're ranking 1, C+B, 2, B+D, 3, C.

> *Dave Longley:* +1 to just C, I don't think there's a significant difference in MUST/SHOULD with C and having a default language with D, the people that don't want to do it won't do it either way -- and only tools will stop them.

> *Dmitri Zagidulin:* wait D is in the VC core context or in the VC itself?

> *Joe Andrieu:* +1 to C.

> *Manu Sporny:* D is in the VC itself.

> *Kristina Yasuda:* D is in VC itself I think.

> *Shigeya Suzuki:* I think slightly C+B better but C+D is also acceptable.

> *Dave Longley:* that's an insufficient description of B and D ...

**Dmitri Zagidulin:** can we remind people of the difference between D and B?  

**Manu Sporny:** D creates a new feature, B uses an existing JSON-LD feature.  

> *Dave Longley:* B will use a JSON-LD feature that will apply a language to EVERY string field, even non-language fields.

> *Dmitri Zagidulin:* in that case, C+B.

> *Orie Steele:* I think I agree with what dimitri is saying though...

> *Dave Longley:* D will invent something new for VCs but only apply it to natural language text fields.

> *Orie Steele:* ^ yeah... that.

> *Phillip Long:* Ori is up.

**Ivan Herman:** dlongley said that every field will get a language tag with B. However, if we had LD tags for datatypes, that won't be an issue.  

> *Dmitri Zagidulin:* agree with ivan.

**Ivan Herman:** It's not as bad as it looks considering the existence of those JSON-LD datatypes.  

> *Dmitri Zagidulin:* an app somehow interpreting language & direction on a base64 string or whatever is /not/ a realistic problem.

**Dave Longley:** I agree with you that we should use datatypes, but the JOSE and COSE parts do not have data types defined.  

> *Orie Steele:* also not necessary... to do... because the data model is COMPACT JSON_LD !!!!

**Sebastian Crane:** It is a bit involved, I'll write to the mailing list, can we delay a vote for day or two to engage w/ email.  

**Kristina Yasuda:** We appreciate Addison's time. Thank you!  

---

</details>

longpd (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-10-17-vcwg-special#section2-2) on 2023-10-17

- no resolutions were taken

#### 2.2. Add guidance to not use JSON-LD keywords in `@context` value. (pr vc-data-model#1271)

_See github pull request [vc-data-model#1271](https://github.com/w3c/vc-data-model/pull/1271)._




**Brent Zundel:** Manu, you raised this one -- is there a short path toward consensus for this PR?  

**Manu Sporny:** Yeah, it's marked "do not merge", but as you noted there's only one request for changes and no objections. But we should be able to get this forward.  
… I need to talk with Benjamin to see what his concerns were and then maybe this gets merged.  

**Brent Zundel:** When?  

**Manu Sporny:** By next week, I do changes and if there are no objections we can merge, otherwise we can close.  
… This is somewhat attached to the internationalization section and about using `@language` globally in the context.  
… When we ran our poll, there's only two options, one was: use `@language`, `@value`, `@direction` to express language information. There were a number of people that said they wanted to see globals. We want to see a preference poll on those two options.  
… I think the i18n group wants us to be able to both fine grained and default language and we need to figure out what we'll be doing.  

**Brent Zundel:** Is that i18n section a blocker for this?  

**Manu Sporny:** No.  

**Brent Zundel:** Ok, you will have a conversation with Benjamin and if you can resolve that we will get this merged otherwise recommend closing.  

</details>

brianorwhatever:
It doesn't look like `@base` was mentioned on the call. I would like to see this feature not recommended against.

msporny:
@brianorwhatever wrote:
> It doesn't look like `@base` was mentioned on the call. I would like to see this feature not recommended against.

Yes, we didn't discuss `@base` explicitly. The general guidance we're trying to give is: In general, don't use JSON-LD features that affect the way values are processed in a global way because the processors that do limited JSON-LD processing (aka what we used to call "JSON Processing") will almost certainly not detect/use `@base`, `@language`, and `@direction`. We don't forbid the usage with this PR, we just strongly advise against it.

At present, both you and @OR13 seem to be indicating that you would prefer that we DO NOT merge this PR. At this point, we need to know if your input is blocking or not. If it's blocking, and since we're in the endgame now, we'll close this PR and note in both issues that the WG couldn't come to consensus on recommending against JSON-LD keyword usage, close the issues, and leave it at that. On the one hand, this might harm interoperability between processors that do "Minimal JSON-LD Processing" and ones that do "Full JSON-LD Processing"... the upside being that there is a very clear internationalization story (use the `@language` and `@direction` keywords), and the ability to have relative URLs that are not bound to a particular web domain (use the `@base` keyword).

A middle ground might be just warning people that if they use these keywords, interop might be harmed... so we don't recommend for/against... we just let people know that "here be dragons".

At this point, I need explicit and concrete guidance from both @OR13 and @brianorwhatever on 1) whether or not they're blocking this PR, and 2) if they are blocking, concrete changes that would unblock this PR.

OR13:
I suggest recommending people use JSON-LD language features, or withdrawing the PR.

I object to telling people not to use JSON-LD language features.

dlongley:
Recommending that people use a simpler profile when many options are available is good for interop. That doesn't mean people can't use something more complex -- their data will just be rejected (not even consumed at all) by people who only support the simpler profile.

brianorwhatever:
I am not blocking this PR

msporny:
@OR13 wrote:
> I suggest recommending people use JSON-LD language features, or withdrawing the PR.
> I object to telling people not to use JSON-LD language features.

Given your objection, and @brianorwhatever's concern, I've updated the PR to NOT recommend that people don't use JSON-LD language features. The updated PR warns implementers that using JSON-LD keywords in `@context` might reduce interoperability among processors that don't process those values. 

Thus, the PR no longer recommends that people don't use JSON-LD keywords in `@context` values... it merely warns them that if they do so, interop might be harmed among a subset of the processors.

How do deal w/ globally setting language with be handled in another PR... we're down to two options there... 1) use the JSON-LD language features, or 2) create new properties for credential-level language and direction defaults.

Requesting a re-review from @OR13.

dlongley (commented):
I've made some suggestions to help clarify how the interoperability might actually occur -- as I think some people have been (mistakenly) thinking that static processors would accept documents with extra rules to be applied and just ignore / misapply them, when that is not permitted and not what a conforming consumer would do. Instead, they would just reject the documents, i.e., opt not to consume any document with extra `@context` rules that dynamically change the values throughout the document, particularly from `@context` values that they have not written their static processors against.

index.html
```diff
@@ -3987,6 +3987,16 @@ <h3>JSON-LD</h3>
...
         </p>
 
+        <p>
+As elaborated upon in Section <a href="#json-processing"></a>, some software
+applications might not perform full JSON-LD processing.
+Authors of <a>conforming documents</a> are advised that using JSON-LD keywords
+in the `@context` value to globally affect values in documents, such as
+globally setting the keywords `@base` or `@language`, might reduce
+interoperability with certain processors that do not perform full JSON-LD
+processing.
```
OR13, line 3997:
```suggestion
globally setting the keywords `@base` or `@language`, might reduce
interoperability.
```

OR13 (changes requested):
Please don't suggest that JSON vs JSON-LD processing will impact interoperability, we should be saying consistently that JSON-LD processing is required.

index.html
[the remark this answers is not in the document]
TallTed:
```suggestion
globally setting the keywords `@base` or `@language`, might reduce
interoperability with static processors that reject documents that
can only be properly consumed after applying additional rules to 
values within the document.
```

index.html
```diff
@@ -3987,6 +3987,15 @@ <h3>JSON-LD</h3>
 base media type, `application/vc+ld+json`.
         </p>
 
+        <p>
+As elaborated upon in Section <a href="#json-processing"></a>, some software
```
msporny, line 3991:
```suggestion
As elaborated upon in Section <a href="#static-processing"></a>, some software
```

msporny:
> Please don't suggest that JSON vs JSON-LD processing will impact interoperability, we should be saying consistently that JSON-LD processing is required.

I applied your change suggestion, but can't tell if that addresses your concern.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-10-31-vcwg-special#section1-12) on 2023-10-31

- no resolutions were taken

#### 1.12. Warn authors that usage of global JSON-LD keywords might harm interoperability (pr vc-data-model#1271)

_See github pull request [vc-data-model#1271](https://github.com/w3c/vc-data-model/pull/1271)._




**Brent Zundel:** Warn usage of global JSON-LD keywords may harm interop.  
… My proposal is to label this post-CR and talk about it at that point.  
… If I've misinterpreted the changes requested in this PR please let me know -- if this should be pre-CR.  

**Manu Sporny:** It's attached to a pre-CR issue, is the only item I'm noting.  

**Brent Zundel:** Ok, it's attached to the i18n review.  
… What do we do?  

**Manu Sporny:** We are probably going to need to spend some time about what we're doing the JSON-LD features as a mechanism as a mechanism to express the default language and text direction. Using those in an `@context` value or defining our own credential language / direction fields.  
… That seems to be the decision we need to make, once we make that decision, that drives the pre-CR issue.  
… I will note we asked for a review from Orie but he hasn't provided one yet.  
… The other approach is to see if Orie will object to this.  
… It's still pre-CR, we have to make a decision about default language/direction and that requires discussion. Then we can decide what to do with this PR and the issues attached to it.  

**Brent Zundel:** I'm going to mark as discuss and get rid of "do not merge". I will set aside call time tomorrow to talk about this.  
… Ideally with Orie on the call tomorrow.  
… And see what we can find out.  

**Ivan Herman:** I also see a separate discussion on IRC ... but we had a special call, some weeks ago.  
… About this problem and I thought that the agreement ... not this PR itself, but the underlying i18n issue ... and I thought that the idea was that we'll use the global JSON-LD facilities and that we put in a PR along those lines and we try to move on with that.  
… That's my recollection, but my memory may be off.  

**Manu Sporny:** We did have a conversation that Ivan is referencing. I don't think we got to that conclusion though. We got 80% of the way there and I think Andres, you, maybe Gabe, maybe Phil or Dmitri ... some subset of people saying we want new properties.  

> *Phillip Long:* I'm also find with JSON-LD mechanisms.

**Manu Sporny:** Then the question was are the people that don't want the JSON-LD keywords going to object. The other thing was that the i18n group might want us to strongly recommend a default language / direction in all VCs. I'm uncomfortable with that.  
… We need to know if anyone will object to using default JSON-LD terms and do you have to do that for all VCs.  

**Ivan Herman:** I think it would be perfectly fine to put in a PR with those changes -- with the global mechanism and I agree with requiring to set a global language for all VCs for all the text fields ... are in a strong minority anyway, we wouldn't put it there.  
… I think we can move ahead by coming up with a PR and take it from there.  
… And not discuss the issue again and again.  

**Brent Zundel:** PR 1271 should be closed in favor of a new PR?  

**Ivan Herman:** I don't think so. There are also issues about other globals settings.  
… That's a separate one, sorry.  

**Manu Sporny:** This PR could be changed to remove the `@language` and we could pick some other global keywords, I still expect Orie's objection on any feature, we need Orie on the call.  

**Brent Zundel:** So the original course of action is that we'll talk about this a bit tomorrow, whether or not other PRs show up to address the underlying issues.  
… Thank you all, this was a good call, look forward to seeing you on tomorrow, thank you Dave for scribing and Manu as backup.  

---

</details>

OR13 (changes requested):
It would be good to explain why or how interoperability would be reduced.

For example, if an issuer wants to use these features, and they are allowed to use them, and a verifier is required to understand them, why is interoperability lost?

index.html
```diff
@@ -3987,6 +3987,15 @@ <h3>JSON-LD</h3>
 base media type, `application/vc+ld+json`.
         </p>
 
+        <p>
+As elaborated upon in Section <a href="#static-processing"></a>, some software
+applications might not perform dynamic JSON-LD processing.
+Authors of <a>conforming documents</a> are advised that using JSON-LD keywords
+in the `@context` value to globally affect values in documents, such as
+globally setting the keywords `@base` or `@language`, might reduce
+interoperability.
```
TallTed, line 3996: This is an attempt to more clearly open the door to use of `@language` or `@base` where they may increase usability for specific deployments, when used with properties that are not mapped by the default `@context` and/or with URIs that are not relative to inferred `@base`.... I am more than open to improvements on this language. I will not block if this is not merged, though I would like to have an issue (open or closed) on the topic labeled for "VCDM.next" or similar.
```suggestion
Authors of <a>conforming documents</a> are advised that interoperability
might be reduced if JSON-LD keywords in the `@context` value are used to
globally affect values in `@context` or other documents, such as by
globally setting the keywords `@base` or `@language`. Using these
keywords with specific focus, as when aligned with one or more values
of one or more other keywords, may broaden the utility of specific
`@context` files and/or <a>conforming documents</a>.
```

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-11-14-vcwg-special#section1-1) on 2023-11-14

- no resolutions were taken

#### 1.1. Warn authors that usage of global JSON-LD keywords might harm interoperability (pr vc-data-model#1271)

_See github pull request [vc-data-model#1271](https://github.com/w3c/vc-data-model/pull/1271)._




**Brent Zundel:** orie requested explanation of why and how interop would be reduced.  

**Manu Sporny:** I had volunteered but have not had a chance as I though this PR would be closed.  
… one thing worth mentioning, there are other language PRs that Benjamin has raised. we will probably strike telling people they shouldn't use that language in context property.  
… and that they should use `@base`.  
… once done, we should ask Orie for re-review.  

**Brent Zundel:** agreed, if we don't hear by end of this week we merge and indicate we have attempted to address change requests.  

**Manu Sporny:** I'll work on update today.  

</details>

msporny:
@TallTed I accepted and then heavily modified your last suggestion. Please re-review. While I understood what you were trying to say with your last sentence, and it was correct, it was also such a generalized statement that non-expert readers would have a hard time understanding what it meant (or worse, would misinterpret it to mean something entirely different from what was intended). I'm asserting that we can safely not say what you said in your last sentence. In any case, please re-review and see if you're ok w/ the updated language.

@OR13 I have provided an example of how interop can be harmed when using `@base`.

TallTed (approved):
I can live with it.

iherman (approved):

Issue #1206, opened 2023-07-19
Allowing expanded `type` values in conforming documents


OR13:
@TallTed  wrote here: https://github.com/w3c/vc-jose-cose/issues/76#issuecomment-1642334035

> `type` is RDF Class, this question seems more about the core data model.

Given that ``` `type` is RDF Class ```, its value MUST be a URI, not a literal.

[examples/vc-2.0/vc-ld-jwt.jose](https://github.com/w3c/vc-jwt/blob/main/examples/vc-2.0/vc%2Bld%2Bjwt.jose) no longer resolves to a file, so I cannot check its actual content, and say what it *should* decode to contain, but that certainly is not —
```
  "type": [
    "VerifiableCredential",
    "UniversityDegreeCredential"
  ],
```

That _should_ be something more like this, with fully qualified URIs —
```
  "http://www.w3.org/1999/02/22-rdf-syntax-ns#type": [
    "http://example.com/ontology/VerifiableCredential",
    "http://example.com/ontology/UniversityDegreeCredential"
  ],
```

... edit from @or13 below ...

While I think the left hand side is expected to be `type` based on the compact form requirement... I don't know if you are allowed to use the right hand side values as expanded when compact form is used.

This would seem useful to "none JSON-LD processors" who want to distinguish:

Cases like this:

```
  "type": [
    "https://www.w3.org/2018/credentials#VerifiableCredential",
    "https://vendor1.example/ontology/UniversityDegreeCredential",
    "https://vendor2.example/ontology/UniversityDegreeCredential",
  ],
```

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-07-26-vcwg#section2-2) on 2023-07-26

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 2.2. Allowing expanded `type` values in conforming documents (issue vc-data-model#1206)

_See github issue [vc-data-model#1206](https://github.com/w3c/vc-data-model/issues/1206)._




**Manu Sporny:** It's fine to use full URIs in normal JSON-LD documents, however if publishing in JSON-only mode implementors would have to add support for both shortened and full URIs.  

**Ted Thibodeau Jr.:** This reminds me of a discussion about MIME types for VCs, and that to be able to use the JSON media type for JSON-LD documents the URI has to be expanded.  

[...]

PR #1276, opened 2023-09-11
Provide guidance to use compact terms/types instead of full URLs

msporny:
This PR attempts to address issue #1206 by advising that document authors do not use full URLs, such as `https://schema.org/Person` for terms/types, and instead use the compact form (`Person`), in order to increase compatibility with processors that don't do full JSON-LD processing.

### DIFF
--- a/index.html
+++ b/index.html
@@ -4001,6 +4001,11 @@ Not using JSON-LD keywords in the `@context` value that modify document
 processing, such as globally setting `@base` and `@language`, in order to
 increase interoperability.
           </li>
+          <li>
+Always using the compact form for JSON-LD terms and types and always ensuring
+that those terms and types are defined in application-specific JSON-LD context
+files.
+          </li>
        </ul>
 
         </p>
--- a/index.html
+++ b/index.html
@@ -3989,10 +3989,20 @@ base media type, `application/vc+ld+json`.
 
         <p>
 As elaborated upon in Section <a href="#json-processing"></a>, some software
-applications might not perform full JSON-LD processing.
-<a>Conforming document</a> authors are urged to not use JSON-LD keywords that
-affect document processing in the `@context` value, such as globally setting
-`@base` and `@language`, in order to increase interoperability.
+applications might not perform full JSON-LD processing. In order to increase
+interoperability, <a>conforming document</a> authors are urged to not use
+JSON-LD features that are not easily detected when not performing full
+JSON-LD processing. This guidance includes:
+        </p>
+
+        <ul>
+          <li>
+Not using JSON-LD keywords in the `@context` value that modify document
+processing, such as globally setting `@base` and `@language`, in order to
+increase interoperability.
+          </li>
+       </ul>
+
         </p>
 
         <section>


iherman (approved):

decentralgabe (approved):
are there any static validation tools implementers can use to not footgun themselves? maybe better suited for an impl guide...but worth having

dlongley (approved):

index.html
```diff
@@ -3989,10 +3989,25 @@ <h3>JSON-LD</h3>
...
         <p>
 As elaborated upon in Section <a href="#json-processing"></a>, some software
-applications might not perform full JSON-LD processing.
-<a>Conforming document</a> authors are urged to not use JSON-LD keywords that
-affect document processing in the `@context` value, such as globally setting
-`@base` and `@language`, in order to increase interoperability.
+applications might not perform full JSON-LD processing. In order to increase
+interoperability, <a>conforming document</a> authors are urged to not use
+JSON-LD features that are not easily detected when not performing full
+JSON-LD processing. This guidance includes:
```
TallTed, line 3995:
```suggestion
applications might not perform full JSON-LD processing. To increase
interoperability, <a>conforming document</a> authors are urged to not use
JSON-LD features that are not easily detected unless performing full
JSON-LD processing. These features include:
```

index.html
```diff
@@ -3989,10 +3989,25 @@ <h3>JSON-LD</h3>
...
+        </p>
+
+        <ul>
+          <li>
+Not using JSON-LD keywords in the `@context` value that modify document
+processing, such as globally setting `@base` and `@language`, in order to
+increase interoperability.
```
TallTed, line 4002:
```suggestion
Use of JSON-LD keywords in the `@context` value that modify document
processing, such as global settings of `@base` and `@language`
```

index.html
```diff
@@ -3989,10 +3989,25 @@ <h3>JSON-LD</h3>
...
+          </li>
+          <li>
+Always using the compact form for JSON-LD terms and types and always ensuring
+that those terms and types are defined in application-specific JSON-LD context
+files.
```
TallTed, line 4007:
```suggestion
Use of the compact form for JSON-LD terms and types and defining those
terms and types in application-specific JSON-LD context files
```

TallTed (changes requested):

TallTed:
I *think* I preserved intended meaning. Negation of negative negation is challenging.

index.html
pchampin, line 4002:
```suggestion
processing, such as globally setting `@base` and `@language`.
```
This has been stated already in the paragraph before the bullet list (and this actually applies to all items).

TallTed: See https://github.com/w3c/vc-data-model/pull/1276#discussion_r1325510657

OR13 (changes requested):
This still does not address what the value of the non compact JSON-LD is... and if that value is not elaborated on substantially, we should recommend no JSON-LD processing at all.

I would like to see language that specifically addressed application n-quads, or JSON-LD with `@graph`

OR13:
I recognize it might be better addressed in the value of JSON-LD PR. Before we add more text saying you don't need to do JSON-LD processing, and should avoid creating problems when JSON-LD processing occurs we need to address the value of JSON-LD processing, and in particular the value of RDF... which is ultimately what JSON-LD is about.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-09-14-vcwg#section1-7) on 2023-09-14

- no resolutions were taken

#### 1.7. Provide guidance to use compact terms/types instead of full URLs (pr vc-data-model#1276)

_See github pull request [vc-data-model#1276](https://github.com/w3c/vc-data-model/pull/1276)._




**Brent Zundel:** Few requests for changes from TallTed. Not a lot of review yet, raised 2 days ago.  
� We'd like to see how to move this forward.  

**Manu Sporny:** This is in response to an issue that DavidC raised.  
� We're suggesting to people that they do not use full URLs for things that have json-ld terms. Doing so would harm interop. Not illegal, just a suggestion to not do it.  

**Brent Zundel:** Please review folks.  

</details>

index.html
David-Chadwick: For implementors doing JSON only processing, then having to define an application specific JSON-LD context file is perhaps asking too much, and they should be allowed to use URLs for the new application specific types

David-Chadwick: Otherwise what is the point of having the '@vocab' keyword if every term and type must have a compact form defined in an `@context` file

David-Chadwick, line 4007: Suggest replacing the above three lines by:

Always using the compact form for JSON-LD terms and types when these are defined in JSON-LD context files.
Application specific terms and types MAY use the long (URI) form when the compact forms are not defined in any JSON-LD context file.

David-Chadwick (changes requested):

index.html
TallTed: @David-Chadwick -- I'm confused by your comments. Note that the `Use ... files` in green in my change request, which replaced the three lines you're suggesting be replaced by something else, is an item in unordered list of "avoid using these JSON-LD features" ... and you seem to be saying "don't tell implementers to use JSON-LD context files" which is already happening. Part of my confusion stems, I think, from your having made 3 comments on the same bullet point, which comments don't seem to be fully connected thoughts, nor to take my existing change request into consideration.

My sense is that you're reading the PR Preview, which doesn't reflect suggested changes until/unless the suggestions are merged.... Please take another look at the HTML source, and base your suggestion on the total context of that unordered list, starting with [line 3991](https://github.com/w3c/vc-data-model/pull/1276/files#r1325509056) and continuing to line 4011?

TallTed: > [@David-Chadwick] it appears to only remove the word 'always'

I strongly suggest that you re-examine [my change request for that paragraph (bullet point)](https://github.com/w3c/vc-data-model/pull/1276#discussion_r1325512116), and tell me again whether it appears to only remove the word 'always'. If so, I will have little choice but to ignore your comments, as you are not responding to the full flesh of my own; indeed, you're acting as if my change requests do not exist.

PRs on GitHub are conversations. PR Preview is a *helper* but it only shows the result of the PR as initially drafted, and as/when conversation on the PR results in _revision_ of the PR (which is to revise the text of the document). Ignoring the conversational aspect of this tool does a disservice to all involved, yourself included.

(To be most explicit: My change request significantly rewords the content of that bullet, which is to make it make sense in context of my small rewording of the intro to those bullets, along with my suggested changes to the other bullets in the list, which are to make them all agree with one another. Your own change only applies to the single bullet, not to the intro, nor the other bullets, and does not appear to me to help in bringing clarity to this segment of the document.)

msporny: > For implementors doing JSON only processing, then having to define an application specific JSON-LD context file is perhaps asking too much, and they should be allowed to use URLs for the new application specific types

They don't have to do this because we now have an "issuer defined" `@vocab` value in the base context.

I'll try to figure out how to make this more clear -- remember, these are not normative MUSTs -- they are recommendations to increase interoperability. They don't /have to/ define "terms and types in application-specific JSON-LD context files" -- can you suggest some language that you'd be ok with?

msporny: @David-Chadwick wrote:
> Always using the compact form for JSON-LD terms and types when these are defined in JSON-LD context files.

This seems fine.

> Application specific terms and types MAY use the long (URI) form when the compact forms are not defined in any JSON-LD context file.

Hrm, but we have a base `@vocab` that makes it so they don't have to do that. I believe there were objections in the group to using long-form URLs and adding the text you suggest above would re-raise those objections. @OR13, I believe you were one of the people that wanted clear guidance here -- thoughts?

OR13:
@msporny I think this comment addresses the impact that `@vocab` has on `@type`... https://github.com/w3c/vc-data-model/issues/1290#issuecomment-1733969379

^ if folks don't understand what the `@vocab` is doing in the v2 context... they are probably not ready to use W3C Verifiable Credentials... the context is now normative.

index.html
OR13: Sorry, I escaped the thread: https://github.com/w3c/vc-data-model/pull/1276#issuecomment-1739187061

The linked comment is probably a better way to address this issue.... the default behavior should not appear to be illegal, imo.

dlongley: My comments here are relevant: https://github.com/w3c/vc-data-model/issues/1290#issuecomment-1737778877

David-Chadwick: `Application specific terms and types MAY use the long (URI) form when the compact forms are not defined in any JSON-LD context file.`
The reason for adding this text is that it still allows compact local names to be used because of the `@vocab` term, but it also allows implementors who wish to use a global long (URI) name to use that as well, ensuring that no other VC Issuer will have a name clash with their invented compact local name when they do not want to start creating their own `@context` files

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-10-11-vcwg#section1-13) on 2023-10-11

- no resolutions were taken

#### 1.13. Provide guidance to use compact terms/types instead of full URLs (pr vc-data-model#1276)

_See github pull request [vc-data-model#1276](https://github.com/w3c/vc-data-model/pull/1276)._




**Brent Zundel:** Guidance for the use of compact terms and types rather than full URLs. I know this is related to the light processing PRs.  

**Manu Sporny:** There's an issue raised to provide clear guidance on using compact terms vs. full URLs. I think Orie is saying that this PR doesn't address this issue and it might be better address in the value of JSON-LD PR. I don't know what is needed to make it so that Orie is ok with it. It is a PR that would be closed because we couldn't get to consensus.  

**Brent Zundel:** Thank you, Manu. We're at time for the meeting. Appreciate folks joining, kinda crazy with people at IIW, etc.  
… Thank you for scribing, Dave, thanks for all the work people have done, see you later!  

---


</details>

longpd (approved):

msporny:
@TallTed @OR13 @David-Chadwick Your change requests have resulted in this PR being marked as pending close. If you feel you can save this PR w/ change suggestions, please do, if not, the PR will be closed and the issue will go unresolved in this iteration of the WG.

OR13:
Apply or dismiss suggestions, seems the PR could be merged , with issue markers, in case there is no consensus.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-10-17-vcwg-special#section2-3) on 2023-10-17

- no resolutions were taken

#### 2.3. Provide guidance to use compact terms/types instead of full URLs (pr vc-data-model#1276)

_See github pull request [vc-data-model#1276](https://github.com/w3c/vc-data-model/pull/1276)._




**Brent Zundel:** This PR three different folks requesting changes and an equal number who have approved. Not clear what the path toward consensus is.  
… Chairs would move toward closing it -- but would like a recommendation toward coming to agreement.  

**Manu Sporny:** +1 to close this, I think we should let Orie, David, and Ted that their change requests are causing this to not go in and it's up to them to propose a change within a week and this will be otherwise closed.  
… We will, at DB, -1 any further attempt to change this in this WG if there's no consensus here.  

</details>

index.html
```diff
@@ -3989,10 +3989,23 @@ <h3>JSON-LD</h3>
...
         <p>
 As elaborated upon in Section <a href="#json-processing"></a>, some software
-applications might not perform full JSON-LD processing.
-<a>Conforming document</a> authors are urged to not use JSON-LD keywords that
-affect document processing in the `@context` value, such as globally setting
-`@base` and `@language`, in order to increase interoperability.
+applications might not perform full JSON-LD processing. To increase
+interoperability, <a>conforming document</a> authors are urged to avoid using
+JSON-LD features that are easily detected only when performing full
+JSON-LD processing. These features include:
```
msporny, line 3995:
```suggestion
applications might not perform full JSON-LD processing. 
<a>Conforming document</a> authors are advised that 
usage of certain JSON-LD features might reduce 
interoperability. These features include:
```

index.html
```diff
@@ -3989,22 +3989,20 @@ <h3>JSON-LD</h3>
...
         <p>
 As elaborated upon in Section <a href="#json-processing"></a>, some software
-applications might not perform full JSON-LD processing. In order to increase
-interoperability, <a>conforming document</a> authors are urged to not use
-JSON-LD features that are not easily detected when not performing full
-JSON-LD processing. This guidance includes:
+applications might not perform full JSON-LD processing. To increase
+interoperability, <a>conforming document</a> authors are urged to avoid using
+JSON-LD features that are easily detected only when performing full
+JSON-LD processing. These features include:
```
```suggestion
applications might not perform dynamic JSON-LD processing, electing to
use static processing that will, for example, reject documents with unrecognized
contexts. To increase interoperability, <a>conforming document</a> authors are
urged to avoid using JSON-LD features that are easily detected only when
performing dynamic JSON-LD processing. <a>Conforming document</a> authors
are advised that usage of certain JSON-LD features might reduce 
interoperability. These features include:
```

msporny:
@OR13 wrote:
> Apply or dismiss suggestions, seems the PR could be merged , with issue markers, in case there is no consensus.

I have merged @TallTed's suggestions and have attempted to address @David-Chadwick and @OR13's concerns by doing the following:

1. To address @OR13's concerns, we don't advise against using JSON-LD keywords (but do warn that using them might reduce interoperability), and 
2. To address @David-Chadwick's concerns, we don't make using full URLs "illegal" (but do warn that using them might reduce interoperability)

Requesting re-review from both of you.

index.html
```diff
@@ -3989,22 +3989,20 @@ <h3>JSON-LD</h3>
...
         </p>
 
         <ul>
           <li>
-Not using JSON-LD keywords in the `@context` value that modify document
-processing, such as globally setting `@base` and `@language`, in order to
-increase interoperability.
+Use of JSON-LD keywords in the `@context` value that modify document
+processing, such as global settings of `@base` and `@language`
```
dlongley, line 4001:
```suggestion
Use of JSON-LD keywords in the `@context` value that globally modify document
value processing, such as global settings of `@base` and `@language`
```

index.html
```diff
@@ -3989,22 +3989,20 @@ <h3>JSON-LD</h3>
...
           </li>
           <li>
-Always using the compact form for JSON-LD terms and types and always ensuring
-that those terms and types are defined in application-specific JSON-LD context
-files.
+Use of the compact form for JSON-LD terms and types and defining those
+terms and types in application-specific JSON-LD context files
```
dlongley: This doesn't make sense with the suggestions above (it inverts what should be done since this is to be "avoided", so suggest changing to:

```suggestion
Use of the non-compact form for JSON-LD terms and types. Instead prefer using
compact form and define terms and types in application-specific JSON-LD context files.
```

index.html
```diff
@@ -3989,10 +3989,24 @@ <h3>JSON-LD</h3>
...
         <p>
 As elaborated upon in Section <a href="#json-processing"></a>, some software
-applications might not perform full JSON-LD processing.
-<a>Conforming document</a> authors are urged to not use JSON-LD keywords that
-affect document processing in the `@context` value, such as globally setting
-`@base` and `@language`, in order to increase interoperability.
+applications might not perform full JSON-LD processing. 
+<a>Conforming document</a> authors are advised that 
+usage of certain JSON-LD features might reduce 
+interoperability. These features include:
```
dlongley, line 3995:
```suggestion
applications might not perform dynamic JSON-LD processing, electing to
use static processing that will, for example, reject documents with unrecognized
contexts. To increase interoperability, <a>conforming document</a> authors are
urged to avoid using JSON-LD features that are easily detected only when
performing dynamic JSON-LD processing. <a>Conforming document</a> authors
are advised that usage of certain JSON-LD features might reduce 
interoperability. These features include:
```

index.html
```diff
@@ -3989,10 +3989,27 @@ <h3>JSON-LD</h3>
...
-<a>Conforming document</a> authors are urged to not use JSON-LD keywords that
-affect document processing in the `@context` value, such as globally setting
-`@base` and `@language`, in order to increase interoperability.
+applications might not perform dynamic JSON-LD processing, electing to
+use static processing that will, for example, reject documents with unrecognized
+contexts. To increase interoperability, <a>conforming document</a> authors are
+urged to avoid using JSON-LD features that are easily detected only when
+performing dynamic JSON-LD processing. <a>Conforming document</a> authors
+are advised that usage of certain JSON-LD features might reduce 
+interoperability. These features include:
```
msporny, line 3998:
```suggestion
contexts. <a>Conforming document</a> authors are advised 
that usage of certain JSON-LD features might reduce 
interoperability. These features include:
```

index.html
```diff
@@ -3989,10 +3989,25 @@ <h3>JSON-LD</h3>
 
         <p>
 As elaborated upon in Section <a href="#json-processing"></a>, some software
```
msporny, line 3991:
```suggestion
As elaborated upon in Section <a href="#static-processing"></a>, some software
```

msporny:
@OR13 wrote:
> Apply or dismiss suggestions, seems the PR could be merged , with issue markers, in case there is no consensus.

I have merged @TallTed's  and @dlongley's suggestions. I have attempted to address @David-Chadwick and @OR13's concerns by doing the following:

1. To address @OR13's concerns, we don't advise against using JSON-LD keywords (but do warn that using them might reduce interoperability), and 
2. To address @David-Chadwick's concerns, we don't make using full URLs "illegal" (but do warn that using them might reduce interoperability)
3. To try to get past the "JSON Processing" vs. "JSON-LD Processing" naming issue... we try to use new terminology "Static Processing" (e.g. JSON Schema-based approaches) vs. "Dynamic Processing" (e.g., JSON-LD compact/expand approaches).

Requesting re-review from each of you.

index.html
msporny: I've attempted to address this issue in the latest series of edits, please re-review @David-Chadwick.

David-Chadwick (changes requested):
"Using full URLs for JSON-LD terms and types instead of the compact form of those
values defined in application-specific JSON-LD context files" still does not address my concern that it is OK to use a full URL when there is no compact form defined in the context file

msporny:
@David-Chadwick wrote:
> "Using full URLs for JSON-LD terms and types instead of the compact form of those values defined in application-specific JSON-LD context files" still does not address my concern that it is OK to use a full URL when there is no compact form defined in the context file

Please provide some concrete text that would address your concern, noting that it seems like @OR13 is objecting to the text that you want.

The current PR does not make the thing you want to do illegal, @David-Chadwick... but it warns that interop might suffer if you start using full URLs in places (in order to address @OR13's concerns).

If you guys can't come to an agreement on a path forward, the specification will continue to stay silent on the matter of using full URLs or not. Please work together to get to specification text that is acceptable to both of you.

index.html
```diff
@@ -3885,6 +3885,29 @@ <h3>JSON-LD</h3>
...
         </p>
 
+        <p>
+As elaborated upon in Section <a href="#static-processing"></a>, some software
+applications might not perform dynamic JSON-LD processing, which utilizes
+features such as JSON-LD compaction and expansion. Static processors will
+elect to, for example, reject unrecognized URL values in the
+`@context` property. <a>Conforming document</a> authors are also advised that
+usage of certain dynamic JSON-LD processor features might reduce
+interoperability. These features include:
```
TallTed, line 3895:
```suggestion
As elaborated upon in Section <a href="#static-processing"></a>, some software
applications might not perform dynamic JSON-LD processing, which uses features
such as JSON-LD compaction and expansion. Static processors can elect to reject
unrecognized URL values in the `@context` property, for example. Authors of
<a>conforming documents</a> are also advised that use of certain dynamic
JSON-LD processor features might reduce interoperability. These features
include the following:
```

index.html
```diff
@@ -3885,6 +3885,29 @@ <h3>JSON-LD</h3>
...
+        </p>
+
+        <ul>
+          <li>
+Use of JSON-LD keywords in the `@context` value that globally modify document
+value processing, such as global settings of `@base` and `@language`
+          </li>
+          <li>
+Using full URLs for JSON-LD terms and types instead of the compact form of those
+values defined in application-specific JSON-LD context files
```
TallTed, line 3905:
```suggestion
Use of full URLs for JSON-LD terms and types, instead of the compact forms of
those values as are defined in application-specific JSON-LD context files
```

TallTed (changes requested):
Editorial prose massage

index.html
David-Chadwick: We can also add another interop problem, specifically

"Use of full URLs for JSON-LD terms and types, when the compact forms of
these values are not defined in application-specific JSON-LD context files"

This is acknowledging that an application may do this but it could cause interop problems.

David-Chadwick: @OR13 are you happy for the above to be added

msporny:
> If the spec doesn't say anything different, whatever is legal in JSON-LD is legal, so my argument remains, reflect what JSON-LD does by default or close the PR and have the same effect.

Note that the whole reason we're working on this PR is to address issue #1206, which you raised. If you feel that we don't need to say anything to address that issue, I'm happy to close both this PR and that issue, and allow @David-Chadwick to continue doing what he was doing (injecting full URLs wherever he pleases).

This PR doesn't say using any JSON-LD feature illegal; it's saying that using advanced/non-common features could have negative interop consequences (just like using any advanced language features can potentially lead to interop consequences in any language).

> Telling people the core data model's major benefit is JSON-LD / RDF and then telling people not to use it... makes very little sense to me.

That's not what this PR is saying. It's saying that using advanced / non-common features might have negative results when it comes to interop.

For example:

* Using C++20 language features will ensure that you won't be able to compile in ecosystems that only go as high as C++11.
* Writing to a college-level of education will fail to convey your message to the subset of the population that isn't at a college level of reading.
* Mandating that all new vehicles are powered by hydrogen doesn't help sell cars to areas of the population that don't have hydrogen refueling facilities.

Just because /you/ can do something fancy/great doesn't mean that everyone else wants to do that fancy/great thing and creates problems if you want to integrate into an ecosystem that doesn't have the advanced tooling that you have yet (or want to use it).

This is about communicating among an ecosystem and maximizing interop -- and if you use advanced features, you're going to end up leaving some subset of that ecosystem behind, and that's (generally) a bad thing.

So, this PR is trying to say: If you care about broad interop, don't use certain features that will harm broad interop.

@OR13, please provide concrete feedback to this PR if you'd like to see changes. It's not clear if you're blocking at the moment or not.

index.html
```diff
@@ -3885,6 +3885,29 @@ <h3>JSON-LD</h3>
...
+        </p>
+
+        <ul>
+          <li>
+Use of JSON-LD keywords in the `@context` value that globally modify document
+value processing, such as global settings of `@base` and `@language`
+          </li>
+          <li>
+Use of full URLs for JSON-LD terms and types, instead of the compact forms of
+those values as are defined in application-specific JSON-LD context files
```
David-Chadwick, line 3905: What about
"instead of the compact forms of those values whether or not they are defined in application-specific JSON-LD context files"
This covers the cases when a new context file is created or when a new context file is not created.

David-Chadwick (changes requested):

index.html
TallTed: There *is* no "compact form" of such URL values, if they're not mapped (I think "defined" remains the wrong verb for what `@context` files do, whether I accidentally re-introduced it or overlooked it, in a relevant comment or suggestion related to this thread) by an `@context` file, no matter whether that file is old, new, generic, application-specific, etc.

The thing called "compact form" here — probably better to be called the "unmapped form" — is the naked word, the non-URL. The full URL is what that "unmapped form" is mapped to by *some* `@context` identified by the JSON-LD document in question, which is probably one of the documents cited in the VC specifications, unless the "unmapped form" in question is not mapped by any of those VC `@context` files.

These "unmapped forms" *might* be validly referred to as "JSON keys" given that a JSON file contains some number of key/value pairs; and a JSON-LD file contains some number of key/value pairs where the JSON-LD-relevant keys are mapped by a file identified by the `@context` property, and some additional number of key/value pairs involving keys that remain un-mapped by any identified `@context` are ignored as non-conforming or uncrecognized.

```suggestion
those values as are mapped by application-specific JSON-LD context files
```

Beyond that, I don't believe I drafted these bullets nor their introductory paragraph, though I did content-edit both, so I'm hoping earlier author(s) will chime in on whether they see their meaning as being preserved, lost, clarified, distorted, etc., by current efforts.

index.html
David-Chadwick: Replace with
"Use of full URLs for JSON-LD terms and types, instead of the compact forms of
those values which may or may not be defined in application-specific JSON-LD context files"

David-Chadwick (changes requested):

index.html
TallTed: This suggestion makes no sense in context of the existing thread (started by [your 19 hour older comment](https://github.com/w3c/vc-data-model/pull/1276#discussion_r1375782617), followed on by [my 10 hour older comment](https://github.com/w3c/vc-data-model/pull/1276#discussion_r1376988222)). The verb tense here ("Using") breaks from the intro preceding the `<ul>` and differs from that in both bullets including this one. I had much to say about your "values which may or may not be defined"/"values whether or not they are defined" in [my previous comment](https://github.com/w3c/vc-data-model/pull/1276#discussion_r1376988222), but that comment appears to have been ignored completely.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-10-31-vcwg-special#section1-11) on 2023-10-31

- no resolutions were taken

#### 1.11. Warn that using dynamic processor features might lead to less interoperability (pr vc-data-model#1276)

_See github pull request [vc-data-model#1276](https://github.com/w3c/vc-data-model/pull/1276)._




**Brent Zundel:** This one is already marked pending close.  
… David is requesting changes, Orie is requesting changes.  
… It's been two weeks.  
… We did get some more feedback from David.  
… Orie seems fine with it being closed. That's where we're at. I'm planning on closing after the call at this point.  

**Manu Sporny:** The reason it was marked pending close is because Orie was objecting but just didn't use the button to object. I don't see a way around ... we need Orie to respond or it's getting closed and I expect closure.  
… If this is closed, we should also close the feeder issue, which was raised by Orie.  
… There are a subset of people in the WG that are pushing for all JSON-LD features and not to say anything else. This PR was about saying that interop will be better if you don't use all features all the time.  
… This is why we were going to say that if you use fewer features and profile down that you'll get more interop and it's fine and legal to do that both that and with more features. This PR was saying that the more features you use, the less interop you may see.  
… I don't know how that affects the original issue Orie raised, but if we close this PR, we should close that other issue as well.  
… Noting what Dave is saying -- this is a post-CR issue, we could mark it as that.  
… That's another path forward here. I want to make it clear that David Chadwick has been saying he wants to use full URLs in certain places, there's no spec text stopping that, it's just going to lead to interop issues and the spec won't have had any warnings about that.  
… We will continue to just stay vague.  

> *Manu Sporny:* This is also related to [https://github.com/w3c/vc-data-model/issues/1206](https://github.com/w3c/vc-data-model/issues/1206).

**Brent Zundel:** I appreciate Dave pointing out that this is a post-CR PR.  
… It would be inline with the other PRs that we don't have consensus on to mark it as post-CR if we feel like it's possible to have a path forward that leads to some sort of consensus.  

**Dave Longley:** If I recall, Orie's objection was around recommending that people don't use certain JSON-LD features. I think we've changed the text to say it's not the goal... we want to inform people that if they use a lot of features, they might get less interop.  
… I don't know if we've gotten sufficient feedback from Orie on that, but that might be a path forward in the post-CR phase.  

**Brent Zundel:** Thank you, Dave, I appreciate that comment, I've marked it post-CR.  

</details>

index.html
David-Chadwick: I get your point. But I am not sure you get the point I am trying to make, which is: an issuer that only uses the default at context files, but defines their own application types, wishes to use the full URLs for these, since the VCDM requires them to be URLs. The issuer does not want to use a local string that maps via the at vocab into a URL, since another issuer could use the same local string that maps into the same URL via the at vocab, whilst it has a different semantic. Using the full URL is conformant to the VCDM but the proposed text says that it might affect interoperability. However using a local string could cause a bigger car crash because recipients would interpret two different application types to be the same one by mapping them both via the at vocab.

index.html
TallTed, line 3905:
```suggestion
Use of full URLs for JSON-LD terms and types (e.g.,
`https://www.w3.org/2018/credentials#VerifiableCredential`), instead of the
unmapped forms of any such values (e.g., `VerifiableCredential`) that are
among active JSON-LD `@context` mappings (e.g.,
`https://www.w3.org/ns/credentials/v2`)
```

index.html
TallTed: I believe I understand what you're trying to say here. I think it may not be possible (nor necessary) to wedge all that nuance into this list item. I have a [new draft for the list item](https://github.com/w3c/vc-data-model/pull/1276/files#r1382158148), which changes both existing lines, so it's not going directly on this thread (which is only attached to one line of text). I suggest trying to craft a new list item, if the existing list items do not now cover your concern as elaborated above.

index.html
```diff
@@ -3885,6 +3885,29 @@ <h3>JSON-LD</h3>
...
+unrecognized URL values in the `@context` property, for example. Authors of
+<a>conforming documents</a> are also advised that use of certain dynamic
+JSON-LD processor features might reduce interoperability. These features
+include the following:
+        </p>
+
+        <ul>
+          <li>
+Use of JSON-LD keywords in the `@context` value that globally modify document
+value processing, such as global settings of `@base` and `@language`
```
msporny, line 3901:
```suggestion
value processing, such as global settings of `@base`
```

index.html
```diff
@@ -3885,6 +3885,32 @@ <h3>JSON-LD</h3>
...
+
+        <ul>
+          <li>
+Use of JSON-LD keywords in the `@context` value that globally modify document
+value processing, such as global settings of `@base`
+          </li>
+          <li>
+Use of full URLs for JSON-LD terms and types (e.g.,
+`https://www.w3.org/2018/credentials#VerifiableCredential`), instead of the
+unmapped forms of any such values (e.g., `VerifiableCredential`) that are
```
David-Chadwick, line 3906: (e.g., `https://www.w3.org/2018/credentials#VerifiableCredential` or `https://example.org/somenewtype`) instead of the unmapped forms of any such values (e.g., `VerifiableCredential` or `somenewtype`) that are

index.html
```diff
@@ -3885,6 +3885,32 @@ <h3>JSON-LD</h3>
...
+among active JSON-LD `@context` mappings (e.g.,
+`https://www.w3.org/ns/credentials/v2`)
```
David-Chadwick, line 3908: add
" through either explicit mappings or the `@vocab` feature that covers all other unlisted values"

David-Chadwick (changes requested):

index.html
David-Chadwick: Alternatively we could change the text of this PR to be more precise, which you could help by wordsmithing the following e.g.

"using full URLs for JSON-LD terms and types instead of the compact form of those
values that are either explicitly defined in application-specific JSON-LD context files, or are implicitly defined by the  `@vocab` feature of the base `@context` file `https://www.w3.org/ns/credentials/v2`

David-Chadwick (changes requested):
It would be helpful to use an example of the problem that I am raising

index.html
TallTed: @David-Chadwick -- "that are" what? [Your comment above](https://github.com/w3c/vc-data-model/pull/1276#discussion_r1382566565) seems to have been cut off.

David-Chadwick: no its not. This is a replacement for lines 3905 and 3906 so that 3907 continues as before

decentralgabe (approved):

TallTed:
> [@David-Chadwick] It would be helpful to use an example of the problem that I am raising

Please do provide such an example!

index.html
David-Chadwick: Thankyou. I will try this next time. I was not aware of the feature. I only thought you could edit line by line

index.html
TallTed: > [@David-Chadwick ] add 
> " through either explicit mappings or the @vocab feature that covers all other unlisted values"

I don't understand what you intend to communicate by this addition, so I cannot suggest revision. My inability to decipher your meaning might suggest a need for rephrasing such that others — who do not have even the benefit of our discussions to date — to understand it.

David-Chadwick:
@TallTed Can you please wordsmith this to be precise

```
Use of full URLs for JSON-LD terms and types (e.g., 
`https://www.w3.org/2018/credentials#VerifiableCredential` or `https://example.org/somenewtype`) instead of the
unmapped forms of any such values (e.g., `VerifiableCredential` or `somenewtype`) that either are
among explicit JSON-LD `@context` mappings (e.g.,
`https://www.w3.org/ns/credentials/v2`) or are implicitly mapped via the @vocab feature that covers all unlisted values.
```

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-11-14-vcwg-special#section1-2) on 2023-11-14

- no resolutions were taken

#### 1.2. Warn that using dynamic processor features might lead to less interoperability (pr vc-data-model#1276)

_See github pull request [vc-data-model#1276](https://github.com/w3c/vc-data-model/pull/1276)._




**Brent Zundel:** this is post-cr, so would like to understand status today.  

**Manu Sporny:** I believe DavidC was asking for changes and manu requested and received re-review with some subsequent discussion.  
… I think that we need an example from DavidC to move this along.  
… sounds like DavidC is opposed to this PR as it stands, would like some changes, and we need explicit change requests.  

**David Chadwick:** I did produce the text, but TallTed had suggested reformat as a block.  
… at the moment there is one example, which uses a term defined in `@context` and uses the alias for that.  
… I wanted to add a second example with an alias not explicitly defined in `@context` that would be turned into a URI using `@vocab`.  

**Ted Thibodeau Jr.:** that explanation is helpful, but I don't see that communicated in the current text.  
… if you update your change request with this text I think it will be easier to move forward.  

**David Chadwick:** I had some difficulty with the GitHub UI trying to add multiple lines to the change request.  
… I will try just adding a block of text to bypass this (maybe broken) UI issue.  

**Manu Sporny:** Can we change this one to pre-cr? I've been ignoring post-cr issues, but this one seems like it may have a resolution this week.  

**Brent Zundel:** It might make sense to do that, yes.  
… it seems like we are close, and ideally this one will get merged this week.  

</details>

msporny:
@David-Chadwick I've edited your content here: https://github.com/w3c/vc-data-model/pull/1276#issuecomment-1810623057

msporny:

Issue #1185, opened 2023-07-05
Conformance requirements for normative context


OR13:
`@context` value is now normative.

What should conformant processors expect regarding operation that apply the context.

Should `jsonld.expand` succeed? 

If `jsonld.frame` fails, should a verify raise an error?

etc...

Its not clear why the value is normative, and what part of it, an implementer must understand to conform.

It seems like a few sentences about `frame` or `expand` would be enough to cover this.

dlongley:
We already reference that `@context` is defined by the JSON-LD spec (https://w3c.github.io/vc-data-model/#json-ld) and that spec has its own [conformance section](https://www.w3.org/TR/json-ld11/#conformance). Talking about performing any JSON-LD-related transformations is something that would be done in other specs such as a data integrity cryptosuite spec. So, IMO, I think we're already covered for the core data model spec and can close this.

OR13:
Concrete example, an issuer redefines a protected term in a second context, but doesn't actually apply the context when issuing a credential.

2 verifiers both receive the credential, and verify it successfully but one tries to store it as RDF and encounters the protected term redefined error... The other stores it as JSON.

I guess this highlights that this is a securing spec issue, not a core data model issue?

dlongley:
@OR13,

> Concrete example, an issuer redefines a protected term in a second context, but doesn't actually apply the context when issuing a credential.
> 
> 2 verifiers both receive the credential, and verify it successfully but one tries to store it as RDF and encounters the protected term redefined error... The other stores it as JSON.

In order to [process a VC as "plain" JSON](https://w3c.github.io/vc-data-model/#semantic-interoperability), its contexts must be expected values:

> JSON-based processors MUST process the `@context` key, ensuring the expected values exist in the expected order for the credential type being processed.

This means that the contexts are well-known and have been vetted at least once and either stored permanently or matched up against some expected content hash. It doesn't require that the processing software apply a runtime JSON-LD transform (nor even include a JSON-LD library as a dependency), just that the contexts have been previously vetted (perhaps involving a one-time use of a JSON-LD transform performed by a trusted party) and put into some allow list that the processing software uses.

> I guess this highlights that this is a securing spec issue, not a core data model issue?

It is true that some securing mechanisms will automatically detect problems for unexpected (or unknown) contexts, but for others, if all of the contexts are expected / well-known (and therefore vetted), it isn't a requirement to run any transforms during the consumption or verification process.

OR13:
[...]

PR #1281, opened 2023-09-13
Add normative requirements related to context processing

msporny:
This PR is an attempt at addressing #1185 by noting how the normative context is used and to make sure that if that optional processing is performed, if an error occurs, verification MUST fail.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1281.html" title="Last updated on Sep 26, 2023, 9:26 PM UTC (6247ecf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1281/d300869...6247ecf.html" title="Last updated on Sep 26, 2023, 9:26 PM UTC (6247ecf)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5804,6 +5804,17 @@ if a cryptographic hash value for a resource does not match the expected hash
 value.
         </p>
         <p>
+Implementations that apply the base context above, as well as other contexts
+and values in any `@context` property, during operations such as
+<a href="https://www.w3.org/TR/json-ld11-api/#expansion-algorithm">
+JSON-LD Expansion</a> or
+<a href="https://www.w3.org/TR/json-ld11/#serializing-deserializing-rdf">
+transformation to RDF</a> are expected to do so without experiencing any
+errors. If these operations are performed, and the result is an error, the
+<a>verifiable credential</a> or <a>verifiable presentation</a> MUST result in
+a verification failure.
+        </p>
+        <p>
 It is possible to confirm the SHA-384 digest above by running the following
 command from a modern Unix command interface line:
 `curl -s https://www.w3.org/ns/credentials/v2 | openssl dgst -sha384 -binary | openssl base64 -A`
--- a/index.html
+++ b/index.html
@@ -5810,9 +5810,9 @@ and values in any `@context` property, during operations such as
 JSON-LD Expansion</a> or
 <a href="https://www.w3.org/TR/json-ld11/#serializing-deserializing-rdf">
 transformation to RDF</a> are expected to do so without experiencing any
-errors. If these operations are performed, and the result is an error, the
-<a>verifiable credential</a> or <a>verifiable presentation</a> MUST result in
-a verification failure.
+errors. If operations of this type are performed, and the result is an error,
+the <a>verifiable credential</a> or <a>verifiable presentation</a> MUST result
+in a verification failure.
         </p>
         <p>
 It is possible to confirm the SHA-384 digest above by running the following
--- a/index.html
+++ b/index.html
@@ -5809,8 +5809,8 @@ and values in any `@context` property, during operations such as
 <a href="https://www.w3.org/TR/json-ld11-api/#expansion-algorithm">
 JSON-LD Expansion</a> or
 <a href="https://www.w3.org/TR/json-ld11/#serializing-deserializing-rdf">
-transformation to RDF</a> are expected to do so without experiencing any
-errors. If operations of this type are performed, and the result is an error,
+transformation to RDF</a>, are expected to do so without experiencing any
+errors. If such operations are performed and result in an error,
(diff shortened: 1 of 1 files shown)


Wind4Greg (approved):

iherman (approved):

decentralgabe (approved):

index.html
```diff
@@ -5804,6 +5804,17 @@ <h3>Base Context</h3>
 value.
         </p>
         <p>
+Implementations that apply the base context above, as well as other contexts
+and values in any `@context` property, during operations such as
+<a href="https://www.w3.org/TR/json-ld11-api/#expansion-algorithm">
+JSON-LD Expansion</a> or
+<a href="https://www.w3.org/TR/json-ld11/#serializing-deserializing-rdf">
+transformation to RDF</a> are expected to do so without experiencing any
```
TallTed, line 5812:
```suggestion
transformation to RDF</a>, are expected to do so without experiencing any
```

index.html
```diff
@@ -5804,6 +5804,17 @@ <h3>Base Context</h3>
...
+errors. If operations of this type are performed, and the result is an error,
```
TallTed, line 5813:
```suggestion
errors. If such operations are performed and result in an error,
```

TallTed (changes requested):

brentzundel (approved):

Sakurann (approved):

jandrieu (approved):
The language here is fine. It wasn't touching on the aspect I thought you brought up in your summary during the meting. 

Sorry for the confusion.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-09-14-vcwg#section1-11) on 2023-09-14

- no resolutions were taken

#### 1.11. Add normative requirements related to context processing (pr vc-data-model#1281)

_See github pull request [vc-data-model#1281](https://github.com/w3c/vc-data-model/pull/1281)._




**Manu Sporny:** Orie requested that, because context is normative, we explain what that means.  
� this PR says that if you utilize the context when doing full json-ld processing, and it results in an error, then all further operations should also result in an error.  
� In other words: if there is an error, you should bubble it up.  
� We know exactly what the contents of the context should be, so errors should be detected and raised across the stack.  

**Sebastian Crane:** From a security perspective you don't want to have errors leaking through. What is the difference between the context and any other validation?  

**Manu Sporny:** One example is the daterange validFrom and validUntil. E.g. a validFrom value that said use "validFrom tuesday". It has nothing to do with context. It's a non-context based error. Validation would fails.  
� We do have that throughout the spec. they have nothing to do with the context.  
� That's separate from the context file. It comes into play when doing full json-ld processing. But it's something that's optional.  
� not sure if that answers your question. Just noting that we have places in the spec where we expect errors to be raised.  

**Joe Andrieu:** I think your question is the difference between verification vs validation. I don't think you can verify with an invalid context.  

**Sebastian Crane:** Why is it so specific to the context file?  

**Manu Sporny:** I think you are correct in suggesting "any validation errors should be bubbled up".  
� Some orgs may keep processing the VC even if an error happens.  
� the reason this PR doesn't call this out is because this is only about what @context should have.  
� To Joe's point, the context comes into play when you're verifying, but only when you're doing full JSON-ld processing. When you're not doing it, you assume that the context is correct.  
� This means that you might miss point in time errors for a VC.  
� Validation is similar in this case.  

**Joe Andrieu:** I want to disagree when the processing you're suggesting to do when you're not doing json-ld processing.  
� My understanding was that you should fails if the context is something you don't expect to understand.  

**Manu Sporny:** the full json-ld processor will utilize the full extent of the @context file.  
� Without full processing, you presume that someone has done that elsewhere.  

**Joe Andrieu:** I understand that nuance. You still need to understand that context even if you aren't doing full processing.  

**Manu Sporny:** that does not happen. You presume it's valid.  

> *Joe Andrieu:* we are talking past each other.

> *Joe Andrieu:* I'm not talking about processing in a JSON-LD manner, I'm talking about processing of the context property. If MUST be checked to confirm that you think you understand it. Not that you apply those contexts.

> *Manu Sporny:* ah, yes ^.

**Manu Sporny:** It's happening in the web right now. the same file may yield different results depending on whether full processing is doing.  

**Brent Zundel:** the VCDM already states that context needs to have certain values.  

> *Joe Andrieu:* yep.

**Manu Sporny:** I think Joe, you're saying that the `@context` property is always checked, even if you're not looking as the contents of the dereferenced file (or all the values).  
� What changes do you think should be made, in order to address your concern?  

**Joe Andrieu:** Not sure yet, but I think we need to still look at the property. Will look at the PR and m ake suggestions.  

</details>

dlongley (approved):

msporny:
Normative, multiple reviews, changes requested and made, no objections merging.

PR #1283, opened 2023-09-15
Add normative text requiring protecting VCs and VPs

### LINKED ISSUE #1282: Normative requirement for securing a VC or VP
The current language in the section [Securing Verifiable Credentials](https://www.w3.org/TR/vc-data-model-2.0/#securing-verifiable-credentials) make it so that a VC or VP need not be secured. I think this is unintentional. 

We have normative text in the section https://www.w3.org/TR/vc-data-model-2.0/#conformance, but it doesn't use normative language. It does not require VC nor VPs to be secured. Specifically, we currently say

> Digital proof mechanisms, a subset of which are digital signatures, are required to ensure the protection of a [verifiable credential](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verifiable-credential). Having and validating proofs, which may be dependent on the syntax of the proof (for example, using the JSON Web Signature of a JSON Web Token for proofing a key holder), are an essential part of processing a [verifiable credential](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verifiable-credential).

They are required to ensure the protection, but protecting a VC or VP is not mandatory.

Do we want text like "verifiable credentials or verifiable presentations MUST be secured"? Or is my understanding incorrect?

andresuribe87:
Fixes #1282

***
<a href="https://pr-preview.s3.amazonaws.com/andresuribe87/vc-data-model/pull/1283.html" title="Last updated on Sep 26, 2023, 9:30 PM UTC (9087738)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1283/fc5a741...andresuribe87:9087738.html" title="Last updated on Sep 26, 2023, 9:30 PM UTC (9087738)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -583,7 +583,7 @@ be ignored by <a>issuers</a> and/or <a>verifiers</a>.
 
         <p>
 <a>Verifiable credential</a> and <a>verifiable presentation</a> MUST be protected using a digital proof
-mechanism such as digital signatures. Having and
+mechanism such as a digital signature. Having and
 verifying proofs, which may be dependent on the syntax of the proof
 (for example, using the JSON Web Signature of a JSON Web Token for proofing a
 key holder), are an essential part of processing <a>verifiable credentials</a> and <a>verifiable presentations</a>.
@@ -612,8 +612,8 @@ specification does not standardize on any single digital signature mechanism.
 One of the goals of this specification is to provide a data model that can be
 protected by a variety of current and future digital proof mechanisms.
 Conformance to this specification does not depend on the details of a particular
-proof mechanism; it requires clearly identifying the mechanisms used by
-<a>verifiable credentials</a> and <a>verifiable presentations</a>.
+proof mechanism; it requires clearly identifying the mechanism used by
+each <a>verifiable credential</a> and <a>verifiable presentation</a>.
         </p>
 
         <p>
@@ -1994,7 +1994,7 @@ It should be noted that these two classes of securing mechanisms are not
 mutually exclusive.
         </p>
         <p>
-<a>Verifiable credential</a> and <a>verifiable presentations</a> MUST be secured by at least
+<a>Verifiable credentials</a> and <a>verifiable presentations</a> MUST be secured by at least
 one securing method.
         </p>
         <p>
--- a/index.html
+++ b/index.html
@@ -584,9 +584,9 @@ be ignored by <a>issuers</a> and/or <a>verifiers</a>.
         <p>
 <a>Verifiable credential</a> and <a>verifiable presentation</a> MUST be protected using a digital proof
 mechanism such as digital signatures. Having and
-validating proofs, which may be dependent on the syntax of the proof
+verifying proofs, which may be dependent on the syntax of the proof
 (for example, using the JSON Web Signature of a JSON Web Token for proofing a
-key holder), are an essential part of processing a <a>verifiable credential</a> or a <a>verifiable presentation</a>.
+key holder), are an essential part of processing <a>verifiable credentials</a> and <a>verifiable presentations</a>.
 At the time of publication, Working Group members had implemented
 such protection using at least three proof mechanisms:
         </p>
@@ -612,8 +612,8 @@ specification does not standardize on any single digital signature mechanism.
 One of the goals of this specification is to provide a data model that can be
 protected by a variety of current and future digital proof mechanisms.
 Conformance to this specification does not depend on the details of a particular
-proof mechanism; it requires clearly identifying the mechanism a
-<a>verifiable credential</a> or a <a>verifiable presentation</a> uses.
+proof mechanism; it requires clearly identifying the mechanisms used by
+<a>verifiable credentials</a> and <a>verifiable presentations</a>.
         </p>
 
         <p>
@@ -1994,7 +1994,7 @@ It should be noted that these two classes of securing mechanisms are not
 mutually exclusive.
         </p>
         <p>
-A <a>verifiable credential</a> or a <a>verifiable presentation</a> MUST be secured by at least
+<a>Verifiable credential</a> and <a>verifiable presentations</a> MUST be secured by at least
 one securing method.
         </p>
         <p>
--- a/index.html
+++ b/index.html
@@ -582,13 +582,13 @@ be ignored by <a>issuers</a> and/or <a>verifiers</a>.
         </p>
 
         <p>
-Digital proof mechanisms, a subset of which are digital signatures, are required
-to ensure the protection of a <a>verifiable credential</a>. Having and
+<a>Verifiable credential</a> and <a>verifiable presentation</a> MUST be protected using a digital proof
+mechanism such as digital signatures. Having and
 validating proofs, which may be dependent on the syntax of the proof
 (for example, using the JSON Web Signature of a JSON Web Token for proofing a
-key holder), are an essential part of processing a <a>verifiable credential</a>.
+key holder), are an essential part of processing a <a>verifiable credential</a> or a <a>verifiable presentation</a>.
 At the time of publication, Working Group members had implemented
-<a>verifiable credentials</a> using at least three proof mechanisms:
+such protection using at least three proof mechanisms:
         </p>
 
         <ul>
@@ -613,7 +613,7 @@ One of the goals of this specification is to provide a data model that can be
 protected by a variety of current and future digital proof mechanisms.
 Conformance to this specification does not depend on the details of a particular
 proof mechanism; it requires clearly identifying the mechanism a
-<a>verifiable credential</a> uses.
+<a>verifiable credential</a> or a <a>verifiable presentation</a> uses.
         </p>
 
         <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -582,13 +582,13 @@ <h3>Use Cases and Requirements</h3>
...
 
         <p>
-Digital proof mechanisms, a subset of which are digital signatures, are required
-to ensure the protection of a <a>verifiable credential</a>. Having and
+<a>Verifiable credential</a> and <a>verifiable presentation</a> MUST be protected using a digital proof
+mechanism such as digital signatures. Having and
 validating proofs, which may be dependent on the syntax of the proof
 (for example, using the JSON Web Signature of a JSON Web Token for proofing a
-key holder), are an essential part of processing a <a>verifiable credential</a>.
+key holder), are an essential part of processing a <a>verifiable credential</a> or a <a>verifiable presentation</a>.
```
decentralgabe, line 589:
```suggestion
key holder), are an essential part of processing <a>verifiable credentials</a> and <a>verifiable presentations</a>.
```

index.html
```diff
@@ -582,13 +582,13 @@ <h3>Use Cases and Requirements</h3>
         </p>
 
         <p>
-Digital proof mechanisms, a subset of which are digital signatures, are required
-to ensure the protection of a <a>verifiable credential</a>. Having and
+<a>Verifiable credential</a> and <a>verifiable presentation</a> MUST be protected using a digital proof
+mechanism such as digital signatures. Having and
 validating proofs, which may be dependent on the syntax of the proof
```
jandrieu, line 587:
```suggestion
verifying proofs, which may be dependent on the syntax of the proof
```
I'd like to avoid the verification/validation confusion here. The English isn't *wrong* but proofs are used for verification, so that seems a less confusing word choice. Also, I know that wasn't part of this PR, but let's fix it.

index.html
```diff
@@ -613,7 +613,7 @@ <h3>Use Cases and Requirements</h3>
 protected by a variety of current and future digital proof mechanisms.
 Conformance to this specification does not depend on the details of a particular
 proof mechanism; it requires clearly identifying the mechanism a
-<a>verifiable credential</a> uses.
+<a>verifiable credential</a> or a <a>verifiable presentation</a> uses.
```
decentralgabe, line 616:
```suggestion
<a>verifiable credentials</a> and <a>verifiable presentations</a>.
```

index.html
```diff
@@ -613,7 +613,7 @@ <h3>Use Cases and Requirements</h3>
 protected by a variety of current and future digital proof mechanisms.
 Conformance to this specification does not depend on the details of a particular
 proof mechanism; it requires clearly identifying the mechanism a
```
decentralgabe, line 615:
```suggestion
proof mechanism; it requires clearly identifying the mechanisms used by
```

index.html
```diff
@@ -1994,6 +1994,10 @@ <h3>Securing Verifiable Credentials</h3>
 mutually exclusive.
         </p>
         <p>
+A <a>verifiable credential</a> or a <a>verifiable presentation</a> MUST be secured by at least
+one securing method.
+        </p>
+        <p>
```
decentralgabe, line 2000:
```suggestion<a>verifiable credentials</a> and <a>verifiable presentations</a> MUST be secured by at least
one securing method.
        </p>
        <p>
```

index.html
```diff
@@ -1994,6 +1994,10 @@ <h3>Securing Verifiable Credentials</h3>
 mutually exclusive.
         </p>
         <p>
+A <a>verifiable credential</a> or a <a>verifiable presentation</a> MUST be secured by at least
```
decentralgabe, line 1997:
```suggestion
<a>Verifiable credential</a> and <a>verifiable presentations</a> MUST be secured by at least
```

decentralgabe (approved):
great to have this

index.html
```diff
@@ -582,13 +582,13 @@ <h3>Use Cases and Requirements</h3>
         </p>
 
         <p>
-Digital proof mechanisms, a subset of which are digital signatures, are required
-to ensure the protection of a <a>verifiable credential</a>. Having and
-validating proofs, which may be dependent on the syntax of the proof
+<a>Verifiable credential</a> and <a>verifiable presentation</a> MUST be protected using a digital proof
+mechanism such as digital signatures. Having and
```
TallTed, line 586:
```suggestion
mechanism such as a digital signature. Having and
```

index.html
```diff
@@ -612,8 +612,8 @@ <h3>Use Cases and Requirements</h3>
 One of the goals of this specification is to provide a data model that can be
 protected by a variety of current and future digital proof mechanisms.
 Conformance to this specification does not depend on the details of a particular
-proof mechanism; it requires clearly identifying the mechanism a
-<a>verifiable credential</a> uses.
+proof mechanism; it requires clearly identifying the mechanisms used by
+<a>verifiable credentials</a> and <a>verifiable presentations</a>.
```
TallTed, line 616:
```suggestion
proof mechanism; it requires clearly identifying the mechanism used by
each <a>verifiable credential</a> and <a>verifiable presentation</a>.
```

index.html
```diff
@@ -1994,6 +1994,10 @@ <h3>Securing Verifiable Credentials</h3>
 mutually exclusive.
         </p>
         <p>
+<a>Verifiable credential</a> and <a>verifiable presentations</a> MUST be secured by at least
```
TallTed, line 1997:
```suggestion
<a>Verifiable credentials</a> and <a>verifiable presentations</a> MUST be secured by at least
```

TallTed (changes requested):

index.html
```diff
@@ -582,13 +582,13 @@ <h3>Use Cases and Requirements</h3>
         </p>
 
         <p>
-Digital proof mechanisms, a subset of which are digital signatures, are required
-to ensure the protection of a <a>verifiable credential</a>. Having and
-validating proofs, which may be dependent on the syntax of the proof
+<a>Verifiable credential</a> and <a>verifiable presentation</a> MUST be protected using a digital proof
```
msporny, line 585: This is creating a normative statement in an informative section, which we can't do. If we want to say this normatively, we'll have to say it in a normative section in the specification. You might see if some of this text can be moved here: https://w3c.github.io/vc-data-model/#securing-verifiable-credentials

msporny (changes requested):
There is one normative statement made that's in an informative section; that needs to be fixed. The rest of the PR looks good.

index.html
andresuribe87: Unless I'm missing something, this is inside the `conformance` section, which is marked as normative. See https://github.com/w3c/vc-data-model/blob/9087738b6a3249b6f3702a522d4aeb183685cda1/index.html#L556

TallTed: I see the problem... 

```
<section id="conformance" class="normative">
``` 
is parallel to (but has no `<h3>`!) 
```
      <section class="informative">
        <h3>Use Cases and Requirements</h3>
```
and both are within
```
    <section class="informative">
      <h2>Introduction</h2>
```
Really, all the 1.x sub-sections should NOT be labeled as non-normative, because they inherit this from 1.0, and there REALLY should be a way to label a section as normative!

Also, the normative `Conformance` section REALLY cannot be within the informative `Introduction`, so the whole section needs to move, not just the sentence seen above on line 585.

andresuribe87: @TallTed thanks for the explanation.

I'm happy to move sections to more appropriate places, even though that's separate from the scope of the changes intended to be done with this PR. 

Regarding

> Really, all the 1.x sub-sections should NOT be labeled as non-normative, because they inherit this from 1.0, and there REALLY should be a way to label a section as normative!

Labelling the `Introduction` section as normative seems to be what you're suggesting. If not, can you clarify? 


Re:
> Also, the normative Conformance section REALLY cannot be within the informative Introduction, so the whole section needs to move, not just the sentence seen above on line 585.

This depends on the previous point. Assuming that we keep the `Introduction` section as normative, is creating a new entire section appropriate? Where? I fear that moving the whole Conformace section would break the flow of the document, and will need a lot of editorial love.

TallTed: >> [@TallTed] Really, all the 1.x sub-sections should NOT be labeled as non-normative, because they inherit this from 1.0, and there REALLY should be a way to label a section as normative!
>
> [@andresuribe87] Labelling the Introduction section as normative seems to be what you're suggesting. If not, can you clarify?

I wasn't suggesting that, but it's the only way to make the current `Conformance` sub-section normative within the (currently non-normative) `Introduction`, because normative sub-sections are not allowed within non-normative sections.

@iherman may be able to tell us whether `Introduction` can be made normative, such that `Conformance` may remain where it is *and* be made normative, given that all the *other* sub-sections of `Introduction` are already labeled as non-normative.

If `Introduction` must remain non-normative, @msporny can probably point out a good place to move `Conformance` to, such that it can be normative.

msporny: > Also, the normative Conformance section REALLY cannot be within the informative Introduction, so the whole section needs to move, not just the sentence seen above on line 585.

Err, hold up -- you can have a normative section buried deep in a non-normative section; remember, we got all the way through two RECs in this state. That said, we could also move Conformance up to it's own section... I think I moved it initially because having a top-level section, just for conformance classes, felt like it was a bit much.

I'd prefer to leave the conformance section as it is... which brings me back around to why I objected to this change... @andresuribe87 is indeed correct, the normative statement is added in the conformance section, which is normative, so his initial text is fine and I'm retracting my change request and approving the PR for merge.

msporny (approved):

msporny (approved):

msporny:

PR #1302, opened 2023-10-04
Propose better JSON-LD processing text

OR13:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1302.html" title="Last updated on Nov 29, 2023, 4:29 PM UTC (9959a04)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1302/286c2b2...9959a04.html" title="Last updated on Nov 29, 2023, 4:29 PM UTC (9959a04)">Diff</a>

### DIFF
--- a/common.js
+++ b/common.js
@@ -93,6 +93,14 @@ var vcwg = {
       status:   'CG-DRAFT',
       publisher:  'Credentials W3C Community Group'
     },
+    // todo: replace all references to RDF-NORMALIZATION, with this.
+    'RDF-CANON': {
+      title: 'RDF Dataset Canonicalization',
+      href: 'https://www.w3.org/TR/rdf-canon/',
+      authors: ['Dave Longley', 'Gregg Kellogg', 'Dan Yamamoto'],
+      status:   'WG-DRAFT',
+      publisher:  'RDF Dataset Canonicalization and Hash Working Group'
+    },
     'DEMOGRAPHICS': {
       title: 'Simple Demographics Often Identify People Uniquely',
       href: 'https://dataprivacylab.org/projects/identifiability/paper1.pdf',
--- a/common.js
+++ b/common.js
@@ -93,14 +93,6 @@ var vcwg = {
       status:   'CG-DRAFT',
       publisher:  'Credentials W3C Community Group'
     },
-    // todo: replace all references to RDF-NORMALIZATION, with this.
-    'RDF-CANON': {
-      title: 'RDF Dataset Canonicalization',
-      href: 'https://www.w3.org/TR/rdf-canon/',
-      authors: ['Dave Longley', 'Gregg Kellogg', 'Dan Yamamoto'],
-      status:   'WG-DRAFT',
-      publisher:  'RDF Dataset Canonicalization and Hash Working Group'
-    },
     'DEMOGRAPHICS': {
       title: 'Simple Demographics Often Identify People Uniquely',
       href: 'https://dataprivacylab.org/projects/identifiability/paper1.pdf',
--- a/index.html
+++ b/index.html
@@ -4326,7 +4326,7 @@ lightweight manner, without having to use RDF processing libraries.
         <section>
           <h2>Advanced RDF Processing</h2>
           <p>
-Issuers should be aware that while conforming documents are expressed as compact JSON-LD,
+Issuers are advised that while conforming documents are expressed as compact JSON-LD,
 not all Holders or Verifiers will understand expressions of intention that are only visible 
 after RDF processing has occured.
           </p>
--- a/index.html
+++ b/index.html
@@ -4371,9 +4371,13 @@ graphs; in particular, these are used to separate the graphs related to
 <a href="#dfn-embedded-proof">embedded proofs</a>.
           </p>
           <p>
-Issuers and Holders might apply these same techniques to unambigiously communicate their intention regarding structured claims in 
-<a>verifiable credentials</a> and <a>verifiable presentations</a>. Verifiers that do not understand or process JSON-LD, will not 
-be aware of these differences, and confusion over the intention of the issuer and the holder could lead to unexpected processing behavior by verifiers.
+<a>Issuers</a> and <a>holders</a> might apply these same techniques to
+unambiguously communicate their intention regarding structured claims in
+<a>verifiable credentials</a> and <a>verifiable presentations</a>.
+<a>Verifiers</a> that do not understand or process JSON-LD will not
+be aware of these differences, and confusion between the intentions of the
+<a>issuer</a> and the <a>holder</a> could lead to unexpected processing
+behavior by <a>verifiers</a>.
           </p>
         </section>
       </section>
--- a/index.html
+++ b/index.html
@@ -4321,9 +4321,10 @@ implementing the rule above. This can ensure proper term identification,
 typing, and order, when a JSON document is processed as JSON-LD.
           </p>
           <p>
-However, the intention of the issuer is not preserved, unless the <code>@context</code>
-is applied, and the resulting <a href="#verifiable-credential-graphs">Verifiable Credential Graphs</a>,
-and their exact RDF types, and node and edge structure is understood by the verifier.
+However, the intention of the issuer is not preserved, unless the rules expressed
+in the <code>@context</code> are followed, and the resulting <a href="#verifiable-credential-graphs">Verifiable Credential Graphs</a>, and their exact RDF types, and node and edge structure is understood by the verifier. By following the rules laid out in this section, applications can be hard-coded against
+specific contexts, just as they are commonly hard-coded to consume other
+application-specific JSON formats.
           </p>
 
           <p>
--- a/index.html
+++ b/index.html
@@ -4344,7 +4344,7 @@ should review <a href="https://www.w3.org/TR/json-ld11/#node-identifiers">Node I
           </p>
           <p>
 Implementers should be aware that object member, and array element order are not preserved by default in JSON-LD,
-and as such, Issuers intending to communicate array order need to leverage the langauge features of JSON-LD.
+and as such, Issuers intending to communicate array order need to leverage the language features of JSON-LD.
 See <a href="https://www.w3.org/TR/json-ld11/#sets-and-lists">Value Ordering</a>.
           </p>
           <p>
--- a/index.html
+++ b/index.html
@@ -4314,7 +4314,7 @@ and their exact RDF types, and node and edge structure is understood by the veri
 These rules alone do not guarantee semantic interoperability between JSON and JSON-LD for
 literal JSON keys mapped to URIs by the `@context` mechanism.
           </p>
-While some RDF processors will use the langauge features of JSON-LD and can verify that all
+While some RDF processors will use the language features of JSON-LD and can verify that all
 terms are correctly specified, JSON-LD processors implicitly accept the same
 semantics without performing any RDF transformations, but instead by
 applying the above rules. In other words, the context in which the data exchange
--- a/index.html
+++ b/index.html
@@ -4349,7 +4349,7 @@ See <a href="https://www.w3.org/TR/json-ld11/#sets-and-lists">Value Ordering</a>
           </p>
           <p>
 Implementers should be aware that while naive JSON interpretation of claims might make it appear that claims are related,
-the <code>https://www.w3.org/ns/credentials/v2</code> leverages <code>@container</code> and <code>@graph</code> to seperate information graphs.
+the <code>https://www.w3.org/ns/credentials/v2</code> leverages <code>@container</code> and <code>@graph</code> to separate information graphs.
 In particular, to seperate the graphs related to <a>verifiable credentials</a>, <a>verifiable presentations</a>
 and <code>data integrity proofs</code>.
           </p>
--- a/index.html
+++ b/index.html
@@ -4350,7 +4350,7 @@ See <a href="https://www.w3.org/TR/json-ld11/#sets-and-lists">Value Ordering</a>
           <p>
 Implementers should be aware that while naive JSON interpretation of claims might make it appear that claims are related,
 the <code>https://www.w3.org/ns/credentials/v2</code> leverages <code>@container</code> and <code>@graph</code> to separate information graphs.
-In particular, to seperate the graphs related to <a>verifiable credentials</a>, <a>verifiable presentations</a>
+In particular, to separate the graphs related to <a>verifiable credentials</a>, <a>verifiable presentations</a>
 and <code>data integrity proofs</code>.
           </p>
           <p>
--- a/index.html
+++ b/index.html
@@ -4356,7 +4356,7 @@ Implementers are advised that <code>id</code> and <code>type</code> have been
 aliased to <code>@id</code> and <code>@type</code>, and are marked <code>@protected</code>.
 Depending on the software libraries used, attempts to redefine these terms might raise processing errors.
 Implementers interested in understanding what <code>@id</code> means 
-should review <a href="https://www.w3.org/TR/json-ld11/#node-identifiers">Node Identifiers</a>, and
+are advised to review <a href="https://www.w3.org/TR/json-ld11/#node-identifiers">Node Identifiers</a>, and
 <a href="https://www.w3.org/TR/json-ld11/#specifying-the-type">Specifying the Type</a>.
           </p>
           <p>
--- a/index.html
+++ b/index.html
@@ -4335,7 +4335,7 @@ Some of the best features of JSON-LD are only possible by applying its language 
 terms defined in the <code>@context</code>.
           </p>
           <p>
-Implementers should be aware that <code>id</code> and <code>type</code> have been
+Implementers are advised that <code>id</code> and <code>type</code> have been
 aliased to <code>@id</code> and <code>@type</code>, and are marked <code>@protected</code>.
 Depending on the software libraries used, attempts to redefine these terms may raise processing errors.
 Implementers interested in understanding what <code>@id</code> means 
--- a/index.html
+++ b/index.html
@@ -4246,7 +4246,24 @@ While the media types describing conforming documents defined in this
 specification always express JSON-LD, RDF processing is not required to be
 performed, since JSON-LD is a concrete RDF syntax as described in [RDF11-CONCEPTS]. 
 Hence, a JSON-LD document is both an RDF document and a JSON document and correspondingly represents an instance of an RDF data model.
+          </p>
+           <p>
+RDF is a technology for modeling graphs of statements. Each statement is a single subject-property-value
+relationship, which is referred to as a <a>claim</a> in this specification. JSON-LD is a technology that
+enables the expression of RDF using idiomatic JSON, enabling developers familiar with JSON to write
+applications to consume RDF as JSON. In general, subjects are expressed as JSON objects with each property and value of the subject as a JSON key and value, respectively. A special affordance is made to express an identifier of a subject, if present, using the <code>@id</code> JSON key, which is aliased in this specification to <code>id</code>.
 See <a href="JSON-LD11##relationship-to-rdf">Relationship to RDF</a> for more details.
+          </p>
+           <p>
+As JSON can be used to express all different kinds of information, a consumer of a particular JSON
+document can only properly interpret the authors intent if they possess information that contextualizes
+and disambiguates it from other possible expressions. Information to assist with this interpretation can either 
+be wholly external to the JSON document or linked from within it. Compact JSON-LD documents include a <code>@context</code> property that internally expresses or links to contextual information and, per the JSON-LD specification, these documents follow object-oriented organizational principles to express <a>claims</a>. These features enable generalized processors to be written to convert JSON-LD documents from one context to another, but this is not needed when consumers receive JSON-LD documents that already use the context and shape that they expect. Authors of JSON-LD documents, such as <a>issuers</a> of <a>verifiable credentials</a>, are required to provide proper JSON-LD contexts and follow these rules in order to facilitate interoperability.
+          </p>
+           <p>
+The text below helps consumers understand how to ensure a JSON-LD document is expressed in a context
+and shape that their application already understands such that they do not need to transform it in
+order to consume its contents. Notably, this does not mean that consumers do not need to understand any context at all, rather, consuming applications only need to understand a chosen set of contexts and document shapes to work with and not others. Issuers can publish contexts and information about their <a>verifiable credentials</a> to aid consumers who do not use generalized processors, just like would be done with any other JSON-formatted data.
           </p>
           <p>
 Processing <a>verifiable credentials</a> or <a>verifiable presentations</a>
--- a/index.html
+++ b/index.html
@@ -4365,10 +4365,10 @@ and as such, Issuers intending to communicate array order need to leverage the l
 See <a href="https://www.w3.org/TR/json-ld11/#sets-and-lists">Value Ordering</a>.
           </p>
           <p>
-Implementers should be aware that while naive JSON interpretation of claims might make it appear that claims are related,
+Implementers are advised that while minimal processing of a document make it appear that claims are related,
 the <code>https://www.w3.org/ns/credentials/v2</code> leverages <code>@container</code> and <code>@graph</code> to separate information graphs.
-In particular, to separate the graphs related to <a>verifiable credentials</a>, <a>verifiable presentations</a>
-and <code>data integrity proofs</code>.
+In particular, to separate the graphs related to <a>verifiable credentials</a>, <a>verifiable presentations</a>,
+and <a href="#dfn-embedded-proof">embedded proofs</a>.
           </p>
           <p>
 Issuers and Holders might apply these same techniques to unambigiously communicate their intention regarding structured claims in 
--- a/index.html
+++ b/index.html
@@ -4240,47 +4240,54 @@ content type, or return an error code such as
         </section>
 
         <section class="informative">
-          <h2>JSON Processing</h2>
-
+          <h2>JSON-LD Processing</h2>
           <p>
 While the media types describing conforming documents defined in this
-specification always express JSON-LD, JSON-LD processing is not required to be
-performed, since JSON-LD is JSON. Some scenarios where processing a
-<a>verifiable credential</a> or a <a>verifiable presentation</a> as JSON is
+specification always express JSON-LD, RDF processing is not required to be
+performed, since JSON-LD is a concrete RDF syntax as described in [RDF11-CONCEPTS]. 
+Hence, a JSON-LD document is both an RDF document and a JSON document and correspondingly represents an instance of an RDF data model.
+See <a href="JSON-LD11##relationship-to-rdf">Relationship to RDF</a> for more details.
+          </p>
+          <p>
+Some scenarios where processing a <a>verifiable credential</a> or a <a>verifiable presentation</a> as RDF is
 desirable include, but are not limited to:
           </p>
-
           <ul>
             <li>
-Before securing or after verifying content
-that requires <a href="https://csrc.nist.gov/glossary/term/data_integrity">data
-integrity</a>, such as a
-<a>verifiable credential</a> or <a>verifiable presentation</a>.
+Before securing or after verifying <a>verifiable credentials</a> or <a>verifiable presentations</a>,
+in cases where the issuer has leveraged the <code>@context</code> to provide details regarding 
+their intended expression of the <a href="#verifiable-credential-graphs">Verifiable Credential Graphs</a>.
             </li>
             <li>
-When performing JSON Schema validation, as described in Section
-<a href="#data-schemas"></a>.
+After performing JSON Schema validation, as described in Section
+<a href="#data-schemas"></a>, in order to mitigate attacks 
+on the canonicalization process as described in <a href="https://w3c.github.io/rdf-canon/spec/#dataset-poisoning">Dataset Poisoning</a>
             </li>
             <li>
 When serializing or deserializing <a>verifiable credentials</a> or
-<a>verifiable presentations</a> into systems that store or index their contents.
+<a>verifiable presentations</a> into systems that store or index claims about subject in
+<code>text/turtle</code>, <code>application/n-quads</code>, <code>application/ld+json</code>. 
             </li>
             <li>
-When operating on <a>verifiable credentials</a> or <a>verifiable
-presentations</a> in a software application, after verification or validation
-is performed for securing mechanisms that require an understanding of
-and/or processing of JSON-LD.
+When validating <a>verifiable credentials</a> or <a>verifiable
+presentations</a> in a software application, after verification 
+is performed for securing mechanisms that do not require an understanding
+or processing of JSON-LD.
             </li>
             <li>
-When an application chooses to process the media type using the `+json`
+When an application chooses to process the media type using the `+ld+json`
 structured media type suffix.
             </li>
           </ul>
 
           <p>
-That is, JSON processing is allowed as long as the document being consumed or
-produced is a <a>conforming document</a>. If JSON processing is desired, an
-implementer is advised to follow the following rule:
+Naive JSON processing is allowed as long as the document being consumed or
+produced is a <a>conforming document</a>, however, this can lead to interoperability issues 
+and confusion about the specific information that has been secured between JSON-LD and RDF processors.
+          </p>
+
+          <p>
+Implementers unfamilar with JSON-LD / RDF are advised:
           </p>
 
           <ul>
@@ -4297,19 +4304,60 @@ Using static context files with a JSON Schema is one acceptable approach to
 implementing the rule above. This can ensure proper term identification,
 typing, and order, when a JSON document is processed as JSON-LD.
           </p>
-
           <p>
-The rule above guarantees semantic interoperability between JSON and JSON-LD for
-literal JSON keys mapped to URIs by the `@context` mechanism. While JSON-LD
-processors will use the specific mechanism provided and can verify that all
-terms are correctly specified, JSON-based processors implicitly accept the same
-semantics without performing any JSON-LD transformations, but instead by
-applying the above rules. In other words, the context in which the data exchange
-happens is explicitly stated for both JSON and JSON-LD by using the same
-mechanism. With respect to JSON-based processors, this is achieved in a
-lightweight manner, without having to use JSON-LD processing libraries.
+However, the intention of the issuer is not preserved, unless the <code>@context</code>
+is applied, and the resulting <a href="#verifiable-credential-graphs">Verifiable Credential Graphs</a>,
+and their exact RDF types, and node and edge structure is understood by the verifier.
           </p>
 
+          <p>
+These rules alone do not guarantee semantic interoperability between JSON and JSON-LD for
+literal JSON keys mapped to URIs by the `@context` mechanism.
+          </p>
+While some RDF processors will use the langauge features of JSON-LD and can verify that all
+terms are correctly specified, JSON-LD processors implicitly accept the same
+semantics without performing any RDF transformations, but instead by
+applying the above rules. In other words, the context in which the data exchange
+happens is explicitly stated for both RDF and JSON-LD by using the same
+mechanism. With respect to JSON-LD-based processors, this is achieved in a
+lightweight manner, without having to use RDF processing libraries.
+          </p>
+        </section>
+        <section>
+          <h2>Advanced RDF Processing</h2>
+          <p>
+Issuers should be aware that while conforming documents are expressed as compact JSON-LD,
+not all Holders or Verifiers will understand expressions of intention that are only visible 
+after RDF processing has occured.
+          </p>
+          <p>
+Some of the best features of JSON-LD are only possible by applying its langauge features to the 
+terms defined in the <code>@context</code>.
+          </p>
+          <p>
+Implementers should be aware that <code>id</code> and <code>type</code> have been
+aliased to <code>@id</code> and <code>@type</code>, and are marked <code>@protected</code>.
+Depending on the software libraries used, attempts to redefine these terms may raise processing errors.
+Implementers interested in the details being what <code>@id</code> means 
+should review <a href="https://www.w3.org/TR/json-ld11/#node-identifiers">Node Identifiers</a>, and
+<a href="https://www.w3.org/TR/json-ld11/#specifying-the-type">Specifying the Type</a>.
+          </p>
+          <p>
+Implementers should be aware that object member, and array element order are not preserved by default in JSON-LD,
+and as such, Issuers intending to communicate array order need to leverage the langauge features of JSON-LD.
+See <a href="https://www.w3.org/TR/json-ld11/#sets-and-lists">Value Ordering</a>.
+          </p>
+          <p>
+Implementers should be aware that while naive JSON interpretation of claims might make it appear that claims are related,
+the <code>https://www.w3.org/ns/credentials/v2</code> leverages <code>@container</code> and <code>@graph</code> to seperate information graphs.
+In particular, to seperate the graphs related to <a>verifiable credentials</a>, <a>verifiable presentations</a>
+and <code>data integrity proofs</code>.
+          </p>
+          <p>
+Issuers and Holders might apply these same techniques to unambigiously communicate their intention regarding structured claims in 
+<a>verifiable credentials</a> and <a>verifiable presentations</a>. Verifiers that do not understand or process JSON-LD, will not 
+be aware of these differences, and confusion over the intention of the issuer and the holder could lead to unexpected processing behavior by verifiers.
+          </p>
         </section>
       </section>
 
--- a/index.html
+++ b/index.html
@@ -4296,7 +4296,7 @@ structured media type suffix.
           </ul>
 
           <p>
-Naive JSON processing is allowed as long as the document being consumed or
+Instead of performing general JSON-LD processing, application-specific processing is allowed as long as the document being consumed or
 produced is a <a>conforming document</a>, however, this can lead to interoperability issues 
 and confusion about the specific information that has been secured between JSON-LD and RDF processors.
           </p>
--- a/index.html
+++ b/index.html
@@ -4356,7 +4356,9 @@ are advised to review <a href="https://www.w3.org/TR/json-ld11/#node-identifiers
           </p>
           <p>
 Implementers are advised that object member, and array element order are not preserved by default in JSON-LD,
-and as such, Issuers intending to communicate array order need to leverage the language features of JSON-LD.
+therefore, <a>issuers</a> intending to communicate the orders of object
+members and/or array elements might need to leverage language and/or other
+advanced features of JSON-LD.
 See <a href="https://www.w3.org/TR/json-ld11/#sets-and-lists">Value Ordering</a>.
           </p>
           <p>
--- a/index.html
+++ b/index.html
@@ -4271,8 +4271,7 @@ Turtle [[?TURTLE]], N-Quads [[?N-QUADS]], or JSON-LD [[?JSON-LD11]].
             <li>
 When consuming <a>verifiable credentials</a> or <a>verifiable
 presentations</a> in a software application, after verification 
-is performed for securing mechanisms that do not require an understanding
-or processing of JSON-LD.
+is performed for securing mechanisms that do not require JSON-LD transformations.
             </li>
             <li>
 When an application chooses to process the media type using the `+ld+json`
--- a/index.html
+++ b/index.html
@@ -4240,7 +4240,7 @@ content type, or return an error code such as
         </section>
 
         <section class="informative">
-          <h2>JSON-LD Processing</h2>
+          <h2>Processing JSON-LD</h2>
           <p>
 While the media types describing conforming documents defined in this
 specification always express JSON-LD, RDF processing is not required to be
@@ -4249,8 +4249,8 @@ Hence, a JSON-LD document is both an RDF document and a JSON document and corres
 See <a href="JSON-LD11##relationship-to-rdf">Relationship to RDF</a> for more details.
           </p>
           <p>
-Some scenarios where processing a <a>verifiable credential</a> or a <a>verifiable presentation</a> as RDF is
-desirable include, but are not limited to:
+Processing <a>verifiable credentials</a> or <a>verifiable presentations</a>
+by transforming JSON-LD to RDF might be desirable under the following conditions:
           </p>
           <ul>
             <li>
@@ -4323,7 +4323,7 @@ lightweight manner, without having to use RDF processing libraries.
           </p>
         </section>
         <section>
-          <h2>Advanced RDF Processing</h2>
+          <h2>Processing RDF</h2>
           <p>
 Issuers are advised that while conforming documents are expressed as compact JSON-LD,
 not all Holders or Verifiers will understand expressions of intention that are only visible 
--- a/index.html
+++ b/index.html
@@ -4302,8 +4302,7 @@ produced is a <a>conforming document</a>, however, this can lead to interoperabi
 and confusion about the specific information that has been secured between JSON-LD and RDF processors.
           </p>
 
-          <p>
-Implementers unfamilar with JSON-LD / RDF are advised:
+         <h3>Application-specific Processing</h3>
           </p>
 
           <ul>
--- a/index.html
+++ b/index.html
@@ -4255,7 +4255,7 @@ desirable include, but are not limited to:
           <ul>
             <li>
 Before securing or after verifying <a>verifiable credentials</a> or <a>verifiable presentations</a>,
-in cases where the issuer has leveraged the <code>@context</code> to provide details regarding 
+in cases where the issuer made use of the underlying RDF Dataset model to provide details regarding 
 their intended expression of the <a href="#verifiable-credential-graphs">Verifiable Credential Graphs</a>.
             </li>
             <li>
--- a/index.html
+++ b/index.html
@@ -4277,8 +4277,7 @@ their intended expression of the <a href="#verifiable-credential-graphs">Verifia
             </li>
             <li>
 After performing JSON Schema validation, as described in Section
-<a href="#data-schemas"></a>, in order to mitigate attacks 
-on the canonicalization process as described in <a href="https://w3c.github.io/rdf-canon/spec/#dataset-poisoning">Dataset Poisoning</a>
+<a href="#data-schemas"></a>.
             </li>
             <li>
 When serializing or deserializing <a>verifiable credentials</a> or
--- a/index.html
+++ b/index.html
@@ -4265,8 +4265,8 @@ on the canonicalization process as described in <a href="https://w3c.github.io/r
             </li>
             <li>
 When serializing or deserializing <a>verifiable credentials</a> or
-<a>verifiable presentations</a> into systems that store or index claims about subject in
-<code>text/turtle</code>, <code>application/n-quads</code>, <code>application/ld+json</code>. 
+<a>verifiable presentations</a> into systems that store or index data in 
+Turtle [[?TURTLE]], N-Quads [[?N-QUADS]], or JSON-LD [[?JSON-LD11]].
             </li>
             <li>
 When validating <a>verifiable credentials</a> or <a>verifiable
--- a/index.html
+++ b/index.html
@@ -4362,10 +4362,13 @@ advanced features of JSON-LD.
 See <a href="https://www.w3.org/TR/json-ld11/#sets-and-lists">Value Ordering</a>.
           </p>
           <p>
-Implementers are advised that while minimal processing of a document make it appear that claims are related,
-the <code>https://www.w3.org/ns/credentials/v2</code> leverages <code>@container</code> and <code>@graph</code> to separate information graphs.
-In particular, to separate the graphs related to <a>verifiable credentials</a>, <a>verifiable presentations</a>,
-and <a href="#dfn-embedded-proof">embedded proofs</a>.
+Implementers are advised that while minimal processing of a document might
+make it appear that claims are related,
+the <code>https://www.w3.org/ns/credentials/v2</code> `@context` leverages
+<code>@container</code> and <code>@graph</code> to separate information
+graphs; in particular, these are used to separate the graphs related to
+<a>verifiable credentials</a>, <a>verifiable presentations</a>, and
+<a href="#dfn-embedded-proof">embedded proofs</a>.
           </p>
           <p>
 Issuers and Holders might apply these same techniques to unambigiously communicate their intention regarding structured claims in 
--- a/index.html
+++ b/index.html
@@ -4331,7 +4331,7 @@ not all Holders or Verifiers will understand expressions of intention that are o
 after RDF processing has occured.
           </p>
           <p>
-Some of the best features of JSON-LD are only possible by applying its langauge features to the 
+Some of the best features of JSON-LD are only possible by applying its language features to the 
 terms defined in the <code>@context</code>.
           </p>
           <p>
--- a/index.html
+++ b/index.html
@@ -4297,8 +4297,7 @@ structured media type suffix.
 
           <p>
 Instead of performing general JSON-LD processing, application-specific processing is allowed as long as the document being consumed or
-produced is a <a>conforming document</a>, however, this can lead to interoperability issues 
-and confusion about the specific information that has been secured between JSON-LD and RDF processors.
+produced is a <a>conforming document</a>.
           </p>
 
          <h3>Application-specific Processing</h3>
--- a/index.html
+++ b/index.html
@@ -4338,7 +4338,7 @@ terms defined in the <code>@context</code>.
 Implementers should be aware that <code>id</code> and <code>type</code> have been
 aliased to <code>@id</code> and <code>@type</code>, and are marked <code>@protected</code>.
 Depending on the software libraries used, attempts to redefine these terms may raise processing errors.
-Implementers interested in the details being what <code>@id</code> means 
+Implementers interested in understanding what <code>@id</code> means 
 should review <a href="https://www.w3.org/TR/json-ld11/#node-identifiers">Node Identifiers</a>, and
 <a href="https://www.w3.org/TR/json-ld11/#specifying-the-type">Specifying the Type</a>.
           </p>
--- a/index.html
+++ b/index.html
@@ -4325,8 +4325,6 @@ application-specific JSON formats.
           </p>
 
           <p>
-These rules alone do not guarantee semantic interoperability between JSON and JSON-LD for
-literal JSON keys mapped to URIs by the `@context` mechanism.
           </p>
 While some RDF processors will use the language features of JSON-LD and can verify that all
 terms are correctly specified, JSON-LD processors explicitly accept the same
--- a/index.html
+++ b/index.html
@@ -4355,7 +4355,7 @@ are advised to review <a href="https://www.w3.org/TR/json-ld11/#node-identifiers
 <a href="https://www.w3.org/TR/json-ld11/#specifying-the-type">Specifying the Type</a>.
           </p>
           <p>
-Implementers should be aware that object member, and array element order are not preserved by default in JSON-LD,
+Implementers are advised that object member, and array element order are not preserved by default in JSON-LD,
 and as such, Issuers intending to communicate array order need to leverage the language features of JSON-LD.
 See <a href="https://www.w3.org/TR/json-ld11/#sets-and-lists">Value Ordering</a>.
           </p>
--- a/index.html
+++ b/index.html
@@ -4315,7 +4315,7 @@ These rules alone do not guarantee semantic interoperability between JSON and JS
 literal JSON keys mapped to URIs by the `@context` mechanism.
           </p>
 While some RDF processors will use the language features of JSON-LD and can verify that all
-terms are correctly specified, JSON-LD processors implicitly accept the same
+terms are correctly specified, JSON-LD processors explicitly accept the same
 semantics without performing any RDF transformations, but instead by
 applying the above rules. In other words, the context in which the data exchange
 happens is explicitly stated for both RDF and JSON-LD by using the same
--- a/index.html
+++ b/index.html
@@ -4269,7 +4269,7 @@ When serializing or deserializing <a>verifiable credentials</a> or
 Turtle [[?TURTLE]], N-Quads [[?N-QUADS]], or JSON-LD [[?JSON-LD11]].
             </li>
             <li>
-When validating <a>verifiable credentials</a> or <a>verifiable
+When consuming <a>verifiable credentials</a> or <a>verifiable
 presentations</a> in a software application, after verification 
 is performed for securing mechanisms that do not require an understanding
 or processing of JSON-LD.
--- a/index.html
+++ b/index.html
@@ -4337,7 +4337,7 @@ terms defined in the <code>@context</code>.
           <p>
 Implementers are advised that <code>id</code> and <code>type</code> have been
 aliased to <code>@id</code> and <code>@type</code>, and are marked <code>@protected</code>.
-Depending on the software libraries used, attempts to redefine these terms may raise processing errors.
+Depending on the software libraries used, attempts to redefine these terms might raise processing errors.
 Implementers interested in understanding what <code>@id</code> means 
 should review <a href="https://www.w3.org/TR/json-ld11/#node-identifiers">Node Identifiers</a>, and
 <a href="https://www.w3.org/TR/json-ld11/#specifying-the-type">Specifying the Type</a>.


decentralgabe (approved):
nice and clear

index.html
```diff
@@ -4297,19 +4304,60 @@ <h2>JSON Processing</h2>
...
           <p>
-The rule above guarantees semantic interoperability between JSON and JSON-LD for
-literal JSON keys mapped to URIs by the `@context` mechanism. While JSON-LD
-processors will use the specific mechanism provided and can verify that all
-terms are correctly specified, JSON-based processors implicitly accept the same
-semantics without performing any JSON-LD transformations, but instead by
+These rules alone do not guarantee semantic interoperability between JSON and JSON-LD for
+literal JSON keys mapped to URIs by the `@context` mechanism.
+          </p>
+While some RDF processors will use the langauge features of JSON-LD and can verify that all
```
brianorwhatever, line 4317:
```suggestion
While some RDF processors will use the language features of JSON-LD and can verify that all
```

index.html
```diff
@@ -4297,19 +4304,60 @@ <h2>JSON Processing</h2>
...
+        </section>
+        <section>
+          <h2>Advanced RDF Processing</h2>
+          <p>
+Issuers should be aware that while conforming documents are expressed as compact JSON-LD,
+not all Holders or Verifiers will understand expressions of intention that are only visible 
+after RDF processing has occured.
+          </p>
+          <p>
+Some of the best features of JSON-LD are only possible by applying its langauge features to the
```
brianorwhatever, line 4334:
```suggestion
Some of the best features of JSON-LD are only possible by applying its language features to the 
```

index.html
```diff
@@ -4297,19 +4304,60 @@ <h2>JSON Processing</h2>
...
+<a href="https://www.w3.org/TR/json-ld11/#specifying-the-type">Specifying the Type</a>.
+          </p>
+          <p>
+Implementers should be aware that object member, and array element order are not preserved by default in JSON-LD,
+and as such, Issuers intending to communicate array order need to leverage the langauge features of JSON-LD.
+See <a href="https://www.w3.org/TR/json-ld11/#sets-and-lists">Value Ordering</a>.
+          </p>
+          <p>
+Implementers should be aware that while naive JSON interpretation of claims might make it appear that claims are related,
+the <code>https://www.w3.org/ns/credentials/v2</code> leverages <code>@container</code> and <code>@graph</code> to seperate information graphs.
```
brianorwhatever, line 4352:
```suggestion
the <code>https://www.w3.org/ns/credentials/v2</code> leverages <code>@container</code> and <code>@graph</code> to separate information graphs.
```

index.html
```diff
@@ -4297,19 +4304,60 @@ <h2>JSON Processing</h2>
...
+In particular, to seperate the graphs related to <a>verifiable credentials</a>, <a>verifiable presentations</a>
```
brianorwhatever, line 4353:
```suggestion
In particular, to separate the graphs related to <a>verifiable credentials</a>, <a>verifiable presentations</a>
```

index.html
```diff
@@ -4297,19 +4304,60 @@ <h2>JSON Processing</h2>
...
-The rule above guarantees semantic interoperability between JSON and JSON-LD for
-literal JSON keys mapped to URIs by the `@context` mechanism. While JSON-LD
-processors will use the specific mechanism provided and can verify that all
-terms are correctly specified, JSON-based processors implicitly accept the same
-semantics without performing any JSON-LD transformations, but instead by
+These rules alone do not guarantee semantic interoperability between JSON and JSON-LD for
+literal JSON keys mapped to URIs by the `@context` mechanism.
+          </p>
+While some RDF processors will use the language features of JSON-LD and can verify that all
+terms are correctly specified, JSON-LD processors implicitly accept the same
```
OR13, line 4318:
```suggestion
terms are correctly specified, JSON-LD processors explicitly accept the same
```

OR13: this is left over, and is still a bit awkward, the intention here is essentially repeat the "relationship to RDF phrasing", but without implying that RDF processing has occurred.

index.html
```diff
@@ -4297,19 +4304,60 @@ <h2>JSON Processing</h2>
...
+          </p>
+          <p>
+Some of the best features of JSON-LD are only possible by applying its language features to the 
+terms defined in the <code>@context</code>.
+          </p>
+          <p>
+Implementers should be aware that <code>id</code> and <code>type</code> have been
+aliased to <code>@id</code> and <code>@type</code>, and are marked <code>@protected</code>.
+Depending on the software libraries used, attempts to redefine these terms may raise processing errors.
+Implementers interested in the details being what <code>@id</code> means
```
OR13, line 4341:
```suggestion
Implementers interested in understanding what <code>@id</code> means 
```

common.js
```diff
@@ -93,6 +93,14 @@ var vcwg = {
       status:   'CG-DRAFT',
       publisher:  'Credentials W3C Community Group'
     },
+    // todo: replace all references to RDF-NORMALIZATION, with this.
+    'RDF-CANON': {
```
davidlehn, line 97: I don't think there are an RDF-NORMALIZATION references?  Either way, shouldn't need the local RDF-CANON biblio entry since it's in specref: https://www.specref.org/?q=rdf-canon.

common.js
```diff
@@ -93,6 +93,14 @@ var vcwg = {
...
+      title: 'RDF Dataset Canonicalization',
+      href: 'https://www.w3.org/TR/rdf-canon/',
+      authors: ['Dave Longley', 'Gregg Kellogg', 'Dan Yamamoto'],
+      status:   'WG-DRAFT',
+      publisher:  'RDF Dataset Canonicalization and Hash Working Group'
+    },
```
msporny, line 103:
```suggestion
```

Remove since specref already has the most up-to-date reference.

index.html
```diff
@@ -4240,47 +4240,54 @@ <h2>HTTP</h2>
...
+When an application chooses to process the media type using the `+ld+json`
 structured media type suffix.
             </li>
           </ul>
 
           <p>
-That is, JSON processing is allowed as long as the document being consumed or
-produced is a <a>conforming document</a>. If JSON processing is desired, an
-implementer is advised to follow the following rule:
+Naive JSON processing is allowed as long as the document being consumed or
```
msporny, line 4284: "Naive JSON processing" isn't defined, so it's not clear what is being talked about here. Suggest we point this to the renamed "JSON Processing" section and call it "Minimal Processing" (under the JSON-LD Processing section).

andresuribe87 (approved):
This is great for implementers

msporny (changes requested):
Overall, happy with 80% of the text in this PR. The rest of it either needs to be made less vague, be corrected, or be removed. Concrete suggestions have been made where possible, questions raised for vague statements to get further clarity.

index.html
```diff
@@ -4240,47 +4240,54 @@ <h2>HTTP</h2>
...
           <p>
 While the media types describing conforming documents defined in this
-specification always express JSON-LD, JSON-LD processing is not required to be
-performed, since JSON-LD is JSON. Some scenarios where processing a
-<a>verifiable credential</a> or a <a>verifiable presentation</a> as JSON is
+specification always express JSON-LD, RDF processing is not required to be
+performed, since JSON-LD is a concrete RDF syntax as described in [RDF11-CONCEPTS]. 
+Hence, a JSON-LD document is both an RDF document and a JSON document and correspondingly represents an instance of an RDF data model.
+See <a href="JSON-LD11##relationship-to-rdf">Relationship to RDF</a> for more details.
+          </p>
```
dlongley, line 4250:
```suggestion
          </p>
           <p>
RDF is a technology for modeling graphs of statements. Each statement is a single subject-property-value
relationship, which is referred to as a <a>claim</a> in this specification. JSON-LD is a technology that
enables the expression of RDF using idiomatic JSON, enabling developers familiar with JSON to write
applications to consume RDF as JSON. In general, subjects are expressed as JSON objects with each property and value of the subject as a JSON key and value, respectively. A special affordance is made to express an identifier of a subject, if present, using the <code>@id</code> JSON key, which is aliased in this specification to <code>id</code>.
See <a href="JSON-LD11##relationship-to-rdf">Relationship to RDF</a> for more details.
          </p>
           <p>
As JSON can be used to express all different kinds of information, a consumer of a particular JSON
document can only properly interpret the authors intent if they possess information that contextualizes
and disambiguates it from other possible expressions. Information to assist with this interpretation can either 
be wholly external to the JSON document or linked from within it. Compact JSON-LD documents include a <code>@context</code> property that internally expresses or links to contextual information and, per the JSON-LD specification, these documents follow object-oriented organizational principles to express <a>claims</a>. These features enable generalized processors to be written to convert JSON-LD documents from one context to another, but this is not needed when consumers receive JSON-LD documents that already use the context and shape that they expect. Authors of JSON-LD documents, such as <a>issuers</a> of <a>verifiable credentials</a>, are required to provide proper JSON-LD contexts and follow these rules in order to facilitate interoperability.
          </p>
           <p>
The text below helps consumers understand how to ensure a JSON-LD document is expressed in a context
and shape that their application already understands such that they do not need to transform it in
order to consume its contents. Notably, this does not mean that consumers do not need to understand any context at all, rather, consuming applications only need to understand a chosen set of contexts and document shapes to work with and not others. Issuers can publish contexts and information about their <a>verifiable credentials</a> to aid consumers who do not use generalized processors, just like would be done with any other JSON-formatted data.
          </p>
```

index.html
```diff
@@ -4240,47 +4240,54 @@ <h2>HTTP</h2>
...
+          <p>
+Some scenarios where processing a <a>verifiable credential</a> or a <a>verifiable presentation</a> as RDF is
 desirable include, but are not limited to:
```
dlongley, line 4253:
```suggestion
Some scenarios where limiting processing to a limited set of expected
<a>verifiable credentials</a> or a <a>verifiable presentations</a> is
desirable include, but are not limited to:
```

index.html
```diff
@@ -4240,47 +4240,54 @@ <h2>HTTP</h2>
...
             </li>
             <li>
-When operating on <a>verifiable credentials</a> or <a>verifiable
-presentations</a> in a software application, after verification or validation
-is performed for securing mechanisms that require an understanding of
-and/or processing of JSON-LD.
+When validating <a>verifiable credentials</a> or <a>verifiable
+presentations</a> in a software application, after verification 
+is performed for securing mechanisms that do not require an understanding
+or processing of JSON-LD.
```
dlongley, line 4275:
```suggestion
is performed for securing mechanisms that do not require JSON-LD transformations.
```

index.html
```diff
@@ -4240,47 +4240,54 @@ <h2>HTTP</h2>
...
             </li>
           </ul>
 
           <p>
-That is, JSON processing is allowed as long as the document being consumed or
-produced is a <a>conforming document</a>. If JSON processing is desired, an
-implementer is advised to follow the following rule:
+Naive JSON processing is allowed as long as the document being consumed or
+produced is a <a>conforming document</a>, however, this can lead to interoperability issues 
+and confusion about the specific information that has been secured between JSON-LD and RDF processors.
```
dlongley, line 4286: No, "Naive JSON processing" is never allowed ... with any JSON. I guess I don't know what this means. When you are processing JSON ... or anything generally, you shouldn't be naive. There is also no special danger of confusion  or deviation of interpretation from generalized processors if the rules are followed, so we should focus there.

An attempt at rewording:

```suggestion
<a>Conforming documents</a> that are expressed using contexts and shapes that are
well-known to a consuming application can be consumed like any other JSON
document, that is, according to the rules of its associated specifications and interpreted
according to the <a>issuers</a> intent. If the <a>issuer</a> has used <code>@context</code>
to provide details regarding the expression of their <a>verifiable credential</a>, then
consumers are advised to follow these rules to ensure that the document matches
their accepted contexts and document shapes:
```

index.html
```diff
@@ -4297,19 +4304,60 @@ <h2>JSON Processing</h2>
 implementing the rule above. This can ensure proper term identification,
 typing, and order, when a JSON document is processed as JSON-LD.
           </p>
+          <p>
+However, the intention of the issuer is not preserved, unless the <code>@context</code>
+is applied, and the resulting <a href="#verifiable-credential-graphs">Verifiable Credential Graphs</a>,
+and their exact RDF types, and node and edge structure is understood by the verifier.
```
dlongley, line 4310:
```suggestion
However, the intention of the issuer is not preserved, unless the rules expressed
in the <code>@context</code> are followed, and the resulting <a href="#verifiable-credential-graphs">Verifiable Credential Graphs</a>, and their exact RDF types, and node and edge structure is understood by the verifier. By following the rules laid out in this section, applications can be hard-coded against
specific contexts, just as they are commonly hard-coded to consume other
application-specific JSON formats.
```

index.html
```diff
@@ -4297,19 +4304,60 @@ <h2>JSON Processing</h2>
...
+          </p>
 
           <p>
-The rule above guarantees semantic interoperability between JSON and JSON-LD for
-literal JSON keys mapped to URIs by the `@context` mechanism. While JSON-LD
-processors will use the specific mechanism provided and can verify that all
-terms are correctly specified, JSON-based processors implicitly accept the same
-semantics without performing any JSON-LD transformations, but instead by
+These rules alone do not guarantee semantic interoperability between JSON and JSON-LD for
+literal JSON keys mapped to URIs by the `@context` mechanism.
```
dlongley, line 4315: I think this is true now with some clean up above and it's simpler:

```suggestion
The rules above guarantee semantic interoperability between applications
that are hard-coded to specific contexts and applications that more
generally consume JSON-LD documents.
```

index.html
```diff
@@ -4297,19 +4304,60 @@ <h2>JSON Processing</h2>
...
+          </p>
+While some RDF processors will use the language features of JSON-LD and can verify that all
```
dlongley, line 4317: Ran out of time down here -- will have to do more review later.

dlongley (changes requested):

index.html
```diff
@@ -4240,47 +4240,54 @@ <h2>HTTP</h2>
...
+          <h2>JSON-LD Processing</h2>
           <p>
 While the media types describing conforming documents defined in this
-specification always express JSON-LD, JSON-LD processing is not required to be
-performed, since JSON-LD is JSON. Some scenarios where processing a
-<a>verifiable credential</a> or a <a>verifiable presentation</a> as JSON is
+specification always express JSON-LD, RDF processing is not required to be
+performed, since JSON-LD is a concrete RDF syntax as described in [RDF11-CONCEPTS]. 
+Hence, a JSON-LD document is both an RDF document and a JSON document and correspondingly represents an instance of an RDF data model.
+See <a href="JSON-LD11##relationship-to-rdf">Relationship to RDF</a> for more details.
```
iherman, line 4249: The notion of "RDF processing" is not defined, and it can mean many things. Better not go there. Besides, it took me some time to understand what (I think) you meant in this section... What about a rewrite of the whole paragraph along the lines of:

> While the media types describing conforming documents defined in this specification all express JSON-LD, developers should be aware that JSON-LD is a serialization of the abstract RDF Model [[RDF11-CONCEPTS]]. Hence, a JSON-LD document is both a JSON document and the representation of an abstract RDF Dataset. See <a href="JSON-LD11#relationship-to-rdf">Relationship to RDF</a> for more details. 

Also, in view of what you write later, _I have the feeling_ that you have in mind two type of "processing":

- "RDF Processing" when you transform the JSON-LD into RDF using the JSON-LD rules
- "JSON-LD Processing" when you operate on the JSON level only.

if that is correct, this may be the place to define these, because otherwise the text below is very unclear. That being said, I do not like the two terms; for me, "JSON-LD Processing" means to perform the processing steps defined in the [JSON-LD 1.1 Processing Algorithms and API](https://www.w3.org/TR/json-ld11-api/) specification, and this seems to contradict this.

index.html
```diff
@@ -4240,47 +4240,54 @@ <h2>HTTP</h2>
...
-
           <ul>
             <li>
-Before securing or after verifying content
-that requires <a href="https://csrc.nist.gov/glossary/term/data_integrity">data
-integrity</a>, such as a
-<a>verifiable credential</a> or <a>verifiable presentation</a>.
+Before securing or after verifying <a>verifiable credentials</a> or <a>verifiable presentations</a>,
+in cases where the issuer has leveraged the <code>@context</code> to provide details regarding 
+their intended expression of the <a href="#verifiable-credential-graphs">Verifiable Credential Graphs</a>.
```
iherman, line 4259:
```suggestion
in cases where the issuer made use of the underlying RF Dataset model to provide details regarding 
their intended expression of the <a href="#verifiable-credential-graphs">Verifiable Credential Graphs</a>.
```

At least that is how I understood what I _think_ you meant...

index.html
```diff
@@ -4240,47 +4240,54 @@ <h2>HTTP</h2>
...
             </li>
             <li>
-When performing JSON Schema validation, as described in Section
-<a href="#data-schemas"></a>.
+After performing JSON Schema validation, as described in Section
+<a href="#data-schemas"></a>, in order to mitigate attacks 
+on the canonicalization process as described in <a href="https://w3c.github.io/rdf-canon/spec/#dataset-poisoning">Dataset Poisoning</a>
```
iherman, line 4264: Two comments on the same section

- I am not sure why JSON schema is relevant in this respect. I guess the issue is that if the issuer decided to use the RDFC 1.0 algorithm for its security related processing, and that is completely orthogonal whether he/she also uses JSON schemas. The two are unrelated
- RDFC 1.0 takes care of the dataset positing issue, so "mitigate may not be appropriate here. The bullet point may be:
```suggestion
In order to fine tune the <a href="https://www.w3.org/TR/rdf-canon/#canon-algo-algo">defense against the dataset poisoning issue</a> built-in the RDFC 1.0 algorithm. [[RDF-CANON]]
```

index.html
```diff
@@ -4240,47 +4240,54 @@ <h2>HTTP</h2>
...
             </li>
             <li>
 When serializing or deserializing <a>verifiable credentials</a> or
-<a>verifiable presentations</a> into systems that store or index their contents.
+<a>verifiable presentations</a> into systems that store or index claims about subject in
+<code>text/turtle</code>, <code>application/n-quads</code>, <code>application/ld+json</code>.
```
iherman, line 4269: Systems do not store RDF datasets in text/turtle and friends. Those are all serialization formats. RDF runtimes like RDFLib, rdf-js, Jena, etc, or RDF Databases like Virtuoso or Amazon Neptune all store the data in their propriatery format.

```suggestion
<a>verifiable presentations</a> into an RDF Data Store. 
```

B.t.w., the counterpart of n-quads should be text/trig and not text/turtle

index.html
iherman, line 4315: I do not understand what you mean

index.html
```diff
@@ -4297,19 +4304,60 @@ <h2>JSON Processing</h2>
...
-happens is explicitly stated for both JSON and JSON-LD by using the same
-mechanism. With respect to JSON-based processors, this is achieved in a
-lightweight manner, without having to use JSON-LD processing libraries.
+happens is explicitly stated for both RDF and JSON-LD by using the same
+mechanism. With respect to JSON-LD-based processors, this is achieved in a
+lightweight manner, without having to use RDF processing libraries.
+          </p>
+        </section>
+        <section>
+          <h2>Advanced RDF Processing</h2>
```
iherman, line 4327: Can we find an alternative subtitle? There is nothing "Advanced" in what is in the section, at least no from an RDF point of view...

index.html
```diff
@@ -4297,19 +4304,60 @@ <h2>JSON Processing</h2>
...
+          </p>
+          <p>
+Implementers should be aware that object member, and array element order are not preserved by default in JSON-LD,
+and as such, Issuers intending to communicate array order need to leverage the langauge features of JSON-LD.
+See <a href="https://www.w3.org/TR/json-ld11/#sets-and-lists">Value Ordering</a>.
+          </p>
+          <p>
+Implementers should be aware that while naive JSON interpretation of claims might make it appear that claims are related,
+the <code>https://www.w3.org/ns/credentials/v2</code> leverages <code>@container</code> and <code>@graph</code> to separate information graphs.
+In particular, to separate the graphs related to <a>verifiable credentials</a>, <a>verifiable presentations</a>
```
iherman, line 4353: AFAIK there is no property marked as `"@container":"@graph"` for VP-s. Only credentials and proofs are "separated" in their own graphs.

index.html
iherman: I saw now @dlongley's comment that the whole bullet point should be removed in view of the recent changes in RDFC; I would be o.k. with that, too.

RolandoMorales22 (approved):

index.html
OR13: The intention was to clearly communicate that there is "JSON-LD processing" where you don't apply the context (This was previously called "JSON Processing"... and there is JSON-LD processing, where you do apply the context (this is currently titled "Advanced RDF Processing".).

I'm not attached to the names, I am attached to signaling that both processing rules are in the context of RDF.

index.html
```diff
@@ -4297,19 +4304,60 @@ <h2>JSON Processing</h2>
...
+not all Holders or Verifiers will understand expressions of intention that are only visible 
+after RDF processing has occured.
+          </p>
+          <p>
+Some of the best features of JSON-LD are only possible by applying its language features to the 
+terms defined in the <code>@context</code>.
+          </p>
+          <p>
+Implementers should be aware that <code>id</code> and <code>type</code> have been
+aliased to <code>@id</code> and <code>@type</code>, and are marked <code>@protected</code>.
```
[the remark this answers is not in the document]
OR13: I agree, do you you have any concrete text, or maybe its better for us to just refer to https://www.w3.org/TR/json-ld11/#protected-term-definitions ?

index.html
OR13: https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v2#L175

^ This one... and 

https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v2#L139

Is what I was referring too.

index.html
iherman: I would not call the first "JSON-LD Processing", because it is very misleading. The term "JSON-LD Processing" is a term coined by the JSON-LD WG and [has a well-defined meaning](https://www.w3.org/TR/json-ld11-api/). It is actually closer to what you call "Advanced RDF Processing".

What we describe in the spec handling VC files in JSON-LD format but _without_ a full JSON-LD Processing is, though fairly general, proprietary to the VC specification. Maybe using something like "Basic VC Processing" would do.

index.html
iherman: What those mean is that a Verifiable Credential, resp. a proof, is put into a separate graph. You are right that this means, in practice, that the Verifiable Presentation that refers to the credential is therefore kept in its own (default) graph, but the sentence is nevertheless misleading.

OR13:
In case you missed this, https://github.com/w3ctag/design-principles/issues/239#issuecomment-1748162752

Very relevant to the guidance we are attempting on this PR, and on https://github.com/w3c/vc-data-model/pull/1270

dlongley:
> In case you missed this, [w3ctag/design-principles#239 (comment)](https://github.com/w3ctag/design-principles/issues/239#issuecomment-1748162752)
> 
> Very relevant to the guidance we are attempting on this PR, and on #1270

I don't think it's relevant, but I see how people could get confused that it is. Let's not add to that. Part of this confusion, I believe, comes from the fact that we keep trying to use "processing" terminology that is either under-defined or differs from how it is defined in the JSON-LD specification. This is one of Ivan's points above.

So, I think the most troublesome terminology is "JSON processing" and "JSON-LD processing", which everyone attributes their own (different) meaning to. We ought to avoid those terms. We are not presenting two different models for interpretation, just one. There can be different steps to get there based on localized optimizations, but whether you do steps X or Y, everyone must arrive at the same interpretation; that's what this section is about. This is a common concept for interfaces: the implementation details behind them should not matter.

In this section, we're telling people how to write software to get a consistent view of what pieces of the data they choose to consume. You can either consume JSON-LD by first transforming it to a context your application speaks (i.e., do "JSON-LD processing" as defined by the JSON-LD API spec by running compaction, etc.) or by limiting your inputs to data that your application natively accepts. Both of these must result in an interpretation that is according to "the context", whether that interpretation is hard-coded into your application like it is for every "application-specific vanilla JSON" consumer or if your application can work with multiple contexts and / or data shapes.

Your application either accepts only one specific expression that you know how to interpret properly and rejects everything else, or it accepts other expressions that it can transform, but once transformed, will result in the same interpretation as if it had started that way. The former is the only way software can be written when working with "application-specific vanilla JSON", because it does not have "transformation capabilities" (this is something that JSON-LD adds). But this is additive. You can still write applications that do not use these capabilities, you just can't pretend that you can interpret however you want. But that is never true with any format; you must always interpret according to the author's intent or else you will not interoperate. That the author can tag their data with machine-processable context helps widen where it can be used and accepted.

The fact that you can transform to another context doesn't mean you must nor does it mean that you must accept multiple expressions. In order to increase interoperability, the onus is on the producer to express the data in a way most will accept, which means in a way that simple applications that do not perform transforms will accept -- because those that do perform them can handle that expression too. The advantage for generalized processors is that they won't need to read your spec to transform to a context they accept -- because, with JSON-LD, your data carries that contextual information.

iherman:
Adding to what @dlongley says: 

> In this section, we're telling people how to write software to get a consistent view of what pieces of the data they choose to consume. You can either consume JSON-LD by first transforming it to a context your application speaks (i.e., do "JSON-LD processing" as defined by the JSON-LD API spec by running compaction, etc.) or by limiting your inputs to data that your application natively accepts. 

There is one aspect that should be emphasized: "limiting your inputs to data that your application natively accepts" means that what we are defining here is a processing _strictly limited to Verifiable Credentials and Presentations_. Ie, we have a general mechanism (JSON-LD) _and_ we have a specialized version thereof for our purposes. That is why I wondered about the "Basic VC Processing" term for the latter. (I am not saying @dlongley did not put this in his comment, but it should be emphasized better nevertheless.)

Maybe an example from the past helps. In the [Publication Manifest specification](https://www.w3.org/TR/2020/REC-pub-manifest-20201110/#manifest-processing) we defined manifest metadata in terms of JSON-LD (we did not even have a choice because we defined a profile and extension to schema.org). The dilemma about the general vs. specialized processing was the same. We were very meticulous, and we even defined the specialized processing steps of a manifest in great details (see [§7.4 of that spec](https://www.w3.org/TR/2020/REC-pub-manifest-20201110/#processing-algorithm)), and we also had two independent implementations for that algorithm to prove its correctness. *I am not saying we have to go through the same level of details in the VCDM spec*, but maybe it explains the general thought process.

awoie:
IMO, nothing in the specification should suggest that JSON-only processors can be VCDM 2.0 conformant. Because this would mean all normative statements can be checked without RDF/JSON-LD processing. `type` having to be a URL is one example. To make this absolutely clear this should be explicitly put into the specification. Putting JSON-only processors into the mix is actually harmful.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-10-11-vcwg#section1-6) on 2023-10-11

- no resolutions were taken

#### 1.6. Propose better JSON-LD processing text (pr vc-data-model#1302)

_See github pull request [vc-data-model#1302](https://github.com/w3c/vc-data-model/pull/1302)._




**Brent Zundel:** A proposal for better JSON-LD processing text. It has an approval and a bunch of comments. There's been quite a bit of conversation here.  

**Manu Sporny:** Yeah, so I think around 80% of the text is good. These are good modifications to make. It's clear that people don't like the "JSON processing" terminology so we're trying to say you don't have to do "full blown JSON-LD processing" and still be ok. The 20% of the text is being debated in the PR. 80% is good clarifications to make, we need to get to consensus on the 20%.  
… All kinds of concrete suggestions I made there.  

**Dave Longley:** I feel similarly, there are some good improvements here, good suggestions made by others, would like to see others changes pulled in.  
… One of the big things out of this PR is that talking about "JSON Processing" and "JSON-LD Processing" is problematic, because those terms mean different things to different people. If we can avoid using that terminology, and just talk about specifics on consuming data, that would be best.  
… Some text suggested in this PR would help us get there.  

**Brent Zundel:** Not seeing anyone else on the queue, next steps would be Orie jumping in and accepting suggested changes or continuing the conversation there.  
… This PR looks to be in a decent spot.  

</details>

common.js
OR13: I removed this.

index.html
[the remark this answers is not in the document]
OR13: I like this suggestion, I will update to it

index.html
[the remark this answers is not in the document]
OR13: The point is that RDF processing of a conforming document and JSON processing of a conforming document can lead to different information graphs.

This is caused the context contains ANY information that is critical to the understanding of the claims.

A JSON verifier will not have this context.

An RDF verifier will.

index.html
```diff
@@ -4240,47 +4240,54 @@ <h2>HTTP</h2>
...
-That is, JSON processing is allowed as long as the document being consumed or
-produced is a <a>conforming document</a>. If JSON processing is desired, an
-implementer is advised to follow the following rule:
+Naive JSON processing is allowed as long as the document being consumed or
+produced is a <a>conforming document</a>, however, this can lead to interoperability issues 
+and confusion about the specific information that has been secured between JSON-LD and RDF processors.
+          </p>
+
+          <p>
+Implementers unfamilar with JSON-LD / RDF are advised:
```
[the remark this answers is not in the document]
OR13: I'd caution against encouraging processing of conforming documents as different media types, its JSON-LD processing either way... the difference is if you process the context or not.

Perhaps processing JSON-LD without understanding `@context` ?

index.html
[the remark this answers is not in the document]
OR13: No, because in the case the issuer included structure in the context, and a verifier ignores the context, the verifier ignores the issuers intent.

The verifier has no way to know if the issuer meant for them to ignore the context or not.

index.html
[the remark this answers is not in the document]
OR13: https://w3c.github.io/vc-data-model/#example-a-simple-example-of-the-contents-of-a-verifiable-presentation

```json
{
  "@context": [
    // imagine 50% of the time this is in examples/v2 ^ and 50% of the time its not
    { "name": { "@id": "https://bar.example", "@type": "@id" }} 
  ],
  "name": "bob",
  "verifiableCredential": [{
    "@context": [
    ],
    "id": "http://university.example/credentials/1872",
    "type": ["VerifiableCredential", "ExampleAlumniCredential"],
    "credentialSubject": {
      "alumniOf": {
        "id": "did:example:c276e12ec21ebfeb1f712ebc6f1",
        "name": "Example University"
      }
    }
  }]
}
```

```json
{
  "@graph": [
    {
      "@type": "https://www.w3.org/2018/credentials#VerifiablePresentation",
      "https://bar.example": {
        "@id": "bob"
      },
      "https://www.w3.org/2018/credentials#verifiableCredential": {
        "@id": "http://university.example/credentials/1872",
        "@type": [
          "https://www.w3.org/2018/credentials#VerifiableCredential",
          "https://www.w3.org/ns/credentials/examples#ExampleAlumniCredential"
        ],
        "https://www.w3.org/2018/credentials#credentialSubject": {
          "@id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
          "https://www.w3.org/ns/credentials/examples#alumniOf": {
            "@id": "did:example:c276e12ec21ebfeb1f712ebc6f1",
            "https://www.w3.org/ns/credentials/examples#name": "Example University"
          }
        },
        "https://www.w3.org/2018/credentials#issuer": {
          "@id": "https://university.example/issuers/565049"
        },
        "https://www.w3.org/2018/credentials#validFrom": {
          "@type": "http://www.w3.org/2001/XMLSchema#dateTime",
          "@value": "2010-01-01T19:23:24Z"
        }
      }
    },
    {
      "@id": "bob"
    },
    {
      "@id": "did:example:c276e12ec21ebfeb1f712ebc6f1",
      "https://www.w3.org/ns/credentials/examples#name": "Example University"
    },
    {
      "@id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
      "https://www.w3.org/ns/credentials/examples#alumniOf": {
        "@id": "did:example:c276e12ec21ebfeb1f712ebc6f1",
        "https://www.w3.org/ns/credentials/examples#name": "Example University"
      }
    },
    {
      "@id": "http://university.example/credentials/1872",
      "@type": [
        "https://www.w3.org/2018/credentials#VerifiableCredential",
        "https://www.w3.org/ns/credentials/examples#ExampleAlumniCredential"
      ],
      "https://www.w3.org/2018/credentials#credentialSubject": {
        "@id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
        "https://www.w3.org/ns/credentials/examples#alumniOf": {
          "@id": "did:example:c276e12ec21ebfeb1f712ebc6f1",
          "https://www.w3.org/ns/credentials/examples#name": "Example University"
        }
      },
      "https://www.w3.org/2018/credentials#issuer": {
        "@id": "https://university.example/issuers/565049"
      },
      "https://www.w3.org/2018/credentials#validFrom": {
        "@type": "http://www.w3.org/2001/XMLSchema#dateTime",
        "@value": "2010-01-01T19:23:24Z"
      }
    },
    {
      "@id": "https://university.example/issuers/565049"
    }
  ]
}
```

The point is that processing the context leads to an (often but not necessarily always) better interpretation of the issuer or holder's intended claims.

It's not safe to say that you can ignore context processing and "know the same things"... its trust that assuming you preserve context, and context does not change, you can "come to the same conclusions later".

In the same I show, the JSON values are literally mapped differently when the context is applied... We know that context can change, perhaps when eratta are published, this is why we set the cache policy for contexts, and why we care about treating contexts as previously dereferenced.

I the context in the example above changes, the semantic interop changes, including how term definitions are mapped.

The only way the statement holds that committing to a context preserves semantic interop is when the context is immutable, and we know thats not the case, even if its "usually" the case... since this is a security spec, the bar should not be "mostly true", the bar should be "always true", or "not always true".

index.html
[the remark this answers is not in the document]
OR13: I applied manu's suggestion, and this comment is now outdated, can you review the latest and suggestion additional changes if needed?

index.html
```diff
@@ -4240,47 +4240,54 @@ <h2>HTTP</h2>
...
+<a>verifiable presentations</a> into systems that store or index claims about subject in
+<code>text/turtle</code>, <code>application/n-quads</code>, <code>application/ld+json</code>. 
             </li>
             <li>
-When operating on <a>verifiable credentials</a> or <a>verifiable
-presentations</a> in a software application, after verification or validation
-is performed for securing mechanisms that require an understanding of
-and/or processing of JSON-LD.
+When validating <a>verifiable credentials</a> or <a>verifiable
+presentations</a> in a software application, after verification
```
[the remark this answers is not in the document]
OR13: haha yea...

index.html
```diff
@@ -4297,19 +4304,60 @@ <h2>JSON Processing</h2>
...
+mechanism. With respect to JSON-LD-based processors, this is achieved in a
+lightweight manner, without having to use RDF processing libraries.
+          </p>
+        </section>
+        <section>
+          <h2>Advanced RDF Processing</h2>
+          <p>
+Issuers should be aware that while conforming documents are expressed as compact JSON-LD,
+not all Holders or Verifiers will understand expressions of intention that are only visible 
+after RDF processing has occured.
```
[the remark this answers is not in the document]
OR13: issuers make claims about subjects, their intention is to be understood consistently by verifiers, or their intention is to equivocate.

index.html
OR13: I'd take "Some scenarios where limiting RDF processing"... not I won't take removal of RDF entirely.

index.html
OR13: Thanks, my general intention is to make it clear, you need to understand JSON-LD.

index.html
OR13: I don't think your suggestion makes it clearer. The issuer is required to use `@context`, since it's required in conforming documents.

Perhaps you mean additional context beyond the v2 core context? can you adjust your suggestion to take that into account?

index.html
OR13: "applications can be hard-coded against specific contexts, just as they are commonly hard-coded to consume other application-specific JSON formats."

^ can you update this to account for when the contexts change, and how that impacts what the verifier will understand  based on if the verifier is hard coded or is consuming a context that has errata published?

index.html
OR13: This is not true, but could be true modulo some defense language around contexts changing, and it would be nice to say it this way.

index.html
OR13: Please re-review, I've applied some suggestions from Manu, and the discussion on them is relevant to your suggestions which I have not applied.

As I see it the main challenge remaining is to address the "hard coded" vs "errata published" scenario for contexts, in the case where the issue does not intend to equivocate but is forced to by their commitment to a context hosted by a third party (including W3C as a 3rd party).

index.html
OR13: ^ this is correct, my goal is to basically say this:

1. You need to care about JSON-LD
2. You can care about it and ignore the context, you will have a bad time.
3. You can care about it and pay attention to the context, and W3C or any other 3rd party context host can cause you to have a bad time (by changing the context after you commit to it).
4. In the case that all contexts are immutable and always available, you can still have equivocation, but at least its limited to differences between RDF and JSON, not RDF and JSON-LD (where there will be NO equivocation in this very special case).

index.html
OR13: Yes, thank you, much better

index.html
OR13: I think its better to give direct security advice here, conforming documents are JSON, schema validation can protect against poisoning... telling people to read the RDF spec without spelling this out is harmful to the security posture of our document.

index.html
OR13: @iherman I applied Manu's suggestion, which I thing better addresses this

index.html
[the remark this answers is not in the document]
OR13: Outdates, hopefully the updates are not comprehensible.

index.html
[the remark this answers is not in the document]
OR13: its what used to be called "JSON processing" or in DID Core was called "Pure JSON processing".... its basically ignoring the spec and treating JSON-LD as JSON, see https://tess.oconnor.cx/2023/09/polyglots-and-interoperability

index.html
```diff
@@ -4297,19 +4304,60 @@ <h2>JSON Processing</h2>
 implementing the rule above. This can ensure proper term identification,
 typing, and order, when a JSON document is processed as JSON-LD.
           </p>
+          <p>
+However, the intention of the issuer is not preserved, unless the <code>@context</code>
```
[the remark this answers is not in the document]
OR13: The intention of the issue cannot be to do "pure json" processing, conforming documents are marked as JSON-LD.

index.html
OR13: I clarified in https://github.com/w3c/vc-data-model/pull/1302#discussion_r1358952470

index.html
```diff
@@ -4297,19 +4304,60 @@ <h2>JSON Processing</h2>
...
-literal JSON keys mapped to URIs by the `@context` mechanism. While JSON-LD
-processors will use the specific mechanism provided and can verify that all
-terms are correctly specified, JSON-based processors implicitly accept the same
-semantics without performing any JSON-LD transformations, but instead by
+These rules alone do not guarantee semantic interoperability between JSON and JSON-LD for
+literal JSON keys mapped to URIs by the `@context` mechanism.
+          </p>
+While some RDF processors will use the language features of JSON-LD and can verify that all
+terms are correctly specified, JSON-LD processors explicitly accept the same
+semantics without performing any RDF transformations, but instead by
```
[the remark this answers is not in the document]
OR13: I think we addressed this, and we'll need to resolve things that are addressed so I can process the remaining blockers. can you confirm?

index.html
OR13: Manu suggested calling in RDF processing, I think thats a fair compromise... its the "JSON-LD... but with transformations that care about effort put into the context" processing kind...

index.html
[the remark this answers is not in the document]
OR13: I think we addressed this, and we'll need to resolve things that are addressed so I can process the remaining blockers. can you confirm?

index.html
OR13: Its more misleading to ignore that RDF transformations make this distinction but JSON-LD (Without context processing) does not... how do you suggest we proceed?

OR13:
Thanks for all the valuable comments, the ones I could accept, I have.

In cases where I couldn't accept clarification was provided I have provided.

The PR will quickly get un-reviewable.

My suggestion would be for anything where we are still stuck, we file issue markers, and open smaller more focused PRs to address them.

I expect the whole document still needs a lot of updates, so the faster we merge PRs (and the more smaller PRs we can get) the better.

index.html
dlongley: > You can care about it and ignore the context, you will have a bad time.

No, this is not a thing. You never "ignore" `@context`. You can hard-code your application to only work with some specific contexts, that's fine. But you don't "ignore" `@context`, that is not at all the same thing. Perhaps this is part of a core misunderstanding here?

The comments about context immutability are a different kind of issue. We can have text here that says: "if you're going to hard-code to specific contexts, they must be immutable in order to guarantee the same interpretation as someone who doesn't hard-code". That is a simple fact and encouraging the use of immutable contexts for VCs is a good thing, IMO.

index.html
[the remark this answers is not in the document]
dlongley: > Perhaps processing JSON-LD without understanding `@context`?

See my latest comment below; this is not a thing. You never "ignore" `@context`. You can write applications that only accept specific contexts and read only the properties mapped therein, but you cannot ever consume properties from contexts that are not understood.

This section was never about doing that -- and that is not supported. This section is about telling people how to hard-code simple applications to specific, well-known contexts, making it more clear to newcomers that performing context transformations using a JSON-LD library that implements the JSON-LD API is not needed when you limit your data inputs to data that your application natively understands.

index.html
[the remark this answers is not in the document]
dlongley: > No, because in the case the issuer included structure in the context, and a verifier ignores the context, the verifier ignores the issuers intent.

No verifier can ignore `@context`; that is not conformant processing. You can only consume properties from contexts you understand. If that's only the core context, that's fine -- but you can't read properties from other contexts unless you understand them.

index.html
iherman: Dataset poisoning is an issue in general but, I have the impression, it is _very rarely_ an issue with v. credentials or presentations, and the text we are writing here relates to those.  (It has been my suspicion for a long time that a VC/VP dataset is, from an RDF point of view, ["well-behaved"](http://dbooth.org/2013/well-behaved-rdf/Booth-well-behaved-rdf.pdf) 90% of the time, which means that the "tricky" part of the RDFC 1.0 algorithm, that may run amok, is not applied in the first place. But I cannot prove that.) The VCDM spec does not even talk about defining a blank node explicitly, something that is usually done to create poisoned datasets.

(Let alone the fact that, per RDFC 1.0 spec, each implementation may have its own way of defending itself against poisoned datasets, whose parameters (that the user may fiddle with) are therefore also implementation dependent. The RDFC standard only stipulates that _a defense must exist_ and have a reasonable default parameter setting; the test suite does include a few poisoned datasets that the implementation should pass.)

As a conclusion, I do not believe that going to the details of the RDFC 1.0 is really an issue _for this text_.

index.html
iherman: I do not see any change. The text still says 

> When serializing or deserializing [verifiable credentials](http://localhost:8001/LocalData/github/VC/vc-data-model/#dfn-verifiable-credential) or [verifiable presentations](http://localhost:8001/LocalData/github/VC/vc-data-model/#dfn-verifiable-presentation) into systems that store or index data in Turtle [[TURTLE](http://localhost:8001/LocalData/github/VC/vc-data-model/#bib-turtle)], N-Quads [[N-QUADS](http://localhost:8001/LocalData/github/VC/vc-data-model/#bib-n-quads)], or JSON-LD [[JSON-LD11](http://localhost:8001/LocalData/github/VC/vc-data-model/#bib-json-ld11)].

which is what I objected to: most of the (RDF) data does not store their data in serialized form. With that exclusion the list item is meaningless...

index.html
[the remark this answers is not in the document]
iherman: I see the renaming, it is o.k. for now (we will have to re-read the text altogether anyway before merge...)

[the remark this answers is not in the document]
iherman: However... the term "Naive JSON processing is still in the text".

index.html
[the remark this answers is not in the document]
iherman: I can live with this, for now...

index.html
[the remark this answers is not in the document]
iherman: I *think* it is all right...

index.html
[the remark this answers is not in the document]
iherman: Where is this addressed? The text has not changed...

To be more precise: in §6.1.1 of the spec, the role of `@protected` is indeed explained (in the last bullet), so is the fact that `@id` and `@type` are aliased. In fact, this paragraph does not add any information that would not be explicitly stated in §6.1.1. Maybe the best is to remove it altogether.

index.html
iherman: Remove the reference to verifiable presentation. References to proofs and to verifiable credentials result in separate graphs, as you say in the text, and that is it.

iherman:
> Thanks for all the valuable comments, the ones I could accept, I have.
> 
> In cases where I couldn't accept clarification was provided I have provided.
> 
> The PR will quickly get un-reviewable.

Yep :-(

> 
> My suggestion would be for anything where we are still stuck, we file issue markers, and open smaller more focused PRs to address them.
> 
> I expect the whole document still needs a lot of updates, so the faster we merge PRs (and the more smaller PRs we can get) the better.

I would rather propose to do a shortcut. You'd produce a new text based on all the open issues, mark all the issues that relate to explicit changes as outdated, and we should all re-review the text. From scratch.

I do not believe spawning numerous small issues and PRs makes it easier to review.

index.html
dlongley: Please add this suggestion as it provides more important background that makes clear what this section is about and that the consumer is always consuming RDF expressed in a JSON format when they consume JSON-LD -- and that contextual information must always be understood. This has been a consistent point of confusion and it's important to have this text.

index.html
dlongley: I've made an adjustment to say "Issues might use additional contexts... and consumers are advised...".

index.html
iherman: MInor:

```
<a href="JSON-LD11##relationship-to-rdf"
```

should be

```
<a href="JSON-LD11#relationship-to-rdf"
```

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-10-17-vcwg-special#section2-9) on 2023-10-17

- no resolutions were taken

#### 2.9. Propose better JSON-LD processing text (pr vc-data-model#1302)

_See github pull request [vc-data-model#1302](https://github.com/w3c/vc-data-model/pull/1302)._




**Brent Zundel:** This PR has requests for changes from Dave and Manu. What concrete changes ... are the concrete changes that you're looking for already reflected in your comments?  
… Do we have a path forward.  

**Manu Sporny:** I think we have a path forward, not all comments made have been addressed.  
… 80% of this PR is headed in a good direction, Orie hasn't responded to some of the other stuff, it looks like there needs to be some more negotiation in here. Orie needs to do an updated processing on this.  

**Brent Zundel:** Orie four days ago looks like he did a rejiggering of it. I'm not that the ball is in Orie's court at this point.  

**Manu Sporny:** I see Dave responded after that, I will look at it again.  

**Dave Longley:** I looked at this again after Orie provided feedback, made adjustments, hope those changes go in (up to Orie), also highlighted important section that I'd like to see go in (important to clarify some things that were highlighted some stuff that seemed to be confusing to Orie).  
… If that text goes in, and if stuff I was responsive to Orie goes in, I didn't get to tail end of this PR, but I think we're in good shape. Ivan had a good suggestion on changes/re-reviews.  

**Ivan Herman:** I had some comments and requests for changes and did look at the version...  
… A few days ago ... the things I requested have not been done yet.  
… From my point of view, the ball is in Orie's court. But he's right in saying it's difficult to do things because the github process is that when you look at the changes it gives you all the changes back to the first version. I have to agree it's difficult to review, but as far as I'm concerned there are still open issues on the PR.  

**Brent Zundel:** Thank you, Ivan. This is one of those where -- we really need to see movement.  
… It's been open a couple of weeks, the path toward consensus isn't entirely clear to me.  
… Is it changes to the PR or additional clarifications related to the PR that are primarily being desired?  
… Is this where additional PRs could be added?  

**Ivan Herman:** I was wondering about that. Orie's proposal is to merge as it with open suggestions and then we'd create more issues for each suggestion and we'd all lose.  
… I think we should ask Orie to do all the changes he can and then open a new PR with all the changes to start from scratch to help with review.  
… There are some things outdated and by-discussions ... interesting but not helpful. I think that making a new text with a new PR and closing the other one is a possibility.  
… Orie isn't here, so maybe we have to settle this with him tomorrow.  

**Brent Zundel:** Thank you, Ivan, I'm not fond of that suggestion, personally.  
… If the group feels that that would be the most efficient way forward, I wouldn't oppose it.  
… I will plan on having a discussion on this PR tomorrow. Hopefully Orie will be present tomorrow and that will lead to clarity.  

</details>

brentzundel:
We did not have an opportunity to discuss this issue during the 2023-10-18 VCWG call.

If a clear path forward finds consensus before next week's WG meetings, then we may still merge this PR. Otherwise it likely will be closed.

dlongley:
@brentzundel,

I'll note that this is all informative text in the spec, if that helps alleviate any timeline pressures at all.

brentzundel:
> @brentzundel,
> 
> I'll note that this is all informative text in the spec, if that helps alleviate any timeline pressures at all.

Thank you, Dave, that does help (and should have been the first consideration during our calls this week).

index.html
msporny: @dlongley suggested "Static Processing" vs. "Dynamic Processing"... where "Static Processing" covers things like "Use a JSON Schema to check the well-formed-ness of the payload and then process it like you would a JSON payload... and where "Dynamic Processing" is using a JSON-LD processor to compact/expand or convert to NQuads.

Thoughts?

index.html
msporny: @OR13, please respond to this.

index.html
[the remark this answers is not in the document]
msporny: @OR13, please respond to this.

msporny: @OR13, please respond to this.

index.html
msporny: @OR13, please respond to this.

index.html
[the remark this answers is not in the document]
msporny: I think I'm going to take a shot at a PR that attempts to just use different wording wrt. "JSON Processing" vs. "JSON-LD Processing". That seems to be the key issue that we need to resolve before we start trying to dive into the stuff you're saying above (which only holds in cases where implementers don't follow the guidance we give them).

msporny: @OR13 please respond to this comment.

index.html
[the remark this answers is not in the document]
msporny: @OR13 thoughts on the above?

index.html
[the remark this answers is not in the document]
msporny: @OR13, thoughts?

msporny: @OR13, thoughts?

index.html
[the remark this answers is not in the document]
msporny: I don't think anyone is saying "ignore context processing"... we're saying "if these N things hold true, then you can guarantee that the semantics are the same". So, all we need to do is list the N things that need to hold true. You seem to be saying that the hash for an "immutable" context can change if there are errata... but even then, software can check hashes to detect that -- perhaps that's the one rule that we'd need to add for the statement around the semantics being the same between static processing vs. dynamic processing to hold?

index.html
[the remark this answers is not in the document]
msporny: I still don't understand what you're saying.

index.html
```diff
@@ -4297,19 +4304,60 @@ <h2>JSON Processing</h2>
...
+          <p>
+Some of the best features of JSON-LD are only possible by applying its language features to the 
+terms defined in the <code>@context</code>.
+          </p>
+          <p>
+Implementers should be aware that <code>id</code> and <code>type</code> have been
+aliased to <code>@id</code> and <code>@type</code>, and are marked <code>@protected</code>.
+Depending on the software libraries used, attempts to redefine these terms may raise processing errors.
+Implementers interested in understanding what <code>@id</code> means 
+should review <a href="https://www.w3.org/TR/json-ld11/#node-identifiers">Node Identifiers</a>, and
```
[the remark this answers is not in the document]
msporny: @OR13, thoughts?

index.html
msporny: @OR13, thoughts?

index.html
```diff
@@ -4297,19 +4304,60 @@ <h2>JSON Processing</h2>
...
+          <p>
+Implementers should be aware that object member, and array element order are not preserved by default in JSON-LD,
+and as such, Issuers intending to communicate array order need to leverage the langauge features of JSON-LD.
+See <a href="https://www.w3.org/TR/json-ld11/#sets-and-lists">Value Ordering</a>.
+          </p>
+          <p>
+Implementers should be aware that while naive JSON interpretation of claims might make it appear that claims are related,
+the <code>https://www.w3.org/ns/credentials/v2</code> leverages <code>@container</code> and <code>@graph</code> to separate information graphs.
+In particular, to separate the graphs related to <a>verifiable credentials</a>, <a>verifiable presentations</a>
+and <code>data integrity proofs</code>.
```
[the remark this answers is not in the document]
msporny: @OR13, thoughts?

index.html
```diff
@@ -4297,19 +4304,60 @@ <h2>JSON Processing</h2>
...
+          </p>
+          <p>
+Issuers and Holders might apply these same techniques to unambigiously communicate their intention regarding structured claims in 
+<a>verifiable credentials</a> and <a>verifiable presentations</a>. Verifiers that do not understand or process JSON-LD, will not 
+be aware of these differences, and confusion over the intention of the issuer and the holder could lead to unexpected processing behavior by verifiers.
```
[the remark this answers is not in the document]
msporny: @OR13, thoughts?

msporny (changes requested):

index.html
iherman: I am not sure about the "Dynamic" vs. "Static". I do not lie down on the road over this, but these two terms do not speak to me in this context.

OR13:
@msporny I can't tell what you are asking me to respond to... it seems like there might have been some kind of bug in github when you posted comments.

msporny:
@OR13 wrote:
> @msporny I can't tell what you are asking me to respond to... it seems like there might have been some kind of bug in github when you posted comments.

There are a series of comments that have been outstanding for a while now that need responses from you to see if you are going to address the comments or not:

https://github.com/w3c/vc-data-model/pull/1302/files

index.html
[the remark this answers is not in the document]
OR13: Decline to implement feedback, reasoning is that JSON that is not validated with a schema can always be used to exploit an RDF processor.

OR13: Decline to implement feedback, but perhaps a stronger rewarding regarding the security impact would be acceptable

index.html
OR13: I don't support any form of encouraging implementers of the spec to ignore / not understand conforming documents.

I think the language here is effective, but perhaps not clear enough, and could be fixed in an editorial pass in the future.

index.html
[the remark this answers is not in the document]
OR13: Decline to implement feedback, again, reasoning is security, if we encourage multiple inconsistent interpretations of the same information, we open the ecosystem up to confusion attacks.

I think the language here is effective, but could be improved in a future editorial PR

OR13: Need to define "shapes that are well-known", assuming some version of this PR will land:

https://github.com/w3c/vc-data-model/pull/1320

I can accept this suggestion as it exists.

index.html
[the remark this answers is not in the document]
OR13: > This section was never about doing that -- and that is not supported. This section is about telling people how to hard-code simple applications to specific, well-known contexts, making it more clear to newcomers that performing context transformations using a JSON-LD library that implements the JSON-LD API is not needed when you limit your data inputs to data that your application natively understands.

This section is about warning people that hardcoded contexts don't contain information that has been updated in the context, for which you don't know if the issuer intended to support or not.

I'm not sure anything in this thread is actionable, I suggest we resolve it and focus on concrete suggestions.

index.html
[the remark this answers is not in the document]
OR13: thanks, resolving.

index.html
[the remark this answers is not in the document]
OR13: Decline to implement feedback, although as I said in the other comments, while the current language is effective, it might be improved in a future editorial PR.

OR13: I'm accepting this suggestion, but its misleading in the case that a hard coded context does not match one for which errata has been published.

index.html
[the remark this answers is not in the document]
OR13: decline to implement feedback, but I think we may be able to adjust text in a future editorial pass

OR13: @dlongley  @msporny I am not intending to accept this suggestion, in the interest of getting this PR to mergable state, I suggest we resolve any threads that we feel don't need further discussion on the PR, if you need issue markers to resolve a thread, please file issues, and I will address all issues filed in a subsequent PR.

index.html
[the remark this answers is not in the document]
OR13: I think the suggestions that @dlongley made that I applied cover it. I am resolving this thread.

index.html
OR13: I don't think this is thread is actionable.

The context controls the shape of the RDF, not processing it (the latest version published by the authority, such as W3C) will lead to inconsistent processing.

I suggest we resolve this thread, or file seperate issues for follow up

index.html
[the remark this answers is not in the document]
OR13: I can live with this, but I think "minimal processing" is not well defined.

index.html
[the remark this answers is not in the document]
OR13: I think this has been addressed by now

OR13:
@msporny thanks, that view is much easier, I think I covered everything. I did not merge all suggestions.

I suggest we resolve everything that we can resolve, so we can see what is still blocking.

index.html
dlongley: @OR13,

> I can accept this suggestion as it exists.

Yes, please, thanks!

index.html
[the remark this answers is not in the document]
dlongley: Note that any exploit that relies on a non-conformant RDF-CANON implementation isn't relevant here. In that light, what is an "RDF processor" and what is an example of the kind of exploit you're referring to? We need to ensure we don't add unjustified fears / risks to the spec.

index.html
```diff
@@ -4297,19 +4320,61 @@ <h2>JSON Processing</h2>
...
+          <p>
+Implementers are advised that <code>id</code> and <code>type</code> have been
+aliased to <code>@id</code> and <code>@type</code>, and are marked <code>@protected</code>.
+Depending on the software libraries used, attempts to redefine these terms might raise processing errors.
+Implementers interested in understanding what <code>@id</code> means 
+are advised to review <a href="https://www.w3.org/TR/json-ld11/#node-identifiers">Node Identifiers</a>, and
+<a href="https://www.w3.org/TR/json-ld11/#specifying-the-type">Specifying the Type</a>.
+          </p>
+          <p>
+Implementers should be aware that object member, and array element order are not preserved by default in JSON-LD,
```
msporny, line 4363:
```suggestion
Implementers are advised that object member, and array element order are not preserved by default in JSON-LD,
```

index.html
msporny: I have tried to use "application-specific", which you use in your prose. No need to delay clarification. There is a concrete suggestion there, are you taking it, offering an alternative, or rejecting it?

dlongley: I have another suggestion here that @OR13 said he would accept:

https://github.com/w3c/vc-data-model/pull/1302/files#r1347713087

Can you apply that suggestion, @OR13?

index.html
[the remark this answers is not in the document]
msporny:
```suggestion
         <h3>Application-specific Processing</h3>
```

What about this, instead?

msporny (changes requested):
I have resolved all conversations that I could. There are still a handful of statements that are vague or inaccurate. Removal of the changes, or the statements, would result in a +1 to merge the PR. I still believe this PR is moving in an acceptable direction.

index.html
iherman: > The context controls the shape of the RDF, not processing it (the latest version published by the authority, such as W3C) will lead to inconsistent processing.

I do not understand this statement.

index.html
TallTed:
```suggestion
Implementers should be aware that the orders of object members and array elements are not preserved by default in JSON-LD;
```

index.html
```diff
@@ -4297,19 +4320,61 @@ <h2>JSON Processing</h2>
...
+and as such, Issuers intending to communicate array order need to leverage the language features of JSON-LD.
```
TallTed, line 4364:
```suggestion
therefore, <a>issuers</a> intending to communicate the orders of object
members and/or array elements might need to leverage language and/or other
advanced features of JSON-LD.
```

index.html
```diff
@@ -4297,19 +4320,61 @@ <h2>JSON Processing</h2>
...
+See <a href="https://www.w3.org/TR/json-ld11/#sets-and-lists">Value Ordering</a>.
+          </p>
+          <p>
+Implementers are advised that while minimal processing of a document make it appear that claims are related,
+the <code>https://www.w3.org/ns/credentials/v2</code> leverages <code>@container</code> and <code>@graph</code> to separate information graphs.
+In particular, to separate the graphs related to <a>verifiable credentials</a>, <a>verifiable presentations</a>,
+and <a href="#dfn-embedded-proof">embedded proofs</a>.
```
TallTed, line 4371:
```suggestion
Implementers are advised that while minimal processing of a document might
make it appear that claims are related,
the <code>https://www.w3.org/ns/credentials/v2</code> `@context` leverages
<code>@container</code> and <code>@graph</code> to separate information
graphs; in particular, these are used to separate the graphs related to
<a>verifiable credentials</a>, <a>verifiable presentations</a>, and
<a href="#dfn-embedded-proof">embedded proofs</a>.
```

index.html
```diff
@@ -4297,19 +4320,61 @@ <h2>JSON Processing</h2>
...
+          </p>
+          <p>
+Issuers and Holders might apply these same techniques to unambigiously communicate their intention regarding structured claims in 
+<a>verifiable credentials</a> and <a>verifiable presentations</a>. Verifiers that do not understand or process JSON-LD, will not 
+be aware of these differences, and confusion over the intention of the issuer and the holder could lead to unexpected processing behavior by verifiers.
```
TallTed, line 4376:
```suggestion
<a>Issuers</a> and <a>holders</a> might apply these same techniques to
unambiguously communicate their intention regarding structured claims in
<a>verifiable credentials</a> and <a>verifiable presentations</a>.
<a>Verifiers</a> that do not understand or process JSON-LD will not
be aware of these differences, and confusion between the intentions of the
<a>issuer</a> and the <a>holder</a> could lead to unexpected processing
behavior by <a>verifiers</a>.
```

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-11-14-vcwg-special#section1-3) on 2023-11-14

- no resolutions were taken

#### 1.3. Propose better JSON-LD processing text (pr vc-data-model#1302)

_See github pull request [vc-data-model#1302](https://github.com/w3c/vc-data-model/pull/1302)._




**Brent Zundel:** my understanding is that this is still not closer to consensus than it was. is this at all closer and worth talking about, or should we move on?  
… moving on.  

</details>

OR13:
I accepts all suggestions, but I have lost the energy to continue to work on this, I suggest the PR be closed.

index.html
```diff
@@ -4240,47 +4240,67 @@ <h2>HTTP</h2>
...
+When an application chooses to process the media type using the `+ld+json`
 structured media type suffix.
             </li>
           </ul>
 
           <p>
-That is, JSON processing is allowed as long as the document being consumed or
-produced is a <a>conforming document</a>. If JSON processing is desired, an
-implementer is advised to follow the following rule:
+Instead of performing general JSON-LD processing, application-specific processing is allowed as long as the document being consumed or
```
TallTed, line 4299: just a line break
```suggestion
Instead of performing general JSON-LD processing, application-specific
processing is allowed as long as the document being consumed or
```

index.html
```diff
@@ -4297,19 +4317,68 @@ <h2>JSON Processing</h2>
 implementing the rule above. This can ensure proper term identification,
 typing, and order, when a JSON document is processed as JSON-LD.
           </p>
+          <p>
+However, the intention of the issuer is not preserved, unless the rules expressed
+in the <code>@context</code> are followed, and the resulting <a href="#verifiable-credential-graphs">Verifiable Credential Graphs</a>, and their exact RDF types, and node and edge structure is understood by the verifier. By following the rules laid out in this section, applications can be hard-coded against
```
TallTed, line 4322: just line breaks
```suggestion
in the <code>@context</code> are followed, and the resulting
<a href="#verifiable-credential-graphs">Verifiable Credential Graphs</a>,
and their exact RDF types, and node and edge structure is understood
by the verifier. By following the rules laid out in this section,
applications can be hard-coded against
```

index.html
```diff
@@ -4297,19 +4317,68 @@ <h2>JSON Processing</h2>
...
+While some RDF processors will use the language features of JSON-LD and can verify that all
+terms are correctly specified, JSON-LD processors explicitly accept the same
+semantics without performing any RDF transformations, but instead by
 applying the above rules. In other words, the context in which the data exchange
-happens is explicitly stated for both JSON and JSON-LD by using the same
-mechanism. With respect to JSON-based processors, this is achieved in a
-lightweight manner, without having to use JSON-LD processing libraries.
+happens is explicitly stated for both RDF and JSON-LD by using the same
+mechanism. With respect to JSON-LD-based processors, this is achieved in a
+lightweight manner, without having to use RDF processing libraries.
```
TallTed, line 4335:
```suggestion
mechanism. JSON-LD-based processors achieve this in a lightweight manner,
without having to use RDF processing libraries.
```

index.html
```diff
@@ -4297,19 +4317,68 @@ <h2>JSON Processing</h2>
...
+          </p>
+        </section>
+        <section>
+          <h2>Processing RDF</h2>
+          <p>
+Issuers are advised that while conforming documents are expressed as compact JSON-LD,
+not all Holders or Verifiers will understand expressions of intention that are only visible 
+after RDF processing has occured.
```
TallTed, line 4343:
```suggestion
some <a>holders</a> and <a>verifiers</a> will not understand expressions of
intention that are only visible after RDF processing has occured.
```

index.html
```diff
@@ -4297,19 +4317,68 @@ <h2>JSON Processing</h2>
...
+          </p>
+          <p>
+Some of the best features of JSON-LD are only possible by applying its language features to the 
+terms defined in the <code>@context</code>.
```
TallTed, line 4347:
```suggestion
Some of the most powerful features of JSON-LD are only accessible by applying
its language features to the terms mapped in the <code>@context</code>.
```

index.html
```diff
@@ -4297,19 +4317,68 @@ <h2>JSON Processing</h2>
...
+          <p>
+Implementers are advised that <code>id</code> and <code>type</code> have been
+aliased to <code>@id</code> and <code>@type</code>, and are marked <code>@protected</code>.
+Depending on the software libraries used, attempts to redefine these terms might raise processing errors.
+Implementers interested in understanding what <code>@id</code> means 
+are advised to review <a href="https://www.w3.org/TR/json-ld11/#node-identifiers">Node Identifiers</a>, and
+<a href="https://www.w3.org/TR/json-ld11/#specifying-the-type">Specifying the Type</a>.
+          </p>
+          <p>
+Implementers are advised that object member, and array element order are not preserved by default in JSON-LD,
```
TallTed, line 4358:
```suggestion
Implementers are advised that JSON-LD does not preserve the orders of 
object members or array elements by default;
```

index.html
```diff
@@ -4297,19 +4317,68 @@ <h2>JSON Processing</h2>
...
+<a href="#dfn-embedded-proof">embedded proofs</a>.
+          </p>
+          <p>
+<a>Issuers</a> and <a>holders</a> might apply these same techniques to
+unambiguously communicate their intention regarding structured claims in
+<a>verifiable credentials</a> and <a>verifiable presentations</a>.
+<a>Verifiers</a> that do not understand or process JSON-LD will not
+be aware of these differences, and confusion between the intentions of the
+<a>issuer</a> and the <a>holder</a> could lead to unexpected processing
+behavior by <a>verifiers</a>.
```
TallTed, line 4380:
```suggestion
be aware of these differences, and confusion about the intentions of
<a>issuers</a> and/or <a>holders</a> could lead to unexpected processing
behavior by subsequent <a>holder</a> and/or <a>verifiers</a>, respectively.
```

brentzundel:
It is my understanding that this PR has been somewhat overtaken by events. Some of the changes sought here have already been included in other PRs, and there are a large number of conflicts that would need to be addressed in order to see what the remaining set of proposed changes looks like.

I am marking this PR `pending close` and will give it a week to see if anyone steps forward to try and clean it up in a way that is more likely to gain consensus.

BigBlueHat:
@brentzundel here is a PR that resolves the conflicts and brings in the text with @TallTed's suggestions:
https://github.com/w3c/vc-data-model/pull/1413

brentzundel:
Closing in favor of #1413

PR #1305, opened 2023-10-04
Remove CL Signature Example

OR13:
This PR adds a reference to vc-di-bbs, which is expected to be the only zkp credential system commented on by the WG, and is currently stalled, awaiting new leadership. 

See: 

- https://github.com/w3c/vc-di-bbs/pull/91
- https://github.com/w3c/vc-di-bbs/issues/84

I've retained normative statements surrounding the example, but removed the examples.

Ideally, we get some bbs examples in the spec here soon, although I personally think it would be better to drop the bbs work item, and focus on improving the specification and test suites related to the existing security mechansisms.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1305.html" title="Last updated on Oct 4, 2023, 10:32 PM UTC (0d7bdca)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1305/286c2b2...0d7bdca.html" title="Last updated on Oct 4, 2023, 10:32 PM UTC (0d7bdca)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -598,9 +598,6 @@ Securing Verifiable Credentials using JOSE and COSE [[VC-JOSE-COSE]].
           <li>
 Securing Verifiable Credentials using Data Integrity Proofs [[VC-DATA-INTEGRITY]].
           </li>
-          <li>
-Camenisch-Lysyanskaya Zero-Knowledge Proofs [[?CL-SIGNATURES]].
-          </li>
         </ul>
 
         <p>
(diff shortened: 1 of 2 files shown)


decentralgabe (approved):

index.html
```diff
@@ -3546,68 +3543,11 @@ <h3>Zero-Knowledge Proofs</h3>
...
-The next example utilizes the <a>verifiable credential</a> above to generate a
-new derived <a>verifiable credential</a> with a privacy-preserving proof. The
-derived <a>verifiable credential</a> is then placed in a
-<a>verifiable presentation</a>, so that the <a>verifiable credential</a>
-discloses only the <a>claims</a> and additional credential metadata that the
-<a>holder</a> intended. To do this, all of the following requirements are
-expected to be met:
+        <p class="issue">
+Examples of leveraging <a href="https://w3c.github.io/vc-di-bbs/">vc-di-bbs</a>,
+will be added here in the future, or this section will be removed.
```
BigBlueHat, line 3548: Maybe just leave this out until it's ready and open an issue instead?

If it stays here, though, it should get the "at-risk" label so it doesn't get overlooked during editing.

BigBlueHat (changes requested):
I don't know all the backstory here, but why is this content being removed and a "we may add this later" paragraph about something seemingly unrelated (to those not embedded in the work of both/either thing) being added in its place?

I could see this happening as too separate PRs with explanations instead--which might help the interplay here be clearer.

Thanks!
🎩

index.html
OR13: IMO this current text accomplishes the same goal, but happy to apply a suggestion if you have one, and the group likes it.

brentzundel (approved):

OR13:
multiple approvals, no objections, open for over 1 week.

Issue #1316, opened 2023-10-13
Update 4.9 Securing Verifiable Credentials


OR13:
It was requested to update this section to clarify that it is only relevant to Data Integrity and not VC JOSE COSE.

https://github.com/w3c/vc-data-model/pull/1308#pullrequestreview-1662838622

OR13:
![Screen Shot 2023-10-13 at 1 37 58 PM](https://github.com/w3c/vc-data-model/assets/8295856/9f3b7e04-5016-4dd6-905b-931cc9855aaa)

https://w3c.github.io/vc-data-model/#securing-verifiable-credentials

We didn't get all the change needed in #1308 and we will need to do a follow up PR to align the text above the examples

msporny:
> It was requested to update this section to clarify that it is only relevant to Data Integrity and not VC JOSE COSE.

That section is relevant to both Data Integrity and VC JOSE COSE. Not clear what this issue is attempting to address.

OR13:
@Sakurann can add details, essentially we removed proof from the first and second examples of the spec, the sections around those examples need to be updated, and since `proof` (the JSON-LD property of a conforming document) is only required  / recommended for use with data integrity proofs, this issue tracks completing the requested updates regarding `proof` that started on #1308

brentzundel:
Chair hat off, I do not agree with removing the proof property from VCDM.
I see it as an extension point, which has been utilized by Data Integrity, but which could be utilized by any other mechanism that wishes to. 

IMO, section 4.9 does not only apply to Data Integrity and it should not be modified to make it only apply to Data Integrity.

brentzundel:
What is the relationship between this issue and #1312 ?
Are both needed?

OR13:
I suggest we attempt to resolve this after all the "verificaton algorithm" issues are resolved

msporny:
PR #1389 has been raised to address the "before-CR" portion of this issue.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-12-19-vcwg-special#section2-8) on 2023-12-19

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 2.8. Clarify that "proof" is an extension point (pr vc-data-model#1389)

_See github pull request [vc-data-model#1389](https://github.com/w3c/vc-data-model/pull/1389)._




[...]

PR #1308, opened 2023-10-06
Remove "proof" clauses from initial examples

selfissued:
This PR needs to be merged to add the vc-jose-cose examples at https://github.com/w3c/vc-jose-cose/pull/162#pullrequestreview-1662677250

***
<a href="https://pr-preview.s3.amazonaws.com/selfissued/vc-data-model/pull/1308.html" title="Last updated on Oct 13, 2023, 6:12 PM UTC (7878d79)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1308/286c2b2...selfissued:7878d79.html" title="Last updated on Oct 13, 2023, 6:12 PM UTC (7878d79)">Diff</a>

PR #1317, opened 2023-10-13
Additional updates regarding proof

OR13:
This PR is follow up on:  https://github.com/w3c/vc-data-model/pull/1308

And addresses comments tracked in: https://github.com/w3c/vc-data-model/issues/1316

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1317.html" title="Last updated on Oct 13, 2023, 10:44 PM UTC (118b6ba)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1317/d3568d7...118b6ba.html" title="Last updated on Oct 13, 2023, 10:44 PM UTC (118b6ba)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2132,13 +2132,6 @@ If the <code>holder</code> <a>property</a> is absent, information about the
 <a>holder</a> is expected to either be obtained via the securing mechanism, or
 to not pertain to the <a>validation</a> of the <a>verifiable presentation</a>.
           </dd>
-          <dt><var>proof</var></dt>
-          <dd>
-The <a>verifiable presentation</a> MAY include a <code>proof</code>
-<a>property</a>. If present, the value SHOULD be used to express a securing
-mechanism, such as those listed in the
-Verifiable Credentials Specifications Directory [[?VC-SPECS]].
-          </dd>
         </dl>
 
         <p>
@@ -2153,8 +2146,7 @@ The example below shows a <a>verifiable presentation</a>:
   ],
   "id": "urn:uuid:3978344f-8596-4c3a-a978-8fcaba3903c5",
   "type": ["VerifiablePresentation", "ExamplePresentation"],
-  <span class="highlight">"verifiableCredential": [{ <span class="comment">...</span> }],
-  "proof": [{ <span class="comment">...</span> }]</span>
+  <span class="highlight">"verifiableCredential": [{ <span class="comment">...</span> }]
 }
         </pre>
 
--- a/index.html
+++ b/index.html
@@ -1959,11 +1959,6 @@ mutually exclusive.
 <a>Verifiable credentials</a> and <a>verifiable presentations</a> MUST be secured by at least
 one securing method.
         </p>
-        <p>
-Methods of securing <a>verifiable credentials</a> or <a>verifiable
-presentations</a> that use an external proof MAY use the <code>proof</code>
-<a>property</a>.
-        </p>
         <dl>
           <dt><var>proof</var></dt>
           <dd>
--- a/index.html
+++ b/index.html
@@ -1961,15 +1961,7 @@ one securing method.
         </p>
         
 
-        <p>
-Because the method used for a mathematical proof varies by representation
-language and the technology used, the set of name-value pairs that is expected
-as the value of the <code>proof</code> <a>property</a> will vary accordingly.
-For example, if digital signatures are used for the proof mechanism, the
-<code>proof</code> <a>property</a> is expected to have name-value pairs that
-include a signature, a reference to the signing entity, and a representation of
-the signing date. The example below uses Ed25519 digital signatures.
-        </p>
+        
 
         <pre class="example nohighlight"
           title="Usage of the proof property on a verifiable credential">
--- a/index.html
+++ b/index.html
@@ -1959,15 +1959,7 @@ mutually exclusive.
 <a>Verifiable credentials</a> and <a>verifiable presentations</a> MUST be secured by at least
 one securing method.
         </p>
-        <dl>
-          <dt><var>proof</var></dt>
-          <dd>
-One or more cryptographic proofs that can be used to detect tampering and verify
-the authorship of a <a>verifiable credential</a> or <a>verifiable
-presentation</a>. The specific method used for an <a>embedded proof</a> MUST be
-identified using the <code>type</code> <a>property</a>.
-          </dd>
-        </dl>
+        
 
         <p>
 Because the method used for a mathematical proof varies by representation
--- a/index.html
+++ b/index.html
@@ -2272,7 +2272,7 @@ self-asserted <a>verifiable credential</a> that is secured using the same
 mechanism as the <a>verifiable presentation</a>.
           </p>
 
-          <pre class="example nohighlight" title="A verifiable presentation, secured with an embedded proof, with a self-asserted verifiable credential">
+          <pre class="example nohighlight" title="Contents of a verifiable presentation, with a self-asserted verifiable credential">
 {
   "@context": [
     "https://www.w3.org/ns/credentials/v2",
@@ -2289,7 +2289,7 @@ mechanism as the <a>verifiable presentation</a>.
     },
     { <span class="comment">...</span> }
   }],
-  "proof": [{ <span class="comment">...</span> }]
+  <span class="comment">...</span>
 }
           </pre>
           <p>
@@ -2299,7 +2299,7 @@ self-asserted <a>verifiable credential</a> that holds <a>claims</a> about the
 <a>verifiable presentation</a>.
           </p>
 
-          <pre class="example nohighlight" title="A verifiable presentation, secured with an embedded proof, with a self-asserted verifiable credential about the verifiable presentation">
+          <pre class="example nohighlight" title="Contents of a verifiable presentation, with a self-asserted verifiable credential about the verifiable presentation">
 {
   "@context": [
     "https://www.w3.org/ns/credentials/v2",
@@ -2318,7 +2318,7 @@ self-asserted <a>verifiable credential</a> that holds <a>claims</a> about the
     },
     { <span class="comment">...</span> }
   }],
-  "proof": [{ <span class="comment">...</span> }]
+  <span class="comment">...</span>
 }
           </pre>
         </section>
@@ -2462,7 +2462,7 @@ see Section <a href="#zero-knowledge-proofs"></a>.
     "id": "https://example.org/examples/degree",
     "type": "ZkpExampleSchema2018"
   }</span>,
-  "proof": { <span class="comment">...</span> }
+  <span class="comment">...</span>
 }
         </pre>
 
@@ -3342,7 +3342,7 @@ Verifiable Credentials Implementation Guidelines [[VC-IMP-GUIDE]] document.
     "documentPresence": "Physical",
     "licenseNumber": "123AB4567"
   }]</span>,
-  "proof": { <span class="comment">...</span> }
+  <span class="comment">...</span>
 }
         </pre>
 
--- a/index.html
+++ b/index.html
@@ -1960,9 +1960,6 @@ mutually exclusive.
 one securing method.
         </p>
         
-
-        
-
         <pre class="example nohighlight"
           title="Usage of the proof property on a verifiable credential">
 {
--- a/index.html
+++ b/index.html
@@ -1960,11 +1960,6 @@ mutually exclusive.
 one securing method.
         </p>
         <p>
-Methods of securing <a>verifiable credentials</a> or
-<a>verifiable presentations</a> that embed a proof in the data model MUST use
-the <code>proof</code> <a>property</a>.
-        </p>
-        <p>
 Methods of securing <a>verifiable credentials</a> or <a>verifiable
 presentations</a> that use an external proof MAY use the <code>proof</code>
 <a>property</a>.
--- a/index.html
+++ b/index.html
@@ -1956,8 +1956,7 @@ It should be noted that these two classes of securing mechanisms are not
 mutually exclusive.
         </p>
         <p>
-<a>Verifiable credentials</a> and <a>verifiable presentations</a> MUST be secured by at least
-one securing method.
+<a>Verifiable credentials</a> MUST be secured by at least one securing method.
         </p>
         
         <pre class="example nohighlight"


index.html
```diff
@@ -1956,39 +1956,9 @@ <h3>Securing Verifiable Credentials</h3>
 mutually exclusive.
         </p>
         <p>
-<a>Verifiable credentials</a> and <a>verifiable presentations</a> MUST be secured by at least
```
OR13, line 1959: presentations have not ever required at least one proof historically, I believe this was accidentally changed in a previous PR.

index.html
```diff
@@ -1956,39 +1956,9 @@ <h3>Securing Verifiable Credentials</h3>
...
-one securing method.
+<a>Verifiable credentials</a> MUST be secured by at least one securing method.
         </p>
-        <p>
-Methods of securing <a>verifiable credentials</a> or
-<a>verifiable presentations</a> that embed a proof in the data model MUST use
```
OR13, line 1964: This is data integrity proof specific, and should be a normative statement in that specification.

index.html
```diff
@@ -1956,39 +1956,9 @@ <h3>Securing Verifiable Credentials</h3>
...
-the <code>proof</code> <a>property</a>.
-        </p>
-        <p>
-Methods of securing <a>verifiable credentials</a> or <a>verifiable
-presentations</a> that use an external proof MAY use the <code>proof</code>
```
OR13, line 1969: Removing this sentence does not change what implementations have always been allowed to do.

index.html
```diff
@@ -1956,39 +1956,9 @@ <h3>Securing Verifiable Credentials</h3>
...
-<a>property</a>.
-        </p>
-        <dl>
-          <dt><var>proof</var></dt>
```
OR13, line 1973: as proof is an attribute of data integrity secured conforming documents, it should be defined in the data integrity spec.

index.html
```diff
@@ -1956,39 +1956,9 @@ <h3>Securing Verifiable Credentials</h3>
...
-          <dd>
-One or more cryptographic proofs that can be used to detect tampering and verify
-the authorship of a <a>verifiable credential</a> or <a>verifiable
-presentation</a>. The specific method used for an <a>embedded proof</a> MUST be
```
OR13, line 1977: This is a normative requirement of data integrity.

index.html
```diff
@@ -1956,39 +1956,9 @@ <h3>Securing Verifiable Credentials</h3>
...
-        </dl>
-
-        <p>
-Because the method used for a mathematical proof varies by representation
-language and the technology used, the set of name-value pairs that is expected
-as the value of the <code>proof</code> <a>property</a> will vary accordingly.
-For example, if digital signatures are used for the proof mechanism, the
-<code>proof</code> <a>property</a> is expected to have name-value pairs that
-include a signature, a reference to the signing entity, and a representation of
-the signing date. The example below uses Ed25519 digital signatures.
```
OR13, line 1989: extraneous text, that is not longer relevant to the example below

index.html
```diff
@@ -2162,13 +2132,6 @@ <h3>Presentations</h3>
 <a>holder</a> is expected to either be obtained via the securing mechanism, or
 to not pertain to the <a>validation</a> of the <a>verifiable presentation</a>.
           </dd>
-          <dt><var>proof</var></dt>
```
OR13, line 2165: proof can be present in ANY JSON-LD object secured by data integrity, not just the "VerifiablePresentation" RDF class.

index.html
```diff
@@ -2162,13 +2132,6 @@ <h3>Presentations</h3>
...
-          <dd>
-The <a>verifiable presentation</a> MAY include a <code>proof</code>
```
OR13, line 2167: redundant text, implementers are already allowed to do this.

index.html
```diff
@@ -2162,13 +2132,6 @@ <h3>Presentations</h3>
...
-<a>property</a>. If present, the value SHOULD be used to express a securing
```
OR13, line 2168: normative statement point to non normative note.... and specific to data integrity.

index.html
```diff
@@ -2183,8 +2146,7 @@ <h3>Presentations</h3>
   ],
   "id": "urn:uuid:3978344f-8596-4c3a-a978-8fcaba3903c5",
   "type": ["VerifiablePresentation", "ExamplePresentation"],
-  <span class="highlight">"verifiableCredential": [{ <span class="comment">...</span> }],
-  "proof": [{ <span class="comment">...</span> }]</span>
```
OR13, line 2187: removed empty proof from example

index.html
```diff
@@ -2310,7 +2272,7 @@ <h4>Presentations Including Holder Claims</h4>
 mechanism as the <a>verifiable presentation</a>.
           </p>
 
-          <pre class="example nohighlight" title="A verifiable presentation, secured with an embedded proof, with a self-asserted verifiable credential">
```
OR13, line 2313: updated title to be consistent with previous example title changes

index.html
```diff
@@ -2327,7 +2289,7 @@ <h4>Presentations Including Holder Claims</h4>
     },
     { <span class="comment">...</span> }
   }],
-  "proof": [{ <span class="comment">...</span> }]
```
OR13, line 2330: replace proof with ellipses, many other optional properties could be present here.

selfissued (approved):
This continues moving us in the right direction of being securing-format agnostic.

index.html
decentralgabe: do they have any utility without a proof? are they even verifiable (no)?

decentralgabe (approved):

index.html
dlongley: Proof isn't specific to data integrity, but rather data integrity is a particular implementation an embedded proof. This is similar to "credentialStatus", "evidence", etc. So I think any changes in this PR that treat it differently from that should probably be adjusted back / removed.

dlongley (changes requested):

dlongley:
We have a number of properties in the spec that we've been referring to as "extension points" where people have complained about a lack of information in the core spec about the general use of those properties. These particular properties have been targeted because they did not have well defined or interoperable specific uses -- and there's been a desire, as we document these specific uses, to ensure that the spec sufficiently speaks about them and their general purpose.

The "proof" property is sort of the inversion of that situation. It is a property that has been in use for years with multiple interoperable implementations for specific use and it has had statements about its general properties and purpose in the spec.

This PR seems to be undoing what we aim to do with other properties to make them better documented. I don't think the fact that you could "do something external to secure a VC / link it to a proof" should cause us to remove documentation here or change this property -- nor do I think it should be changed just because we have a well documented specific use of the property. Again, this upends what we try to do with everything else.

What if someone invents an external way to attach "evidence" or "terms of use" or "credential status" to a VC -- should we go back and strip out everything about those properties here? This seems like the wrong approach.

Now, if some of the text changes here are correcting bugs, those are fine, but we shouldn't be undoing the very same kind of work we strive to do in other places.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-10-17-vcwg-special#section2-11) on 2023-10-17

- no resolutions were taken

#### 2.11. Additional updates regarding proof (pr vc-data-model#1317)

_See github pull request [vc-data-model#1317](https://github.com/w3c/vc-data-model/pull/1317)._




**Dave Longley:** This PR does a number of different things at once, it might be fixing some bugs in the spec, that part of the PR is good.  
… We also have this inconsistency that's related to the other part of the PR -- this PR also removes the proof property in core data model based on idea that there are multiple ways to secure a VC -- external mechanism, or internal mechanism, comments on PR are...  
… we've been talking a lot about different properties of spec... extension points -- we want to make sure we want to say what an extension point does/is in the core spec and we only keep it around/reserved if we have a specific use of it in other specs. `proof` is just another extension point -- it happened in an inverted way, we've had proof around and we've always had it in Data Integrity and now we're defining it in the spec. We had proof, had good statements about it, and had specific implementations/tests around it -- we have all the things we said we want to have in the data model.  
… We have now said "the external thing exists" and we should remove the property... we have said we want more documentation around points of extension -- so I think we're going in the wrong direction w/ part of this PR. What happens if we get an external mechanism for credentialStatus, would we remove it? Goes for any properties we have. We want to talk about properties generally in core spec, and make sure we have other properties in other specs that talk about them in specific ways.  

**Ivan Herman:** I don't know if this is addressed in the PR of Orie, I didn't look at it. There is one thing that personally bothered me for a long time. "proof" as a term is formally defined in the DI vocabulary. It's unclear where really it belongs. It's in the two at the same time. That might be a reason for the confusion.  

**Manu Sporny:** I did look at this PR -- just didn't respond yet. I agree, there are good things in this PR and largely agree that proof is an extension point we need to document. This process is like a "feeler PR" that goes in and then there are other PRs that go in and remove more.  
… The concern here is that the VCDM becomes just about the data and not securing it -- and lots of people have complained about the VCDM not allowing you to do anything, I think we should have concrete securing examples. We should have both data integrity and VC-JOSE-COSE examples in the spec.  
… There seems to be an undercurrent in the PR to stop talking about how to verify a VC and that's a mistake, IMO. If those parts are taken out, +1 to the PR.  

**Brent Zundel:** Would someone be willing to summarize what the next step would be for this? I recognize that this PR is only a week old... we're still in the window, 4 days. It hasn't been languishing, we still need to know what the concrete path forward is.  
… Do we see hope for consensus?  
… Do we need Orie present?  

**Manu Sporny:** Orie needs to be present to see what he finds acceptable.  
… If this PR focused on bug fixes, Orie is correct to fix some bugs here -- I would expect that PR to go in.  
… If it just does that, it would be ok. I think a PR that removes proof from the spec wouldn't work, but talking about proof as an extension point might address some of his concerns.  
… The examples where we have a proof with a bunch of ellipsis statements and it may be fine to remove those things if replaced with examples showing how to secure with DI and with VC-JOSE-COSE.  
… So those things, I think, could possibly get to consensus.  

**Brent Zundel:** Thank you all, thanks, Dave for scribing, thanks for all the review, please be timely, we do need to wrap things up. Hope will see you all on the call tomorrow.  

</details>

index.html
msporny: Why is it no longer relevant?

index.html
msporny: `proof` is an extension point in the VC specification, just like `credentialStatus` and `issuer`. Conflating Data Integrity and `proof` as the same thing is a mistake. Verifiable Presentations have an extension point for embedded proofs, that extension point is `proof`. Verifiable Credentials have an extension point for embedded proofs, that extension point is `proof`.

-1 to removing it from Verifiable Presentation.

index.html
msporny: Split the sentence into two sentences, then. 

-1 for removing the statement.

index.html
msporny: Then convert this into a real example that shows a minimal proof using both securing mechanisms. We don't gain anything by NOT showing people what a secured a VC or VP looks like.

index.html
msporny: -1 for removing securing examples from the specification. Please convert into an example that shows how it's secured.

index.html
msporny: Then we should show what a secured VP looks like.

index.html
```diff
@@ -2356,7 +2318,7 @@ <h4>Presentations Including Holder Claims</h4>
...
```
msporny, line 2359: Replace w/ real secured examples.

index.html
```diff
@@ -2500,7 +2462,7 @@ <h3>Data Schemas</h3>
     "id": "https://example.org/examples/degree",
     "type": "ZkpExampleSchema2018"
   }</span>,
-  "proof": { <span class="comment">...</span> }
```
msporny, line 2503: Replace w/ real secured examples.

index.html
```diff
@@ -3380,7 +3342,7 @@ <h3>Evidence</h3>
     "documentPresence": "Physical",
     "licenseNumber": "123AB4567"
   }]</span>,
-  "proof": { <span class="comment">...</span> }
```
msporny, line 3383: Replace w/ real secured examples.

msporny (changes requested):
This PR removes a number of items from the specification based on a confused interpretation of the `proof` value as something directly tied to Data Integrity, when it is not -- it's an extension point for any embedded proof mechanisms. I've gone through and +1'd the things that are good changes and -1'd the items that are problematic. If we narrowed the scope of this PR to the +1'd items, we could merge it.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-10-31-vcwg-special#section1-5) on 2023-10-31

- no resolutions were taken

#### 1.5. Additional updates regarding proof (pr vc-data-model#1317)

_See github pull request [vc-data-model#1317](https://github.com/w3c/vc-data-model/pull/1317)._




**Brent Zundel:** This PR was raised by Orie.  
… It modifies the proof section. It gets rid of a lot of it.  
… I believe makes it apply only to data integrity mechanisms for securing the spec. I may be inaccurate on that.  

**Manu Sporny:** Yeah, some of the changes in this PR are ok, others I objected to. I think we can clean this up in a better way, but this takes a hatchet to the spec in a way that's unacceptable. I suggest we mark this pending close.  
… I doubt it's going to get into a mergeable state any time soon.  

**Brent Zundel:** Chair hat off, I don't agree with the direction this PR has taken. It's always been my position that "proof" is an extension point that any securing mechanism can use and just because data integrity does already doesn't mean others now can't or never can.  
… Chair hat back on, the plan is to mark this as pending close.  

> *Gabe Cohen:* +1 TBD has a use case to make use of the proof property without using DI proofs.

> *Manu Sporny:* ^ excellent, so 3 uses of proof property extension point, only one of them is Data Integrity.

**Manu Sporny:** Agree, also wanted to note that the CL signature stuff that province of BC is using this -- either they will use the data integrity pattern or their own, and they are currently using their own.  

**Brent Zundel:** I'm going to mark this pending close due to my determination that consensus has not been reached and it seems unlikely that it will be reached.  

</details>

brentzundel:
As discussed in the WG call on 31 October, this PR is not on a path toward consensus and is being closed.

PR #1320, opened 2023-10-18
Add definition of well formed

OR13:
We refer to "conforming documents" and "well formed documents", it seems a term definition for a well formed document will be helpful.

### DIFF
--- a/terms.html
+++ b/terms.html
@@ -185,4 +185,10 @@ or fetching, a URL are defined by the URL [=url/scheme=]. This specification
 does not use the term URI or IRI because those terms have been deemed to be
 confusing to Web developers.
   </dd>
+  <dt><dfn data-lt="well-formed">well-formed</dfn></dt>
+  <dd>
+A well-formed document is a conforming document, 
+and an instance of the core data model defined via a context,
+which produces no JSON-LD processing errors.
+  </dd>
 </dl>


terms.html
```diff
@@ -185,4 +185,10 @@
 does not use the term URI or IRI because those terms have been deemed to be
 confusing to Web developers.
   </dd>
+  <dt><dfn data-lt="well-formed">well-formed</dfn></dt>
+  <dd>
+A well-formed document is a conforming document, 
+and an instance of the core data model defined via a context,
+which produces no JSON-LD processing errors.
```
jandrieu, line 192: This definition is in conflict with the definition of well-formed relative to the credentialSchema property.

Here's the definition from wikipedia for xml well-formedness https://en.wikipedia.org/wiki/Well-formed_document :
> adheres to the [syntax](https://en.wikipedia.org/wiki/Syntax) rules specified by the [XML](https://en.wikipedia.org/wiki/XML) 1.0 specification in that it must satisfy both physical and logical structures

Perhaps more odd in your definition is the inclusion of JSON-LD processing, which is orthogonal to well-formedness. In the XML notion of well-formedness, the semantics are not included, just the syntax. 

> The rules of [validity](https://en.wikipedia.org/wiki/Validity_(logic)) mean that a document complies with the restraints stated within a DTD.

The semantics are a matter of validity, not syntax.

I suggest
```suggestion
A well-formed Verifiable Credential is a conforming document which satisfies the syntax defined by its credentialSchema property, if present.
```

msporny: Hrm, this isn't just about VCs, it's about all document types specified by the spec (VCs and VPs).

msporny, line 192: We should work well-formed-ness into the definition of a conforming document, and then update all mentions of "well-formed" in the specification to "conforming document". Let's try to reduce vocabulary here, if possible.

msporny (changes requested):
Hrm, we only use the word "well formed" (as it relates to our spec) in one place in the specification:

https://w3c.github.io/vc-data-model/

Let's just get rid of the usage of that word and/or wrap it into the definition of "conforming document".

Separately, we can just say that a conforming document MUST also be a conforming JSON-LD Document as defined in JSON-LD 1.1... and really, that's already true because of the media type. IOW, I don't think we need to say what a well formed document is... and if we feel we need to do that, we can just point to the JSON-LD11 spec from the "conforming document" definition.

So, let's delete "well formed" (so we don't have to define it). I don't really think we need to do much else than that. We could re-visit the definition of "conforming document", though I'm pretty sure it already requires that it "produces no JSON-LD processing errors".

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-10-31-vcwg-special#section1-4) on 2023-10-31

- no resolutions were taken

#### 1.4. Add definition of well formed (pr vc-data-model#1320)

_See github pull request [vc-data-model#1320](https://github.com/w3c/vc-data-model/pull/1320)._




**Brent Zundel:** Adds a definition of "well-formed". Last I looked, there was a difference of opinion about whether it should be added at all.  

**Manu Sporny:** I am objecting to this PR. We use the word "well-formed" once in the specification and should just delete it. Conforming documents should be well formed and it would be weird if they were ill formed. We should just use "conforming document", and I'm happy to raise a PR to do that. I don't think it's useful to have two different definitions.  
… I think we should mark this pending close.  

**Michael Jones:** I'm fine with Manu's suggestion to just delete the use of the word.  

**Joe Andrieu:** This originally came out of discussions around adherence to the credentialSchema -- does "well-formedness" and conformance are good enough for that?  

**Manu Sporny:** That's a good point, we should figure out the language we want to use there ... like "the document matches the schema".  

**Joe Andrieu:** "Well-formed" is what I used for that.  

> *Phillip Long:* pdl-asu has joined #vcwg-special.

**Manu Sporny:** I don't know -- I feel that the discussion around that will take some time ... I don't think defining it until we have the discussion and decide if we even need it is premature. There's one use of the word in the spec and we've written a lot without it.  
… If we want to introduce the word we'll have to define it and whether it refers to credentialSchema, a JSON-LD doc, so on. I think it's premature.  

**Ted Thibodeau Jr.:** Just to say that there are two instances of hyphenated "well-formed" in the spec.  

> *Manu Sporny:* Thanks Ted, I didn't check for hyphenated forms of the word... yes, get rid of all of them.

**Ted Thibodeau Jr.:** I think it's better to get rid of all if we get rid of one.  

**Gabe Cohen:** I think JSON schema is a good tool for checking "well-formedness". Another way I've framed it is spec-compliant. Some way of capturing this is important.  

**Brent Zundel:** Can you say more about this PR in particular?  

**Gabe Cohen:** I need to take a close look at it.  

**Manu Sporny:** Gabe -- the word we use for spec-compliant is "conforming document", it's defined as a conformance class.  
… I'm looking at the landing page in the JSON schema docs and it doesn't say "well-formed" -- it says things like "validate", "JSON data consistency", "validity", etc.  
… I think the PR is premature... if we want to focus on the PR, we're having a conversation around it and that tells me we need more conversation.  

**Joe Andrieu:** I agree, maybe we need more conversation before merging this in. This came from being able to define "how you validate" and there's a narrative loop and we can't get out of it by just saying validate again, but we need more conversation.  

**Brent Zundel:** Ok, marking this PR as pending close... more conversation needed.  

**Joe Andrieu:** Sorry, don't agree with pending close.  

**Brent Zundel:** Oh, I thought that's where we're at, go ahead Manu.  

**Manu Sporny:** I think we communicate something to the group -- the editors are deeply concerned with how long it's taking to pulling in PRs and process issues to the point that we're putting the core spec in danger -- by marking things as pre-CR and then saying we'll discuss later.  
… Please correct me if I'm wrong on the timeline, chair/staff contact, we need to speed things up. We are well past the date we wanted for getting the core spec into CR. As a result, these PRs that gather 100s of comments and stay out for a long time. My personal opinion as an editor -- we don't have time to process things in that manner any more.  
… The chairs have been warning us that we have to make some hard decisions on what we can address and what we can't -- which means we have to get more aggressive on whether a PR will make it in or not. If we have people objecting to PRs no matter who they are, that makes it highly likely it will be closed. Granted, there are still things we have to do -- and have to discuss, especially if things will result in a formal objection. But a lot of PRs are stuck in the queue.  
… Marking as pending close is a way to try and get us to the point where we can go to CR with the VCDM 2.0 spec. That's my view on why we're processing issues and PRs in that manner, please correct me if I'm wrong.  

**Brent Zundel:** Thank you, Manu, I should have taken a few moments in the start of the call to say that. We're a month over due, things without a clear path to consensus we're going to close. If consensus can be closed, great, we can try another PR, but we just don't have time to have long discussions that we hope to get in.  

**Joe Andrieu:** Frankly, I don't mind getting rid of this PR, but what I heard is that we need more discussion and that we want to remove well-formed elsewhere. I heard you keeping it alive, Manu and more work. I'm happy to get rid of it and leaving two forms of well-formed in the spec that are undefined in the spec.  

**Brent Zundel:** This is non-normative, arguable it's normative but I don't think so, I'd like to mark it post-CR and we can discuss later.  

**Manu Sporny:** My hopes and dreams are that we get full discussion with all the things and everyone is happy with what goes into the spec. The tough reality is we have a ticking clock and we're behind and we have to decide what to do talk about, what to hold nose on, and what to close or defer.  

**Brent Zundel:** Thank you, Manu.  

</details>

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-11-14-vcwg-special#section1-4) on 2023-11-14

- no resolutions were taken

#### 1.4. Add definition of well formed (pr vc-data-model#1320)

_See github pull request [vc-data-model#1320](https://github.com/w3c/vc-data-model/pull/1320)._




**Brent Zundel:** another post cr, don't believe any changes have been made since last discussion.  
… propose we move on, as there doesn't seem to be additional consensus.  

**Joe Andrieu:** my sense from last week was that we were headed towards closing this, should we add pending close?  

**Brent Zundel:** you are not wrong, that's why we added post-cr in case we see a path towards consensus later that hadn't been available before.  

**Joe Andrieu:** I'm okay with that.  

</details>

brentzundel:
Does anyone see a path toward consensus on this PR?
I am going to mark it `pending close`, see where the conversation goes, and make a determination in a week.

Issue #1248, opened 2023-08-20
signatures should be on quoted graphs


bblfish:
In [issue 204 on the N3 CG](https://github.com/w3c/N3/issues/204#issuecomment-1685313961), I tried to translate 2 VC model examples. This revealed a  few minor errors and some deeper conceptual problems, which are much easier to see in N3 than in Json-LD. These conceptual problems revolve around what is in a graph context and what is not. The JsonLD seems to place the contexts incorrectly, leading to unclarity about what is and what is not signed.

**Update:** the problem is perhaps most concisely expressed in [a comment below in response to @msporny's feedback](https://github.com/w3c/vc-data-model/issues/1248#issuecomment-1693975080).

The [N3 notation](https://w3c.github.io/N3/reports/20230703/) is to NQuads what Turtle is to NTriples: it makes it easy to read and reduces repetition for the author and the reader so that one can easily get an overview of what is being said. N3 makes it easy to read and write contexts. It also adds many other things, such as the ability to write rules for logical inferences... Where JsonLD makes it easy for developers happy with JSON to use the result, N3 makes it easy to see the data structure and reason about it. It especially makes it easy to see contexts, which are very difficult to see in jsonld.

##  1. Mapping the JSonLD of example 1 to N3

To take [example 1](https://www.w3.org/TR/vc-data-model/#example-a-simple-example-of-a-verifiable-credential), which I placed here https://bblfish.net/tmp/2023/08/vcdm.ex1.jsonld the result after fixing a few errors with namespaces, explained in detail in   [N3 issue 204](https://github.com/w3c/N3/issues/204#issuecomment-1685313961)  is that it gives me the following N3:

```Turtle
@prefix sec:     <https://w3id.org/security#> .
@prefix cred:    <https://www.w3.org/2018/credentials#> .
@prefix eg:      <https://example.org/examples#> .
@prefix xsd:     <http://www.w3.org/2001/XMLSchema#> .
@prefix rdfs:    <http://www.w3.org/2000/01/rdf-schema#> .

<did:example:c276e12ec21ebfeb1f712ebc6f1>
    <http://schema.org/name>
        "Exemple d'Université"@fr , 
        "Example University"@en .

<did:example:ebfeb1f712ebc6f1c276e12ec21>
    <http://schema.org/alumniOf>
        <did:example:c276e12ec21ebfeb1f712ebc6f1> .

<http://example.edu/credentials/1872>
[...]

PR #1321, opened 2023-10-19
Replacement of the diagrams (figure 6 and 8), and their descriptions

iherman:
Both [Figure 6](https://www.w3.org/TR/vc-data-model-2.0/#info-graph-vc) and [Figure 8](https://www.w3.org/TR/vc-data-model-2.0/#info-graph-vc), and their corresponding description in the text, are incorrect and misleading insofar as to where a graph is used and where it isn't. This PR proposes alternative diagrams for both, as well changes in the corresponding descriptions.

Note that we may have to consider making a distinction in the terminology for a "collection of claims", which has no additional structure, and a "graph", that has a distinct identity in the RDF sense. I know this may lead to some editorial complexity, but the misunderstanding between, on the one hand, just a loose combination of claims (a.k.a. RDF statements) and, on the other hand, a (named) graph in the RDF sense was the source of several confusions in the past, and the figures were just a reflection of that confusion. That being said, the proposed changes are all in an informative section (hence my labeling set to post CR), and maybe there is no need for further changes elsewhere. 

(As an aside, the diagrams are deliberately poor in colours, to make them usable for dark mode presentation style.)

Alas! the official, generated preview below is not usable here, because it does not include the diagrams. Please, use the [preview on githack](https://raw.githack.com/w3c/vc-data-model/updating-explanatory-figures/index.html#credentials) instead.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1321.html" title="Last updated on Oct 19, 2023, 3:29 PM UTC (3110016)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1321/87a4db3...3110016.html" title="Last updated on Oct 19, 2023, 3:29 PM UTC (3110016)">Diff</a>


iherman:
> Note that we may have to consider making a distinction in the terminology for a "collection of claims", which has no additional structure, and a "graph", that has a distinct identity in the RDF sense. I know this may lead to some editorial complexity, but the misunderstanding between, on the one hand, just a loose combination of claims (a.k.a. RDF statements) and, on the other hand, a (named) graph in the RDF sense was the source of several confusions in the past, and the figures were just a reflection of that confusion. That being said, the proposed changes are all in an informative section (hence my labeling set to post CR), and maybe there is no need for further changes elsewhere.

Alternatively... we can continue using the terminology of a graph everywhere (instead of this loose term of "collection of claims"), in which case we can also refer to a "Verifiable Credential Graph" [§4.11.1](https://www.w3.org/TR/vc-data-model-2.0/#verifiable-credential-graphs) in the description (and we would have to make this explicit in Figure 6, too). However, I believe that, to make it proper, we have to bring in the terminology of a [default graph](https://w3c.github.io/rdf-concepts/spec/#dfn-default-graph). I.e., if a JSON-LD document expresses a verifiable credential, then the Verifiable Credential Graph is its default graph (which clearly separates it from the proof graph). If the document expresses a Verifiable Presentation, then the Verifiable Credential Graphs are "created" by virtue of using the `verifiableCredential` property, whose definition in the `@context` is such that its value is a [named graph](https://w3c.github.io/rdf-concepts/spec/#dfn-named-graph) (incidentally, these named graphs have all blank nodes as a "name", but that is a detail). Note that we also have then a Verifiable Presentation Graph (which is the default graph of the relevant document).

By using this terminology, things may become clearer. I know we wanted to hide the complexity of RDF Datasets but this seems to backfire, so maybe we have to add this term into the spec.

Note that this makes the definition of `proof`, when used on VC/VPs, possibly clearer. Indeed:

- The `proof` property, referred to by an object of type Verifiable Credential, is a proof of the corresponding Verifiable Credential Graph 
- The `proof` property, referred to by an object of type Verifiable Presentation, is a proof of the corresponding Verifiable Presentation Graph, the collection of Verifiable Credential Graphs referred to from the VP, and the Verifiable Credential Proof Graphs, referred to from the aforementioned Verifiable Credentials Graphs.

At this moment, these statements are not clearly defined in the spec imho, which I think is at the core of the discussions in #1248.

@dlongley @msporny

iherman:
@msporny @dlongley, going along the lines of https://github.com/w3c/vc-data-model/pull/1321#issuecomment-1772083341 would require more surgery on the VCDM spec. If you agree to go down that route, I am happy to create a separate PR going along those lines (which would supersede this one).

iherman:
> @iherman,
> 
> I do think following comment [#1321 (comment)](https://github.com/w3c/vc-data-model/pull/1321#issuecomment-1772083341) is probably best because it clears up some of the technical details here and highlights how we're relying on default graphs for isolation / separation in documents that either express VPs or VCs.

Ok. I will try to come up with a replacement PR next week.

iherman:
Closing this, without merge, in favor of #1326

Issue #1267, opened 2023-09-05
Encourage the use of OHTTP


awoie:
From https://github.com/w3cping/privacy-request/issues/121#issuecomment-1682691192:
> One additional point that was highlighted in https://github.com/w3cping/privacy-request/issues/119 would be to utilize OHTTP more when needing to resolve contexts, status lists, credential schemas, etc from central servers to reduce IP based correlation.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-09-06-vcwg#section3-1) on 2023-09-06

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 3.1. Encourage the use of OHTTP (issue vc-data-model#1267)

_See github issue [vc-data-model#1267](https://github.com/w3c/vc-data-model/issues/1267)._




**Brent Zundel:** oliver, can you walk us through #1267 and why it should be pre- or post-CR.  

**awoie:** This issue resulted from ping group review, and a clarification or addition to security considerations would probably cover it, I believe.  
… I can't think of any normative changes that this would necessitate, and I don't have strong opinion on pre- versus post.  

**Brent Zundel:** if non-normative, post- should be fine.  

**Sebastian Crane:** OHTTP hasn't cleared IETF yet, so even if we wanted to add a normative reference we can't yet.  

> *Kristina Yasuda:* ohttp has not been published by ietf yet!?

> *Manu Sporny:* [https://w3c.github.io/vc-data-integrity/#fingerprinting-network-requests](https://w3c.github.io/vc-data-integrity/#fingerprinting-network-requests).

> *Sebastian Crane:* OHTTP === 'oblivious' HTTP.

> *Oliver Terbu:* +1 manu.

**Manu Sporny:** the precedent is in the link i've just shared.  

> *Dave Longley:* [https://datatracker.ietf.org/doc/draft-ietf-ohai-ohttp/10/](https://datatracker.ietf.org/doc/draft-ietf-ohai-ohttp/10/).

**Sebastian Crane:** I can take this issue and add a note post-CR.  

</details>

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-09-15-vcwg#section3-1) on 2023-09-15

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

[...]

PR #1322, opened 2023-10-21
Add suggestion to use Oblivious HTTP when fetching resources.

msporny:
This PR attempts to address issue #1267 by suggesting that implementers consider Oblivious HTTP when fetching external resources. /cc @kdenhartog

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1322.html" title="Last updated on Oct 29, 2023, 11:35 PM UTC (0234a2b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1322/cc5cd92...0234a2b.html" title="Last updated on Oct 29, 2023, 11:35 PM UTC (0234a2b)">Diff</a>

### DIFF
--- a/common.js
+++ b/common.js
@@ -85,7 +85,7 @@ var vcwg = {
     },
     'OHTTP': {
       title: 'Oblivious HTTP ',
-      href: 'https://ietf-wg-ohai.github.io/oblivious-http/draft-ietf-ohai-ohttp.html',
+      href: 'https://datatracker.ietf.org/doc/html/draft-ietf-ohai-ohttp',
       authors: ['Martin Thomson', 'Christopher A. Wood'],
       status: 'Working Group Draft',
       publisher: 'IETF Oblivious HTTP Application Intermediation'
--- a/common.js
+++ b/common.js
@@ -82,6 +82,13 @@ var vcwg = {
       authors: ['Daniel Buchner', 'Brent Zundel', 'Martin Riedel', 'Kim Hamilton Duffy'],
       status: 'DIF Ratified Specification',
       publisher: 'Decentralized Identity Foundation'
+    },
+    'OHTTP': {
+      title: 'Oblivious HTTP ',
+      href: 'https://ietf-wg-ohai.github.io/oblivious-http/draft-ietf-ohai-ohttp.html',
+      authors: ['Martin Thomson', 'Christopher A. Wood'],
+      status: 'Working Group Draft',
+      publisher: 'IETF Oblivious HTTP Application Intermediation'
     }
   }
 };
--- a/index.html
+++ b/index.html
@@ -4408,13 +4408,13 @@ transmit <a>verifiable credentials</a> on behalf of a <a>holder</a>.
         </p>
 
         <p>
-One mechanism that implementers might consider utilizing when fetching external
-resources that are associated with a <a>verifiable credential</a> or a
-<a>verifiable presentation</a> is the Oblivious HTTP protocol [[?OHTTP]].
+The Oblivious HTTP protocol [[?OHTTP]] is one mechanism that implementers might
+consider using when fetching external resources that are associated with a
+<a>verifiable credential</a> or a <a>verifiable presentation</a>.
 Oblivious HTTP allows a client to make multiple requests to an origin server
-without that server being able to link those requests to the client or to
-identify the requests as having come from the same client, while placing only
-limited trust in the nodes used to forward the messages. Using Oblivious HTTP
+without that server being able to link those requests to that client or even to
+identify those requests as having come from a single client, while placing only
+limited trust in the nodes used to forward the messages. Hence, Oblivious HTTP
 is one privacy-preserving mechanism that can be used to reduce the possibility
 of device tracking and fingerprinting.
         </p>
--- a/index.html
+++ b/index.html
@@ -4385,7 +4385,7 @@ shared.
       </section>
 
       <section class="informative">
-        <h3>Device Fingerprinting</h3>
+        <h3>Device Tracking and Fingerprinting</h3>
 
         <p>
 There are mechanisms external to <a>verifiable credentials</a> that are used to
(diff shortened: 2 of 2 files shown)


iherman (approved):

dlongley (approved):

index.html
```diff
@@ -4406,6 +4406,18 @@ <h3>Device Fingerprinting</h3>
...
+        <p>
+One mechanism that implementers might consider utilizing when fetching external
+resources that are associated with a <a>verifiable credential</a> or a
+<a>verifiable presentation</a> is the Oblivious HTTP protocol [[?OHTTP]].
+Oblivious HTTP allows a client to make multiple requests to an origin server
+without that server being able to link those requests to the client or to
+identify the requests as having come from the same client, while placing only
+limited trust in the nodes used to forward the messages. Using Oblivious HTTP
+is one privacy-preserving mechanism that can be used to reduce the possibility
+of device tracking and fingerprinting.
```
TallTed, line 4419:
```suggestion
Oblivious HTTP allows a client to make multiple requests to an origin server
without that server being able to link those requests to that client or even to
identify those requests as having come from a single client, while placing only
limited trust in the nodes used to forward the messages. Hence, Oblivious HTTP
is one privacy-preserving mechanism that can be used to reduce the possibility
of device tracking and fingerprinting.
```

TallTed (changes requested):

decentralgabe (approved):

OR13 (approved):

martinthomson:
What is your goal here?  Are you seeking to protect the link between network identity and interest in a given identity?  Is that the only factor that might be relevant from a privacy perspective?

common.js
```diff
@@ -82,6 +82,13 @@ var vcwg = {
       authors: ['Daniel Buchner', 'Brent Zundel', 'Martin Riedel', 'Kim Hamilton Duffy'],
       status: 'DIF Ratified Specification',
       publisher: 'Decentralized Identity Foundation'
+    },
+    'OHTTP': {
+      title: 'Oblivious HTTP ',
+      href: 'https://ietf-wg-ohai.github.io/oblivious-http/draft-ietf-ohai-ohttp.html',
```
ShivanKaul, line 88: I'd suggest using the official IETF datatracker link: https://datatracker.ietf.org/doc/html/draft-ietf-ohai-ohttp

msporny, line 88:
```suggestion
      href: 'https://datatracker.ietf.org/doc/html/draft-ietf-ohai-ohttp',
```

msporny:
@martinthomson wrote:
> What is your goal here? Are you seeking to protect the link between network identity and interest in a given identity? Is that the only factor that might be relevant from a privacy perspective?

The goal is to do as much as possible to protect the link between the entity evaluating a Verifiable Credential and any associated party of the Verifiable Credential. For example:

* For Holders using Digital Wallets, we want to reduce the chances that a Holder interacting with a Verifiable Credential in the digital wallet provides tracking information to a 3rd party. For example, a digital wallet fetching and rendering externally linked images, checking the validity of a Verifiable Credential (checking revocation lists stored externally to the credential), or visually displaying a particular aspect of the Verifiable Credential (links to external images, PDF files, or thumbnails).
* For Verifiers, we want to reduce signalling to an Issuer that the Verifier has received the Verifiable Credential -- for example,  while checking the validity of a Verifiable Credential (checking revocation lists stored externally to the credential).

There are other factors that we care about from a privacy perspective (see the Privacy Considerations in the Table of Contents): https://www.w3.org/TR/vc-data-model-2.0/

The use of OHTTP was suggested by PING to be helpful from a "network signal that an attacker might use" standpoint.

common.js
msporny: Done in 0c37885017d1365be7e7a1a3165e29f24c6c30e1 , thank you!

kdenhartog:
Noting this would close https://github.com/w3c/vc-bitstring-status-list/issues/80 as well. It may be useful to point that status list back to here, but not necessary since it should be a given that people looking at those specs will refer back here.

@martinthomson During the original PING review, the main goal I was trying to prevent here was the resources hosted by the issuer (e.g. vc-status-list) can recreate the phone-home problem that this is trying to avoid. @msporny pointed out the main cases for this such as vc-status-list which is often hosted by the issuer such that they can analyze the IP traffic to see who's using the VC. It wouldn't be able to prevent time based correlation of when the verifier is verifying the status of the credential, but it would be able to help a bit was the idea.

kdenhartog:
> We're going to speak to this in the status list specification... I was on the fence about mentioning it here.

I think that makes sense to make that a focus there. Particularly because you still run into a liveness tradeoff if the holder is utilzing a cached status list or are faced with a temporal correlation (e.g. how often it's used) if the holder's UA decides to retrieve it each time before presentation. Given that's very specific to that spec that makes sense to do it separately there and here just more generally speak about the decorrelation of network identities and credential identities.

chris-wood (approved):

TallTed (approved):

msporny:
@martinthomson wrote:
> What is your goal here? Are you seeking to protect the link between network identity and interest in a given identity? Is that the only factor that might be relevant from a privacy perspective?

I added new text in 0234a2b372900dfa1bfbd42001cb69a46edd3f0e to provide concrete examples on the benefits of using OHTTP in certain scenarios.

msporny:

index.html
```diff
@@ -4416,8 +4416,26 @@ <h3>Device Tracking and Fingerprinting</h3>
 identify those requests as having come from a single client, while placing only
 limited trust in the nodes used to forward the messages. Hence, Oblivious HTTP
 is one privacy-preserving mechanism that can be used to reduce the possibility
-of device tracking and fingerprinting.
+of device tracking and fingerprinting. Concrete examples for how Oblivious HTTP
+can benefit ecosystem participants are included below.
```
TallTed, line 4420:
```suggestion
of device tracking and fingerprinting. Concrete examples of ways that Oblivious
HTTP can benefit ecosystem participants are described below.
```

index.html
```diff
@@ -4416,8 +4416,26 @@ <h3>Device Tracking and Fingerprinting</h3>
...
         </p>
+
+        <ul>
+          <li>
+A <a>holder</a> using a digital wallet can reduce the chances that they
+will be tracked by a 3rd party when accessing external links within a
```
TallTed, line 4426:
```suggestion
A <a>holder</a> using a digital wallet can reduce their chances of being
tracked by a third party when accessing external links within a
```

index.html
```diff
@@ -4416,8 +4416,26 @@ <h3>Device Tracking and Fingerprinting</h3>
...
+<a>verifiable credential</a> stored in their digital wallet.
+For example, a digital wallet might fetch and render linked images, or
+check the validity of a <a>verifiable credential</a> by fetching an
+externally linked revocation list.
```
TallTed, line 4430:
```suggestion
For example, a digital wallet might use Oblivious HTTP to fetch linked
images for rendering, or to fetch an externally linked revocation list
for use in checking the validity of a <a>verifiable credential</a>.
```

index.html
```diff
@@ -4416,8 +4416,26 @@ <h3>Device Tracking and Fingerprinting</h3>
...
+          </li>
+          <li>
+A <a>verifier</a> can reduce signalling to an <a>issuer</a> that the
+<a>verifier</a> has received a specific <a>verifiable credential</a>.
+For example, a <a>verifier</a> might fetch an externally linked revocation
+list while performing status checks on a <a>verifiable credential</a>.
```
TallTed, line 4436:
```suggestion
For example, a <a>verifier</a> might use Oblivious HTTP to fetch an
externally linked revocation list while performing status checks on a
<a>verifiable credential</a>.
```

TallTed (commented):
@msporny — Actionable as comments, or do you want a PR for these?

Issue #1318, opened 2023-10-18
Updating figures 6 and 7


iherman:
As discussed in https://github.com/w3c/vc-data-model/issues/1248, both [Figure 6](https://www.w3.org/TR/vc-data-model-2.0/#info-graph-vc) and [Figure 8](https://www.w3.org/TR/vc-data-model-2.0/#info-graph-vc) are incorrect and misleading. I have come up with an alternative for both which is also in line with exact semantics of the vocabulary. I can do a PR at some point, but I thought it is worth discussion it in an issue whether approach works.

Compared to the current diagrams:

- the nodes properly separate literals from objects (ie, entities with their own URLs); it was messy in the original
- the term "graph" is reserved for what is indeed a graph in the RDF sense, and the property arrows reflect these choices. The separate entity (eg, for claim) is a "conceptual" partition of the terms, but not reflected in the properties.
- stylistically, I keep to black-and-white so that these diagrams could work properly with various dark modes; that is to be experimented with a separate PR (if it gets to the point).

Note that the description in the text should also be reviewed as for the usage of "graph". As I said, if we keep to the precise vocabulary meaning, the "claim" is not a separate graph (in the RDF sense, that is), so we should probably change the description.

Here are the proposed new versions:

![Figure 6](https://github.com/w3c/vc-data-model/assets/520723/985aaf44-41d2-4e1a-b0ee-adb9be79f91e)

and

![Figure 8](https://github.com/w3c/vc-data-model/assets/520723/41085318-d462-4ae9-9eba-95a095c83fb2)

dlongley:
I'm concerned about the "Claim" label -- as each of the other statements are also claims. What if the label said "Example Claim" or "One Claim" ... or something? Maybe it's best to just remove that particular circle. It also seems to imply there's a separate container thing there of some *other* type when there isn't.

iherman:
See #1321 for a PR.

Also, due to my combined stupidity and clumsiness, some comments have been misplaced. The real target for 
https://github.com/w3c/vc-data-model/issues/1248#issuecomment-1770490256 and https://github.com/w3c/vc-data-model/issues/1248#issuecomment-1771154697 and, I suppose, https://github.com/w3c/vc-data-model/issues/1248#issuecomment-1771100088 is this current issue...

msporny:
PR #1326 was raised to address this issue, closing.

PR #1326, opened 2023-10-24
Improve the explanation of using graphs in VCDM

iherman:
This PR is a replacement for #1321, following up on https://github.com/w3c/vc-data-model/pull/1321#issuecomment-1772083341. Globally, it takes a more thorough approach and also attempts to provide a more complete solution to #1248 and #1318.

What it does is:

- brings to the spec the RDF terms "named graphs" and "default graph", although in simpler formats
- [Figure 6](https://www.w3.org/TR/vc-data-model-2.0/#info-graph-vc) and [Figure 8](https://www.w3.org/TR/vc-data-model-2.0/#info-graph-vc) both fully exploit these terms, both in the diagram and in the preceding informal description
- the `proof` property has now a proper definition of _what_ is being proven, i.e., which claims; this is done by using the graph terminology.

All this required some more, relatively minor changes on the text for a better coherence. The most visible change is that what used to be [§4.11.1 Verifiable Credential Graphs](https://www.w3.org/TR/vc-data-model-2.0/#verifiable-credential-graphs) has been moved to the top level of the Basic Concepts section, and is now known as [§4.9](https://raw.githack.com/w3c/vc-data-model/improve-graph-usage-explanation/index.html#verifiable-credential-graphs) (with some minor changes).

(As an aside, the PR also puts the sources of the new versions of Figures 6 and 8 to the repository, using the drawio format. Better have all these at the same place.)

For reviewers: the automatic preview below does not work properly because it does not process included files (e.g., the terminology section or the diagrams). You can use these pointers instead:

- [Preview](https://raw.githack.com/w3c/vc-data-model/improve-graph-usage-explanation/index.html)
- [Diff](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FTR%2Fvc-data-model-2.0%2F&doc2=https%3A%2F%2Flabs.w3.org%2Fspec-generator%2F%3Ftype%3Drespec%26url%3Dhttps%3A%2F%2Fraw.githack.com%2Fw3c%2Fvc-data-model%2Fimprove-graph-usage-explanation%2Findex.html%23terminology#terminology) (take a deep breath when using this, and be patient: there is no caching of the processing behind this, i.e., respec generates a final form, which is followed by an HTML diff on two fairly large documents…)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1326.html" title="Last updated on Oct 25, 2023, 9:49 AM UTC (c946571)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1326/cc5cd92...c946571.html" title="Last updated on Oct 25, 2023, 9:49 AM UTC (c946571)">Diff</a>


iherman:
I marked it as post-CR, insofar as the PR does not introduce any normative change, just tightens up the specification (e.g., for the usage of the `proof` property). I was hesitating, though; I think it would be way better to get this done before going to CR.

I leave the decision to the chairs. @brentzundel @Sakurann

OR13 (approved):

index.html
```diff
@@ -1930,6 +1945,48 @@ <h4>Representing Time</h4>
...
+          document contains a single credential with, possibly, associated proofs,
+          this graph is the <a>default graph</a>.
+          For <a>presentations</a>, each value associated with the
+          `verifiableCredential` property of the <a>presentation</a> is a separate <a>named graph</a> of type
+          <dfn class="lint-ignore">VerifiableCredentialGraph</dfn> which contains a single
+          <a>verifiable credential</a>.
+        </p>
+        <p>
+          Using these <a>graphs</a> has a concrete effect on full
+          JSON-LD processing, which properly separates graph node identifiers in one graph
```
dlongley, line 1978: Whether it is "full" or not doesn't change this, and "full" is not a good name. There is static / application-specific processing that only accepts certain shapes and contexts, but that doesn't change any of the facts about the data model and this separation.

```suggestion
          Using these <a>graphs</a> has a concrete effect when performing
          JSON-LD processing, which properly separates graph node identifiers in one graph
```

index.html
```diff
@@ -1930,6 +1945,48 @@ <h4>Representing Time</h4>
...
+          from those in another graph. Implementers that do not fully process JSON-LD will
+          need to keep this in mind if they merge data from one <a>verifiable
```
dlongley, line 1980:
```suggestion
          from those in another graph. Implementers that limit their inputs to application-specific
          JSON-LD documents will also need to keep this in mind if they merge data from one <a>verifiable
```

index.html
```diff
@@ -1969,10 +2026,19 @@ <h3>Securing Verifiable Credentials</h3>
...
+              the authorship of a <a>verifiable credential</a> or a <a>verifiable
+              presentation</a>. Each proof is a separate <a>named graph</a>
+              (referred to as a <dfn>proof graph</dfn>) containing a single proof.
+              The specific method used for an <a>embedded proof</a> MUST be
+              identified using the <code>type</code> <a>property</a>.
+            </p>
+            <p>
+              A proof for a <a>verifiable credential</a> proves the validity of all <a>claims</a>
+              included in the corresponding <a>verifiable credential graph</a>. See <a href="#presentations-0"></a>
+              for the case when the property is used for a <a>verifiable presentation</a>.
```
dlongley, line 2040: We shouldn't use the term "validity" here, it leads to confusion. I suggest:

```suggestion
              A proof for a <a>verifiable credential</a> covers all <a>claims</a>
              included in the corresponding <a>verifiable credential graph</a>. See <a href="#presentations-0"></a>
              for the case when the property is used for a <a>verifiable presentation</a>.
```

index.html
```diff
@@ -2161,10 +2234,20 @@ <h3>Presentations</h3>
...
-Verifiable Credentials Specifications Directory [[?VC-SPECS]].
+            <p>
+                The <a>verifiable presentation</a> MAY include a <code>proof</code>
+                <a>property</a>, which refers to a separate <a>named graph</a> containing a single proof.
+                The specific method used for the MUST be
+                identified using the <code>type</code> <a>property</a>.
+                <!-- If present, the value SHOULD be used to express a securing
+                mechanism, such as those listed in the
+                Verifiable Credentials Specifications Directory [[?VC-SPECS]]. -->
+                The proof proves the validity of all claims in the following <a>graphs</a>:
```
dlongley, line 2245:
```suggestion
                The proof covers of all claims in the following <a>graphs</a>:
```

dlongley (changes requested):
A few suggestions, then I'm supportive. Thanks!

index.html
iherman: I agree (b.t.w., this text was in the spec before, I just copied it :-)

index.html
iherman: Same as above...

dlongley (approved):

TallTed:
> * [@iherman] brings to the spec the RDF terms "named graphs" and "default graph", although in simpler formats

I am concerned that neither of the things now called "the default graph" in the new diagrams actually is "the default graph" in any scenario. If anything, I would expect each of these instead to be *another* "named graph". (In at least one RDF quadstore, the UNION of all the graphs it holds is "the default graph".)

At minimum, I think some discussion of how and why these come to be "the default graph" here is necessary, certainly in the text neighboring, and possibly _within_, each diagram. Possibly the `the default graph` label should just be changed to `a named graph` in parallel to the other named graph(s) in each diagram.

index.html
```diff
@@ -758,25 +758,33 @@ <h3>Credentials</h3>
...
 
         <figure id="info-graph-vc">
           <img style="margin: auto; display: block; width: 100%;"
-               src="diagrams/vc-graph.svg" alt="diagram with a
-               Credential Graph on top connected via a proof to a
-               Proof Graph on the bottom.  The Credential Graph has
-               Credential 123 with 4 properties: 'type' of value
+               src="diagrams/vc-graph.svg" alt="Diagram with a
+               collections of claims for a 'verifiable credential graph'
+               on top connected via a proof to a
```
TallTed, line 772:
```suggestion
               on top connected via a proof property (or predicate) to a
```

index.html
```diff
@@ -758,25 +758,33 @@ <h3>Credentials</h3>
...
+               The claims for a verifiable credential include
+               'Credential 123' as a subject with 4 properties: 'type' of value
                ExampleAlumniCredential, 'issuer' of Example University,
                'validFrom' of 2010-01-01T19:23:24Z, and
-               credentialSubject of Pat, who has an alumniOf property
-               with value of Example University.  The Proof Graph has
-               Signature 456 with 5 properties: 'type' of
+               credentialSubject of Pat, who also has an alumniOf property
+               with value of Example University.  The verifiable credential proof graph 
+               has and object 'Signature 456' subject with 5 properties: 'type' of
```
TallTed, line 780: I know `and` should be `an`, but I'm not sure I understand this sentence, so don't want to try to reword it immediately. I *think* it should say something like, "The verifiable credential proof graph has a `proof` predicate with 'Signature 456' as its  object; the subject 'Signature 456' itself has 5 properties: ..." but I don't think that this text nor the associated diagram is optimal yet.

I think each named graph (including what is currently labeled as "the default graph") should be identified, explicitly, via an example URI or a blank node identifier.
```suggestion
               has an object 'Signature 456' subject with 5 properties: 'type' of
```

TallTed (changes requested):
I'll have more, but the direction I go in from here will depend on the outcome of the comments already made, so I'll wait for the more...

iherman:
Thanks for your comments, @TallTed.

As for:

> 
> I am concerned that neither of the things now called "the default graph" in the new diagrams actually is "the default graph" in any scenario. If anything, I would expect each of these instead to be _another_ "named graph". (In at least one RDF quadstore, the UNION of all the graphs it holds is "the default graph".)
> 

I use the term "default graph" in the way it is defined in the [RDF 1.1 concepts spec](https://www.w3.org/TR/rdf11-concepts/#dfn-default-graph). This is in line with the way the term is used in [JSON-LD 1.1](https://www.w3.org/TR/json-ld11/#terms-imported-from-other-specifications) or [N-Quads](https://www.w3.org/TR/n-quads/#simple-triples) (both are relevant to us). The succinct way N-Quads describes the quads in a default graph is (when defining a simple quad term):   
"The graph label IRI can be omitted, in which case the triples are considered part of the default graph of the RDF dataset."

I'm afraid the usage of the term "default graph" in that quadstore may not be aligned with the RDF specs, but I cannot judge it.

> At minimum, I think some discussion of how and why these come to be "the default graph" here is necessary, certainly in the text neighboring, and possibly _within_, each diagram. 

I've not brought in anything new. I just describe, in more precise terms, what is already implicitly defined by the spec. E.g., let us look at [Example 5](https://www.w3.org/TR/vc-data-model-2.0/#example-usage-of-the-proof-property-on-a-verifiable-credential) in the spec, which is the case of a "simple" credential (i.e., no presentation). The same dataset in N-Quads can be seen on the [JSON-LD playground example](https://tinyurl.com/ntz5k7jj). The terms that are part of the "core" of the credentials are all quads without a graph label, i.e., part of the default graph of the dataset per the N-Quads specification.

> Possibly the `the default graph` label should just be changed to `a named graph` in parallel to the other named graph(s) in each diagram.

I don't think so. As the aforementioned playground example shows, those quads are different; there is also a real named graph in that example (the proof graph), and the quads all have an explicit graph label `_:b0`. If the goal was to turn what is now the default graph into a genuine named graph, we would have to have a different (and more complicated) JSON-LD `@context`, if not a different JSON-LD structure altogether. It would not reflect today's specification.

B.t.w., to make things complete and coherent I did make some changes in other places of the spec making use of the default graph/named graph terminology. See, for example, the [section on verifiable credential graphs](https://raw.githack.com/w3c/vc-data-model/improve-graph-usage-explanation/index.html#verifiable-credential-graphs) or, primarily, the definition of the `proof` property in [§4.10](https://raw.githack.com/w3c/vc-data-model/improve-graph-usage-explanation/index.html#securing-verifiable-credentials) and [§4.12](https://raw.githack.com/w3c/vc-data-model/improve-graph-usage-explanation/index.html#presentations-0). I believe the specification was incomplete with regard to the `proof` property, and that addition was necessary (this is, in my view, the core of the discussion in #1248).

I believe the real goal of all this PR is the issue around the `proof` property's definition (i.e., the incompleteness thereof). I think introducing the default graph/named graph terminology into the spec is the simplest way of doing that.

decentralgabe (approved):

TallTed:
Perhaps there is a way to qualify "the default graph" so it doesn't appear to be "*the only* default graph", but rather is "the default graph *of this dataset*"? I will think further on this. I raise the question here so others who might have good ideas might voice them.

_I realized that I mis-typed above, where I said '(In at least one RDF quadstore, the `UNION` of all the graphs it holds is "the default graph".)' It's not the `UNION`, it's the `MERGE`._

iherman:
> Perhaps there is a way to qualify "the default graph" so it doesn't appear to be "_the only_ default graph", but rather is "the default graph _of this dataset_"? I will think further on this. I raise the question here so others who might have good ideas might voice them.

I take your point, and I am happy to change this. The problem is that "the default graph of this dataset" is a bit of a mouthful.  "Dataset default graph"? "Local default graph"?

index.html
```diff
@@ -2161,10 +2234,20 @@ <h3>Presentations</h3>
           </dd>
           <dt><var>proof</var></dt>
           <dd>
-The <a>verifiable presentation</a> MAY include a <code>proof</code>
-<a>property</a>. If present, the value SHOULD be used to express a securing
-mechanism, such as those listed in the
-Verifiable Credentials Specifications Directory [[?VC-SPECS]].
+            <p>
+                The <a>verifiable presentation</a> MAY include a <code>proof</code>
+                <a>property</a>, which refers to a separate <a>named graph</a> containing a single proof.
```
OR13, line 2239: my understanding is that multiple proofs are allowed here, and that in the case of BBS, the secure these information graphs in very different ways.

decentralgabe (commented):
Just a comment - not blocking - it would be great to have a similar set of graphics/text that highlight a graph VC with a vc-jose-cose based proof

index.html
```diff
@@ -2161,10 +2234,20 @@ <h3>Presentations</h3>
           </dd>
           <dt><var>proof</var></dt>
           <dd>
-The <a>verifiable presentation</a> MAY include a <code>proof</code>
-<a>property</a>. If present, the value SHOULD be used to express a securing
-mechanism, such as those listed in the
-Verifiable Credentials Specifications Directory [[?VC-SPECS]].
+            <p>
+                The <a>verifiable presentation</a> MAY include a <code>proof</code>
```
OR13, line 2238: I feel that proof remains a property of data integrity, and would prefer to see normative guidance related to data integrity done in the data integrity specification.

index.html
iherman: @OR13 I have no opinion on this, and I would prefer if this was raised and discussed as a separate issue. The text in this version simply reused what is currently in the official draft.

Could you raise a separate issue and mark this conversation as resolved?

index.html
iherman: @OR13 similar comment as above: I've just reused the current draft, and I do not have a strong opinion either way. Would it be possible to move this into a separate issue and not hold up this one on this?

iherman:
> Just a comment - not blocking - it would be great to have a similar set of graphics/text that highlight a graph VC with a vc-jose-cose based proof

@decentralgabe I agree with that, and I am also happy looking into this, but I would prefer to do that once this PR is merged. Truth must be said, I am not really deeply familiar with the JWT world, so I do not know whether there is an accepted "style" of diagrams. Do you know?

decentralgabe:
@iherman no problem if you want to take a shot at it after this PR is merged that would be greatly appreciated! Perhaps @selfissued knows if there's prior art on such diagrams, if not, I think a similar form to what you've already done would work nicely.

msporny (approved):

msporny:

Issue #1047, opened 2023-02-16
NIST defines "credential" differently


OR13:
https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-63-4.ipd.pdf

<img width="661" alt="Screen Shot 2023-02-16 at 2 48 09 PM" src="https://user-images.githubusercontent.com/8295856/219472355-af1f40d5-d906-496e-b303-00019873ae17.png">

EDITED to clarify the ask:

The working group should provide citations for other vocabularies, and explain why their definitions are useable or not.

We should also refer to https://www.ietf.org/rfc/rfc4949.txt

mwherman2000:
@or13 It's the old, legacy, person-centric view of identifiers and identities ...like the early, early versions of the DID spec.

PatStLouis:
Defines differently than what definition?

Furthermore, this definition of credential relates exclusively to the realm of authentication within the scope of this document, as stated in the Appendix A.1 section:

> "A wide variety of terms is used in the realm of authentication. [...] Many of these terms lack a single, consistent definition, warranting careful attention to how the terms are defined here."

The dictionary defines credentials as: _That which entitles one to confidence, credit, or authority._

In the digital realm this could be further interpreted as a mean to authenticate an entity or to prove claims about an entity.

Can you clarify how you feel this poses an issue and to what degree?

RieksJ:
What's the problem? It is widely known and accepted that the meaning of a term is scoped, i.e. restricted to a particular scope/context. 

Lawyers understand this. In legal texts (I've seen Dutch and EU such texts), it is common to have a terminology section that defines terms that are used within the scope of the particular text. While typically efforts are made not to be too far off with terminology in related scopes (other legal texts), the whole idea is that people (including lawyers, judges, etc.) have the same understanding of what is (not) an instance of a term.

Software engineers know this. They have namespaces, where the name in a namespace refers e.g., to a specific variable or function, that have well-defined meanings (or else the software doesn't work). Namespaces (the equivalent of a Terminology) can be inherited, and their meanings renamed, or terms redefined. 

The VCDM lives in the specific context in which we want to talk, e.g., about sets of digital representations of of statements that a single party makes about a variety of entities (for which the VCDM uses the term 'credential'), and such things that also have the property of being tamper-evident and a proof of who this single party is, where the proof can be cryptographically verified (which in the VCDM terminology is a 'verifiable credential'). 

[...]

PR #1332, opened 2023-10-31
Add comparison to NIST definition

OR13:
Addresses: https://github.com/w3c/vc-data-model/issues/1047

### DIFF
--- a/terms.html
+++ b/terms.html
@@ -11,6 +11,11 @@ An assertion made about a <a>subject</a>.
   <dd>
 A set of one or more <a>claims</a> made by an <a>issuer</a>.
 The <a>claims</a> in a credential can be about different <a>subjects</a>.
+<p>
+  Our definition of credential differs from, 
+  <a href="https://csrc.nist.gov/glossary/term/credential">NIST's definitions of credential</a>, 
+  in particular our definition does not require any binding to identity.
+</p>
   </dd>
   <dt><dfn>data minimization</dfn></dt>
   <dd>
--- a/terms.html
+++ b/terms.html
@@ -13,8 +13,7 @@ A set of one or more <a>claims</a> made by an <a>issuer</a>.
 The <a>claims</a> in a credential can be about different <a>subjects</a>.
 <p>
   Our definition of credential differs from, 
-  <a href="https://csrc.nist.gov/glossary/term/credential">NIST's definitions of credential</a>, 
-  in particular our definition does not require any binding to identity.
+  <a href="https://csrc.nist.gov/glossary/term/credential">NIST's definitions of credential</a>.
 </p>
   </dd>
   <dt><dfn>data minimization</dfn></dt>


terms.html
```diff
@@ -11,6 +11,11 @@
   <dd>
 A set of one or more <a>claims</a> made by an <a>issuer</a>.
 The <a>claims</a> in a credential can be about different <a>subjects</a>.
+<p>
+  Our definition of credential differs from, 
+  <a href="https://csrc.nist.gov/glossary/term/credential">NIST's definitions of credential</a>, 
+  in particular our definition does not require any binding to identity.
```
andresuribe87, line 17:
```suggestion
  in particular, our definition does not require any binding to identity.
```

andresuribe87 (approved):

brentzundel (approved):

terms.html
jandrieu: What does "identity" mean?

dlongley: I agree with Joe -- we should just strike the "in particular" part, IMO.

terms.html
```diff
@@ -11,6 +11,11 @@
   <dd>
 A set of one or more <a>claims</a> made by an <a>issuer</a>.
 The <a>claims</a> in a credential can be about different <a>subjects</a>.
+<p>
+  Our definition of credential differs from, 
+  <a href="https://csrc.nist.gov/glossary/term/credential">NIST's definitions of credential</a>,
```
TallTed, line 16:
```suggestion
  Our definition of credential differs from
  <a href="https://csrc.nist.gov/glossary/term/credential">NIST's definitions of credential</a>;
```

terms.html
TallTed: I think it's not our "identity", here; it's NIST's ["identity"](https://csrc.nist.gov/glossary/term/identity) to which NIST is requiring [their credentials](https://csrc.nist.gov/glossary/term/credential) be bound.

Perhaps ["identity"](https://csrc.nist.gov/glossary/term/identity) should be a data cite in our doc? Or just a link out?

On the other hand, I'm not sure that "our definition does not require any binding to identity" is quite accurate. We require that a credential have a subject, whether URI (explicit) or blank node (implicit), which is, in a sense, a binding to that identifier (in our terminology) which it appears NIST might consider an identity, making this not a difference between their credentials and ours....

jandrieu: > On the other hand, I'm not sure that "our definition does not require any binding to identity" is quite accurate.

I'm not sure if that's meant as an interpretation of my comment. If so, that's not quite what I meant, but I do agree that quote would not be accurate, if only because we don't have a formal definition. 

*My* definition of identity would lead to the position that VCs *do* bind to identity because identity is how we recognize, remember, and respond to specific people and things. VCs do that with identifiers and proofs and confidenceMethod.  But we don't require that VCs are bound to "physical and behavioral characteristics" of individuals which is the heart of that NIST definition of Identity.

So, I would argue our definition of identity is what is out of alignment. Hence, my question.

That said, I don't expect that the WG would consider my definition as canon. We certainly haven't defined it in spec text.

I like @dlongley's proposal just to remove the clause "in particular our definition does not require any binding to identity."

longpd: > What does "identity" mean?
> 
> I don't mean to be obtuse here, but I believe we do bind to identity in that we use identifiers as a means to bind a credential so that we can recognize, remember, and respond to the subject of the VC.
> 
> In contrast, I expect what NIST is doing is a technical "binding" that they believe binds the VC to a particular individual. Which is also a form of identity--and not something VCs require. But to say VCs don't bind identity is incorrect without a specific definition of identity.

I agree with Joe, Dave, and Ted that there is a binding of an identifier with VCs. We are not binding to an individual person, but the unique URL, or a DID, with  a proof or other confidence method. 

I would strike the entire sentence, 

> "in particular, our definition does not require any binding to identity."

TallTed (approved):
*IFF* @dlongley's suggested change is applied.

decentralgabe (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-11-14-vcwg-special#section1-5) on 2023-11-14

- no resolutions were taken

#### 1.5. Add comparison to NIST definition (pr vc-data-model#1332)

_See github pull request [vc-data-model#1332](https://github.com/w3c/vc-data-model/pull/1332)._




**Brent Zundel:** there was one change request on this one, which folks seemed fine with. I think we are waiting on Orie to respond.  
… I will ping Orie directly to ask him to respond (he is not available for these Tuesday calls).  

**Joe Andrieu:** seems like there was reasonable consensus to remove that last line, but I do think it is appropriate to bring in Orie on this one.  

> *Sebastian Crane:* +1 to JoeAndrieu (IRC) - I think we have consensus. Can we note this for Orie's benefit?

**Brent Zundel:** I think there is broad agreement on this one if the recommended change is made.  

</details>

terms.html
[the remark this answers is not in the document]
OR13: I think the lack of identity binding is important security information. Our data model does not enforce security properties, we rely on the presence of optional embedded or external proofs.

That being, said I am applying this suggestion.

OR13: its gone, I'm resolving this thread.

OR13:
I believe all suggestions have been applied.

brentzundel:
non-normative changes, open for review for several weeks, requested changes applied, merging

terms.html
```diff
@@ -11,6 +11,10 @@
   <dd>
 A set of one or more <a>claims</a> made by an <a>issuer</a>.
 The <a>claims</a> in a credential can be about different <a>subjects</a>.
+<p>
+  Our definition of credential differs from,
```
TallTed, line 15: strange place for a comma...
```suggestion
  Our definition of credential differs from
```

Issue #1289, opened 2023-09-20
Update i18n example


OR13:
https://w3c.github.io/vc-data-model/#example-usage-of-the-name-and-description-property-0

```json
...
  "description": [{
    "value": "2015 Bachelor of Science and Arts Degree",
    "lang": "en"
  }, {
    "value": "2015 Baccalauréat Scientifique et Arts",
    "lang": "fr"
  }, {
    "value": "2015 بكالوريوس العلوم والآداب",
    "lang": "ar",
    "dir": "rtl"
  }],
  "credentialSubject": {
    "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
    "degree": {
      "type": "ExampleBachelorDegree",
      "subtype": "Bachelor of Science and Arts" // this should be updated
    }
  }
...
```

Also just a comment that `type` in the example above is actually `@type`, but `subtype` is not an RDF type expression, it is similar to https://schema.org/additionalType

When localizing JSON-LD, you can't localize `@type` without breaking the graph... but you can localize `additionalType`.

msporny:
PR #1340 was raised to address this issue and changes `subtype` back to `name`, which was the original intention of the field, which was to express the human-readable name of the degree (not a `subtype` of the degree). That PR also updates the `@language` values appropriately.

Marking this as pending close.

PR #1333, opened 2023-10-31
Add comment on localizing vocabulary terms.

OR13:
Addresses: https://github.com/w3c/vc-data-model/issues/1289

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1333.html" title="Last updated on Oct 31, 2023, 9:35 PM UTC (4d06377)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1333/19fb1e3...4d06377.html" title="Last updated on Oct 31, 2023, 9:35 PM UTC (4d06377)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1532,6 +1532,10 @@ an object that contains the `value`, `lang`, and `dir` properties to express
 the text value, language tag, and text direction, respectively.
 See <a href="#language-and-base-direction"></a> for further information.
         </p>
+        <p>
+The vocabulary required by this specification relies on RDF classes and predicates
+that cannot be localized, in order to ensure consistent semantic processing.
+        </p>
 
         <pre class="example nohighlight"
           title="Usage of the name and description property">
@@ -1594,7 +1598,17 @@ See <a href="#language-and-base-direction"></a> for further information.
     "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
     "degree": {
       "type": "ExampleBachelorDegree",
-      "subtype": "Bachelor of Science and Arts"
+      "subtype": [{
+        "value": "Bachelor of Science and Arts Degree",
+        "lang": "en"
+      }, {
+        "value": "Baccalauréat Scientifique et Arts",
+        "lang": "fr"
+      }, {
+        "value": "بكالوريوس العلوم والآداب",
+        "lang": "ar",
+        "dir": "rtl"
+      }],
     }
   }
 }


iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-11-01-vcwg#section1-3) on 2023-11-01

- no resolutions were taken

#### 1.3. Add comment on localizing vocabulary terms. (pr vc-data-model#1333)

_See github pull request [vc-data-model#1333](https://github.com/w3c/vc-data-model/pull/1333)._




**Brent Zundel:** Add comment on localizing vocabulary terms. Raised in response to issue 1289. This is a before CR PR. Its a relatively minor change. Comments?  

> *Dave Longley:* +1 not to merge because the example won't be what we do.

**Manu Sporny:** The second part of the change is fine except that it is not what we will end up doing. We can't alias value because schema.org uses it. Queued for a future call. The commentary around the vocab is weird. Not sure why we are saying that. Suggest that we don't put that language in there. Its a low level detail we are raising.  

</details>

decentralgabe (approved):

index.html
```diff
@@ -1532,6 +1532,10 @@ <h3>Names and Descriptions</h3>
 the text value, language tag, and text direction, respectively.
 See <a href="#language-and-base-direction"></a> for further information.
         </p>
+        <p>
+The vocabulary required by this specification relies on RDF classes and predicates
+that cannot be localized, in order to ensure consistent semantic processing.
+        </p>
```
msporny, line 1538: It's not clear what this means. What does "cannot be localized" mean? Do you mean:

> The vocabulary used by this specification defines terms in a way that is globally unambiguous and in a way that ensures consistent semantic processing.

The only thing that throws a wrench into those words is the whole "issuer-defined" vocabulary stuff, which results in ambiguity (but, at least the ambiguity is constrained to people that don't want to define semantics for the terms they use via an inline mechanism).

msporny (changes requested):
It's not clear what the paragraph in this PR is attempting to convey, I'd suggest a re-write, but I can't be sure that's what is meant by this PR.

index.html
TallTed: This paragraph does not make sense to me. The lexical values of RDF class and predicate URIs should be treated as opaque, i.e., without literal value in and of themselves.

Further, each RDF class and predicate URI may be associated with multiple [`schema:name`](https://schema.org/name), [`rdfs:label`](http://www.w3.org/2000/01/rdf-schema#label), [`dcterms:description`](http://purl.org/dc/terms/description), [`rdfs:comment`](http://www.w3.org/2000/01/rdf-schema#comment), [`dcterms:title`](http://purl.org/dc/terms/title), etc., values, whereby there may be a set of these covering each language of concern.

Yes, it may be challenging to get such multi-lingual properties into some externally managed vocabularies/ontologies, so this handling might be reserved for vocabularies managed by the immediate user, and they might want to set up a bunch of `{ localvocab:comment` [`skos:broader`](http://www.w3.org/2004/02/skos/core#broader) `rdfs:comment }` or `{ localvocab:comment` [`owl:equivalentProperty`](http://www.w3.org/2002/07/owl#equivalentProperty) `rdfs:comment }` or even `{ localvocab:comment` [`owl:sameAs`](http://www.w3.org/2002/07/owl#sameAs) `rdfs:comment }`, to really put their reasoner to the test.

But I do not think that "RDF classes and predicates that cannot be localized" is an accurate way of describing this (or any) vocabulary.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-11-14-vcwg-special#section1-6) on 2023-11-14

- no resolutions were taken

#### 1.6. Add comment on localizing vocabulary terms. (pr vc-data-model#1333)

_See github pull request [vc-data-model#1333](https://github.com/w3c/vc-data-model/pull/1333)._




**Brent Zundel:** no approvals, and one change request. We still have not had any response from Orie based on the change request.  

**Manu Sporny:** I don't understand the language in the PR.  
… I don't know what Orie means by "localized", maybe I18N?  

> *Ivan Herman:* I do not understand that comment either.

**Brent Zundel:** I will add this to the things that I ask Orie about.  

**Sebastian Crane:** I can't speak for Orie, but I think what he means is that the actual names of the RDF properties can't be translated into, e.g., French, without overhead. Similarly to how you have to use the english terms.  
… "for" and "while" for loops in C.  

**Ivan Herman:** that is indeed the intention, this is not RDF specific, this is a general feature for all the apis used on the web.  

**Ted Thibodeau Jr.:** as commented in the thread, this appears to be about the language used for URI construction, which should be considered opaque anyway. There are ways to "localize" via multiple schema:name, rdfs:label, dcterms:description, rdfs:comment, dcterms:title, etc.  

> *Dave Longley:* +1 to ivan, this is true for everything, this isn't specific to anything in ... computer science.

</details>

index.html
OR13: What I mean is that we don't expect all terms to be localized.

For example, we don't expect `VerifiableCredential` or `VerifiablePresentation` to be localized, because doing so would break the information model.

Please feel free to suggest alternative text.

TallTed: Neither `VerifiableCredential` nor `VerifiablePresentation` is either an RDF class or an RDF predicate.

If you meant [`https://www.w3.org/2018/credentials#VerifiableCredential`](https://www.w3.org/2018/credentials#VerifiableCredential) or [`https://www.w3.org/2018/credentials#VerifiablePresentation`](https://www.w3.org/2018/credentials#VerifiablePresentation), then I say again -- these are URIs, which means that their lexical values *SHOULD* be treated as opaque.

At present, each of these is described with a single [`rdfs:label`](http://www.w3.org/2000/01/rdf-schema#label) — `"Verifiable credential"@en` and `"Verifiable presentation"@en`, respectively. These can be localized by the addition of more language-tagged literal values of that property, without breaking the information model. Optimally, such additional literal values would be added to the authoritative ontology, but they could be deployment-local additions.

What am I missing?

OR13: @TallTed thats exactly my point... many people don't understand RDF / JSON-LD... and they might start localizing all strings... if they localize strings that become URIs that are supposed to be opaque, they will break things.

OR13:
Closing as conflicts exist, and I have lost the energy to work on this part of the spec.

Issue #1300, opened 2023-10-03
Clarify what conformance means to the issuers/verifiers


Sakurann:
from Issue #1285

> 1.4 Conformance
>
> I think you have 2 or 3 conformance classes that aren't mentioned here. Since the spec includes statements like "An issuer MUST" and "A verifier SHOULD", software can conform to this spec by being a conforming issuer or a conforming verifier. And you should probably have more statements that constrain verifiers, for example to enforce the properties desired in 7.10 Validity Checks. I suspect that holders (or credential repositories?) don't need to be a conformance class, but you might want to turn statements like "Holders should be warned by their software" into active requirements on the repositories, which would make them a conformance class.

consider new sections that clarify what conformance to the spec requires from verifiers and/or issuers

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-11-01-vcwg#section2-4) on 2023-11-01

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 2.4. Clarify what conformance means to the issuers/verifiers (issue vc-data-model#1300)

_See github issue [vc-data-model#1300](https://github.com/w3c/vc-data-model/issues/1300)._




**Brent Zundel:** clarify what conformance means to issuers and verifiers. Assigned to Manu and Joe. What is the status?  

**Manu Sporny:** Still needs a PR written. I intend to do this, but cannot give a timeline. We should try to address this. Maybe someone else can get to it before I do.  

**Brent Zundel:** thank you. Agree we should not close and it should have a PR.  

</details>

msporny:
PR #1336 has been raised to address this issue. This issue will be closed once PR #1336 has been merged.

msporny:
PR #1336 has been merged, closing.

PR #1336, opened 2023-11-07
Add conformance classes for issuers and verifiers.

msporny:
This PR attempts to address issue #1300 by creating new conformance classes for "conforming issuer processors" and "conforming verifier processors". I expect we'll need to fine tune the language before this PR can be merged.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1336.html" title="Last updated on Nov 15, 2023, 8:55 PM UTC (7a68810)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1336/998a3dc...7a68810.html" title="Last updated on Nov 15, 2023, 8:55 PM UTC (7a68810)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -569,41 +569,34 @@ serialization format.
         </p>
 
         <p>
-A <dfn>conforming processor</dfn> is any algorithm realized as software and/or
-hardware that generates and/or consumes a <a>conforming document</a>. <a>Conforming
-processors</a> MUST produce errors when non-conforming documents are
-encountered.
+A <dfn class="lint-ignore">conforming issuer implementation</dfn> produces
+<a>conforming documents</a>, MUST include all required properties in the
+<a>conforming documents</a> that it produces, and MUST secure the <a>conforming
+documents</a> it produces using a securing mechanism as described in Section <a
+href="#securing-verifiable-credentials"></a>.
         </p>
 
         <p>
-A <dfn class="lint-ignore">conforming issuer processor</dfn> is a <a>conforming
-processor</a> that produces <a>conforming documents</a>. A <a>conforming
-issuer processor</a> MUST include all required fields in the <a>conforming
-documents</a> that it produces. A <a>conforming issuer processor</a> MUST
-secure the <a>conforming documents</a> it produces using a securing mechanism
-as described in Section <a href="#securing-verifiable-credentials"></a>.
+A <dfn class="lint-ignore">conforming verifier implementation</dfn>
+consumes <a>conforming documents</a>, MUST perform <a>verification</a> on a
+<a>conforming document</a> as described in Section
+<a href="#securing-verifiable-credentials"></a>, and MUST check that each
+required property satisfies the normative requirements for that property.
         </p>
 
         <p>
-A <dfn class="lint-ignore">conforming verifier processor</dfn> is a
-<a>conforming processor</a> that consumes <a>conforming documents</a>. A
-<a>conforming verifier processor</a> SHOULD check that each required field
-satisfies the normative requirements for that field. A <a>conforming verifier
-processor</a> MUST perform <a>verification</a> on a <a>conforming document</a>.
+This specification includes both required and optional properties. Optional
+properties MAY be ignored by <a>conforming issuer implementations</a> and/or
+<a>conforming verifier implementations</a>.
         </p>
 
         <p>
-This specification includes both required and optional fields. Optional fields
-MAY be ignored by <a>conforming issuer processors</a> and/or
-<a>conforming verifier processors</a>.
-        </p>
-
-        <p>
-<a>Verifiable credential</a> and <a>verifiable presentation</a> MUST be
-protected using a digital proof mechanism such as a digital signature. Having
-and verifying proofs, which may be dependent on the syntax of the proof (for
-example, using the JSON Web Signature of a JSON Web Token for proofing a key
-holder), are an essential part of processing <a>verifiable credentials</a> and
+<a>Verifiable credentials</a> and <a>verifiable presentations</a> MUST be
+protected using a securing mechanism as described in Section
+<a href="#securing-verifiable-credentials"></a>. Having and verifying proofs,
+which may be dependent on the syntax of the proof (for example, using the JSON
+Web Signature of a JSON Web Token for proofing a key holder), are an essential
+part of processing <a>verifiable credentials</a> and
 <a>verifiable presentations</a>. At the time of publication, Working Group
 members had implemented such protection using at least three proof mechanisms:
         </p>
--- a/index.html
+++ b/index.html
@@ -570,8 +570,9 @@ serialization format.
 
         <p>
 A <dfn>conforming processor</dfn> is any algorithm realized as software and/or
-hardware that generates or consumes a <a>conforming document</a>. <a>Conforming
-processors</a> MUST produce errors when non-conforming documents are consumed.
+hardware that generates and/or consumes a <a>conforming document</a>. <a>Conforming
+processors</a> MUST produce errors when non-conforming documents are
+encountered.
         </p>
 
         <p>
@@ -593,7 +594,7 @@ processor</a> MUST perform <a>verification</a> on a <a>conforming document</a>.
 
         <p>
 This specification includes both required and optional fields. Optional fields
-MAY be ignored by <a>conforming issuer processors</a> and
+MAY be ignored by <a>conforming issuer processors</a> and/or
 <a>conforming verifier processors</a>.
         </p>
 
--- a/index.html
+++ b/index.html
@@ -592,7 +592,7 @@ processor</a> MUST perform <a>verification</a> on a <a>conforming document</a>.
         </p>
 
         <p>
-This specification includes both required and optional fields.  Optional fields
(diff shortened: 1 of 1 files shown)


decentralgabe (approved):

index.html
```diff
@@ -569,18 +569,32 @@ <h3>Use Cases and Requirements</h3>
...
+        <p>
+A <dfn class="lint-ignore">conforming verifier processor</dfn> is a
+<a>conforming processor</a> that consumes <a>conforming documents</a>. A
+<a>conforming verifier processor</a> SHOULD check that each required field
+satisfies the normative requirements for that field. A <a>conforming verifier
+processor</a> MUST perform <a>verification</a> on a <a>conforming document</a>.
+        </p>
+
+        <p>
+This specification includes both required and optional fields.  Optional fields
```
decentralgabe, line 595:
```suggestion
This specification includes both required and optional fields. Optional fields
```

iherman (approved):

brentzundel (approved):

TallTed (changes requested):
Tweaks for clarity...

TallTed (approved):

index.html
```diff
@@ -569,18 +569,33 @@ <h3>Use Cases and Requirements</h3>
...
+documents</a> that it produces. A <a>conforming issuer processor</a> MUST
+secure the <a>conforming documents</a> it produces using a securing mechanism
+as described in Section <a href="#securing-verifiable-credentials"></a>.
+        </p>
+
+        <p>
+A <dfn class="lint-ignore">conforming verifier processor</dfn> is a
+<a>conforming processor</a> that consumes <a>conforming documents</a>. A
+<a>conforming verifier processor</a> SHOULD check that each required field
+satisfies the normative requirements for that field. A <a>conforming verifier
```
andresuribe87, line 591: Above, we state that `<a>Conforming
processors</a> MUST produce errors when non-conforming documents are
encountered`

If you MUST produce errors, then the SHOULD in this line doesn't have any effect. It appears to me that the two lines are at odds. Am I misunderstanding this?

seabass-labrax (approved):

index.html
David-Chadwick:
```suggestion
<a>conforming verifier processor</a> MUST check that each required field
satisfies the normative requirements for that field. A <a>conforming verifier
```

David-Chadwick (changes requested):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-11-14-vcwg-special#section1-7) on 2023-11-14

- no resolutions were taken

#### 1.7. Add conformance classes for issuers and verifiers. (pr vc-data-model#1336)

_See github pull request [vc-data-model#1336](https://github.com/w3c/vc-data-model/pull/1336)._




**Brent Zundel:** this PR has nothing but approvals, and has been open for more than a week.  

**Manu Sporny:** was going to mention the comment from andres.  
… I took a shot at naming the conformance class. Right now we call it ConformingIssuerProcessor, but could make an editorial change in the future.  
… JoeAndrieu would like your thoughts on changing to ConformingIssuer, but expect pushback.  

**Joe Andrieu:** I think we need a good set of vocab to refer to these aside from roles.  

**Michael Jones:** suggest we call it a ConformingIssuerImplementation.  

> *Dave Longley:* +1 to "software".

> *Dave Longley:* is fine with "implementation".

> *Joe Andrieu:* +1 to software or implementation.

> *Sebastian Crane:* +1 to Mike from me.

> *Dave Longley:* +1 to "implementation" generally over "processor".

**Brent Zundel:** would anyone be opposed to ConformingIssuerImplementation?  

> *Brent Zundel:* Seems to be support for that, which Mike suggested.

**Andres Uribe:** my interpretation is that the "should" becomes moot.  

> *Sebastian Crane:* +1 to andres from me also. I wouldn't mind making that a MUST.

**Manu Sporny:** you are correct, that needs to be struck.  

> *Dave Longley:* +1 to remove the top-level one.

> *Andres Uribe:* +1 to that as well.

> *Brent Zundel:* +1 to remove top-level one.

**Manu Sporny:** probably what we need to do here... I'm wondering if we should just remove the classes and just have ConformingIssuer and ConformingVerifier.  

> *Dave Longley:* and then leave the others alone.

**Manu Sporny:** e.g., remove the top-level ones and keep the more specific ones.  
… the reason the statement was "SHOULD" - some implementations may want to continue processing something with an error in the vc.  
… "date" is a great example because we have new language in the spec saying you must produce dates with utc timestamp.  
… if the date is wrong in a way that the verifier feels is recoverable, we should allow them to recover. That is why it is a "SHOULD" instead of a "MUST".  

> *Manu Sporny:* "conforming issuer implementation" <-- we'll use that.

**Andres Uribe:** wondering what the value is of introducing normative statements for those classes.  
… seems to me that creating a label for conformance class is simply a grouping of which normative statements are correctly implemented by a specific implementation.  
… just wondering what the value is vs. just saying "implementation X does actually pass all normative statements that are testable".  

**Sebastian Crane:** responding to manu from earlier regarding conformance requirement where MUST is appropriate.  

**Manu Sporny:** yes, specifically I'm saying the time value is corrupt, not wrong.  
… for whatever reason the verifier may know they can recover.  

**Sebastian Crane:** thanks, real world example: vc from phone when accessing turnstile.  
… for some reason the qr is corrupted. by saying MUST in the specification, we are saying that somewhere in the system some alarm must be raised.  
… what is not being said is that the turnstyle cannot open.  
… therefore I don't think it unreasonable to say the implementation should not produce errors (via MUST).  

**Joe Andrieu:** responding to manu - we are assuming processing stops on error. I don't think that's what we are talking about.  
… only that software has to report the error.  
… if the verifier needs to make a business decision, that's validation. If it is something always checked, that is verification and should be a MUST.  

**Brent Zundel:** practically, because we had a Google rep review and say we really needed normative conformance classes. This is in response to that direct feedback.  

**Manu Sporny:** in addition, having conformance classes makes it clear to vendors what kind of software they are producing.  
… sometimes a customer will ask if you are conformant to a standard, and you want to say with confidence that we produce conforming issuer implementations.  
… often w3c specs may talk about different sets of software and requirements.  

> *Ivan Herman:* +1 to Manu.

**Manu Sporny:** e.g., conforming issuer implementations, status list implementations, etc.  
… I am hearing we should just make this a MUST, and we may decrease interop if these corrupt examples are being accepted.  

> *Phillip Long:* +1 to Joe's comment that this should be made "must" if it is a verification.

**Manu Sporny:** agreed we are saying must produce errors, not halt processing.  

> *Sebastian Crane:* +1 to manu and to use of 'MUST' here.

**Brent Zundel:** path forward seems to be keep MUST, get rid of first tier conformance classes, and just keep IssuerImplementation and VerifierImplementation.  

**David Chadwick:** I agree with the MUST, but seabass' use case was slightly wrong - conflating validation and verification.  

> *Joe Andrieu:* +1 to validation can ignore the errors.

**Brent Zundel:** this should be merged by end of week at the latest.  

</details>

msporny:
I have applied the following changes based on the call yesterday:

* Applied the "implementation" language as suggested by @selfissued.
* Addressed @andresuribe87 comment by removing the conflicting normative MUST/SHOULD.
* Applied @David-Chadwick's requested changes in 857f422125a07216ec4d3b2cd5e14b7ebf4b02c4.

I believe the PR now reflects the consensus of the group, requesting a re-review from a few of you that requested changes.

index.html
msporny: Applied in 857f422125a07216ec4d3b2cd5e14b7ebf4b02c4.

andresuribe87 (approved):

index.html
```diff
@@ -569,26 +569,35 @@ <h3>Use Cases and Requirements</h3>
...
         </p>
 
         <p>
-<a>Verifiable credential</a> and <a>verifiable presentation</a> MUST be
-protected using a digital proof mechanism such as a digital signature. Having
-and verifying proofs, which may be dependent on the syntax of the proof (for
-example, using the JSON Web Signature of a JSON Web Token for proofing a key
-holder), are an essential part of processing <a>verifiable credentials</a> and
+<a>Verifiable credentials</a> and <a>verifiable presentations</a> MUST be
+protected using a securing mechanism as described in Section
```
seabass-labrax, line 596: Why was "digital proof mechanism" replaced with "securing mechanism"? This question does not affect my approval of this pull request; I am merely curious to know whether this is a matter of preference, or whether it is part of a change in terminology agreed upon by the VCWG.

seabass-labrax (approved):

index.html
msporny: The "digital proof mechanism" was from the v1.0 and v1.1 days, and that's what the VCWG (at that time) had settled upon for the language. The VCWG v2.0 charter shifted to using the term "Securing": https://www.w3.org/2022/06/verifiable-credentials-wg-charter.html#deliverables

... and all of our "securing" specifications have "securing", either in the title or the subtitle. I still need to make an editorial pass over the entire spec to update the language, but it seems the new direction is "securing" and "securing mechanisms".

Issue #1335, opened 2023-11-04
Recommend how authors set default language and direction


msporny:
Part of the Internationalization Horizontal Review noted that we need to suggest how a Verifiable Credential's default language and direction is specified. This issue is to ensure that we specify how this is done before we enter the CR phase.

msporny:
Implement [Option B](https://github.com/w3c/vc-data-model/issues/1264#issuecomment-1712807665).

msporny:
PR #1339 has been raised to address this issue. This issue will be closed once PR #1339 has been merged.

aphillips:
Adding our tracking label, since #1264 was tracked and is closed in favor of this issue.

msporny:
PR #1339 has been merged, closing.

PR #1339, opened 2023-11-09
Explain default language expression.

BigBlueHat:
This PR seeks to address https://github.com/w3c/vc-data-model/issues/1335 by presenting how authors may provide a default language and direction using standard JSON-LD language affordances.

***
<a href="https://pr-preview.s3.amazonaws.com/BigBlueHat/vc-data-model/pull/1339.html" title="Last updated on Nov 17, 2023, 2:43 PM UTC (b9d9115)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1339/998a3dc...BigBlueHat:b9d9115.html" title="Last updated on Nov 17, 2023, 2:43 PM UTC (b9d9115)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -5678,7 +5678,7 @@ transformation which is otherwise optional. See the <a
 data-cite="JSON-LD#string-internationalization">String Internationalization</a>
 section of the [[JSON-LD]] specification for more information. Natural language
 string values that do not have a language associated with them SHOULD be
-treated as if the language value is `Undefined` (language tag "`und`"). Natural
+treated as if the language value is `undefined` (language tag "`und`"). Natural
 language string values that do not have a base direction associated with them
 SHOULD be treated as if the direction value is "`auto`".
         </p>
--- a/index.html
+++ b/index.html
@@ -5665,7 +5665,7 @@ doing this is shown below.
         </pre>
       </section>
 
-      <section class="informative">
+      <section class="substantive">
         <h3>Providing Default Language and Direction</h3>
 
         <p>
--- a/index.html
+++ b/index.html
@@ -5672,7 +5672,7 @@ doing this is shown below.
 A default language and direction MAY be provided for the entire credential.
 However, doing so introduces a higher demand on creating a compliant processor
 and fully compliant context files. Consequently, using default language and
-direction should be limited to scenarios where JSON-LD expansion-based
+direction is expected to be be limited to scenarios where JSON-LD expansion-based
 transformation can be assured. See the
 <a href="https://www.w3.org/TR/json-ld11/#string-internationalization">String
 Internationalization</a> section of the [[JSON-LD]] specification for more
--- a/index.html
+++ b/index.html
@@ -5670,8 +5670,8 @@ doing this is shown below.
 
         <p>
 A default language and direction MAY be provided for the entire credential.
-However, doing so introduces a higher demand on creating a compliant processor
-and fully compliant context files. Consequently, using default language and
+However, doing so introduces a higher demand on context creators and 
+downstream processors. Consequently, using default language and
 direction is expected to be be limited to scenarios where JSON-LD expansion-based
 transformation can be assured. See the
 <a href="https://www.w3.org/TR/json-ld11/#string-internationalization">String
--- a/index.html
+++ b/index.html
@@ -5663,7 +5663,21 @@ doing this is shown below.
   "@id": "https://www.w3.org/2018/credentials/examples#title"
 }
         </pre>
+      </section>
 
+      <section class="informative">
+        <h3>Providing Default Language and Direction</h3>
+
+        <p>
+A default language and direction MAY be provided for the entire credential.
+However, doing so introduces a higher demand on creating a compliant processor
+and fully compliant context files. Consequently, using default language and
+direction should be limited to scenarios where JSON-LD expansion-based
+transformation can be assured. See the
+<a href="https://www.w3.org/TR/json-ld11/#string-internationalization">String
+Internationalization</a> section of the [[JSON-LD]] specification for more
+information.
+        </p>
       </section>
 
       <section class="informative">
--- a/index.html
+++ b/index.html
@@ -5674,7 +5674,7 @@ However, doing so introduces a higher demand on context creators and
 downstream processors. Consequently, using default language and
 direction is expected to be be limited to scenarios where JSON-LD expansion-based
 transformation can be assured. See the
-<a href="https://www.w3.org/TR/json-ld11/#string-internationalization">String
+<a date-cite="JSON-LD#string-internationalization">String
 Internationalization</a> section of the [[JSON-LD]] specification for more
 information.
         </p>
--- a/index.html
+++ b/index.html
@@ -5669,14 +5669,18 @@ doing this is shown below.
         <h3>Providing Default Language and Direction</h3>
 
         <p>
-A default language and direction MAY be provided for the entire credential.
-However, doing so introduces a higher demand on context creators and 
-downstream processors. Consequently, using default language and
-direction is expected to be be limited to scenarios where JSON-LD expansion-based
-transformation can be assured. See the
-<a date-cite="JSON-LD#string-internationalization">String
-Internationalization</a> section of the [[JSON-LD]] specification for more
-information.
+The language and base direction of each natural language string property SHOULD
+be provided, either via the language value structure for each property or via a
+default language and base direction for the entire credential. Using the
+language value structure is preferred, because document defaults can result in
+a requirement that downstream processors perform JSON-LD expansion-based
+transformation which is otherwise optional. See the <a
+data-cite="JSON-LD#string-internationalization">String Internationalization</a>
+section of the [[JSON-LD]] specification for more information. Natural language
+string values that do not have a language associated with them SHOULD be
+treated as if the language value is `Undefined` (language tag "`und`"). Natural
+language string values that do not have a base direction associated with them
+SHOULD be treated as if the direction value is "`auto`".
         </p>
       </section>
 


dlongley (approved):

index.html
```diff
@@ -5663,7 +5663,21 @@ <h3>Language and Base Direction</h3>
...
 
+        <p>
+A default language and direction MAY be provided for the entire credential.
+However, doing so introduces a higher demand on creating a compliant processor
+and fully compliant context files. Consequently, using default language and
+direction should be limited to scenarios where JSON-LD expansion-based
+transformation can be assured. See the
+<a href="https://www.w3.org/TR/json-ld11/#string-internationalization">String
+Internationalization</a> section of the [[JSON-LD]] specification for more
+information.
```
aphillips, line 5679: Hmm... we're almost back where we started.

The `name` and `description` fields really must have a language and base direction and consumers of credentials need to be able to find out what it is. Either the document has a default for these or it uses the language value structure. "Neither" should not be an option, but that is what this MAY (combined with the rest of the document) results in--particularly since this paragraph makes it sound like a bad idea to provide the default.

index.html
```diff
@@ -5663,7 +5663,21 @@ <h3>Language and Base Direction</h3>
   "@id": "https://www.w3.org/2018/credentials/examples#title"
 }
         </pre>
+      </section>
+
+      <section class="informative">
```
aphillips, line 5668: this says "informative", but the section contains a 2119 keyword `MAY`?

aphillips (changes requested):
"Houston, we have a problem."

index.html
msporny, line 5668:
```suggestion
      <section class="substantive">
```

index.html
```diff
@@ -5663,7 +5663,21 @@ <h3>Language and Base Direction</h3>
...
+        <h3>Providing Default Language and Direction</h3>
 
+        <p>
+A default language and direction MAY be provided for the entire credential.
+However, doing so introduces a higher demand on creating a compliant processor
+and fully compliant context files. Consequently, using default language and
+direction should be limited to scenarios where JSON-LD expansion-based
```
msporny, line 5675:
```suggestion
direction is expected to be be limited to scenarios where JSON-LD expansion-based
```

Avoid use of RFC 2119 language, even if it's lowercase; it makes the normative intent of the prose unclear.

index.html
```diff
@@ -5663,7 +5663,21 @@ <h3>Language and Base Direction</h3>
...
         </pre>
+      </section>
+
+      <section class="informative">
+        <h3>Providing Default Language and Direction</h3>
 
+        <p>
+A default language and direction MAY be provided for the entire credential.
+However, doing so introduces a higher demand on creating a compliant processor
+and fully compliant context files. Consequently, using default language and
```
msporny, line 5674: Not sure this is about "compliance", which is around syntax... you can have a data model that's fully compliant, but where the language information and semantics are wrong. It's the latter that we're concerned about.

```suggestion
However, doing so introduces a higher demand on context creators and 
downstream processors. Consequently, using default language and
```

index.html
```diff
@@ -5663,7 +5663,21 @@ <h3>Language and Base Direction</h3>
...
+direction should be limited to scenarios where JSON-LD expansion-based
+transformation can be assured. See the
+<a href="https://www.w3.org/TR/json-ld11/#string-internationalization">String
```
msporny, line 5677: I think we can do something like this instead (please check the syntax, don't know if it's accurate):

```suggestion
<a date-cite="JSON-LD#string-internationalization">String
```

index.html
msporny: Perhaps we could do a "MUST either provide human-readable fields via the language value structure (preferred) or via document defaults"?

index.html
BigBlueHat: Fixed.

BigBlueHat:
Suggests processed. Thanks @aphillips and @msporny!

index.html
```diff
@@ -5663,7 +5663,21 @@ <h3>Language and Base Direction</h3>
...
 
+        <p>
+A default language and direction MAY be provided for the entire credential.
+However, doing so introduces a higher demand on context creators and 
+downstream processors. Consequently, using default language and
+direction is expected to be be limited to scenarios where JSON-LD expansion-based
+transformation can be assured. See the
+<a date-cite="JSON-LD#string-internationalization">String
+Internationalization</a> section of the [[JSON-LD]] specification for more
+information.
```
aphillips, line 5679: I don't agree that this text fixes the problem, as it is still "MAY" with harsh sounding warning afterwards. Perhaps:

```suggestion
The language and direction of each natural language string field MUST be provided 
via the language value structure or via a default language and direction
for the entire credential. Using the language value structure is preferred because document
defaults can result in the need to perform JSON-LD expansion-based transformation by
downstream processors. See the <a data-cite="JSON-LD#string-internationalization">String
Internationalization</a> section of the [[JSON-LD]] specification for more information.

**Review by aphillips — CHANGES_REQUESTED** [review]
See suggestion

File: index.html (hunk 6 above)
Comment (BigBlueHat) [reply index.html:5679]: @aphillips I'm not sure we can make this a MUST as it would effectively force all v2 Verifiable Credentials to contain language metadata which they may not need/use/etc. It's a bit analogous to if `lang` on `<html>` were to suddenly become a requirement for pages to render.

I'd prefer all data come with proper language metadata...but I'm not sure this is the layer at which to force it into existence.

**Review by decentralgabe — APPROVED** [review]

File: index.html (hunk 1 above)
Comment (TallTed) [reply index.html]: `SHOULD provide language and base direction for human-readable fields either via the language value structure (preferred) or via document defaults` because the MUST horse has [long since left the barn](https://github.com/w3c/vc-data-model/pull/1339#discussion_r1391390907).

**Comment by iherman** [comment]
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-11-14-vcwg-special#section1-9) on 2023-11-14

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 1.9. Explain default language expression. (pr vc-data-model#1339)

_See github pull request [vc-data-model#1339](https://github.com/w3c/vc-data-model/pull/1339)._




**Brent Zundel:** Benjamin, can you give status update?  

**Benjamin Young:** this one got good feedback from Addison, and there is a suggestion about requiring language definition, which would be a major change for everybody.  

> *Chris Abernethy:* .. that is the piece that needs discussion - otherwise approvals throughout.

**Manu Sporny:** my one concern is that there are two ways to specify.  
… best way is to be precise. The other way is `@language` and `@direction` in the `@context` which will set it blanket for the entire doc.  
… both of Benjamin's PRs say use one or the other, up to you. Addison is saying no to that.  
… question to group: do we want to mandate saying you MUST set the `@language` for every type of VC that you issue?  

> *Dave Longley:* IMO, you can't force people to do the right thing here -- you will just get bad language data (e.g., English as the default language, but the content will be in another language).

> *Dave Longley:* +1 to Joe, MUST is aspirational that won't be attained.

**Joe Andrieu:** I think MUST is something we probably would not be able to realize, and would lead to lots of non-conforming implementations and ignoring on the front-end.  
… agree it should be there, just don't think MUST is feasible.  

**Benjamin Young:** +1 to what was said, HTML, JSON, JS don't do this. Providing language as a capability is good, requiring it is a "bridge too far".  

**Sebastian Crane:** as we are releasing 2.0 to the world, this is the best time to promote language support.  
… maybe people are right, but this is probably the best time to try and support making this a MUST.  

**Ted Thibodeau Jr.:** making this a MUST means derivative certs are not doable.  
… SHOULD is almost a MUST, e.g., do it unless you have a good reason not to.  
… VCs are not the path to force I18N on the world.  

**Dave Longley:** agree with TallTed, this will hurt more than help, and we can't force people to do this.  

**Sebastian Crane:** question - where are the difficulties expected to be here?  
… I know what language I am using, so it seems easy. Can someone explain where the pain points are here?  

**Brent Zundel:** encourage folks to respond to seabass in the PR.  

</details>

File: index.html (hunk 6 above)
Comment (dlongley) [reply index.html:5679]: @jandrieu,

> I agree with SHOULD on this

Well, if we change it from MAY to SHOULD, then we need it to be you SHOULD use the default language if and only if you don't use the localized language objects (as those are preferred). So we'd need some additional tweaks, we can't just change MAY => SHOULD here.

Comment (aphillips) [reply index.html:5679]: Since this is 2.0, I can understand making MUST => SHOULD in my suggested text as a sop to existing credentials that do not include language metadata. 

I would clarify that this part of @BigBlueHat's comment:

> as it would effectively force all v2 Verifiable Credentials to contain language metadata which they may not need/use/etc.

I don't care if "they" use the metadata. Some consumers are just passing the data or checking the validity dates, after all. If the right credential was received, the display name isn't important. But I do care that the metadata is present for when consumers do access the natural language strings. That's not really a "SHOULD" thing.

What should be added if we make this a SHOULD is some additional text at the end:

```suggestion
The language and direction of each natural language string field SHOULD be provided 
via the language value structure or via a default language and direction
for the entire credential. Using the language value structure is preferred because document
defaults can result in the need to perform JSON-LD expansion-based transformation by
downstream processors. See the <a data-cite="JSON-LD#string-internationalization">String
Internationalization</a> section of the [[JSON-LD]] specification for more information.
Natural language string values that do not have a language associated with them
SHOULD treat the language as Undefined (language tag `und`). Natural language string values
that do not have a base paragraph direction associated with them SHOULD treat the value
as having a direction of "auto".
```

dlongley: +1 to @aphillips suggestion.

msporny: Excellent, thank you @aphillips -- I believe that will achieve consensus given the discussion the group had yesterday.

TallTed: Tweaks to @aphillips [suggestion](https://github.com/w3c/vc-data-model/pull/1339/files#r1393063076)

```suggestion
The language and base direction of each natural language string property SHOULD
be provided, either via the language value structure for each property or via a
default language and base direction for the entire credential. Using the
language value structure is preferred, because document defaults can result in
a requirement that downstream processors perform JSON-LD expansion-based
transformation which is otherwise optional. See the <a
data-cite="JSON-LD#string-internationalization">String Internationalization</a>
section of the [[JSON-LD]] specification for more information. Natural language
string values that do not have a language associated with them SHOULD be
treated as if the language value is `Undefined` (language tag "`und`"). Natural
language string values that do not have a base direction associated with them
SHOULD be treated as if the direction value is "`auto`".
```

dlongley (approved):

index.html
```diff
@@ -5663,7 +5663,25 @@ <h3>Language and Base Direction</h3>
...
+The language and base direction of each natural language string property SHOULD
+be provided, either via the language value structure for each property or via a
+default language and base direction for the entire credential. Using the
+language value structure is preferred, because document defaults can result in
+a requirement that downstream processors perform JSON-LD expansion-based
+transformation which is otherwise optional. See the <a
+data-cite="JSON-LD#string-internationalization">String Internationalization</a>
+section of the [[JSON-LD]] specification for more information. Natural language
+string values that do not have a language associated with them SHOULD be
+treated as if the language value is `Undefined` (language tag "`und`"). Natural
```
msporny, line 5681:
```suggestion
treated as if the language value is `undefined` (language tag "`und`"). Natural
```

msporny (approved):

msporny:

PR #1340, opened 2023-11-09
Use JSON-LD keywords for languages.

BigBlueHat:
This PR uses the non-aliased values for `@language`, `@direction`, and `@value` to prevent confusion that these aliased terms may be used anywhere for any purpose--instead highlighting their sole application for providing language metadata for a single string value.

***
<a href="https://pr-preview.s3.amazonaws.com/BigBlueHat/vc-data-model/pull/1340.html" title="Last updated on Nov 20, 2023, 2:37 PM UTC (6c008dc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1340/3fb62f8...BigBlueHat:6c008dc.html" title="Last updated on Nov 20, 2023, 2:37 PM UTC (6c008dc)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -123,7 +123,7 @@
             w3cid: 46156
           }
         ],
-        xref: ["url"],
+        xref: ["url", "i18n-glossary"],
         maxTocLevel: 2,
         inlineCSS: true
       };
@@ -1560,9 +1560,9 @@ without their having to look through the entirety of the <a>claims</a>.
 
         <p>
 Names and descriptions also support expressing content in different languages.
-To express a string with language and text direction information, one can use
+To express a string with language and <a>base direction</a> information, one can use
 an object that contains the `@value`, `@language`, and `@direction` properties
-to express the text value, language tag, and text direction, respectively.
+to express the text value, language tag, and base direction, respectively.
 See <a href="#language-and-base-direction"></a> for further information.
         </p>
 
@@ -5658,7 +5658,7 @@ document [[STRING-META]] that implementers might be interested in reading.
 Data publishers are strongly encouraged to read the section on
 Cross-Syntax Expression in the <em>Strings on the Web: Language and Direction
 Metadata</em> document [[STRING-META]] to ensure that the expression of
-language and base direction information is possible across multiple expression
+language and <a>base direction</a> information is possible across multiple expression
 syntaxes, such as [[JSON-LD]], [[JSON]], and CBOR [[?RFC7049]].
         </p>
 
@@ -5681,7 +5681,7 @@ When the language value object is used in place of a string value, the object
 MUST contain a `@value` property whose value is a string, and SHOULD contain a
 `@language` property whose value is a string containing a valid
 `Language-Tag` as defined by [[BCP47]], and MAY contain a `@direction` property
-whose value is a text direction string defined by the `@direction` attribute in
+whose value is a <a>base direction</a> string defined by the `@direction` attribute in
 [[JSON-LD]]. The language value object MUST NOT include any other keys beyond
 `@value`, `@language`, and `@direction`.
         </p>
@@ -5714,7 +5714,7 @@ base direction of right-to-left.
         <p class="note">
 The text above would most likely be rendered incorrectly as left-to-right
 without the explicit expression of language and direction because many systems
-use the first character of a text string to determine text direction.
+use the first character of a text string to determine its <a>base direction</a>.
         </p>
 
         <p>
--- a/index.html
+++ b/index.html
@@ -5760,10 +5760,9 @@ SHOULD be treated as if the direction value is "`auto`".
         <h3>Complex Language Markup</h3>
 
         <p>
-When a single natural language string contains multiple languages, base
-directions, and annotations, a more complex mechanism is typically required.
-It is possible to use markup languages, such as HTML, to encode text with
-multiple languages and base directions. It is also possible to use the
+When a single natural language string contains multiple languages or annotations,
+the contents of the string might require additional structure or markup in order to be
+presented correctly. It is possible to use markup languages, such as HTML, to label spans of text in different languages or to supply string-internal markup needed for proper display of <a>bidirectional text</a>. It is also possible to use the
 <code>rdf:HTML</code> datatype to encode such values accurately in JSON-LD.
         </p>
 
--- a/index.html
+++ b/index.html
@@ -5679,7 +5679,7 @@ specific base direction.
         <p>
 When the language value object is used in place of a string value, the object
 MUST contain a `@value` property whose value is a string, and SHOULD contain a
-`@language` property whose value is a string containing a valid
+`@language` property whose value is a string containing a well-formed
 `Language-Tag` as defined by [[BCP47]], and MAY contain a `@direction` property
 whose value is a <a>base direction</a> string defined by the `@direction` attribute in
 [[JSON-LD]]. The language value object MUST NOT include any other keys beyond
--- a/index.html
+++ b/index.html
@@ -5705,7 +5705,7 @@ base direction of right-to-left.
 
         <pre class="example nohighlight" title="Arabic text with a base direction of right-to-left">
 "title": {
-  "@value": "<span class="highlight">HTML و CSS: تصميم و إنشاء مواقع الويب</span>",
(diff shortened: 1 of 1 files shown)


dlongley (approved):

OR13 (approved):

index.html
```diff
@@ -5610,21 +5613,31 @@ <h3>Language and Base Direction</h3>
...
+  "@direction": "<code>DIRECTION</code>"
 }
         </pre>
 
+        <p>
+When the language value object is used in place of a string value, the object
+MUST contain a `@value` property whose value is a string, and SHOULD contain a
+`@language` property whose value is a string containing a well-formed
+`Language-Tag` as defined by [[BCP47]], and MAY contain a `@direction` property
+whose value is a text direction string defined by the `@direction` attribute in
```
aphillips, line 5627: Consider linking "text direction" to the definition in the I18N glossary.

index.html
```diff
@@ -5610,21 +5613,31 @@ <h3>Language and Base Direction</h3>
...
+  "@value": "<span class="highlight">The string value</span>",
+  "@language": "<code>LANGUAGE</code>"
+  "@direction": "<code>DIRECTION</code>"
 }
         </pre>
 
+        <p>
+When the language value object is used in place of a string value, the object
+MUST contain a `@value` property whose value is a string, and SHOULD contain a
+`@language` property whose value is a string containing a well-formed
```
aphillips, line 5625: Consider making "well-formed" be "valid" here. Conformance should only require well-formed, but authors should choose valid language tags. If it is too difficult to make the emit/consume distinction here, this text is fine.

index.html
```diff
@@ -5635,9 +5648,9 @@ <h3>Language and Base Direction</h3>
 
         <pre class="example nohighlight" title="Arabic text with a base direction of right-to-left">
 "title": {
-  "value": "<span class="highlight">HTML و CSS: تصميم و إنشاء مواقع الويب</span>",
-  "lang": "<code>ar</code>"
-  "dir": "<code>rtl</code>"
+  "@value": "<span class="highlight">HTML و CSS: تصميم و إنشاء مواقع الويب</span>",
```
aphillips, line 5651: You want a `dir` on the span here so the string displays RTL correctly :-)

```suggestion
  "@value": "<span class="highlight" dir="rtl" lang="ar">HTML و CSS: تصميم و إنشاء مواقع الويب</span>",
```

index.html
```diff
@@ -5648,32 +5661,33 @@ <h3>Language and Base Direction</h3>
...
-  "@id": "https://www.w3.org/2018/credentials/examples#title"
-}
+        <pre class="example nohighlight" title="Multiple language texts provided for title">
+"title": [
+  {
+    "@value": "<span class="highlight">HTML and CSS: Designing and Creating Websites</span>",
+    "@language": "<code>en</code>"
+  },
+  {
+    "@value": "<span class="highlight">HTML و CSS: تصميم و إنشاء مواقع الويب</span>",
```
aphillips, line 5675: see above

index.html
```diff
@@ -5648,32 +5661,33 @@ <h3>Language and Base Direction</h3>
...
 
         <p>
-When multiple languages, base directions, and annotations are used in a single
-natural language string, more complex mechanisms are typically required. It is
-possible to use markup languages, such as HTML, to encode text with multiple
-languages and base directions. It is also possible to use the
+When a single natural language string contains multiple languages, base
+directions, and annotations, a more complex mechanism is typically required.
+It is possible to use markup languages, such as HTML, to encode text with
+multiple languages and base directions. It is also possible to use the
```
aphillips, line 5690: The text here about base direction is wrong. Any paragraph--even one in multiple languages and scripts--has only one base paragraph direction. Inline directional changes that are part of the paragraph do exist and can require care on the part of the author, but this is outside the scope of string metadata. Perhaps:

```suggestion
When a single natural language string contains multiple languages or annotations,
the contents of the string might require additional structure or markup in order to be
presented correctly. It is possible to use markup languages, such as HTML, to label spans of text in different languages or to supply string-internal markup needed for proper display of <a>bidirectional text</a>. It is also possible to use the
```

aphillips (changes requested):
Good start.

Where is the information on credential-level default language/direction (for cases where `name` and `direction` are just strings)?

index.html
msporny, line 5625: The change below applies @aphillips request:

```suggestion
`@language` property whose value is a string containing a valid
```

We are trying very hard to stay away from any usage of "well-formed" text in the spec.

index.html
msporny: > You want a `dir` on the span here so the string displays RTL correctly :-)

Nope, that "highlight" is ReSpec markup on the example itself, not the markup that will be displayed in the spec. :)

We're highlighting the text that `@language` and `@direction` is being applied to... we're not using HTML to express `lang` and `dir`... and yes, it's difficult to see that just by staring at the source code in the spec. :)

index.html
msporny: See response to "see above" :) -- https://github.com/w3c/vc-data-model/pull/1340/files#r1390255038

index.html
aphillips: Not sure I understand. The `<span>` does end up in the rendered page (Respec is using it to style the string). By adding the dir and lang it makes the string display correctly (as if the @language and @direction had been applied):

image

index.html
aphillips: "well-formed" and "valid" are defined terms in BCP47 (see [here](https://www.rfc-editor.org/rfc/rfc5646.html#section-2.2.9)). They are also defined terms in the i18n glossary. Generally VC shouldn't require implementations to check for validity, because that involves reading the IANA Language Subtag Registry--you actually want conformance to hinge on well-formed language tags. But you wan users to pick language tags that are valid.

In this case, though, I can agree with the change provided you're aware of what you want to require people to demonstrate.

index.html
BigBlueHat: @aphillips there's no discrete "text direction" definition in the I18N glossary. Would you recommend I reference [base direction](https://www.w3.org/TR/i18n-glossary/#dfn-base-direction) or [paragraph direction](https://www.w3.org/TR/i18n-glossary/#dfn-paragraph-direction) instead?

aphillips: @BigBlueHat Use base direction here.

index.html
BigBlueHat: Fixed as @aphillips suggested.

index.html
BigBlueHat: k. Putting this back to "well-formed".

index.html
```diff
@@ -5635,45 +5648,45 @@ <h3>Language and Base Direction</h3>
 
         <pre class="example nohighlight" title="Arabic text with a base direction of right-to-left">
 "title": {
-  "value": "<span class="highlight">HTML و CSS: تصميم و إنشاء مواقع الويب</span>",
-  "lang": "<code>ar</code>"
-  "dir": "<code>rtl</code>"
+  "@value": "<span class="highlight">HTML و CSS: تصميم و إنشاء مواقع الويب</span>",
```
aphillips, line 5651: You didn't implement the fix here? (It is correct on line 5675)

BigBlueHat:
> Where is the information on credential-level default language/direction (for cases where `name` and `direction` are just strings)?

It's clear you found #1339. 😃 

Thanks for all the help on these, @aphillips and @msporny. All your requests should be addressed now. Cheers!

index.html
BigBlueHat: I did in https://github.com/w3c/vc-data-model/pull/1340/commits/0222763398dd0d2d505ad570598e0a10b05aa410

BigBlueHat: Oh. Missed the other line. My bad. Fixing.

aphillips (approved):
Looks good! Thanks @BigBlueHat

dlongley (approved):

brianorwhatever (approved):

msporny:
This needs a rebase before I can merge it, @BigBlueHat.

BigBlueHat:
@msporny rebased and ready when you are.

msporny:

Issue #1343, opened 2023-11-09
Confused by "all times MUST be specified as offsets against Universal Coordinated Time (UTC)."


jyasskin:
In https://w3c.github.io/vc-data-model/#representing-time, we have

> In order to reduce misinterpretations between different regions, all times MUST be specified as offsets against Universal Coordinated Time (UTC).

I thought that might be saying that the timezoneFrag had to be just "Z", but the example in the next paragraph has `-07:00`. Then I thought you might be excluding use of named timezones, but dateTimeStamp doesn't allow those at all. So what constraint is this line trying to add?

msporny:
The normative statement is attempting to say "You MUST NOT specify a datetime that does not include a time zone offset.". That is, `dateTime` allows times like `2023-01-01T00:00:00` (no timezone offset against UTC), which means you interpret the time according to your local timezone (which could be off by 14 hours -- I think -- based on weird daylight savings times differences on opposite sides of the world). 

I'll try to raise a PR that's more clear with the intent... feel free to propose some text that would have not led to you being puzzled by the statement.

jyasskin:
I think you just don't need anything normative here, now that the spec uses `dateTimeStamp` instead of `dateTime`.

msporny:
PR #1344 has been raised to address this issue. This issue will be closed once PR #1344 has been merged.

msporny:
> I think you just don't need anything normative here, now that the spec uses `dateTimeStamp` instead of `dateTime`.

The statement is meant to cover /any/ time value expressed in a VC. I understand that is quite a reach for a spec, we could downgrade it to a SHOULD, or get rid of it.

Thoughts @aphillips and @r12a?

jyasskin:
Ooooh. That's plausible, but I'd expect it to be in a separate section that talks about the claims that a VC can make, rather than a subsection of the `validFrom` and `validUntil` section.

Making this a MUST increases the difficulty of having a verifier enforce that a document is conformant, since it says that the verifier has to recognize all the ways a VC might encode a time, and check that those ways include a timezone. E.g. do verifiers have to notice that a field contains an encoded ASN.1 [GeneralizedTime](https://en.wikipedia.org/wiki/GeneralizedTime)?

aphillips:
[...]

PR #1344, opened 2023-11-09
Clarify normative statement related to timezone offsets.

msporny:
This PR attempts to address issue #1343 by clarifying the normative statement related to timezone offsets.

/cc @jyasskin

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1344.html" title="Last updated on Nov 12, 2023, 10:42 PM UTC (3cb0665)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1344/998a3dc...3cb0665.html" title="Last updated on Nov 12, 2023, 10:42 PM UTC (3cb0665)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1917,8 +1917,11 @@ combinations of dates and times, and reduce the potential for
 misinterpretation of time values, by utilizing the
 <code>dateTimeStamp</code> construction first established by the [<a
 data-cite="XMLSCHEMA11-2#dateTimeStamp">XMLSCHEMA11-2</a>] specification. In
-order to reduce misinterpretations between different regions, all times MUST be
-specified as offsets against Universal Coordinated Time (UTC).
+order to reduce misinterpretations between different time zones, any time
+value MUST specify a time zone offset relative to Universal Coordinated Time
+(UTC). Examples of valid time zone offsets relative to UTC include
+`Z`, `+01:00`, `-08:00`, and `+14:00`. See the regular expression at the
+end of this section for a formal definition of all acceptable values.
         </p>
 
         <p>
--- a/index.html
+++ b/index.html
@@ -1917,10 +1917,12 @@ combinations of dates and times, and reduce the potential for
 misinterpretation of time values, by utilizing the
 <code>dateTimeStamp</code> construction first established by the [<a
 data-cite="XMLSCHEMA11-2#dateTimeStamp">XMLSCHEMA11-2</a>] specification. In
-order to reduce misinterpretations between different time zones, any time
-value MUST specify a time zone offset relative to Universal Coordinated Time
-(UTC). Examples of valid time zone offsets relative to UTC include
-`Z`, `+01:00`, `-08:00`, and `+14:00`. See the regular expression at the
+order to reduce misinterpretations between different time zones, all time values
+SHOULD be specified in `dateTimeStamp` format, either in Universal Coordinated
+Time (UTC), denoted by a `Z` at the end of the value, or with a time zone offset
+relative to UTC. Time values that are incorrectly serialized without an offset
+MUST be interpreted as UTC. Examples of valid time zone offsets relative to UTC
+include `Z`, `+01:00`, `-08:00`, and `+14:00`. See the regular expression at the
 end of this section for a formal definition of all acceptable values.
         </p>
 
--- a/index.html
+++ b/index.html
@@ -1918,12 +1918,13 @@ misinterpretation of time values, by utilizing the
 <code>dateTimeStamp</code> construction first established by the [<a
 data-cite="XMLSCHEMA11-2#dateTimeStamp">XMLSCHEMA11-2</a>] specification. In
 order to reduce misinterpretations between different time zones, all time values
-SHOULD be specified in `dateTimeStamp` format, either in Universal Coordinated
-Time (UTC), denoted by a `Z` at the end of the value, or with a time zone offset
-relative to UTC. Time values that are incorrectly serialized without an offset
-MUST be interpreted as UTC. Examples of valid time zone offsets relative to UTC
-include `Z`, `+01:00`, `-08:00`, and `+14:00`. See the regular expression at the
-end of this section for a formal definition of all acceptable values.
+expressed in <a>conforming documents</a> SHOULD be specified in `dateTimeStamp`
+format, either in Universal Coordinated Time (UTC), denoted by a `Z` at the end
+of the value, or with a time zone offset relative to UTC. Time values that are
+incorrectly serialized without an offset MUST be interpreted as UTC. Examples of
+valid time zone offsets relative to UTC include `Z`, `+01:00`, `-08:00`, and
+`+14:00`. See the regular expression at the end of this section for a formal
+definition of all acceptable values.
         </p>
 
         <p>


jyasskin (approved):
Thanks, this is clearer.

index.html
```diff
@@ -1917,8 +1917,11 @@ <h4>Representing Time</h4>
 misinterpretation of time values, by utilizing the
 <code>dateTimeStamp</code> construction first established by the [<a
 data-cite="XMLSCHEMA11-2#dateTimeStamp">XMLSCHEMA11-2</a>] specification. In
-order to reduce misinterpretations between different regions, all times MUST be
-specified as offsets against Universal Coordinated Time (UTC).
+order to reduce misinterpretations between different time zones, any time
+value MUST specify a time zone offset relative to Universal Coordinated Time
+(UTC). Examples of valid time zone offsets relative to UTC include
+`Z`, `+01:00`, `-08:00`, and `+14:00`. See the regular expression at the
+end of this section for a formal definition of all acceptable values.
```
aphillips, line 1924: This is fine, albeit a lot of detail for your spec. 

This doesn't say what happens when a given VC credential ignores it (is the credential invalid??) One work around for this would be to specify that time values must include an offset (with UTC preferred) and separately specify that values without an offset will be interpreted as UTC. 

Also, note that temporal serializations are coming, which include true time zone IDs (offsets have many problems). The I-D isn't an RFC yet, so you can't reference it here, but be aware of it for revisions in the future. The WG there is called [SEDATE](https://datatracker.ietf.org/doc/draft-ietf-sedate-datetime-extended/)

iherman (approved):

index.html
msporny: > This is fine, albeit a lot of detail for your spec.

Yes. :P

When people raise issues, we try to be crystal clear in the changes, and that has led to a verbose specification (that some people are cranky about)... but if we remove the detail, we end up not addressing the concerned party's needs/concerns. Such is the specification editing process. :)

> This doesn't say what happens when a given VC credential ignores it (is the credential invalid??) One work around for this would be to specify that time values must include an offset (with UTC preferred) and separately specify that values without an offset will be interpreted as UTC.

Good idea, that covers all of our bases AFAICT. I'll implement it in an upcoming modification to this PR.

It would allow us to weaken the existing MUST to a SHOULD, but still ensure a single "correct" way to interpret all time values. I'm on the fence wrt. weakening the existing MUST to a SHOULD -- thoughts, @jyasskin and @aphillips? I could go either way.

> Also, note that temporal serializations are coming, which include true time zone IDs (offsets have many problems). The I-D isn't an RFC yet, so you can't reference it here, but be aware of it for revisions in the future. The WG there is called [SEDATE](https://datatracker.ietf.org/doc/draft-ietf-sedate-datetime-extended/)

Interesting... if we were to switch to IXDTF syntax in the future, we'd have to move away from XML Schema dateTimeStamp (and the ref to that spec) and move back to RFC3339 (which we had to move away from for reasons I can't remember right now). That is definitely a VC v3.0 thing, so let's let things settle in SEDATE-land before we make any changes. I don't forsee any spec issues if we had to switch given that dateTimeStamps are a subset of IXDTF syntax. The biggest issue is deployed infrastructure choking on IXDTF datetimes... we'd have to make sure IXDTF parsing is in a large variety of datetime libraries before shipping that feature.

aphillips: > It would allow us to weaken the existing MUST to a SHOULD, but still ensure a single "correct" way to interpret all time values. I'm on the fence wrt. weakening the existing MUST to a SHOULD -- thoughts, @jyasskin and @aphillips? I could go either way.

I'd be reluctant to reduce MUST to SHOULD as well. Perhaps something like:

```suggestion
All time values MUST be specified as dateTimeStamp, either in UTC (denoted by `Z`) 
or with a time zone offset relative to Universal Coordinated Time. 
Time values incorrectly serialized without an offset MUST be interpreted as UTC.
```

(above suggestion is not meant to be merged, as it needs references and such)

> Interesting... if we were to switch to IXDTF syntax in the future, we'd have to move away from XML Schema dateTimeStamp (and the ref to that spec) and move back to RFC3339 (which we had to move away from for reasons I can't remember right now). That is definitely a VC v3.0 thing, so let's let things settle in SEDATE-land before we make any changes. I don't forsee any spec issues if we had to switch given that dateTimeStamps are a subset of IXDTF syntax. The biggest issue is deployed infrastructure choking on IXDTF datetimes... we'd have to make sure IXDTF parsing is in a large variety of datetime libraries before shipping that feature.

I suspect Temporal will be as popular in JS as it has been in other programming environments. This will likely lead to various serializations getting updated too. RFC3339 is just ahead of the curve here. You have the right idea: wait for the dust to settle.

TallTed (approved):
LGTM after tweaks as requested by @aphillips

index.html
msporny: Applied your changes in 91339e127d533724a276663a805ad7db882043d7.

longpd (approved):

OR13 (approved):

aphillips (approved):
LGTM! Thanks!

dlongley (approved):

decentralgabe (approved):

jandrieu:
Seems overly broad to *all* time values. Would infect custom values.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-11-14-vcwg-special#section1-10) on 2023-11-14

- no resolutions were taken

#### 1.10. Clarify normative statement related to timezone offsets. (pr vc-data-model#1344)

_See github pull request [vc-data-model#1344](https://github.com/w3c/vc-data-model/pull/1344)._




**Brent Zundel:** this PR is broadly approved with no unresolved change requests.  
… we are going to merge in a couple of days.  

**Manu Sporny:** heads up to the group, this is also significant. When you express a time value in any VC, you should express it as a date timestamp formatted value.  
… anyone verifying, if they do not see a relative UTC offset, has to interpret it as UTC.  
… this means you cannot express (should not express) a date that does not have a timezone offset at the end of it for any time value in a VC.  

**Joe Andrieu:** also commented in github - would like to see this scoped to just the properties we are defining in the VCDM. Don't want to see this on custom properties, or.  

**Manu Sporny:** that is the intention. we are trying to increase interop.  

**Brent Zundel:** let's continue the conversation on the PR.  
… please be aware of and review other PRs that we didn't get a chance to review today.  

---

</details>

msporny:
> @msporny You seemed to say on the call today that it is the intention of this PR to apply to included vocabularies used for claims in VCs. So that we are explicitly discouraging extending a VC with a formally defined @context with its own definitions for various temporal properties. [Ivan thought you misheard me, so please clarify if I misheard you.]

I think you misheard me, but maybe not... here's what I was saying:

I want people to pay special attention to this PR because we're providing guidance on how people should express time values. We are saying two things:

1. If you're generating conforming documents, you SHOULD use dateTimeStamp to express time values. 
2. If you are consuming conforming documents, and someone has expressed a time in a way that doesn't specify any timezone/timeoffset information, you MUST presume the timezone/timeoffset is UTC.

We are not saying that someone can't extend the data model and have properties for `myTimeIsBetterThanYours` or `asnValidFrom`. We're just saying that when they do that, please use `dateTimeStamp` to express the value, and regardless of whether you do that or not, that time value, when read, and if it doesn't include a timezone, will be interpreted as UTC.

So, we're trying to strike a balance here to 1) not be overbearing in how extensions express time values (but provide strong guidance), and 2) be very strict in giving people fair warning that if they don't include a timezone/offset, that their times will be interpreted as UTC.

> I believe this would be the first time we are suggesting restricting extended vocabularies beyond locking down specific properties of the VCDM.

We already tell people where to put properties that express things on the VC vs. the credential subject, and where to put fields that are associated with extension points like `credentialStatus` -- those objects MUST contain (`type`, at a minimum). So, I do think we specify what extensions MUST do in some cases.

> The result would be to bias the ecosystem against vocabularies that have a reason to represent time differently.

I mean, we still allow it... what's the compelling use case for representing time differently? Remember that we've had decades of standards on trying to NOT represent time differently (ISO8601, RFC3339, XML Datetime) -- they all (more or less) follow the same format with some fairly esoteric differences.

> That would be an untenable position, given our commitment to allow anyone to say anything.

We still allow anyone to say anything. I don't know how this PR would prevent that?

> This restriction would say that although you are conceptually ALLOWED to say anything, you SHOULDN'T. Instead you SHOULD only use vocabularies that meet certain rules.

The statement above seems much stronger than what the PR is doing... don't know if any of the explanatory text above helped?

> This is not just a bad idea, IMO, it violates the core premise and promise of VCs. As long as the extensions to the VCDM do not override required values, they should be encouraged as first class uses of the technology instead of biasing the ecosystem against such extendibility.

I don't think we're biasing against extensibility -- we're just saying, if you express time values, you SHOULD do them in this way. We're doing that to increase interop, because this specific thing (not including timezone information) has shown up in deployed systems and it's just a matter of time before it starts biting the ecosystem.

Hope the above helps, if not, and your concern is a blocking concern, let me know sooner than later.

dlongley:
> We are not saying that someone can't extend the data model and have properties for `myTimeIsBetterThanYours` or `asnValidFrom`. We're just saying that when they do that, please use `dateTimeStamp` to express the value, and regardless of whether you do that or not, that time value, when read, and if it doesn't include a timezone, will be interpreted as UTC.

I would tweak this by saying: "...when read, and if it doesn't include a timezone, and no custom advice is given for `myTimeIsBetterThanYours` for how to handle that, it will be interpreted as UTC."

TallTed:
I'm with @dlongley, with this addition --
`s/when read, and if it doesn't/when read, if it doesn't/`
(i.e., drop the misplaced `and`)

msporny:
I'm waiting to hear back from @jandrieu to ensure he's not objecting before merging this PR.

jandrieu (approved):
I'm good with this as a SHOULD.

msporny:

Issue #1342, opened 2023-11-09
Issuer appears to be defined using a non-JSON-LD data model


jyasskin:
https://w3c.github.io/vc-data-model/#issuer says

> The value of the issuer [property](https://w3c.github.io/vc-data-model/#dfn-property) MUST be either a [URL](https://w3c.github.io/vc-data-model/#dfn-url) or an object containing an id [property](https://w3c.github.io/vc-data-model/#dfn-property).

But unless I've misunderstood how JSON-LD works, `"some:url"` is exactly equivalent to `{"@id": "some:url"}`, and the VC context makes `"id"` a shorthand for `"@id"`.

msporny:
Your interpretation is correct. I'm struggling to see where you see the misalignment with JSON-LD?

Maybe you were expecting us to say more about what the value of the `id` property should be?

jyasskin:
No, I think it's that I'm not expecting it to say the property MUST be ~"X or something-equivalent-to-X". An "or" in a spec makes me think there will be two different things there. In the terminology from https://www.w3.org/TR/json-ld11/#data-model-overview, I think you're requiring that the value has to be a [resource](https://www.w3.org/TR/rdf11-concepts/#dfn-resource) that is not a [blank node](https://www.w3.org/TR/rdf11-concepts/#dfn-blank-node). There might be a shorter way to say that, or you could introduce a local "non-blank resource" term to cover it.

Going down to the terminology from https://www.w3.org/TR/rdf11-concepts/#section-rdf-graph, I think you're saying that the value has to be an [IRI](https://www.w3.org/TR/rdf11-concepts/#dfn-iri) rather than a [literal](https://www.w3.org/TR/rdf11-concepts/#dfn-literal) or a [blank node](https://www.w3.org/TR/rdf11-concepts/#dfn-blank-node) ... but that terminology is very bad, because it says both that ["An IRI ... is a Unicode string"](https://www.w3.org/TR/rdf11-concepts/#dfn-iri) and that ["http://example.org/ as a string literal is [not] equal to http://example.org/ as an IRI"](https://www.w3.org/TR/rdf11-concepts/#dfn-rdf-term), which if we substitute in the definitions says "http://example.org/ as a string literal is [not] equal to http://example.org/ as a Unicode string".

msporny:
[...]

PR #1338, opened 2023-11-08
Add verification algorithm to specification.

msporny:
This PR attempts to address issue #1337 by adding a verification algorithm to the specification. The PR attempts to do this in a way that defers to the securing specifications on verifying the securing mechanism and defers to the rest of the normative statements in the specification (instead of re-stating them in the algorithm).

/cc @jyasskin (since he raised this issue as a major item of concern in his review)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1338.html" title="Last updated on Dec 7, 2023, 11:18 PM UTC (6041145)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1338/fcf867c...6041145.html" title="Last updated on Dec 7, 2023, 11:18 PM UTC (6041145)">Diff</a>

PR #1349, opened 2023-11-12
Update normative language related to issuer value.

msporny:
This PR attempts to address issue #1342 by clarifying the normative language related to the issuer value. 

/cc @jyasskin

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1349.html" title="Last updated on Nov 25, 2023, 11:44 PM UTC (ccb65a7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1349/998a3dc...ccb65a7.html" title="Last updated on Nov 25, 2023, 11:44 PM UTC (ccb65a7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1727,12 +1727,13 @@ A <a>verifiable credential</a> MUST have an <code>issuer</code> <a>property</a>.
         <dl>
           <dt><var id="defn-issuer">issuer</var></dt>
           <dd>
-The value of the <code>issuer</code> <a>property</a> MUST be either a
-<a>URL</a> or an object containing an <code>id</code> <a>property</a>. It is
-RECOMMENDED that the <a>URL</a> in the <code>issuer</code> or its
-<code>id</code> be one which, if dereferenced, results in a document containing
-machine-readable information about the <a>issuer</a> that can be used to
-<a>verify</a> the information expressed in the <a>credential</a>.
+The value of the <code>issuer</code> <a>property</a> MUST be a
+<a>URL</a>, or an object containing an <code>id</code> <a>property</a> whose
+value is a <a>URL</a>, which identifies the issuer in a globally unambiguous
+way. It is RECOMMENDED that the <a>URL</a> be one which, if dereferenced,
+results in a document containing machine-readable information about the
+<a>issuer</a> that can be used to <a>verify</a> the information expressed in the
+<a>credential</a>.
           </dd>
         </dl>
 
--- a/index.html
+++ b/index.html
@@ -1729,8 +1729,8 @@ A <a>verifiable credential</a> MUST have an <code>issuer</code> <a>property</a>.
           <dd>
 The value of the <code>issuer</code> <a>property</a> MUST either be a
 <a>URL</a>, or be an object containing an <code>id</code> <a>property</a>
-whose value is a <a>URL</a>; in either case, the <a>URL</a> MUST identify
-the issuer in a globally unambiguous
+whose value is a <a>URL</a>; in either case, the issuer MUST select this
+<a>URL</a> to identify the issuer in a globally unambiguous
 way. It is RECOMMENDED that the <a>URL</a> be one which, if dereferenced,
 results in a document containing machine-readable information about the
 <a>issuer</a> that can be used to <a>verify</a> the information expressed in the
--- a/index.html
+++ b/index.html
@@ -1727,9 +1727,10 @@ A <a>verifiable credential</a> MUST have an <code>issuer</code> <a>property</a>.
         <dl>
           <dt><var id="defn-issuer">issuer</var></dt>
           <dd>
-The value of the <code>issuer</code> <a>property</a> MUST be a
-<a>URL</a>, or an object containing an <code>id</code> <a>property</a> whose
-value is a <a>URL</a>, which identifies the issuer in a globally unambiguous
+The value of the <code>issuer</code> <a>property</a> MUST either be a
+<a>URL</a>, or be an object containing an <code>id</code> <a>property</a>
+whose value is a <a>URL</a>; in either case, the <a>URL</a> MUST identify
+the issuer in a globally unambiguous
 way. It is RECOMMENDED that the <a>URL</a> be one which, if dereferenced,
 results in a document containing machine-readable information about the
 <a>issuer</a> that can be used to <a>verify</a> the information expressed in the
--- a/index.html
+++ b/index.html
@@ -1729,8 +1729,8 @@ A <a>verifiable credential</a> MUST have an <code>issuer</code> <a>property</a>.
           <dd>
 The value of the <code>issuer</code> <a>property</a> MUST either be a
 <a>URL</a>, or be an object containing an <code>id</code> <a>property</a>
-whose value is a <a>URL</a>; in either case, the issuer MUST select this
-<a>URL</a> to identify the issuer in a globally unambiguous
+whose value is a <a>URL</a>; in either case, the issuer selects this
+<a>URL</a> to identify itself in a globally unambiguous
 way. It is RECOMMENDED that the <a>URL</a> be one which, if dereferenced,
 results in a document containing machine-readable information about the
 <a>issuer</a> that can be used to <a>verify</a> the information expressed in the


iherman (approved):

seabass-labrax (approved):
Looks good to me!

longpd (approved):
This removes some potential ambiguity. Looks good.

index.html
```diff
@@ -1727,12 +1727,13 @@ <h3>Issuer</h3>
...
           <dd>
-The value of the <code>issuer</code> <a>property</a> MUST be either a
-<a>URL</a> or an object containing an <code>id</code> <a>property</a>. It is
-RECOMMENDED that the <a>URL</a> in the <code>issuer</code> or its
-<code>id</code> be one which, if dereferenced, results in a document containing
-machine-readable information about the <a>issuer</a> that can be used to
-<a>verify</a> the information expressed in the <a>credential</a>.
+The value of the <code>issuer</code> <a>property</a> MUST be a
+<a>URL</a>, or an object containing an <code>id</code> <a>property</a> whose
+value is a <a>URL</a>, which identifies the issuer in a globally unambiguous
```
OR13, line 1732: and we are asserting that DIDs are URLs here right?

OR13 (approved):

index.html
msporny: Well, the DID spec asserts that a DID URL is a URL. This text just says "any URL will do"... so, yes, a DID is a valid value here (as is any HTTP URL).

dlongley (approved):

index.html
TallTed, line 1732:
```suggestion
The value of the <code>issuer</code> <a>property</a> MUST either be a
<a>URL</a>, or be an object containing an <code>id</code> <a>property</a>
whose value is a <a>URL</a>; in either case, the <a>URL</a> MUST identify
the issuer in a globally unambiguous
```

TallTed (changes requested):
Intended to be an editorial tweak for clarification...

index.html
```diff
@@ -1727,12 +1727,14 @@ <h3>Issuer</h3>
...
           <dt><var id="defn-issuer">issuer</var></dt>
           <dd>
-The value of the <code>issuer</code> <a>property</a> MUST be either a
-<a>URL</a> or an object containing an <code>id</code> <a>property</a>. It is
-RECOMMENDED that the <a>URL</a> in the <code>issuer</code> or its
-<code>id</code> be one which, if dereferenced, results in a document containing
-machine-readable information about the <a>issuer</a> that can be used to
-<a>verify</a> the information expressed in the <a>credential</a>.
+The value of the <code>issuer</code> <a>property</a> MUST either be a
+<a>URL</a>, or be an object containing an <code>id</code> <a>property</a>
```
jyasskin, line 1731: So if you encode the issuer as `{"@id": "the:url"}`, that's not a valid VC? Or if your `@context` establishes another shorthand for `@id`, it's not conforming to use that? How would a pure JSON-LD processor detect those issues: do they need to check the incoming representation before expanding it to RDF?

index.html
```diff
@@ -1727,12 +1727,14 @@ <h3>Issuer</h3>
...
+whose value is a <a>URL</a>; in either case, the <a>URL</a> MUST identify
+the issuer in a globally unambiguous
+way. It is RECOMMENDED that the <a>URL</a> be one which, if dereferenced,
```
jyasskin, line 1734: How should the verifier check that the `id` is globally unambiguous during the [verification algorithm](https://github.com/w3c/vc-data-model/pull/1338#discussion_r1388509458)?

decentralgabe (approved):

index.html
msporny: VCs are a subset of JSON-LD documents. Every VC is a valid JSON-LD document, but not every JSON-LD document is a valid VC.

> So if you encode the issuer as {"@id": "the:url"}, that's not a valid VC? 

Correct, it is not, because [the Verifiable Credentials specification requires that JSON-LD compact document form is used](https://w3c.github.io/vc-data-model/#json-ld).

> Or if your @context establishes another shorthand for @id, it's not conforming to use that?

No, it wouldn't be a valid VC because it wouldn't follow the data model provided in the specification. It would, however, be a valid JSON-LD document and the semantics wouldn't be any different from a valid VC.

> How would a pure JSON-LD processor detect those issues: do they need to check the incoming representation before expanding it to RDF?

Yeah, most every implementation that we know of uses JSON Schema, or hard-coded checks that are roughly equivalent to JSON Schema, to check the incoming form for the VCs they expect to process.

index.html
msporny: Hrm, you're right, it can't. I was trying to make that "MUST" equate to "you have to be able to uniquely identify the issuer". What if we replace it w/ this:

```suggestion
whose value is a <a>URL</a>; in either case, the <a>URL</a> is used to identify
the issuer using a globally unambiguous identifier.
It is RECOMMENDED that the <a>URL</a> be one which, if dereferenced,
```

index.html
jyasskin: Ok, thanks, and sorry for assuming a different model for these constraints than you actually had. I'll update my understanding.

index.html
jyasskin: Or:

```suggestion
whose value is a <a>URL</a>; in either case, the <a>URL</a> is assumed to
be a globally unambiguous identifier for the the issuer.
It is RECOMMENDED that the <a>URL</a> be one which, if dereferenced,
```

Or you could keep the tighter constraint on just the issuer, like:

```suggestion
whose value is a <a>URL</a>; in either case, the issuer MUST select this
<a>URL</a> to identify the issuer in a globally unambiguous
way. It is RECOMMENDED that the <a>URL</a> be one which, if dereferenced,
```

jyasskin (approved):

index.html
msporny: No apologies necessary... it's important to understand how folks are interpreting the mental model for VCs differently. One of the things we're finding is that people that go beyond "I'll just copy/paste an example" (which is the mental model that made schema.org so successful) into actually trying to reconcile the subset of JSON-LD that we're using here tend to get tripped up in areas such as this.

In general, the approach we're taking is to optimize for the average developer that's going to be copy/pasting examples into their code/systems. We're certainly not there yet wrt. ecosytem maturity. Developers still have to think too much when using VCs today, but we're hoping that a schema.org-like approach in different market verticals will help here.

We have some of the constraints in the VC spec that we do because we want people to be able to process these objects just like they would process any other JSON object (after they've verified the digital signature). IOW, in the general case, we don't want to force them to understand JSON-LD intricacies or go to RDF to start using the data. Many of the implementations today only break out the JSON-LD processors and go to RDF when they verify the digital signature -- once that's done, they just work w/ the VC as they would most other JSON input into their systems.

index.html
msporny: I like the stronger, more focused, constraint you proposed on the issuer. That has been merged, thank you!

index.html
```diff
@@ -1727,12 +1727,14 @@ <h3>Issuer</h3>
...
-<a>URL</a> or an object containing an <code>id</code> <a>property</a>. It is
-RECOMMENDED that the <a>URL</a> in the <code>issuer</code> or its
-<code>id</code> be one which, if dereferenced, results in a document containing
-machine-readable information about the <a>issuer</a> that can be used to
-<a>verify</a> the information expressed in the <a>credential</a>.
+The value of the <code>issuer</code> <a>property</a> MUST either be a
+<a>URL</a>, or be an object containing an <code>id</code> <a>property</a>
+whose value is a <a>URL</a>; in either case, the issuer MUST select this
+<a>URL</a> to identify the issuer in a globally unambiguous
+way. It is RECOMMENDED that the <a>URL</a> be one which, if dereferenced,
```
msporny, line 1734:
```suggestion
whose value is a <a>URL</a>; in either case, the issuer selects this
<a>URL</a> to identify itself in a globally unambiguous
way. It is RECOMMENDED that the <a>URL</a> be one which, if dereferenced,
```

Hrm... I can see the testing team complaining that the "MUST" statement here is not testable (and we're trying to avoid non-testable MUST statements). Would a rewording like the above be ok with you @jyasskin? I believe it is still aligned w/ what you wanted to say, but in a way that doesn't create a MUST statement we have to defend against the "... and how are you testing this MUST statement?" conversation w/ the VCWG and implementers.

Wind4Greg (approved):

index.html
msporny: I'm applying this @jyasskin and merging. Please let me know if you disagree, we can do the change you want in a future PR.

msporny:

Issue #1352, opened 2023-11-15
Add mechanism to embed externally secured VCs in a VP


brentzundel:
Based on the [VCWG discussion this past week](https://github.com/w3c/vc-data-model/issues/1265#issuecomment-1812938709), it was decided that issue #1265 would be split into two issues. This issue attempts to propose a mechanism to express externally-secured VCs in a VP.

dlongley:
I propose a new property: `envelopedVerifiableCredential` (or `verifiableCredentialEnvelope`, whichever others prefer) to add to Verifiable Presentations. The value of this property must be a data URL expressing an externally-secured (enveloped) Verifiable Credential. For consumers, when the securing envelope (such as a JWS, etc.) is removed, the resulting payload must be a conforming document.

msporny:
IIUC, this is what these sorts of presentations could look like:

```json
{
  "@context": [
    "https://www.w3.org/ns/credentials/v2",
    "https://www.w3.org/ns/credentials/examples/v2"
  ],
  "id": "urn:uuid:3978344f-8596-4c3a-a978-8fcaba3903c5",
  "type": "VerifiablePresentation",
  "envelopedVerifiableCredential": [
    "data:application/jwt;base64,QzVjVWYzMDBOb3haW...EFrdTNkYkhEMjZvV1drVmZRMjUK==",
    "data:application/cwt;base64,ZmlOWW81U2poUjhQc...FFEbTBnb3NaZEF6NFdhWkRpYzMK=",
    "data:application/sd-jwt;base64,TmxFRW1Td0I2QWxHZ...WFrTkFZY0FzVWEzM0ptaUh4amUK===",
    "data:application/acdc;base64,TlhFeDZjSTd6YTk4eUtXRk...FnR3RLNFVLRzh0Q0luYTIK",
    "data:application/gordian;base64,ZDZNM2ZkUnNacjJXSn...FnQnFnbHdobkdzZ0tFTEhjRkkK"
   ]
}
```

Any securing specification can say how an `envelopedVerifiableCredential` for any media type should be processed. It must specify 1) how the securing mechanism on the envelope is verified and 2) how the secured content type and payload is extracted (for example, the secured content exists in the `payload` field and the content type exists in the `cty` field and MUST be `application/vc+ld+json` (or provide instructions on how to transform to that content type).

msporny:
I'll prep a PR for consideration by the WG on the approach provided above.

msporny:
[...]

PR #1358, opened 2023-11-19
Add mechanism to embed enveloped VCs in VPs.

msporny:
This PR attempts to address issue #1352 by adding a new feature for embedding "enveloped" Verifiable Credentials in order to support JWT/CWT/Gordian/ACDC-secured VCs in presentations. It also renames the "external proof" securing mechanism class to "enveloping proof" and the "embedded proof" securing mechanism class to "watermarking proof" in order to convey the concept using better analogies.

/cc @m00sey @christophera @SmithSamuelM

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1358.html" title="Last updated on Nov 20, 2023, 2:26 PM UTC (3ef9512)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1358/cb3696e...3ef9512.html" title="Last updated on Nov 20, 2023, 2:26 PM UTC (3ef9512)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2022,11 +2022,11 @@ using it.
         </p>
         <p>
 This specification recognizes two classes of securing mechanisms: those that use
-external proofs and those that use embedded proofs. An
-<dfn class="export">external proof</dfn> is one that wraps an expression of
+enveloping proofs and those that use watermarking proofs. An
+<dfn class="export">enveloping proof</dfn> is one that wraps an expression of
 this data model, such as via a JSON Web Token, which is elaborated on in the
 <em>Securing Verifiable Credentials using JOSE and COSE</em> [[VC-JOSE-COSE]] specification.
-An <dfn class="export">embedded proof</dfn> is a mechanism where the proof is
+An <dfn class="export">watermarking proof</dfn> is a mechanism where the proof is
 included in the data model, such as a Data Integrity Proof, which is elaborated
 on in <em>Verifiable Credential Data Integrity</em> [[VC-DATA-INTEGRITY]].
         </p>
@@ -2045,7 +2045,7 @@ the <code>proof</code> <a>property</a>.
         </p>
         <p>
 Methods of securing <a>verifiable credentials</a> or <a>verifiable
-presentations</a> that use an external proof MAY use the <code>proof</code>
+presentations</a> that use an enveloping proof MAY use the <code>proof</code>
 <a>property</a>.
         </p>
         <dl>
@@ -2056,7 +2056,7 @@ One or more cryptographic proofs that can be used to detect tampering and verify
 the authorship of a <a>verifiable credential</a> or a <a>verifiable
 presentation</a>. Each proof is a separate <a>named graph</a>
 (referred to as a <dfn class="export">proof graph</dfn>) containing a single
-proof. The specific method used for an <a>embedded proof</a> MUST be identified
+proof. The specific method used for a <a>watermarking proof</a> MUST be identified
 using the <code>type</code> <a>property</a>.
             </p>
             <p>
@@ -2245,6 +2245,13 @@ verifiable credential graphs</a> in a cryptographically <a>verifiable</a> format
 See Section <a href="#verifiable-credential-graphs"></a> for further details on
 this topic.
           </dd>
+          <dt><var id="defn-envelopedVerifiableCredential">envelopedVerifiableCredential</var></dt>
+          <dd>
+The <code>envelopedVerifiableCredential</code> <a>property</a> MAY be present.
+The value MUST be an array of one or more <a>URLs</a> using the `data:` URL
+scheme where each value contains a <a>verifiable credential</a> that
+is secured using an <a>enveloping proof</a>.
+          </dd>
           <dt><var id="defn-holder">holder</var></dt>
           <dd>
 The <a>verifiable presentation</a> MAY include a <code>holder</code>
@@ -2278,7 +2285,8 @@ to not pertain to the <a>validation</a> of the <a>verifiable presentation</a>.
         </dl>
 
         <p>
-The example below shows a <a>verifiable presentation</a>:
+The example below shows a <a>verifiable presentation</a> that uses
+<a>watermarking proofs</a>:
         </p>
 
         <pre class="example nohighlight" title="Basic structure of a presentation">
@@ -2303,6 +2311,26 @@ An example of a <a>verifiable presentation</a> using the JWT proof mechanism is
 provided in the Securing Verifiable Credentials using JOSE and COSE
 [[?VC-JOSE-COSE]] specification.
         </p>
+
+        <p>
+The example below shows a <a>verifiable presentation</a> that contains
+<a>verifiable credentials</a> that are protected using <a>enveloping proofs</a>:
+        </p>
+
+        <pre class="example nohighlight" title="A presentation that contains verifiable credentials secured using enveloping proofs">
+{
+  "@context": [
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
+  ],
+  "id": "urn:uuid:5ec137ea-871e-11ee-a783-ef96a4397a9a",
+  "type": ["VerifiablePresentation", "ExamplePresentation"],
+  "envelopedVerifiableCredential": [
+    "data:application/jwt;base64,QzVjV...RMjUK==",
+    "data:application/cwt;base64,ZmlOW...pYzMK="
+ ]
+}
+        </pre>
 
         <section>
           <h4>Presentations Using Derived Credentials</h4>
@@ -2381,7 +2409,7 @@ self-asserted <a>verifiable credential</a> that is secured using the same
 mechanism as the <a>verifiable presentation</a>.
           </p>
 
-          <pre class="example nohighlight" title="A verifiable presentation, secured with an embedded proof, with a self-asserted verifiable credential">
+          <pre class="example nohighlight" title="A verifiable presentation, secured with a watermarking proof, with a self-asserted verifiable credential">
 {
   "@context": [
     "https://www.w3.org/ns/credentials/v2",
@@ -2408,7 +2436,7 @@ self-asserted <a>verifiable credential</a> that holds <a>claims</a> about the
 <a>verifiable presentation</a>.
           </p>
 
-          <pre class="example nohighlight" title="A verifiable presentation, secured with an embedded proof, with a self-asserted verifiable credential about the verifiable presentation">
+          <pre class="example nohighlight" title="A verifiable presentation, secured with a watermarking proof, with a self-asserted verifiable credential about the verifiable presentation">
 {
   "@context": [
     "https://www.w3.org/ns/credentials/v2",
@@ -6528,7 +6556,7 @@ Credential implementations that justify the use of a specific media type.
         </p>
 
         <p>
-This media type can be used with credentials secured using an <a>external
+This media type can be used with credentials secured using an <a>enveloping
 proof</a>.
         </p>
         <p>
@@ -6590,7 +6618,7 @@ Credential implementations that justify the use of a specific media type.
 
         <p>
 This media type can be used with presentations secured using an
-<a>external proof</a>.
+<a>enveloping proof</a>.
         </p>
         <p>
 A [[JSON-LD]] context is expected to be present in the body of the document, and
--- a/index.html
+++ b/index.html
@@ -2026,7 +2026,7 @@ enveloping proofs and those that use watermarking proofs. An
 <dfn class="export">enveloping proof</dfn> is one that wraps an expression of
 this data model, such as via a JSON Web Token, which is elaborated on in the
 <em>Securing Verifiable Credentials using JOSE and COSE</em> [[VC-JOSE-COSE]] specification.
-An <dfn class="export">watermarking proof</dfn> is a mechanism where the proof is
+A <dfn class="export">watermarking proof</dfn> is a mechanism where the proof is
 included in the data model, such as a Data Integrity Proof, which is elaborated
 on in <em>Verifiable Credential Data Integrity</em> [[VC-DATA-INTEGRITY]].
         </p>


index.html
```diff
@@ -2022,11 +2022,11 @@ <h3>Securing Verifiable Credentials</h3>
...
         <p>
 This specification recognizes two classes of securing mechanisms: those that use
-external proofs and those that use embedded proofs. An
-<dfn class="export">external proof</dfn> is one that wraps an expression of
+enveloping proofs and those that use watermarking proofs. An
+<dfn class="export">enveloping proof</dfn> is one that wraps an expression of
 this data model, such as via a JSON Web Token, which is elaborated on in the
 <em>Securing Verifiable Credentials using JOSE and COSE</em> [[VC-JOSE-COSE]] specification.
-An <dfn class="export">embedded proof</dfn> is a mechanism where the proof is
+An <dfn class="export">watermarking proof</dfn> is a mechanism where the proof is
```
dlongley, line 2029:
```suggestion
A <dfn class="export">watermarking proof</dfn> is a mechanism where the proof is
```

dlongley (approved):

TallTed (commented):
Seems OK overall. There's some bikeshedding in the back of my head about `watermarking proofs` vs `watermark proofs` and `watermarks`, and `enveloping proofs` vs `envelope proofs` and `envelopes` or maybe `wrapper proofs` and `wrappers`.

msporny:
> Seems OK overall. There's some bikeshedding in the back of my head about `watermarking proofs` vs `watermark proofs` and `watermarks`, and `enveloping proofs` vs `envelope proofs` and `envelopes` or maybe `wrapper proofs` and `wrappers`.

Yes, I went back and forth on those alternatives as well. I've seen "enveloping" thrown around in various communities, and have heard others in the VCWG start to use that terminology, so wanted to try that terminology out on this PR. Happy for others to jump in w/ bikeshedding suggestions. We're not changing the concepts, just trying to find a better way to convey them.

The real world analogy of an "envelope" and a "watermark" seem to be the closest so far. The JOSE, COSE, ACDC, and Gordian approaches put the data to be secured in an envelope (like putting a letter in an envelope that is sealed w/ an official wax seal). 

Data Integrity watermarks the data by adding something to the data (like how holograms and UV ink watermarks are used to secure the authenticity of a physical passport or driver's license).

Those analogies feel like they pass the "Explain the concept in a sentence or two to the general public." test.

longpd:
> > Seems OK overall. There's some bikeshedding in the back of my head about `watermarking proofs` vs `watermark proofs` and `watermarks`, and `enveloping proofs` vs `envelope proofs` and `envelopes` or maybe `wrapper proofs` and `wrappers`.
> 
> Yes, I went back and forth on those alternatives as well. I've seen "enveloping" thrown around in various communities, and have heard others in the VCWG start to use that terminology, so wanted to try that terminology out on this PR. Happy for others to jump in w/ bikeshedding suggestions. We're not changing the concepts, just trying to find a better way to convey them.
> 
> The real world analogy of an "envelope" and a "watermark" seem to be the closest so far. The JOSE, COSE, ACDC, and Gordian approaches put the data to be secured in an envelope (like putting a letter in an envelope that is sealed w/ an official wax seal).
> 
> Data Integrity watermarks the data by adding something to the data (like how holograms and UV ink watermarks are used to secure the authenticity of a physical passport or driver's license).
> 
> Those analogies feel like they pass the "Explain the concept in a sentence or two to the general public." test.

I'm not a fan of enveloping as a descriptor as it also means obscures or shrouds.  I'd prefer "proof envelope" but I'm not going object if this is the preference of the group.

msporny:
@longpd wrote:
> I'm not a fan of enveloping as a descriptor as it also means obscures or shrouds.

Well, that's exactly what the enveloping proof formats do. :)

They take something that looks like this:

```
{
  "test": "Hello World!"
}
```

and turn it into something that looks like this (the envelope):

```
eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ0ZXN0IjoiSGVsbG8gV29ybGQhIn0.sHNFdvKnbR-PNe6P1lDDSjWPVmmgqZN-5krJvdaTUOI
```

So the data is obscured/shrouded, as you say ^

decentralgabe (approved):
nice to have this!

andresuribe87:
This PR made it clear to me that we need to provide more clarity to readers on what is and what is not a [verifiable credential graph](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verifiable-credential-graph). Implementers should be able to easily and quickly answer the question "Is X a verifiable credential graph?". 

I'd like to start by providing some examples, and what I believe the answer is for each. 
* [1] IS NOT a verifiable credential graph. Why not? Because it is not a verifiable credential, since, according to the spec, a verifiable credential "MUST be protected using a securing mechanism" ([source](https://www.w3.org/TR/vc-data-model-2.0/#dfn-conforming-document:~:text=MUST%20be%20protected%20using%20a%20securing%20mechanism)). This example is not. Note that this example is part of the spec [here](https://www.w3.org/TR/vc-data-model-2.0/#example-usage-of-the-id-property), and we're incorrectly labeling it as a `Verifiable Credential`.
* [2] IS NOT a verifiable credential graph. Why not? Well.... this is a bit murky. I think the answer is that, while it is a verifiable credential, it is not a graph (or apologies if I missed the normative definition of a graph).
* [3] IS a verifiable credential credential graph. Why? It's a [conformant document](https://www.w3.org/TR/vc-data-model-2.0/#dfn-conforming-document), it's secured, and it's a graph. Admittedly, I don't know how to answer the question of why [2] is a graph.

If we're in agreement with the above, I'd like to suggest that:
* We provide guidance to what is a `verifiable credential graph`, presumably by having a normative definition on what is a `graph`.
* We rename the `verifiablePresentation.verifiableCredentials` property to `verifiablePresentation.verifiableCredentialGraphs`.

I'd be happy to raise both points above as separate issued. It seemed to me that this PR was the correct place, as it is building on top of those missing definitions. Or perhaps this was already discussed elsewhere and I missed it. 

[1]
```json
{
  "@context": [
  ],
  "id": "http://university.example/credentials/3732",
  "type": ["VerifiableCredential", "ExampleDegreeCredential"],
  "validFrom": "2010-01-01T00:00:00Z",
  "credentialSubject": {
    "degree": {
    }
  }
}
```

[2]
```jwt
eyJhbGciOiJFUzI1NiIsInR5cCI6IkpXVCJ9.eyJ2YyI6eyJAY29udGV4dCI6WyJodHRwczovL3
d3dy53My5vcmcvbnMvY3JlZGVudGlhbHMvdjIiLCJodHRwczovL3d3dy53My5vcmcvbnMvY3JlZ
GVudGlhbHMvZXhhbXBsZXMvdjIiXSwiaWQiOiJodHRwOi8vdW5pdmVyc2l0eS5leGFtcGxlL2Ny
ZWRlbnRpYWxzLzM3MzIiLCJ0eXBlIjpbIlZlcmlmaWFibGVDcmVkZW50aWFsIiwiRXhhbXBsZUR
lZ3JlZUNyZWRlbnRpYWwiXSwiaXNzdWVyIjoiaHR0cHM6Ly91bml2ZXJzaXR5LmV4YW1wbGUvaX
NzdWVycy81NjUwNDkiLCJ2YWxpZEZyb20iOiIyMDEwLTAxLTAxVDAwOjAwOjAwWiIsImNyZWRlb
nRpYWxTdWJqZWN0Ijp7ImlkIjoiZGlkOmV4YW1wbGU6ZWJmZWIxZjcxMmViYzZmMWMyNzZlMTJl
YzIxIiwiZGVncmVlIjp7InR5cGUiOiJFeGFtcGxlQmFjaGVsb3JEZWdyZWUiLCJuYW1lIjoiQmF
jaGVsb3Igb2YgU2NpZW5jZSBhbmQgQXJ0cyJ9fX0sImlzcyI6Imh0dHBzOi8vdW5pdmVyc2l0eS
5leGFtcGxlL2lzc3VlcnMvNTY1MDQ5IiwianRpIjoiaHR0cDovL3VuaXZlcnNpdHkuZXhhbXBsZ
S9jcmVkZW50aWFscy8zNzMyIiwic3ViIjoiZGlkOmV4YW1wbGU6ZWJmZWIxZjcxMmViYzZmMWMy
NzZlMTJlYzIxIn0.9sRuyNnImQQZnJAb00TGEzLmMhXXHmDvVvf7qwzb34hCxANEwRt2eBhL3fC
LzblZM2r9yb37_2jprJywVU_RgQ
```

[3]
```json
{
  "@context": [
    "https://w3id.org/security/suites/ed25519-2020/v1"
  ],
  "id": "http://university.example/credentials/3732",
  "type": [
    "VerifiableCredential",
    "ExampleDegreeCredential"
  ],
  "validFrom": "2010-01-01T00:00:00Z",
  "credentialSubject": {
    "degree": {
    }
  },
  "proof": {
    "type": "Ed25519Signature2020",
    "created": "2023-11-26T23:04:13Z",
    "verificationMethod": "https://university.example/issuers/565049#key-1"
,
    "proofValue": "zbZuLBB1qkoK3KShFs1USRrhXn6bTmQ2tQhANtYYNM5zntoygX5Q3zJUKZaQpWXzRPd4Cd8iYxq1PobiNXk97SUW"
  }
}
```

iherman:
> There's some bikeshedding in the back of my head about `watermarking proofs` vs `watermark proofs` and `watermarks`, and `enveloping proofs` vs `envelope proofs` and `envelopes` or maybe `wrapper proofs` and `wrappers`.

I do not have a strong opinion about the 'envelope' analogy, which may indeed be better than the 'external' one. But I am not really sure what the problem is with 'embedded'. It worked for me at least...

iherman:
@andresuribe87,

reflecting on https://github.com/w3c/vc-data-model/pull/1358#issuecomment-1828340329, forgive me if I am very pedantic. A credential (verifiable or not) is a set of statements which is _always_ part of (or is) a [graph](https://www.w3.org/TR/vc-data-model-2.0/#dfn-graphs). (That is true to any RDF structure, not specific to credentials.) That graph may be explicitly "created" (or designated) as a [named graph](https://www.w3.org/TR/vc-data-model-2.0/#dfn-named-graphs), or it is the so-called [default graph](https://www.w3.org/TR/vc-data-model-2.0/#dfn-default-graph). These terms are actually defined in the [RDF specification](https://www.w3.org/TR/rdf11-concepts/#section-dataset), but were taken over into the terminology of the VCDM by leaving out some aspects that VCDM does not use.

With that, some of my comments:

> [1] IS NOT a verifiable credential graph. Why not? Because it is not a verifiable credential, since, according to the spec, a verifiable credential "MUST be protected using a securing mechanism" (source). This example is not. Note that this example is part of the spec here, and we're incorrectly labeling it as a Verifiable Credential.

I think that we had this discussion before. The way it was discussed was whether we need two separate types, one for `Credential` and the other as `VerifiableCredential`. (It came up when cleaning up the formal vocabulary.) We finally decided that, though not precise, we'll live with the ambiguity of using credentials and verifiable credentials interchangeably; the required changes in the spec were considered to be too extensive.

 @msporny will remember the details of the discussion. But I do not think we should reopen this issue.

> [2] IS NOT a verifiable credential graph. Why not? Well.... this is a bit murky. I think the answer is that, while it is a verifiable credential, it is not a graph (or apologies if I missed the normative definition of a graph).

As I said above, that is not correct. The question is whether it is a separate named graph or is part of the single (default) graph. Provided that [2] is an encoded version of something like [1], it is a default graph. 

(Named graphs appear in the spec with verifiable presentations.)

> [3] IS a verifiable credential credential graph. Why? It's a conformant document, it's secured, and it's a graph. Admittedly, I don't know how to answer the question of why [2] is a graph.

Per our specification, [3] is actually a collection of two graphs. The credential part, which is the default graph, and the proof graph, which (per definition of the `proof` property) is a separate, named graph. See also [the new figure 6](https://www.w3.org/TR/vc-data-model-2.0/#info-graph-vc), with an example very similar to [3].

> If we're in agreement with the above, I'd like to suggest that:
>
> • [...] presumably by having a normative definition on what is a graph.

The normative definition of graphs is in the spec (admittedly, this is a recent change on the document).

That said, your comments do raise a very important question that is relevant to this PR. The `verifiableCredential` property is defined to specify a (new) _named graph_ that would contain a single credential. The details are in [§4.9](https://www.w3.org/TR/vc-data-model-2.0/#verifiable-credential-graphs). ***I believe the same should be true for `envelopedVerifiableCredential`:*** its definition should clarify that and should also refer to §4.9 (and §4.9 should also be modified referring to the new property).

cc @dlongley @msporny

> • We rename the verifiablePresentation.verifiableCredentials property to verifiablePresentation.verifiableCredentialGraphs.

In an ideal world, I would agree with you. Yes, the term `verifiableCredential` is misleading insofar as the graph naming aspect is hidden (and this misconception actually may have led to a bug in v1.0). The problem I see is that renaming a term that is already widely used in implementations out there might be the source of problems. I leave the judgement to others, but, pragmatically, we may have to live with the current term.

andresuribe87:
> Forgive me if I am very pedantic reflecting on [#1358 (comment)](https://github.com/w3c/vc-data-model/pull/1358#issuecomment-1828340329).

I appreciate the extra detail and thoughtful response @iherman, thanks!

>  A credential (verifiable or not) is a set of statements which is _always_ part of (or is) a [graph](https://www.w3.org/TR/vc-data-model-2.0/#dfn-graphs). (That is true to any RDF structure, not specific to credentials.) That graph may be explicitly "created" (or designated) as a [named graph](https://www.w3.org/TR/vc-data-model-2.0/#dfn-named-graphs), or it is the so-called [default graph](https://www.w3.org/TR/vc-data-model-2.0/#dfn-default-graph). These terms are actually defined in the [RDF specification](https://www.w3.org/TR/rdf11-concepts/#section-dataset), but were taken over into the terminology of the VCDM by leaving out some aspects that VCDM does not use.
> 
> With that, some of my comments:
> 
> > [1] IS NOT a verifiable credential graph. Why not? Because it is not a verifiable credential, since, according to the spec, a verifiable credential "MUST be protected using a securing mechanism" (source). This example is not. Note that this example is part of the spec here, and we're incorrectly labeling it as a Verifiable Credential.
> 
> I think that we had this discussion before. The way it was discussed was whether we need two separate types, one for `Credential` and the other as `VerifiableCredential`. (It came up when cleaning up the formal vocabulary.) We finally decided that, though not precise, we'll live with the ambiguity of using credentials and verifiable credentials interchangeably; the required changes in the spec were considered to be too extensive.
> 
> @msporny will remember the details of the discussion. But I do not think we should reopen this issue.
> 

Are you saying that [1] IS a verifiable credential graph? If so, would you say that [1] IS a verifiable credential? I consider it very important to align on examples, because they leave less space for ambiguity. I'm trying to use the terms "credential" and "verifiable credential" in the way that's not ambiguous, with the distinction that we've made in the spec under [section 3](https://www.w3.org/TR/vc-data-model-2.0/#core-data-model).    

> > [2] IS NOT a verifiable credential graph. Why not? Well.... this is a bit murky. I think the answer is that, while it is a verifiable credential, it is not a graph (or apologies if I missed the normative definition of a graph).
> 
> As I said above, that is not correct. The question is whether it is a separate named graph or is part of the single (default) graph. Provided that [2] is an encoded version of something like [1], it is a default graph.
> 
> (Named graphs appear in the spec with verifiable presentations.)
> 

I think I follow your argument, but if that's the case, then why aren't we able to have [2] be an item of a `verifiablePresentation.verifiableCredential`?  Could you point to which step is incorrect in the below inference steps? I've included the decoded payload of [2] below for clarity, which is referenced as [4].

1. [2] is a verifiable credential because it: a) encodes [4], which is a conforming document, and b) is secured using vc-jose-cose.
2. A [graph](https://www.w3.org/TR/vc-data-model-2.0/#dfn-graphs) is a "A set of claims, forming a network of information composed of [subjects](https://www.w3.org/TR/vc-data-model-2.0/#dfn-subjects) and their relationship to other [subjects](https://www.w3.org/TR/vc-data-model-2.0/#dfn-subjects) or data". [4] is a set of claims, and thus is a graph.
3. [2] is a verifiable credential (from step 1) and is a graph (from step 2), thus [2] is a verifiable credential graph.
4. Related to `[verifiablePresentation](https://www.w3.org/TR/vc-data-model-2.0/#presentations-0).verifiableCredential`: "The value MUST be an array of one or more [verifiable credential graphs](https://www.w3.org/TR/vc-data-model-2.0/#verifiable-credential-graphs) in a cryptographically [verifiable](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verify) format. See Section [4.9 Verifiable Credential Graphs](https://www.w3.org/TR/vc-data-model-2.0/#verifiable-credential-graphs) for further details on this topic.". We've established in step 3 that [2] is a verifiable credential graph. Thus we can include [2] as an item of the array.

> > [3] IS a verifiable credential credential graph. Why? It's a conformant document, it's secured, and it's a graph. Admittedly, I don't know how to answer the question of why [2] is a graph.
> 
> Per our specification, [3] is actually a collection of two graphs. The credential part, which is the default graph, and the proof graph, which (per definition of the `proof` property) is a separate, named graph. See also [the new figure 6](https://www.w3.org/TR/vc-data-model-2.0/#info-graph-vc), with an example very similar to [3].
> 
> > If we're in agreement with the above, I'd like to suggest that:
> > • [...] presumably by having a normative definition on what is a graph.
> 
> The normative definition of graphs is in the spec (admittedly, this is a recent change on the document).
> 
> That said, your comments do raise a very important question that is relevant to this PR. The `verifiableCredential` property is defined to specify a (new) _named graph_ that would contain a single credential. The details are in [§4.9](https://www.w3.org/TR/vc-data-model-2.0/#verifiable-credential-graphs). _**I believe the same should be true for `envelopedVerifiableCredential`:**_ its definition should clarify that and should also refer to §4.9 (and §4.9 should also be modified referring to the new property).

I've read [section 9](https://www.w3.org/TR/vc-data-model-2.0/#verifiable-credential-graphs) now multiple times, and I'm still not able to define what a "verifiable credential graph" is. That section is introducing a concept, but it's not defining it. Could you explain further what it is? Happy to open a separate issue to elaborate on this further.

> 
> > • We rename the verifiablePresentation.verifiableCredentials property to verifiablePresentation.verifiableCredentialGraphs.
> 
> In an ideal world, I would agree with you. Yes, the term `verifiableCredential` is misleading insofar as the graph naming aspect is hidden (and this misconception actually may have led to a bug in v1.0). The problem I see is that renaming a term that is already widely used in implementations out there might be the source of problems. I leave the judgement to others, but, pragmatically, we may have to live with the current term.

I would argue that if renaming fixes these sort of bugs, that's exactly what we should do in 2.0 :)

[4]
```json
{
  "vc": {
    "@context": [
    ],
    "id": "http://university.example/credentials/3732",
    "type": [
      "VerifiableCredential",
      "ExampleDegreeCredential"
    ],
    "validFrom": "2010-01-01T00:00:00Z",
    "credentialSubject": {
      "degree": {
      }
    }
  },
  "iss": "https://university.example/issuers/565049",
  "jti": "http://university.example/credentials/3732",
  "sub": "did:example:ebfeb1f712ebc6f1c276e12ec21"
}
```

iherman:
> > 
> > > [1] IS NOT a verifiable credential graph. Why not? Because it is not a verifiable credential, since, according to the spec, a verifiable credential "MUST > be protected using a securing mechanism" (source). This example is not. Note that this example is part of the spec here, and we're incorrectly labeling it as > a Verifiable Credential.
> > 
> > I think that we had this discussion before. The way it was discussed was whether we need two separate types, one for `Credential` and the other as > `VerifiableCredential`. (It came up when cleaning up the formal vocabulary.) We finally decided that, though not precise, we'll live with the ambiguity of > using credentials and verifiable credentials interchangeably; the required changes in the spec were considered to be too extensive.
> 
> Are you saying that [1] IS a verifiable credential graph? If so, would you say that [1] IS a verifiable credential? I consider it very important to align on examples, because they leave less space for ambiguity. I'm trying to use the terms "credential" and "verifiable credential" in the way that's not ambiguous, with the distinction that we've made in the spec under [section 3](https://www.w3.org/TR/vc-data-model-2.0/#core-data-model). 

My comment was on the question whether it is a ***graph or not a graph***. And I do say that yes, it is a graph.

The question you seem to ask is whether it is a ***verifiable credential*** or just a plain ***credential***. What I said was that this is a different issue: do we want to make a very clear distinction between these two terms. I.e., should we say that it is a *verifiable* credential if and only if there is either a `proof` statement or it is part of an external mechanism like JWT. This is a different issue that should not be discussed in this PR (and was discussed before).

[…]

> 
> I think I follow your argument, but if that's the case, then why aren't we able to have [2] be an item of a `verifiablePresentation.verifiableCredential`?  Could you point to which step is incorrect in the below inference steps? I've included the decoded payload of [2] below for clarity, which is referenced as [4].

> 
> 1. [2] is a verifiable credential because it: a) encodes [4], which is a conforming document, and b) is secured using vc-jose-cose.
> 2. A [graph](https://www.w3.org/TR/vc-data-model-2.0/#dfn-graphs) is a "A set of claims, forming a network of information composed of [subjects](https://www.w3.org/TR/vc-data-model-2.0/#dfn-subjects) and their relationship to other [subjects](https://www.w3.org/TR/vc-data-model-2.0/#dfn-subjects) or data". [4] is a set of claims, and thus is a graph.
> 3. [2] is a verifiable credential (from step 1) and is a graph (from step 2), thus [2] is a verifiable credential graph.
> 4. Related to [verifiablePresentation](https://www.w3.org/TR/vc-data-model-2.0/#presentations-0).`verifiableCredential`: "The value MUST be an array of one or more [verifiable credential graphs](https://www.w3.org/TR/vc-data-model-2.0/#verifiable-credential-graphs) in a cryptographically [verifiable](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verify) format. See Section [4.9 Verifiable Credential Graphs](https://www.w3.org/TR/vc-data-model-2.0/#verifiable-credential-graphs) for further details on this topic.". We've established in step 3 that [2] is a verifiable credential graph. Thus we can include [2] as an item of the array.

If my understanding is correct, your question is why we need the separate property introduced in this PR, i.e., why isn't it enough to use the `verifiablePresentation.verifiableCredential` property? If so, the problem is with the word `encodes`. From a purely RDF point of view, that statement does not make sense: if there is a URL (let that be a data URL) or a literal (whether it encodes something or not), that resource is opaque for RDF. 

That extra action must be specified separately if this is really the intention. Technically, the semantics of the new property would have to say explicitly: "you decode the content of the URL and you create a set of RDF statements that you consider as such...".

Alternatively, the semantics should clearly say that the literal is to be taken as an opaque entity for the purpose of, say, the signature of a VP (by "opaque", I mean that the signature should simply sign, among other things, the _encoded_ form).

And you are right: this may have to be defined with more care than it is now. I will raise a separate comment as part of my formal review.

[…]

> 
> I've read [section 9](https://www.w3.org/TR/vc-data-model-2.0/#verifiable-credential-graphs) now multiple times, and I'm still not able to define what a "verifiable credential graph" is. That section is introducing a concept, but it's not defining it. Could you explain further what it is? Happy to open a separate issue to elaborate on this further.

The formal definition is really in the RDF spec, so I am forced to go back to the RDF world. Hopefully, in this case, this helps. And my apologies for the verbosity of my answer.

The reason we are talking about _named_ graphs in RDF is because the concept is, basically, to have a graph with its own, distinct identity, i.e., a _name_ which, in the RDF world, is a URL. To use a somewhat different syntax, one could have something like:

```
<https://example.org/name1> {
    <https://> type  <Verifiable Credential>;
               issuer <https://university.example/issuers/565049>;
               ...
}

<https://example.org/name2> {
    <https://> type  <Verifiable Credential>;
               issuer <https://university.example/issuers/01234>;
               ...
}
```

We have here two, distinct credentials, i.e., credential graphs. If I use the URL `https://example.org/name1` I know I refer to the terms in the first block and not in the second. I.e., if I have to create a signature for `name1`, then I know which terms I have to consider: those that are in the first graph and not in the second. That "separation" is what is described in [section 9](https://www.w3.org/TR/vc-data-model-2.0/#verifiable-credential-graphs).

What muddles the waters a bit is that, as it is the case for RDF in general, I can also use a blank node instead of a clear URL, which is something like saying "we know there is a distinct identifier but we do not care what it is". (It is an "existential term" in logic). So I could have: 

```
_:1234 {
    <https://> type  <Verifiable Credential>;
               issuer <https://university.example/issuers/565049>;
               ...
}
```

where `_:...` is the syntax used in most of the RDF serializations for blank nodes. The choice of `_:1234` is arbitrary: an RDF implementation can choose any term there, just be careful that distinct graphs should have distinct bnode identifiers. (This does create practical problems when merging two different graph sets: the bnodes have to be renamed to ensure unicity.)

The way the JSON-LD `@context` (for VC-s) is defined is to instruct the JSON-LD processor to generate something like this for the `verifiableCredential` term:

```
<id for the VP> verifiableCredential _:1234 .
_:1234 {
     <https://> type  <Verifiable Credential>;
               issuer <https://university.example/issuers/565049>;
               ...
}
```

I.e., only blank nodes are used. If there are more than one credentials, then each gets its own blank node, ie, they are all mutually separated.

Does this help a bit?

And yes, I believe any further clarification should be subject of a separate issue...

> > In an ideal world, I would agree with you. Yes, the term `verifiableCredential` is misleading insofar as the graph naming aspect is hidden (and this misconception actually may have led to a bug in v1.0). The problem I see is that renaming a term that is already widely used in implementations out there might be the source of problems. I leave the judgement to others, but, pragmatically, we may have to live with the current term.
> 
> I would argue that if renaming fixes these sort of bugs, that's exactly what we should do in 2.0 :)

As I said: in theory yes, pragmatically probably not. But that is a really separate issue, too, I believe.

index.html
```diff
@@ -2244,6 +2244,13 @@ <h3>Presentations</h3>
 verifiable credential graphs</a> in a cryptographically <a>verifiable</a> format.
 See Section <a href="#verifiable-credential-graphs"></a> for further details on
 this topic.
+          </dd>
+          <dt><var id="defn-envelopedVerifiableCredential">envelopedVerifiableCredential</var></dt>
+          <dd>
+The <code>envelopedVerifiableCredential</code> <a>property</a> MAY be present.
+The value MUST be an array of one or more <a>URLs</a> using the `data:` URL
+scheme where each value contains a <a>verifiable credential</a> that
+is secured using an <a>enveloping proof</a>.
```
iherman, line 2253: As the [discussion](https://github.com/w3c/vc-data-model/pull/1358#issuecomment-1828340329) with @andresuribe87 shows: this term may not be properly defined. The question is: what is exactly the semantics and the behavior v.a.v. the encoded verifiable credentials. This may become a question if the enclosing VP has to be signed, for example. There are two alternatives that I may see:

1. The idea is that the data URL is "dereferenced" which, in this case, would mean that data part is decoded, and the resulting VC is out into its own VC graph (just like what happens in the case of the `verifiableCredential` term).
2. The data URL content is, essentially, opaque; the signature of the VP happens with that term taken verbatim, i.e., as a string. Everything else, i.e., what to do with that piece of credential, falls into the application domain.

I think that to properly specify (1) is problematic, and requires quite a lot of additional specification text. That approach would not really go well with the Linked Data aspect: it looks like having a "plugin" processor put into the JSON-LD processing.

Alternative (2) is clean (although with a caveat, see below), but I am not sure whether it is what the expectation is with that property.

As for the caveat: the usage of a URL, i.e., a data URL, is a bit of a question, mainly in the case of alternative (2). See my separate [comment on the relevant issue](https://github.com/w3c/vc-data-model/issues/1352#issuecomment-1827948258)

iherman (changes requested):

andresuribe87:
> > > > [1] IS NOT a verifiable credential graph. Why not? Because it is not a verifiable credential, since, according to the spec, a verifiable credential "MUST > be protected using a securing mechanism" (source). This example is not. Note that this example is part of the spec here, and we're incorrectly labeling it as > a Verifiable Credential.
> > > 
> > > 
> > > I think that we had this discussion before. The way it was discussed was whether we need two separate types, one for `Credential` and the other as > `VerifiableCredential`. (It came up when cleaning up the formal vocabulary.) We finally decided that, though not precise, we'll live with the ambiguity of > using credentials and verifiable credentials interchangeably; the required changes in the spec were considered to be too extensive.
> > 
> > 
> > Are you saying that [1] IS a verifiable credential graph? If so, would you say that [1] IS a verifiable credential? I consider it very important to align on examples, because they leave less space for ambiguity. I'm trying to use the terms "credential" and "verifiable credential" in the way that's not ambiguous, with the distinction that we've made in the spec under [section 3](https://www.w3.org/TR/vc-data-model-2.0/#core-data-model).
> 
> My comment was on the question whether it is a _**graph or not a graph**_. And I do say that yes, it is a graph.
> 
> The question you seem to ask is whether it is a _**verifiable credential**_ or just a plain _**credential**_. What I said was that this is a different issue: do we want to make a very clear distinction between these two terms. I.e., should we say that it is a _verifiable_ credential if and only if there is either a `proof` statement or it is part of an external mechanism like JWT. This is a different issue that should not be discussed in this PR (and was discussed before).
> 
> […]
> 

I see what you mean now, agreed.

> > I think I follow your argument, but if that's the case, then why aren't we able to have [2] be an item of a `verifiablePresentation.verifiableCredential`?  Could you point to which step is incorrect in the below inference steps? I've included the decoded payload of [2] below for clarity, which is referenced as [4].
> 
> > 1. [2] is a verifiable credential because it: a) encodes [4], which is a conforming document, and b) is secured using vc-jose-cose.
> > 2. A [graph](https://www.w3.org/TR/vc-data-model-2.0/#dfn-graphs) is a "A set of claims, forming a network of information composed of [subjects](https://www.w3.org/TR/vc-data-model-2.0/#dfn-subjects) and their relationship to other [subjects](https://www.w3.org/TR/vc-data-model-2.0/#dfn-subjects) or data". [4] is a set of claims, and thus is a graph.
> > 3. [2] is a verifiable credential (from step 1) and is a graph (from step 2), thus [2] is a verifiable credential graph.
> > 4. Related to [verifiablePresentation](https://www.w3.org/TR/vc-data-model-2.0/#presentations-0).`verifiableCredential`: "The value MUST be an array of one or more [verifiable credential graphs](https://www.w3.org/TR/vc-data-model-2.0/#verifiable-credential-graphs) in a cryptographically [verifiable](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verify) format. See Section [4.9 Verifiable Credential Graphs](https://www.w3.org/TR/vc-data-model-2.0/#verifiable-credential-graphs) for further details on this topic.". We've established in step 3 that [2] is a verifiable credential graph. Thus we can include [2] as an item of the array.
> 
> If my understanding is correct, your question is why we need the separate property introduced in this PR, i.e., why isn't it enough to use the `verifiablePresentation.verifiableCredential` property? If so, the problem is with the word `encodes`. From a purely RDF point of view, that statement does not make sense: if there is a URL (let that be a data URL) or a literal (whether it encodes something or not), that resource is opaque for RDF.
> 
> That extra action must be specified separately if this is really the intention. Technically, the semantics of the new property would have to say explicitly: "you decode the content of the URL and you create a set of RDF statements that you consider as such...".
> 
> Alternatively, the semantics should clearly say that the literal is to be taken as an opaque entity for the purpose of, say, the signature of a VP (by "opaque", I mean that the signature should simply sign, among other things, the _encoded_ form).
> 
> And you are right: this may have to be defined with more care than it is now. I will raise a separate comment as part of my formal review.
> 
> […]
> 

Perfect, that all makes a ton of sense. Looking forward to clarifying that piece in the spec.

> > I've read [section 9](https://www.w3.org/TR/vc-data-model-2.0/#verifiable-credential-graphs) now multiple times, and I'm still not able to define what a "verifiable credential graph" is. That section is introducing a concept, but it's not defining it. Could you explain further what it is? Happy to open a separate issue to elaborate on this further.
> 
> The formal definition is really in the RDF spec, so I am forced to go back to the RDF world. Hopefully, in this case, this helps. And my apologies for the verbosity of my answer.
> 
> The reason we are talking about _named_ graphs in RDF is because the concept is, basically, to have a graph with its own, distinct identity, i.e., a _name_ which, in the RDF world, is a URL. To use a somewhat different syntax, one could have something like:
> 
> ```
> <https://example.org/name1> {
>     <https://> type  <Verifiable Credential>;
>                issuer <https://university.example/issuers/565049>;
>                ...
> }
> 
> <https://example.org/name2> {
>     <https://> type  <Verifiable Credential>;
>                issuer <https://university.example/issuers/01234>;
>                ...
> }
> ```
> 
> We have here two, distinct credentials, i.e., credential graphs. If I use the URL `https://example.org/name1` I know I refer to the terms in the first block and not in the second. I.e., if I have to create a signature for `name1`, then I know which terms I have to consider: those that are in the first graph and not in the second. That "separation" is what is described in [section 9](https://www.w3.org/TR/vc-data-model-2.0/#verifiable-credential-graphs).
> 
> What muddles the waters a bit is that, as it is the case for RDF in general, I can also use a blank node instead of a clear URL, which is something like saying "we know there is a distinct identifier but we do not care what it is". (It is an "existential term" in logic). So I could have:
> 
> ```
> _:1234 {
>     <https://> type  <Verifiable Credential>;
>                issuer <https://university.example/issuers/565049>;
>                ...
> }
> ```
> 
> where `_:...` is the syntax used in most of the RDF serializations for blank nodes. The choice of `_:1234` is arbitrary: an RDF implementation can choose any term there, just be careful that distinct graphs should have distinct bnode identifiers. (This does create practical problems when merging two different graph sets: the bnodes have to be renamed to ensure unicity.)
> 
> The way the JSON-LD `@context` (for VC-s) is defined is to instruct the JSON-LD processor to generate something like this for the `verifiableCredential` term:
> 
> ```
> <id for the VP> verifiableCredential _:1234 .
> _:1234 {
>      <https://> type  <Verifiable Credential>;
>                issuer <https://university.example/issuers/565049>;
>                ...
> }
> ```
> 
> I.e., only blank nodes are used. If there are more than one credentials, then each gets its own blank node, ie, they are all mutually separated.
> 
> Does this help a bit?
> 
> And yes, I believe any further clarification should be subject of a separate issue...
> 

This makes a ton of sense, thanks for taking the time to explain. I've raised https://github.com/w3c/vc-data-model/issues/1365

> > > In an ideal world, I would agree with you. Yes, the term `verifiableCredential` is misleading insofar as the graph naming aspect is hidden (and this misconception actually may have led to a bug in v1.0). The problem I see is that renaming a term that is already widely used in implementations out there might be the source of problems. I leave the judgement to others, but, pragmatically, we may have to live with the current term.
> > 
> > 
> > I would argue that if renaming fixes these sort of bugs, that's exactly what we should do in 2.0 :)
> 
> As I said: in theory yes, pragmatically probably not. But that is a really separate issue, too, I believe.

Raised https://github.com/w3c/vc-data-model/issues/1366

index.html
msporny: > Alternative (2) is clean (although with a caveat, see below), but I am not sure whether it is what the expectation is with that property.

Agreed, and I believe it's aligned w/ the expectation. 

IMHO, that /is/ a separate graph of information (or rather, should be treated as one) and we can say that in the specification. IOW, it's not presumed that one would take that signed object and merge it into the default graph. There is also an expectation that it contains a conforming document that is secured using the rules for the media type.

iherman: > > Alternative (2) is clean (although with a caveat, see below), but I am not sure whether it is what the expectation is with that property.
> 
> Agreed, and I believe it's aligned w/ the expectation.
> 

👍

> IMHO, that /is/ a separate graph of information (or rather, should be treated as one) and we can say that in the specification. 

We have to be ***very*** cautious then how we formulate that. From a Linked Data perspective and, consequently, from the terminology used in the specification, it is ***not*** a separate graph, it is "just" a more-or-less opaque string. The application level, could (should? must?) do additional checks on whether it can be decoded into a graph and whether that graph is conforming to our specification. But that is on a different layer.

Very specifically. If I sign, using RCH or via JWS, a VP that uses this term, that signature applies ***on the encoded string, i.e., the value of the property*** and not on the decoded graph. This must be clearly specified in the spec, otherwise we run into problems imho.

> IOW, it's not presumed that one would take that signed object and merge it into the default graph. There is also an expectation that it contains a conforming document that is secured using the rules for the media type.

Ok. I think we are aligned. Now comes the simple task of formalizing that succinctly in the document 😀.

iherman: @msporny, you did not react on the exact specification of the term's range, ie, whether we define a new datatype or whether leave it as a URL...

iherman:
> @iherman and @andresuribe87 -- it's not clear to me what changes you want to be made to this PR. I have read your discussion, and have philosophical thoughts/input, but none of which I'm confident of applying to this PR.

_I_ am not requiring any change. The question is whether the fact of @andresuribe87 not understanding things can be explained by some ambiguities in the text, or whether it can be avoided by some improvements. @andresuribe87, only you can tell us that...

index.html
msporny: > Now comes the simple task of formalizing that succinctly in the document

The above gives me enough to make an attempt at a PR modification, thank you.

> specification of the term's range, ie, whether we define a new datatype or whether leave it as a URL...

Let's leave it as a URL for now, because it is a URL. You are signing over an opaque URL and it's up to the application to interpret that URL, and we'll provide guidance on exactly how applications should interpret that URL (evaluate the URL, it should result in a separate graph that contains a VC that is secured using some recognized securing mechanism). I'll try to be less hand-wavy in the PR contents.

iherman: Ok, let us try to go the simple way and keep it as a URL... Let us see where this goes.

index.html
```diff
@@ -2022,11 +2022,11 @@ <h3>Securing Verifiable Credentials</h3>
         </p>
         <p>
 This specification recognizes two classes of securing mechanisms: those that use
-external proofs and those that use embedded proofs. An
-<dfn class="export">external proof</dfn> is one that wraps an expression of
+enveloping proofs and those that use watermarking proofs. An
```
OR13, line 2025: watermarking typically implies some difficulty in removing the mark... this is not true of embedded proofs.

I think the term embedded proofs was better here...

index.html
```diff
@@ -2304,6 +2312,26 @@ <h3>Presentations</h3>
...
+{
+  "@context": [
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
+  ],
+  "id": "urn:uuid:5ec137ea-871e-11ee-a783-ef96a4397a9a",
+  "type": ["VerifiablePresentation", "ExamplePresentation"],
+  "envelopedVerifiableCredential": [
+    "data:application/jwt;base64,QzVjV...RMjUK==",
+    "data:application/cwt;base64,ZmlOW...pYzMK="
```
OR13, line 2330: A CWT can never be a"verifiable credential"... CWTs secure a CBOR map as a payload.

Verifiable Credentials secure a Compact JSON-LD Document as a payload.

index.html
```diff
@@ -2304,6 +2312,26 @@ <h3>Presentations</h3>
...
+
+        <pre class="example nohighlight" title="A presentation that contains verifiable credentials secured using enveloping proofs">
+{
+  "@context": [
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
+  ],
+  "id": "urn:uuid:5ec137ea-871e-11ee-a783-ef96a4397a9a",
+  "type": ["VerifiablePresentation", "ExamplePresentation"],
+  "envelopedVerifiableCredential": [
```
OR13, line 2328: I take it this is needed because dynamic RDF Prcessors struggle with this:

```
{
  "@context": [
  ],
  "id": "urn:uuid:5ec137ea-871e-11ee-a783-ef96a4397a9a",
  "verifiableCredential": [
    "data:application/jwt;base64,QzVjV...RMjUK==",
    "data:application/cwt;base64,ZmlOW...pYzMK="
 ]
}
```

```
<urn:uuid:5ec137ea-871e-11ee-a783-ef96a4397a9a> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://www.w3.org/2018/credentials#VerifiablePresentation> .
<urn:uuid:5ec137ea-871e-11ee-a783-ef96a4397a9a> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://www.w3.org/ns/credentials/examples#ExamplePresentation> .
<urn:uuid:5ec137ea-871e-11ee-a783-ef96a4397a9a> <https://www.w3.org/2018/credentials#verifiableCredential> "data:application/cwt;base64,ZmlOW...pYzMK=" .
<urn:uuid:5ec137ea-871e-11ee-a783-ef96a4397a9a> <https://www.w3.org/2018/credentials#verifiableCredential> "data:application/jwt;base64,QzVjV...RMjUK==" .
```

I don't think these changes are helpful.

OR13 (changes requested):
I don't think these changes are necessary, or improve the quality of the document.

They also conflate the securing representations with the core data model.

We don't need a new property to process the examples as RDF, as is shown in the nquads I shared.

index.html
iherman: I do not understand the remark "dynamic RDF Processors struggle with this". Actually, I do not even understand what you mean by "dynamic RDF Processor"...

From a pure RDF point of view, there is no struggle. A data URL is a URL, no difference between that one an any other URL.

iherman:
@OR13,

On the [call of Nov. 15 ](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-11-15-vcwg#section3-3) the issue of adding a property to VP to represent externally defined proofs came up. This PR is a proposal to handle that issue. With that:

> I don't think these changes are necessary, or improve the quality of the document.

Do you mean that, after all, such extra property is not necessary, or that the proposed approach is not the right one? If the former, make this statement clearly, and we can, if there is an agreement, close this issue and move on. If the latter, what is your alternative.

> 
> They also conflate the securing representations with the core data model.

This goes back to what I asked above. Do you mean that we do not need that feature, or that this feature would be harmful?

> 
> We don't need a new property to process the examples as RDF, as is shown in the nquads I shared.

I do not understand that remark.

 I am not taking side on the issue, but I am not an VC application developer, hence I cannot judge whether that extra feature is necessary or not.

index.html
iherman: ***If*** what you mean is that one could use an absolute URL (e.g., a data URL) as a value for the `verifiableCredential` property, hence there is no need for a new property, that by itself is a correct statement. But I do not think it is what is meant by the original discussion, and it is actually related to the discussion I had with @andresuribe87.

The value of the `verifiableCredential` is _not_ a URL for a credential. It is the URL for a named graph that contains a credential. I believe the original issue in the WG was to have a property that refers (directly) to an externally secured credential (e.g., JOSE-COSE). These two are not the same.

Looking at your examples

```
<urn:uuid:5ec137ea-871e-11ee-a783-ef96a4397a9a> <https://www.w3.org/2018/credentials#verifiableCredential> "data:application/cwt;base64,ZmlOW...pYzMK=" .
```
is actually wrong. The proper representation in nquads would be:

```
<urn:uuid:5ec137ea-871e-11ee-a783-ef96a4397a9a> <https://www.w3.org/2018/credentials#verifiableCredential> _:b1 .
<data:application/cwt;base64,ZmlOW...pYzMK=> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://www.w3.org/2018/credentials#VerifiableCredential> _:b1 .
```

It is questionable whether this is what we want, semantically. The usage of a separate property side-steps this issue, which sounds o.k. to me _if this is what the WG wants_.

See also my separate discussion on the feature [a few days ago](https://github.com/w3c/vc-data-model/pull/1358#discussion_r1409060339) and in what follows.

See my [comment below](https://github.com/w3c/vc-data-model/pull/1358#issuecomment-1840161412): we have to decide whether that feature is really necessary or not.

BigBlueHat: @OR13's example doesn't use the approach proposed here and pulls in the examples context with it's "catch-all" vocab...so it doesn't accurately represent the changes proposed here, but rather shows what's broken without these changes.

The context changes that would be made once #1372 is merged would look more like the following--which uses the term proposed here defined as a URL/IRI as expressed in #1372:
```json
{
  "@context": [
    {"envelopedVerifiableCredential": {"@type": "@id"}}
  ],
  "id": "urn:uuid:5ec137ea-871e-11ee-a783-ef96a4397a9a",
  "envelopedVerifiableCredential": [
    "data:application/jwt;base64,QzVjV...RMjUK==",
    "data:application/cwt;base64,ZmlOW...pYzMK="
 ]
}
```

Resulting quads look like--which reads correctly to me based on this PR and the vocab changes proposed in #1372:
```n3
<urn:uuid:5ec137ea-871e-11ee-a783-ef96a4397a9a> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://www.w3.org/2018/credentials#VerifiablePresentation> .
<urn:uuid:5ec137ea-871e-11ee-a783-ef96a4397a9a> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://www.w3.org/ns/credentials/issuer-dependent#ExamplePresentation> .
<urn:uuid:5ec137ea-871e-11ee-a783-ef96a4397a9a> <https://www.w3.org/ns/credentials/issuer-dependent#envelopedVerifiableCredential> <data:application/cwt;base64,ZmlOW...pYzMK=> .
<urn:uuid:5ec137ea-871e-11ee-a783-ef96a4397a9a> <https://www.w3.org/ns/credentials/issuer-dependent#envelopedVerifiableCredential> <data:application/jwt;base64,QzVjV...RMjUK==> .
```

BigBlueHat:
> I don't think these changes are necessary, or improve the quality of the document.
> 
> They also conflate the securing representations with the core data model.
> 
> We don't need a new property to process the examples as RDF, as is shown in the nquads I shared.

@OR13, I've updated your example to reflect what's proposed here. Hopefully that makes these improvements more obvious.

andresuribe87:
@msporny I am not asking for specific changes in this PR. My concerns are being discussed in https://github.com/w3c/vc-data-model/issues/1365

iherman:
Not being at the call, my reaction on the [discussion on the 5th of December](https://github.com/w3c/vc-data-model/pull/1358#issuecomment-1842457719).

My understanding of the minutes is to do something like this:

```json
"verifiableCredential" : [
   {
      … this is a 'traditional' credential, i.e., with an embedded proof
   },
   {
     "id": "data:...."
     "type": "EnvelopedVerifiableCredential"
   }
]
```

Did we consider the fact that `verifiableCredential` generates a separate graph? Does this create any additional complication? If nothing else, it is a bit of a sledgehammer for those data URL-s...

Also, can I put a non-data URL in the "id" field? Per JSON or JSON-LD, I can. What does that mean in terms of proofs? (Per default, the signature through DI only signs the URL string, nothing else.)

Maybe all this works, but I just wanted to raise a flag...

iherman:
Just putting this out there (and this is only an opinion and not a formal objection): unless things fall into place quickly and easily, I wonder whether we should not close this PR and declare the issue as deferred. After all, this is a new feature request that popped up a few weeks ago only; per our own agreements, we have frozen the spec and accepting a new feature request was not supposed to happen in the first place...

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-12-06-vcwg#section3-1) on 2023-12-06

- no resolutions were taken

#### 3.1. Add mechanism to embed externally secured VCs in a VP (issue vc-data-model#1352)

_See github issue [vc-data-model#1352](https://github.com/w3c/vc-data-model/issues/1352)._




**Brent Zundel:** This issue is being addressed by the PR we discussed yesterday.  
… I believe that yesterday we decided that the verifiableCredential property should be expanded to allow each object in the array to express its type: whether it is a normal VC or an enveloped VC (actual name yet to be decided).  

---

</details>

msporny:
This PR is marked as pending close. I don't think any variation of it will be acceptable to @OR13.

We have an alternate proposal that I will try to do a PR for related to [a discussion during the WG call](https://github.com/w3c/vc-data-model/pull/1358#issuecomment-1842457719).

If that fails, we will close issue #1352 and note that the WG couldn't come to consensus on embedding externally secured documents in a VP (even though multiple workable proposals were put forward).

brentzundel:
Closing, superseded by #1379

PR #1367, opened 2023-11-30
Update context to match spec for name/description.

BigBlueHat:
This removes the `@value`, `@language`, and `@direction` aliases from
`name` and `description` to match the work done in
https://github.com/w3c/vc-data-model/pull/1340


OR13 (approved):

iherman (approved):

longpd (approved):

msporny (approved):

brentzundel (approved):

decentralgabe (approved):

contexts/credentials/v2
```diff
@@ -211,19 +185,8 @@
...
-
-        "id": "@id",
-        "type": "@type",
-
-        "description": "https://schema.org/description",
-        "name": "https://schema.org/name"
-      }
-    },
+    "StatusList2021Credential":
+      "https://w3id.org/vc/status-list#StatusList2021Credential",
```
OR13, line 189: It would be nice if W3C hosted these files... I don't like including references to w3id.org, without some understanding that W3C will buyout / host these terms forever.

BigBlueHat: @OR13 https://w3id.org/ is run by a W3C CG for pre-publication identifiers (among other things):
> The purpose of this website is to provide a secure, permanent [URL](http://en.wikipedia.org/wiki/URL) re-direction service for Web applications. This service is run by the [W3C Permanent Identifier Community Group](http://www.w3.org/community/perma-id/).

iherman: @OR13, while I would have preferred myself to use direct W3C URL-s for this vocabulary as well, this issue has been discussed many times earlier this year (including at the time when the CR was approved). The main issue is and was the backward incompatibility with deployed software that would be created if this URL changed, and the consensus was to keep w3id.org. I think we should not reopen a discussion again.

brentzundel:
Open for a week, multiple positive reviews, no changes requested, merging.

Issue #1360, opened 2023-11-23
Add mechanism to cryptographically secure non-credential VP properties (contexts etc)


dmitrizagidulin:
(Extracting from issue #1265, as requested on the previous call.)
Verifiable Presentations need a way to cryptographically secure externally linked resources (such as `@context`s) that is similar to the VC `relatedProperties`/`digestSRI` mechanism.

dmitrizagidulin:
My proposal - we add the `relatedProperties` field to the VP data model. (That is, extend the domain of the `relatedProperties` field to be _either_ VC or VP.)

OR13:
this is not an issue for vc-jose-cose, additional claims can be present, and if not understood they are ignored.

in the context of rdf, they get assigned "issuer dependent terms".

OR13:
@dmitrizagidulin that functionality is still supported by what I said... and it still needs to be interpreted consistently by issuers and verifiers to be useful.

dmitrizagidulin:
@OR13 I agree that schema-wise, it's not an issue for jose/cose serializations. However, we should still highlight its use in VPs, _so that_ verifiers interpret it consistently

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-11-28-vcwg-special#section1-2) on 2023-11-28

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 1.2. Add mechanism to cryptographically secure non-credential VP properties (contexts etc) (issue vc-data-model#1360)

_See github issue [vc-data-model#1360](https://github.com/w3c/vc-data-model/issues/1360)._




**Dmitri Zagidulin:** We want to be able to lock these down cryptographically.  
… We have only added this for verifiable credentials.  
… But not for related resources.  
… We need to do this for VPs.  

**Brent Zundel:** Can the solution be as simple as saying "this section also refers to VPs".  

**Manu Sporny:** Do you mean "related resources"?  
… I can volunteer to write a PR for this.  

> *Joe Andrieu:* correct. id we need to secure contexts we should do that. arbitrary resources is a layer violation.

> *Phillip Long:* pdl-asu has joined #vcwg-special.

> *Joe Andrieu:* happy to review pr.

</details>

msporny:
PR #1370 has been raised to address this issue. This issue will be closed once issue #1370 has been merged.

msporny:
PR #1370 has failed to gain consensus.

[...]

PR #1370, opened 2023-12-03
Add `relatedResource` property to `VerifiablePresentation`

msporny:
This PR attempts to address issue #1360 by allowing the `relatedResource` property to be used on a `VerifiablePresentation`. It also adds an "at risk" issue marker noting that the feature might be removed in the Candidate Recommendation phase based on implementer feedback.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1370.html" title="Last updated on Dec 4, 2023, 4:23 PM UTC (de83559)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1370/62a667c...de83559.html" title="Last updated on Dec 4, 2023, 4:23 PM UTC (de83559)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3080,7 +3080,7 @@ The requirement that contexts be listed in `relatedResource` is currently being
 debated in the VCWG. This requirement might be removed in future iterations of
 the specification.
         </p>
-        <p class="issue at-risk" title="Use of `relatedResource` in `VerifiablePresentation`">
+        <p class="issue atrisk" title="Use of `relatedResource` in `VerifiablePresentation`">
 The Working Group is seeking feedback from implementers on whether the
 `relatedResource` property is useful when used in `VerifiablePresentation`
 objects. Based on feedback, the Working Group might determine that the property
--- a/index.html
+++ b/index.html
@@ -3080,6 +3080,13 @@ The requirement that contexts be listed in `relatedResource` is currently being
 debated in the VCWG. This requirement might be removed in future iterations of
 the specification.
         </p>
+        <p class="issue at-risk" title="Use of `relatedResource` in `VerifiablePresentation`">
+The Working Group is seeking feedback from implementers on whether the
+`relatedResource` property is useful when used in `VerifiablePresentation`
+objects. Based on feedback, the Working Group might determine that the property
+is not useful and will then remove the feature during the Candidate
+Recommendation phase.
+        </p>
         <p>
 Each object in the <code>relatedResource</code> array MUST contain the
 following: the [[URL]] to the resource named <code>id</code> and the
--- a/index.html
+++ b/index.html
@@ -3054,25 +3054,26 @@ JSON-LD implementers seeking interoperability.
       <section>
         <h2>Integrity of Related Resources</h2>
         <p>
-When including a link to an external resource in a <a>verifiable credential</a>,
+When including a link to an external resource in a <a>conforming document</a>,
 it is desirable to know whether the resource that is pointed to is the same at
 signing time as it is at verification time. This applies to cases where there is
 an external resource that is remotely retrieved as well as to cases where the
-<a>issuer</a> and/or <a>verifier</a> may have local cached copies of a resource.
+<a>issuer</a> and/or <a>verifier</a> might have local cached copies of a
+resource.
         </p>
         <p>
 It is also desirable to know that the contents of the JSON-LD context(s) used in
-the <a>verifiable credential</a> are the same when used by both the
+a <a>conforming document</a> are the same when used by both the
 <a>issuer</a> and <a>verifier</a>.
         </p>
         <p>
-To validate that a resource referenced by a <a>verifiable credential</a> is the
+To validate that a resource referenced by a <a>conforming document</a> is the
 same at verification time as it is at issuing time, an implementer MAY include a
 property named <code id="defn-relatedResource">relatedResource</code> that
 stores an array of objects that describe additional integrity metadata about
-each resource referenced by the <a>verifiable credential</a>. If
+each resource referenced by the <a>conforming document</a>. If
 <code>relatedResource</code> is present, there MUST be an object in the array
-for each remote resource for each context used in the verifiable credential.
+for each remote resource used in the verifiable credential.
         </p>
         <p class="issue" title="Mandatory listing of contexts in relatedResouce are under debate.">
 The requirement that contexts be listed in `relatedResource` is currently being
@@ -3118,7 +3119,7 @@ Header.
         </ul>
 
         <p>
-Any object in the <a>verifiable credential</a> that contains an `id` [[URL]]
+Any object in a <a>conforming document</a> that contains an `id` [[URL]]
 property MAY be annotated with integrity information as specified in this
 section by inclusion of <code>digestSRI</code>
 in the object.
@@ -3140,7 +3141,7 @@ Implementers are urged to consult appropriate sources, such as the
 FIPS 180-4 Secure Hash Standard</a> and the
 <a href="https://media.defense.gov/2022/Sep/07/2003071834/-1/-1/0/CSA_CNSA_2.0_ALGORITHMS_.PDF">
 Commercial National Security Algorithm Suite 2.0</a> to ensure that they are
-chosing a current and reliable hash algorithm. At the time of this writing
+choosing a current and reliable hash algorithm. At the time of this writing
 `sha384` SHOULD be considered the minimum strength hash algorithm for use by
 implementers.
         </p>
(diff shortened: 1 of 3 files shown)


index.html
```diff
@@ -3054,30 +3054,38 @@ <h4>Semantic Interoperability</h4>
...
 <code>relatedResource</code> is present, there MUST be an object in the array
-for each remote resource for each context used in the verifiable credential.
+for each remote resource used in the verifiable credential.
         </p>
         <p class="issue" title="Mandatory listing of contexts in relatedResouce are under debate.">
 The requirement that contexts be listed in `relatedResource` is currently being
 debated in the VCWG. This requirement might be removed in future iterations of
 the specification.
+        </p>
+        <p class="issue at-risk" title="Use of `relatedResource` in `VerifiablePresentation`">
```
iherman, line 3083:
```suggestion
        <p class="issue at-risk" title="Use of `relatedResource` in `VerifiablePresentation` is at risk">
```

I think it is better to make the "at risk" features very explicit when reading the document. (I have not checked whether this is the case for the other, at risk features in the spec.)

iherman (approved):

index.html
msporny: I got the "atrisk" class wrong and fixed it in de8355957fc69fe3b390caf93f48234bddf3fde5.

Which results in this rendering, does that work for you @iherman ?

image

andresuribe87 (approved):

brentzundel (approved):

dlongley (approved):

vocab/credentials/v2/vocabulary.yml
```diff
@@ -163,7 +163,7 @@ property:
 
   - id: relatedResource
     label: Related resource
-    domain: cred:VerifiableCredential
+    domain: [cred:VerifiableCredential, cred:VerifiablePresentation]
```
OR13, line 166: Given this points to credentials by URL, which can't you use a Data URI here?

```
{
  "@context": [
  ],
  "id": "urn:uuid:5ec137ea-871e-11ee-a783-ef96a4397a9a",
  "relatedResource": [
    "data:application/jwt;base64,QzVjV...RMjUK==",
    "data:application/cwt;base64,ZmlOW...pYzMK="
 ]
}
```

index.html
```diff
@@ -3054,30 +3054,38 @@ <h4>Semantic Interoperability</h4>
...
-To validate that a resource referenced by a <a>verifiable credential</a> is the
+To validate that a resource referenced by a <a>conforming document</a> is the
 same at verification time as it is at issuing time, an implementer MAY include a
 property named <code id="defn-relatedResource">relatedResource</code> that
 stores an array of objects that describe additional integrity metadata about
-each resource referenced by the <a>verifiable credential</a>. If
+each resource referenced by the <a>conforming document</a>. If
 <code>relatedResource</code> is present, there MUST be an object in the array
-for each remote resource for each context used in the verifiable credential.
+for each remote resource used in the verifiable credential.
```
OR13, line 3076: data URIs are IRIs.

vocab/credentials/v2/vocabulary.yml
OR13: confusing when comparing to https://github.com/w3c/vc-data-model/pull/1372/files#diff-dfe00db9d510a45f71289312990ee36df870f36c02094551bfba625dcd0d8e96R193

jandrieu (changes requested):
Remove this. It isn't a good adjustment.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-12-05-vcwg-special#section1-1) on 2023-12-05

- no resolutions were taken

#### 1.1. Add `relatedResource` property to `VerifiablePresentation` (pr vc-data-model#1370)

_See github pull request [vc-data-model#1370](https://github.com/w3c/vc-data-model/pull/1370)._




**Brent Zundel:** only has approvals.  
… likely to get merged within the week.  

**Joe Andrieu:** I still think this is a layering error. I don't know why Verifier would know how to interpret related resource, they didn't ask for it, it's not a VC, I think most use cases could be handled w/ a VC or how to secure contexts for ... generic mechanism seems wrong.  

**Manu Sporny:** I largely agree with Joe on his use cases. We should probably have a context securing mechanism.  
… and have something for enveloped VCs.  
… a fear is that some implementers are going to use related resources as a dumping ground for all sorts of things.  
… I want to express XYZ, I'm going to put that in there.  
… All that to say, Joe has a point.  
… But if folks want to use the data model in that way, they can.  

**Joe Andrieu:** I want to note that anyone can add to the context and they can add properties. What we're talking about here is that everyone has to respect... we aren't preventing people that want to use relatedResource. I don't see reason for extensibility in this way.  

**Brent Zundel:** I see the benefit of having more specific mechanisms for the different things.  
… would the related resource be seen as a stepping stone?  

**Joe Andrieu:** I think a lot of it is lessons learned from extensible HTTP headers, which was "X-something"... if people start using it, they won't stop using it. If burden is you have your own context that does something specific. That's the path towards getting new things into VCDM. I'm afraid of setting the wrong direction here.  

> *Dave Longley:* +1 to Joe that once people start using things they won't stop ... there are advantages for putting slight speed bumps in the way to make people do things in better ways.

**Brent Zundel:** this PR has nothing but approvals.  

</details>

jandrieu (changes requested):
This should not go in. It's a layer violation.

msporny:
@jandrieu wrote:
> This should not go in. It's a layer violation.

You noted on the call that you were a non-blocking objection for this PR. 

However, your comment above could be seen as objecting to this PR.

Are you ok with it being merged, or are you objecting to it being merged?

jandrieu:
That was a miscommunication.

I do object to this going in.

msporny:
@jandrieu wrote:
> I do object to this going in.

Ok, marking this as pending close, then.

vocab/credentials/v2/vocabulary.yml
msporny: @decentralgabe, would you find this mechanism more acceptable? https://github.com/w3c/vc-data-model/pull/1379

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-12-12-vcwg-special#section2-1) on 2023-12-12

- no resolutions were taken

#### 2.1. Add `relatedResource` property to `VerifiablePresentation` (pr vc-data-model#1370)

_See github pull request [vc-data-model#1370](https://github.com/w3c/vc-data-model/pull/1370)._




**Brent Zundel:** First up, 1370.  
… Looks like Joe is opposed.  
… We don't have consensus on this. Looking to close.  
… happy to take comments. if folks are opposed, we can timebox a conversation.  

**Manu Sporny:** As a heads up, I curated the PRs and put pending close on any thing that has an objection without a work-around.  
… we need to get more aggressive about closing.  
… Just an explanation about why things are marked pending closed.  
… Another thing to ask is whether or not the objector would make an FO on this change.  
… My concern about this PR is that if we don't put Related Resource will not have a way of adding a hash for a JSON LD context in addition to the base context.  
… It has been asserted that you don't have to do that if you have static contexts.  
… Orie's not here, who might be objecting.  

**Joe Andrieu:** Yes, I am opposed to this, I would Formally Object, I think this is a huge layer violation. It creates an opportunity for surveillance through related resource, I think we can solve context securing in a different way.  
… Manu, noted that VC-JOSE-COSE folks want to use JSON-LD Contexts... they should just use JSON-LD.  

**Ivan Herman:** I'm neutral on the original thing. but what was just proposed to Joe seems the wrong thing to do.  
… saying that it can only be used for certain purposes sounds bad.  
… that requires additional restrictions through new properties.  
… I can't object as a W3C staff, but I don't like it.  

> *Manu Sporny:* FWIW, I agree with Ivan, I was just trying to find a way through this where we don't end up w/ FOs.

**Brent Zundel:** I'm seeing no consensus here.  
… We don't have Orie, so we can't ask him.  
… based on the information we have, we should close this PR.  
… I'd like to close it now, but we just put the pending close label on this three days ago.  

**Manu Sporny:** note, if Orie is going to formally object (because you can't secure context).  
… our way through this would be to have a "relatedContext" property.  
… I think Joe would not be opposed to that. And maybe Orie would find that ok.  

**Ivan Herman:** I want to remind people that there was a feature freeze a while ago.  
… what you just proposed feels like a new feature. It's just too late.  
… We can't work that way or we'll never finish.  

**Brent Zundel:** appreciate the reminders.  

</details>

msporny:
PR failed to achieve consensus, marked "pending close" for over a week, closing.

PR #1372, opened 2023-12-04
Add the `EnvelopedVerifiableCredential` class to the vocabulary

iherman:
This is a PR on top of PR #1358

To be merged by @msporny into that PR if acceptable...

Note: this PR does _not_ include a change on the vocabulary diagram. If and when #1358 _and_ #1370, I will do a separate PR with the necessary modification for both. Otherwise the drawio file will have ugly merge conflicts...

PR #1379, opened 2023-12-09
Add support for Enveloped Verifiable Credentials (2nd attempt)

msporny:
This PR attempts to address issue #1352 by adding a new feature for embedding "enveloped" Verifiable Credentials in Verifiable Presentations. This PR integrates @iherman's PR #1372.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1379.html" title="Last updated on Dec 15, 2023, 5:23 PM UTC (3884d4e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1379/40be9e0...3884d4e.html" title="Last updated on Dec 15, 2023, 5:23 PM UTC (3884d4e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2362,7 +2362,7 @@ enveloped <a>verifiable credential</a>:
   "type": ["VerifiablePresentation", "ExamplePresentation"],
   <span class="highlight">"verifiableCredential": [{
     "@context": "https://www.w3.org/ns/credentials/v2",
-    "id": "data:application/vc+ld+json+sd-jwt;base64,QzVjV...RMjUK==",
+    "id": "data:application/vc+ld+json+sd-jwt;QzVjV...RMjU",
     "type": "EnvelopedVerifiableCredential"
   }]</span>
 }
--- a/index.html
+++ b/index.html
@@ -2342,7 +2342,7 @@ Used to associate an object containing an enveloped <a>verifiable credential</a>
 with the `verifiableCredential` property in a <a>verifiable presentation</a>.
 The `@context` value of the object MUST follow all of the "MUST" statements in
 Section <a href="#contexts"></a>. The `id` value of the object MUST be a `data:`
-URL that expresses a secured <a>verifiable credential</a> using an enveloping
+URL [[RFC2397]] that expresses a secured <a>verifiable credential</a> using an enveloping
 security scheme, such as [[[VC-JOSE-COSE]]] [[VC-JOSE-COSE]]. The `type` value
 of the object MUST be `EnvelopedVerifiableCredential`.
             </dd>
--- a/index.html
+++ b/index.html
@@ -2325,6 +2325,52 @@ provided in the Securing Verifiable Credentials using JOSE and COSE
         </p>
 
         <section>
+          <h4>Enveloped Verifiable Credentials</h4>
+
+          <p>
+It is possible to include one or more <a>verifiable credentials</a> in a
+<a>verifiable presentation</a> that have been secured using a securing mechanism
+that "envelopes" the payload, such as [[[?VC-JOSE-COSE]]] [[?VC-JOSE-COSE]].
+This can be accomplished by associating the `verifiableCredential` property with
+an object that has a `type` of `EnvelopedVerifiableCredential`.
+          </p>
+
+          <dl>
+            <dt id="defn-EnvelopedVerifiableCredential">EnvelopedVerifiableCredential</dt>
+            <dd>
+Used to associate an object containing an enveloped <a>verifiable credential</a>
+with the `verifiableCredential` property in a <a>verifiable presentation</a>.
+The `@context` value of the object MUST follow all of the "MUST" statements in
+Section <a href="#contexts"></a>. The `id` value of the object MUST be a `data:`
+URL that expresses a secured <a>verifiable credential</a> using an enveloping
+security scheme, such as [[[VC-JOSE-COSE]]] [[VC-JOSE-COSE]]. The `type` value
+of the object MUST be `EnvelopedVerifiableCredential`.
+            </dd>
+          </dl>
+
+        <p>
+The example below shows a <a>verifiable presentation</a> that contains an
+enveloped <a>verifiable credential</a>:
+        </p>
+
+        <pre class="example nohighlight" title="Basic structure of a presentation">
+{
+  "@context": [
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
+  ],
+  "type": ["VerifiablePresentation", "ExamplePresentation"],
+  <span class="highlight">"verifiableCredential": [{
+    "@context": "https://www.w3.org/ns/credentials/v2",
+    "id": "data:application/vc+ld+json+sd-jwt;base64,QzVjV...RMjUK==",
+    "type": "EnvelopedVerifiableCredential"
+  }]</span>
+}
+        </pre>
+
+        </section>
+
+        <section>
           <h4>Presentations Using Derived Credentials</h4>
 
           <p>
--- a/index.html
+++ b/index.html
@@ -2328,8 +2328,8 @@ provided in the Securing Verifiable Credentials using JOSE and COSE
           <h4>Enveloped Verifiable Credentials</h4>
 
           <p>
-It is possible to include one or more <a>verifiable credentials</a> in a
-<a>verifiable presentation</a> that have been secured using a securing mechanism
+It is possible for a <a>verifiable presentation</a> to include one or more
+<a>verifiable credentials</a> that have been secured using a securing mechanism
 that "envelopes" the payload, such as [[[?VC-JOSE-COSE]]] [[?VC-JOSE-COSE]].
 This can be accomplished by associating the `verifiableCredential` property with
 an object that has a `type` of `EnvelopedVerifiableCredential`.
--- a/contexts/credentials/v2
+++ b/contexts/credentials/v2
@@ -77,6 +77,9 @@
     "description": "https://schema.org/description",
     "name": "https://schema.org/name",
 
+    "EnvelopedVerifiableCredential":
+      "https://www.w3.org/2018/credentials#EnvelopedVerifiableCredential",
+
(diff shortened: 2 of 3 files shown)


index.html
```diff
@@ -2324,6 +2324,52 @@ <h3>Presentations</h3>
 [[?VC-JOSE-COSE]] specification.
         </p>
 
+        <section>
+          <h4>Enveloped Verifiable Credentials</h4>
+
+          <p>
+It is possible to include one or more <a>verifiable credentials</a> in a
+<a>verifiable presentation</a> that have been secured using a securing mechanism
+that "envelopes" the payload, such as [[[?VC-JOSE-COSE]]] [[?VC-JOSE-COSE]].
```
iherman, line 2333:
```suggestion
that "envelopes" the payload, such as [[[VC-JOSE-COSE]]] [[VC-JOSE-COSE]].
```

The reference to VC-JOSE-COSE is normative elsewhere in the document, including in this section, to putting it here as an informative reference does not have any effect. It would just mess up other editors now or later...

index.html
```diff
@@ -2324,6 +2324,52 @@ <h3>Presentations</h3>
...
+an object that has a `type` of `EnvelopedVerifiableCredential`.
+          </p>
+
+          <dl>
+            <dt id="defn-EnvelopedVerifiableCredential">EnvelopedVerifiableCredential</dt>
+            <dd>
+Used to associate an object containing an enveloped <a>verifiable credential</a>
+with the `verifiableCredential` property in a <a>verifiable presentation</a>.
+The `@context` value of the object MUST follow all of the "MUST" statements in
+Section <a href="#contexts"></a>. The `id` value of the object MUST be a `data:`
```
iherman, line 2344:
```suggestion
The possible `@context` value of the object MUST follow all of the "MUST" statements in
Section <a href="#contexts"></a>. The `id` value of the object MUST be a `data:`
```

Is it really _necessary_ to include a `@context` at this point? The only reason to have it is to use the `id` alias, but that would be inherited from the contexts above, wouldn't it?

index.html
```diff
@@ -2324,6 +2324,52 @@ <h3>Presentations</h3>
...
+URL that expresses a secured <a>verifiable credential</a> using an enveloping
```
iherman, line 2345:
```suggestion
URL [[RFC2397]] that expresses a secured <a>verifiable credential</a> using an enveloping
```

Just adding a reference to the Data URL...

iherman (approved):
None of my comments are show-stoppers.

index.html
msporny: It's cited here as an informative. Throughout the entire document, I try to be very careful about when we cite something normatively vs. informatively. This is important when sections are added/removed, etc. Over time, some citations that were normative become informative. On that basis, I am going to not take this change.

index.html
msporny: Yes, it's necessary, because we clear all context values inside the Verifiable Credential Graph. We have to do this in order to ensure that VCDM v1.1 VCs can be put inside VCDMV v2.0 VPs. If we don't do that, JSON-LD safe mode kicks in and alerts us that we're re-defining some terms (which is the correct behaviour).

index.html
iherman: As I said, this is not a show-stopper.

index.html
iherman: Ah. Right, my bad.

(For other reviewers, in case they do not follow: there is a `"@context": null` statement within the definition of the `VerifiablePresentation.verifiableCredential` property in the V2 `@context` file. The reason is what @msporny describes above.)

Let me picky, though. The text says:

> The @context value of the object MUST follow all of the "MUST" statements in Section [4.2 Contexts](https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1379.html#contexts).

However, §4.2 only says:

> Verifiable credentials and verifiable presentations MUST include a @context [property](https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1379.html#dfn-property).

Technically, i.e., in the vocabulary, an `EnvelopedVerifiableCredential` is _not_ a `VerifiableCredential`, meaning that the statement in §4.2 does not apply, i.e., it is not said that a `@context` file must be present. You can either change the text in §4.2, or add a statement in the new section whereby a `@context` property MUST be added.

Also, continuing to be picky, and looking at §5.12, the text says:

> For presentations, each value associated with the verifiableCredential property of the presentation is a separate named graph of type VerifiableCredentialGraph ***which contains a single verifiable credential.***

(Emphasis is mine.) I believe it should say something like

> which contains a single verifiable credential or and enveloped verifiable credential.

(Or something for that effect.)

msporny: Yes, I believe your statements above are correct, I'll make those modifications in the next iteration of this PR.

index.html
```diff
@@ -2324,6 +2324,52 @@ <h3>Presentations</h3>
...
+security scheme, such as [[[VC-JOSE-COSE]]] [[VC-JOSE-COSE]]. The `type` value
```
OR13, line 2346: Are there 2 here intentionally?

index.html
```diff
@@ -2324,6 +2324,52 @@ <h3>Presentations</h3>
...
+{
+  "@context": [
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
+  ],
+  "type": ["VerifiablePresentation", "ExamplePresentation"],
+  <span class="highlight">"verifiableCredential": [{
+    "@context": "https://www.w3.org/ns/credentials/v2",
+    "id": "data:application/vc+ld+json+sd-jwt;base64,QzVjV...RMjUK==",
+    "type": "EnvelopedVerifiableCredential"
```
OR13, line 2366: This is interesting... I think with some discussion this could be acceptable.

OR13 (approved):

index.html
```diff
@@ -2324,6 +2324,52 @@ <h3>Presentations</h3>
...
+        <pre class="example nohighlight" title="Basic structure of a presentation">
+{
+  "@context": [
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
+  ],
+  "type": ["VerifiablePresentation", "ExamplePresentation"],
+  <span class="highlight">"verifiableCredential": [{
+    "@context": "https://www.w3.org/ns/credentials/v2",
+    "id": "data:application/vc+ld+json+sd-jwt;base64,QzVjV...RMjUK==",
```
OR13, line 2365:
```suggestion
    "id": "data:application/vc+ld+json+sd-jwt;QzVjV...RMjUK==",
```

brianorwhatever:
This PR is great and does a good job in finally disambiguating things that are quite clearly different.

index.html
OR13: afaik, you are not required to double encode media types, for example:

`data:text/csv;charset=utf-8,` does not base64 encode.

decentralgabe (approved):
I'm supportive of this compromise. Thank you.

OR13:
Does this mean that you can't use SD-JWT to create v1 presentations or credentials?

Or that the v2 context is required to make v1 presentations?

index.html
OR13: cc @selfissued @Sakurann please confirm the media type for SD-JWT vc is correct here, and the data URI encoding is correct as well.

index.html
OR13: I don't think the second context is required here... if the v2 context defines things correctly.

dlongley: @OR13,

As explained in the comments above, the second context is indeed necessary as the context is intentionally cleared to support other contexts (such as 1.1). So a context declaration is required.

dlongley (approved):
+1 to @iherman's comments.

index.html
iherman: @OR13 

We can put it another way: if the second context is not used, then the right terms for that object must be `@type` and `@id` instead of `type` and `id`, respectively (this is how the current, v2 version of `@context` works, and I do not see any way to avoid that.)

It is indeed unfortunate that the usage of those aliases force us to use the second context here. But that is water under the bridge, we decided not to use the original, `@`-prefixed values.

index.html
iherman: @OR13, this is a respec trick.

- [[[VC-JOSE-COSE]]] is converted into the official title of the spec
- [[VC-JOSE-COSE]] is converted into a reference to the References' section.

So the short answer to your question is: yes.

index.html
```diff
@@ -2324,6 +2324,52 @@ <h3>Presentations</h3>
 [[?VC-JOSE-COSE]] specification.
         </p>
 
+        <section>
+          <h4>Enveloped Verifiable Credentials</h4>
+
+          <p>
+It is possible to include one or more <a>verifiable credentials</a> in a
+<a>verifiable presentation</a> that have been secured using a securing mechanism
```
TallTed, line 2332:
```suggestion
It is possible for a <a>verifiable presentation</a> to include one or more
<a>verifiable credentials</a> that have been secured using a securing mechanism
```

dmitrizagidulin (approved):
👍

mkhraisha (approved):

msporny:

iherman:
@msporny the PR has been merged, but I do not see any additions reflecting on what I wrote in https://github.com/w3c/vc-data-model/pull/1379#discussion_r1421763931. A quick search on the term `EnvelopedVerifiableCredential` reveals that the only place the term appears is in [§4.11.1 Enveloped Verifiable Credentials](https://w3c.github.io/vc-data-model/#enveloped-verifiable-credentials).

index.html
msporny: @iherman, I applied your requests in 2ffb884853a93cc1195a114267ce25b31ff157e7.

msporny:
> @msporny the PR has been merged, but I do not see any additions reflecting on what I wrote in [#1379 (comment)](https://github.com/w3c/vc-data-model/pull/1379#discussion_r1421763931). 

Woops, apologies. Fixed in 2ffb884853a93cc1195a114267ce25b31ff157e7.

Issue #1374, opened 2023-12-07
Specify guarantees that all securing mechanisms must provide.


msporny:
In PR #1338, @awoie [noted](https://github.com/w3c/vc-data-model/pull/1338/files#r1415684650):
> It would be immensely helpful if we could establish precise requirements for this [robust requirements for securing mechanisms], as verifiers would then have at least some baseline security guarantees.

This PR is to track that concern and possibly add guarantees that all securing mechanisms must provide in order to be "conforming securing mechanisms".

msporny:
@awoie, would making these normative requirements on securing mechanism specifications work for you? For example:

* Securing mechanism specifications MUST have protected all the data in the conforming document returned by the securing mechanism verification algorithm. ALTERNATIVE: Non-protected data MUST NOT be returned in the conforming document returned by the securing mechanism verification algorithm.
* Securing mechanism specifications SHOULD protect information referenced by a URL that is critical to validation. Mechanisms that can achieve this protection include: `relatedResource`, `digestSRI`, `digestMultibase`, well-known permanently cached URLs (such as JSON-LD Context URLs), and RDF Canonicalization (for JSON-LD Context URLs).

OR13:
I am concerned about the "verifiable presentation" side of this.

In relation to securing protocols that use "audience / domain", "nonce / challenge".

If these protocol parameters are not secured, or checked during presentation verification, there can be serious security issues impacting authentication.

msporny:
> I am concerned about the "verifiable presentation" side of this.
> 
> In relation to securing protocols that use "audience / domain", "nonce / challenge".
> 
> If these protocol parameters are not secured, or checked during presentation verification, there can be serious security issues impacting authentication.

Yes, and the language that has been proposed covers those cases. What concrete text are you looking to have added to the specification to cover your concern?

msporny:
PR #1380 has been raised to address this issue. This issue will be closed once PR #1380 has been merged.

awoie:
I'd also miss something like the following:

- A securing mechanism MUST protect the integrity of the verifiable credential
- A securing mechanism MUST verify the authorship of the verifiable credential (although this could be a requirement for the algorithm but in the VCDM)

Are we intentionally allowing strange securing mechanism? These are extreme examples but the current definition would allow securing mechanisms such as phoning home to the issuer; having to call some random number on the phone etc.

awoie:
[...]

PR #1380, opened 2023-12-10
Add requirements for securing mechanism specifications.

msporny:
This PR is an attempt at addressing issue #1374 by adding normative requirements for securing mechanism specifications.

/cc @jyasskin

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1380.html" title="Last updated on Dec 15, 2023, 6:57 PM UTC (5e767d9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1380/40be9e0...5e767d9.html" title="Last updated on Dec 15, 2023, 6:57 PM UTC (5e767d9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2016,11 +2016,6 @@ T(([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9](\.[0-9]+)?|(24:00:00(\.0+)?))
         <h3>Securing Mechanisms</h3>
 
         <p>
-A securing mechanism MUST express the details necessary to evaluate whether a
-<a>verifiable credential</a> or <a>verifiable presentation</a> has been secured
-using it.
-        </p>
-        <p>
 This specification recognizes two classes of securing mechanisms: those that use
 external proofs and those that use embedded proofs. An
 <dfn class="export">external proof</dfn> is one that wraps an expression of
@@ -2035,9 +2030,37 @@ It should be noted that these two classes of securing mechanisms are not
 mutually exclusive.
         </p>
         <p>
-<a>Verifiable credentials</a> and <a>verifiable presentations</a> MUST be secured by at least
-one securing method.
+<a>Verifiable credentials</a> and <a>verifiable presentations</a> MUST be
+secured by at least one securing mechanism.
         </p>
+        <p>
+Securing mechanism specifications:
+        </p>
+
+        <ul>
+          <li>
+MUST document normative algorithms that can be used to secure and verify
+<a>verifiable credentials</a> and <a>verifiable presentations</a>.
+          </li>
+          <li>
+MUST provide a verification mechanism that returns only the information in the
+<a>conforming document</a> that has been secured and MAY provide additional
+information that might be helpful to developers. For the avoidance of doubt,
+the securing mechanism specification MUST NOT return any information that was
+not secured in the <a>conforming document</a>. The concrete interface that is
+expected to be provided is detailed in the
+<a href="#verify-securing-mechanism">securing mechanism verification
+algorithm</a> section.
+          </li>
+          <li>
+SHOULD provide integrity protection for any information referenced by a URL that
+is critical to validation. Mechanisms that can achieve this protection are
+discussed in Section <a href="#integrity-of-related-resources"></a> and Section
+<a href="#base-context"></a>.
+          </li>
+        </ul>
+        </p>
+
         <p>
 Methods of securing <a>verifiable credentials</a> or
 <a>verifiable presentations</a> that embed a proof in the data model MUST use
--- a/index.html
+++ b/index.html
@@ -2044,10 +2044,9 @@ MUST document normative algorithms that can be used to secure and verify
           </li>
           <li>
 MUST provide a verification mechanism that returns only the information in the
-<a>conforming document</a> that has been secured and MAY provide additional
-information that might be helpful to developers. For the avoidance of doubt,
-the securing mechanism specification MUST NOT return any information that was
-not secured in the <a>conforming document</a>. The concrete interface that is
+<a>conforming document</a> that has been secured. It MAY provide an
+additional mechanism to receive other information that might be helpful
+to developers. The concrete interface that is
 expected to be provided is detailed in the
 <a href="#verify-securing-mechanism">securing mechanism verification
 algorithm</a> section.
--- a/index.html
+++ b/index.html
@@ -2039,14 +2039,15 @@ Securing mechanism specifications:
 
         <ul>
           <li>
-MUST document normative algorithms that can be used to secure and verify
-<a>verifiable credentials</a> and <a>verifiable presentations</a>.
+MUST document normative algorithms that provide content integrity protection
+for <a>conforming documents</a>. The algorithms MAY be general in nature and
+MAY be used to secure data other than <a>conforming documents</a>.
           </li>
           <li>
 MUST provide a verification mechanism that returns only the information in the
 <a>conforming document</a> that has been secured. It MAY provide an
 additional mechanism to receive other information that might be helpful
-to developers. The concrete interface that is
+during validation or for debugging purposes. The concrete interface that is
 expected to be provided is detailed in the
 <a href="#verify-securing-mechanism">securing mechanism verification
 algorithm</a> section.


andresuribe87 (approved):

index.html
```diff
@@ -2035,9 +2030,37 @@ <h3>Securing Mechanisms</h3>
...
+        </p>
+        <p>
+Securing mechanism specifications:
         </p>
+
+        <ul>
+          <li>
+MUST document normative algorithms that can be used to secure and verify
+<a>verifiable credentials</a> and <a>verifiable presentations</a>.
+          </li>
```
OR13, line 2044: I don't like this text.

Whats the intention here?

OR13 (approved):
I'm approving assuming that normative algorithms don't need to understand the vc data model.

index.html
```diff
@@ -2035,9 +2030,37 @@ <h3>Securing Mechanisms</h3>
...
+          <li>
+MUST provide a verification mechanism that returns only the information in the
+<a>conforming document</a> that has been secured and MAY provide additional
+information that might be helpful to developers. For the avoidance of doubt,
+the securing mechanism specification MUST NOT return any information that was
+not secured in the <a>conforming document</a>. The concrete interface that is
```
dlongley, line 2050: This seems contradictory to returning additional helpful information. Maybe this:

```suggestion
MUST provide a verification mechanism that returns only the information in the
<a>conforming document</a> that has been secured. It MAY provide an
additional mechanism to receive other information that might be helpful
to developers. The concrete interface that is
```

This way additional information to help developers can be returned (as stated) via an opt-in and clearly specified mechanism.

dlongley (approved):
Approving assuming my clarifying suggestion will be included.

TallTed (commented):
lgtm

index.html
msporny: > My challenge is this seems to create a normative requirement in a foreign spec, to force that spec to "understand" credentials and presentations.

Right, I can see how we don't want to convey that the securing specs need to understand VCs and VPs in order to secure them; they shouldn't. However, "securing bytes" isn't necessarily the right thing to say either (because Data Integrity does more than (on interpretation of) "secure bytes").

I'll try to think of a way to reword this so that it doesn't imply that the securing specifications need to understand VCs/VPs.

index.html
```diff
@@ -2035,9 +2030,36 @@ <h3>Securing Mechanisms</h3>
...
+<a href="#verify-securing-mechanism">securing mechanism verification
+algorithm</a> section.
+          </li>
+          <li>
+SHOULD provide integrity protection for any information referenced by a URL that
+is critical to validation. Mechanisms that can achieve this protection are
+discussed in Section <a href="#integrity-of-related-resources"></a> and Section
+<a href="#base-context"></a>.
+          </li>
+        </ul>
```
awoie, line 2060:
```suggestion
          <li>
MUST provide integrity protection for the <a>conforming document</a>.
          </li>
          <li>
MUST verify the authorship of the <a>verifiable credential</a> or <a>verifiable presentation</a>.
          </li>          
        </ul>        
```

index.html
```diff
@@ -2035,9 +2030,36 @@ <h3>Securing Mechanisms</h3>
 mutually exclusive.
         </p>
         <p>
-<a>Verifiable credentials</a> and <a>verifiable presentations</a> MUST be secured by at least
-one securing method.
+<a>Verifiable credentials</a> and <a>verifiable presentations</a> MUST be
+secured by at least one securing mechanism.
+        </p>
+        <p>
```
awoie, line 2036:
```suggestion
        </p>
        <p>
Verifiers MUST validate and trust the specific securing method, understanding its security implications. This is important because security guarantees MAY vary depending on the type of security mechanisms within a securing class and between securing mechanism classes.
        </p>
        <p>        
```

awoie (changes requested):

index.html
awoie: If they shouldn't understand VCs and VPs, then how can running the securing mechanism verification algorithm verify the authorship of a VC or VP?

awoie: A verifiable credential is "a tamper-evident credential that has authorship that can be cryptographically verified.". I thought the securing mechanism would do that?

index.html
awoie: These two relate to:
> A verifiable credential is a tamper-evident credential that has authorship that can be cryptographically verified.

or
> A [verifiable credential](https://w3c.github.io/vc-data-model/#dfn-verifiable-credential) is a set of tamper-evident [claims](https://w3c.github.io/vc-data-model/#dfn-claims) and metadata that cryptographically prove who issued it.

awoie: IMO, it would be valuable that the verifier knows the `issuer` issued the VC, or the `holder` presented the VP after running the securing mechanism verification algorithm.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-12-13-vcwg#section2-13) on 2023-12-13

- no resolutions were taken

#### 2.13. Specify guarantees that all securing mechanisms must provide. (issue vc-data-model#1374)

_See github issue [vc-data-model#1374](https://github.com/w3c/vc-data-model/issues/1374)._




**Brent Zundel:** specify requirements for securing mechanisms.  
… a PR exists.  

_See github pull request [vc-data-model#1380](https://github.com/w3c/vc-data-model/pull/1380)._




**Brent Zundel:** there is a request for changes from oliver.  

**Manu Sporny:** seems we are on a good trajectory, one thing that is concerning, he is saying verifier needs to know who the issuer of a VC is.  
… that sounds like validation.  
… I will try to make that a part of it, but I don't want to cover trust frameworks, or trust lists.  
… the current text can be made clearer... the securing mechanism should not need to understand our data model.  
… I will try to address oliver's suggestions.  

</details>

jyasskin:
Another possible requirement (that might make sense to discuss in a separate issue) is that the securing specification should/must say what kind of processing is safe on the returned document. In particular, if the securing mechanism secures the JSON representation, like vc-jose-cose and ecdsa-jcs-2019, then it's not safe to subsequently process the document with a generic JSON-LD processor because some of the contexts might have changed since the signature was made. On the other hand, if the securing mechanism secures the RDF, like ecdsa-rdfc-2019, then it's not safe to subsequently process the document as JSON, because it's possible to move properties into unexpected other objects without breaking the signature, and the JSON processor might miss something the issuer needed them to find.

H/T https://medium.com/@markus.sabadello/json-ld-vcs-are-not-just-json-4488d279be43 and see https://github.com/w3c/vc-jose-cose/issues/188.

awoie:
> The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-12-13-vcwg#section2-> 
> * no resolutions were taken
> 
> View the transcript
> **Manu Sporny:** seems we are on a good trajectory, one thing that is concerning, he is saying verifier needs to know who the issuer of a VC is. … that sounds like validation. … I will try to make that a part of it, but I don't want to cover trust frameworks, or trust lists. … the current text can be made clearer... the securing mechanism should not need to understand our data model. … I will try to address oliver's suggestions.

IMO, the securing mechanism verification algorithm should provide assurance that the `issuer` authored the VC, or the `holder` (if present) authored the VP. This is what I understand by "authorship" which is mentioned several times throughout the VCDM specification. If the securing mechanism is not responsible for doing this, then I'd like to understand whose responsibility it is and then we should add a normative statement for that.

awoie:
> > The issue was discussed in a [meeting]([https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-12-13-vcwg#section2->](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-12-13-vcwg#section2-%3E)
> > 
> > * no resolutions were taken
> > 
> > View the transcript
> > **Manu Sporny:** seems we are on a good trajectory, one thing that is concerning, he is saying verifier needs to know who the issuer of a VC is. … that sounds like validation. … I will try to make that a part of it, but I don't want to cover trust frameworks, or trust lists. … the current text can be made clearer... the securing mechanism should not need to understand our data model. … I will try to address oliver's suggestions.
> 
> IMO, the securing mechanism verification algorithm should provide assurance that the `issuer` authored the VC, or the `holder` (if present) authored the VP. This is what I understand by "authorship" which is mentioned several times throughout the VCDM specification. If the securing mechanism is not responsible for doing this, then I'd like to understand whose responsibility it is and then we should add a normative statement for that.

I understand that there is a potential violation of layers but how else would a verifier be able to verify the authorship without validating the specifics of a particular proof, e.g., `verificationMethod`, `iss` etc.?

msporny:
@awoie wrote:
> If validating those profiles and validating the X509 chain is not part of the securing mechanism verification algorithm because it is related to a trust framework, then we have to describe that which means verifiers need to do something on top of what we have defined today (perhaps a validation algorithm?) to use issuer and holder in a way that guarantees authorship.

Right, so I think I understand where you're coming from better than I did before. Thank you for taking the time to explain that, @awoie.

I'll note that we kinda-sorta have a section that speaks to this in the VCDM specification here (see paragraph 2 in this section):

https://w3c.github.io/vc-data-integrity/#relationship-to-verifiable-credentials

and we do talk about `issuer` validation here:

https://w3c.github.io/vc-data-model/#issuer-0

... but you want us to say more, which I agree with, we should say more.

There are three ways that I can see us (partially) addressing the concern:

1. Each securing mechanism has to state clearly how it ties the `verificationMethod` to the `issuer` value. I believe @OR13 would object to this approach, because he doesn't want the securing mechanisms to contain details related to VCs.
2. We normatively define how `issuer` validation works in the VC Data Model for each securing mechanism. I believe @OR13 and @selfissued would object to this approach as well, because they don't want the vc-data-model to contain details specific to the securing mechanisms.
3. We tell implementers that they need to validate the binding in some way, by using the output of the verification algorithm (it succeeded) and the input to the verification algorithm (the `verificationMethods` associated w/ the proofs), which are then bound to the issuer value in some way. We could then provide one example of how it can be done, by comparing the `controller` in the controller document to the `issuer` property. We can't say that's the only way to do it because there might be out-of-band trust lists that establish the link between the digital signature and the issuer in another way using purely out-of-band information.

I'm going to raise a new issue for this, because it's not about requirements for securing mechanism specifications (except for maybe an interface that returns the controller document associated w/ a verification method?), but it's more about validation on the data model that goes beyond document conformance:

Raised issue #1386.

index.html
msporny: Hrm, so this section is about normative requirements for securing mechanism specifications. 

The text above is about normative requirements for the Verifier role. I have raised issue #1386 to track your concern above.

index.html
msporny: I've raised issue #1386 to track this concern.

I took the suggestion for the first bullet item in commit 5e767d9f44f341ccd423b6a2e99268a493db3e9d.

index.html
msporny: I raised issue #1386 to track @awoie's concern.

I made an attempt at addressing @OR13's concern in commit 5e767d9f44f341ccd423b6a2e99268a493db3e9d.

awoie:
> @awoie wrote:
> 
> > If validating those profiles and validating the X509 chain is not part of the securing mechanism verification algorithm because it is related to a trust framework, then we have to describe that which means verifiers need to do something on top of what we have defined today (perhaps a validation algorithm?) to use issuer and holder in a way that guarantees authorship.
> 
> Right, so I think I understand where you're coming from better than I did before. Thank you for taking the time to explain that, @awoie.
> 
> I'll note that we kinda-sorta have a section that speaks to this in the DI specification here (see paragraph 2 in this section):
> 
> https://w3c.github.io/vc-data-integrity/#relationship-to-verifiable-credentials
> 
> and we do talk about `issuer` validation in the VCDM here:
> 
> https://w3c.github.io/vc-data-model/#issuer-0
> 
> ... but you want us to say more, which I agree with, we should say more.
> 
> There are three ways that I can see us (partially) addressing the concern:
> 
> 1. Each securing mechanism has to state clearly how it ties the `verificationMethod` to the `issuer` value. I believe @OR13 would object to this approach, because he doesn't want the securing mechanisms to contain details related to VCs.
> 2. We normatively define how `issuer` validation works in the VC Data Model for each securing mechanism. I believe @OR13 and @selfissued would object to this approach as well, because they don't want the vc-data-model to contain details specific to the securing mechanisms.
> 3. We tell implementers that they need to validate the binding in some way, by using the output of the verification algorithm (it succeeded) and the input to the verification algorithm (the `verificationMethods` associated w/ the proofs), which are then bound to the issuer value in some way. We could then provide one example of how it can be done, by comparing the `controller` in the controller document to the `issuer` property. We can't say that's the only way to do it because there might be out-of-band trust lists that establish the link between the digital signature and the issuer in another way using purely out-of-band information.
> 
> I'm going to raise a new issue for this, because it's not about requirements for securing mechanism specifications (except for maybe an interface that returns the controller document associated w/ a verification method?), but it's more about validation on the data model that goes beyond document conformance:
> 
> Raised issue #1386.

Perhaps I'm missing something but I have some concerns that the layering might not be right since the `issuer` value could be different from the issuer of the `proof` of the securing mechanism. Especially, if there are multiple signatures in data integrity. IMO, verifying the securing mechanism allows the verifier to trust the `issuer` value because they trust the issuer of the `proof` and the securing mechanism itself. The challenge ist to allow the verifier to understand who the issuer of the proof is without having to understand all the details of the securing mechanism and proofs themselves.

awoie (commented):
I'm fine with discussing the issuer further in https://github.com/w3c/vc-data-model/issues/1386

msporny:
@jyasskin wrote:
> Another possible requirement (that might make sense to discuss in a separate issue) is that the securing specification should/must say what kind of processing is safe on the returned document.

Tracking in issue #1388.

msporny:

Issue #1375, opened 2023-12-07
Remove repetition of normative statements in verification algorithm


msporny:
@OR13 has asserted that the current verification algorithm (arguably) [repeats a number of normative statements throughout the specification](https://github.com/w3c/vc-data-model/pull/1338#pullrequestreview-1763413263). Those statements are in there because @jyasskin requested that we add them to ensure that implementers had a definitive list of items to check.

The Working Group needs to come to consensus on whether we are keeping that text, removing it, or modifying it in a way that achieves consensus.

jyasskin:
I don't want to insist on exactly this arrangement of the [verifier's data model conformance checks](https://w3c.github.io/vc-data-model/#data-model-conformance-checking). My request that led to this was in https://github.com/w3c/vc-data-model/pull/1338#discussion_r1388509458, and I hope it's clear that the pattern of "ensure that the _X_ property conforms to the "MUST" statements in Section _Y_" is only a little better than "Check whether <var>verified</var>.<var>document</var> is a [=conforming document=]" with the current definition of "conforming document". @OR13's [suggestion](https://github.com/w3c/vc-data-model/pull/1338#discussion_r1414492504) to have the "conforming document" definition name the "MUST" statements brings them closer together.

I'd rather have an algorithm that returns "conforming" or "nonconforming" for a document, probably by calling down to a sub-algorithm for each property, but I won't object to the more declarative version that has the verifiers grep for "MUST".

Which is a long way of saying that the route to consensus is probably to do what @OR13's asking.

msporny:
PR #1381 has been raised to address this issue. This issue will be closed once PR #1381 has been merged.

msporny:
PR #1381 has been merged, closing.

PR #1381, opened 2023-12-10
Remove document conformance checking algorithm section

msporny:
This PR attempts to address issue #1375 by removing the document conformance checking algorithm section, clarifying the definition of "document conformance", and then using the updated definition to reduce the (arguably) repeated conformance language.

/cc @jyasskin

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1381.html" title="Last updated on Dec 17, 2023, 2:09 AM UTC (a515e6f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1381/40be9e0...a515e6f.html" title="Last updated on Dec 17, 2023, 2:09 AM UTC (a515e6f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -585,15 +585,15 @@ license is suspended).
 
       <section id="conformance" class="normative">
         <p>
-A <dfn>conforming document</dfn> is any concrete expression of the data model
-that complies with the normative statements in this specification.
-Specifically, all relevant normative statements in Sections
+A <dfn>conforming document</dfn> is a compact JSON-LD document that
+complies with all of the normative "MUST" statements in this specification.
+Specifically, the relevant normative "MUST" statements in Sections
 <a href="#basic-concepts"></a>, <a href="#advanced-concepts"></a>, and
-<a href="#syntaxes"></a> of this document MUST be enforced. A serialization
-format for the conforming document MUST be deterministic, bi-directional,
-and lossless as described in Section <a href="#syntaxes"></a>.
-The <a>conforming document</a> MAY be transmitted or stored in any such
-serialization format.
+<a href="#syntaxes"></a> of this document MUST be enforced. Other serialization
+formats for the conforming document MAY be used, and if they are, any
+transformation to or from a conforming document MUST be deterministic,
+bi-directional, and lossless. The <a>conforming document</a> MAY be transmitted
+or stored in any such serialization format.
         </p>
 
         <p>
--- a/index.html
+++ b/index.html
@@ -594,10 +594,7 @@ Sections <a href="#basic-concepts"></a>, <a href="#advanced-concepts"></a>, and
 A conforming document is either a <a>verifiable credential</a> that MUST be
 serialized using the `application/vc+ld+json` media type or a
 <a>verifiable presentation</a> that MUST be serialized using the
-`application/vp+ld+json` media type. Other serialization formats for the
-conforming document MAY be used, and if they are, further guidance in Section <a
-href="#ecosystem-compatibility"></a> MUST be followed. The <a>conforming
-document</a> MAY be transmitted or stored in any such serialization format.
+`application/vp+ld+json` media type.
         </p>
 
         <p>
--- a/index.html
+++ b/index.html
@@ -585,15 +585,16 @@ license is suspended).
 
       <section id="conformance" class="normative">
         <p>
-A <dfn>conforming document</dfn> is a compact JSON-LD document that
-complies with all of the relevant "MUST" statements in this specification.
-Specifically, the relevant normative "MUST" statements in Sections
-<a href="#basic-concepts"></a>, <a href="#advanced-concepts"></a>, and
+A <dfn>conforming document</dfn> is a
+<a data-cite="JSON-LD11-API#compaction-algorithms">compacted</a> JSON-LD
+document that complies with all of the relevant "MUST" statements in this
+specification. Specifically, the relevant normative "MUST" statements in
+Sections <a href="#basic-concepts"></a>, <a href="#advanced-concepts"></a>, and
 <a href="#syntaxes"></a> of this document MUST be enforced. Other serialization
 formats for the conforming document MAY be used, and if they are, further
-guidance in Section <a href="#ecosystem-compatibility"></a> MUST be
-followed. The <a>conforming document</a> MAY be transmitted
-or stored in any such serialization format.
+guidance in Section <a href="#ecosystem-compatibility"></a> MUST be followed.
+The <a>conforming document</a> MAY be transmitted or stored in any such
+serialization format.
         </p>
 
         <p>
@@ -4016,7 +4017,7 @@ storage engines.
 
         <p>
 [[!JSON-LD]] is useful when extending the data model described in this
-specification. Instances of the data model are encoded in JSON-LD compact
+specification. Instances of the data model are encoded in JSON-LD compacted
 form [[!JSON-LD]] and include the <code>@context</code> <a>property</a>. The
 <a href="https://www.w3.org/TR/json-ld/#the-context">JSON-LD context</a>
 is described in detail in the [[!JSON-LD]] specification and its use is
@@ -4037,7 +4038,7 @@ Model is available at <code>https://www.w3.org/2018/credentials</code>.
         <p>
 This specification restricts the usage of JSON-LD representations of
 the data model. JSON-LD <a
-href="https://www.w3.org/TR/json-ld/#compacted-document-form">compact document
+href="https://www.w3.org/TR/json-ld/#compacted-document-form">compacted document
 form</a> MUST be utilized for all representations of the data model in the
 base media type, `application/vc+ld+json`.
         </p>
@@ -7218,7 +7219,7 @@ Add table of reserved properties for properties that are not yet standardized
 or are at risk for removal.
         </li>
         <li>
-Restrict data model expression to JSON-LD in compact document form.
+Restrict data model expression to JSON-LD in compacted document form.
         </li>
         <li>
 Update ZKP section to remove older content.
--- a/index.html
+++ b/index.html
@@ -4503,25 +4503,12 @@ information related to the verification failure.
           </li>
           <li>
 Ensure that the <var>result</var>.<var>document</var> is a
-<a>conforming document</a> by performing the following steps:
-            <ol class="algorithm">
-              <li>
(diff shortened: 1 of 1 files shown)


dlongley (approved):

iherman (approved):

decentralgabe (approved):

OR13 (approved):

index.html
```diff
@@ -585,15 +585,15 @@ <h3>Use Cases and Requirements</h3>
...
-<a href="#syntaxes"></a> of this document MUST be enforced. A serialization
-format for the conforming document MUST be deterministic, bi-directional,
-and lossless as described in Section <a href="#syntaxes"></a>.
-The <a>conforming document</a> MAY be transmitted or stored in any such
-serialization format.
+<a href="#syntaxes"></a> of this document MUST be enforced. Other serialization
+formats for the conforming document MAY be used, and if they are, further
+guidance in Section <a href="#ecosystem-compatibility"></a> MUST be
+followed. The <a>conforming document</a> MAY be transmitted
+or stored in any such serialization format.
```
jyasskin, line 596: I'm not sure what this means for the result type of the verification algorithm. Does that need to be a compact JSON-LD document, or can it be one of the other serialization formats?

index.html
```diff
@@ -585,15 +585,15 @@ <h3>Use Cases and Requirements</h3>
...
+Specifically, the relevant normative "MUST" statements in Sections
 <a href="#basic-concepts"></a>, <a href="#advanced-concepts"></a>, and
-<a href="#syntaxes"></a> of this document MUST be enforced. A serialization
-format for the conforming document MUST be deterministic, bi-directional,
-and lossless as described in Section <a href="#syntaxes"></a>.
-The <a>conforming document</a> MAY be transmitted or stored in any such
-serialization format.
+<a href="#syntaxes"></a> of this document MUST be enforced. Other serialization
+formats for the conforming document MAY be used, and if they are, further
+guidance in Section <a href="#ecosystem-compatibility"></a> MUST be
```
jyasskin, line 594: I noticed that https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1381.html#ecosystem-compatibility includes "A conforming document is either a verifiable credential serialized as the application/vc+ld+json media type or a verifiable presentation serialized as the application/vp+ld+json media type." That piece of the definition should move here.

index.html
```diff
@@ -585,15 +585,15 @@ <h3>Use Cases and Requirements</h3>
 
       <section id="conformance" class="normative">
         <p>
-A <dfn>conforming document</dfn> is any concrete expression of the data model
-that complies with the normative statements in this specification.
-Specifically, all relevant normative statements in Sections
+A <dfn>conforming document</dfn> is a compact JSON-LD document that
```
jyasskin, line 588: Is there an algorithm somewhere that verifiers can use to check that something is a compact JSON-LD document? I see https://w3c.github.io/json-ld-api/#compaction-algorithms defines how to compact a JSON-LD document: is the "is a compact JSON-LD document" predicate something like "https://w3c.github.io/json-ld-api/#compaction-algorithms returns a document that's identical to the original"?

index.html
TallTed, line 596: "Other serialization formats for the conforming document MAY be used" except that it's just said that "A <dfn>conforming document</dfn> is a compact JSON-LD document" which implies rather strongly that "A <dfn>conforming document</dfn> is" not in any "Other serialization formats".

How a reader is to determine which statements (normative or otherwise) are _relevant_ would be a challenge we're now laying out to them, rather than solving it ourselves, by making the relevance of such statements unquestionably obvious.

_NOTE: The suggested revision cannot be applied automatically, because it includes both of the line ranges shown above as inserts 588-591 and 592-596 (and below as deletions 588-591 and 597-601), as well as the line range shown both above and below as deletions 592-596._
```suggestion
A <dfn>conforming document</dfn> is, or can be deterministically transformed
into, a compact JSON-LD document that complies with all of the relevant
normative "MUST" statements in this specification. More specifically, the
document MUST comply with all relevant normative "MUST" statements in Sections
<a href="#basic-concepts"></a>, <a href="#advanced-concepts"></a>, and
<a href="#syntaxes"></a> of this document. If the document is serialized as
anything other than compact JSON-LD, further guidance in Section
<a href="#ecosystem-compatibility"></a> MUST be followed.
```

TallTed (changes requested):

index.html
dlongley: @TallTed, I don't think we want to change the definition of the conforming document -- but we may want to say that the conforming document can be stored / transported / enveloped using other formats provided that the conforming document can be deterministically reproduced. I think that's the aim here.

TallTed: @msporny @dlongley 

The definition is here being changed from the previous text to the new, and in ways which seem (potentially) problematic to me. To start with, from --
> any concrete expression of the data model
that complies with the normative statements in this specification.

-- to --
> a compact JSON-LD document that
complies with all of the relevant "MUST" statements in this specification.

The first is VERY broad and open. The second is VERY narrow and closed.

As I said in my intro to the change suggested above, the new text starts by saying that only Compact (where I capitalize the `C` because it may easily be overlooked otherwise, suggesting that any form of JSON-LD is sufficient) JSON-LD _might_ (if it complies with all the MUSTs) be a conforming document.

Only later does the new text suggest that other formats *might* be usable, but only (it appears to me) for storage or transmission (so evaluation, verification, etc., first requires transformation to Compact JSON-LD), making me question what value there could be in such storage of transmission and further wonder why not simply require that it always be kept in Compact JSON-LD?

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-12-12-vcwg-special#section2-5) on 2023-12-12

- no resolutions were taken

#### 2.5. Remove document conformance checking algorithm section (pr vc-data-model#1381)

_See github pull request [vc-data-model#1381](https://github.com/w3c/vc-data-model/pull/1381)._




**Brent Zundel:** now to 1381.  
… remove document checking section.  

**Manu Sporny:** Orie requested changes on the verification algorithm, specifically that part that checked document conformance.  
… It did it in a way to call out every single property to check.  
… That's what Jeffrey asked for.  
… Orie disagreed and wants it to be restored.  
… So, basically "read the doc, do the MUSTs".  
… Jeffrey was ok with that. This PR does that by becoming much more concise, accurate, about what a conforming document is or is not.  
… document: must be compact JSON-LD that follows all of the MUST statements in the specification, pointing to the sections with those MUST sections.  
… and it does a few more things TallTed is worried about.  
… it also aligns the discussion about "Big Tent" into the credential type specific processing section.  
… Also, "you may, transmit and store a conforming document in a variety of serialization formats" as long you can bring that back.  
… that invites others to profile VCs.  

> *Ivan Herman:* Manu, look at [https://github.com/w3c/vc-data-model/pull/1381#discussion_r1424109226](https://github.com/w3c/vc-data-model/pull/1381#discussion_r1424109226) for possible bike shedding on terminology.

**Ted Thibodeau Jr.:** I don't see the path to resolving my concerns. My comments said the definition change is substantial.  
… The degree of substantialness...  
… yes you can wrap it, but since you have to bring it back, what's the point.  
… I'm not understanding the motivations.  

**Brent Zundel:** would it be fair to say since a conforming document is well described why talk about other serializations?  

**Manu Sporny:** one path is to remove the language about serializations. because it should be obvious.  
… but we spent a long time discussing that, so maybe we should put it in there, even though it's blindingly obvious.  

> *Dave Longley:* i think we need to assuage the concerns of people who want to transform/envelope/whatever as long as you can come back again (even if it's obvious you can do that).

> *Dave Longley:* +1 to Manu, it's been a concern from others.

> *Ted Thibodeau Jr.:* "Deterministic and 100% reversible transformation to or encapsulation in other media types is permissible.".

**Joe Andrieu:** I share TallTed's concerns, I appreciate we want to have roadmarkers for conversations we've had before, we should have language that says "other serializations" are not VCs. I think we need to make it clear that those other things aren't VCs. I don't want people to think that they can call those other things VCs.  

**Brent Zundel:** note we do say that "serialization must be ..." we already say that.  
… what we can say, we have already agreed on and said.  
… So I'm leaning towards just keeping what we have.  

**Manu Sporny:** that language is being deleted in this PR.  

**Brent Zundel:** can we not delete it?  

**Manu Sporny:** yes. that's an option.  
… people are saying reasonable things.  

**Brent Zundel:** looks like we have a path forward for 1381.  

> *Dave Longley:* "another serialization that transforms a conforming document is not a conforming document itself... and it must be (bi-directional, yada-yada)" maybe.

</details>

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-12-13-vcwg#section2-8) on 2023-12-13

- no resolutions were taken

#### 2.8. Remove repetition of normative statements in verification algorithm (issue vc-data-model#1375)

_See github issue [vc-data-model#1375](https://github.com/w3c/vc-data-model/issues/1375)._




**Brent Zundel:** labeled with PR exists, manu can you give an update?  

_See github pull request [vc-data-model#1381](https://github.com/w3c/vc-data-model/pull/1381)._




**Manu Sporny:** lots of approvals, some unaddressed changes and questions, does not seem to have anything blocking.  
… it would remove a large chunk of duplicated normative statements.  
… seems this will merge over the weekend.  

</details>

index.html
msporny: @jyasskin wrote:
> is the "is a compact JSON-LD document" predicate something like "https://w3c.github.io/json-ld-api/#compaction-algorithms returns a document that's identical to the original"?

Yes, exactly. 

@iherman wrote:
> there is a slight terminology mismatch, and we may want to use the term 'Compacted' rather than 'Compact'.

Fixed in 80bf0fadbd3a89a56eb2e8221badb48539b8254d

index.html
msporny: Good catch, fixed in: 7edc347c9afe0d25e3ff11f10a84a5e533566c09

index.html
msporny: The WG didn't like that language, as of the last meeting. It's been removed in a515e6fbfed5b28bd31af56fac790836567b8242.

msporny: The problematic text has been removed in commit (based on consensus reached in the last WG meeting): a515e6fbfed5b28bd31af56fac790836567b8242

msporny:

Issue #1378, opened 2023-12-07
Update algorithms section to normatively depend on the INFRA specification


msporny:
@jyasskin and @iherman [requested that the specification uses terminology from INFRA](https://github.com/w3c/vc-data-model/pull/1338#discussion_r1394242350) to [define data structures](https://github.com/w3c/vc-data-model/pull/1338#discussion_r1388489174) and [algorithms](https://github.com/w3c/vc-data-model/pull/1338#discussion_r1388488567). @OR13 noted that he would not object to such usage of INFRA.

This issue is to track that concern and ensure that we normatively depend on INFRA before we enter the Candidate Recommendation phase.

OR13:
The result of verifying a secured credential or presentation is an INFRA map expressing a conforming document.

Conforming document implies compact JSON-LD document implementing all normative statements associated with credentials or presentations.

msporny:
PR #1383 has been raised to address this issue. This issue will be closed once PR #1383 has been merged.

msporny:
PR #1383 has been merged, closing.

PR #1383, opened 2023-12-10
Add normative references to INFRA in algorithms

msporny:
This PR attempts to address issue #1378 by normatively referring to INFRA in the algorithms section as well as using data types defined in INFRA.

/cc @jyasskin

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1383.html" title="Last updated on Dec 17, 2023, 2:27 AM UTC (ec10cca)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1383/09fce9b...ec10cca.html" title="Last updated on Dec 17, 2023, 2:27 AM UTC (ec10cca)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4524,8 +4524,8 @@ This section contains an algorithm that <a>conforming verifier
 implementations</a> MUST run when verifying a <a>verifiable credential</a> or a
 <a>verifiable presentation</a>. This algorithm takes a sequence of bytes
 ([=byte sequence=] <var>inputBytes</var>) and a media type
-([=string=] <var>inputMediaType</var>) as inputs, and produces a result
-([=map=] <var>result</var>) that contains the following:
+([=string=] <var>inputMediaType</var>) as inputs, and returns a [=map=]
+that contains the following:
         </p>
 
         <ul>
--- a/index.html
+++ b/index.html
@@ -606,9 +606,7 @@ A <dfn class="lint-ignore">conforming issuer implementation</dfn> produces
 <a>conforming documents</a>, MUST include all required properties in the
 <a>conforming documents</a> that it produces, and MUST secure the <a>conforming
 documents</a> it produces using a securing mechanism as described in Section
-<a href="#securing-mechanisms"></a>. Conformance requirements
-phrased as algorithms or specific steps MAY be implemented in any manner, so
-long as the end result is equivalent.
+<a href="#securing-mechanisms"></a>.
         </p>
 
         <p>
@@ -4488,8 +4486,11 @@ Verifiable Credential Data Integrity [[VC-DATA-INTEGRITY]].
       <h2>Algorithms</h2>
 
       <p>
-This section contains algorithms that can be used by implementations to
-perform common operations.
+This section contains algorithms that can be used by implementations to perform
+common operations, such as verification. Conformance requirements phrased as
+algorithms or specific steps utilize normative concepts from the [[[INFRA]]]
+[[INFRA]]. See the section on <a data-cite="INFRA#conformance">Conformance</a>
+in the [[[INFRA]]] for more guidance on implementation requirements.
       </p>
 
       <p class="issue atrisk" title="Issues need resolution before Candidate Recommendation">
--- a/index.html
+++ b/index.html
@@ -133,7 +133,7 @@
             companyURL: "https://www.transmute.industries/",
             note: "v2.0", w3cid: 109171}
         ],
-        xref: ["url", "i18n-glossary"],
+        xref: ["URL", "I18N-GLOSSARY", "INFRA"],
         lint: { "informative-dfn": false },
         maxTocLevel: 2,
         inlineCSS: true
@@ -4522,27 +4522,27 @@ versions of this specification.
         <p>
 This section contains an algorithm that <a>conforming verifier
 implementations</a> MUST run when verifying a <a>verifiable credential</a> or a
-<a>verifiable presentation</a>. This algorithm takes a byte sequence (bytes
-<var>inputBytes</var>) and a media type (string <var>inputMediaType</var>) as
-inputs, and produces an object (<var>result</var>) that contains the
-following:
+<a>verifiable presentation</a>. This algorithm takes a sequence of bytes
+([=byte sequence=] <var>inputBytes</var>) and a media type
+([=string=] <var>inputMediaType</var>) as inputs, and produces a result
+([=map=] <var>result</var>) that contains the following:
         </p>
 
         <ul>
           <li>
-a status (boolean <var>status</var>)
+a status ([=boolean=] <var>status</var>)
           </li>
           <li>
-a <a>conforming document</a> (object <var>document</var>)
+a <a>conforming document</a> ([=map=] <var>document</var>)
           </li>
           <li>
-a media type (string <var>mediaType</var>)
+a media type ([=string=] <var>mediaType</var>)
           </li>
           <li>
-zero or more warnings (array of <a>ProblemDetails</a> <var>warnings</var>)
+zero or more warnings ([=list=] of <a>ProblemDetails</a> <var>warnings</var>)
           </li>
           <li>
-zero or more errors (array of <a>ProblemDetails</a> <var>errors</var>)
+zero or more errors ([=list=] of <a>ProblemDetails</a> <var>errors</var>)
           </li>
         </ul>
 
@@ -4552,7 +4552,7 @@ The verification algorithm is as follows:
 
         <ol class="algorithm">
           <li>
-Set <var>result</var> to an empty object.
+Set <var>result</var> to an empty [=map=].
           </li>
           <li>
 Ensure that the securing mechanism has properly protected the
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -4703,7 +4575,7 @@ <h3>Verify Securing Mechanism</h3>
 <var>inputBytes</var> as an input parameter, parses the value as JSON into
 an object, removes the `proof` property, and returns the result as an object.
 If JSON parsing fails, a <a href="#PARSING_ERROR">PARSING_ERROR</a> is appended
```
iherman, line 4577: (@jyasskin has probably more experience in using infra, so he should be the arbiter on whether the following comment is justified or not...)

In step 5.1) below, it says, that _result.document_ is set to the output of this algorithm. In the introduction of this section, _result.document_ is set to an (infra) _map_. This means that the _extractDocument_ should produce, formally, a _map_ and not an "object" as it says here.

Fortunately, that is also covered by the infra standard: it defines the ["parse JSON bytes to an infra value"](https://infra.spec.whatwg.org/#parse-json-bytes-to-an-infra-value) step that formally provides that "bridge" between the input bytes into what is needed in infra. I think this is what should be used here.

Note that the same comment applies to 3.1.1)

index.html
```diff
@@ -4743,13 +4615,13 @@ <h3>Verify Securing Mechanism</h3>
 Set <var>securedDocument</var> to the result of a JSON parsing function that
```
iherman, line 4615: GitHub does not let me comment on text that has not changed, so I put it here; it concerns the 3.1) step. That text uses the terminology "by utilizing the `proof`.`type` property". However, these operations happen on the result of bridging the JSON bytes onto Infra, so (I presume) we are talking about considering an infra _map_ at this step. If so, the terminology is "key" and not "property".

decentralgabe (approved):

index.html
```diff
@@ -4446,27 +4447,27 @@ <h3>Verification</h3>
...
 This section contains an algorithm that <a>conforming verifier
 implementations</a> MUST run when verifying a <a>verifiable credential</a> or a
-<a>verifiable presentation</a>. This algorithm takes a byte sequence (bytes
-<var>inputBytes</var>) and a media type (string <var>inputMediaType</var>) as
-inputs, and produces an object (<var>result</var>) that contains the
-following:
+<a>verifiable presentation</a>. This algorithm takes a sequence of bytes
+([=byte sequence=] <var>inputBytes</var>) and a media type
+([=string=] <var>inputMediaType</var>) as inputs, and produces a result
+([=map=] <var>result</var>) that contains the following:
```
jyasskin, line 4453: For wording, consider

```suggestion
([=string=] <var>inputMediaType</var>) as inputs, and returns a [=map=]
that contains the following:
```

Including `<var>result</var>` is fine, but readers don't usually need to know what variable the algorithm uses to store its return value.

This return type is also more like a [struct](https://infra.spec.whatwg.org/#structs) than a map, and the fact that it's repeated here and in Verify Securing Mechanism, and that I copied it to [w3c/vc-data-integrity#227](https://github.com/w3c/vc-data-integrity/pull/227/files#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051R3050-R3066), makes me think it should be given a name and moved to https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1383.html#algorithms. Doing that in a separate PR is fine.

Note that, strictly, you refer to a struct's [items](https://infra.spec.whatwg.org/#struct-item) with `|myStruct|'s [=struct type/item name=]`, and you refer to map [entry](https://infra.spec.whatwg.org/#map-entry) [values](https://infra.spec.whatwg.org/#map-value) with <code>|myMap|["`key`"]</code>, but I think it's fine to use `.` notation with both: I can't imagine implementations diverging based on that.

index.html
jyasskin: That's correct, but I'm also hoping to replace _extractDocument_ entirely with https://github.com/w3c/vc-data-integrity/pull/227, so you might not want to spend time fine-tuning this bit of the algorithm.

index.html
jyasskin: Not all web spec authors agree with me here (e.g. I think the Infra editors disagree), but I don't think it's super important to always use the right terminology for the things inside maps, structs, lists, objects, etc. The important thing is that we avoid any possibility for implementers to understand different things from our text, and I haven't been able to come up with a misinterpretation based on this choice of words.

That said, the other words here are somewhat ambiguous. `proof.type` in what object? It'll be _securedDocument_["`proof`"]["`type`"] (or _securedDocument_.proof.type to be less formal)., but _securedDocument_ hasn't been initialized yet. So the text should probably move below its initialization ... except that https://github.com/w3c/vc-data-integrity/pull/227 teaches vc-data-integrity to do this dispatch based on _inputBytes_, so it's probably not worth changing anything in this PR.

jyasskin (approved):
All my comments are things to think about, not requests.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-12-12-vcwg-special#section2-4) on 2023-12-12

- no resolutions were taken

#### 2.4. Add normative references to INFRA in algorithms (pr vc-data-model#1383)

_See github pull request [vc-data-model#1383](https://github.com/w3c/vc-data-model/pull/1383)._




**Brent Zundel:** a few comments. mostly editorial concerns raised.  

**Manu Sporny:** this PR looks like it will make it.  
… To clarify, we are now taking a normative dependence on INFRA, limited to the algorithm section.  
… We are NOT planning on rewriting the entire spec. We could later.  
… But for now, just the algorithm section.  

> *Ivan Herman:* +1 to Manu on infra.

**Ted Thibodeau Jr.:** my comment lines 588-596 are changed twice.  
… conversation at 1381 is the same.  
… So this should be rebased.  

**Manu Sporny:** plus one. that's probably a mistake on my part. I'll remove it from the PR. Not meant to be in there.  

**Brent Zundel:** so this PR is looking like it will be approved. So, in five days from now, it will likely be merged.  

**Ivan Herman:** to be precise, pending changes are still expected.  

> *Manu Sporny:* Agree with Ivan, things need to change in this PR to go in.

</details>

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-12-13-vcwg#section2-11) on 2023-12-13

- no resolutions were taken

#### 2.11. Update algorithms section to normatively depend on the INFRA specification (issue vc-data-model#1378)

_See github issue [vc-data-model#1378](https://github.com/w3c/vc-data-model/issues/1378)._




_See github pull request [vc-data-model#1383](https://github.com/w3c/vc-data-model/pull/1383)._




**Brent Zundel:** add infra to algorithms, PR exists.  
… seems to be on a good track.  
… seems this will merge, and things will close.  

</details>

index.html
msporny: Acknowledged on the above, using your rewording. I expect these algorithms will continue to need to be cleaned up (editorially) during CR. One of those items will need to be consistency among all of the algorithms here (and in other specifications).

index.html
```diff
@@ -4446,27 +4447,27 @@ <h3>Verification</h3>
         <p>
 This section contains an algorithm that <a>conforming verifier
 implementations</a> MUST run when verifying a <a>verifiable credential</a> or a
-<a>verifiable presentation</a>. This algorithm takes a byte sequence (bytes
-<var>inputBytes</var>) and a media type (string <var>inputMediaType</var>) as
-inputs, and produces an object (<var>result</var>) that contains the
-following:
+<a>verifiable presentation</a>. This algorithm takes a sequence of bytes
+([=byte sequence=] <var>inputBytes</var>) and a media type
```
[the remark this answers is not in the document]
msporny: My concern w/ `[=MIME type=]` is that I'm not sure that developers implementing this specification type media types in that way (they probably just treat them as strings). I don't know if it will cause them to think that the check is going to be anything other than a straight string compare (in most implementations). At least, that's my current thinking wrt. `inputMediaType` being a string. Happy to be swayed another way if there is a stronger argument for stronger typing.

index.html
msporny: @iherman I'm going to wait and land @jyasskin's PR https://github.com/w3c/vc-data-integrity/pull/227 and then come back and clean up the interfaces here.

index.html
msporny: Yes, waiting on https://github.com/w3c/vc-data-integrity/pull/227 to settle before coming back here and fixing the language. This PR was just to get INFRA integrated in a basic fashion, we'll tighten things up in another pass, after https://github.com/w3c/vc-data-integrity/pull/227 has landed.

index.html
```diff
@@ -585,25 +585,23 @@ <h3>Use Cases and Requirements</h3>
...
-<a href="#syntaxes"></a> of this document MUST be enforced. A serialization
-format for the conforming document MUST be deterministic, bi-directional,
-and lossless as described in Section <a href="#syntaxes"></a>.
-The <a>conforming document</a> MAY be transmitted or stored in any such
-serialization format.
+<a href="#syntaxes"></a> of this document MUST be enforced. Other serialization
+formats for the conforming document MAY be used, and if they are, further
+guidance in Section <a href="#ecosystem-compatibility"></a> MUST be
+followed. The <a>conforming document</a> MAY be transmitted
+or stored in any such serialization format.
```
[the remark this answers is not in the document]
msporny: I accidentally included some #1381 items in this PR. Those items have been removed now. Thanks for the catch.

msporny:
Arguably normative, multiple reviews, changes requested and made, no objections, merging.

PR #1387, opened 2023-12-16
Updating the vocab diagram

iherman:
This PR completes #1379 (which has been merged) in adding the `EnvelopedVerifiableCredential` type to the vocabulary diagram.

Additionally, a minor change is in the vocabulary file itself: instead using a local reference entry to the VCDM spec, it now uses the "official" reference to ensure an up-to-date entry.

Because the preview does not properly handle inputs, etc, I will add a copy of the new SVG diagram into my comment below.

This is an editorial change, and not overly important at that, so I did not mark this PR as pre-CR. For consistency's sake it would be better to merge it, though.

### DIFF
--- a/vocab/credentials/v2/template.html
+++ b/vocab/credentials/v2/template.html
@@ -21,18 +21,6 @@
     </script>
     <script class="remove">
       var respecConfig = {
-        localBiblio: {
-          "VC-DATA-MODEL": {
-            title: "Verifiable Credentials Data Model v2.0",
-            href: "https://www.w3.org/TR/vc-data-model-2.0/",
-            authors: [
-              "Manu Sporny", "Grant Noble", "Dave Longley",
-              "Daniel C. Burnett", "Brent Zundel", "Kyle Den Hartog",
-              "Orie Steel", "Michael B. Jones", "Gabe Cohen"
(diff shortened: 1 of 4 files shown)


iherman:
Here is the new diagram:

vocabulary

decentralgabe (approved):

andresuribe87 (approved):

brentzundel:
perhaps a naïve question, but why isn't `proof` listed as a property of a `VerifiableCredential`?

iherman:
@brentzundel 
> perhaps a naïve question, but why isn't `proof` listed as a property of a `VerifiableCredential`?

Because this diagram is showing exclusively the terms defined in the [credentials vocabulary](https://www.w3.org/2018/credentials#). The `proof` property is defined in the [security vocabulary](https://w3id.org/security).

There may be issues around this division (see, e.g., my [comment elsewhere](https://github.com/w3c/vc-data-model/pull/1389#issuecomment-1860849474)), but that is not the subject of these changes and this PR.

msporny (approved):

msporny:
Editorial, multiple reviews, no changes requested, no objections, merging.

PR #1389, opened 2023-12-17
Clarify that "proof" is an extension point

msporny:
This PR is an attempt to partially address the "before-CR" concerns raised in issue #1316. It makes the following changes to the specification:

1. Separates the "Securing Mechanisms" section from the "Proofs" extension point section.
2. Clarifies that the `proof` property is an extension point that is not specific to Data Integrity. It just so happens that Data Integrity utilizes the extension point, but that doesn't preclude non-Data Integrity approaches from using the same extension point.
3. Removes duplicated normative statements between the Securing Mechanisms section and the Conformance section. There was some hand-wavy content around securing mechanisms that we had in v1.1 because securing mechanisms weren't as strongly defined as they are now in v2.0.
4. Cleans up the text from an editorial perspective from the v1.1 text (and the heavy edits that went in over the past two months around securing mechanisms and verification.

While it might seem like we're making a bunch of normative changes in this PR, we're not. We're just moving normative text around and removing duplicated text. I believe there is one new normative SHOULD (about adding securing mechanisms to the VC Specs Directory), but by and large, this PR is largely editorial (or is supposed to be).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1389.html" title="Last updated on Dec 26, 2023, 3:44 PM UTC (8c5f22a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1389/edcbf7e...8c5f22a.html" title="Last updated on Dec 26, 2023, 3:44 PM UTC (8c5f22a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2024,8 +2024,8 @@ T(([01][0-9]|2[0-3]):[0-5][0-9]:[0-5][0-9](\.[0-9]+)?|(24:00:00(\.0+)?))
         <p>
 This specification recognizes two classes of securing mechanisms: those that use
 enveloping proofs and those that use embedded proofs. An
-<dfn class="export">enveloping proof</dfn> is one that wraps an expression of this
-data model. One such enveloping proof mechanism is JSON Web Token,
+<dfn class="export">enveloping proof</dfn> is one that wraps a serialization of
+this data model. One such enveloping proof mechanism is JSON Web Token,
 which is elaborated on in the [[[VC-JOSE-COSE]]] [[VC-JOSE-COSE]]
 specification. An <dfn class="export">embedded proof</dfn> is a mechanism where
 the proof is included in the data model, using the extension point
@@ -7290,7 +7290,7 @@ Add table of reserved properties for properties that are not yet standardized
 or are at risk for removal.
         </li>
         <li>
-Restrict data model expression to JSON-LD in compacted document form.
+Restrict data model serialization to JSON-LD in compacted document form.
         </li>
         <li>
 Update ZKP section to remove older content.
--- a/index.html
+++ b/index.html
@@ -2053,34 +2053,36 @@ of the [[[?VC-SPECS]]] [[?VC-SPECS]].
         <h3>Proofs</h3>
 
         <p>
-Methods of securing <a>verifiable credentials</a> or
-<a>verifiable presentations</a> that embed a proof in the data model MUST use
-the <code>proof</code> <a>property</a>.
+This specification defines the `proof` property for the expression of
+<a href="#dfn-embedded-proof">embedded proofs</a>.
         </p>
-        <p>
-Methods of securing <a>verifiable credentials</a> or <a>verifiable
-presentations</a> that use an <a>enveloping proof</a>
-MAY use the <code>proof</code> <a>property</a>.
-        </p>
+
         <dl>
           <dt><var>proof</var></dt>
           <dd>
-            <p>
 One or more cryptographic proofs that can be used to detect tampering and verify
 the authorship of a <a>verifiable credential</a> or a <a>verifiable
 presentation</a>. Each proof is a separate <a>named graph</a>
 (referred to as a <dfn class="export">proof graph</dfn>) containing a single
 proof. The specific method used for an <a>embedded proof</a> MUST be identified
 using the <code>type</code> <a>property</a>.
-            </p>
-            <p>
+          </dd>
+        </dl>
+
+        <p>
 A proof for a <a>verifiable credential</a> covers all <a>claims</a>
 included in the corresponding <a>verifiable credential graph</a>.
 See <a href="#presentations-0"></a> for the case when the property is used for a
 <a>verifiable presentation</a>.
-            </p>
-          </dd>
-        </dl>
+        </p>
+
+        <p>
+Methods of securing <a>verifiable credentials</a> or
+<a>verifiable presentations</a> that embed a proof in the data model MUST use
+the <code>proof</code> <a>property</a>. Methods of securing <a>verifiable
+credentials</a> or <a>verifiable presentations</a> that use an <a>enveloping
+proof</a> MAY use the <code>proof</code> <a>property</a>.
+        </p>
 
         <p>
 Because the method used for a mathematical proof varies by representation
@@ -2111,12 +2113,12 @@ the signing date. The example below uses Ed25519 digital signatures.
     }
   },
   <span class="highlight">"proof": {
-    "type": "Ed25519Signature2020",
+    "type": "DataIntegrityProof",
+    "cryptosuite": "eddsa-rdfc-2022",
     "created": "2021-11-13T18:19:39Z",
     "verificationMethod": "https://university.example/issuers/14#key-1",
     "proofPurpose": "assertionMethod",
-    "proofValue": "z58DAdFfa9SkqZMVPxAQpic7ndSayn1PzZs6ZjWp1CktyGesjuTSwRdo
-                   WhAfGFCF5bppETSTojQCrfFPP2oumHKtz"
+    "proofValue": "z58DAdFfa9SkqZMVPxAQp...jQCrfFPP2oumHKtz"
   }</span>
 }
         </pre>
--- a/index.html
+++ b/index.html
@@ -2036,8 +2036,8 @@ mechanisms are not mutually exclusive.
 
         <p>
 It is expected that securing mechanism specifications other than the ones
-referred to above will be created as necessary. It is important that the
-authors of any securing mechanism specification:
+referred to above will be created as necessary. A securing mechanism
+specification:
         </p>
 
         <ul>
--- a/index.html
+++ b/index.html
@@ -2068,6 +2068,20 @@ of the [[[?VC-SPECS]]] [[?VC-SPECS]].
           </li>
         </ul>
 
+        <p class="note"
+           title="Choice in securing mechanism is use-case dependent">
+There are multiple acceptable securing mechanisms and this specification does
+not mandate any particular securing mechanism for use with
+<a>verifiable credentials</a> or <a>verifiable presentations</a>.
+The Working Group that produced this specification did standardize two
+securing mechanism options, which are:
+[[[VC-DATA-INTEGRITY]]] [[VC-DATA-INTEGRITY]] and [[[VC-JOSE-COSE]]]
+[[VC-JOSE-COSE]]. Other securing mechanisms that are known to the community
+can be found in the
+<a data-cite="?VC-SPECS#securing-mechanisms">Securing Mechanisms</a> section
+of the [[[?VC-SPECS]]] [[?VC-SPECS]].
+        </p>
+
       </section>
 
       <section>
@@ -2141,17 +2155,6 @@ the signing date. The example below uses Ed25519 digital signatures.
   }</span>
 }
         </pre>
-
-        <p class="note">
-As discussed in Section <a href="#conformance"></a>, there are multiple viable
-proof mechanisms, and this specification does not standardize nor recommend any
-single proof mechanisms for use with <a>verifiable credentials</a> or
-<a>verifiable presentations</a>. For more information about the
-<code>proof</code> mechanisms, see the following specifications: Data Integrity
-[[VC-DATA-INTEGRITY]], Securing Verifiable Credentials using JOSE and COSE
-[VC-JOSE-COSE], and the "Proofs" section of the Verifiable Credential Specifications
-Directory [[VC-SPECS]].
-        </p>
 
       </section>
 
--- a/index.html
+++ b/index.html
@@ -1194,8 +1194,7 @@ to be present; this property is defined by [[JSON-LD]].
       "id": "did:example:c276e12ec21ebfeb1f712ebc6f1",
       "name": "Example University"
     }
-  },
-  "proof": { <span class="comment">...</span> }
+  }
 }
         </pre>
 
@@ -2335,8 +2334,7 @@ The example below shows a <a>verifiable presentation</a>:
   ],
   "id": "urn:uuid:3978344f-8596-4c3a-a978-8fcaba3903c5",
   "type": ["VerifiablePresentation", "ExamplePresentation"],
-  <span class="highlight">"verifiableCredential": [{ <span class="comment">...</span> }],
-  "proof": [{ <span class="comment">...</span> }]</span>
+  <span class="highlight">"verifiableCredential": [{ <span class="comment">...</span> }]
 }
         </pre>
 
@@ -2519,9 +2517,9 @@ self-asserted <a>verifiable credential</a> that holds <a>claims</a> about the
       <span class="highlight">"id": "urn:uuid:313801ba-24b7-11ee-be02-ff560265cf9b"</span>,
       "assertion": "This VP is submitted by the subject as evidence of a legal right to drive"
     },
-    { <span class="comment">...</span> }
+    "proof": { <span class="comment">...</span> }
   }],
-  "proof": [{ <span class="comment">...</span> }]
+  "proof": { <span class="comment">...</span> }
 }
           </pre>
         </section>
@@ -3468,11 +3466,10 @@ by a trusted issuer higher in the EBSI trust chain [?EBSI].
     "credentialSubject": {
       "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
       "degree": {
-      "type": "ExampleBachelorDegree",
-      "name": "Bachelor of Science and Arts"
+        "type": "ExampleBachelorDegree",
+        "name": "Bachelor of Science and Arts"
+      }
     }
-    },
-    "proof": { <span class="comment">...</span> }
   }],
   <span class="highlight">"termsOfUse": [{
     "type": "HolderPolicy",
@@ -3484,8 +3481,7 @@ by a trusted issuer higher in the EBSI trust chain [?EBSI].
       "target": "http://university.example/credentials/3732",
       "action": ["3rdPartyCorrelation"]
     }]
-  }]</span>,
-  "proof": [ ... ]
+  }]</span>
 }
         </pre>
 
@@ -3583,7 +3579,7 @@ Verifiable Credentials Implementation Guidelines [[VC-IMP-GUIDE]] document.
       "name": "Bachelor of Science and Arts"
     }
   },
-  <span class="highlight">"evidence": [{
+  <span class="highlight">"evidence": {
     "id": "https://university.example/evidence/f2aeec97-fc0d-42bf-8ca7-0548192d4231",
     "type": ["DocumentVerification"],
     "verifier": "https://university.example/issuers/14",
@@ -3591,8 +3587,7 @@ Verifiable Credentials Implementation Guidelines [[VC-IMP-GUIDE]] document.
     "subjectPresence": "Physical",
     "documentPresence": "Physical",
     "licenseNumber": "123AB4567"
-  }]</span>,
-  "proof": { <span class="comment">...</span> }
+  }</span>
 }
         </pre>
 
@@ -5893,8 +5888,7 @@ probably ought to be:
       "id": "did:example:c276e12ec21ebfeb1f712ebc6f1",
       "name": "Example University"
     }
-  },
-  "proof": { <span class="comment">...</span> }
+  }
 }
         </pre>
 
@@ -5919,8 +5913,7 @@ protected are advised to use URL schemes that enforce content integrity.
       "id": "did:example:c276e12ec21ebfeb1f712ebc6f1",
       "name": "Example University"
     }
-  },
-  "proof": { <span class="comment">...</span> }
+  }
 }
         </pre>
 
--- a/index.html
+++ b/index.html
@@ -2033,8 +2033,11 @@ defined in Section <a href="#proofs"></a>. One such embedded proof mechanism
 is [[[VC-DATA-INTEGRITY]]] [[VC-DATA-INTEGRITY]]. These two classes of securing
 mechanisms are not mutually exclusive.
         </p>
+
         <p>
-Securing mechanism specifications:
+It is expected that securing mechanism specifications other than the ones
+referred to above will be created as necessary. It is important that the
+authors of any securing mechanism specification:
         </p>
 
         <ul>
@@ -2057,6 +2060,11 @@ SHOULD provide integrity protection for any information referenced by a URL that
 is critical to validation. Mechanisms that can achieve this protection are
 discussed in Section <a href="#integrity-of-related-resources"></a> and Section
 <a href="#base-context"></a>.
+          </li>
+          <li>
+SHOULD register the securing mechanism in the
+<a data-cite="?VC-SPECS#securing-mechanisms">Securing Mechanisms</a> section
+of the [[[?VC-SPECS]]] [[?VC-SPECS]].
           </li>
         </ul>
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -624,39 +626,6 @@ <h3>Use Cases and Requirements</h3>
...
-these mechanisms, as well as new ones, to mature independently and become
-standardized in time. Given there are multiple valid proof mechanisms, this
-specification does not standardize on any single digital signature mechanism.
-One of the goals of this specification is to provide a data model that can be
-protected by a variety of current and future digital proof mechanisms.
-Conformance to this specification does not depend on the details of a particular
-proof mechanism; it requires clearly identifying the mechanism used by
-each <a>verifiable credential</a> and <a>verifiable presentation</a>.
-        </p>
-
```
msporny, line 659: The text above is removed because it came from v1.1 and we've made great strides in v2.0 to clarify the things this language was hand waving around.

index.html
```diff
@@ -2123,27 +2113,16 @@ <h3>Securing Mechanisms</h3>
...
-        <p class="note">
-As discussed in Section <a href="#conformance"></a>, there are multiple viable
-proof mechanisms, and this specification does not standardize nor recommend any
-single proof mechanisms for use with <a>verifiable credentials</a> or
-<a>verifiable presentations</a>. For more information about the
-<code>proof</code> mechanisms, see the following specifications: Data Integrity
-[[VC-DATA-INTEGRITY]], Securing Verifiable Credentials using JOSE and COSE
-[VC-JOSE-COSE], and the "Proofs" section of the Verifiable Credential Specifications
-Directory [[VC-SPECS]].
-        </p>
```
msporny, line 2145: This text wasn't deleted, it was moved up into the Securing Mechanisms section.

index.html
```diff
@@ -598,7 +598,9 @@ <h3>Use Cases and Requirements</h3>
 A conforming document is either a <a>verifiable credential</a> that MUST be
 serialized using the `application/vc+ld+json` media type or a
 <a>verifiable presentation</a> that MUST be serialized using the
-`application/vp+ld+json` media type.
+`application/vp+ld+json` media type. A conforming document MUST be
+secured by at least one securing mechanism as described in Section
+<a href="#securing-mechanisms"></a>.
```
msporny, line 603: By moving this text from the Securing Mechanisms section to the definition of a conforming document, we make it clear that a conforming document MUST be secured. Doing that also enables us to remove a bunch of hand waving we were doing around securing mechanisms.

dlongley (approved):

iherman:
(As usual, I comment on the Linked Data/RDF aspect of the specification.)

***At the moment***, the `proof` property is formally defined in the [DI specification](https://www.w3.org/TR/vc-data-integrity/#proof-sets), and the official URL for the property is `https://w3id.org/security#proof`. The current version of the VCDM only "mentions", so to say, the vocabulary term, but the term itself "belongs" to the DI spec, to its vocabulary, etc.

The proposed PR text shifts the definition. It says, in [§4.10](https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1389.html#proofs):

> This specification defines the proof property for the expression of [embedded proofs](https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1389.html#dfn-embedded-proof).

It goes on by making the semantics statements whereby each proof is a separate named graph, and it makes it clear that the DI usage of the term is only one of its possible usages. I believe that is in contradiction with the fact that the `proof` property is formally defined in the DI specification.

To be clear, ***I have no problem with that "shift"***, but then we have to change the VCDM-DI relationships on that issue. Specifically:

- The `proof` property must be added to the VCDM vocabulary, and that vocabulary items should refer to §4.10 of the VCDM spec for its formal definition. The current `proof` property must be removed from the DI vocabulary. This also means changing the relevant `@context` files as well.
- We have to propose modifications in the DI specification, which should then "refer" to the specification of the `proof` in the VCDM. (This would probably lead to a new CR snapshot for the DI spec. But that is all right, we can do that relatively easily.) 

I realize, however, that it may create problems with deployed specifications.

***Alternatively***, i.e., if we do not want to change the current setup of the DI spec, then, in this PR, the `proof` extension point must be formulated as ***the*** extension point for the DI approach, and if other kinds of embedded proofs are defined, they would be required to define their own version of a property, which would be _something like the `proof` property_.

decentralgabe (approved):
thanks for this. looks clear to me.

index.html
```diff
@@ -2020,24 +1988,21 @@ <h3>Securing Mechanisms</h3>
...
-It should be noted that these two classes of securing mechanisms are not
-mutually exclusive.
-        </p>
-        <p>
-<a>Verifiable credentials</a> and <a>verifiable presentations</a> MUST be
-secured by at least one securing mechanism.
+enveloping proofs and those that use embedded proofs. An
+<dfn class="export">enveloping proof</dfn> is one that wraps a serialization of
+this data model. One such enveloping proof mechanism is JSON Web Token,
+which is elaborated on in the [[[VC-JOSE-COSE]]] [[VC-JOSE-COSE]]
```
selfissued, line 1994: What does the syntax [[[VC-JOSE-COSE]]] [[VC-JOSE-COSE]] do?  That looks odd to me and it may be a mistake.

index.html
```diff
@@ -2020,24 +1988,21 @@ <h3>Securing Mechanisms</h3>
...
+specification. An <dfn class="export">embedded proof</dfn> is a mechanism where
+the proof is included in the data model, using the extension point
+defined in Section <a href="#proofs"></a>. One such embedded proof mechanism
+is [[[VC-DATA-INTEGRITY]]] [[VC-DATA-INTEGRITY]]. These two classes of securing
```
selfissued, line 1998: Is this also a mistake?

index.html
```diff
@@ -2020,24 +1988,21 @@ <h3>Securing Mechanisms</h3>
...
+mechanisms are not mutually exclusive.
         </p>
+
         <p>
-Securing mechanism specifications:
+It is expected that securing mechanism specifications other than the ones
+referred to above will be created as necessary. A securing mechanism
```
selfissued, line 2004:
```suggestion
Securing mechanism specifications other than the ones
referred to above may also be defined as necessary. A securing mechanism
```

index.html
```diff
@@ -2061,38 +2026,63 @@ <h3>Securing Mechanisms</h3>
...
+          </li>
         </ul>
-        </p>
 
-        <p>
-Methods of securing <a>verifiable credentials</a> or
-<a>verifiable presentations</a> that embed a proof in the data model MUST use
-the <code>proof</code> <a>property</a>.
+        <p class="note"
+           title="Choice in securing mechanism is use-case dependent">
```
selfissued, line 2037:
```suggestion
           title="Choice of securing mechanism is use-case dependent">
```

index.html
```diff
@@ -2061,38 +2026,63 @@ <h3>Securing Mechanisms</h3>
...
+
+      <section>
+        <h3>Proofs</h3>
+
         <p>
-Methods of securing <a>verifiable credentials</a> or <a>verifiable
-presentations</a> that use an external proof MAY use the <code>proof</code>
-<a>property</a>.
+This specification defines the `proof` property for the expression of
+<a href="#dfn-embedded-proof">embedded proofs</a>.
```
selfissued, line 2057:
```suggestion
This specification defines the `proof` property extension point for the expression of
<a href="#dfn-embedded-proof">embedded proofs</a>.
It is up to each securing mechanism whether or not to use it, and if so, how to use it.
```

index.html
```diff
@@ -2061,38 +2026,63 @@ <h3>Securing Mechanisms</h3>
...
 <a>verifiable presentation</a>.
-            </p>
-          </dd>
-        </dl>
+        </p>
+
+        <p>
+Methods of securing <a>verifiable credentials</a> or
+<a>verifiable presentations</a> that embed a proof in the data model MUST use
+the <code>proof</code> <a>property</a>. Methods of securing <a>verifiable
```
selfissued, line 2082:
```suggestion
<a>verifiable presentations</a> that embed a proof in the data model MAY use
the <code>proof</code> <a>property</a>. Methods of securing <a>verifiable
```

selfissued (changes requested):
I will approve this PR after my suggestions are accepted.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-12-19-vcwg-special#section2-8) on 2023-12-19

- no resolutions were taken

#### 2.8. Clarify that "proof" is an extension point (pr vc-data-model#1389)

_See github pull request [vc-data-model#1389](https://github.com/w3c/vc-data-model/pull/1389)._




**Brent Zundel:** Clarify that `proof` is an extension point. This PR is addressing issue 1316. I'm going to let Manu talk about it.  

_See github issue [vc-data-model#1316](https://github.com/w3c/vc-data-model/issues/1316)._




**Manu Sporny:** There was a part of #1316 that I thought we could potentially address before CR. This PR attempts to clarify that `proof` is an extension point and always has been, and DI uses it in a way that's compatible, but DI isn't the only way. Other specs can use it too.  
… The only requirement is that you specify the `type` of the `proof`.  
… The concern here is that if we make `proof` only about DI then it's bound explicitly to that and we wanted to avoid that.  
… It makes some non-normative and small normative changes to clarify that. This PR also splits the securing mechanism section into two parts, one is to talk about securing mechanisms generally.  
… We talk about envelope and embedded proofs. The separate part we create the `proof` extension section explicitly.  
… It makes it much clearer that it's non-exclusive.  
… There are some concerns that Ivan has and Mike has concerns, but we shouldn't look at this as an either-or with the PRs that Mike raised and this one. I think we can merge this one and Mike can come back in an make his edits.  
… The key split here is that to talk about securing mechanisms and their specs wholly separately from the `proof` extension point.  

**Brent Zundel:** Thank you, Manu.  

**Michael Jones:** The strange thing I ran into when doing my PR was a statement that all embedded proofs must use the `proof` property. Which seems it's overly tying the hands of people writing proof mechanisms that we haven't thought of.  
… You could easily imagine some future ZKP mechanism embedding a different identifier that's specific to the mechanism. It's fine for it to be an extension point, but it shouldn't be a mandatory one for all embedded proofs. I'll review with that hat on.  

**Manu Sporny:** Totally agree, that is exactly the normative change I was hoping we could make. We're in agreement on that.  
… The previous text forced anyone doing an embedded securing mechanism to utilize proof and we should relax that and +1 to that, Mike says he agrees.  

**Michael Jones:** Ok, we're violently agreeing again, that's good news.  

**Brent Zundel:** Makes me happy.  

</details>

index.html
iherman: It is not a mistake; it is a respec trick.

- [[[VC-JOSE-COSE]]] extracts the exact title of the publication directly into the text
- [[VC-JOSE-COSE]] creates a reference into the "references" section, and includes a formal document reference into the latter

Though useful, it is ugly... but that is beyond our control.

index.html
iherman: see above

index.html
TallTed:
```suggestion
Securing mechanism specifications other than those
referred to above may also be defined, as necessary. A securing mechanism
```

index.html
```diff
@@ -2061,38 +2026,63 @@ <h3>Securing Mechanisms</h3>
...
         </ul>
-        </p>
 
-        <p>
-Methods of securing <a>verifiable credentials</a> or
-<a>verifiable presentations</a> that embed a proof in the data model MUST use
-the <code>proof</code> <a>property</a>.
+        <p class="note"
+           title="Choice in securing mechanism is use-case dependent">
+There are multiple acceptable securing mechanisms and this specification does
```
TallTed, line 2038:
```suggestion
There are multiple acceptable securing mechanisms, and this specification does
```

index.html
TallTed:
```suggestion
This specification defines the `proof` property as an extension point for the
expression of <a href="#dfn-embedded-proof">embedded proofs</a>.
It is up to each securing mechanism to specify whether or not to use it, and
if so, how it is to be used.
```

index.html
```diff
@@ -7115,8 +7089,8 @@ <h2>application/vc+ld+json</h2>
         </p>
 
         <p>
-This media type can be used with credentials secured using an <a>external
-proof</a>.
+This media type can be used with credentials secured using an
```
TallTed, line 7092:
```suggestion
This media type can be used for credentials secured using an
```

index.html
```diff
@@ -7177,7 +7151,7 @@ <h2>application/vp+ld+json</h2>
 
         <p>
 This media type can be used with presentations secured using an
```
TallTed, line 7153:
```suggestion
This media type can be used for presentations secured using an
```

TallTed (changes requested):

msporny:
@selfissued wrote:
> I will approve this PR after my suggestions are accepted.

I have applied all of your change suggestions, as well as @TallTed's change suggestions. Requesting a re-review.

msporny:
@iherman wrote:
> it makes it clear that the DI usage of the term is only one of its possible usages. I believe that is in contradiction with the fact that the `proof` property is formally defined in the DI specification.

When it was created, many years ago, the `proof` property existed "outside" of both the "Verifiable Credentials" and "Data Integrity" specifications (because those specifications didn't exist in their current form then). The `proof` property was first used as an extension point in VCs and then DIDs (for secured DID Documents), but it was never meant to be "owned" by the VC or DID specs. The DI spec ended up formally defining /one/ way of using the `proof` extension point, because that's the only standardization path we have, but was never intended to be the ONLY securing mechanism that would use it. Ideally, `proof` would be defined in the vocabulary document, but those aren't normative, only the specs are... so, I agree that we need to align all the specifications with these realities, and not everything is going to line up cleanly. 

Things are the way they are now because these technologies ended up being standardized in the wrong order and in ways that special-purposed technologies that were always meant to be more generalized. The other thing that complicates all of this is that we do have production deployments of this stuff on a fairly significant scale and so attempting to re-align the formalities of how this stuff is defined is challenging:

* If we change the vocabulary URLs, we break many implementations out there (and there *are* many of them today). I'll assert that `https://w3id.org/security#proof` *is* the proper URL for this value (because it was always meant to be a general solution), so it shouldn't change to being a VC vocabulary-based URL.
* The formal definition of `proof` as "a general extension point for embedded proofs" shouldn't be in the VC spec, though the use of it as a general extension point should be in the VC spec.

The best specification we have for proof is still probably the Data Integrity specification, though, that does send the wrong message (it's not just for Data Integrity usage -- it's for ANY securing mechanism that would like to embed a proof without mandating that ALL securing mechanisms use the `proof` mechanism).

> * The `proof` property must be added to the VCDM vocabulary, and that vocabulary items should refer to §4.10 of the VCDM spec for its formal definition. The current `proof` property must be removed from the DI vocabulary. This also means changing the relevant `@context` files as well.

-1 for the reasons above. We have two less than ideal places to define `proof`... VC or DI. At present, DI remains the best place to put the formal definition (even though it's not ideal) because DI is a more generalized solution than VCs. We could add language to DI to say that `proof` is a generalized extension mechanism to make the point clear. VC can formalize the use of that extension mechanism for the purposes of VCs and we can have normative text in the VC spec about that (basically, this PR).

> * We have to propose modifications in the DI specification, which should then "refer" to the specification of the `proof` in the VCDM. (This would probably lead to a new CR snapshot for the DI spec. But that is all right, we can do that relatively easily.)

-1, because that makes the references backwards. The architectural layering gets all wonky. DI isn't supposed to know anything about VCs to function properly, it would be a layering violation for that to happen.

> _**Alternatively**_, i.e., if we do not want to change the current setup of the DI spec, then, in this PR, the `proof` extension point must be formulated as _**the**_ extension point for the DI approach, and if other kinds of embedded proofs are defined, they would be required to define their own version of a property, which would be _something like the `proof` property_.

Yes, we could do that. I prefer this approach. Can you please suggest some text that would adequately implement what you state above, @iherman?

decentralgabe (approved):

iherman:
@msporny,

I do not want to go into the details of your comments (I do not agree with everything you wrote, but this may not be relevant at this point). I do accept (and actually anticipated) the argument that any change in the exact URL of the `proof` property may create problems with deployments, and therefore we have to adopt a pragmatic, though in my view suboptimal, approach. In other words, I agree with you on:

>> **Alternatively**, i.e., if we do not want to change the current setup of the DI spec, then, in this PR, the `proof` extension point must be formulated as the extension point for the DI approach, and if other kinds of embedded proofs are defined, they would be required to define their own version of a property, which would be _something like the `proof` property_.  
>
> Yes, we could do that. I prefer this approach. Can you please suggest some text that would adequately implement what you state above, @iherman?

Here are some thoughts that I think should be followed (the editorial details are still to be worked out, and I would trust you more than myself to find the right text). 

The core conceptual change is the following: the extension point is ***not*** the `proof` property. *The extension point is the general concept, whereby a particular embedded proof algorithm defines a property that must be used for that particular algorithm.* This specification does not specify the name or the URL of that property (it can be, e.g., `signature`, `secured`, `proof`, etc.). 

The property used by a particular embedded proof algorithm _MUST_ have the following characteristics:

- the property _MUST_ be applicable for both verifiable credentials or verifiable presentations
- the property _MUST_ relate the verifiable credential or the verifiable presentation to a separate named graph containing the details of the proof, referred to as the _proof graph_. The specification of the property must also provide, through `@context` files and corresponding vocabulary specifications, the definition of all terms that are used by that proof graph.
- the property _MUST_ secure:
  - in the verifiable credential case: the default graph of the credential
  - in the verifiable presentation case: the default graph of the presentation, all credential name graphs, as well as all proof graphs related to verifiable credentials

The [[VC-DATA-INTEGRITY]] specification, defined alongside with the VCDM specification, is a standard approach for an embedded proof, relying on the [`proof` property](https://www.w3.org/TR/vc-data-integrity/#dfn-proof); it is this algorithm (method?) that is used as an example for an embedded proof throughout this specification.

Note: it is conceivable (is it?) that, in the verifiable presentation case, a _different_ proof algorithm is used for the verifiable credentials and for the presentation itself. (If this is ***not*** a good idea, then this should be added to the bullet points above.)

We can also mention that the default context of VCDM includes the mapping of the `proof` property to make the usage of DI easier, just as it contains a number of JOSE terms to make the usage of JOSE-COSE easier.

Would that work, @msporny?

iherman:
Just a minor layperson's question here: for enveloped proofs we do not define such a conceptual extension point; we simply refer to JOSE-COSE. I simply do not know whether there are other enveloped proof approaches out there, which may warrant a similar 'extension point' terminology. I am a little bit bothered by the asymmetry here...

msporny:
> Would that work, @msporny?

The approach you suggest could put us back into a situation that the WG was admonished for during v1.0 and v1.1, namely, that we didn't normatively specify how to secure the payloads. The formulation you have above might create this meta-framework on what an embedded securing mechanism MUST do, but stops short in recommending any such mechanism. I'm a -1 to that potential path forward. In the v2.0 WG, we've gone to great lengths to define normative securing mechanisms.

If we were to integrate what you say, we'd go back to "not normatively specifying how to secure the payloads", which I expect would raise formal objections this time around (because we were specifically chartered to standardize securing mechanisms this time around). I expect the push-back will be: "What good is a normative verifiable credentials data model w/o at least one normative mechanism to secure the data model?" ... and that would be a fair question.

What we could do is some variation of the text that you propose (and perhaps this is what you meant by "standard approach for an embedded proof"), as further requirements for securing specifications (started in PR #1380). We could then RECOMMEND at least VC-JOSE-COSE and DI (as we do today in the spec via the normative references). That would give implementers clear normative guidance on at least two recognized ways of securing the data model without saying that other mechanisms with equivalent characteristics are not allowed.

If that works for you @iherman, I can re-formulate this PR to do what you say as long as no one else in this PR discussion objects to that direction.

> Note: it is conceivable (is it?) that, in the verifiable presentation case, a different proof algorithm is used for the verifiable credentials and for the presentation itself. (If this is not a good idea, then this should be added to the bullet points above.)

I don't think we should weigh in on whether or not that is a good idea. It's conceivable that there are presentation mechanisms that could reformulate the VCs in a way that moves the proofs to the presentation, or ones where you use selective disclosure proofs on the VCs and a non-selective disclosure proof on the presentation. There are use cases where combinations of these things are valid today and so I don't think it would be good for the group to suggest one approach or another.

@iherman wrote:
> Just a minor layperson's question here: for enveloped proofs we do not define such a conceptual extension point; we simply refer to JOSE-COSE. I simply do not know whether there are other enveloped proof approaches out there, which may warrant a similar 'extension point' terminology.

If we do the above set of changes, we could add specification requirements for "enveloped securing mechanisms" and "embedded securing mechanisms". Most of the requirements would be the same between the two (separate graphs, define all properties, etc.)

iherman:
> > Would that work, @msporny?
> 
> The approach you suggest could put us back into a situation that the WG was admonished for during v1.0 and v1.1, namely, that we didn't normatively specify how to secure the payloads. The formulation you have above might create this meta-framework on what an embedded securing mechanism MUST do, but stops short in recommending any such mechanism. I'm a -1 to that potential path forward. In the v2.0 WG, we've gone to great lengths to define normative securing mechanisms.
> 

I am not sure that is the case. The big difference between then and now is that we (will) have a standard embedded proof mechanism defined by the WG, namely DI. Back than that was not true.

> 
> What we could do is some variation of the text that you propose (and perhaps this is what you meant by "standard approach for an embedded proof"), as further requirements for securing specifications (started in PR #1380). 

I *think* (if I understand well what you write) this is also what I meant :-)

> We could then RECOMMEND at least VC-JOSE-COSE and DI (as we do today in the spec via the normative references). That would give implementers clear normative guidance on at least two recognized ways of securing the data model without saying that other mechanisms with equivalent characteristics are not allowed.

Yes, that is what I wrote above. We define the general mechanism as an extension point, without referring to the `proof`, and then we recommend the mechanism that we standardize.

Ie, somebody _MAY_ come, at some point in the future, with a different mechanism, define a property call, say, `securedBy`, and combine it with VCDM declaring that `securedBy` "plugs" into VCDM via that extension point. And that would still be fine. But we do recommend DI (and JOSE-COSE, but that is a different matter).

> 
> If that works for you @iherman, I can re-formulate this PR to do what you say as long as no one else in this PR discussion objects to that direction.

I have the impression that we violently agree.

> 
> > Note: it is conceivable (is it?) that, in the verifiable presentation case, a different proof algorithm is used for the verifiable credentials and for the presentation itself. (If this is not a good idea, then this should be added to the bullet points above.)
> 
> I don't think we should weigh in on whether or not that is a good idea. It's conceivable that there are presentation mechanisms that could reformulate the VCs in a way that moves the proofs to the presentation, or ones where you use selective disclosure proofs on the VCs and a non-selective disclosure proof on the presentation. There are use cases where combinations of these things are valid today and so I don't think it would be good for the group to suggest one approach or another.

Ok. Forget I asked :-)

> 
> @iherman wrote:
> 
> > Just a minor layperson's question here: for enveloped proofs we do not define such a conceptual extension point; we simply refer to JOSE-COSE. I simply do not know whether there are other enveloped proof approaches out there, which may warrant a similar 'extension point' terminology.
> 
> If we do the above set of changes, we could add specification requirements for "enveloped securing mechanisms" and "embedded securing mechanisms". Most of the requirements would be the same between the two (separate graphs, define all properties, etc.)

+1

msporny:
@iherman wrote:
> I have the impression that we violently agree.

I tried re-writing this PR and it caused a huge number of issues for the other 4 PRs that depend on this PR. So, here's what I'm going to do: I have raised issue #1400 to track your concern @iherman and we'll deal w/ it in a future PR (to be completed this week). 

We'll process the 4 other PRs (that have approvals now) and then apply your suggestions on top of all the other ones that landed.

msporny:
Normative, multiple reviews, changes requested and made, remaining issue tracked in #1400, no objections, merging.

Issue #1386, opened 2023-12-15
Specify that it is important to validate the `issuer` value


msporny:
@awoie noted that we don't specify how one validates the `issuer` value and that might need to be a part of the securing mechanism specifications. @OR13 and @selfissued noted that the securing mechanisms shouldn't know much/anything about VCs. The group needs to figure out how to address @awoie's concern without raising layering objections from @OR13 or @selfissued.

awoie:
Is the following assumption on layering correct? The `issuer` value could be different from the issuer of the proof of the securing mechanism. Especially, if there are proof values, e.g., multiple data integrity signatures. 

In that case, verifying the securing mechanism allows the verifier to trust the `issuer` value because they trust the issuer of the proof and the securing mechanism itself. The challenge is to allow the verifier to understand who the issuer of the proof is without having to understand all the details of the securing mechanism and proofs themselves. If the verifier needs to know the details of the securing mechanism, then this would also represent a layer violation.

awoie:
I think it does not matter where the `issuer` validation happens there is a layer violation either in the VCDM or in the securing mechanism. Not saying that I have a preference where this layer violation happens but I wanted to say that I cannot see a way without that.

awoie:
Perhaps it helps to complete the following sentences ...

- As a verifier I run the securing mechanism verification algorithm, so I can rely on (... ?).
- As a verifier I do (...?) , so I can trust the information in the VC comes from an issuer I trust.

msporny:
> Is the following assumption on layering correct? The `issuer` value could be different from the issuer of the proof of the securing mechanism. Especially, if there are proof values, e.g., multiple data integrity signatures.

Correct, the `issuer` value could be different from the `verificationMethod` expressed in the `proof`. Usually, there is a way of "following your nose" from the `verificationMethod` back to the issuer, such as:

verificationMethod -> controller document -> controller of controller document ==> related to issuer in some way. To provide some concrete examples:

| verificationMethod  | Controller Document | Controller |
| ------------- | ------------- | ------------- |
| `https://issuer.example/#key-1` -> | `https://issuer.example/` -> | `https://issuer.example` |
| `did:example:abcd#key-1` -> | _VDR Entry_  -> | `did:example:abcd` |
| `did:jwk:yMS0x...MWV` -> | _ephemeral_  -> | `did:jwk:yMS0x...MWV` |
[...]

PR #1393, opened 2023-12-17
Clarify how issuer validation occurs.

msporny:
This PR is an attempt to address issue #1386 by clarifying how issuer validation could occur. It establishes an interface between the VCDM and securing specifications for the securing mechanism verification algorithms. It also adds more detail to the issuer validation section in the specification.

NOTE: The algorithms haven't been updated to use the new interface (that will be done in a separate PR). What's important in this PR is that the controller and controller document is being returned by the securing mechanism verification algorithm such that a check between the `issuer` property and the `controller` is possible.

/cc @jyasskin (this PR starts to define the interface between the Securing Mechanisms and the VCDM)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1393.html" title="Last updated on Dec 26, 2023, 4:16 PM UTC (a59c20b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1393/7b22399...a59c20b.html" title="Last updated on Dec 26, 2023, 4:16 PM UTC (a59c20b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -6443,11 +6443,11 @@ to identify an <a>issuer</a> that is known to and trusted by the
 Relevant metadata that is related to the <code>issuer</code> <a>property</a>
 is available to the <a>verifier</a> through the
 <a href="#verify-securing-mechanism">securing mechanism verification
-algorithm</a>. This information includes the controller, which is typically the
-`issuer`, of the verification method used by the securing mechanism to secure
-each <a>verifiable credential</a>. When verifying a
-<a>verifiable presentation</a>, the controller of the verification method is
-typically the `holder`.
+algorithm</a> as defined in Section <a href="#securing-mechanisms"></a>. This
+information includes the controller, which is typically the `issuer`, of the
+verification method used by the securing mechanism to secure each <a>verifiable
+credential</a>. When verifying a <a>verifiable presentation</a>, the controller
+of the verification method is typically the `holder`.
         </p>
 
         <p>
--- a/index.html
+++ b/index.html
@@ -4640,28 +4640,6 @@ zero or more errors ([=list=] of <a>ProblemDetails</a> <var>errors</var>)
         </ul>
 
         <p>
-A securing mechanism's verification algorithm returns a verification result
-with at least the following [=struct/items=]:
-        </p>
-        <ul>
-          <li>
-a verification status ([=boolean=] <var>status</var>)
-          </li>
-          <li>
-a verified document ([=map=] <var>document</var>)
-          </li>
-          <li>
-a media type ([=string=] <var>mediaType</var>)
-          </li>
-          <li>
-a verification method controller ([=string=] <var>controller</var>)
-          </li>
-          <li>
-a controller document ([=map=] <var>controllerDocument</var>)
-          </li>
-        </ul>
-
-        <p>
 The securing mechanism verification algorithm is as follows:
         </p>
 
--- a/index.html
+++ b/index.html
@@ -2016,8 +2016,8 @@ Securing mechanism specifications MUST provide a verification mechanism that
 returns only the information in the <a>conforming document</a> that has been
 secured, without any securing mechanism information included, such as `proof` or
 JOSE/COSE metadata. Specifications MAY provide additional mechanisms to convey
-other information that might be helpful during validation or for debugging
-purposes, such as securing mechanism metadata. A securing mechanism's
+other information that might be helpful (for example, during validation or for debugging
+purposes), such as securing mechanism metadata. A securing mechanism's
 verification algorithm MUST provide an interface that receives a sequence of
 bytes ([=byte sequence=] |inputBytes|) or a document ([=map=] |inputDocument|)
 and a media type ([=string=] |inputMediaType|) as inputs and returns a
--- a/index.html
+++ b/index.html
@@ -2014,12 +2014,14 @@ algorithms MAY be general in nature and MAY be used to secure data other than
         <p>
 Securing mechanism specifications MUST provide a verification mechanism that
 returns only the information in the <a>conforming document</a> that has been
-secured. It MAY provide an additional mechanism to receive other information
-that might be helpful during validation or for debugging purposes. A securing
-mechanism's verification algorithm MUST provide an interface that
-receives an sequence of bytes ([=byte sequence=] |inputBytes|) and a media type
-([=string=] |inputMediaType|) as inputs and returns a verification result with
-at least the following [=struct/items=]:
+secured, without any securing mechanism information included, such as `proof` or
+JOSE/COSE metadata. Specifications MAY provide additional mechanisms to convey
+other information that might be helpful during validation or for debugging
+purposes, such as securing mechanism metadata. A securing mechanism's
+verification algorithm MUST provide an interface that receives a sequence of
+bytes ([=byte sequence=] |inputBytes|) and a media type ([=string=]
+|inputMediaType|) as inputs and returns a verification result with at least the
+following [=struct/items=]:
         </p>
 
         <dl>
--- a/index.html
+++ b/index.html
@@ -6413,11 +6413,20 @@ to identify an <a>issuer</a> that is known to and trusted by the
         </p>
 
         <p>
-Relevant metadata about the <code>issuer</code> <a>property</a> is expected
-to be available to the <a>verifier</a>. For example, an <a>issuer</a> can
-publish information containing the public keys it uses to digitally sign
-<a>verifiable credentials</a> that it issued. This metadata is relevant when
-checking the proofs on the <a>verifiable credentials</a>.
+Relevant metadata that is related to the <code>issuer</code> <a>property</a>
+is available to the <a>verifier</a> through the
+<a href="#verify-securing-mechanism">securing mechanism verification
+algorithm</a>. This information includes the controller, which is typically the
+`issuer`, of the verification method used by the securing mechanism to secure
+each <a>verifiable credential</a>. When verifying a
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2001,44 +2001,64 @@ <h3>Securing Mechanisms</h3>
...
-MUST document post-processing considerations for data returned from verification
-processes. This includes, but is not limited to, what information can and cannot
-be used for decision making, for what period of time after verification, when
-additional information is required, and when care is necessary even when the
-returned data was properly secured.
-          </li>
-          <li>
-SHOULD provide integrity protection for any information referenced by a URL that
-is critical to validation. Mechanisms that can achieve this protection are
-discussed in Section <a href="#integrity-of-related-resources"></a> and Section
```
msporny, line 2033: This text hasn't been deleted, but has instead been rewritten below in non-list form.

index.html
```diff
@@ -2001,44 +2001,64 @@ <h3>Securing Mechanisms</h3>
...
+        </p>
+
+        <p>
+Securing mechanism specifications MUST provide a verification mechanism that
+returns only the information in the <a>conforming document</a> that has been
+secured. It MAY provide an additional mechanism to receive other information
+that might be helpful during validation or for debugging purposes. A securing
+mechanism's verification algorithm MUST provide an interface that
+receives an sequence of bytes ([=byte sequence=] |inputBytes|) and a media type
+([=string=] |inputMediaType|) as inputs and returns a verification result with
```
dlongley, line 2021: I think it must be also possible for securing mechanisms to be valid if they accept a JSON document / parsed JSON document (Map/dictionary?) instead of bytes and an input media type. Many implementations will be written that way for embedded securing mechanisms (and already are), based on common software pipeline practices that will have JSON already being parsed into a document / object before it hits this API.

We should allow for either of these approaches to be used, especially if we are requiring these algorithms to be normative in the securing mechanism specifications.

index.html
```diff
@@ -2001,44 +2001,73 @@ <h3>Securing Mechanisms</h3>
...
+Securing mechanism specifications MUST document normative algorithms that
+provide content integrity protection for <a>conforming documents</a>. The
+algorithms MAY be general in nature and MAY be used to secure data other than
+<a>conforming documents</a>.
+        </p>
+
+        <p>
+Securing mechanism specifications MUST provide a verification mechanism that
+returns only the information in the <a>conforming document</a> that has been
+secured. It MAY provide an additional mechanism to receive other information
```
selfissued, line 2017: How does this sentence apply when Data Integrity is used?  Does it mean that the "proof" member must not be returned in the secured information?  That would be my interpretation but this could be clearer.

msporny: All current Data Integrity specifications secure all the information in the `proof` value. There is no such thing as "unprotected" in Data Integrity. 

That said, we defer what to do to the securing mechanism specifications. It is their remit wrt. what is returned. For example, vc-jose-cose might choose to strip out `proof` values that were not verified during the securing process (but by doing that, it'll make it impossible for upstream processors to verify those proofs). So, there will need to be work done in vc-jose-cose to determine what to do there.

For Data Integrity, I expect all the current cryptosuites we have to return `proof` in the returned data because all information in the `proof` values ARE protected and verified as a part of the securing mechanism verification process.

dlongley: In DI, all verified proofs are secured -- so DI securing mechanisms can return all proofs there were verified using the verifier's accepted cryptosuites from the base verification algorithm.

selfissued: The plain meaning of the text above implies that `proof` is used to secure the secured data and is not in the secured data.  Please be clear that `proof`, if present, must be removed by this algorithm.

selfissued (changes requested):
I will approve this PR after my suggestion about not returning `proof` as part of the secured data is accepted.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-12-19-vcwg-special#section2-4) on 2023-12-19

- no resolutions were taken

#### 2.4. Clarify how issuer validation occurs. (pr vc-data-model#1393)

_See github pull request [vc-data-model#1393](https://github.com/w3c/vc-data-model/pull/1393)._




**Brent Zundel:** "Clarify how issuer validation occurs".  

**Manu Sporny:** This PR is an attempt at addressing Oliver's concern where Oliver is saying that the spec needs to be clear about how you verify that a VC came from a particular issuer.  
… Some people in the group would argue that's validation.  
… And that we don't do that and don't normatively specify that. So what I tried to do is ... we're now specifying an interface between the securing mechanisms and the VCDM verify algorithm -- and it now returns the controller and controller document. That gives you info you could use to check against the issuer.  
… However, that's not the only way to do validation, you could also have a trust list with some keys and you could just trust that -- including the statements in the VC, including the issuer field.  
… In that case the issuer isn't directly tied to the verification method and controller, etc. -- so I tried to address Oliver's concern to ensure the information is available if the issuer validation works one way.  
… I added some non-normative language as well to help address Oliver's concerns. I don't know if these changes together address his concerns.  

**Brent Zundel:** Any comments on this PR?  
… If all of these, if it's possible to find consensus, if the language needs tweaks or not, that is our goal. If you have comments to make please do or say the PR is great, etc. Let us know what we need to change for you if you need it.  

</details>

index.html
```diff
@@ -6414,11 +6443,20 @@ <h3>Issuer</h3>
...
-<a>verifiable credentials</a> that it issued. This metadata is relevant when
-checking the proofs on the <a>verifiable credentials</a>.
+Relevant metadata that is related to the <code>issuer</code> <a>property</a>
+is available to the <a>verifier</a> through the
+<a href="#verify-securing-mechanism">securing mechanism verification
+algorithm</a> as defined in Section <a href="#securing-mechanisms"></a>. This
+information includes the controller, which is typically the `issuer`, of the
+verification method used by the securing mechanism to secure each <a>verifiable
+credential</a>. When verifying a <a>verifiable presentation</a>, the controller
+of the verification method is typically the `holder`.
```
TallTed, line 6453: I *hope* this edit makes this (single) paragraph somewhat clearer.

I wonder whether the first sentence of this paragraph should discuss ```metadata related to the `holder` property``` in addition to ```metadata related to the `issuer` property```.

It might be better to cover verifiable credentials (and issuers) in one paragraph, and verifiable presentations (and holders) in another.

```suggestion
Metadata related to the `issuer` <a>property</a>
is available to the <a>verifier</a> through the
<a href="#verify-securing-mechanism">verification algorithm of the securing
mechanism</a> as defined in Section <a href="#securing-mechanisms"></a>.
This metadata includes identification of the controller of the verification
method used by the securing mechanism to secure each <a>verifiable credential</a>
or <a>verifiable presentation</a>, of which the controller is typically the
respective `issuer` or `holder`.
```

index.html
```diff
@@ -6414,11 +6443,20 @@ <h3>Issuer</h3>
...
-Relevant metadata about the <code>issuer</code> <a>property</a> is expected
-to be available to the <a>verifier</a>. For example, an <a>issuer</a> can
-publish information containing the public keys it uses to digitally sign
-<a>verifiable credentials</a> that it issued. This metadata is relevant when
-checking the proofs on the <a>verifiable credentials</a>.
+Relevant metadata that is related to the <code>issuer</code> <a>property</a>
+is available to the <a>verifier</a> through the
+<a href="#verify-securing-mechanism">securing mechanism verification
+algorithm</a> as defined in Section <a href="#securing-mechanisms"></a>. This
+information includes the controller, which is typically the `issuer`, of the
```
awoie, line 6450:
```suggestion
information includes the verified controller, which is typically the `issuer`, of the
```

awoie (changes requested):
I assume the securing mechanism verifies the returned controller is the controller of the securing mechanism?

index.html
msporny: @awoie wrote:
> I assume the securing mechanism verifies the returned controller is the controller of the securing mechanism?

Yes, correct. If that's not done, you can't trust the securing mechanism. 

Merging your change suggestion.

index.html
msporny: Applied in af271281d8218238c8b03ee014d8d1cb05b4bcba.

msporny:
@awoie and @TallTed, I have applied both of your changes, requesting re-review from each of you.

index.html
msporny: @selfissued fixed in 8c16fec9990f0f5b4b0d16d1fd8d587fa59a5461.

index.html
msporny: Fixed in 02c3744873614afe5ab0ae0cc14d5b0918c15916.

index.html
```diff
@@ -2001,44 +2001,75 @@ <h3>Securing Mechanisms</h3>
...
+<a>conforming documents</a>.
+        </p>
+
+        <p>
+Securing mechanism specifications MUST provide a verification mechanism that
+returns only the information in the <a>conforming document</a> that has been
+secured, without any securing mechanism information included, such as `proof` or
+JOSE/COSE metadata. Specifications MAY provide additional mechanisms to convey
+other information that might be helpful during validation or for debugging
+purposes, such as securing mechanism metadata. A securing mechanism's
```
dlongley, line 2020: To clarify that information can only be helpful for more than just the given two examples:

```suggestion
JOSE/COSE metadata. Specifications MAY provide additional mechanisms to convey
other information that might be helpful (for example, during validation or for debugging
purposes), such as securing mechanism metadata. A securing mechanism's
```

dlongley (approved):

msporny:

Issue #1377, opened 2023-12-07
Rewrite verification algorithm in a way that does not cause layer violations


msporny:
Both @selfissued and @OR13 [raised architectural layering concerns](https://github.com/w3c/vc-data-model/pull/1338/files#r1412874049) around [the way that the verification algorithm is written](https://github.com/w3c/vc-data-model/pull/1338/files#r1409574359). This issue is being raised to track their concerns.

OR13:
I think this is solved so long as there is no language in the core data model about validating proof, or extracting proof from a credential or presentation.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-12-13-vcwg#section2-6) on 2023-12-13

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 2.6. Rewrite verification algorithm in a way that does not cause layer violations (issue vc-data-model#1377)

_See github issue [vc-data-model#1377](https://github.com/w3c/vc-data-model/issues/1377)._




**Brent Zundel:** rewrite verification algorithm in a way that does not cause layering violations.  
… can we close this issue.  

**Manu Sporny:** I think the text in the spec is not optimal, and it needs to align with the new interface langauge to align with the securing mechanism, and its done.  
… I could not figure out how to write that text, you orie, please propose some text.  

**Orie Steele:** Yes, seems like we're in the process of addressing follow-up PRs on verification algorithm. I expect those follow up PRs to resolve this. I'm not sure I have special text to add. Current language isn't acceptable, heading in a good direction.  

**Brent Zundel:** Can you make a concrete suggestion that would make the text acceptable?  

**Orie Steele:** Yes, I think Mike Jones is planning to do that already? I'll leave it to him.  

**Brent Zundel:** can you make a suggestion for text?  

**Manu Sporny:** I prefer to explicitly assign mike jones to this.  

**Brent Zundel:** Mike is assigned, orie, hopefully you can communicate to mike that he is assigned.  

</details>

msporny:
PR #1393 has started to define an interface that, if merged, will enable this PR to be closed (by utilizing an interface that won't cause layer violations).

msporny:
PR #1394 has been raised to address this issue. This issue will be closed when PR #1394 has been merged.

iherman:
[...]

PR #1394, opened 2023-12-18
Use new Securing Mechanism Verify Proof interface

msporny:
This PR attempts to address issue #1377 by normatively defining and then using the new Securing Mechanism Verify Proof interface. The new interface has resulted in a significant reduction in complexity and steps in the algorithms.

/cc @jyasskin

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1394.html" title="Last updated on Dec 26, 2023, 4:33 PM UTC (49abccd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1394/8d81752...49abccd.html" title="Last updated on Dec 26, 2023, 4:33 PM UTC (49abccd)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4572,36 +4572,43 @@ The verification algorithm is as follows:
 
         <ol class="algorithm">
           <li>
-Set <var>result</var> to an empty [=map=].
-          </li>
-          <li>
 Ensure that the securing mechanism has properly protected the
 <a>conforming document</a> by performing the following steps:
             <ol class="algorithm">
-              <li>
-Set <var>securingMechanismVerified</var> to the result of executing the
-<a href="#verify-securing-mechanism">securing mechanism verification
-algorithm</a> with <var>inputBytes</var> and <var>mediaType</var> passed as
-inputs to the algorithm.
+             <li>
+Set the |verifyProof| function by using the |inputMediaType| and the
+<a data-cite="?VC-SPECS#securing-mechanisms">Securing Mechanisms</a> section of
+the [[[?VC-SPECS]]] [[?VC-SPECS]], or other mechanisms known to the
+implementation, to determine the cryptographic suite to use when verifying
+the securing mechanism. The |verifyProof| function MUST implement the interface
+described in <a href="#securing-mechanisms"></a>.
+                <div class="issue"
+                     title="Mechanism for 'determining' is being detailed">
+At present, the Working Group is concerned that the algorithm for "determining"
+might need to be more formally defined. At present, no implementation has
+had an issue determining the proper |verifyProof| algorithm to use, but the
+Working Group is attempting to see if saying more here would be worthwhile.
+                </div>
               </li>
               <li>
-Set <var>status</var>, <var>document</var>, <var>mediaType</var>,
-<var>controller</var>, <var>controllerDocument</var>,
-<var>warnings</var>, and <var>errors</var> in <var>result</var> to their
-respective properties in <var>securingMechanismVerified</var>.
+Set <var>result</var> to the result of passing <var>inputBytes</var> and
+|inputMediaType| to the <var>verifyProof</var> function. If the call was
+successful, <var>result</var> will contain the <var>status</var>,
+<var>document</var>, <var>mediaType</var>, <var>controller</var>,
+<var>controllerDocument</var>, <var>warnings</var>, and <var>errors</var>
+properties.
               </li>
               <li>
-If <var>result</var>.<var>status</var> is `false`, the provided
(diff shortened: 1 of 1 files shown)


TallTed (approved):

index.html
```diff
@@ -4564,35 +4572,43 @@ <h3>Verification</h3>
...
-<a href="#verify-securing-mechanism">securing mechanism verification
-algorithm</a> with <var>inputBytes</var> and <var>mediaType</var> passed as
-inputs to the algorithm.
+             <li>
+Set the |verifyProof| function by using the |inputMediaType| and the
+<a data-cite="?VC-SPECS#securing-mechanisms">Securing Mechanisms</a> section of
+the [[[?VC-SPECS]]] [[?VC-SPECS]], or other mechanisms known to the
+implementation, to determine the cryptographic suite to use when verifying
+the securing mechanism. The |verifyProof| function MUST implement the interface
+described in <a href="#securing-mechanisms"></a>.
```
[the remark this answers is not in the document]
iherman: I do not think that is the case. The charter refers to "normative specification of APIs" as being out of scope. At least in my mind (and I believe that was the concept used for the charter) this means defining, e.g., a WebIDL interface, that can be converted, with property names, arguments, etc, into a Javascript or Python interface. At least in other Working Groups at W3C this is what the term "API" is used for. (Typically, those APIs are supposed to be implemented by browsers. Actually, and in my experience with other groups, the very reason these types of clauses are added to charters is exactly to avoid the impression that the underlying technology is an implementation requirement for browsers…)

In this text, if you follow the link, it gets you to [§4.9 Securing Mechanisms](https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1394.html#securing-mechanisms) which remains still fairly high level:

> Securing mechanism specifications MUST provide a verification mechanism that returns only the information in the [conforming document](https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1394.html#dfn-conforming-document) that has been secured. […] A securing mechanism's verification algorithm MUST provide an interface that receives an sequence of bytes ([byte sequence](https://infra.spec.whatwg.org/#byte-sequence) inputBytes) and a media type ([string](https://infra.spec.whatwg.org/#string) inputMediaType) as inputs and returns a verification result with at least the following [items](https://infra.spec.whatwg.org/#struct-item):

How exactly the securing mechanism is built up for a specific case and specific implementation remains open and very high level. Which, I believe, is perfectly fine charter-wise.

For the avoidance of any doubt, maybe the text can formulate the requirement a bit differently, and not use the term "interface".

iherman (approved):

brentzundel (approved):

longpd (approved):

index.html
```diff
@@ -4564,35 +4572,43 @@ <h3>Verification</h3>
...
+                <div class="issue"
+                     title="Mechanism for 'determining' is being detailed">
+At present, the Working Group is concerned that the algorithm for "determining"
+might need to be more formally defined. At present, no implementation has
+had an issue determining the proper |verifyProof| algorithm to use, but the
+Working Group is attempting to see if saying more here would be worthwhile.
```
dlongley, line 4590:
```suggestion
Working Group is attempting to see if saying more here would be worthwhile.
Additional example language could be added that says that an implementation
might have an allow list of acceptable cryptosuites -- and these will be used as
inputs for finding matching proofs to be verified.
```

index.html
```diff
@@ -4564,35 +4572,43 @@ <h3>Verification</h3>
...
+                </div>
               </li>
               <li>
-Set <var>status</var>, <var>document</var>, <var>mediaType</var>,
-<var>warnings</var>, and <var>errors</var> in <var>result</var> to their
-respective properties in <var>securingMechanismVerified</var>.
+Set <var>result</var> to the result of passing <var>inputBytes</var> and
```
dlongley, line 4594: This algorithm could be written to receive a document instead of bytes (i.e., the bytes have already been parsed) -- and we should allow this. I suspect most implementations already work this way and are built to run on top of a pipeline that has already parsed the bytes as JSON.

msporny: There is a PR this this PR is built on top of that does this: https://github.com/w3c/vc-data-model/pull/1393/commits/02c3744873614afe5ab0ae0cc14d5b0918c15916

I expect that PR to be merged and this PR will inherit those changes that allow bytes or a document. I believe it has to be both because of enveloping data formats, which are often sequences of bytes that don't have a document representation that are passed to the verification algorithm.

msporny:

PR #1396, opened 2023-12-19
Rename "Controller Document" to "Entity Document".

msporny:
This PR renames the "Controller Document" terminology to "Entity Document" because the language for "controller documents" is getting increasingly confusing to use and reason about in this specification (and others, for reasons elaborated on below). 

ActivityPub also uses the concept of a "Controller Document", but calls them "Actor objects". This is an important consideration because, with the announcement of new implementers for ActivityPub -- that is, Threads (141 million users) and Flipboard (145 million users), it's important that we get the naming here right (such that it's easier to reason about the alignment in the future).

Here's how this language change is intended to play out:

* "Entity Documents" describe an entity (a thing with independent existence). 
* The description of an entity can contain properties that describe how to communicate with it in a secure manner (verification relationships, verification material, and service endpoints). 
* "Verification Methods" have controllers, which are entities. 
* "Actors" are a type of Entity that can perform an activity, they are expressed using an "Actor object", which could be viewed as being a subclass of "Entity Document" (we could provide explanatory text describing the the alignment in ActivityPub v2).
* Decentralized Identifier Documents (DID Documents) are also a type of "Entity Document" (which we could rename in DID Core v2).

All this to say, "Actor objects" and "DID Documents" are types of "Entity Documents".

If we continue to use the "Controller Document" terminology, the liklihood of making the statements above, and it making sense to people that have a common understanding of English, is low. It's just too hard to connect the dots. I imagine some people reading this, without the tribal knowledge from the last decade or so in these spaces, will express surprise that these things are related at all.

While this is an editorial change, it's an important one to make, IMHO. It's marked "Before CR" because we need to decide on the name before publishing the VC-CONTROLLER/ENTITY-DOCUMENT specification to TR space, which we have to do before VCDM goes into CR, because VCDM will depend on the VC-CONTROLLER/ENTITY-DOCUMENT (normatively) due to PR #1393.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1396.html" title="Last updated on Jan 6, 2024, 4:36 PM UTC (2d7d65b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1396/b59c955...2d7d65b.html" title="Last updated on Jan 6, 2024, 4:36 PM UTC (2d7d65b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1746,9 +1746,9 @@ The value of the `issuer` [=property=] MUST be either a
 whose value is a [=URL=]; in either case, the issuer selects this
 [=URL=] to identify itself in a globally unambiguous
 way. It is RECOMMENDED that the [=URL=] be one which, if dereferenced,
-results in a controller document [[?VC-CONTROLLER-DOCUMENT]] about the
-[=issuer=] that can be used to [=verify=] the information expressed in
-the [=credential=].
+results in a controller document, as defined in [[VC-DATA-INTEGRITY]] or
+[[VC-JOSE-COSE]], about the [=issuer=] that can be used to [=verify=] the
+information expressed in the [=credential=].
           </dd>
         </dl>
 
@@ -4000,11 +4000,13 @@ A media type as defined in [[RFC6838]].
           </dd>
           <dt>[=string=] |controller|</dt>
           <dd>
-A verification method controller as defined in [[VC-CONTROLLER-DOCUMENT]].
+A verification method controller as defined in [[VC-DATA-INTEGRITY]] or
+[[VC-JOSE-COSE]].
           </dd>
           <dt>[=map=] |controllerDocument|</dt>
           <dd>
-A controller document as defined in [[VC-CONTROLLER-DOCUMENT]].
+A controller document as defined in [[VC-DATA-INTEGRITY]] or
+[[VC-JOSE-COSE]].
           </dd>
         </dl>
 
@@ -4023,7 +4025,7 @@ of the [[[?VC-SPECS]]] [[?VC-SPECS]].
         </p>
 
         <p class="note"
-            title="Choice of securing mechanism is use-case dependent">
+           title="Choice of securing mechanism is use-case dependent">
 There are multiple acceptable securing mechanisms, and this specification does
 not mandate any particular securing mechanism for use with
 [=verifiable credentials=] or [=verifiable presentations=].
@@ -4034,6 +4036,17 @@ securing mechanism options, which are:
 can be found in the
 <a data-cite="?VC-SPECS#securing-mechanisms">Securing Mechanisms</a> section
 of the [[[?VC-SPECS]]] [[?VC-SPECS]].
+        </p>
+
+        <p class="issue atrisk"
+           title="Controller document reference might change">
+The Working Group is currently attempting to align the definitions of a
+controller document between [[[?DID-CORE]]], [[[VC-DATA-INTEGRITY]]], and
+[[[VC-JOSE-COSE]]]. The goal is to have one specification that each of the
+previously stated specifications, and this specification, can reference for
+the normative statements related to controller documents. The normative
+references to controller documents are expected to change during the
+Candidate Recommendation phase.
         </p>
 
       </section>


OR13 (approved):

decentralgabe (approved):
This is a clarifying change.

dlongley (approved):

brianorwhatever (approved):

jandrieu:
It seems like we should find the semantics that relate the entity in question to the VC under discussion.

In particular, I don't think these documents are about entities. Unless you define that entity as control record in a VDR. 

I think what we are talking about are "documents that tell you how to securely interact with reference to an identifier".

Obviously this is an ugly bikeshedding opportunity, but seriously, asserting that a controller document is about an entity is essentially making the DID Document an assertion about that entity, i.e. a VC.

But that is not what controller documents are. They describe how you interact with relation to a given identifier. Assumptions about the entity that is represented by that identifier or who controls that identifier are a huge problem in this space.

I'll repeat what I've said about DID Documents. The statements in DID Documents (and hence controller documents) are NOT statements about the entity. They are statements about how to securely interact with regard to that entity.

jandrieu (changes requested):
These are not "entity" documents. They are identifier documents at best.

See my comment in the main thread.

msporny:
> The statements in DID Documents (and hence controller documents) are NOT statements about the entity. They are statements about how to securely interact with regard to that entity.

... which are statements about the entity, aren't they?

> They are identifier documents at best.

These documents express more than just identifiers, they are descriptive about a thing that has existence. The most general form of what we're talking about is a "resource", but it's not quite as general as that.

Given the choice, @jandrieu, would you want to keep "Controller Document" as the name, or try to settle on something that's better?

jandrieu:
> > The statements in DID Documents (and hence controller documents) are NOT statements about the entity. They are statements about how to securely interact with regard to that entity.
> 
> ... which are statements about the entity, aren't they?

No. But maybe I should have been more precise "They are statements about how to securely interact with regard to that identifier."  That might mean how you authenticate a current session wrt the identifier or how you can verify an attestation is made "by the entity in control of that identifier". However, IMO, arbitrary statements about the entity do not belong in the DID Document. They belong in VCs.  

This is a major layer conflation that we continue to propagate. What's in the DID Document is what we need to interact with the identifier. The expectation that the identifier is consistently referring to some entity is a goal and/or assumption. It is not a fact. When we treat it as a fact, we blind ourselves to real security and privacy problems. The binding of an identifier to an entity is a constructive action that cannot be assumed.

> > They are identifier documents at best.
> 
> These documents express more than just identifiers, they are descriptive about a thing that has existence (the entity). The most general form of what we're talking about is a "resource", but it's not quite as general as that.

My take is different. The concepts referred to by a subject identifier may or may not have existence other than as a fictional concept. 

> Given the choice, @jandrieu, would you want to keep "Controller Document" as the name, or try to settle on something that's better?

Yes.

Controller document at least expresses the semantic relationship of that document to the VC and avoids the layer conflation of thinking of a DID Document as a vehicle for statements about entities.

selfissued:
It would be useful to know in which specifications the term "Controller Document" is used.  Despite my expectations, https://www.w3.org/TR/did-core/ doesn't have any instances of the string "Controller Document".

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-12-19-vcwg-special#section2-2) on 2023-12-19

- no resolutions were taken

#### 2.2. Rename "Controller Document" to "Entity Document". (pr vc-data-model#1396)

_See github pull request [vc-data-model#1396](https://github.com/w3c/vc-data-model/pull/1396)._




**Brent Zundel:** This PR renames "Controller Document" to "Entity Document", it has a good number of approvals and it does have request for changes from Joe.  
… Please jump on the queue.  

**Manu Sporny:** This PR came about because I'm trying to address Oliver's concerns around the spec and how issuer validation occurs and how you get the data for it -- and it meant talking about Controller Documents and connecting keys being used and so on.  
… That means talking about that language around there and it was awkward; we've been using "Controller Document" for a while and it's not lining up quite right. This is a bikeshedding exercise so we don't want to spend too much time.  
… The ActivityPub community uses Actor objects and uses or relies on this stuff -- and we have a separate spec for this and the name isn't quite right to match things up with other communities that want to share.  
… I think the main point here is that "Controller Document" is a bit confusing and it's hard to see that a "controller document" and a "DID document" and an "actor object" refer to the same concept.  
… This wouldn't be an issue if we didn't have to publish the controller document spec before we publish VCDM. Let's bikeshed in the PR, not today on the call.  

**Brent Zundel:** Thank you, Manu, thanks for the background.  
… Just noting, 4 approvals, request for changes from Joe to just keep calling it that.  

**Joe Andrieu:** One is -- I think "Entity Document" loses the semantic relationship to the VC and secondly, I don't think DID documents are the same as Actor objects, they are not about the subject but about how to interact with its identifier.  
… The binding between the identifier and the entity is the hard part in this work. A DID doc works even if you don't have that binding clear.  
… I think the naming here can be confusing / conflating in the marketplace.  

**Brent Zundel:** Others make comments in the PR, please, thank you folks.  

</details>

TallTed:
> [@jandrieu] What's in the DID Document is what we need to interact with the identifier.

I do not believe we interact with the identifier, which is a URI/IRI.

I believe we interact with the entity (a/k/a resource) denoted by the identifier, which may be almost anything (human, fictional, notional, software application, etc.), which I think makes the need for a DID Document clearer.

msporny:
@selfissued wrote:
> Controller document at least expresses the semantic relationship of that document to the VC and avoids the layer conflation of thinking of a DID Document as a vehicle for statements about entities.

The DID specification never defined the term "controller document". It defined "[DID Document](https://www.w3.org/TR/did-core/#dfn-did-documents)", which then became "Controller Document" when we wanted to convey that it wasn't necessary to use a DID to identify the subject in such a document. 

That is, we wanted to convey that one could any URL, it didn't have to be a DID. When the VC v2.0 work started, we kept sort of lazily using the "Controller Document" terminology without really defining it, and now that we're starting to define it, it's becoming apparent (at least to me) that we picked the term based on tribal knowledge and sloppiness that's going to bite us down the road. 

It's not easy for someone new to the space to understand what a "Controller Document" is, just from the name alone, nor is "Controlling" central to the purpose of these documents (across all use cases). The purpose of the documents is as a general mechanism to describe properties of an entity (real or fictional) -- most notably cryptographic material associated with them and services that you can use to communicate with them... what entity controls cryptographic material is just ONE use case among many.

msporny:
@jandrieu wrote:
> Controller document at least expresses the semantic relationship of that document to the VC 

That's a stretch... I challenge the notion that anyone reading about a "Controller document" would understand it to mean that the "identifier controls the VC". In fact, there are cases where that's a dangerous assumption to make (you can have keys listed in a "Controller document" that /are not/ controlled by the identifier listed (expired keys kept around for historical purposes, references to keys that are controlled by other entities, and so on). There are also service descriptions in DID Documents that have nothing to do w/ "control", they are statements about the entity identified by the DID.

My point here is that there are entries that don't have anything to do with a "Controller" in these things that we're calling "Controller Documents".

> avoids the layer conflation of thinking of a DID Document as a vehicle for statements about entities.

The DID WG defined DID Document as this:

> DID document
> A set of data describing the [DID subject](https://www.w3.org/TR/did-core/#dfn-did-subjects), including mechanisms, such as cryptographic public keys, that the [DID subject](https://www.w3.org/TR/did-core/#dfn-did-subjects) or a [DID delegate](https://www.w3.org/TR/did-core/#dfn-did-delegate) can use to [authenticate](https://www.w3.org/TR/did-core/#dfn-authenticated) itself and prove its association with the [DID](https://www.w3.org/TR/did-core/#dfn-decentralized-identifiers).

Note: "describing the DID subject" (general purpose nature) and "including mechanisms" (not exclusive).

and then this:

> DID subject
> The entity identified by a [DID](https://www.w3.org/TR/did-core/#dfn-decentralized-identifiers) and described by a [DID document](https://www.w3.org/TR/did-core/#dfn-did-documents). Anything can be a DID subject: person, group, organization, physical thing, digital thing, logical thing, etc.

again, "The ***entity*** identified by" ... and then we highlight things like this:

> A [DID](https://www.w3.org/TR/did-core/#dfn-decentralized-identifiers) is an identifier assigned by a [DID controller](https://www.w3.org/TR/did-core/#dfn-did-controllers) to refer to a [DID subject](https://www.w3.org/TR/did-core/#dfn-did-subjects) and resolve to a [DID document](https://www.w3.org/TR/did-core/#dfn-did-documents) that describes the [DID subject](https://www.w3.org/TR/did-core/#dfn-did-subjects).

"that describes the DID subject". 

For all of those reasons, I don't buy this notion that a DID Document is /only/ about "what we need to interact with the identifier". 

> The expectation that the identifier is consistently referring to some entity is a goal and/or assumption. It is not a fact. 

It's a fact per the data model we're using. :)

Identifiers refer to entities with independent existence. A thing with independent existence is often called an "entity" in the English language.

> avoids the layer conflation of thinking of a DID Document as a vehicle for statements about entities.

I get that we don't want to convey that people should fill these documents with a bunch of arbitrary statements and then put them on a blockchain for full view in the public. I think that's partly where you're coming from (if I'm reading between the lines). I do agree with that notion of that as a best practice.

However, we are also dealing w/ an open world data model, and we cannot prohibit people from using things other than VCs to convey this sort of information. It's also true that VCs are unnecessary in some of these cases, such as ActivityPub Actor objects, where the entity is fine w/ expressing the information about how to refer to them in an "Entity Document"/Actor object, and in how to get in touch with them being public information.

At this point, I think I've convinced myself that moving forward with "Controller Document" is worth objecting to... let's find a better word (it doesn't have to be Entity, but it shouldn't be 'Controller' either).

In order to buy us some breathing room to deal w/ this issue (and not create a fire drill right before CR), we could keep the reference to "controller document" and point it at Data Integrity (which we already point to normatively and is already CR), with an explicit "AT RISK" marker stating that the WG is going to change the reference to a separate "controller document" specification in the future (and the construction of that document isn't expected to change) and is debating what to name that concept (or we might just stay w/ the current name). 

What the above paragraph enables us to do is go into CR and make the change later while keeping ***A*** normative reference to what a controller document should look like (it's basically what already in the VC-JOSE-COSE and DI specs). I expect that attempting to go to PR with the normative reference to Data Integrity for controller documents would fail until we button up this issue and publish a separate spec w/ references updated to that spec.

Thoughts?

jandrieu:
> > [@jandrieu] What's in the DID Document is what we need to interact with the identifier.
> 
> I do not believe we interact with the identifier, which is a URI/IRI.

That's fair. "Interact with the identifier" leaves a lot to be defined. When we see the identifier (in a VC, in a VP, in whatever) we then use that identifier to get cryptographic material, which we interact with. Some people would accept that you interact with a URI by dereferencing it, and prior to that, resolving a domain if present. This is the sense that I mean we interact with the identifier: we use it to verify some action taken with respect to that identifier.

> I believe we interact with the entity (a/k/a resource) denoted by the identifier, which may be almost anything (human, fictional, notional, software application, etc.), which I think makes the need for a DID Document clearer.

This is the fundamental flaw in many mental models and one I think keeps pulling us into layer violations. We conflate two things: (1) the mechanisms related to the identifier and (2) interactions with the entity allegedly referred to by that identifier. That latter acknowledges the inherent gap between what we can do with an identifier and what we can *know* about the referent. DIDs allow us to bootstrap strong notions of an entity, either through its observable, verifiable actions taken in its name, or through attestations about that entity using that identifier as subject. However, the DID (and DID infrastructure) cannot itself distinguish between the entity that controls a DID Document and the entity that is referred to by that DID Documents DID. Indeed, there is no guarantee or even conformant requirement that the DID only ever be used to refer to a single entity for all time. 

All you can know is that you are verifying actions of verification relationships as defined by the controller of the DID Document, as taken at a particular point in time.

There is no guarantee *at all* about whether or not the controller of the DID Document is the entity denoted by the identifier.

In fact, when I use a DID that refers to my dog, any processing of DID actions (attestations, encryption, invocations, delegations, and authentication) you are NEVER interacting with the dog. The dog can't attest, encrypt, invoke, delegate, nor authenticate. You're interacting with me, the controller of the DID Document. This is a vital distinction if you are designing systems that depend on interacting with the dog. You cannot assume that the dog you think you are interacting with is the same dog that the verifiable action was about, e.g., a VC of a rabies vaccination of the dog. You need to have additional mechanisms, such as a confidenceMethod to convince yourself that the dog in front of you is the one who received the vaccination, rather than just taking the word of the DID Controller that this dog is the same dog they took to the vet last year.

jandrieu:
> @jandrieu wrote:
> 
> > Controller document at least expresses the semantic relationship of that document to the VC
> 
> That's a stretch... I challenge the notion that anyone reading about a "Controller document" would understand it to mean that the "identifier controls the VC". In fact, there are cases where that's a dangerous assumption to make (you can have keys listed in a "Controller document" that /are not/ controlled by the identifier listed (expired keys kept around for historical purposes, references to keys that are controlled by other entities, and so on). There are also service descriptions in DID Documents that have nothing to do w/ "control", they are statements about the entity identified by the DID.

Whether or not people understand it isn't a good argument. It's our job to explain what is happening here. And what is happening is we are saying that this document is the one that should be used to get the cryptographic material for verifying actions taken on behalf of that identifier, especially with regard to attestation verification. "Controller" is a better approximation of that semantic than "Entity". Although, to be fair, Controller has its own semantic problems, as the property, the role, and the document all have subtly different meanings.

I disagree that statements of service endpoints are about the entity defined by the DID. Rather they are about mechanisms which use can use to interact with SOMETHING in relation to the DID. The endpoint may not even be related to the DID, it could just be put there by the controller because it is convenient.

In an ideal realization, service endpoints would be services where you could engage in verifiable interactions with regard to the DID. We don't have that requirement, but that's largely because there are no standard service paradigms that guarantee verifiability.

> My point here is that there are entries that don't have anything to do with a "Controller" in these things that we're calling "Controller Documents".

To the extent that is true, that's a layer violation I've been fighting from the inception of DIDs. We have VCs for statements about entities. What goes in a DID document is what you need to verify interactions relative to that identifier. Placing statements about the entity in the DID document, such as a name, a state of being (corporation/person/animal), or biometrics, are huge privacy and confidentiality problems. Once you start jamming statements about the entity into the DID document, you lose the flexibility and guarantees of the verification relationships that make DIDs DIDs.

> > avoids the layer conflation of thinking of a DID Document as a vehicle for statements about entities.
> 
> The DID WG defined DID Document as this:
> 
> > DID document
> > A set of data describing the [DID subject](https://www.w3.org/TR/did-core/#dfn-did-subjects), including mechanisms, such as cryptographic public keys, that the [DID subject](https://www.w3.org/TR/did-core/#dfn-did-subjects) or a [DID delegate](https://www.w3.org/TR/did-core/#dfn-did-delegate) can use to [authenticate](https://www.w3.org/TR/did-core/#dfn-authenticated) itself and prove its association with the [DID](https://www.w3.org/TR/did-core/#dfn-decentralized-identifiers).
> 
> Note: "describing the DID subject" (general purpose nature) and "including mechanisms" (not exclusive).

Yes. This is a layer violation that I regularly spoke up about. It's still wrong. 

In fact, that language is what has led you to propose this exaggeration of the conflation.

Controller documents describe how control is verified.

They do not describe the controller.

That's the simple explanation of what is going on here.

If you want other options, consider Verification Document or Verifying Document. As in, this document is how you verify actions taken using one of the verification relationships specified therein.

> and then this:
> 
> > DID subject
> > The entity identified by a [DID](https://www.w3.org/TR/did-core/#dfn-decentralized-identifiers) and described by a [DID document](https://www.w3.org/TR/did-core/#dfn-did-documents). Anything can be a DID subject: person, group, organization, physical thing, digital thing, logical thing, etc.
> 
> again, "The _**entity**_ identified by" ... and then we highlight things like this:
> 
> > A [DID](https://www.w3.org/TR/did-core/#dfn-decentralized-identifiers) is an identifier assigned by a [DID controller](https://www.w3.org/TR/did-core/#dfn-did-controllers) to refer to a [DID subject](https://www.w3.org/TR/did-core/#dfn-did-subjects) and resolve to a [DID document](https://www.w3.org/TR/did-core/#dfn-did-documents) that describes the [DID subject](https://www.w3.org/TR/did-core/#dfn-did-subjects).
> 
> "that describes the DID subject".
> 
> For all of those reasons, I don't buy this notion that a DID Document is /only/ about "what we need to interact with the identifier".

Yes. The spec unfortunately says all of those things. I spoke against it at the time and my concerns were minimally addressed. The layer error is still there. And I'll keep pointing it out.

Statements about entities belong in VCs. Statements about how to verifiably interact with regard to that identifier belong in the DID Document.

> > The expectation that the identifier is consistently referring to some entity is a goal and/or assumption. It is not a fact.
> 
> It's a fact per the data model we're using. :)

No. It isn't.  There is no normative language that states the subject of a DID is the subject FOREVER. 

In fact, that was one of the initial four points defining DIDs. And it was as wrong then as it is now. My opposition to that first principle of "persistence" as initially stated was one of the most common corrections I repeatedly made when Drummond and Markus spoke of the permanence of the link between DIDs and their subject. We corrected that in the use case documents and tried to promote that distinction through the rest of the work.

To the exact extent that DIDs represent some permanent and unappealable referent to an entity, they are a delusion without guarantee. The nature of symbols is the separation of the signifier and the signified. Assurance of the binding between an identifier and its referent is the hard part, not the given part. 

To the extent this DIDs are using in that manner, DIDs represent fundamental PII because in fact, that permanence would be a privacy and confidentiality nightmare for any entity. It is the fact that entities may have any number of different DIDs and DIDs can be used to refer to different things at different times that gets us out of GDPR jail.

You can't even be sure that the DID refers to the same entity over very short periods of time. See my description of DIDs for dogs in my previous comment. https://github.com/w3c/vc-data-model/pull/1396#issuecomment-1875706741

Perhaps more importantly, I can issue a VC using a DID that *to me* referred to President Biden as a distinct human being, Joseph Robinette Biden Jr. Someone else could misinterpret that referent as referring to the President of the United States, regardless of which human being holds that office, and issue a VC with that same subject DID.

Now, which referent is the correct one? You cannot know with just the mechanism of DIDs. I could have misinterpreted the intention of the DID controller when I issued my VC. Maybe I'm wrong. It's a fundamental boundary of knowability. The semantic drift is not resolved by DIDs, it just makes it more tractable by giving mechanisms of verification based on objective math rather than subjective assertions by corruptible and fallible authorities. In fact, the linguistic answer to these kinds of ambiguity is simply more communication to attempt to resolve any perceived ambiguity. "Wait, that statements seems be be about the office of the president, not  Joe Biden himself. Is that what you meant?"

We don't get around these semiotic challenges just because we can verify certain actions taken with regard to an identifier.

> Identifiers refer to entities with independent existence. A thing with independent existence is often called an "entity" in the English language.
> 
> > avoids the layer conflation of thinking of a DID Document as a vehicle for statements about entities.
> 
> I get that we don't want to convey that people should fill these documents with a bunch of arbitrary statements and then put them on a blockchain for full view in the public. I think that's partly where you're coming from (if I'm reading between the lines). I do agree with that notion of that as a best practice.

That's right. Statements about entities belong in VCs, not in controller documents.

> However, we are also dealing w/ an open world data model, and we cannot prohibit people from using things other than VCs to convey this sort of information. It's also true that VCs are unnecessary in some of these cases, such as ActivityPub Actor objects, where the entity is fine w/ expressing the information about how to refer to them in an "Entity Document"/Actor object, and in how to get in touch with them being public information.

Respectfully, we are not just dealing with an open world data model. We are defining a standard for people to create interoperable implementations.  If all we cared about were open world modeling, we wouldn't have ANY conformant requirements beyond JSON-LD. But, in fact, our conversation is precisely to discover and document specific features that we can all use to satisfy our use cases.

Yes, anyone *can* add anything they want to any JSON-LD document, including statements about the entity, or about anything really: the weather on the day it was issued, the current president of the school that controls the DID, etc.

But exactly to the extent that non-standard terms are added to a DID Document is the extent that that DID document lacks interoperability with conformant DID software. 

> At this point, I think I've convinced myself that moving forward with "Controller Document" is worth objecting to... let's find a better word (it doesn't have to be Entity, but it shouldn't be 'Controller' either).

Sure. As long as we find something that captures the semantic about the relationship between that Document and the VC, rather than attempting to define that Document as being about Entities. Because that is NOT what DID Documents are and if we are extending that paradigm, we should get it right without continue the layer violations of unverifiable attestations in DID Documents.

> In order to buy us some breathing room to deal w/ this issue (and not create a fire drill right before CR), we could keep the reference to "controller document" and point it at Data Integrity (which we already point to normatively and is already CR), with an explicit "AT RISK" marker stating that the WG is going to change the reference to a separate "controller document" specification in the future (and the construction of that document isn't expected to change) and is debating what to name that concept (or we might just stay w/ the current name).

If we are committed to a new specification, this seems like a reasonable step.

> What the above paragraph enables us to do is go into CR and make the change later while keeping _**A**_ normative reference to what a controller document should look like (it's basically what already in the VC-JOSE-COSE and DI specs). I expect that attempting to go to PR with the normative reference to Data Integrity for controller documents would fail until we button up this issue and publish a separate spec w/ references updated to that spec.
> 
> Thoughts?

My biggest concern is that we are basically trying to quickfire make a new specification for something called a "TBD" Document. That is not a short piece of work. Precisely because all of these issues the DID WG created with conflation about attributes in the DID document, we will have significant work to clean that up. I, for one, am not willing to exacerbate the errors of DID Documents without a legitimate attempt to resolve them. That legitimate attempt probably does not fit into our timeline.

In hindsight, we probably should not have adopted this strategy at this stage. This is a completely new specification suggested AFTER feature freeze, which we should honor by not creating new conformance requirements. Adding a new deliverable at this stage feels more like a poison pill for the work than a quick fix for an easy problem.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-01-03-vcwg#section1-7) on 2024-01-03

- no resolutions were taken

#### 1.7. Rename "Controller Document" to "Entity Document". (pr vc-data-model#1396)

_See github pull request [vc-data-model#1396](https://github.com/w3c/vc-data-model/pull/1396)._




**Brent Zundel:** next PR: 1396 - rename "controller document" to "entity document".  
… a request for changes from Joe. conversation has continued back and forth. does not look like we have consensus. happy to take comments before marking pending close.  

**Manu Sporny:** clearly disagreement on this item. the problem this creates is : what is the short name we publish, if we need to before VCDM v2 goes to CR. today, VCDM depends normatively on the controller document spec. we broke this out from DI. have yet to do FPWD with it.  
… need to do at least 1 of 2 things. 1 - figure out the new short name, publish as a FPWD on the TR track for controller document. feels like a tall order.  
… OR in the VCDM spec we could update the references to controller document -> data integrity with an AT RISK issue marker, noting we are publishing a new controller document spec.  
… that allows us some breathing room, avoids debate for now. will not need to publish controller document as FPWD before taking VCDM before CR. easiest thing is to update refs and note they will change in the future.  
… seems more difficult to have this debate which could take a few weeks, delay VCDM to CR for weeks or more.  

**Brent Zundel:** proposed by Manu, update refs to the controller doc spec to point to the current text in the DI spec. with a note that we will update them.  

**Joe Andrieu:** concerns: on semantics, feel entity name lost the relationship b/w that document and the VC. and introduces new semantics which I think are incorrect. could take a while to hash out.  
… more concerned about time. introduced this after the VCDM feature freeze. think the language itself is good as long as it has a "we may not get to it" and can still publish the VCDM even if we don't get closer on a new controller document spec.  

**Brent Zundel:** my understanding that is the path we are on.  

**Michael Jones:** I feel like I'm echoing Orie, but we shouldn't be creating normative deps on the securing specs from the VCDM. full stop.  

> *Gabe Cohen:* +1 from me.

**Manu Sporny:** we already have normative dependencies. has been true for a while now. this is not new.  

**Michael Jones:** not sure what you're referring to. should not exacerbate the problem.  

**Manu Sporny:** how does this exacerbate the problem? current controller doc is an exact copy+paste of what's in the DI spec. there's no difference in content currently. we would be adding text to outline our intention.  
… e.g. point here today, will point somewhere else tomorrow. we don't want to delay. could take another month to get the controller doc spec in shape. don't see how this exacerbates the issue. helps us get VCDM v2 in CR sooner than later.  

**Michael Jones:** exacerbates the problem by taking a normative dep on the controller document language in the DI spec. talked about this 2 months ago. agreed to break this out with you to a new spec. recognizing not everything in the DI language will survive.  
… if you take a tentative decision, someone will say "we can't reconsider this" - so not willing to do that.  

**Brent Zundel:** I don't believe it matters what we call this...it matters...but what we're pointing to. we're pointing to something with language specifying a document specifying properties. we've been calling it controller document, maybe should be called something else. what we call it matters a bit, but really not much.  
… as long as it's not the jelly bean spec! (might be more fun, though). right now the concern is we want to get to CR with the core data model. we are pointing normatively to a thing that's not a thing.  
… either need to make that thing a thing as fast as possible. or point to something else, fully intending to make that thing a thing.  

**Manu Sporny:** position I'm arguing for is your second option. "we intend to make this thing a thing ASAP" - express our intent and move forward. 3rd option is to point to DID Core, which is a REC.  
… have looked at references we have. we talk about very similar things. each of these sentences could point to DID Core, with a caveat that these fields can be URLs, not DIDs. that's the only difference we're talking about here.  

**Michael Jones:** option, where I thought we were, wasn't enumerated. there is language in each of the securing specs. we will leave it there until we pass CR. then we will pass CR and harmonize. I thought we were at a place where both security docs define this. no need in the core data model.  

**Manu Sporny:** no that's not where we are Mike. if you are OK with that...we can go in and modify where we reference controller doc in the spec and point to "as defined in VC JOSE COSE or VC DATA INTEGRITY" - that's aligned with what you said, and I'm fine with that.  

**Michael Jones:** yes, thank you. thought that's where we landed.  

**Joe Andrieu:** I'd like to see the aperture of that defined as "in securing specs..." to not exclude other securing specs.  

**Manu Sporny:** I will raise a PR and Joe can try to modify the language.  

**Joe Andrieu:** perfect.  

**Manu Sporny:** just 3 instance with this language. then we are clear to go into CR rapidly.  
… will raise a PR in lieu of 1396 that attempts this.  

</details>

selfissued (changes requested):
I don't love the term Controller Document, but it's the term we have across multiple specifications and in people's minds.  It doesn't seem to me that changing it to "Entity Document", which also isn't a great term, is an improvement worth the costs of the change.

msporny:
This PR has been updated to implement the consensus position we reached [during the last telecon](https://github.com/w3c/vc-data-model/pull/1396#issuecomment-1876565728), which was to:

1. NOT rename the concept of controller documents and leave the terminology alone for now.
2. Update the references to controller documents and state that either definition in VC-JOSE-COSE or VC-DATA-INTEGRITY can be used.

This PR now does that.

msporny:
Normative, multiple reviews, changes requested and made, all objections addressed, consensus position of group applied, merging.

PR #1413, opened 2024-01-17
Add intro to Credential Type-Specific Processing.

BigBlueHat:
This pulls from the text highlighted via the following link from PR #1302. It also brings in the pending suggestions made by @TallTed to that highlighted text:
https://github.com/w3c/vc-data-model/pull/1302/files#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051R4251-R4266

***
<a href="https://pr-preview.s3.amazonaws.com/BigBlueHat/vc-data-model/pull/1413.html" title="Last updated on Feb 10, 2024, 8:30 PM UTC (34f40bc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1413/2980929...BigBlueHat:34f40bc.html" title="Last updated on Feb 10, 2024, 8:30 PM UTC (34f40bc)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4482,7 +4482,7 @@ content type, or return an error code such as
           <p>
 RDF [[?RDF11-CONCEPTS]] is a technology for modeling graphs of statements. Each statement is a
 single <i>subject&#x2192;property&#x2192;value</i> (also known as
-<i>entity&#x2192;attribute&#x2192;value</i>) relationship, which is referred to
+<i>entity→attribute→value</i>) relationship, which is referred to
 as a <a>claim</a> in this specification. JSON-LD [[?JSON-LD11]] is a technology that enables
 the expression of RDF using idiomatic JSON, enabling developers familiar with
 JSON to write applications that consume RDF as JSON. In general, subjects are
--- a/index.html
+++ b/index.html
@@ -4181,8 +4181,8 @@ or an array of values.
         <p>
 [[!JSON-LD11]] is a JSON-based format used to serialize
 <a href="https://www.w3.org/TR/ld-glossary/#linked-data">Linked Data</a>. The
-syntax is designed to easily integrate into deployed systems already using JSON,
-and provides a smooth upgrade path from JSON to [[!JSON-LD11]]. It is primarily
+syntax is designed to easily integrate into deployed systems already using JSON.
+It is primarily
 intended to be a way to use Linked Data in Web-based programming environments,
 to build interoperable Web services, and to store Linked Data in JSON-based
 storage engines.
--- a/index.html
+++ b/index.html
@@ -4480,26 +4480,26 @@ content type, or return an error code such as
           <h2>Credential Type-Specific Processing</h2>
 
           <p>
-RDF is a technology for modeling graphs of statements. Each statement is a
+RDF [[?RDF11-CONCEPTS]] is a technology for modeling graphs of statements. Each statement is a
 single <i>subject&#x2192;property&#x2192;value</i> (also known as
 <i>entity&#x2192;attribute&#x2192;value</i>) relationship, which is referred to
-as a <a>claim</a> in this specification. JSON-LD is a technology that enables
+as a <a>claim</a> in this specification. JSON-LD [[?JSON-LD11]] is a technology that enables
 the expression of RDF using idiomatic JSON, enabling developers familiar with
-JSON to write applications to consume RDF as JSON. In general, subjects are
+JSON to write applications that consume RDF as JSON. In general, subjects are
 expressed as JSON objects with each property and value of the subject as a JSON
 key and value, respectively. A special affordance is made to express an
-identifier of a subject, if present, using the <code>@id</code> JSON key, which
-is aliased in this specification to <code>id</code>. See
-<a data-cite="?JSON-LD11#relationship-to-rdf">Relationship to RDF</a> for more
-details.
+identifier of a subject, if necessary, using the <code>id</code> key in this
+specification. See
+<a data-cite="?JSON-LD11#relationship-to-rdf">Relationship of JSON-LD to RDF</a>
+for more details.
           </p>
 
           <p>
-As JSON can be used to express all different kinds of information, a consumer of
-a particular JSON document can only properly interpret the authors intent if they
+As JSON can be used to express different kinds of information, a consumer of
+a particular JSON document can only properly interpret the author's intent if they
 possess information that contextualizes and disambiguates it from other possible
 expressions. Information to assist with this interpretation can either be wholly
-external to the JSON document or linked from within it. Compact JSON-LD documents
+external to the JSON document or linked from within it. Compacted JSON-LD documents
 include a <code>@context</code> property that internally expresses or links to
 contextual information and, per the JSON-LD specification, these documents follow
 object-oriented organizational principles to express <a>claims</a>. These features
@@ -4516,10 +4516,10 @@ The text below helps consumers understand how to ensure a JSON-LD document is
 expressed in a context and shape that their application already understands
 such that they do not need to transform it in order to consume its contents.
 Notably, this does not mean that consumers do not need to understand any
-context at all, rather, consuming applications only need to understand a chosen
+context at all; rather, consuming applications only need to understand a chosen
 set of contexts and document shapes to work with and not others. Issuers can
 publish contexts and information about their <a>verifiable credentials</a> to
-aid consumers who do not use generalized processors, just like would be done
+aid consumers who do not use generalized processors, just as can be done
 with any other JSON-formatted data.
           </p>
 
--- a/index.html
+++ b/index.html
@@ -4191,7 +4191,7 @@ storage engines.
         <p>
 Linked Data is modeled using Resource Description Framework (RDF).
 RDF [[?RDF11-CONCEPTS]] is a technology for modeling graphs of statements. Each statement is a
-single <i>subject&#x2192;property&#x2192;value</i> (also known as
+single <i>subject→property→value</i> (also known as
 <i>entity→attribute→value</i>) relationship, which is referred to
 as a <a>claim</a> in this specification. JSON-LD [[?JSON-LD11]] is a technology that enables
 the expression of RDF using idiomatic JSON, enabling developers familiar with
--- a/index.html
+++ b/index.html
@@ -4180,18 +4180,10 @@ or an array of values.
 
         <p>
 [[!JSON-LD11]] is a JSON-based format used to serialize
-<a href="https://www.w3.org/TR/ld-glossary/#linked-data">Linked Data</a>. The
-syntax is designed to easily integrate into deployed systems already using JSON.
-It is primarily
-intended to be a way to use Linked Data in Web-based programming environments,
-to build interoperable Web services, and to store Linked Data in JSON-based
-storage engines.
-        </p>
-
-        <p>
+<a href="https://www.w3.org/TR/ld-glossary/#linked-data">Linked Data</a>.
 Linked Data is modeled using Resource Description Framework (RDF).
 RDF [[?RDF11-CONCEPTS]] is a technology for modeling graphs of statements. Each statement is a
-single <i>subject→property→value</i> (also known as
+single <i>subject&#x2192;property&#x2192;value</i> (also known as
 <i>entity→attribute→value</i>) relationship, which is referred to
 as a <a>claim</a> in this specification. JSON-LD [[?JSON-LD11]] is a technology that enables
 the expression of RDF using idiomatic JSON, enabling developers familiar with
--- a/index.html
+++ b/index.html
@@ -4480,6 +4480,50 @@ content type, or return an error code such as
           <h2>Credential Type-Specific Processing</h2>
 
           <p>
+RDF is a technology for modeling graphs of statements. Each statement is a
+single <i>subject&#x2192;property&#x2192;value</i> (also known as
+<i>entity&#x2192;attribute&#x2192;value</i>) relationship, which is referred to
+as a <a>claim</a> in this specification. JSON-LD is a technology that enables
+the expression of RDF using idiomatic JSON, enabling developers familiar with
+JSON to write applications to consume RDF as JSON. In general, subjects are
+expressed as JSON objects with each property and value of the subject as a JSON
+key and value, respectively. A special affordance is made to express an
+identifier of a subject, if present, using the <code>@id</code> JSON key, which
+is aliased in this specification to <code>id</code>. See
+<a data-cite="?JSON-LD11#relationship-to-rdf">Relationship to RDF</a> for more
+details.
+          </p>
+
+          <p>
+As JSON can be used to express all different kinds of information, a consumer of
+a particular JSON document can only properly interpret the authors intent if they
+possess information that contextualizes and disambiguates it from other possible
+expressions. Information to assist with this interpretation can either be wholly
+external to the JSON document or linked from within it. Compact JSON-LD documents
+include a <code>@context</code> property that internally expresses or links to
+contextual information and, per the JSON-LD specification, these documents follow
+object-oriented organizational principles to express <a>claims</a>. These features
+enable generalized processors to be written to convert JSON-LD documents from one
+context to another, but this is not needed when consumers receive JSON-LD documents
+that already use the context and shape that they expect. Authors of JSON-LD
+documents, such as <a>issuers</a> of <a>verifiable credentials</a>, are required
+to provide proper JSON-LD contexts and follow these rules in order to facilitate
+interoperability.
+          </p>
+
+          <p>
+The text below helps consumers understand how to ensure a JSON-LD document is
+expressed in a context and shape that their application already understands
+such that they do not need to transform it in order to consume its contents.
+Notably, this does not mean that consumers do not need to understand any
+context at all, rather, consuming applications only need to understand a chosen
+set of contexts and document shapes to work with and not others. Issuers can
+publish contexts and information about their <a>verifiable credentials</a> to
+aid consumers who do not use generalized processors, just like would be done
+with any other JSON-formatted data.
+          </p>
+
+          <p>
 <dfn>General JSON-LD processing</dfn> is defined as a mechanism that utilizes a
 JSON-LD software library to process a [=conforming document=] by performing
 various <a data-cite="?JSON-LD11#forms-of-json-ld">transformations</a>.
--- a/index.html
+++ b/index.html
@@ -4501,8 +4501,7 @@ possess information that contextualizes and disambiguates it from other possible
 expressions. Information to assist with this interpretation can either be wholly
 external to the JSON document or linked from within it. Compacted JSON-LD documents
 include a <code>@context</code> property that internally expresses or links to
-contextual information and, per the JSON-LD specification, these documents follow
-object-oriented organizational principles to express <a>claims</a>. These features
+contextual information to express <a>claims</a>. These features
 enable generalized processors to be written to convert JSON-LD documents from one
 context to another, but this is not needed when consumers receive JSON-LD documents
 that already use the context and shape that they expect. Authors of JSON-LD
--- a/index.html
+++ b/index.html
@@ -4189,7 +4189,22 @@ storage engines.
         </p>
 
         <p>
-[[!JSON-LD11]] is useful when extending the data model described in this
+Linked Data is modeled using Resource Description Framework (RDF).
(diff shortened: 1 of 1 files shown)


dlongley (approved):

index.html
```diff
@@ -4479,6 +4479,50 @@ <h2>HTTP</h2>
...
+key and value, respectively. A special affordance is made to express an
+identifier of a subject, if present, using the <code>@id</code> JSON key, which
+is aliased in this specification to <code>id</code>. See
+<a data-cite="?JSON-LD11#relationship-to-rdf">Relationship to RDF</a> for more
+details.
+          </p>
+
+          <p>
+As JSON can be used to express all different kinds of information, a consumer of
+a particular JSON document can only properly interpret the authors intent if they
```
TallTed, line 4499:
```suggestion
a particular JSON document can only properly interpret the author's intent if they
```

index.html
```diff
@@ -4479,6 +4479,50 @@ <h2>HTTP</h2>
...
+possess information that contextualizes and disambiguates it from other possible
+expressions. Information to assist with this interpretation can either be wholly
+external to the JSON document or linked from within it. Compact JSON-LD documents
```
TallTed, line 4502:
```suggestion
external to the JSON document or linked from within it. Compacted JSON-LD documents
```

index.html
```diff
@@ -4479,6 +4479,50 @@ <h2>HTTP</h2>
...
+to provide proper JSON-LD contexts and follow these rules in order to facilitate
+interoperability.
+          </p>
+
+          <p>
+The text below helps consumers understand how to ensure a JSON-LD document is
+expressed in a context and shape that their application already understands
+such that they do not need to transform it in order to consume its contents.
+Notably, this does not mean that consumers do not need to understand any
+context at all, rather, consuming applications only need to understand a chosen
```
TallTed, line 4519:
```suggestion
context at all; rather, consuming applications only need to understand a chosen
```

index.html
```diff
@@ -4479,6 +4479,50 @@ <h2>HTTP</h2>
...
+set of contexts and document shapes to work with and not others. Issuers can
+publish contexts and information about their <a>verifiable credentials</a> to
+aid consumers who do not use generalized processors, just like would be done
```
TallTed, line 4522:
```suggestion
aid consumers who do not use generalized processors, just as can be done
```

TallTed (changes requested):
Small stuff

index.html
```diff
@@ -4479,6 +4479,50 @@ <h2>HTTP</h2>
...
+RDF is a technology for modeling graphs of statements. Each statement is a
+single <i>subject&#x2192;property&#x2192;value</i> (also known as
+<i>entity&#x2192;attribute&#x2192;value</i>) relationship, which is referred to
+as a <a>claim</a> in this specification. JSON-LD is a technology that enables
+the expression of RDF using idiomatic JSON, enabling developers familiar with
+JSON to write applications to consume RDF as JSON. In general, subjects are
+expressed as JSON objects with each property and value of the subject as a JSON
+key and value, respectively. A special affordance is made to express an
+identifier of a subject, if present, using the <code>@id</code> JSON key, which
+is aliased in this specification to <code>id</code>. See
```
iherman, line 4492: I wonder whether for the prospective readers of this specification, mentioning `@id` may do more harm than good.
 
```suggestion
identifier of a subject using, if necessary, the <code>id</code> key in this specification. See
```

index.html
```diff
@@ -4479,6 +4479,50 @@ <h2>HTTP</h2>
...
+<a data-cite="?JSON-LD11#relationship-to-rdf">Relationship to RDF</a> for more
```
iherman, line 4493:
```suggestion
<a data-cite="?JSON-LD11#relationship-to-rdf">Relationship of JSON-LD to RDF</a> for more
```

index.html
```diff
@@ -4479,6 +4479,50 @@ <h2>HTTP</h2>
         <section class="informative">
           <h2>Credential Type-Specific Processing</h2>
 
+          <p>
+RDF is a technology for modeling graphs of statements. Each statement is a
+single <i>subject&#x2192;property&#x2192;value</i> (also known as
+<i>entity&#x2192;attribute&#x2192;value</i>) relationship, which is referred to
+as a <a>claim</a> in this specification. JSON-LD is a technology that enables
```
iherman, line 4486: Maybe add references to RDF and JSON-LD?
```suggestion
RDF [[?RDF11-CONCEPTS]] is a technology for modeling graphs of statements. Each statement is a
single <i>subject&#x2192;property&#x2192;value</i> (also known as
<i>entity&#x2192;attribute&#x2192;value</i>) relationship, which is referred to
as a <a>claim</a> in this specification. JSON-LD [[?JSON-LD11]] is a technology that enables
```

index.html
```diff
@@ -4479,6 +4479,50 @@ <h2>HTTP</h2>
...
+expressed as JSON objects with each property and value of the subject as a JSON
+key and value, respectively. A special affordance is made to express an
+identifier of a subject, if present, using the <code>@id</code> JSON key, which
+is aliased in this specification to <code>id</code>. See
+<a data-cite="?JSON-LD11#relationship-to-rdf">Relationship to RDF</a> for more
+details.
+          </p>
+
+          <p>
+As JSON can be used to express all different kinds of information, a consumer of
```
iherman, line 4498:
```suggestion
As JSON can be used to express different kinds of information, a consumer of
```

index.html
```diff
@@ -4479,6 +4479,50 @@ <h2>HTTP</h2>
...
+a particular JSON document can only properly interpret the authors intent if they
+possess information that contextualizes and disambiguates it from other possible
+expressions. Information to assist with this interpretation can either be wholly
+external to the JSON document or linked from within it. Compact JSON-LD documents
+include a <code>@context</code> property that internally expresses or links to
+contextual information and, per the JSON-LD specification, these documents follow
+object-oriented organizational principles to express <a>claims</a>. These features
```
iherman, line 4505: I am not sure what "object-oriented organizational principles" means. More exactly... I of course know that we refer to issues like using `@protected`, local `@context`, etc., but these may be meaningless for a casual reader, who may associate these terms to programming language classes, their class inheritance, properties, etc. 

I think this side remark can be safely removed.

```suggestion
contextual information to express the correct <a>claims</a>. These features
```

index.html
```diff
@@ -4479,6 +4479,50 @@ <h2>HTTP</h2>
...
+enable generalized processors to be written to convert JSON-LD documents from one
+context to another, but this is not needed when consumers receive JSON-LD documents
+that already use the context and shape that they expect. Authors of JSON-LD
```
iherman, line 4508: The whole issue/information about generalized processors converting JSON-LD documents seems superfluous to me altogether. Average VC application developers do not even understand what you refer to at this point. I believe the whole sentence can be safely deleted.

index.html
BigBlueHat: I'm fine to take it out, but I think the original intent was to highlight in this context what RDF is providing "under-the-hood". @dlongley iirc, noting this was your suggestion, so your input here would be helpful before I take it out. 😃

iherman: Yeah, the question is who the target is for this. My assumption is that the target is those who do not really know and, actually, do not really care about the RDF aspect, just want to get things work for them without too many questions asked. For an audience who know RDF & Co, the whole text is, I believe, unnecessary anyway...

dlongley: I think this is good text to help people understand that statements aren't expressed in either a flat or bespoke format, but JSON objects really are used to express objects (things) where each JSON key is a property and value of said thing. You can design your JSON in totally different ways from this where no information can be gleaned from the JSON structure alone -- but JSON-LD requires it to work in a way that is true to this object oriented nature.

So I'd prefer to keep it; I don't think we come out and say this anywhere else and it's helpful to understand this overarching principle in how JSON-LD expresses claims.

index.html
iherman: I won't lie down the road on this, it is an informational text anyway.

index.html
iherman: Though not convinced, I won't lie down the road on this.

Can we at least find another formulation? "Object-oriented organizational principles" bothers my ears, to be honest...

BigBlueHat: @iherman I've come around to your way of thinking. I think the current statement is a bit non sequitur as well.

@dlongley if it's important to express the OOP nature of JSON-LD vs. random JSON structures, I think we need to do it elsewhere in this text. Happy to take a crack at it, if you think it's important.

dlongley: If we want to strike that one phrase and add some other text to address it in another way in a follow-on PR, that's fine. I don't want to hold up the other good text here.

msporny (approved):

index.html
```diff
@@ -4479,6 +4479,50 @@ <h2>HTTP</h2>
         <section class="informative">
           <h2>Credential Type-Specific Processing</h2>
 
+          <p>
+RDF [[?RDF11-CONCEPTS]] is a technology for modeling graphs of statements. Each statement is a
+single <i>subject&#x2192;property&#x2192;value</i> (also known as
```
msporny, line 4484: Any particular reason we can't use the `→` character here?

index.html
```diff
@@ -4479,6 +4479,50 @@ <h2>HTTP</h2>
...
+<i>entity&#x2192;attribute&#x2192;value</i>) relationship, which is referred to
+as a <a>claim</a> in this specification. JSON-LD [[?JSON-LD11]] is a technology that enables
+the expression of RDF using idiomatic JSON, enabling developers familiar with
+JSON to write applications that consume RDF as JSON. In general, subjects are
+expressed as JSON objects with each property and value of the subject as a JSON
+key and value, respectively. A special affordance is made to express an
+identifier of a subject, if necessary, using the <code>id</code> key in this
+specification. See
+<a data-cite="?JSON-LD11#relationship-to-rdf">Relationship of JSON-LD to RDF</a>
+for more details.
```
msporny, line 4494: Doesn't some of this repeat stuff already in https://w3c.github.io/vc-data-model/#claims?

Why are we saying that a "special affordance" is made to express an identifier of a subject? We say that here, right? https://w3c.github.io/vc-data-model/#syntactic-sugar

index.html
```diff
@@ -4479,6 +4479,50 @@ <h2>HTTP</h2>
...
+          </p>
```
msporny, line 4495: I'm concerned that this is going to come across as word salad for many readers, who probably don't care about RDF (or have heard bad things about it, some of it completely justified). There are phrases here like: 

* "technology for modeling graphs of statements" -- not experienced by most CS students much less web developers.
* "enables the expression of RDF using idiomatic JSON" 
* "consume RDF as JSON" -- They can read the JSON-LD spec to learn more about the details there...

IOW, I'm concerned that the opening paragraph is going to not land well -- most readers are probably reading this section because they /don't want to care/ about all that stuff. They just want to issue a VC, they don't want to pull in a bunch of unnecessary RDF stuff... or be reminded of it (because, ideally, they don't have to think about the vast majority of the stuff under the hood... just follow the fairly straight-forward instructions in this section.

index.html
msporny: FWIW, I agree with Ivan... it's not clear what "object-oriented organizational principles" means... is it talking about encapsulation? abstractions? design patterns? I suggest just striking ", per the JSON-LD specification, these documents follow object-oriented organizational principles to express <a>claims</a>" (because we already explain that several times throughout the specification, starting in the "Core Data Model" section.

msporny (changes requested):
The first paragraph is a rough entry -- too much RDF / graph/ idiomatic JSON talk -- the PR would probably be fine if the first paragraph was deleted entirely (or re-written to be more approachable). The last two paragraphs are good (modulo the vague "object oriented" statement).

Please rework the first paragraph (or delete) and the "object oriented" statement.

index.html
dlongley: > Why are we saying that a "special affordance" is made to express an identifier of a subject?

Because the preceding sentence might imply to some readers that `id` is a property of a subject rather than being special in that it holds the identifier for the subject instead. We can strike the sentence if we think that readers won't get confused over that.

index.html
dlongley: I think this text provides some brief context (no pun intended) that has been missing in every iteration of the VC data model spec to date. What is says here fills gaps that have continued to come up in questions from various different readers over time. We should keep it.

> IOW, I'm concerned that the opening paragraph is going to not land well -- most readers are probably reading this section because they /don't want to care/ about all that stuff. 

This provides a quick informative summary of what those things are instead of leaving some readers quite confused about why these technologies are used. There are readers who don't care at all (who this text is not for) and there are readers who want to know more but aren't going to read all of the links to the various underlying specs. This text informatively explains (and I think pretty tersely):

RDF lets you model `subject->property->value` statements (oh, that's what "claims" are, so that's why RDF is used). JSON-LD lets you express RDF in JSON (oh, JSON-LD enables RDF to happen in JSON; I know JSON, but how does it look?). Each JSON object is a subject with JSON keys as the properties of the subject and JSON values as the associated values... and the `id` property holds the subject's identifier (oh, so these technologies provide a *standard*, consistent way to express "claims" in JSON, i.e., it is not just "anything goes JSON").

We could explicitly say how this approach is fully compositional (whereas "anything goes JSON" is not), but we can perhaps assume readers will get that -- and we are trying to keep it short.

> ...ideally, they don't have to think about the vast majority of the stuff under the hood...

I agree that it's good when people don't have to know this. However, we also keep not telling the people who do want to know in a concise way -- without forcing them to deep-dive into the other specs. I think people have been asking for this.

dlongley: I *do* think this paragraph feels out of place in the "Credential-Type specific processing" section of the spec. I'd recommend we move this to another subsection under JSON-LD called "Relationship to RDF" so people can read it there.

index.html
TallTed: Presuming we can...
```suggestion
single <i>subject→property→value</i> (also known as
```

index.html
```diff
@@ -4479,6 +4479,50 @@ <h2>HTTP</h2>
         <section class="informative">
           <h2>Credential Type-Specific Processing</h2>
 
+          <p>
+RDF [[?RDF11-CONCEPTS]] is a technology for modeling graphs of statements. Each statement is a
+single <i>subject&#x2192;property&#x2192;value</i> (also known as
+<i>entity&#x2192;attribute&#x2192;value</i>) relationship, which is referred to
```
TallTed, line 4485:
```suggestion
<i>entity→attribute→value</i>) relationship, which is referred to
```

BigBlueHat:
@msporny @dlongley @iherman I move that intro paragraph down into the JSON-LD section. I didn't use a separate section, but was able to find a home for it as the second paragraph there providing more explanation around Linked Data.

Please let me know if you think that's a better fit for it, and I can do another pass (if you want) on smoothing any wrinkles caused by that move.

Thanks!
🎩

index.html
```diff
@@ -4189,7 +4189,22 @@ <h3>JSON-LD</h3>
         </p>
```
iherman, line 4190: (I cannot directly edit the text from github, copying here the problematic part:

> [[!JSON-LD11]] is a JSON-based format used to serialize <a href="https://www.w3.org/TR/ld-glossary/#linked-data">Linked Data</a>. The syntax is designed to easily integrate into deployed systems already using JSON, and provides a smooth upgrade path from JSON to [[!JSON-LD11]].

The 2nd sentence looks a bit strange to me.  It says 

> JSON-LD 'provides a smooth upgrade path from JSON' to… JSON-LD.

I.e., it provides a smooth upgrade to itself... which is a bit circular. My proposal would be to remove the second part of the sentence, to say:

> [[!JSON-LD11]] is a JSON-based format used to serialize <a href="https://www.w3.org/TR/ld-glossary/#linked-data">Linked Data</a>. The syntax is designed to easily integrate into deployed systems already using JSON.

index.html
```diff
@@ -4189,7 +4189,22 @@ <h3>JSON-LD</h3>
...
 
         <p>
-[[!JSON-LD11]] is useful when extending the data model described in this
+Linked Data is modeled using Resource Description Framework (RDF).
+RDF [[?RDF11-CONCEPTS]] is a technology for modeling graphs of statements. Each statement is a
+single <i>subject&#x2192;property&#x2192;value</i> (also known as
```
iherman, line 4194:
```suggestion
single <i>subject→property→value</i> (also known as
```

index.html
```diff
@@ -4189,7 +4189,22 @@ <h3>JSON-LD</h3>
...
+<i>entity→attribute→value</i>) relationship, which is referred to
+as a <a>claim</a> in this specification. JSON-LD [[?JSON-LD11]] is a technology that enables
+the expression of RDF using idiomatic JSON, enabling developers familiar with
+JSON to write applications that consume RDF as JSON. In general, subjects are
```
iherman, line 4198: This sentence seems to repeat what is said in the first paragraph...

(I actually begin to wonder whether the first paragraph should not be removed altogether.)

index.html
```diff
@@ -4189,7 +4189,22 @@ <h3>JSON-LD</h3>
...
+expressed as JSON objects with each property and value of the subject as a JSON
+key and value, respectively. A special affordance is made to express an
+identifier of a subject, if necessary, using the <code>id</code> key in this
+specification. See
+<a data-cite="?JSON-LD11#relationship-to-rdf">Relationship of JSON-LD to RDF</a>
+for more details.
+          </p>
+        <p>
+Consequently, [[!JSON-LD11]] is useful when extending the data model described in this
```
iherman, line 4207:
```suggestion
[[!JSON-LD11]] is useful when extending the data model described in this
```

"Consequently" to what? I think it is unnecessary here.

iherman (approved):
Only minor remarks.

index.html
BigBlueHat: Agreed. It does sound circular. I'll remove that last bit.

index.html
BigBlueHat: I do feel this new "second paragraph" is much clearer about what Linked Data is, what it provides, and how JSON-LD is an implementation of that which empower extensibility. I'm in favor of removing the first paragraph in favor of this new second one.

@msporny @dlongley any objections?

index.html
BigBlueHat: JSON-LD provides all the super-powers described in the preceding paragraph...consequently, it's "useful when extending the data model described in this specification."

If that's not generally clear, though, I can forego the "consequently." 😄

iherman: But this is certainly not a reason to lie down on the road as far as I am concerned... 😀

index.html
msporny: Yes, that should work. Let's try that and see how it reads.

BigBlueHat: Merged the two--kept the first sentence of the first paragraph and removed the rest of the first paragraph:
[409bfcd](https://github.com/w3c/vc-data-model/pull/1413/commits/409bfcdd59c918c9bc2e9cdc142242aef1c0234e)

index.html
iherman: It has been removed from the latest iteration, as far as I can see, so marking this as resolved.

index.html
```diff
@@ -4180,16 +4180,23 @@ <h3>JSON-LD</h3>
...
-<a href="https://www.w3.org/TR/ld-glossary/#linked-data">Linked Data</a>. The
-syntax is designed to easily integrate into deployed systems already using JSON,
-and provides a smooth upgrade path from JSON to [[!JSON-LD11]]. It is primarily
-intended to be a way to use Linked Data in Web-based programming environments,
-to build interoperable Web services, and to store Linked Data in JSON-based
-storage engines.
-        </p>
-
+<a href="https://www.w3.org/TR/ld-glossary/#linked-data">Linked Data</a>.
+Linked Data is modeled using Resource Description Framework (RDF).
```
msporny, line 4184:
```suggestion
Linked Data is modeled using the Resource Description Framework (RDF).
```

index.html
```diff
@@ -4180,16 +4180,23 @@ <h3>JSON-LD</h3>
...
+RDF [[?RDF11-CONCEPTS]] is a technology for modeling graphs of statements. Each statement is a
+single <i>subject&#x2192;property&#x2192;value</i> (also known as
```
msporny, line 4186:
```suggestion
single <i>subject→property→value</i> (also known as
```

index.html
msporny: I agree with @iherman, let's take it out, it read a bit strange to me as well. I get what it's saying, but removing it didn't harm the text, IMHO.

msporny (approved):
Much better, thank you, approving.

msporny:

PR #1416, opened 2024-01-25
Adding SHACL as a credential schema language

clange:
With this PR, I am suggesting to implement the proposal suggested by @bmaier in #1411. Just in the context, for now, not yet in the human-readable specification document.

I am affiliated with W3C member Fraunhofer.

### DIFF
--- a/contexts/credentials/v2
+++ b/contexts/credentials/v2
@@ -193,6 +193,23 @@
       }
     },
 
+    "ShaclSchemaCredential": "https://w3.org/2018/credentials#ShaclSchemaCredential",
+
+    "ShaclSchema": {
+      "@id": "https://w3.org/2018/credentials#ShaclSchema",
+      "@context": {
+        "@protected": true,
+
+        "id": "@id",
+        "type": "@type",
+
+        "shaclSchema": {
+           "@id": "https://w3.org/2018/credentials#shaclSchema",
+           "@type": "@id"
+        }
+      }
+    },
+
     "BitstringStatusListCredential": "https://www.w3.org/ns/credentials/status#BitstringStatusListCredential",
 
     "BitstringStatusList": {


contexts/credentials/v2
```diff
@@ -193,6 +193,23 @@
       }
     },
 
+    "ShaclSchemaCredential": "https://w3.org/2018/credentials#ShaclSchemaCredential",
```
decentralgabe, line 196: I believe this should be added to the [VC Specs Directory](https://w3c.github.io/vc-specs-dir/#credential-schema) before merging this PR to provide additional context on how one would go about using a SHACL schema.

iherman:
The problem I see is that SHACL is not prepared to handle RDF Datasets, only "simple" RDF. On the other hand, the VCDM consists of a number of graphs (in the RDF Dataset sense) see, for example [the graphs in this diagrams](https://www.w3.org/TR/vc-data-model-2.0/#info-graph-vc). In view of this, I am not sure if I see any use of this extra possibility... (unless a new version of SHACL is coming to the fore that handles graphs).

iherman:
(with my admin hat on) @clange, this PR, if accepted, is substantial, and requires to sign off IPR; I believe you have received the references to it. Alternatively, you can join the WG (which you would be welcome to) to represent Fraunhofer...

iherman:
> In view of this, I am not sure if I see any use of this extra possibility... 

To be a bit more positive: SHACL may be used to check a _credential_ but without and embedded proof (the `proof` property refers to a separate graph), and cannot be used for a _presentation_ (the `verifiableCredential` property refers to separate graphs). If this extra facility is accepted, it should be surrounded with warnings as for these limitations.

contexts/credentials/v2
iherman: Additionally, in this form this means extending the official VCDM vocabulary.

Alternatively, a separate URL is used for this property to reflect that this is an extension to the core (if this is the way we go).

clange: Thanks for pointing me to the VC Specs Directory, @decentralgabe. What's the most reasonable way to get this done?
I noticed https://github.com/w3c/vc-specs-dir#adding-new-entries-to-the-directory.

All entries [in this table](https://w3c.github.io/vc-specs-dir/#credential-schema) point to proper specification documents. So, would I have to write such a specification document to explain how a SHACL shapes graph (to use the proper term) should be used to validate a credential? So far, I would have considered that a straightforward application of SHACL's "validation of a focus node \[i.e., the credentialSubject\] against a shape \[for all applicable shapes\]" ([as specified here](https://www.w3.org/TR/shacl/#validation-definition)) to be explained by a few bullet points, but given @iherman's additional comments, maybe that's not the case.

clange:
> (with my admin hat on) @clange, this PR, if accepted, is substantial, and requires to sign off IPR; I believe you have received the references to it. Alternatively, you can join the WG (which you would be welcome to) to represent Fraunhofer...

Not sure – the [validation](https://labs.w3.org/repo-manager/pr/id/w3c/vc-data-model/1416) says I should join the WG, and I'm currently going to Fraunhofer's internal clearance.

clange:
> The problem I see is that SHACL is not prepared to handle RDF Datasets, only "simple" RDF. On the other hand, the VCDM consists of a number of graphs (in the RDF Dataset sense) see, for example [the graphs in this diagrams](https://www.w3.org/TR/vc-data-model-2.0/#info-graph-vc). In view of this, I am not sure if I see any use of this extra possibility... (unless a new version of SHACL is coming to the fore that handles graphs).

What would that diagram look like when serialized, e.g., as JSON-LD? When seeing it in JSON-LD (or Turtle or whatever RDF serialization), I would more easily be able to wrap my head around it and think about what a SHACL validation might look like.

And how would the same thing be validated against a JSON Schema? Wouldn't there also be (different) pitfalls in that case?

clange:
> > In view of this, I am not sure if I see any use of this extra possibility...
> 
> To be a bit more positive: SHACL may be used to check a _credential_ but without and embedded proof (the `proof` property refers to a separate graph),

OK, we are trying to understand that [in a separate thread](https://github.com/w3c/vc-data-model/pull/1416#issuecomment-1910762439).

> and cannot be used for a _presentation_ (the `verifiableCredential` property refers to separate graphs). If this extra facility is accepted, it should be surrounded with warnings as for these limitations.

Isn't this notion of a "separate graph" something conceptual rather than syntactical? E.g., in [this example of a Verifiable Presentation](https://www.w3.org/TR/vc-data-model-2.0/#example-a-simple-example-of-a-verifiable-presentation), the credential is a simple node in the same graph. So is the proof [in this example](https://www.w3.org/TR/vc-data-model-2.0/#example-a-verifiable-credential-utilizing-an-embedded-proof). Both cases seem, to me, straightforward for applying SHACL.

iherman:
I am afraid it is very much conceptual. Consider the [Example 2](https://www.w3.org/TR/vc-data-model-2.0/#example-a-simple-example-of-a-verifiable-presentation) from the spec. I took the JSON-LD, and ran it through the JSON-LD playground (see [the full link](https://tinyurl.com/ylxarodz)). 

```
<did:example:c276e12ec21ebfeb1f712ebc6f1> <https://schema.org/name> "Example University" _:b1 .
<did:example:ebfeb1f712ebc6f1c276e12ec21> <https://www.w3.org/ns/credentials/examples#alumniOf> <did:example:c276e12ec21ebfeb1f712ebc6f1> _:b1 .
<http://university.example/credentials/1872> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://www.w3.org/2018/credentials#VerifiableCredential> _:b1 .
<http://university.example/credentials/1872> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://www.w3.org/ns/credentials/examples#ExampleAlumniCredential> _:b1 .
<http://university.example/credentials/1872> <https://www.w3.org/2018/credentials#credentialSubject> <did:example:ebfeb1f712ebc6f1c276e12ec21> _:b1 .
<http://university.example/credentials/1872> <https://www.w3.org/2018/credentials#issuer> <https://university.example/issuers/565049> _:b1 .
<http://university.example/credentials/1872> <https://www.w3.org/2018/credentials#validFrom> "2010-01-01T19:23:24Z"^^<http://www.w3.org/2001/XMLSchema#dateTime> _:b1 .
_:b0 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://www.w3.org/2018/credentials#VerifiablePresentation> .
_:b0 <https://www.w3.org/2018/credentials#verifiableCredential> _:b1 .
```

Note the last two lines: the VP itself is identified with a blank node `_:b0`, but the `verifiableCredential` property refers to `_:b1` which identifies a separate _graph_ in an [RDF Dataset](https://www.w3.org/TR/rdf11-concepts/#section-dataset), appearing as the fourth entry of all other quads (signalling the fact that those are in a separate graph).

On the JSON-LD side this is achieved in the [context file](https://www.w3.org/ns/credentials/v2), through the [`"@container":"@graph"` directive](https://www.w3.org/TR/rdf11-concepts/#section-dataset) in:

```json
"VerifiablePresentation": {
      "@id": "https://www.w3.org/2018/credentials#VerifiablePresentation",
      "@context": {
        "@protected": true,
        "id": "@id",
        "type": "@type",
         …
        "verifiableCredential": {
          "@id": "https://www.w3.org/2018/credentials#verifiableCredential",
          "@type": "@id",
          "@container": "@graph",
          "@context": null
        },
        …
    },
```

(The same directive is used for `proof`.)

Coming back to SHACL: Alas!, that spec does not address [RDF Datasets](https://www.w3.org/TR/rdf11-concepts/#section-dataset) at this moment. Hence, the restrictive usage for VC/VP as I described in https://github.com/w3c/vc-data-model/pull/1416#issuecomment-1911496907.

contexts/credentials/v2
decentralgabe: A specification that's useful for implementers would be ideal. If some such documents already exist maybe they are fine to point to.

brentzundel:
FYI issue #1419 is raised to track this

TallTed:
> a new version of SHACL ... that handles graphs

Might be a strong argument toward chartering a new Shapes WG to update SHACL to handle graphs *and* RDF 1.2

iherman:
> > a new version of SHACL ... that handles graphs
> 
> Support for VC/VP applications might be a strong argument toward chartering a new Shapes WG to update SHACL to handle graphs _and_ RDF 1.2

Strong +1 to that, but that is definitely not my call...

@pchampin

clange:
> […]
> Note the last two lines: the VP itself is identified with a blank node `_:b0`, but the `verifiableCredential` property refers to `_:b1` which identifies a separate _graph_ in an [RDF Dataset](https://www.w3.org/TR/rdf11-concepts/#section-dataset), appearing as the fourth entry of all other quads (signalling the fact that those are in a separate graph).
> 
> On the JSON-LD side this is achieved in the [context file](https://www.w3.org/ns/credentials/v2), through the [`"@container":"@graph"` directive](https://www.w3.org/TR/rdf11-concepts/#section-dataset) in […]

I see, thank you very much for explaining the details. [BTW here is the relevant part from the JSON-LD spec, which helped me to fully understand it.](https://www.w3.org/TR/json-ld11/#graph-containers) It was not quite obvious from just looking at the JSON-LD listing. Now it's clear to me that, here, we would require a feature that SHACL does not yet support.

Still, how is it possible to validate credentials against a JSON Schema? Because JSON Schema is simply completely ignorant of any graph semantics? 😉

clange:
> > a new version of SHACL ... that handles graphs
> 
> Support for VC/VP applications might be a strong argument toward chartering a new Shapes WG to update SHACL to handle graphs _and_ RDF 1.2

Sounds good to me.

To avoid misunderstandings: am I understanding correctly that support for RDF 1.2 is a different, independent issue? (Which is also worth addressing, of course.) [The definition of RDF datasets was introduced in RDF 1.1](https://www.w3.org/TR/rdf11-new/#datasets), wasn't it?

TallTed:
> [@clange] To avoid misunderstandings: am I understanding correctly that support for RDF 1.2 is a different, independent issue? (Which is also worth addressing, of course.) [The definition of RDF datasets was introduced in RDF 1.1](https://www.w3.org/TR/rdf11-new/#datasets), wasn't it?

Yes, two distinct features to be added to SHACL (to make SHACL 1.1 or 2.0):

1. support for RDF datasets — entirely unsupported by existing SHACL
2. support for RDF 1.2 — existing SHACL is based on RDF 1.1

iherman:
> Still, how is it possible to validate credentials against a JSON Schema? Because JSON Schema is simply completely ignorant of any graph semantics? 😉

Exactly.

iherman:
@clange is o.k. to close this PR without any action?

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-01-31-vcwg#section1-3) on 2024-01-31

- no resolutions were taken

#### 1.3. Adding SHACL as a credential schema language (pr vc-data-model#1416)

_See github pull request [vc-data-model#1416](https://github.com/w3c/vc-data-model/pull/1416)._




**Manu Sporny:** There's a PR opened that suggests adding SHACL as a credential schema language in addition to the JSON schema stuff. This PR is to add it to the core specification.  
… Ivan has noted that there are a number of issues with using SHACL as a credential schema language, notably SHACL doesn't support datasets (only supports 1 graph, not multiple and we use multiple in the core data model).  
… It is not necessary for us to put this as a core part of the VCDM as you can always do this as an extension.  
… This is coming in pretty late, we don't know how many people would implement, feature freeze was a while ago, and I believe Ivan got agreement from Christoph to close it.  

**Brent Zundel:** SHACL doesn't support what our data model requires?  

**Manu Sporny:** Yes, you could be very careful how you use it and ensure certain parts of the data model but not all of it.  

**Brent Zundel:** Ok, so nice to see a PR from people reading our spec to do cool things with it ... but unlikely that this PR will be merged. Any comments before moving on?  

**Joe Andrieu:** I'm just wondering if we could queue up something in the implementation guide to support explaining some of this. How you could extend/use SHACL in this way/be careful about these things ... that might address the request. But I agree with feature freeze, I don't want to change the VCDM in some way.  

**Ted Thibodeau Jr.:** We might want to encourage the person to open issues instead of going straight to PR.  

**Brent Zundel:** In fairness to them they did open a discussion but we don't use that so nobody really engaged because we don't use that feature. So I created an issue based on that discussion.  
… We will be marking it as future.  

**Manu Sporny:** I'd argue against using SHACL at this point in time, so I think it's dangerous to just do partial, you could argue that JSON schema falls into that same category but it's also possible to use it by matching against context to check semantics, etc.  
… I think a future version thing is what it is, we shouldn't merge the PR because of limitations in the current version of SHACL. We should see if this works by having an extension incubate this and try it out and only then later potentially put into a future VCDM.  
… They should use the extension point to attempt this and they need to have an answer to the SHACL doesn't support datasets concern.  

</details>

msporny:
Noting the discussion in this thread that there is interest in incubating the use of SHACL for VCs, but that the WG believes it's premature to do so given that the SHACL support hasn't been incubated yet. This is being tracked in issue https://github.com/w3c/vc-data-model/issues/1419 for work in a future WG. Closing.

Issue #1217, opened 2023-07-26
Add link to Linked Data Best Practices 


msporny:
In https://github.com/w3c/vc-data-model/pull/1203#discussion_r1275176621 @awoie wrote:
> We should also provide a link to what it means to follow linked data best practices. 

This issue is to track the request by @awoie above.

Perhaps the best place to put the advice is in the Extensibility section.

awoie:
The proposal would work for me.

iherman:
@msporny I will come up with a PR to this (and #1210) to move things further along.

msporny:
PR #1420 addressed this issue and has been merged, closing.

Issue #1210, opened 2023-07-23
Document issuer defined vocabulary and use of @vocab


msporny:
@decentralgabe wrote in https://github.com/w3c/vc-data-model/pull/1202/files#r1265764973 :
> Would it help to add a note here around the inclusion of @vocab into the core context making this a bit simpler?

msporny:
There are a number of things that we've done to make it easier to work w/ VCs in a JSON-only mode. Use of `@protected`, use of `@vocab`, aliasing `@id` and `@type`, etc. We do talk about some of these mechanisms in these sections: 

* https://w3c.github.io/vc-data-model/#syntactic-sugar
* https://w3c.github.io/vc-data-model/#getting-started

... but, we don't mention "the issuer defined vocabulary" anywhere in the spec, and we probably should.

iherman:
@msporny @decentralgabe I will come up with PR covering this (and #1217) to move things.

msporny:
PR #1420 has been merged, closing.

Issue #1410, opened 2024-01-06
Does the specification need a normative "Credential Type Specifications" section?


msporny:
Based on comments from @jyasskin in https://github.com/w3c/vc-data-model/issues/1388#issuecomment-1875788296 :+1: 
> We have a recent precedent for non-testably "MUST"ing specifications in https://w3c.github.io/vc-data-model/#securing-mechanism-specifications. I think it's reasonable to write similar requirements for specifications for credential types.

This issue has been opened to track whether or not the WG desires to write such a section, they believe the current statements in the specification are good enough, or if an entry in the implementation guide would be more appropriate.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-01-17-vcwg#section2-1) on 2024-01-17

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 2.1. Does the specification need a normative "Credential Specifications" section? (issue vc-data-model#1410)

_See github issue [vc-data-model#1410](https://github.com/w3c/vc-data-model/issues/1410)._




**Brent Zundel:** We're going to start with a non-triaged issue.  
… Is this something we feel is an issue that needs to be addressed, if so, who is going to work on it?  

**Manu Sporny:** Yes, I can provide insight but Jeffrey is here too.  
… We're trying to provide guidance on things that ... people who extend the data model and write specifications about credentials, like specific types of credentials like driver's licenses, passports, coupons, whatever.  
… I think Jeffrey's concern was that there are times when it's ok using credential type-specific processing and there may times when that's not ok. I think Jeffrey you wanted us to give clearer guidance on that -- on things you should and shouldn't be doing. Like avoid these features / throw errors on this stuff, you should be producing a JSON schema that should be doing xyz and whether the VCDM gives guidance to people writing credential.  
… specs and what they should say about type-specific processing.  
… We give guidance on securing specs and this would give guidance for credential spec writers that say things like "provide a JSON schema for type-specific processors, provide a semantically immutable context, etc.".  

**Jeffrey Yasskin:** Yeah, that does basically capture what I'm worried about. My core worry was broader. If you secure something w/ JSON-LD and process it w/ JSON there are footguns, or vice versa, I didn't find attacks but doesn't mean there aren't any. So guidance so people don't write vulnerabilities.  

**Brent Zundel:** It seems like this is something important to work on, is this something that could be handled editorially, or does it require normative changes?  

[...]

PR #1420, opened 2024-01-30
Updated section on semantic interoperability (issues #1217 and #1210)

iherman:
This PR is an attempt to jointly cover issues #1217 and #1210, by a thorough re-write of [§5.3.1](https://www.w3.org/TR/vc-data-model-2.0/#semantic-interoperability) on Semantic Interoperability.

This is a normative section, ie, special attention should be given to the usage of the _MUST_, _SHOULD_, etc, keywords. My general approach was to reduce the _MUST_ to the strict minimum, but reviewers might differ.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1420.html" title="Last updated on Feb 10, 2024, 8:35 PM UTC (3f5c4d6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1420/8cc968a...3f5c4d6.html" title="Last updated on Feb 10, 2024, 8:35 PM UTC (3f5c4d6)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3079,8 +3079,8 @@ these extension points.
           <h4>Semantic Interoperability</h4>
 
           <p>
-            When defining new terms in an application-specific vocabulary, developers MUST use unique [=URLs=] to identify the terms.
-            Whenever possible, it is RECOMMENDED to re-use terms — and their corresponding URLs — defined by well-known, public vocabularies,
+            When defining new terms in an application-specific vocabulary, developers MUST use globally unambiguous [=URLs=] to identify the terms.
+            Whenever possible, it is RECOMMENDED to re-use terms — and their corresponding URLs — defined by well-known, public vocabularies,
             such as [[[?DCTERMS]]] [[?DCTERMS]] or [[[?schema-org]]] [[?schema-org]]. 
             If that is not possible, developers MUST <em>define</em> a new URL for each term. 
             When doing so, general [[LINKED-DATA]] principles must be followed, in particular:
--- a/index.html
+++ b/index.html
@@ -3081,9 +3081,9 @@ these extension points.
           <p>
             When defining new terms in an application-specific vocabulary, developers MUST use globally unambiguous [=URLs=] to identify the terms.
             Whenever possible, it is RECOMMENDED to re-use terms — and their corresponding URLs — defined by well-known, public vocabularies,
-            such as [[[?DCTERMS]]] [[?DCTERMS]] or [[[?schema-org]]] [[?schema-org]]. 
-            If that is not possible, developers must <em>define</em> a new URL for each term. 
-            When doing so, the general [[LINKED-DATA]] are expected to be followed, in particular:
+            such as [[[?DCTERMS]]] [[?DCTERMS]] or [[[?schema-org]]] [[?schema-org]].
+            If that is not possible, authors MUST <em>define</em> a new URL for each term.
+            When doing so, the general guidelines for [[LINKED-DATA]] are expected to be followed, in particular:
           </p>
 
           <ul>
@@ -3092,48 +3092,48 @@ these extension points.
                 the constraints on the use of each term.
               </li>
               <li>
-                  It is RECOMMENDED to also publish the collection of all new terms as a machine-readable vocabulary 
+                  It is RECOMMENDED to also publish the collection of all new terms as a machine-readable vocabulary
                   using [[[?RDF-SCHEMA]]] [[?RDF-SCHEMA]].
               </li>
               <li>
-                  It SHOULD be possible to dereference the URL of a term, resulting in its description and/or formal definition.   
-              </li> 
-          </ul>       
-             
+                  It SHOULD be possible to dereference the URL of a term, resulting in its description and/or formal definition.
+              </li>
+          </ul>
+
           <p>
             Developers SHOULD follow the <a data-cite="?LD-BP#vocabulary-checklist">detailed checklist</a> in
             [[[?LD-BP]]] [[?LD-BP]] when defining a new vocabulary.
           </p>
           <p>
-              Furthermore, a machine-readable description (that is, a 
+              Furthermore, a machine-readable description (that is, a
               <a data-cite="JSON-LD11#dfn-context">JSON-LD Context document</a>) MUST be published at the URL specified
-              in the `@context` [=property=] for the vocabulary. 
+              in the `@context` [=property=] for the vocabulary.
               This context MUST map each term to its corresponding URL, possibly accompanied by further
               constraints like the type of the property value. A human-readable document describing the expected order of values for
-              the `@context` [=property=] is also expected to be published by any implementer seeking interoperability.     
+              the `@context` [=property=] is also expected to be published by any implementer seeking interoperability.
           </p>
 
           <p class="note">
               When processing the <a data-cite="JSON-LD11#dfn-active-context">active context</a> defined by the base JSON-LD Context
-              document <a href="#dfn-context">defined in this specification</a>, compliant JSON-LD-based processors produce an error when a JSON-LD context 
+              document <a href="#dfn-context">defined in this specification</a>, compliant JSON-LD-based processors produce an error when a JSON-LD context
               <em>redefines</em> any term.
               The only way to change the definition of existing terms is to introduce a new term that clears the active context within
-              the scope of that new term. 
-              Authors that are interested in this feature should read about the <a data-cite="JSON-LD11#protected-term-definitions">`@protected`</a> 
+              the scope of that new term.
+              Authors that are interested in this feature should read about the <a data-cite="JSON-LD11#protected-term-definitions">`@protected`</a>
               keyword in the JSON-LD 1.1 specification.
           </p>
 
           <p>
-              The base JSON-LD Context file for this specification also includes an extra feature, using the 
+              The base JSON-LD Context file for this specification also includes an extra feature, using the
               <a data-cite="JSON-LD11/#default-vocabulary">`@vocab`</a> keyword, which ensures
               that any undefined term in a [=verifiable credential=] or a [=verifiable presentation=] is automatically mapped to a
-              URL prefixed with `https://www.w3.org/ns/credentials/issuer-dependent#`. 
+              URL prefixed with `https://www.w3.org/ns/credentials/issuer-dependent#`.
               This is to allow early experimentation with terms during the development phase, without
-              requiring a formal definition in every cycle of that experimentation. 
+              requiring a formal definition in every cycle of that experimentation.
               Note that developers SHOULD NOT use this feature in production; this
-              could lead to name clashes, yielding semantic ambiguities with other applications. 
+              could lead to name clashes, yielding semantic ambiguities with other applications.
               Instead, they SHOULD define all the
-              terms, as described earlier in this section, to achieve proper interoperability.  
+              terms, as described earlier in this section, to achieve proper interoperability.
           </p>
         </section>
       </section>
--- a/index.html
+++ b/index.html
@@ -3082,8 +3082,8 @@ these extension points.
             When defining new terms in an application-specific vocabulary, developers MUST use globally unambiguous [=URLs=] to identify the terms.
             Whenever possible, it is RECOMMENDED to re-use terms — and their corresponding URLs — defined by well-known, public vocabularies,
             such as [[[?DCTERMS]]] [[?DCTERMS]] or [[[?schema-org]]] [[?schema-org]]. 
-            If that is not possible, developers MUST <em>define</em> a new URL for each term. 
-            When doing so, general [[LINKED-DATA]] principles must be followed, in particular:
+            If that is not possible, developers must <em>define</em> a new URL for each term. 
+            When doing so, the general [[LINKED-DATA]] are expected to be followed, in particular:
           </p>
 
           <ul>
@@ -3093,7 +3093,7 @@ these extension points.
               </li>
               <li>
                   It is RECOMMENDED to also publish the collection of all new terms as a machine-readable vocabulary 
-                  using [[[?RDF-SCHEMA]]] [[?RDF-SCHEMA]] (and, if necessary, OWL2 [[?OWL2-OVERVIEW]]).
+                  using [[[?RDF-SCHEMA]]] [[?RDF-SCHEMA]].
               </li>
               <li>
                   It MUST be possible to dereference the URL of a term, resulting in its description and/or formal definition.   
--- a/index.html
+++ b/index.html
@@ -3078,25 +3078,61 @@ these extension points.
         <section>
           <h4>Semantic Interoperability</h4>
 
+          <p>
+            When defining new terms in an application-dependent vocabulary, developers MUST use unique [=URLs=] to identify the terms.
+            It is RECOMMENDED to re-use, whenever possible, terms—and their corresponding URLs—defined by well-known, public vocabularies,
+            such as [[[?DCTERMS]]] [[?DCTERMS]] or [[[?schema-org]]] [[?schema-org]]. 
+            If that is not possible, developers MUST <em>define</em> a new URL for each term. 
+            When doing so, general [[LINKED-DATA]] principles must be followed, in particular:
+          </p>
+
           <ul>
-            <li>
-JSON-LD-based processors MUST produce an error when a JSON-LD context redefines
-any term in the
-<a href="https://www.w3.org/TR/json-ld/#dfn-active-context">active context</a>.
-The only way to change the definition of existing terms is to introduce a new
-term that clears the active context within the scope of that new term. Authors
-that are interested in this feature should read about the
-`@protected` feature in the JSON-LD 1.1 specification.
-            </li>
-          </ul>
+              <li>
+                A human-readable documentation MUST be published, describing the semantics and
+                the constraints on the use of each term.</li>
+              <li>
+                  It is RECOMMENDED to also publish the collection of all new terms as a machine-readable vocabulary 
+                  using [[[?RDF-SCHEMA]]] [[?RDF-SCHEMA]] (and, if necessary, OWL2 [[?OWL2-OVERVIEW]]).
+              </li>
+              <li>
+                  It MUST be possible to dereference the URL of a term, resulting in its description and/or formal definition.   
+              </li> 
+          </ul>       
+             
+          <p>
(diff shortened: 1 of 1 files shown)


decentralgabe (approved):

index.html
```diff
@@ -3078,25 +3078,61 @@ <h3>Extensibility</h3>
...
-any term in the
-<a href="https://www.w3.org/TR/json-ld/#dfn-active-context">active context</a>.
-The only way to change the definition of existing terms is to introduce a new
-term that clears the active context within the scope of that new term. Authors
-that are interested in this feature should read about the
-`@protected` feature in the JSON-LD 1.1 specification.
-            </li>
-          </ul>
+              <li>
+                A human-readable documentation MUST be published, describing the semantics and
```
TallTed, line 3091:
```suggestion
                Human-readable documentation MUST be published, describing the semantics of and
```

index.html
```diff
@@ -3078,25 +3078,61 @@ <h3>Extensibility</h3>
...
-implementer seeking interoperability. A machine-readable description
-(that is, a normal JSON-LD Context document) is expected to be published
-at the URL specified in the `@context` [=property=] by
-JSON-LD implementers seeking interoperability.
+              The base JSON-LD Context file for this specification also includes an extra feature, using the 
+              <a data-cite="JSON-LD11/#default-vocabulary">`@vocab`</a> keyword, which ensures
+              that any undefined term in a [=verifiable credential=] or a [=verifiable presentation=] is automatically mapped to a
+              URL prefixed with `https://www.w3.org/ns/credentials/issuer-dependent#`. 
+              The purpose is to allow early experimentation with terms during the development phase, without
+              the necessity to provide a formal definition in every cycle of that experimentation.
```
TallTed, line 3131:
```suggestion
              This is to allow early experimentation with terms during the development phase, without
              requiring a formal definition in every cycle of that experimentation. 
```

index.html
```diff
@@ -3078,25 +3078,61 @@ <h3>Extensibility</h3>
...
-(that is, a normal JSON-LD Context document) is expected to be published
-at the URL specified in the `@context` [=property=] by
-JSON-LD implementers seeking interoperability.
+              The base JSON-LD Context file for this specification also includes an extra feature, using the 
+              <a data-cite="JSON-LD11/#default-vocabulary">`@vocab`</a> keyword, which ensures
+              that any undefined term in a [=verifiable credential=] or a [=verifiable presentation=] is automatically mapped to a
+              URL prefixed with `https://www.w3.org/ns/credentials/issuer-dependent#`. 
+              The purpose is to allow early experimentation with terms during the development phase, without
+              the necessity to provide a formal definition in every cycle of that experimentation. 
+              However, developers SHOULD NOT use this feature in production; this
```
TallTed, line 3132:
```suggestion
              Note that developers SHOULD NOT use this feature in production; this
```

TallTed (changes requested):
Editorial...

index.html
```diff
@@ -3078,25 +3078,61 @@ <h3>Extensibility</h3>
         <section>
           <h4>Semantic Interoperability</h4>
 
+          <p>
+            When defining new terms in an application-dependent vocabulary, developers MUST use unique [=URLs=] to identify the terms.
```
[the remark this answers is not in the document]
dlongley: `unique` => `globally unambiguous`

dlongley (commented):
I don't know about the testability of all the MUSTs here and whether, on account of that, they should instead be "expectations" or recommendations.

iherman:
> I don't know about the testability of all the MUSTs here and whether, on account of that, they should instead be "expectations" or recommendations.

The change could be to use SHOULD/RECOMMENDED everywhere.

However. "Testability" in the W3C CR setting is a very general concept; the goal is to prove that a feature is implementable by two, independent implementations (to ensure interoperability). From a CR point of view, this does ***not*** mean that each _MUST_ should have a running test program. (The usual counterexample is the "testing" of a vocabulary, which means that the terms are in use by at least two different implementations.) Taking this view, the few _MUST_ statements in this section are easily testable: it requires at least two VC applications that define their own application specific terms, and that abide to these requirements. Isn't that feasible?

(Personally, I would prefer, from the spec's point of view, to keep those _MUST_-s.)

iherman:
@TallTed:

> Editorial...

All adopted.

index.html
```diff
@@ -3078,25 +3078,61 @@ <h3>Extensibility</h3>
...
-<a href="https://www.w3.org/TR/json-ld/#dfn-active-context">active context</a>.
-The only way to change the definition of existing terms is to introduce a new
-term that clears the active context within the scope of that new term. Authors
-that are interested in this feature should read about the
-`@protected` feature in the JSON-LD 1.1 specification.
-            </li>
-          </ul>
+              <li>
+                Human-readable documentation MUST be published, describing the semantics of and
+                the constraints on the use of each term.</li>
```
davidlehn, line 3092:
```suggestion
                the constraints on the use of each term.
              </li>
```

index.html
```diff
@@ -3078,25 +3078,62 @@ <h3>Extensibility</h3>
         <section>
           <h4>Semantic Interoperability</h4>
 
+          <p>
+            When defining new terms in an application-specific vocabulary, developers MUST use globally unambiguous [=URLs=] to identify the terms.
+            Whenever possible, it is RECOMMENDED to re-use terms — and their corresponding URLs — defined by well-known, public vocabularies,
+            such as [[[?DCTERMS]]] [[?DCTERMS]] or [[[?schema-org]]] [[?schema-org]]. 
+            If that is not possible, developers MUST <em>define</em> a new URL for each term.
```
msporny, line 3085: Doesn't this MUST duplicate the "developers MUST use globally unambiguous [=URLs=]" one above?

index.html
```diff
@@ -3078,25 +3078,62 @@ <h3>Extensibility</h3>
         <section>
           <h4>Semantic Interoperability</h4>
 
+          <p>
+            When defining new terms in an application-specific vocabulary, developers MUST use globally unambiguous [=URLs=] to identify the terms.
+            Whenever possible, it is RECOMMENDED to re-use terms — and their corresponding URLs — defined by well-known, public vocabularies,
+            such as [[[?DCTERMS]]] [[?DCTERMS]] or [[[?schema-org]]] [[?schema-org]]. 
+            If that is not possible, developers MUST <em>define</em> a new URL for each term. 
+            When doing so, general [[LINKED-DATA]] principles must be followed, in particular:
```
msporny, line 3086: Is this a normative MUST? If not, we should say something like: "are expected to be followed"

index.html
```diff
@@ -3078,25 +3078,62 @@ <h3>Extensibility</h3>
...
+
           <ul>
-            <li>
-JSON-LD-based processors MUST produce an error when a JSON-LD context redefines
-any term in the
-<a href="https://www.w3.org/TR/json-ld/#dfn-active-context">active context</a>.
-The only way to change the definition of existing terms is to introduce a new
-term that clears the active context within the scope of that new term. Authors
-that are interested in this feature should read about the
-`@protected` feature in the JSON-LD 1.1 specification.
```
msporny, line 3089: Hmm, IIRC, the WG came to consensus on adding this text to ensure that a JSON-LD processor that didn't throw a `@protected` error was non-conformant wrt. usage for VCs. There were individuals in the group that were concerned about term redefinitions not resulting in an error. We can't remove this language, IMHO.

index.html
```diff
@@ -3078,25 +3078,62 @@ <h3>Extensibility</h3>
...
-            </li>
-          </ul>
+              <li>
+                Human-readable documentation MUST be published, describing the semantics of and
+                the constraints on the use of each term.
```
msporny, line 3092: This section is turning into a "Requirements for Vocabulary Document Authors", which I thought we were going to avoid writing as of last week. IOW, this PR feels like it's trying to also address issue https://github.com/w3c/vc-data-model/issues/1410

That said, this isn't a bad list (although, feels a bit heavy handed)...

index.html
```diff
@@ -3078,25 +3078,62 @@ <h3>Extensibility</h3>
...
+              </li>
+              <li>
+                  It is RECOMMENDED to also publish the collection of all new terms as a machine-readable vocabulary 
+                  using [[[?RDF-SCHEMA]]] [[?RDF-SCHEMA]] (and, if necessary, OWL2 [[?OWL2-OVERVIEW]]).
```
msporny, line 3096: -1 to bringing in OWL2... those that want to do it will do it. We shouldn't place this burden on vocabulary authors (unless the OWL2 and RDF-SCHEMA stuff is done automatically for individuals). It would be better to suggest using yaml2vocab than to point people to these specifications. 

We don't want to make the process so onerous that people cite how difficult it is to create vocabulary documents and then use that as a reason to not use the technology.

index.html
```diff
@@ -3078,25 +3078,62 @@ <h3>Extensibility</h3>
...
+              </li>
+              <li>
+                  It MUST be possible to dereference the URL of a term, resulting in its description and/or formal definition.
```
msporny, line 3099:
```suggestion
                  It SHOULD be possible to dereference the URL of a term, resulting in its description and/or formal definition.   
```

The base level requirement for VCs was always "There exists a URL for each term"... all the other stuff is optional... yes, it's a best practice, but we don't want to create a mountain of work for people just dipping their toes in the water.

index.html
```diff
@@ -3078,25 +3078,62 @@ <h3>Extensibility</h3>
...
+                  using [[[?RDF-SCHEMA]]] [[?RDF-SCHEMA]] (and, if necessary, OWL2 [[?OWL2-OVERVIEW]]).
+              </li>
+              <li>
+                  It MUST be possible to dereference the URL of a term, resulting in its description and/or formal definition.   
+              </li> 
+          </ul>       
+             
+          <p>
+            Developers SHOULD follow the <a data-cite="?LD-BP#vocabulary-checklist">detailed checklist</a> in
+            [[[?LD-BP]]] [[?LD-BP]] when defining a new vocabulary.
```
msporny, line 3105: Yes, this is the right level of suggestion for this section, IMHO.

index.html
```diff
@@ -3078,25 +3078,62 @@ <h3>Extensibility</h3>
...
+          </p>
+          <p>
+              Furthermore, a machine-readable description (that is, a 
+              <a data-cite="JSON-LD11#dfn-context">JSON-LD Context document</a>) MUST be published at the URL specified
+              in the `@context` [=property=] for the vocabulary.
```
msporny, line 3110: We don't need to say this. The system will just not work if a context isn't published at that location.

index.html
```diff
@@ -3078,25 +3078,62 @@ <h3>Extensibility</h3>
...
+              in the `@context` [=property=] for the vocabulary. 
+              This context MUST map each term to its corresponding URL, possibly accompanied by further
+              constraints like the type of the property value. A human-readable document describing the expected order of values for
+              the `@context` [=property=] is also expected to be published by any implementer seeking interoperability.     
+          </p>
+
+          <p class="note">
+              When processing the <a data-cite="JSON-LD11#dfn-active-context">active context</a> defined by the base JSON-LD Context
+              document <a href="#dfn-context">defined in this specification</a>, compliant JSON-LD-based processors produce an error when a JSON-LD context 
+              <em>redefines</em> any term.
```
msporny, line 3119: Hmm, this downgrades a previous normative requirement to just a note. I thought the group felt strongly about the previous normative statement (though, that's "just how modern JSON-LD Processors are expected to work if an `@protected` term is redefined". We should make sure the group is ok with this.

msporny (changes requested):
A few questions and changes requested. Overall, we should decide if we're also addressing issue #1410 with this PR (as we're dangerously close to doing so, IMHO). 

Some of the normative statements feel a bit heavy-handed (and untestable -- or should we test them?).

index.html
iherman: In a way it does indeed. I will change this to a lower case "must"

index.html
iherman: Yeah, I did not want to put a _MUST_ here, the ld principles are fairly long. I changed it to the version you proposed.

index.html
iherman: I am not sure I understand this remark. Yes, the document will be non conformant, because the JSON-LD processor will fail if the `@context` value does not dereference. But if the user does not publish a context document _nor_ does he/she put a `@context` property for the vocabulary, the document will pass, thanks to the `@vocab` in the core context...

index.html
```diff
@@ -3078,25 +3078,62 @@ <h3>Extensibility</h3>
...
+          <p>
+            Developers SHOULD follow the <a data-cite="?LD-BP#vocabulary-checklist">detailed checklist</a> in
+            [[[?LD-BP]]] [[?LD-BP]] when defining a new vocabulary.
+          </p>
+          <p>
+              Furthermore, a machine-readable description (that is, a 
+              <a data-cite="JSON-LD11#dfn-context">JSON-LD Context document</a>) MUST be published at the URL specified
+              in the `@context` [=property=] for the vocabulary. 
+              This context MUST map each term to its corresponding URL, possibly accompanied by further
+              constraints like the type of the property value. A human-readable document describing the expected order of values for
```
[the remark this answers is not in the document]
iherman: Hm. What I wanted to say is that the context defined for the vocabulary MUST map each term in this vocabulary to the corresponding URL (plus define constraints). This pre-empts the effect of `@vocab`. I am not sure what the problem is...

index.html
iherman: What bothered me is that the previous normative sentence seemed to impose a normative requirement on the JSON-LD processor, ie, the way it handles `@protected`. Which is not for this specification to define, this is how JSON-LD 1.1 is defined. Hence this 'note' version which simply states how JSON LD processor behaves.

We may want to add something along the lines of 'per the defition of JSON-LD 1.1' with a reference to that spec. But it still is non-normative in _this_ spec.

iherman:
> Some of the normative statements feel a bit heavy-handed (and untestable -- or should we test them?).

See my comments in https://github.com/w3c/vc-data-model/pull/1420#issuecomment-1918409250.

If we do have two "properly defined" vocabularies around as part of the implementations, then, CR-wise, we are o.k. No reason for explicit "tests" imho.

iherman:
> Overall, we should decide if we're also addressing issue #1410 with this PR (as we're dangerously close to doing so, IMHO).

We are not "dangerously close": we are covering it. A 'type' is part of a vocabulary, whether new properties are defined or not.

I am not sure whether we need to make it explicit in this text, and how. Maybe by defining (in the terms' section) what we mean by 'vocabulary': something that may be as simple as the definition of new types, and as complex as defining whole new properties.

index.html
dlongley: This is a misunderstanding and I think conflated with something else the group discussed. The text by @iherman above is correct.

index.html
dlongley: This is a misunderstanding, see my comment above. The VCDM is using `@protected` according to the JSON-LD 1.1 spec, there is no difference.

index.html
```diff
@@ -3078,25 +3078,62 @@ <h3>Extensibility</h3>
         <section>
           <h4>Semantic Interoperability</h4>
 
+          <p>
+            When defining new terms in an application-specific vocabulary, developers MUST use globally unambiguous [=URLs=] to identify the terms.
+            Whenever possible, it is RECOMMENDED to re-use terms — and their corresponding URLs — defined by well-known, public vocabularies,
+            such as [[[?DCTERMS]]] [[?DCTERMS]] or [[[?schema-org]]] [[?schema-org]]. 
+            If that is not possible, developers must <em>define</em> a new URL for each term. 
+            When doing so, the general [[LINKED-DATA]] are expected to be followed, in particular:
```
msporny, line 3086:
```suggestion
            When doing so, the general guidelines for [[LINKED-DATA]] are expected to be followed, in particular:
```

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-02-07-vcwg#section3-4) on 2024-02-07

- no resolutions were taken

#### 3.4. Updated section on semantic interoperability (issues #1217 and #1210) (pr vc-data-model#1420)

_See github pull request [vc-data-model#1420](https://github.com/w3c/vc-data-model/pull/1420)._




**Manu Sporny:** A number of PRs for VCDM. Huge thanks to ivan for raising a number of them. One we should talk about today (1420). Ivan raised.  
… a question remains. We just entered CR. I think this PR is normative. It does not impact software implementations but does impact vocabulary. Wonder if this triggers a second CR.  

**Ivan Herman:** AFAIK, a new draft with normative changes are OK provided that at some point in time we publish a snapshot with these changes. I don't think there is any problem with these few normative changes here. We know we will republish a snapshot sometime in May. I think we are fine.  

> *Brent Zundel:* +1 to Ivan.

**Manu Sporny:** That is my read as well. I've started tracking this with a normative label so we know we have done a second CR triggering thing and that we can summarize these changes. If you raise a PR, please try to classify as Editorial or Normative.  
… also keep in mind that any normative change needs to be communicated with the test suite workers.  

**Ivan Herman:** If a change is normative in a serious way, then the editor of the PR should add an item to the list of changes at the end of the document or we will forget them.  

> *Manu Sporny:* yes, +1 to that.

**Brent Zundel:** this is something reviews should remind when examining PRs.  
… Any other core data model status?  

**Manu Sporny:** A number of editorial changes are out there and folks should pay attention.  

</details>

index.html
```diff
@@ -3078,25 +3078,62 @@ <h3>Extensibility</h3>
         <section>
           <h4>Semantic Interoperability</h4>
 
+          <p>
+            When defining new terms in an application-specific vocabulary, developers MUST use globally unambiguous [=URLs=] to identify the terms.
+            Whenever possible, it is RECOMMENDED to re-use terms — and their corresponding URLs — defined by well-known, public vocabularies,
+            such as [[[?DCTERMS]]] [[?DCTERMS]] or [[[?schema-org]]] [[?schema-org]]. 
+            If that is not possible, developers must <em>define</em> a new URL for each term.
```
msporny, line 3085:
```suggestion
            If that is not possible, developers MUST <em>define</em> a new URL for each term. 
```

msporny (approved):

msporny:

msporny:

msporny:
BTW, I think this PR just created a new conformance class for vocabulary documents in the spec... If you agree, @iherman, we might want to add that conformance class and point it to the semantic interop section. I'll also note that I still think that this is a partial solution for #1410 because we don't add the text that I believe @jyasskin was looking for.

PR #1427, opened 2024-02-02
Fix URL prefixes.

davidlehn:
- It looks like `JsonSchema` and `JsonSchemaCredentials` ids are typos and should use the same id prefix as everything else.
- `VCDMExplainer.md` is old, but might as well use the correct URLs.
- The `vocab/credentials/*` files are older, and now not used at that endpoint in favor of `.../v2/` files (I think).  Seems worth fixing those files too in case people look at them.  Or perhaps they should be removed?
- The `sec` prefix update appears to be a typo.  Also in old files and appears unused.

### DIFF
--- a/VCDMExplainer.md
+++ b/VCDMExplainer.md
@@ -147,7 +147,7 @@ Decentralized Identifier (DID) scheme.
 ```
 {
   "@context": [
-    "https://w3.org/2018/credentials/v1",
+    "https://www.w3.org/2018/credentials/v1",
     "https://example.com/examples/v1"
   ],
   "id": "http://example.edu/credentials/1872",
@@ -174,7 +174,7 @@ The following is an example verifiable credential that supports ZKPs.
 ```
 {
   "@context": [
-    "https://w3.org/2018/credentials/v1",
+    "https://www.w3.org/2018/credentials/v1",
     "ctx:sov:anoncred:v1",
     "ctx:sov:GppHbMLLeKNYRhcQiXh3GjP2Yh",
   ],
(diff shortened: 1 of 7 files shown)


iherman (approved):
I agree that v2 must be consistent and, therefore, use www.w3.org everywhere.

I am pretty much neutral about the fate of the v1 vocabulary files. I always wondered whether it makes sense to keep them, instead of putting them in a separate area of the repo for archival. After all, the spec itself has changed in the same repo as well. But I yield to @msporny on this one.

dlongley (approved):

davidlehn:
> I am pretty much neutral about the fate of the v1 vocabulary files. I always wondered whether it makes sense to keep them, instead of putting them in a separate area of the repo for archival. After all, the spec itself has changed in the same repo as well. But I yield to @msporny on this one.

I had similar thoughts about the files.  Perhaps a first easy step would be to move the v1 files to a `v1/` directory so there's a clear `v1/` and `v2/` pattern.  I was initially confused looking at those before noticing the current files are in a `v2/` directory.  Would moving those files break anything?  I don't think those are published anywhere linkable?

msporny (approved):

msporny:

msporny:
@davidlehn wrote:
> Perhaps a first easy step would be to move the v1 files to a `v1/` directory so there's a clear `v1/` and `v2/` pattern. 

Done in 7966eb248393053c2d5df467aefb5ccbc82e4613.

... now to see what breaks, if anything. :)

PR #1429, opened 2024-02-02
add link to Use Cases doc, remove use cases section.

### LINKED ISSUE #1169: Consider removing Use Cases summary from core DM

brentzundel:
This PR fixes #1169 by removing the use cases section from the data model spec and adding a link to the use cases and requirements doc.

***
<a href="https://pr-preview.s3.amazonaws.com/brentzundel/vc-data-model/pull/1429.html" title="Last updated on Feb 6, 2024, 4:00 PM UTC (a724dfb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1429/8cc968a...brentzundel:a724dfb.html" title="Last updated on Feb 6, 2024, 4:00 PM UTC (a724dfb)">Diff</a>


index.html
```diff
@@ -259,9 +259,10 @@ <h2>Introduction</h2>
...
 [=verifiable presentations=] are expected to be useful
         </li>
-        <li>
-The use cases and requirements that informed this specification.
-        </li>
+        <p>
+[[[VC-USE-CASES]]][[?VC-USE-CASES]] is also an excellent resource for the use
+cases and requirements that informed this specification.
+        </p>
       </ul>
```
TallTed, line 266:
```suggestion
      </ul>
      <p>
The use cases and requirements that informed this specification may be found
in [[[VC-USE-CASES]]] [[?VC-USE-CASES]].
      </p>
```

TallTed (changes requested):

index.html
msporny: s/may/can/ since MAY is normative language and some people get confused when we use normative language in lower case.

msporny (approved):
+1 to @TallTed's changes.

Goodbye list of "things we wanted to accomplish", you served us well over the past 7+ years.

index.html
marcelruland: Is using normative words in a lower-case, non-normative way generally discouraged in spite of the explicit mention that normative language is always UPPERCASE? There are quite a few other avoidable examples of this in the document.

TallTed: > Is using normative words in a lower-case, non-normative way generally discouraged in spite of the explicit mention that normative language is always UPPERCASE?

Yes, because people either ignore or don't yet absorb the discussion of these words only having their [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119) (issued in March 1997) meaning when/if they're all-caps ([RFC 8174](https://datatracker.ietf.org/doc/html/rfc8174), issued in May 2017). This is not really surprising, as they clearly had their RFC 2119 meaning regardless of case for *20 years* before RFC 8174 clarification only 6 years ago.

TallTed:
Applying my remaining change request will also trigger PR-Preview, which should now work as https://github.com/tobie/pr-preview/issues/149 has been fixed.

decentralgabe (approved):

iherman (approved):

msporny:

PR #1435, opened 2024-02-10
Update refresh service example to use example from CCG spec.

msporny:
This PR attempts to address issue #981 by updating the refresh service example to use the one deployed by TruAge (defined by the[ CCG specification](https://w3c-ccg.github.io/vc-refresh-2021/#verifiablecredentialrefreshservice2021-protocol)).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1435.html" title="Last updated on Feb 18, 2024, 7:28 PM UTC (1dff5c2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1435/d42fd22...1dff5c2.html" title="Last updated on Feb 18, 2024, 7:28 PM UTC (1dff5c2)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3389,12 +3389,9 @@ thereby bypassing the [=holder=].
 The value of the `refreshService` [=property=] MUST be one or more
 refresh services that provides enough information to the recipient's software
 such that the recipient can refresh the [=verifiable credential=]. Each
-`refreshService` value MUST specify its `type` (for
-example, `ManualRefreshService2018`) and its `id`, which
-is the [=URL=] of the service. There is an expectation that machine readable
-information needs to be retrievable from the URL. The precise content of
-each refresh service is determined by the specific `refreshService`
-[=type=] definition.
+`refreshService` value MUST specify its `type`. The precise content of each
+refresh service is determined by the specific `refreshService` [=type=]
+definition.
           </dd>
         </dl>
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -3417,16 +3414,18 @@ <h3>Refreshing</h3>
     }
   },
   <span class="highlight">"refreshService": {
-    "id": "https://university.example/refresh/3732",
-    "type": "ManualRefreshService2018"
+    "type": "VerifiableCredentialRefreshService2021",
```
iherman, line 3417: As it stands, the type will expand to `https://www.w3.org/ns/credentials/examples#VerifiableCredentialRefreshService2021` because the term itself is not defined in the core context file. Is that the intention? Or should we add a CCG-maintained context file somewhere (maybe as a local context)?

iherman (changes requested):

index.html
msporny: Yes, good point, I'll add the refresh service context in an update to this PR.

decentralgabe (approved):

TallTed (approved):
looks OK with @iherman's suggested tweak

index.html
msporny: Fixed in 1dff5c219f2408f92aa0d406f47a6ce233b5d16b.

msporny:

PR #1436, opened 2024-02-11
Update vocabulary and template files

iherman:
yml2vocab has now the ability of adding information on designated `@context` files that contain individual terms to the output. This change is adapted to this vocabulary.

No changes on the vocabulary content itself.

As usual, the automatic preview is useless for this case, see https://w3c.github.io/yml2vocab/previews/vcdm/ for the new versions of the generated files.


msporny (approved):

msporny:

PR #1438, opened 2024-02-14
clarify inputBytes, inputDocument, inputMediaType

### LINKED ISSUE #1402: phrasing and/or punctuation for input "inputBytes or inputDocument and inputMediaType" needs work
As written, it's unclear whether it's "(`inputBytes` or `inputDocument`) and `inputMediaType`" or "`inputBytes` or (`inputDocument` and `inputMediaType`)"; i.e., whether `inputMediaType` is always present (which I believe is correct), or only travels with an `inputDocument` (which I believe is incorrect).

This applies to at least two instances of these inputs (circa lines 2022 and 4540), possibly more.

This can be assigned to me. I wanted to get confirmation of my belief above before going to PR.

TallTed:
fixes #1402

***
<a href="https://pr-preview.s3.amazonaws.com/TallTed/vc-data-model/pull/1438.html" title="Last updated on Feb 14, 2024, 9:08 PM UTC (08db69b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1438/d42fd22...TallTed:08db69b.html" title="Last updated on Feb 14, 2024, 9:08 PM UTC (08db69b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4131,10 +4131,10 @@ secured, without any securing mechanism information included, such as `proof` or
 JOSE/COSE header parameters and signatures. Specifications MAY provide additional mechanisms to convey
 other information that might be helpful (for example, during validation or for
 debugging purposes), such as securing mechanism data. A securing mechanism's
-verification algorithm MUST provide an interface that receives a sequence of
-bytes ([=byte sequence=] |inputBytes|) or a document ([=map=] |inputDocument|)
-and a media type ([=string=] |inputMediaType|) as inputs and returns a
-verification result with at least the following [=struct/items=]:
+verification algorithm MUST provide an interface that receives inputs of a media
+type ([=string=] |inputMediaType|) paired with either a sequence of bytes 
+([=byte sequence=] |inputBytes|) or a document ([=map=] |inputDocument|), and
+returns a verification result with at least the following [=struct/items=]:
         </p>
 
         <dl>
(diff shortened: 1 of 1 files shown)


decentralgabe (approved):

iherman (approved):
A very practical issue aiming to avoid future merge conflicts...

Following the WG agreement, I have created a PR (#1440) changing the property `mediaType` to `encodingFormat`. I was wondering whether it wouldn't be better, for the sake of consistency, to use the term `inputEncodingFormat` instead of `inputMediaType` in this algorithm. It may be wiser if we agreed on this issue in this PR and, possibly, the term was changed here (and my PR concentrates exclusively on the property).

Note that the term `mediaType` also appears in the body of the algorithm. That should either be changed or kept unchanged, in sync with the change of the argument name.

Personally, I am in favour of changing the algorithm argument name, but I won't lie down the road if the agreement is to leave it as is.

TallTed:
Consistency is generally better than not.

That said, I don't like the change to `encodingFormat` because I don't think it's an accurate label nor particularly well defined by schema.org.

I also note (too late for that decision, but perhaps it might be reconsidered) https://www.w3.org/ns/iana/media-types/ which seems like it should be taken into account for W3 purposes.

iherman:
> Consistency is generally better than not.
> 
> That said, I don't like the change to `encodingFormat` because I don't think it's an accurate label nor particularly well defined by schema.org.

I have to agree with you on this. As I said, I can live with inconsistency; we just to agree that we all accept it.

> 
> I also note (too late for that decision, but perhaps it might be reconsidered) https://www.w3.org/ns/iana/media-types/ which seems like it should be taken into account for W3 purposes.

This is a very strange namespace (never saw this before). It seems to assign, for each specific media type, a class. *If* we used this, Example 28 would look like:

```json
  "image": {
     "id": "https:....",
     "digestSRI" : "sha384-......",
     "type" : "https://www.w3.org/ns/iana/media-types/application/svg+xml#Resource"
```

this can work, but I am afraid it would be a bit unnatural to use...

@msporny @dlongley ?

iherman:
> s/"`https://www.w3.org/ns/iana/media-types/application/svg+xml#Resource`"/"`https://www.w3.org/ns/iana/media-types/image/svg+xml#Resource`"/, I think?

Yep, sorry. I will edit the comment just to make it o.k.

> 
> (And, of course, the document at the latter link includes mention of `application/svg+xml` which is not listed in the main table on https://www.w3.org/ns/iana/media-types/. _sighs_)

iherman:
@decentralgabe yes. The issue of `mediaType` remaining is still pending, the other PR has been closed, so it should not affect this PR. Merge ahead...

decentralgabe:
editorial approval, open >1 week, changes addressed. merging!

Issue #1408, opened 2024-01-01
reconsider `@id` for `mediaType` term


gobengo:
Context:
* I'm trying to put W3C ActivityStreams2 inside the `credentialSubject` of a VC.
    * See this related issue I filed on JSON-LD, just to make sure I'm not missing some affordance that could be useful. https://github.com/w3c/json-ld-syntax/issues/424
* the as2 `mediaType` term conflicts with the *`@protected`* vcdm2 `mediaType` term
* I think this means that I cant have valid AS2 (which requires a certain jsonld compaction) inside of a VC. I can only re-compact the as2 mediaType property to something else like `as:mediaType`.

Questions
* Is anyone super attached to defining `mediaType` in terms of `https://schema.org/encodingFormat`?
* Could we entertain the notion of defining it by re-using as2 mediaType?

Straw Man
* https://github.com/w3c/vc-data-model/compare/main...gobengo:vc-data-model:patch-1

gobengo:
Looks like this was added in this PR https://github.com/w3c/vc-data-model/pull/1296 by @iherman. wdyt @iherman ?

iherman:
Just to make it clear for other reviewers in the WG: 

we are talking about making a choice, in the VCDM vocabulary, between two externally defined terms for our internal use, namely:

1. The [schema.org](https://schema.org) term [`encodingFormat`](https://schema.org/encodingFormat); or
2. The [activity streams'](https://www.w3.org/TR/activitystreams-core/) term [`mediaType`](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-mediatype)

Our usage of the term is defined in the [VCDM spec](https://www.w3.org/TR/vc-data-model-2.0/#media-types) that uses the term `mediaType`, mapped (via the `@context` mechanism) to the schema.org term. In other words, so far our choice was (1) above. The usage of the term within the VCDM is related to the [integrity control of related resources](https://www.w3.org/TR/vc-data-model-2.0/#integrity-of-related-resources).

[...]

PR #1440, opened 2024-02-15
Changed the term `mediaType` to `encodingFormat`

iherman:
As agreed on the [WG call](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-02-14-vcwg#section2-6) to handle issue #1408.

See also related comment in https://github.com/w3c/vc-data-model/pull/1438#pullrequestreview-1882041837.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1440.html" title="Last updated on Feb 15, 2024, 8:26 AM UTC (5a6e4ce)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1440/d42fd22...5a6e4ce.html" title="Last updated on Feb 15, 2024, 8:26 AM UTC (5a6e4ce)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3246,8 +3246,8 @@ There MUST NOT be more than one object in the `relatedResource` per
         </p>
         <p>
 An object in the `relatedResource` array MAY contain a property named
-`mediaType` that indicates the expected media type for the indicated
-`resource`. If a `mediaType` is included, its value
+`encodingFormat` that indicates the expected media type for the indicated
+`resource`. If a `encodingFormat` is included, its value
 SHOULD:
         </p>
         <ul>
@@ -3330,7 +3330,7 @@ integrity protected image.
     "id": "https://university.example.org/images/58473",
     "digestSRI":
       "sha384-ZfAwuJmMgoX3s86L7x9XSPi3AEbiz6S/5SyGHJPCxWHs5NEth/c5S9QoS1zZft+J",
-    "mediaType": "application/svg+xml",
+    "encodingFormat": "application/svg+xml",
   },
   ...
 }
--- a/contexts/credentials/v2
+++ b/contexts/credentials/v2
@@ -75,7 +75,7 @@
       "@type": "https://w3id.org/security#multibase"
     },
 
-    "mediaType": {
+    "encodingFormat": {
       "@id": "https://schema.org/encodingFormat"
     },
 


brentzundel (approved):

iherman:
@msporny I am not sure how the SHA value of the context lands in the spec (in §B.1). Is it automatic, or should I generate it before merge?

msporny:
> @msporny I am not sure how the SHA value of the context lands in the spec (in §B.1). Is it automatic, or should I generate it before merge?

It's not automatic, and we really, really, should add some code (to respec?) to make it automatic. Doing that has been on my to-do list for over a year. I was thinking of adding it to the respec-vc tool. /cc @davidlehn :)

But for now, feel free to re-generate the hash by hand (or just leave it alone until we automate the process).

iherman:
> > @msporny I am not sure how the SHA value of the context lands in the spec (in §B.1). Is it automatic, or should I generate it before merge?
> 
> It's not automatic, and we really, really, should add some code (to respec?) to make it automatic. Doing that has been on my to-do list for over a year. I was thinking of adding it to the respec-vc tool. /cc @davidlehn :)
> 
> But for now, feel free to re-generate the hash by hand (or just leave it alone until we automate the process).

I prefer not to do it, because I am not 100% sure I would produce the right value, and that would be worse. In a way, it does not matter so much; this is a new CR draft; what is important to get these values right by PR (there are also hashes for the vocabulary files).

index.html
```diff
@@ -3246,8 +3246,8 @@ <h2>Integrity of Related Resources</h2>
         </p>
         <p>
 An object in the `relatedResource` array MAY contain a property named
-`mediaType` that indicates the expected media type for the indicated
-`resource`. If a `mediaType` is included, its value
+`encodingFormat` that indicates the expected media type for the indicated
+`resource`. If a `encodingFormat` is included, its value
```
TallTed, line 3250:
```suggestion
`resource`. If an `encodingFormat` is included, its value
```

selfissued (changes requested):
I oppose changing from mediaType, which is a well-known term, to encodingFormat, which we'd be making up.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-02-21-vcwg#section3-2) on 2024-02-21

- no resolutions were taken

#### 3.2. Changed the term `mediaType` to `encodingFormat` (pr vc-data-model#1440)

_See github pull request [vc-data-model#1440](https://github.com/w3c/vc-data-model/pull/1440)._




**Ivan Herman:** We had a call on this issue. We decided to change the term "Media Type" to "Encoding Format".  

_See github issue [vc-data-model#1408](https://github.com/w3c/vc-data-model/issues/1408)._




**Ivan Herman:** Discussions are ongoing.  
… We shouldn't spend that much time on this.  

**Manu Sporny:** The goal is for activity streams to be able to use this without changes.  
… We may just be incompatible with the activity streams context.  
… If changing this one thing doesn't fix it, then we shouldn't make the change, since the problem wouldn't be addressed.  
… The way activity streams and schema.org define the context are neither right.  
… We probably don't want to do this.  
… Maybe we should define IANA media type and have it refer to the IANA registry.  
… I'm leaning towards that being my preference.  
… The downside is that we're creating yet another term.  

**Ivan Herman:** Note that the two things you mentioned are orthogonal to one another.  
… What term should we use?  
… Should we define it ourselves?  

> *Ted Thibodeau Jr.:* ianaMediaType was my idea, fwiw. its domain & range remain vital.

**Ivan Herman:** The definition of a data type for media types can be added.  
… It's not a huge deal.  
… I question altogether whether we should do.  

> *Dave Longley:* another option is to go with `encodingFormat` today and then potentially add `ianaMediaType` or `mediaType` in a future WG.

**Michael Jones:** It would be strange to change from a term that is well known "media type" to "encoding format", which we'd be entirely making up.  

**Ivan Herman:** We are not making it up, schema.org defined it.  

**Michael Jones:** That's not authoritative for us.  

**Ivan Herman:** That's debatable.  

**Manu Sporny:** Dmitri is on the call and is chairing the Social Web Community Group.  

**Dmitri Zagidulin:** We're the ones shepherding the activity streams formats.  
… We want to be able to sign activity streams objects.  

**Manu Sporny:** We shouldn't use schema.org.  
… We shouldn't use activity pub.  
… We should point to IETF and IANA and get this right once and for all.  
… It probably shouldn't go in our vocabulary.  
… It could go in our security vocabulary.  
… We should call it something that people understand.  

> *Ivan Herman:* See [IANA pointer](https://www.iana.org/assignments/media-types/media-types.xhtml).

**Ivan Herman:** I have put this pointer into the minutes.  
… From an RDF point of view, would the pointer be the URL of the property?  
… I don't really like that.  
… Instead we can define a media type for RDF.  
… This is where the string format is defined.  
… We define a property in one of our vocabularies.  

**Manu Sporny:** I wouldn't object to that.  
… But we'd be repeating what the social web and schema.org did and we'd be creating another property.  

**Ivan Herman:** I don't know exactly how activity streams defined it.  
… If its compatible with IANA, we could use it.  
… If Dmitri gives me a pointer to the definition, I could look at it.  

> *Manu Sporny:* It's defined here: [https://www.w3.org/TR/activitystreams-vocabulary/#dfn-mediatype](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-mediatype).

**Dmitri Zagidulin:** What's the objection to reusing the mediaType definition?  

**Manu Sporny:** They made it too specific to activity streams (by defining domain restrictions).  

**Dmitri Zagidulin:** That could be changed so it can be applied to any domain.  

**Ivan Herman:** That would be perfect.  

**Dmitri Zagidulin:** We could change that.  

**Michael Jones:** It's not clear to me, are we taking a dependence on an externally defined vocabulary?  

**Manu Sporny:** We already point to a bunch of externally defined vocabularies.  
… We'd be reusing the URL they use for the definition.  
… This would be more correct than using the schema.org encoding.  
… We could actually call this media type.  

**David Lehn:** ?Question?  
… Equivalency checking.  

> *Manu Sporny:* agree with Dmitri, I don't think this is an issue to re-use AS as long as it's aligned.

**David Lehn:** How much do people do full RDF processing?  

**Dmitri Zagidulin:** Zero.  

**Brent Zundel:** The proposal to raise an issue on the activity streams repository sounds right.  
… For our PR, the consensus is to not merge that PR.  

**Ivan Herman:** I'm happy to close it.  
… Who has the action to raise the PR in the right place?  

**Brent Zundel:** I'm willing to do it but I'm not sure I could accurately reflect what we want.  

**Ivan Herman:** I'm willing to do it.  

> *Michael Jones:* The activity streams repository is [https://github.com/w3c/activitystreams/issues/](https://github.com/w3c/activitystreams/issues/).

</details>

Issue #1331, opened 2023-10-29
Add validation to lifecycle details image


msporny:
This issue was raised by @jandrieu in PR #1297. The lifecycle details image needs to include the concept of validation.

jandrieu:
Whoops. We should reopen this.

The issue I created #1328 is different that this one.

TallTed:
I *think* a single dashed line (like the one now labeled "Delete" from Holders) going from Verifiers to upper-right, labeled as "Validate (repeatable)" would serve?

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-02-07-vcwg#section4-6) on 2024-02-07

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 4.6. Add validation to lifecycle details image (issue vc-data-model#1331)

_See github issue [vc-data-model#1331](https://github.com/w3c/vc-data-model/issues/1331)._




**Brent Zundel:** it was closed as completed but Joe re-opened it. I don't know what is going on with this PR.  

**Manu Sporny:** we don't have this in the image now. We have verify but we don;t have validate. This is not done.  

> *Manu Sporny:* See [the diagram's location](https://w3c.github.io/vc-data-model/#lifecycle-details).

**Brent Zundel:** It would be a loop from verifier to verifier that says validate?  

**Manu Sporny:** I don't know but I would expect something like that.  

**Brent Zundel:** Suggest we add a loop from verifiers to verifier that says validate. Anyone want to volunteer?  

**Ivan Herman:** I am not volunteering. but it may be a good opportunity to look at all the figures. We have problems if the document is seen in dark mode.  

**Brent Zundel:** I suggest opening an issue to track image dark-mode compatibility and raised separately.  

**Ivan Herman:** agreed.  

**Brent Zundel:** did not hear anyone volunteer. If no one volunteers, its not going to happen.  
… We are on track with a small extension. Pay atttention as meeting times may change. Closing the meeting.  

---

</details>

msporny:
PR #1441 was raised to address this issue, that PR has been merged, closing.

PR #1441, opened 2024-02-15
Added the "Validate" action to the life cycle details diagram (issue 1331)

iherman:
This is to handle #1331.

The usual preview is useless in this case; to review, you should look at the SVG file directly, see [svg link](https://raw.githack.com/w3c/vc-data-model/validation-on-diagram/diagrams/ecosystemdetail.svg).

- I completely re-did the diagram from scratch, because I do not know where the old source is (and I did not want to mess around with the SVG file itself). Also, the old diagram did not work with dark mode; hopefully this does (I tested it with Brave with Dark Mode switched on, and it looks o.k. there. @dlongley, you should have a look). This is also the reason why the diagram is less "colorful".
- Also, I removed the big title ("Life of a single Verifiable Credential") and I put this into the caption instead.

I am not sure what the 'semantics' of dotted lines is, so the new loop on Validate is just a continuous line.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1441.html" title="Last updated on Mar 3, 2024, 7:44 PM UTC (5fc78ef)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1441/d42fd22...5fc78ef.html" title="Last updated on Mar 3, 2024, 7:44 PM UTC (5fc78ef)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2754,9 +2754,10 @@ how the ecosystem is envisaged to operate.
          src="diagrams/ecosystemdetail.svg" alt="diagram showing how
          credentials flow from issuer to holder, and optionally
          from one holder to another; and how
-         presentations flow from holder to verifier, where
-         all parties can use information from a logical
-         verifiable data registry">
+         presentations flow from holder to verifier, 
+         optionally from one verifier to another through verification and validation.
+         All parties can use information from a logical
+         verifiable data registry. Holder may also delete credential.">
           <figcaption style="text-align: center;">
             Life of a single Verifiable Credential: the roles and information flows for this specification.
           </figcaption>
--- a/index.html
+++ b/index.html
@@ -2759,7 +2759,7 @@ how the ecosystem is envisaged to operate.
          All parties can use information from a logical
          verifiable data registry. Holder may also delete credential.">
           <figcaption style="text-align: center;">
-            Life of a single Verifiable Credential: the roles and information flows for this specification.
+            Lifecycle of a single Verifiable Credential: the roles and information flows for this specification.
           </figcaption>
         </figure>
 
--- a/index.html
+++ b/index.html
@@ -2758,7 +2758,7 @@ how the ecosystem is envisaged to operate.
          all parties can use information from a logical
          verifiable data registry">
           <figcaption style="text-align: center;">
-            The roles and information flows for this specification.
+            Life of a single Verifiable Credential: the roles and information flows for this specification.
           </figcaption>
         </figure>
 
--- a/diagrams/ecosystemdetail.drawio
+++ b/diagrams/ecosystemdetail.drawio
@@ -1,0 +1,185 @@
+<mxfile host="Electron" modified="2024-02-15T13:15:16.149Z" agent="Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) draw.io/23.0.2 Chrome/120.0.6099.109 Electron/28.1.0 Safari/537.36" etag="Vc7W3Dx77ieV5szVuuwu" version="23.0.2" type="device">
+  <diagram name="Page-1" id="6gULVhWcurzz5zKnDaAR">
+    <mxGraphModel dx="1653" dy="997" grid="1" gridSize="10" guides="1" tooltips="1" connect="1" arrows="1" fold="1" page="1" pageScale="1" pageWidth="1169" pageHeight="827" math="0" shadow="0">
+      <root>
+        <mxCell id="0" />
+        <mxCell id="1" parent="0" />
+        <UserObject label="&lt;font style=&quot;&quot;&gt;&lt;span style=&quot;font-size: 20px;&quot;&gt;Issuer&lt;/span&gt;&lt;/font&gt;" linkTarget="_blank" id="r7VOtmBWi9sTdI5Oyx6i-1">
+          <mxCell style="rounded=1;whiteSpace=wrap;html=1;fontSize=16;labelBackgroundColor=none;fillColor=#09555B;strokeColor=#FFFFFF;fontColor=#EEEEEE;" parent="1" vertex="1">
+            <mxGeometry x="210" y="403" width="125.25" height="58" as="geometry" />
+          </mxCell>
+        </UserObject>
+        <UserObject label="&lt;font style=&quot;&quot;&gt;&lt;span style=&quot;font-size: 20px;&quot;&gt;Registry&lt;/span&gt;&lt;/font&gt;" linkTarget="_blank" id="WkVDoU_EPi6P_C-XghKi-11">
+          <mxCell style="rounded=1;whiteSpace=wrap;html=1;fontSize=16;labelBackgroundColor=none;fillColor=#09555B;strokeColor=#FFFFFF;fontColor=#EEEEEE;" parent="1" vertex="1">
+            <mxGeometry x="530" y="593" width="125.25" height="58" as="geometry" />
+          </mxCell>
+        </UserObject>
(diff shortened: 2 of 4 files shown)


davidlehn:
- I think the current one may have been done fully or partly by hand?
- It seems problematic that there are three (or more?) styles of diagrams in the spec.  Would be better if they all used the same design language, style, fonts, weights, and role colors.
- The current one has better font distinction of actions and repeatability comments.  Can the action text here be bold or larger or something?  And perhaps the "might NOT preserve privacy" and "MIGHT preserve privacy" text could be in "(...)" or otherwise distinct so it does read as "MIGHT preserve privacy repeatable"

jandrieu:
Actually, I'd prefer to get rid of this image. It is not reflect most people's understanding of what a registry is. I expect this is largely because it was based on the Soverin approach where they put everything on chain.

However, in most instances:

1. You check status with a status list (not a registry)

2. You verify *at* the verifier, once you've resolved the controller document for the issuer's identifier. For DIDs, that mean you go resolve the DID Document. Non-DID identifiers need their own means for getting a controller document, perhaps using the confidenceMethod for subjects and maybe an allow list for acceptable issuers. However, that allow list is local, managed by the Verifier and NOT something that is expected to be outsourced (although allow lists are expected to be adopted from recognized authorities; those authorities however are not an interactive part of the flow as this "registry" appears to be.

3. The direct status check from the verifier to the issuer is a phone home problem and deserves retraction. We went to a lot of trouble to figure out how to avoid phone home. We should not advocate using a privacy-leaking flow in the specification.

4. Transfers are not a thing. Nowhere in the specification do we provide a mechanism for a user to transfer a VC to a different wallet (and presumably different crypto). We can present them, and we can, as a matter of fact, move the JSON files around, but just having the JSON VC does not mean you can satisfy proof-of-control during a Verifiable Presentation.

In short, I think this image is just out of date and should probably just be removed.

jandrieu (changes requested):
Please just delete this image. It's way out of date with current thinking.

iherman:
> * I think the current one may have been done fully or partly by hand?

No. It was done by the drawio tool, just like the large diagrams I have added recently. 

> * It seems problematic that there are three (or more?) styles of diagrams in the spec.  Would be better if they all used the same design language, style, fonts, weights, and role colors.

I agree it is a bit disturbing to have different styles but, actually, this was already the case. This image was always very different from the others.

> * The current one has better font distinction of actions and repeatability comments.  Can the action text here be bold or larger or something?  And perhaps the "might NOT preserve privacy" and "MIGHT preserve privacy" text could be in "(...)" or otherwise distinct so it does read as "MIGHT preserve privacy repeatable"

I am fine doing these changes and, maybe, removing the colours altogether to make the diagram look like, e.g., diagram 9. But I do not want to do it if the agreement is to follow @jandrieu's advice to remove it altogether and that is a WG choice, not mine.

iherman:
@brentzundel can we put this on the agenda for Wednesday? Not the detailed comments on the diagram (we can handle that off-line) but the answer on @jandrieu's proposal to remove it altogether (or not). If it is to be removed, then this PR is moot and a simple PR can take its place.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-02-21-vcwg#section3-1) on 2024-02-21

- no resolutions were taken

#### 3.1. Added the "Validate" action to the life cycle details diagram (issue 1331) (pr vc-data-model#1441)

_See github pull request [vc-data-model#1441](https://github.com/w3c/vc-data-model/pull/1441)._




**Ivan Herman:** I don't want to get into a discussion of the details of the diagram here.  
… The question is whether we want to keep it.  

**Manu Sporny:** We should keep it. I've referred to it.  

> *Manu Sporny:* lifecycle of VC: [https://w3c.github.io/vc-data-model/#concrete-lifecycle-example](https://w3c.github.io/vc-data-model/#concrete-lifecycle-example).

**Manu Sporny:** We should update the diagram and move it to the lifecycle section.  

**Gabe Cohen:** +1 to what Manu said.  
… I find the diagrams very useful.  
… More pretty pictures please.  

> *Manu Sporny:* +1 to "more pretty pictures please".

**Brent Zundel:** Where in the DM is the diagram currently?  

> *Manu Sporny:* See [current diagram](https://w3c.github.io/vc-data-model/#life-cycle-details).

> *Gabe Cohen:* we also have [https://www.w3.org/TR/vc-data-model-2.0/#concrete-lifecycle-example](https://www.w3.org/TR/vc-data-model-2.0/#concrete-lifecycle-example) without a diagram.

**Ivan Herman:** Let's keep it.  
… I propose that I edit the diagram to incorporate comments from David Lehn, etc.  
… Later we could move it, but I don't want to mix the two issues.  
… I'm happy to do the required changes.  

**Brent Zundel:** We should probably combine sections.  

> *Ivan Herman:* See [The new, temporary version of the diagram](https://raw.githack.com/w3c/vc-data-model/validation-on-diagram/diagrams/ecosystemdetail.svg).

> *Phillip Long:* pdl-asu has joined #vcwg.

**Manu Sporny:** We need to figure out what we're doing with these sections.  
… Let's start by updating the diagram first.  
… Then later editorially move things around.  

> *Ivan Herman:* +1 to manu.

**Brent Zundel:** The PR that modifies the diagram will be merged after Ivan updates it.  
… I will open a PR proposing combining sections.  

</details>

iherman:
I have made an update on the diagram, see the [new version](https://raw.githubusercontent.com/w3c/vc-data-model/validation-on-diagram/diagrams/ecosystemdetail.svg).

@davidlehn,

I have followed your advice on making the action text bold, and I put all secondary texts in one line to make the messages clearer.

I have removed the colours, so the diagram style is in line with other diagrams like Figure 9, 10, etc. The b/w approach (with a transparent background) make the diagrams appear properly on genuine dark mode as well (like the dark mode of the Brave browser).

That being said, older figures, like Figure 1, 2, 3 etc., are indeed of a different style. And... most of them fail on dark mode😒. My proposal is to re-do all those diagrams in the same style as well to make them all properly shown with dark mode on. I am happy to do that if there is a WG agreement on this. (cc @dlongley)

@decentralgabe, 

I have modified the label to be "Verify and Validate". Note that all the actions are verbs ("Transfer", "Delete", etc.) so using verbs at this point seems to be the right choice.

decentralgabe (approved):
an improvement!

index.html
```diff
@@ -2754,9 +2754,10 @@ <h3>Lifecycle Details</h3>
          src="diagrams/ecosystemdetail.svg" alt="diagram showing how
          credentials flow from issuer to holder, and optionally
          from one holder to another; and how
-         presentations flow from holder to verifier, where
-         all parties can use information from a logical
-         verifiable data registry">
+         presentations flow from holder to verifier,
```
TallTed, line 2757:
```suggestion
         presentations flow from holder to verifier, and
```

index.html
```diff
@@ -2754,11 +2754,12 @@ <h3>Lifecycle Details</h3>
...
-         presentations flow from holder to verifier, where
-         all parties can use information from a logical
-         verifiable data registry">
+         presentations flow from holder to verifier, 
+         optionally from one verifier to another through verification and validation.
+         All parties can use information from a logical
+         verifiable data registry. Holder may also delete credential.">
           <figcaption style="text-align: center;">
-            The roles and information flows for this specification.
+            Life of a single Verifiable Credential: the roles and information flows for this specification.
```
msporny, line 2762:
```suggestion
            Lifecycle of a single Verifiable Credential: the roles and information flows for this specification.
```

msporny (approved):

PR #1443, opened 2024-02-16
Proper rendering of sha-512 values

iherman:
In [§B.2](Implementations that depend on RDF vocabulary processing) the SHA3-512 values did not display correctly; the values were not rendered as a `<code>`. This PR removes the `<wbr>` tag which seems to have created problems in the respec->html conversion.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1443.html" title="Last updated on Feb 22, 2024, 8:51 AM UTC (e56dfaa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1443/d42fd22...e56dfaa.html" title="Last updated on Feb 22, 2024, 8:51 AM UTC (e56dfaa)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -6848,8 +6848,8 @@ https://www.w3.org/2018/credentials#<br>
               <td>
 https://www.w3.org/2018/credentials/index.jsonld<br><br>
 sha256: `z52TgKqh2nqTCuACI8lCvhRdjwxQjeVmuOMCDCEijq4=`<br><br>
-sha3-512: `m8Ss+jgZiyL2Ws/ICJcWjHFd9PccJWsXPvMatBOhrH
-h0qCBrzfgO2zO1OQQbTL7zoPgLseIbcxJJpunD2bkoRA==`
+sha3-512: <code>m8Ss+jgZiyL2Ws/ICJcWjHFd9PccJWsXPvMatBOhrH<wbr>
+h0qCBrzfgO2zO1OQQbTL7zoPgLseIbcxJJpunD2bkoRA==</code>
               </td>
             </tr>
             <tr>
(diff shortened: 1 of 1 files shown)


decentralgabe (approved):

msporny (approved):
Hmm, I put the WBR tags in there for a reason, but I can't remember why. It looks like it's rendering just fine w/o them, though. Approving. :)

TallTed:
This appears to me to be one of the places where mixing HTML and Markdown tagging doesn't work so well.

These snippets *look* fine without the `<wbr>` tags, but there's a space in each of the sha3-512 values, which I think should not be there, and which I think will cause problems with anyone who's expecting these values to be accurate ... and if they're not expected to be accurate, why expend the effort to calculate them?

I think that the `<wbr>` tags should be restored, and the linefeed that follows each of them should be removed. Probably the backticks wrapping the sha3-512 values should also be replaced by actual `<code>` tags.

If restoring the `<wbr>` tags without the following linefeeds and with the `<code>` tags replacing the backtick wrappers still results in a ReSpec issue, that demands an issue be raised against ReSpec.

iherman:
> 
> I think that the `<wbr>` tags should be restored, and the linefeed that follows each of them should be removed. Probably the backticks wrapping the sha3-512 values should also be replaced by actual `<code>` tags.

You are right, @TallTed. Restoring `<wbr>` and using `<code>` seems to work properly, and copying the SHA value does not add any extra space. Thanks. Changed it in https://github.com/w3c/vc-data-model/pull/1443/commits/e56dfaa11ad1aeff065110c8d0451af58efc22da.

decentralgabe:
@iherman is this ready to merge? if so, will do so after your OK

iherman:
As far as I am concerned yes, it can be merged

PR #1456, opened 2024-03-09
Registered Structured Syntax Suffix

selfissued:
This registers the structured suffix `+ld+json` so that the registrations of the media types `application/vc+ld+json` and `application/vp+ld+json` can proceed.

### DIFF
--- a/index.html
+++ b/index.html
@@ -7182,6 +7182,9 @@ This section will be submitted to the Internet Engineering Steering Group (IESG)
 for review, approval, and registration with IANA.
       </p>
 
+      <section>
+      <h2>Media Types</h2>
+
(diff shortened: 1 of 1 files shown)


OR13:
It's already been requested to be registered here:

https://w3c.github.io/json-ld-syntax/#structured-extension-ld-json

It's blocked pending the multiple suffixes draft, as are all media types including multiple plus signs.

msporny (changes requested):
As @OR13 noted, we don't need this PR since it's already been integrated into the JSON-LD specification.

selfissued:
This is unnecessary, since it's done elsewhere, as noted above.

PR #1458, opened 2024-03-13
(Editorial) Created a new version for several diagrams 

iherman:
I have re-created the older diagrams with two goals in mind:

- Have some consistency in the "visual language" of the older and newer diagrams (@davidlehn, this was one of your remarks in an older PR)
- All diagrams are now (I hope) dark-mode ready

What dark-mode readiness means:
- Diagrams have a transparent background
- The interior of the various figures (ellipses, rectangles, etc.) are "no-fill", i.e., they are all transparent
- Be careful about colors, because dark mode may mean transforming them into other colors

I have made _no_ change on the diagrams' content. Please, if you review this PR, do _not_ ask here to make a content change on the diagrams; raise a separate issue and I will be happy to take care of it in a separate PR.

All diagrams, except one, are edited using the drawio format. The PR includes both the drawio and the SVG files.

The default preview is useless, because it does not perform the include action. Use the [dedicated preview](https://raw.githack.com/w3c/vc-data-model/iherman-diagrams-for-dark-mode/index.html) for review.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1458.html" title="Last updated on Mar 13, 2024, 11:00 AM UTC (97ba9f0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1458/d41c9af...97ba9f0.html" title="Last updated on Mar 13, 2024, 11:00 AM UTC (97ba9f0)">Diff</a>

### DIFF
--- a/diagrams/claim-example.svg
+++ b/diagrams/claim-example.svg
@@ -1,29 +1,52 @@
-<svg version="1.1" fill="none" viewBox="0 0 300 50"
-    xmlns="http://www.w3.org/2000/svg">
-    <title>A simple example claim</title>
-    <desc>
-A subject, called "Pat", has a property "alumniOf", with the value "Example
-University". The claim expressed is that Pat attended the famous "Example U".
-    </desc>
-    <defs>
-      <style><![CDATA[
-          rect, ellipse {stroke-width: 1;stroke: #000}
-          .literal {font-family: serif}
-          line {stroke: #000;stroke-width: 2;marker-end: url(#arrow)}
-          text {fill: #000;font-size: 10px;font-family: sans-serif}
-          #background {stroke: none;fill: #fff;width:100%;height:100%}
-      ]]></style>
-        <marker id="arrow" viewBox="0 0 6 4" refX="4" refY="2" fill="#000"
-            markerWidth="6" markerHeight="4" orient="auto">
-            <path d="M0 0v4l6 -2z" />
-        </marker>
-    </defs>
-    <rect id="background"/>
-
-    <ellipse fill="#acf" rx="44" ry="20" cx="60" cy="25" />
-    <text x="40" y="24" transform="scale(1.25)">Pat</text>
-    <line x1="105" y1="25" x2="180" y2="25" />
-    <text x="110" y="20">alumniOf</text>
-    <rect x="185" y="15" width="110" height="25" fill="#fe9" />
-    <text class="literal" x="200" y="30">Example University</text>
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="-0.5 -0.5 612 82">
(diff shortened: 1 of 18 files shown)


dlongley (approved):
I looked at the dedicated preview at all of the diagrams in the spec and they all looked good to me from a dark mode perspective. Thanks!

decentralgabe (approved):

longpd (approved):

diagrams/claim.svg
davidlehn, line 1: Was this generated differently?  It has hardcoded width/height and is not indented like the other svgs.

diagrams/zkp-cred-pres.svg
davidlehn, line 1: The text is really close the lines.  Could use some padding.

diagrams/claim-example.svg
davidlehn, line 1: Some of the hand edited (?) SVGs had `title` and `desc` markup.  Good for accessibility I assume.  Does the new tooling have the ability to add such things?  Since this text is already written, would be nice to use it.  The spec markup uses `alt` and `figcaption`, which might be enough in any case.

diagrams/ecosystem.svg
davidlehn, line 1: - Center content and/or shrink the bounding box.  It's offset right and shows up that way in the rendered doc too.  (Note the old version looks offset to the left.)
- Please align the big arrows "Issue Credentials" and "Send Presentation".  The "Send" one is visibly lower.
- The center "Register Identifiers"/"and use Schemas" doesn't have the same font wrapper.  It inherits 16px vs the outside two that are set at 17px.  (Apparently I can notice a 1px font difference, and checked the SVG.)

diagrams/claim-example.svg
iherman: None of the files are hand edited, except for the privacy-spectrum one (that was the only one that, with just a manual change removing the background, was dark-mode ready). Unfortunately (to my spite) the tool does not create any metadata in the svg files. It could be added manually, but that does not scale, so I am forced to leave this as is...

diagrams/ecosystem.svg
iherman: You do have eagle eyes:-) Yes, it was 16px instead of 17px. Changed. The other two issues, too.

diagrams/zkp-cred-pres.svg
iherman: It turned out that there is an error in the svg formatting/processing script I use (called svgo) and the spaces put at the beginning of a text were swallowed. The problem was not only the missing padding, but there was a minimal structure, through indentation, in the original image. I have changed the way those textual contents are created and it should be o.k. now.

msporny (approved):

msporny:
Editorial, multiple positive reviews, all changes that can be made have been made, no objections, merging.

Issue #1348, opened 2023-11-12
Non-normative changes from Jeffrey Yasskin's review


msporny:
This issue tracks the NON-normative changes requested from @jyasskin's review (issue https://github.com/w3c/vc-data-model/issues/1285) of the VC v2.0 specification:

#### 1. Introduction

- [ ] Use property terminology throughout entire specification.

#### 1.1 What is a Verifiable Credential?

- [ ] Specify which properties exist on a VC.

#### 1.3 Use Cases and Requirements

- [ ] Use non-normative wording instead of "should".

#### 3.2 Credentials

- [ ] Mention quoting and verifiable credential graph.

#### 4.1 Getting Started

- [ ] Make section non-normative.

#### 4.2 Contexts

- [ ] Use better language to specify ordered sets.
- [ ] Remove redundant "This specification requires for a @context property to be present;" statement.
- [ ] Rewrite paragraph that contains "Verifiable credentials and verifiable presentations have many attributes and values that are identified by URLs" -- re-write paragraph to introduce contexts in a different way.

#### 4.3 Identifiers

- [ ] Merge the id and the "If the id property is present:" block.

#### 4.4 Types

- [ ] Clarify that you can use a full URL, or you can use a JSON-LD term that maps to a URL. That is, you can't use a number, string, object, etc. 
- [ ] Point to VC-SPECS for some examples of "A valid proof type".
- [ ] Clean up and reword section containing "This enables implementers to rely on values associated with the type property".

#### 5.1 Lifecycle Details

- [ ] Add text on what a "verifiable data registry" and how it is used.
- [ ] Include operations that can be performed in ecosystem.

#### 5.2 Trust Model

- [ ] Wordsmith statements to say is that the verifier trusts the issuer to make the statements it's making after verification has been performed.
- [ ] Elaborate on why entities trust/depend on verifiable data registry.
- [ ] Explain why holder trusts the issuer.
- [ ] Say something about the verifier choosing the issuers they want to trust.

#### 5.3.1 Semantic Interoperability

- [ ] "is expected to be published" diffuses responsibility. Use active voice.

#### 5.4 Integrity of Related Resources

- [ ] Editorial clean-up and simplification of normative statements to match other sections.

#### 5.5 Refreshing

- [ ] Add some text in https://w3c.github.io/vc-data-model/#presentations to say that issuers and verifiers can use presentations to prove who they are to a holder by providing their own credentials via a presentation.

#### 5.11 Ecosystem Compatibility

[...]

Issue #1314, opened 2023-10-11
Default context contains terms outside of the spec


mtaimela:
Hi all.

Currently the https://www.w3.org/ns/credentials/v2 context contains terms regarding the "JWT standard claims" and "SD-JWT claims". These are actually not defined by the VCDM2.0 specification, so I think these should not belong into VCDM2.0 default context.

Could it be possible to remove these, and ask the other specifications to include VCDM2.0 standard claims into their own context definitions. It is not sustainable to include all possible terms from all possible containers into VCDM2.0 default context.

@OR13 I think you might have a strong opinions here, but from objects relation view point the current implementation is inverted. SD-JWT "has-a" VCDM2.0, current context implementation says VCDM2.0 "is-a" SD-JWT and JWT from context mapping.

TL;DR;
1. VCDM2.0 default context should focus into own specification boundaries, remove excess.
1. SD-JWT should define their own terms, and include VCDM2.0 context
2. JWT should define their own terms, and include VCDM2.0 context

This way, the relations are correct, and lifecycles are with-in the specs.

mtaimela:
Hi @brentzundel,

thank you for the historical information. Embedding other specifications vocabs into VCDM2.0 context is clearly violating the architectural boundaries. These boundaries have different lifecycle, thus non-backward compatible changes or conflicts in naming, will propagate up to the dependency tree and cause a new version in the VCDM2.0 context that will destroy all existing VCs.

I fully understand the "let's be nice and accommodate everything and save space", but this approach is clearly wrong from the boundaries, dependencies, and their lifecycle point of view.

I still would like to keep this issue. One option is to remove the "required default context" and say that theVCDM2.0  @vocab "must be included in the chosen context". This way the DIP, JWT and SD-JWT can embed the version of their choice, while having full control of their own terms. Dependencies are correct (all depends on VCDM2.0 vocab), and specs lifecycle does not interfere with the main spec.

dlongley:
The main driver here is the presence of `@vocab` in the core context, which results in erroneously defined terms when people forget to include other contexts.

An option would be to shift that burden onto only those that want to use `@vocab` instead of all VC v2 consumers. If `@vocab` were removed from the core context and instead the current default `@vocab` was provided in a utility context (that could be used by devs running experiments or issuers that don't want to define or reuse other contexts), that would enable the decoupling sought here. This has been proposed before, but there was not consensus to do it.

OR13:
There are controlled by W3C:

```
[...]

Issue #1089, opened 2023-04-17
Why does the Data Model context file define a DataIntegrityProof RDF class?


brentzundel:
Now that Data Integrity is it's own specification, the DataIntegrityProof RDF class in the base data model context should be moved to the Data Integrity context

OR13:
Here are some related contexts we have historically used for data integrity:

- https://w3id.org/security/v2
- https://w3id.org/security/v1

Here is the vocab term definitions for data integrity RDF Class:

- https://w3id.org/security#DataIntegrityProof

Here is the complete security vocabulary:

- https://w3id.org/security

IMO the vc-data-model core context should be restricted to things you MUST understand to implement the spec.

However, currently the URL is normative: https://www.w3.org/ns/credentials/v2

While its contents (resolved content) is not.

Here are a few things that are sorta "implied to be normative" because of this:

- https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v2#L51
- https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v2#L95
- https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v2#L149

Additionally, "data integrity proofs" can be used without "w3c verifiable credentials" to secure arbitrary JSON... so it makes sense that there should a *some other context out there* that includes https://w3id.org/security#DataIntegrityProof but does not include https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v2#L9 .

It seems like it might be easier for us to move some of the "proof" related stuff to data integrity and refer to it from the core data model, as opposed to defining it in the core data model normatively... and then actually maintaining it in data integrity and related vocabs:

> Conformance to this specification does not depend on the details of a particular proof mechanism; it requires clearly identifying the mechanism a [verifiable credential](https://w3c.github.io/vc-data-model/#dfn-verifiable-credential) uses.

https://w3c.github.io/vc-data-model/#conformance

This is sorta not true, if the v2 context is normatively required:

https://w3c.github.io/vc-data-model/#contexts

AND It resolves to a non-normative JSON-LD context object that directly includes:

- https://w3id.org/security#proof
- https://w3id.org/security#DataIntegrityProof 

It would be better to use a second context to add support for "data integrity proofs" to a VC.

This second context would also be useful for applying data integrity proofs to JSON-LD that is not a VC.

This second context can of course be built at any point in the future, so we have no need to bundle these data integrity proof specific terms in the current core data model v2 context.

[...]

PR #1464, opened 2024-03-23
Address Jeffrey Yasskin's review comments that are Editorial

msporny:
This PR is an attempt to address issue #1348 by fixing a large number of editorial concerns that @jyasskin raised. 

Specifically:

#### 1. Introduction

- [x] Use property terminology throughout entire specification. Fixed in c36dd64403f6c50b324a65a4d1ed762c54245068.

#### 1.1 What is a Verifiable Credential?

- [x] Specify which properties exist on a VC. Fixed in 42d38f222469159fa65eb641e1484e0de3a4487b.

#### 1.3 Use Cases and Requirements

- [x] Use non-normative wording instead of "should". This section has been removed by the specification.

#### 3.2 Credentials

- [x] Mention quoting and verifiable credential graph. Fixed in 6938250f25f9d5e7fc66b38955a310c5468e6fb4.

#### 4.1 Getting Started

- [x] Make section non-normative. This was done in a previous PR.

#### 4.2 Contexts

- [x] Use better language to specify ordered sets. Fixed to use INFRA in 19200373bda8f3595b10c99457fa6c7f7d112c65.
- [x] Remove redundant "This specification requires for a @context property to be present;" statement. Fixed in 52562ecdddcdf9e4701c3bfd15da3c4a5d924737.
- [x] Rewrite paragraph that contains "Verifiable credentials and verifiable presentations have many attributes and values that are identified by URLs" -- re-write paragraph to introduce contexts in a different way. Rewritten in 19200373bda8f3595b10c99457fa6c7f7d112c65.

#### 4.3 Identifiers

- [x] Merge the id and the "If the id property is present:" block. Fixed in 7a96f9e222834b5441c45a9ed24fa27cc8479db1.

#### 4.4 Types

- [x] Clarify that you can use a full URL, or you can use a JSON-LD term that maps to a URL. That is, you can't use a number, string, object, etc. Fixed in a45bb7b955bb3a7c03aa4770f78fbd18155cd523.
- [x] Point to VC-SPECS for some examples of "A valid proof type". Discussion of `proof` was removed from this section in a previous PR.
- [x] Clean up and reword section containing "This enables implementers to rely on values associated with the type property". Attempted clean up in c07066f7f5531783518a3f3458fa0f1595fdad62.

#### 5.1 Lifecycle Details

- [x] Add text on what a "verifiable data registry" and how it is used. Addressed in PR #1476.
- [x] Include operations that can be performed in ecosystem. Addressed in PR #1476.

#### 5.2 Trust Model

- [x] Wordsmith statements to say is that the verifier trusts the issuer to make the statements it's making after verification has been performed. Addressed in PR #1469.
- [x] Elaborate on why entities trust/depend on verifiable data registry. Addressed in PR #1469.
- [x] Explain why holder trusts the issuer. Addressed in PR #1469.
- [x] Say something about the verifier choosing the issuers they want to trust. Addressed in PR #1469.

#### 5.3.1 Semantic Interoperability

- [x] "is expected to be published" diffuses responsibility. Use active voice. Fixed in 024994aaf792c2c64a8b2a1bef0577b3d71bca19.

#### 5.4 Integrity of Related Resources

- [x] Editorial clean-up and simplification of normative statements to match other sections. Fixed in b3ac23cafe92eb92980b8c2d4612d0cdd9908b96 and e0c09880ba284d9da44d23b5c151fdf4bf113f7b.

#### 5.5 Refreshing

- [x] Add some text in https://w3c.github.io/vc-data-model/#presentations to say that issuers and verifiers can use presentations to prove who they are to a holder by providing their own credentials via a presentation. Fixed in f4b1e2553cd3e2f337fed599f10000db013a885d.

#### 5.11 Ecosystem Compatibility

- [x] Clarify intent of "rules" in section to be guidance to specification authors that desire compatability. Fixed in df1a45e183e78e73745f5c7a2fc73aa24459c446.

#### 6.3 Proof Formats

- [x] Change "The sections detailing" to "The specifications detailing". The section has been rewritten in a previous PR and the problematic text was removed in the rewrite.

#### 7. Privacy Considerations

- [x] "Implementers of software utilized by holders". This text no longer appears in the specification due to a previous PR.

#### 7.3 Personally Identifiable Information

- [x] Update "Implementers" to "Implementers of software utilized by holders". Fixed in 63d14639073df0c18318112ae3e392e44540fff3.
- [x] Say that people should assume a VC will leak PII unless the credential and securing mechanism is designed specifically to avoid correlation (and there are credential types designed specifically for this purpose). Fixed in cd6116ce4d1110ad682ade5ff2cb73b3a94b73f2.

#### 7.4 Identifier-Based Correlation

- [x] Specify that proof specifications are responsible for avoiding identifier-based correlation. Fixed in 69af3d52632df069e52007121409468c55394f0f.

#### 7.5 Signature-Based Correlation

- [x] Specify who is responsible for signature-based correlation and update section to match modern practices. Fixed in 

#### 7.6 Long-Lived Identifier-Based Correlation

- [x] Fold this section into the "Identifier-based Correlation" section and provide a refresh on the text. Fixed in 1c9ea793803687427e304cccbde8d5f27eef881b.
- [x] Say credential repository software needs to provide protections. Fixed in 1c9ea793803687427e304cccbde8d5f27eef881b.

#### 7.8 Favor Abstract Claims

- [x] Specify who should favor abstract claims -- "issuers should favor abstract claims". Fixed in bdc9c448941a0b09a1305dae314ed286d187252a.

#### 7.9 The Principle of Data Minimization

- [x] Specify that this is advice for issuers and verifiers. Remind credential repositories to disclose what fields the verifier is asking for, so that users can push back. Fixed in aca01b093e1b18e9d500c22572b6e31042fa72b5.

#### 7.11 Validation

- [x] Possibly move section into larger Validation section? Renamed section in 4ee5f171d0cdbb6af934867bef5e189d402234e2.

#### 7.12 Storage Providers and Data Mining

- [x] Say that civil society groups and regulators should play a part in this ecosystem (by doing evaluations and monitoring). Fixed in de2dec680a3b0804cc4f3889f60509b6ce23fb26.

#### 7.14 Usage Patterns

- [x] Localize mitigations to particular actors by saying things like "Authors of specifications for credentialStatus types", and "Using a globally-unique identifier" could be guidance for credential repositories to warn uses when they're re-using an ID. Fixed in 4dc6700220b51fc2929f90b51065f8444298bb7b.

#### 8.3 Content Integrity Protection

- [x] Update outdated text and speak to new `relatedResource` feature. Fixed in 8d9a3159f2b986e46fc15738fc05fdd5c2661d26.

#### A.1 Credential Type

- [x] Specify that the documents at the type's URL are expected to describe the available properties and their usage. The language for this was put in the Types section in  c07066f7f5531783518a3f3458fa0f1595fdad62.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1464.html" title="Last updated on May 5, 2024, 3:01 PM UTC (150bfbb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1464/1153f74...150bfbb.html" title="Last updated on May 5, 2024, 3:01 PM UTC (150bfbb)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1292,30 +1292,23 @@ Software systems that process the kinds of objects specified in this document
 use type information to determine whether or not a provided
 [=verifiable credential=] or [=verifiable presentation=] is appropriate
 for the intended use case. This specification defines a `type`
-[=property=] for the expression of type information. This type information
-can be used during [=validation=] processes as described in Appendix
-<a href="#validation"></a>.
+[=property=] for the expression of object type information. This type
+information can be used during [=validation=] processes as described in Appendix
+[[[#validation-0]]].
         </p>
 
         <p>
-[=Verifiable credentials=] and [=verifiable presentations=] MUST have a
-`type` [=property=]. That is, any [=credential=] or
-[=presentation=] that does not have `type` [=property=]
-<em>is not [=verifiable=]</em>, so is neither a [=verifiable credential=]
-nor a [=verifiable presentation=].
+[=Verifiable credentials=] and [=verifiable presentations=] MUST express a
+`type` [=property=] with an associated value.
         </p>
 
         <dl>
           <dt><dfn class="export" data-lt="type|types">type</dfn></dt>
           <dd>
-The value of the `type` [=property=] MUST be, or map to (through
-interpretation of the `@context` property), one or more [=URLs=].
-If more than one [=URL=] is provided, the [=URLs=] MUST be interpreted
-as an unordered set. Syntactic conveniences SHOULD be used to ease developer
-usage. Such conveniences might include JSON-LD terms. It is RECOMMENDED that
-each [=URL=] in the `type` be one which, if dereferenced, results
-in a document containing machine-readable information about the
-`type`.
+The value of the `type` [=property=] MUST be one or more
+<a href="https://www.w3.org/TR/json-ld/#dfn-term">terms</a> and/or absolute
+[=URLs=] (<a data-cite="URL#absolute-url-string">absolute url string</a>). If
+more than one value is provided, the order does not matter.
           </dd>
         </dl>
 
@@ -1392,7 +1385,7 @@ A valid [=credential=] status [=type=]. For example,<br>
               </td>
               <td>
 A valid terms of use [=type=]. For example,<br>
-`"type": "ExampleTermsPolicy"`)
+`"type": "ExampleTermsPolicy"`
               </td>
             </tr>
 
@@ -1460,11 +1453,10 @@ Title of the degree in the `name` property.
         </ul>
 
         <p>
-This enables implementers to rely on values associated with the
-`type` property for [=verification=] purposes. The expectation of
-[=types=] and their associated properties should be documented in at least a
-human-readable specification, and preferably, in an additional machine-readable
-representation.
+This enables implementers to rely on values associated with the `type` property
+for [=verification=] purposes. [=Types=], and their associated properties, are
+expected to be documented in at least a human-readable specification, and
+preferably, in an additional machine-readable representation.
         </p>
 
         <p class="note">
--- a/index.html
+++ b/index.html
@@ -3098,17 +3098,23 @@ is also desirable to know that the contents of the JSON-LD context(s) used in
 the [=verifiable credential=] are the same when used by both the [=issuer=] and
 [=verifier=].
         </p>
+
+        <p class="issue" title="Mandatory listing of contexts in relatedResouce are under debate.">
+The requirement that contexts be listed in `relatedResource` is currently being
+debated in the VCWG. This requirement might be removed in future iterations of
+the specification.
+        </p>
+
         <p>
-To extend integrity protection to a related resource, a
-[=verifiable credential=], an author MAY include the `relatedResource`
-property:
+To extend integrity protection to a related resource, an [=issuer=] of a
+[=verifiable credential=] MAY include the `relatedResource` property:
         </p>
 
         <dl>
           <dt id="defn-relatedResource">relatedResource</dt>
           <dd>
 The value of the `relatedResource` property MUST be associated with one or
-more objects containing the following values:
+more objects of the following form:
             <table class="simple">
               <thead>
                 <th>Property</th>
@@ -3118,9 +3124,9 @@ more objects containing the following values:
                 <tr>
                   <td>`id`</td>
                   <td>
-The REQUIRED identifier for the resource, as defined in Section
-[[[#identifiers]]]. The value MUST be unique among the list of related
-resource objects.
+The identifier for the resource is REQUIRED and conforms to the format defined
+in Section [[[#identifiers]]]. The value MUST be unique among the list of
+related resource objects.
                   </td>
                 </tr>
                 <tr>
@@ -3145,16 +3151,10 @@ A cryptographic digest, as defined in [[[VC-DATA-INTEGRITY]]].
                 </tr>
               </tbody>
             </table>
-Each value associated with `relatedResource` MUST contain at least a `digestSRI`
-or `digestMultibase` value.
+Each object associated with `relatedResource` MUST contain at least a
+`digestSRI` or `digestMultibase` value.
           </dd>
         </dl>
-
-        <p class="issue" title="Mandatory listing of contexts in relatedResouce are under debate.">
-The requirement that contexts be listed in `relatedResource` is currently being
-debated in the VCWG. This requirement might be removed in future iterations of
-the specification.
-        </p>
 
         <p class="issue" title="Unification of cryptographic hash expression formats are under discussion">
 The Working Group is currently attempting to determine whether cryptographic hash
@@ -3165,7 +3165,7 @@ are arguments for and against unification that the WG is currently debating.
 
         <p>
 If a `mediaType` is listed, implementations that retrieve the resource
-over [[[?HTTP]]] SHOULD:
+using [[[?RFC9110]]] SHOULD:
         </p>
         <ul>
           <li>
@@ -3177,7 +3177,7 @@ use the media type in the `Content-Type` HTTP Header.
         </ul>
 
         <p>
-Any object in the [=verifiable credential=] that contains an `id` [[URL]]
+Any object in the [=verifiable credential=] that contains an `id`
 property MAY be annotated with integrity information as specified in this
 section.
         </p>
--- a/index.html
+++ b/index.html
@@ -798,8 +798,8 @@ time period, a representative image, [=verification material=], status
 information, and so on. A
 [=verifiable credential=] is a set of tamper-evident [=claims=] and metadata
 that cryptographically prove who issued it. Examples of [=verifiable
-credentials=] include digital employee identification cards, digital birth
-certificates, and digital educational certificates.
+credentials=] include, but are not limited to, digital employee identification
+cards, digital birth certificates, and digital educational certificates.
         </p>
 
         <figure id="basic-vc">
@@ -890,10 +890,11 @@ arrow, to a separate rectangle, containing a single text field:
         </figure>
 
         <p class="note">
-[=Credential=] identifiers are often used to identify specific instances
-of a [=credential=]. These identifiers can also be used for correlation. A
-[=holder=] wanting to minimize correlation is advised to use a selective
-disclosure scheme that does not reveal the [=credential=] identifier.
+[=Credential=] identifiers can be used to identify specific instances
+of a [=credential=]. These identifiers could also be used for unwanted
+correlation. A [=holder=] wanting to minimize correlation is advised
+to use a selective disclosure scheme that does not reveal the
+[=credential=] identifier.
         </p>
 
         <p class="note">
@@ -1303,12 +1304,12 @@ use type information to determine whether or not a provided
 [=verifiable credential=] or [=verifiable presentation=] is appropriate
 for the intended use case. This specification defines a `type`
 [=property=] for the expression of object type information. This type
-information can be used during [=validation=] processes as described in Appendix
+information can be used during [=validation=] processes, as described in Appendix
 [[[#validation]]].
         </p>
 
         <p>
-[=Verifiable credentials=] and [=verifiable presentations=] MUST express a
+[=Verifiable credentials=] and [=verifiable presentations=] MUST contain a
 `type` [=property=] with an associated value.
         </p>
 
@@ -1479,7 +1480,7 @@ the same URL.
         <p class="note" title="Creating new verifiable credential types">
 Explaining how to create new types of [=verifiable credentials=] is beyond
 the scope of this specification. Readers that are interested in doing so are
-advised to read the Section on
+advised to read the section on
 <a data-cite="?VC-IMP-GUIDE#creating-new-credential-types">
 Creating New Credential Types</a> in the [[[?VC-IMP-GUIDE]]].
         </p>
@@ -3192,7 +3193,7 @@ Specification authors that write algorithms that fetch a resource based on the
 `id` of an object inside a [=conforming document=] need to consider whether
 that resource's content is vital to the validity of that document. If it is, the
 specification MUST produce a validation error unless the resource matches the
-expected media type and cryptographic digest
+expected media type and cryptographic digest.
         </p>
         <p>
 Implementers are urged to consult appropriate sources, such as the
@@ -5090,7 +5091,7 @@ credential=].
         <p>
 [=Securing mechanism=] specification authors are advised to avoid enabling
 identifier-based correlation by designing their technologies, when possible,
-to avoid the use of correlating identifiers.
+to avoid the use of correlating identifiers that cannot be selectively disclosed.
         </p>
 
         <p>
--- a/index.html
+++ b/index.html
@@ -5211,59 +5211,65 @@ make purchases without revealing more PII than necessary.
 
         <p>
 Privacy violations occur when information divulged in one context leaks into
-another. Accepted best practice for preventing such violations is to limit the
-information requested, and received, to the absolute minimum necessary. This
-data minimization approach is required by regulation in multiple jurisdictions,
-including the Health Insurance Portability and Accountability Act (HIPAA) in the
-United States and the General Data Protection Regulation (GDPR) in the European
-Union.
+another. One accepted best practice for preventing such a violation is for
+[=verifiers=] to limit the information requested, and received, to the absolute
+minimum necessary for a particular transaction. This data minimization approach
+is required by regulation in multiple jurisdictions, including the Health
+Insurance Portability and Accountability Act (HIPAA) in the United States and
+the General Data Protection Regulation (GDPR) in the European Union.
         </p>
         <p>
 With [=verifiable credentials=], data minimization for [=issuers=] means
-limiting the content of a [=verifiable credential=] to the minimum required
-by potential [=verifiers=] for expected use. For [=verifiers=], data
-minimization means limiting the scope of the information requested or
-required for accessing services.
+limiting the content of a [=verifiable credential=] to the minimum required by
+potential [=verifiers=] for expected use. For [=verifiers=], data minimization
+means limiting the scope of the information requested or required for accessing
+services.
         </p>
         <p>
 For example, a driver's license containing a driver's ID number, height, weight,
-birthday, and home address is a [=credential=] containing more information
-than is necessary to establish that the person is above a certain age.
+birthday, and home address expressed as a [=verifiable credential=] contains
+more information than is necessary to establish that the person is above a
+certain age.
         </p>
 
         <p>
-It is considered best practice for [=issuers=] to atomize information or use
-a signature scheme that allows for [=selective disclosure=]. For example, an
+It is considered best practice for [=issuers=] to atomize information or use a
+securing mechanism that allows for [=selective disclosure=]. For example, an
 [=issuer=] of driver's licenses could issue a [=verifiable credential=]
-containing every property that appears on a driver's license, as well as a set
-of [=verifiable credentials=] where every [=verifiable credential=]
-contains only a single property, such as a person's birthday. It could also
-issue more abstract [=verifiable credentials=] (for example, a
-[=verifiable credential=] containing only an `ageOver` property).
-One possible adaptation would be for [=issuers=] to provide secure HTTP
-endpoints for retrieving single-use [=bearer credentials=] that promote the
-pseudonymous usage of [=verifiable credentials=]. Implementers that find this
-impractical or unsafe, should consider using [=selective disclosure=] schemes
-that eliminate dependence on [=issuers=] at proving time and reduce temporal
-correlation risk from [=issuers=].
+containing every property that appears on a driver's license and allow each
+property to be selectively disclosed by the [=holder=]. It could also issue more
+abstract [=verifiable credentials=] (for example, a [=verifiable credential=]
+containing only an `ageOver` property). One possible adaptation would be for
+[=issuers=] to provide secure HTTP endpoints for retrieving single-use [=bearer
+credentials=] that promote the pseudonymous usage of [=verifiable credentials=].
+Implementers that find this impractical or unsafe, might consider using
+[=selective disclosure=] schemes that eliminate dependence on [=issuers=] at
+proving time and reduce temporal correlation risk from [=issuers=].
         </p>
 
         <p>
 [=Verifiers=] are urged to only request information that is absolutely
 necessary for a specific transaction to occur. This is important for at least
-two reasons. It:
+two reasons.
         </p>
 
         <ul>
           <li>
-Reduces the liability on the [=verifier=] for handling highly sensitive
-information that it does not need to.
+It reduces the liability on the [=verifier=] for handling highly sensitive
+information that it does not need to, and
           </li>
           <li>
-Enhances the privacy of the individual by only asking for information required
-for a specific transaction.
+it enhances the privacy of the individual by only asking for information
+required for a specific transaction.
           </li>
         </ul>
+
+        <p>
+Implementers of software used by [=holders=] are urged to disclose what
+information is being requested by a [=verifier=] such that a [=holder=] can
+push back on the over-collection of information that is unnecessary for the
+transaction.
+        </p>
 
         <p class="note">
 While it is possible to practice the principle of minimum disclosure, it might
--- a/index.html
+++ b/index.html
@@ -5023,38 +5023,54 @@ to protect personally identifiable information.
         <h3>Identifier-Based Correlation</h3>
 
         <p>
+[=Verifiable credentials=] might contain long-lived identifiers that could be
+used to correlate individuals. These types of identifiers include [=subject=]
+identifiers, email addresses, government-issued identifiers, organization-issued
+identifiers, addresses, healthcare vitals, and many other sorts of long-lived
+identifiers. Implementers of software used by [=holders=] are advised to strive
+to detect identifiers in [=verifiable credentials=] containing information that
+could be used to correlate individuals and warn [=holders=] when they are
+getting ready to share this information. The rest of this section elaborates
+on guidance related to the use of long-lived identifiers.
+        </p>
+
+        <p>
 [=Subjects=] of [=verifiable credentials=] are identified using the `id`
 property, as defined in Section [[[#identifiers]]], and are used in places such
 as the `credentialSubject.id` property. The identifiers used to identify a
 [=subject=] create a greater risk of correlation when the identifiers are
-long-lived or used across more than one web domain.
+long-lived or used across more than one web domain. Other types of identifiers
+that fall into this category are email addresses, government-issued identifiers,
+and organization-issued identifiers.
         </p>
 
         <p>
 Similarly, disclosing the [=credential=] identifier (such as in
-[[[#example-usage-of-the-id-property]]]) leads to
-situations where multiple [=verifiers=], or an [=issuer=] and a [=verifier=],
-can collude to correlate the [=holder=]. If [=holders=] want to reduce
-correlation, they are advised to use [=verifiable credentials=] from [=issuers=]
-that allow selectively disclosing correlating identifiers in a [=verifiable
-presentation=]. Such approaches expect the [=holder=] to generate the identifier
-and might even allow hiding the identifier from the [=issuer=] through the use
-of techniques like
+[[[#example-usage-of-the-id-property]]]) leads to situations where multiple
+[=verifiers=], or an [=issuer=] and a [=verifier=], can collude to correlate the
+[=holder=].
+        </p>
+
+        <p>
+If [=holders=] want to reduce correlation, they are advised to use [=verifiable
+credentials=] from [=issuers=] that allow selectively disclosing correlating
+identifiers in a [=verifiable presentation=]. Such approaches expect the
+[=holder=] to generate the identifier and might even allow hiding the identifier
+from the [=issuer=] through the use of techniques like
 <a href="https://en.wikipedia.org/wiki/Blind_signature">blind signatures</a>,
 while still keeping the identifier embedded and signed in the [=verifiable
 credential=].
         </p>
 
         <p>
-Securing mechanism specification authors are advised to avoid enabling
+[=Securing mechanism=] specification authors are advised to avoid enabling
 identifier-based correlation by designing their technologies, when possible,
 to avoid the use of correlating identifiers.
         </p>
 
         <p>
-If strong anti-correlation properties are a requirement in a
-[=verifiable credentials=] system, it is strongly advised that identifiers
-are either:
+If strong anti-correlation properties are a requirement in a [=verifiable
+credentials=] system, it is strongly advised that identifiers are either:
         </p>
 
         <ul>
@@ -5116,26 +5132,6 @@ properties of the cryptography used. See Sections
 [[[#long-lived-identifier-based-correlation]]],
 [[[#metadata-based-correlation]]], [[[#correlation-during-validation]]],
 and most of the other sections in Section [[[#privacy-considerations]]].
-        </p>
-      </section>
-
-      <section class="informative">
-        <h3>Long-Lived-Identifier-Based Correlation</h3>
-
-        <p>
-[=Verifiable credentials=] might contain long-lived identifiers that could
-be used to correlate individuals. These types of identifiers include
-[=subject=] identifiers, email addresses, government-issued identifiers,
-organization-issued identifiers, addresses, healthcare vitals,
-[=verifiable credential=]-specific JSON-LD contexts, and many other sorts of
-long-lived identifiers.
-        </p>
-
-        <p>
-Organizations providing software to [=holders=] should strive to identify
-fields in [=verifiable credentials=] containing information that could be
-used to correlate individuals and warn [=holders=] when this information is
-shared.
         </p>
       </section>
 
--- a/index.html
+++ b/index.html
@@ -286,7 +286,7 @@ Information related to the type of [=credential=] this is (for example, a
 Dutch passport, an American driving license, or a health insurance card)
           </li>
           <li>
-Information related to specific attributes or properties being asserted by
+Information related to specific properties being asserted by
 the issuing authority about the [=subject=] (for example, nationality,
 the classes of vehicle entitled to drive, or date of birth)
           </li>
@@ -570,7 +570,7 @@ from them. A holder is often, but not always, a [=subject=] of the
         <dd>
 The means for keeping track of [=entities=] across contexts. Digital
 identities enable tracking and customization of [=entity=] interactions
-across digital contexts, typically using identifiers and attributes. Unintended
+across digital contexts, typically using identifiers and properties. Unintended
 distribution or use of identity information can compromise privacy. Collection
 and use of such information should follow the principle of
 [=data minimization=].
@@ -1081,7 +1081,7 @@ tooling to start issuing, holding, and verifying prototype credentials.
 It is expected that a developer will change `MyPrototypeCredential` below to
 the type of credential they would like to create. Since
 [=verifiable credentials=] talk about subjects, each property-value pair in
-the `credentialSubject` object expresses a particular attribute of the
+the `credentialSubject` object expresses a particular property of the
 credential subject. Once a developer has added a number of these property-value
 combinations, the modified object can be sent to [=verifiable credential=]
 issuer software and a [=verifiable credential=] will be created for the
@@ -1123,7 +1123,7 @@ mean the same thing to each other. This might be referred to as
         </p>
         <p>
 [=Verifiable credentials=] and [=verifiable presentations=] have many
-attributes and values that are identified by [=URLs=] [[URL]]. However,
+properties and values that are identified by [=URLs=] [[URL]]. However,
 those [=URLs=] can be long and not very human-friendly. In such cases,
 short-form human-friendly aliases can be more helpful. This specification uses
 the `@context` [=property=] to map such short-form aliases to the
@@ -1572,9 +1572,9 @@ respectively. See
         </p>
 
         <p class="note">
-The `@direction` attributes in the examples below are not required
+The `@direction` property in the examples below are not required
 for the associated single-language strings, as their default directions are the
-same as those set in the attribute. We include the `@direction` attributes here
+same as those set in the value. We include the `@direction` property here
 for clarity of demonstration, and to make copy+paste+edit deliver functional
 results. Implementers are encouraged to read the section on
 <a data-cite="JSON-LD11#string-internationalization">JSON-LD String Internationalization</a>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2368,7 +2351,74 @@ <h3>Status</h3>
...
+          </dd>
+          <dt>credentialSchema</dt>
+          <dd>
+Defined in Section [[[#data-schemas]]].
+          </dd>
+          <dt>credentialSubject</dt>
+          <dd>
+Defined in Section [[[#credential-subject]]].
+          </dd>
+          <dt>proof</dt>
```
brentzundel, line 2407: Seems odd to me to list `proof` as a property `defined in this specification` then point to data integrity.

index.html
```diff
@@ -2368,7 +2351,74 @@ <h3>Status</h3>
...
+          <dd>
+Defined in [[[VC-DATA-INTEGRITY]]].
+          </dd>
+        </dl>
+
+        <p>
+A [=verifiable credential=] can be extended to have addition properties
+through the extension mechanism defined in Section [[[#extensibility]]].
+        </p>
```
brentzundel, line 2416: maybe the proof property could be mentioned here instead, as an example of an additional property included via an extension mechanism?

index.html
```diff
@@ -4056,9 +4106,9 @@ <h3>Reserved Extension Points</h3>
               <td>`confidenceMethod`</td>
               <td>
 A property used for specifying one or more methods that a verifier might use to
-increase their confidence that the value of an attribute in or of a verifiable
+increase their confidence that the value of a property in or of a verifiable
 credential or verifiable presentation is accurate, including but not limited to
-attributes such as `initialRecipient` (a/k/a `issuee`), `presenter`,
+properties such as `initialRecipient` (a/k/a `issuee`), `presenter`,
 `authorizedPresenter`, `holder`, etc. The associated vocabulary URL MUST be
```
TallTed, line 4112: Here we have ``` `initialRecipient` (a/k/a `issuee`) ``` as well as ``` `authorizedPresenter` ```.

It seems like these should perhaps be added to the VCDMv2 reserved terms list.

I am also concerned that `confidenceMethod` is itself severely [underspecified](https://w3c-ccg.github.io/confidence-method-spec/#confidence-method), and does not appear to discuss any of the above properties.

TallTed (changes requested):

index.html
msporny: Removed from this list in ea0bfb8ec1253218ec1f60faadeba6c61216f206.

index.html
msporny: We link to the VC-DATA-INTEGRITY spec in the Securing Mechanisms section, so we're probably fine to just not include it in this part of the spec.

index.html
msporny: (Editor hat on) The description of confidence method is problematic. Given that almost no work has gone into it since it was introduced, I'm hesitant to reserve properties that it contemplates (that aren't documented at all).

(Digital Bazaar hat on) We have found use cases where confidence method is useful, so expect it to be more fully formed in time. However, for the time being, I'd rather not do anything about this issue in this PR.

We might want to raise another issue about `confidenceMethod`... or just deal with it when we have to make a judgement call on reserving it or not.

index.html
```diff
@@ -5281,42 +5396,44 @@ <h3>Identifier-Based Correlation</h3>
...
         <p>
 If strong anti-correlation properties are required, it is advised that
-signature values and metadata are regenerated each time using technologies like
-third-party pairwise signatures, zero-knowledge proofs, or group signatures.
+[=issuers=] produce [=verifiable credentials=] where the
+signature values and metadata can be regenerated for each
+[=verifiable presentation=] using technologies that support unlinkable
+disclosure, such as the [[[?VC-DI-BBS]]] specification. It is advised that, if
+possible, [=verifiers=] prefer [=verifiable presentations=] that use this
+technology in order to enhance the privacy for [=holders=].
```
David-Chadwick, line 5427:
```suggestion
technology in order to enhance the privacy for [=holders=] and [=subjects=].
```

index.html
```diff
@@ -5404,23 +5521,23 @@ <h3>Device Tracking and Fingerprinting</h3>
...
+[=issuers=] are advised to limit the information published in a [=verifiable
+credential=] to a minimal set needed for the expected purposes. One way to avoid
+placing PII in a [=verifiable credential=] is to use an abstract [=property=]
+that meets the needs of [=verifiers=] without providing overly-specific
+information about a [=subject=].
         </p>
         <p>
 For example, this document uses the `ageOver` [=property=]
-instead of a specific birthdate, which constitutes much stronger PII. If
+instead of a specific birthdate, which represents mroe sensitive PII. If
```
David-Chadwick, line 5534:
```suggestion
instead of a specific birthdate, which represents more sensitive PII. If
```

index.html
```diff
@@ -5404,23 +5521,23 @@ <h3>Device Tracking and Fingerprinting</h3>
...
 retailers in a specific market commonly require purchasers to be older than a
 certain age, an [=issuer=] trusted in that market might choose to offer a
```
David-Chadwick, line 5536:
```suggestion
certain age, an [=issuer=] trusted in that market might choose to offer
```

index.html
```diff
@@ -5404,23 +5521,23 @@ <h3>Device Tracking and Fingerprinting</h3>
...
 [=verifiable credential=] claiming that [=subjects=] have met that
```
David-Chadwick, line 5537:
```suggestion
[=verifiable credentials=] claiming that [=subjects=] have met that
```

index.html
```diff
@@ -5404,23 +5521,23 @@ <h3>Device Tracking and Fingerprinting</h3>
...
 requirement instead of offering [=verifiable credentials=] containing
-[=claims=] about specific birthdates. This enables individual customers to
-make purchases without revealing specific PII.
+[=claims=] about the customer's birthday. This enables individual customers to
```
David-Chadwick, line 5539:
```suggestion
[=claims=] about the customers' birthdays. This enables individual customers to
```

David-Chadwick (changes requested):

David-Chadwick (approved):

index.html
```diff
@@ -3318,110 +3367,136 @@ <h4>Semantic Interoperability</h4>
...
+A cryptographic digest, as defined in [[[VC-DATA-INTEGRITY]]].
+                  </td>
+                </tr>
+              </tbody>
+            </table>
+Each object associated with `relatedResource` MUST contain at least a
+`digestSRI` or `digestMultibase` value.
+          </dd>
+        </dl>
+
```
iherman, line 3436: This new section "smells" like defining a new type, in RDF sense, for a related resource. _At the moment_, in the vocabulary, the range of a `relatedResource` property is IRI, i.e., the only restriction on the property is that it is a an object property (as opposed to a datatype property). This bullet list seems to indicate that we want a new type (`RelatedResource`), in the domain a number of properties.

Do we want to formalize that in the vocabulary? Note that the last statement ("MUST contain at least a `digestSRI` or `digestMultibase` value") would require some OWL wizardry, which we probably should not include in our (simplified) vocabulary. Note also that this would require some changes in the DI vocabulary, because that is where `digestMultibase` is defined.

(If we agree these changes should happen, I am happy to come up with new PR-s at some point.)

index.html
iherman: I agree with @msporny that this should be dealt with in a separate issue.

index.html
```diff
@@ -5429,60 +5546,66 @@ <h3>The Principle of Data Minimization</h3>
...
-another. Accepted best practice for preventing such violations is to limit the
-information requested, and received, to the absolute minimum necessary. This
-data minimization approach is required by regulation in multiple jurisdictions,
-including the Health Insurance Portability and Accountability Act (HIPAA) in the
-United States and the General Data Protection Regulation (GDPR) in the European
-Union.
+another. One accepted best practice for preventing such a violation is for
+[=verifiers=] to limit the information requested, and received, to the absolute
+minimum necessary for a particular transaction. This data minimization approach
+is required by regulation in multiple jurisdictions, including the Health
```
iherman, line 5552:
```suggestion
is required by regulations in multiple jurisdictions, including the Health
```

iherman (approved):

index.html
```diff
@@ -1618,18 +1597,25 @@ <h3>Types</h3>
...
-The type system used in the data model described in this specification allows
-for multiple ways to associate types with data. Implementers and authors are
-urged to read the section on typing in the Verifiable Credentials
-Implementation Guidelines [[?VC-IMP-GUIDE]].
+        <p class="note" title="Creating new verifiable credential types">
+Explaining how to create new types of [=verifiable credentials=] is beyond
+the scope of this specification. Readers that are interested in doing so are
+advised to read the Section on
+<a data-cite="?VC-IMP-GUIDE#creating-new-credential-types">
+Creating New Credential Types</a> in the [[[?VC-IMP-GUIDE]]].
```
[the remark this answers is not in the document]
msporny: Yes, agreed. If we re-publish it as an updated NOTE, we should make a full Editorial pass... there is guidance in there now that is VERY OLD and should be updated.

index.html
msporny: If we formalize that, we should probably formalize both `relatedResource` and its type in the Data Integrity vocabulary. It makes more sense for it to go over there than for it to be defined by the VCDM. If we make this decision, we'll need to make it quickly and affect the change quickly so that we can lock the tests down for CR.

iherman: I am happy to do either way... in any case, both vocabularies will have to be updated.

I would propose to (1) have this PR merged and (2) have DI PR-s, especially https://github.com/w3c/vc-data-integrity/pull/258 handled and possibly merged before doing this. I do not want stupid merge conflicts...

decentralgabe (approved):
thanks for the work!

index.html
```diff
@@ -787,14 +787,15 @@ <h3>Claims</h3>
...
-metadata that cryptographically prove who issued it.
+A [=credential=] is a set of one or more [=claims=] made by the same [=entity=].
+[=Credentials=] might also include an identifier and metadata to describe
+properties of the [=credential=], such as the [=issuer=], the validity date and
+time period, a representative image, [=verification material=], status
+information, and so on. A
+[=verifiable credential=] is a set of tamper-evident [=claims=] and metadata
+that cryptographically prove who issued it. Examples of [=verifiable
+credentials=] include digital employee identification cards, digital birth
+certificates, and digital educational certificates.
```
dlongley, line 798:
```suggestion
credentials=] include, but are not limited to, digital employee identification
cards, digital birth certificates, and digital educational certificates.
```

index.html
```diff
@@ -894,6 +885,13 @@ <h3>Credentials</h3>
           </figcaption>
         </figure>
 
+        <p class="note">
+[=Credential=] identifiers are often used to identify specific instances
```
dlongley, line 889:
```suggestion
[=Credential=] identifiers can be used to identify specific instances
```

index.html
```diff
@@ -894,6 +885,13 @@ <h3>Credentials</h3>
...
+of a [=credential=]. These identifiers can also be used for correlation. A
+[=holder=] wanting to minimize correlation is advised to use a selective
+disclosure scheme that does not reveal the [=credential=] identifier.
```
dlongley, line 892:
```suggestion
of a [=credential=]. These identifiers could also be used for unwanted
correlation. A [=holder=] wanting to minimize correlation is advised
to use a selective disclosure scheme that does not reveal the
[=credential=] identifier.
```

index.html
```diff
@@ -3318,110 +3367,136 @@ <h4>Semantic Interoperability</h4>
...
         </p>
+
         <p>
 Specification authors that write algorithms that fetch a resource based on the
 `id` of an object inside a [=conforming document=] need to consider whether
 that resource's content is vital to the validity of that document. If it is, the
-specification MUST produce a validation error unless the resource has the
-expected media type and its bytes hash to the expected digest.
+specification MUST produce a validation error unless the resource matches the
+expected media type and cryptographic digest
```
dlongley, line 3473:
```suggestion
expected media type and cryptographic digest.
```

index.html
```diff
@@ -5234,38 +5309,78 @@ <h3>Personally Identifiable Information</h3>
...
+from the [=issuer=] through the use of techniques like
+<a href="https://en.wikipedia.org/wiki/Blind_signature">blind signatures</a>,
+while still keeping the identifier embedded and signed in the [=verifiable
+credential=].
+        </p>
+
+        <p>
+[=Securing mechanism=] specification authors are advised to avoid enabling
+identifier-based correlation by designing their technologies, when possible,
+to avoid the use of correlating identifiers.
```
dlongley, line 5371:
```suggestion
to avoid the use of correlating identifiers that cannot be selectively disclosed.
```

dlongley (approved):

index.html
```diff
@@ -1352,37 +1356,20 @@ <h3>Identifiers</h3>
...
-        </p>
-
-        <ul>
-          <li>
-The `id` [=property=] MUST express an identifier that others are
-expected to use when expressing statements about a specific thing identified
-by that identifier.
-          </li>
-          <li>
-The `id` [=property=] MUST NOT have more than one value.
```
selfissued, line 1371: Why is the statement that the `id` must have only one value being deleted?

index.html
```diff
@@ -1392,10 +1379,10 @@ <h3>Identifiers</h3>
         <dl>
           <dt><dfn class="export">id</dfn></dt>
           <dd>
-The value of the `id` [=property=] MUST be a single [=URL=].
-It is RECOMMENDED that the [=URL=] in the `id` be one which, if
-dereferenced, results in a document containing machine-readable information
-about the `id`.
+The `id` [=property=] is OPTIONAL. If present, the value of the `id`
+[=property=] MUST be a single [=URL=] which MAY be dereferenced. It is
```
selfissued, line 1383: This contradicts the above statement that the `id` value can be a GUID.

index.html
```diff
@@ -1450,30 +1436,23 @@ <h3>Types</h3>
 use type information to determine whether or not a provided
 [=verifiable credential=] or [=verifiable presentation=] is appropriate
 for the intended use case. This specification defines a `type`
-[=property=] for the expression of type information. This type information
-can be used during [=validation=] processes as described in Appendix
-<a href="#validation"></a>.
+[=property=] for the expression of object type information. This type
+information can be used during [=validation=] processes as described in Appendix
```
selfissued, line 1440:
```suggestion
information can be used during [=validation=] processes, as described in Appendix
```

index.html
```diff
@@ -1618,18 +1597,25 @@ <h3>Types</h3>
...
 
-        <p class="note">
-The type system used in the data model described in this specification allows
-for multiple ways to associate types with data. Implementers and authors are
-urged to read the section on typing in the Verifiable Credentials
-Implementation Guidelines [[?VC-IMP-GUIDE]].
+        <p class="note" title="Creating new verifiable credential types">
+Explaining how to create new types of [=verifiable credentials=] is beyond
+the scope of this specification. Readers that are interested in doing so are
+advised to read the Section on
```
selfissued, line 1616:
```suggestion
advised to read the section on
```

index.html
selfissued: I agree with not mentioning `proof` here.

index.html
selfissued: It seems that this addition is not editorial.  As one example, it pulls the use of digestMultibase into the VCDM specification, which I object to, per previous discussions.  Please remove these non-editorial additions.

index.html
```diff
@@ -3444,9 +3519,8 @@ <h2>Integrity of Related Resources</h2>
   "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
   "image": {
     "id": "https://university.example.org/images/58473",
-    "digestSRI":
-      "sha384-ZfAwuJmMgoX3s86L7x9XSPi3AEbiz6S/5SyGHJPCxWHs5NEth/c5S9QoS1zZft+J",
     "mediaType": "application/svg+xml",
+    "digestMultibase": "zQmdfTbBqBPQ7VNxZEYEj14VmRuZBkqFbiwReogJgS1zR1n"
```
selfissued, line 3523: Please remove the addition of digestMultibase.  It does not belong in this specification.

index.html
```diff
@@ -5234,38 +5309,78 @@ <h3>Personally Identifiable Information</h3>
...
+        <p>
+In general, individuals are advised to assume that a [=verifiable credential=],
+like most physical credentials, will leak personally identifiable information
+when shared. To combat this leakage, the [=verifiable credential=], and the
+securing mechanism, need to be specifically designed to avoid correlation.
+[=Verifiable credentials=] that are specifically designed to prevent the leakage
+of personally identifiable information do exist. Individuals and implementers
+are urged to prefer these types of credentials over ones that are not designed
+to protect personally identifiable information.
+        </p>
```
selfissued, line 5322: This kind of advice is certainly not normative and arguably belongs in a deployment guide, rather than the specification.  At most, it's non-normative deployment considerations.  Please delete this addition.

index.html
```diff
@@ -5281,42 +5396,44 @@ <h3>Identifier-Based Correlation</h3>
...
         <p>
 If strong anti-correlation properties are required, it is advised that
-signature values and metadata are regenerated each time using technologies like
-third-party pairwise signatures, zero-knowledge proofs, or group signatures.
+[=issuers=] produce [=verifiable credentials=] where the
+signature values and metadata can be regenerated for each
+[=verifiable presentation=] using technologies that support unlinkable
+disclosure, such as the [[[?VC-DI-BBS]]] specification. It is advised that, if
+possible, [=verifiers=] prefer [=verifiable presentations=] that use this
+technology in order to enhance the privacy for [=holders=] and [=subjects=].
```
selfissued, line 5427: This paragraph seems like it's crafted to be an advertisement for a DI spec.  Please revert this proposed change.  The existing text already did the job.

index.html
```diff
@@ -5666,6 +5788,13 @@ <h3>Storage Providers and Data Mining</h3>
 expectations.
           </li>
         </ul>
+
+        <p>
+In addition to the mitigations above, the participation of civil society and
+regulators in the analysis and auditing of vendors can also ensure that there
+are legal protections in place, and enforced, for individuals affected by
+practices that are not aligned with their best interests.
+        </p>
```
selfissued, line 5797: This is far from actionable specification text for implementers.  It's closer to sociology.

Please delete this unnecessary addition.

selfissued (changes requested):
Despite the title of the PR, aspects of this PR are not editorial, as pointed out in several of my individual comments.

For instance, the PR attempts to inject the use of `digestMultibase` into the specification.

It says that an `id` value can be a GUID and then it says that it must be a dereferenceable URL.

It contains unnecessary editorializing about what people should and shouldn't do with VCs in several places.  That's not actionable by implementers of the specification.

It also appears to remove some normative statements that were previously in the specification.

msporny:
> Despite the title of the PR, aspects of this PR are not editorial, as pointed out in several of my individual comments.

As I stated on the call last week and this week, I did touch some normative text, but it was around content that was already in the specification which was either a duplication of existing content or normative statements that were presumed in an at risk marker that weren't in the specification as normative text.

> For instance, the PR attempts to inject the use of `digestMultibase` into the specification.

No, the specification already had an at risk marker that has been in there for a long time stating that the group was trying to decide on what to use, or both. The original relatedResource text was not going to be accepted if both were not still under consideration. So, nothing changed from a normative perspective, but I did notice that the text didn't align w/ the at risk marker and fixed it. The `digestMultibase` and `digestSRI` values have existed in their current form for four months now and were merged, with group consensus, a while ago: 

https://github.com/w3c/vc-data-model/blame/main/contexts/credentials/v2#L69-L76

> It says that an `id` value can be a GUID and then it says that it must be a dereferenceable URL.



> It contains unnecessary editorializing about what people should and shouldn't do with VCs in several places. That's not actionable by implementers of the specification.

This was text that was requested by @jyasskin from Google. I see no issue with the text that they requested; not every statement in a specification needs to be actionable by an implementer.

> It also appears to remove some normative statements that were previously in the specification.

The normative statements removed were duplicative, AFAICT, I'll make another pass to make sure that's true, but that certainly was the intent.

index.html
msporny: It wasn't deleted, it was collapsed into a more succinct description:

https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1464.html#dfn-id

selfissued:
> No, the specification already had an at risk marker that has been in there for a long time stating that the group was trying to decide on what to use, or both. The original relatedResource text was not going to be accepted if both were not still under consideration. So, nothing changed from a normative perspective,

The only use of digestMultibase in the core spec is in an issue marker which says:
> Candidates for this mechanism include digestSRI and digestMultibase.

This PR elevates digestMultibase by including normative use of it in the specification.  That's definitely a normative change.

index.html
msporny: How does it contradict: "urn:uuid:50e8098a-029a-11ef-a8e7-071a217fc2a9"?

index.html
msporny: It was requested that this section was cleaned up.

In doing so, I consolidated all the normative statements to flow like the normative statements in the other sections. The `digestMultibase` value was an at-risk feature, along with `digestSRI` as specified here (see the at risk text starting with "Unification of cryptographic hash expression formats are under discussion" which was added when this section was added to the specification):

https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1464.html#integrity-of-related-resources

Please concretely identify which normative assertions didn't exist in the specification before the change.

index.html
msporny: This change was requested by @jyasskin. Its in a privacy considerations section, which tends to NOT include normative statements but general non-normative deployment or usage considerations.

Please discuss your request to remove this guidance with @jyasskin.

index.html
msporny: More specific guidance on unlinkability and which roles in the ecosystem are responsible for unlinkability was requested by @jyasskin. If VC-JOSE-COSE supported unlinkable disclosure, we could refer to it as well. Speaking to unlinkability, but then not providing an example of it, would not achieve what the issue commenter was requesting to be addressed.

index.html
msporny: This text was explicitly requested by @jyasskin in his review. It's in a privacy considerations section, and it's important to point out that people like @npdoty need to have a say in oversight wrt. digital credentials ecosystems.

msporny:
> This PR elevates digestMultibase by including normative use of it in the specification. That's definitely a normative change.

To quote the entire issue marker:

> The Working Group is currently attempting to determine whether cryptographic hash expression formats can be unified across all of the VCWG core specifications. Candidates for this mechanism include digestSRI and digestMultibase. There are arguments for and against unification that the WG is currently debating.

The document is in Candidate Recommendation. The context includes both `digestSRI` and `digestMultibase`. The text above states that the candidates include both `digestSRI` and `digestMultibase`.

No elevation occurred. It is not a normative change from what existed when we entered CR.

index.html
```diff
@@ -3318,110 +3368,136 @@ <h4>Semantic Interoperability</h4>
...
+                <tr>
+                  <td>`digestSRI`</td>
+                  <td>
+A cryptographic digest, as defined in [[[SRI]]].
+                  </td>
+                </tr>
+                <tr>
+                  <td>`digestMultibase`</td>
+                  <td>
+A cryptographic digest, as defined in [[[VC-DATA-INTEGRITY]]].
```
decentralgabe, line 3428: on second read, I consider that this may view as favoring one securing format ... 
would this work to make the language more generic something like `digestSRI` OR a suitable digest format defined by a securing mechanism?

selfissued:
> The context includes both digestSRI and digestMultibase.

My conclusion from that is that these should be removed from the context, since terms in the context should be normatively defined by the specification, which they aren't in this case.  Unless we achieve a meeting of the minds on this during a working group call, I'll plan to file an issue for their removal, since their inclusion is out of sync with the spec.

index.html
msporny: Saying that would definitely favor `digestSRI` over other formats. So, if the goal here is to not favor any particular digest format, we can define `digestSRI` and `digestMultibase` elsewhere. The language that wouldn't prefer any particular digest format would be:

"A cryptographic digest defined by a securing mechanism, such as `digestSRI` defined in [[[VC-JOSE-COSE]]], or `digestMultibase` as defined in [[[VC-DATA-INTEGRITY]]]". 

If we do that, we'd still want to include both in the base context to make sure that developers don't have to piece together JSON-LD contexts just to have basic content integrity support (which is what we already do).

msporny:
@selfissued wrote:
> My conclusion from that is that these should be removed from the context, since terms in the context should be normatively defined by the specification, which they aren't in this case.

The group discussed this a while ago and came to the conclusion to include things, like a default vocabulary, and terms defined by other specifications, that are useful to developers. 

The original discussion happened here: https://github.com/w3c/vc-data-model/issues/1089
and the topic was re-visited here: https://github.com/w3c/vc-data-model/issues/1314#issuecomment-1765244354

> Unless we achieve a meeting of the minds on this during a working group call, I'll plan to file an issue for their removal, since their inclusion is out of sync with the spec.

That would go against the consensus position we've had in the group since August 2023, which was re-affirmed in October 2023.

index.html
msporny: @decentralgabe ok, please check to see if the group is going to be ok w/ doing this on the call this week (I won't be able to attend), and if so, I can make the change.

index.html
selfissued: Its not dereferenceable

msporny: The text says "If present, the value of the `id` [=property=] MUST be a single [=URL=] which MAY be dereferenced."

We could say:

```suggestion
[=property=] MUST be a single [=URL=], which MAY be dereferenced. It is
```

OR

```suggestion
[=property=] MUST be a single [=URL=], which MAY be dereferenceable. It is
```

Which one would you prefer? Or would you prefer alternate text. If you would prefer alternate text, please suggest something, that will make the chance process go more rapidly.

dlongley: I think this is just a typo that should read `MAY be dereferenceable` (just as the following sentence says).

```suggestion
[=property=] MUST be a single [=URL=] which MAY be dereferenceable. It is
```

index.html
```diff
@@ -4929,22 +5005,22 @@ <h3>Personally Identifiable Information</h3>
...
-with these kinds of characteristics. [=Issuers=] are strongly advised to
-provide privacy-protecting [=verifiable credentials=] when possible. For
-example, issuing `ageOver` [=verifiable credentials=] instead of
-date of birth [=verifiable credentials=] when a [=verifier=] wants to
-determine whether an [=entity=] is over the age of 18.
+Implementers of software used by [=holders=] are strongly advised to warn
+[=holders=] when they share data with these kinds of characteristics.
+[=Issuers=] are strongly advised to provide privacy-protecting [=verifiable
+credentials=] when possible. For example, issuing `ageOver` [=verifiable
+credentials=] instead of date of birth [=verifiable credentials=] when a
```
TallTed, line 5022:
```suggestion
credentials=] instead of `dateOfBirth` [=verifiable credentials=] for use when a
```

index.html
```diff
@@ -4956,38 +5032,78 @@ <h3>Personally Identifiable Information</h3>
...
-[=verifiable credentials=] system, it is strongly advised that identifiers
-are either:
+[=Securing mechanism=] specification authors are advised to avoid enabling
+identifier-based correlation by designing their technologies, when possible,
+to avoid the use of correlating identifiers that cannot be selectively disclosed.
+        </p>
+
+        <p>
+If strong anti-correlation properties are a requirement in a [=verifiable
+credentials=] system, it is strongly advised that identifiers are either:
```
TallTed, line 5099:
```suggestion
credentials=] system, it is strongly advised that identifiers are one or more
of the following:
```

index.html
```diff
@@ -5003,42 +5119,44 @@ <h3>Identifier-Based Correlation</h3>
...
-
-        <p>
-Organizations providing software to [=holders=] should strive to identify
-fields in [=verifiable credentials=] containing information that could be
-used to correlate individuals and warn [=holders=] when this information is
-shared.
+properties of the cryptography used. See Sections
+[[[#personally-identifiable-information]]], [[[#identifier-based-correlation]]],
+[[[#metadata-based-correlation]]], [[[#correlation-during-validation]]],
+and most of the other sections in Section [[[#privacy-considerations]]].
```
TallTed, line 5159:
```suggestion
and most of the other subsections of Section [[[#privacy-considerations]]].
```

index.html
```diff
@@ -5151,60 +5269,66 @@ <h3>The Principle of Data Minimization</h3>
...
+Implementers that find this impractical or unsafe, might consider using
+[=selective disclosure=] schemes that eliminate dependence on [=issuers=] at
+proving time and reduce temporal correlation risk from [=issuers=].
         </p>
 
         <p>
 [=Verifiers=] are urged to only request information that is absolutely
 necessary for a specific transaction to occur. This is important for at least
-two reasons. It:
+two reasons.
```
TallTed, line 5311:
```suggestion
two reasons:
```

index.html
```diff
@@ -5151,60 +5269,66 @@ <h3>The Principle of Data Minimization</h3>
...
         </p>
 
         <ul>
           <li>
-Reduces the liability on the [=verifier=] for handling highly sensitive
-information that it does not need to.
+It reduces the liability on the [=verifier=] for handling highly sensitive
+information that it does not need to, and
```
TallTed, line 5317:
```suggestion
information that it does not need to handle.
```

index.html
```diff
@@ -5151,60 +5269,66 @@ <h3>The Principle of Data Minimization</h3>
...
           </li>
           <li>
-Enhances the privacy of the individual by only asking for information required
-for a specific transaction.
+it enhances the privacy of the individual by only asking for information
+required for a specific transaction.
```
TallTed, line 5321:
```suggestion
It enhances the privacy of the individual by only asking for information
that is necessary for a specific transaction.
```

index.html
```diff
@@ -5151,60 +5269,66 @@ <h3>The Principle of Data Minimization</h3>
...
           </li>
         </ul>
 
+        <p>
+Implementers of software used by [=holders=] are urged to disclose what
+information is being requested by a [=verifier=] such that a [=holder=] can
```
TallTed, line 5327:
```suggestion
information is being requested by a [=verifier=], such that a [=holder=] can
```

index.html
```diff
@@ -5589,9 +5722,9 @@ <h3>Data Theft</h3>
 manage the status and revocation of those credentials. Similarly,
 [=issuers=] are advised to avoid the practice of creating publicly
 resolvable credentials that include personally identifiable information
-(PII) or other sensitive data. Software implementers are advised 
-to safeguarded [=verifiable credentials=] using robust consent 
-and access control measures, ensuring that they remain 
+(PII) or other sensitive data. Software implementers are advised
+to safeguarded [=verifiable credentials=] using robust consent
```
TallTed, line 5726:
```suggestion
to safeguard [=verifiable credentials=] using robust consent
```

index.html
```diff
@@ -5820,72 +5953,28 @@ <h3>Content Integrity Protection</h3>
...
-  }
-}
-        </pre>
-
-        <p class="note">
-It is debatable whether the JSON-LD Contexts above need protection because
-production implementations are expected to ship with static copies of important
-JSON-LD Contexts.
+This specification provides an optional mechanism, contained in Section
+[[[#integrity-of-related-resources]]], that is capable of ensuring the content
```
TallTed, line 5965:
```suggestion
[[[#integrity-of-related-resources]]], that is capable of ensuring content
```

index.html
```diff
@@ -5820,72 +5953,28 @@ <h3>Content Integrity Protection</h3>
...
+that could result in a security issue if the external content changes.
         </p>
 
         <p>
-While the example above is one way to achieve content integrity protection,
-there are other solutions that might be better suited for certain applications.
 Implementers are urged to understand how links to external machine-readable
 content that are not content-integrity protected could result in successful
-attacks against their applications.
+attacks against their applications and utilize the content integrity protection
```
TallTed, line 5975:
```suggestion
attacks against their applications, and utilize the content integrity protection
```

index.html
```diff
@@ -6356,22 +6445,22 @@ <h3>Credential Type</h3>
...
 
         <p>
 When a [=verifier=] requests a [=verifiable credential=] of a specific type,
 there will be a set of mandatory and optional [=claims=] that are associated
-with that type. It is expected that a [=verifier=]'s validation of a
-[=verifiable credential=] will fail when mandatory [=claims=] are not
-included, and that any [=claim=] that is not associated with the specific
-type will be ignored. In other words, a
+with that type. A [=verifier=]'s validation of a [=verifiable credential=] will
+fail when mandatory [=claims=] are not included, and that any [=claim=] that is
```
TallTed, line 6462:
```suggestion
fail when mandatory [=claims=] are not included, and any [=claim=] that is
```

TallTed (changes requested):
Small things...

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-05-01-vcwg#section2-2) on 2024-05-01

- no resolutions were taken

#### 2.2. Address Jeffrey Yasskin's review comments that are Editorial (pr vc-data-model#1464)

_See github pull request [vc-data-model#1464](https://github.com/w3c/vc-data-model/pull/1464)._




**Brent Zundel:** This is Manu's healthy response to a very thorough review from Google's AC rep, Jeffrey Yasskin.  
… The impetus for the review is that we wanted Google to say all the things they wanted fixed to avoid any formal objections from them. That was the gist of the request -- and he provided an acceptable thorough review, the normative changes have already been made, and this is the editorial requests.  
… This is someone who is very versed in Web specs in general and the Web in particular, but not so versed in VCs and he requested clarifications where he thought clarifications would result in a better spec.  
… Some other changes are in a PR from David Chadwick.  
… This PR is quite long and has some requests for changes and some approvals.  
… Timeboxing this one to 7 minutes.  
… Please jump on the queue to talk about 1464.  

**Ted Thibodeau Jr.:** I believe all of my stuff is editorial.  

**Brent Zundel:** Mike you were perhaps saying "don't do this at all"?  

**Michael Jones:** I am ok with the vast majority of this. I'm requesting that the text elevates multihash into normative usage be struck because that's not appropriate in an editorial PR.  
… Manu's tried to say it's already in the context so it doesn't matter. The specification defines what people normatively do, not the context.  

**Brent Zundel:** The one issue I'm running into -- with the additional commits and changes ... trying to find the comments. Down in line 3155, if folks want to read along. That's a conversation we want to have.  

**Gabe Cohen:** I believe it would be more neutral to say that you can use a digest property defined by the securing mechanisms. `digestSRI` isn't defined by either of our securing mechanisms.  
… Manu's suggestion was to move it into VC-JOSE-COSE, but I think on further reflection we shouldn't move it, I think it's in another spec already and neutral. I think a possible way forward is to say you can use either `digestSRI` or a method from another securing spec without saying which spec.  

**Dave Longley:** my memory of consensus was that implementers could use either, and we'd see how that goes.  
… in terms of how digestSRI being in a different spec, that's not quite right...there's lots of work in our spec to shape it as a property and value for VCs.  
… this PR seems to aim at the previous consensus.  

**Brent Zundel:** the digestMultibase in the Data Model does not point to an unspecified thing, but to something we have published.  
… Before going back to the queue -- the `digestMultibase` as referenced, does not point to an unpublished spec somewhere at IETF, it points to the DI spec that our group has published as CR that defines the particular things it is concerned with. In my understanding, there is no normative pointing to something that isn't a spec, except that we're pointing at the DI spec which is underway in becoming a spec.  
… so there is no normative pointing to something that cannot be normatively pointed to.  

**Dave Longley:** +1 to brent.  

**Ted Thibodeau Jr.:** Mike, you are objecting to a thing called multihash which isn't in this document.  

**Michael Jones:** `digestMultibase` sorry.  

**Ted Thibodeau Jr.:** That appears to not be as significant a change as you're making it be.  
… My eyes are different than yours.  

**Michael Jones:** Gabe, thank you, you are correct, I have no objection to `digestSRI` because that's in a W3C REC that defines or largely defines what we would need. Largely for the record, my objections to the whole multihash family is that they fail to make a choice.  
… That's my primary objection to using that at all.  

> *Dmitri Zagidulin:* I think 'multihash family' may be somewhat confusing. I think Mike's main objection is to multiBASE.

**Michael Jones:** With respect to using that at all, with respect to DI standardizing `digestMultibase`, that's fine, but we went through an exercise six months ago or so of making sure that the data model didn't have any normative dependencies on either security spec -- it has informative references.  
… Part of the reasons for doing the controller document work -- an alternative was expressed that maybe we can just reference the controller document section from the data model and call it good. Orie and I and others felt like that was turning things on their heads and the data model should stand on its own.  
… I don't want to change it to have dependencies on either securing spec.  

**Dave Longley:** just a clarification, there are pieces of the SRI spec did have to be added to our vocabulary earlier.  
… I did not mean to imply that it was yet to be done, but to point out that we could not simply point to someone else's specification.  

> *Gabe Cohen:* +1 Brent.

**Brent Zundel:** Since the changes to section 5.3 of related resources is the piece under contention, and I will reach out to Manu here too, the suggestion is that the changes to this section be removed from this PR and made a separate PR that can be further discussed so it doesn't hold up all of the other myriad changes that do not have opposition.  
… Mike, would that alleviate your concerns with the PR?  

**Michael Jones:** Absolutely, once that text is gone, I'll approve.  

**Brent Zundel:** I'll let Manu know we can put those changes in a separate PR.  
… Any other comments on this PR before we move on?  
… I think we covered the main big things we wanted to talk about and maybe we can move to issue processing now.  

</details>

msporny:
> Brent Zundel: Since the changes to section 5.3 of related resources is the piece under contention, and I will reach out to Manu here too, the suggestion is that the changes to this section be removed from this PR and made a separate PR that can be further discussed so it doesn't hold up all of the other myriad changes that do not have opposition.
… Mike, would that alleviate your concerns with the PR?
> Michael Jones: Absolutely, once that text is gone, I'll approve.

I have reverted all changes to Section 5.3: Integrity of Related Resources (in commit 150bfbb2f15c5b7e6f2054ab914a0f43d2e938ea) and raised those changes in PR #1484 for further discussion.

msporny:
Editorial, multiple reviews, most changes requested and made, objections by @selfissued reverted and moved to PR #1484, no remaining objections, merging.

Issue #1467, opened 2024-03-27
Add issuee definition


David-Chadwick:
There have been a number of discussions and pieces of text when the use of the term holder has not been precise enough, since it is the first holder that is specifically being addressed in these cases, rather than any holder.
The proposal is to add the definition of issuee to the data model, it being the first holder that the issuer issues the VC to.
The proposal is also to review the various CRs from the group and to use the newly defined term where it is appropriate to do so (rather than the existing use of holder, or first holder or other phrase that implies the issuee).
This will be a non-normative change to the DM, since there will not be any protocol messages that refer to the issuee.

David-Chadwick:
Here is the straw man proposal for the definition of issuee, so that we can discuss it here and then move the agreed upon definition to the PR #1468 

```
A role an [=entity=] performs when it is the intended recipient of 
a [=verifiable credential=] from
the [=issuer=] and stores the [=verifiable credential=] in its [=repository=]. 
The issuee role is a subclass of the [=holder=] role.
```

And here is the straw man proposal for changing the definition of issuer.

```
A role an [=entity=] performs by asserting [=claims=] about zero, one or
more [=subjects=], creating a [=verifiable credential=] from these
[=claims=], and transmitting the [=verifiable credential=], directly 
or indirectly, to the [=issuee=].
```

TallTed:
All existing mentions of `issuee` within HTML spec docs can be found with this search (which regrettably gets extra things like `issueEvent`, which I can find no way to exclude). I probably should have sorted the results by least-recently-updated; instead, each of these bullet sub-lists are sorted by GitHub's idea of "best match". :-/ Hopefully, still helpful.

* https://github.com/search?q=issuee++user%3Aw3c+language%3AHTML&type=pullrequests&ref=advsearch

There is only *one* instance in the main branch of *one* document, the VCDM itself —
* https://github.com/w3c/vc-data-model/blob/1ab385f1e4daf04a6d6bca44f019d7fe7bc3bda0/index.html#L4068

There are other instances in issues (all on the VCDM) —
* open
  * https://github.com/w3c/vc-data-model/issues/1467
  * https://github.com/w3c/vc-data-model/issues/1359
* closed
  * https://github.com/w3c/vc-data-model/issues/942
  * https://github.com/w3c/vc-data-model/issues/959
  * https://github.com/w3c/vc-data-model/issues/1173
  * https://github.com/w3c/vc-data-model/issues/960
[...]

PR #1468, opened 2024-03-27
Add definition of Issuee

David-Chadwick:
Add issuee and changes due to adding it

***
<a href="https://pr-preview.s3.amazonaws.com/David-Chadwick/vc-data-model/pull/1468.html" title="Last updated on Apr 2, 2024, 8:23 PM UTC (0c20b3c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1468/25c3efa...David-Chadwick:0c20b3c.html" title="Last updated on Apr 2, 2024, 8:23 PM UTC (0c20b3c)">Diff</a>

### DIFF
--- a/diagrams/ecosystem.svg
+++ b/diagrams/ecosystem.svg
@@ -58,7 +58,7 @@
                             <font style="font-size:16px">
                                 <i>
                                     <b>
-                                        Holder
+                                        Issuee/Holder
                                     </b>
                                 </i>
                             </font>
--- a/diagrams/ecosystem.svg
+++ b/diagrams/ecosystem.svg
@@ -58,7 +58,7 @@
                             <font style="font-size:16px">
                                 <i>
                                     <b>
-                                        Issuee/Holder
+                                        Holder
                                     </b>
                                 </i>
                             </font>
--- a/index.html
+++ b/index.html
@@ -370,9 +370,15 @@ from them. Example holders include students, employees, and customers.
           <dd>
 A role an [=entity=] performs by asserting [=claims=] about one or
 more [=subjects=], creating a [=verifiable credential=] from these
-[=claims=], and transmitting the [=verifiable credential=] to a
-[=holder=]. Example issuers include corporations, non-profit organizations,
+[=claims=], and transmitting the [=verifiable credential=] to the
+[=issuee=]. Example issuers include corporations, non-profit organizations,
 trade associations, governments, and individuals.
+          </dd>
+          <dt>[=issuee=]</dt>
+          <dd>
+A role an [=entity=] might perform by receiving one or more
+[=verifiable credentials=] from an [=issuer=]. An issuee is a subclass
+of the holder role.
           </dd>
           <dt>[=subject=]</dt>
           <dd>
@@ -419,6 +425,8 @@ ecosystem.
             The roles and information flows forming the basis for this specification.
           </figcaption>
         </figure>
+
+[PR NOTE. Figure 1 will need updating to replace Holder with Issuee/Holder]
 
         <p class="note">
 <a href="#roles"></a> above provides an example ecosystem in which to ground the
@@ -585,12 +593,17 @@ this document to other specifications. This specification decouples the
 [=identity provider=] concept into two distinct concepts: the [=issuer=]
 and the [=holder=].
         </dd>
+        <dt><dfn class="export" data-lt="issuee|issuee's">issuee</dfn></dt>
+        <dd>
+A role an [=entity=] performs when it receives a [=verifiable credential=] from
+the [=issuer=].
+        </dd>
         <dt><dfn class="export" data-lt="issuers|issuer's">issuer</dfn></dt>
         <dd>
 A role an [=entity=] can perform by asserting [=claims=] about one or
 more [=subjects=], creating a [=verifiable credential=] from these
-[=claims=], and transmitting the [=verifiable credential=] to a
-[=holder=].
+[=claims=], and transmitting the [=verifiable credential=] to the
+[=issuee=].
         </dd>
         <dt><dfn data-lt="named graphs">named graph</dfn></dt>
         <dd>


index.html
```diff
@@ -370,9 +370,15 @@ <h3>Ecosystem Overview</h3>
           <dd>
 A role an [=entity=] performs by asserting [=claims=] about one or
 more [=subjects=], creating a [=verifiable credential=] from these
-[=claims=], and transmitting the [=verifiable credential=] to a
-[=holder=]. Example issuers include corporations, non-profit organizations,
+[=claims=], and transmitting the [=verifiable credential=] to the
+[=issuee=]. Example issuers include corporations, non-profit organizations,
```
TallTed, line 374: Is there only one issuee, by definition? 

Or could the issuer simultaneously transmit the VC to multiple issuees?
```suggestion
[=claims=], and transmitting the [=verifiable credential=] to an
[=issuee=]. Example issuers include corporations, non-profit organizations,
```

dlongley, line 374: Hmm, it could be that an issuer hands the credential to some party *other than the issuee* as the first holder. And that party will / is expected to perform delivery to the `issuee`. At least, I would expect people to think that the `issuee` is different from this intermediary. I worry that this change reduces something more general to a specific case where the issuer interacts directly with the "issuee" whereas the previous text allowed for there to be any number of intermediaries.

If an issuer issues into a repository managed by a third party that holds the credentials for later pickup by eventual recipients -- is that third party the issuee ... or are those recipients the issuees?

Issuers may also issue directly onto public lists, where I don't think the "public list" would be considered an "issuee", but it would be a "holder". There may be no "issuee" in this case ... right?

I think we would need different language here around the "issuee" case being a more specific case (even if it happens quite often) of the more general model. So the issuer always issues to a holder, which may / may not be expressed or recognized as an issuee?

dlongley (changes requested):

index.html
```diff
@@ -370,9 +370,15 @@ <h3>Ecosystem Overview</h3>
...
 trade associations, governments, and individuals.
+          </dd>
+          <dt>[=issuee=]</dt>
+          <dd>
+A role an [=entity=] might perform by receiving one or more
+[=verifiable credentials=] from an [=issuer=]. An issuee is a subclass
+of the holder role.
```
TallTed, line 381:
```suggestion
[=verifiable credentials=] directly from an [=issuer=]. The issuee
role is a subclass of the [=holder=] role.
```

index.html
```diff
@@ -584,13 +592,18 @@ <h2>Terminology</h2>
 this document to other specifications. This specification decouples the
 [=identity provider=] concept into two distinct concepts: the [=issuer=]
 and the [=holder=].
+        </dd>
+        <dt><dfn class="export" data-lt="issuee|issuee's">issuee</dfn></dt>
+        <dd>
+A role an [=entity=] performs when it receives a [=verifiable credential=] from
+the [=issuer=].
         </dd>
```
TallTed, line 600: Does this overlap/repeat lines 376-382?
Does the `data-lt` attribute value (now snigular, plural, and possessive) need to include the (singular) string within the `<dfn>`?
```suggestion
        <dt><dfn class="export" data-lt="issue|issuees|issuee's">issuee</dfn></dt>
        <dd>
A role an [=entity=] performs when it receives a [=verifiable credential=] directly
from an [=issuer=].
        </dd>
```

index.html
```diff
@@ -584,13 +592,18 @@ <h2>Terminology</h2>
...
         <dt><dfn class="export" data-lt="issuers|issuer's">issuer</dfn></dt>
         <dd>
 A role an [=entity=] can perform by asserting [=claims=] about one or
 more [=subjects=], creating a [=verifiable credential=] from these
```
TallTed, line 604: As above, does this echo/repeat lines 369-372? 
Does the `data-lt` attribute value (now singular, plural, and possessive) need to include the (singular) string within the `<dfn>`?
```suggestion
        <dt><dfn class="export" data-lt="issuer|issuers|issuer's">issuer</dfn></dt>
        <dd>
A role an [=entity=] can perform by asserting [=claims=] about one or
more [=subjects=], creating a [=verifiable credential=] from these
```

index.html
```diff
@@ -584,13 +592,18 @@ <h2>Terminology</h2>
...
-[=claims=], and transmitting the [=verifiable credential=] to a
-[=holder=].
+[=claims=], and transmitting the [=verifiable credential=] to the
+[=issuee=].
         </dd>
```
TallTed, line 607: As above, does this echo/repeat lines 373-376?
```suggestion
[=claims=], and transmitting the [=verifiable credential=] to an
[=issuee=].
        </dd>
```

TallTed (changes requested):
It's a start...

Wind4Greg (approved):

index.html
TallTed: @dlongley — Good questions! Definitely need to be addressed!

(Such questions are part of why I suggested there be an issue [per repo] for this, as comments on an issue will be more visible in the future than inline PR review comments.)

decentralgabe (commented):
I like the direction of this PR. I believe language should be added that clarifies the relation (or distinction) between issue and holder. Something like "an issuee can be a holder, but a holder may not be an issuee" with examples for each.

TallTed:
> [@decentralgabe] Something like "an issuee can be a holder, but a holder may not be an issuee" with examples for each.

That (inaccurate) comment shows that such language is definitely needed. An accurate version of that statement would be "All issuees are holders, but many holders are not issuees" (a la "all squares are rectangles, but many rectangles are not squares"). 

Something like this might be a start --
```
An [=issuee=] is a [=holder=] that received a [=verifiable credential=] directly
from its [=issuer=]. A [=holder=] may receive a [=verifiable credential=] from its
[=issuer=], its [=issuee=], or any other [=holder=]. A [=verifiable credentials=]
might or might not identify its [=issuee=] as one of its [=claims=]. If not so
identified, the [=issuee=] cannot be known with certainty.
```

decentralgabe:
> All issuees are holders, but many holders are not issuees

In my understanding, one can be an issuee without being a holder. I can create a credential as an issuer about someone (issuee) and not give it to them. 

Holder speaks to possession. Issuee speaks to intention (whether the party is identified or not).

TallTed:
> [@decentralgabe] In my understanding, one can be an issuee without being a holder. I can create a credential as an issuer about someone (issuee) and not give it to them.

In that scenario, you, as an issuer, have created a credential about someone (who is a/the **subject** of that credential). Since you haven't given it to them, I don't think they're an issuee. I'm not sure whether you, as the issuer of the VC, can also be the issuee, or are just a holder, until you pass that VC to someone else, who (I *think*) would need to be identified as issuee within the VC for others to usefully and reliably treat them as such.

> Holder speaks to possession. Issuee speaks to intention (whether the party is identified or not).

I don't quite understand your use-case/user-story.

I can't imagine a scenario where discussing an unidentified issuee is useful.

There *is* always an issuee, who receives the VC from the issuer, but I think that an issuee who is not identified as such within the VC cannot be usefully referenced as the issuee by anyone other than the issuer, and even their reference seems questionably useful to me.

Perhaps you could offer a concrete change to [my suggested paragraph](https://github.com/w3c/vc-data-model/pull/1468#issuecomment-2023709242)?

And/or discuss _issuee_ with @David-Chadwick, hopefully in a public space, so we can all benefit from that conversation?

index.html
David-Chadwick: I believe that current protocols are 1 to 1, not 1 to many. So "the" is appropriate

index.html
David-Chadwick: The word "directly" is not in the current text so I suggest not to make this change. 
I accept your changes to the role sentence, thanks.

index.html
```diff
@@ -420,6 +426,8 @@ <h3>Ecosystem Overview</h3>
           </figcaption>
         </figure>
 
+[PR NOTE. Figure 1 will need updating to replace Holder with Issuee/Holder]
+
```
David-Chadwick, line 430: This note can be removed now, as Figure 1 has been updated.

index.html
David-Chadwick: I am not sure the proposed change to the data-lt is correct. Lets discuss at a WG meeting.
Again, "directly" is not in the current text, so I prefer not to make this change.

index.html
David-Chadwick: I have not made any edits to this text, so this should be a separate PR on the current CR.

index.html
David-Chadwick: See discussion above

index.html
David-Chadwick: After submitting this PR, I (privately) raised with Manu the issue of making changes to our multiple documents now  and he recommended waiting for this PR to be resolved before making any changes to any other of our documents

David-Chadwick: I believe that all current protocols are 1 to 1 and not 1 to many

TallTed: > I believe that all current protocols are 1 to 1 and not 1 to many

But is that a necessary restriction, to be enshrined in this standard? I think one-to-many is, and should be preserved as, a possibility.

index.html
TallTed: The word "directly" is what ensures that the `issuee` does not receive the VC from an intermediary `holder` ... Or is such a relay possible in your mental model? Is `issuee` defined simply as "the entity identified as such within the VC", rather than reflecting anything about the VC's passage from `issuer` to (`holder` to) `issuee`??

index.html
TallTed: > I am not sure the proposed change to the data-lt is correct. Lets discuss at a WG meeting.

Fine.

> Again, "directly" is not in the current text, so I prefer not to make this change.

This (something which the WG needs to decide) seems more important to discuss in a WG call than the `data-lt` change (which is strictly a question of the technicalities of that attribute, and can be answered by anyone more versed in the finer points of ReSpec magic.)

index.html
dlongley: Maybe an `issuee` is one or more entities that an issuer explicitly identifies as intended holders -- and that's all that needs to be said. I don't expect us to come to consensus on any text that reduces the generality we have today around the flow of VCs, but text that identifies some special case of holder roles on top of it is probably acceptable.

index.html
msporny: I'm a strong -1 to updating the diagram to introduce two terms for a single block in the diagram. This will confuse readers -- "Well, which one is it... 'holder' or 'issuee', or is it both?"

Note: This doesn't mean I'm averse to defining the term, but introducing it as a top-level term will harm more than it helps.

index.html
msporny: -1 to "directly" -- the meaning is vague. 

Is the use of a proxy to receive the credential "directly"? Is the use of a browser to receive it into a digital wallet "directly"? Is the use of a hold-and-pickup-later service "directly"? ... and if we explain what "directly" means, does it add value to the understand-ability of the text?

KDean-GS1 (approved):

index.html
TallTed: OK. So, `issuee` is a role *assigned to* an entity (not *performed by* that entity, as no action on that entity's part is required), which assignment, made by the `issuer` of a VC, is achieved by including a `claim` that sets (a? the?) value of the `issuee` property to an identifier of that entity.

By that definition, transmission of the VC to that entity isn't required, which means that the `issuee` is *not* necessarily a `holder`.

(And there's enough hashing out happening through suggestions on this PR that the discussion should probably be moved to an issue or a non-suggestion thread, because the suggestions will vanish if merged, and become hard to see if resolved.)

decentralgabe:
As an alternative I would propose clarifying what "holder" means in different scenarios. A holder is...
1. Someone who is in possession of the credential – holder
2. Someone who is in possession of the credential, who the credential has been issued to – cryptographic holder
3. Someone who is in possession of the credential, who the credential is about (subject) – holder subject

diagrams/ecosystem.svg
```diff
@@ -58,7 +58,7 @@
                             <font style="font-size:16px">
                                 <i>
                                     <b>
-                                        Holder
+                                        Issuee/Holder
```
TallTed, line 61: Based on the totality of conversation below, this label should revert.
```suggestion
                                        Holder
```
Additionally, the line below which now reads `Acquired, stores, presents VCs` should be changed to `Acquires, stores, presents VCs` ( `Acquired` becomes `Acquires`).

I have also just noticed, and am concerned that, this diagram no longer includes any transfers of VCs from `Holder` to `Holder` — only from `Holder` to `Verifier`. I think the `Holder` to `Holder` transfers need to be included here.

index.html
David-Chadwick: @TallTed -  Good suggestion. The issue is #1467 .  I have moved the proposed definition to this issue so that we can continue the discussion there and hence record all the arguments for posterity.

jandrieu:
> As an alternative I would propose clarifying what "holder" means in different scenarios. A holder is...

This might be a good avenue to help clear up confusion, but I'm not sure your language hits the mark.

> 1. Someone who is in possession of the credential – holder

Agreed. This is the current definition.

> 2. Someone who is in possession of the credential, who the credential has been issued to – cryptographic holder

Unfortunately, we cryptography can't prove *who* a credential was issued to. All it can prove is that a current party has access to a cryptographic secret that can prove control over the identifier used for one of the subjects.  *WHO* the initial VC is "issued to" in the sense of "issuee" is the hard problem of confidenceMethod. We can't blithely assert that the secret holder *is* the issuee. Because, in fact, the controller of an identifier may not be involved in the issuance of a VC using that identifier. Presuming that they are is beyond the guarantees of the spec.

> 3. Someone who is in possession of the credential, who the credential is about (subject) – holder subject

This is a subject who is a holder. Jamming two nouns together and pretending its a new noun most likely doesn't help with the confusion. The subject may not be the controller of the identifier. They may not be the holder. 

We have to be exceptionally careful about the implied guarantees. If you are making a statement about the subject, use "subject". If making a statement about the holder, use "holder". At no point, to my consideration, does it seem better to say "subject holder" rather than just "subject". 

I have the same stylistic take on using the term "presenter" and "recipient" to describe the momentary role of a holder when performing those tasks. We could say "presenter holder" or "recipient holder" but since "holder" is well-defined, referring to the "presenter" or the "recipient" simply highlights the action taken by that role, without confusion, IMO. The extra words certainly don't make it easier to read. In fact, I think it makes it more confusing. Concise language is good language.

IMO, confidenceMethod dealt with this appropriately, given issuers a standard way to tell verifiers how they might establish confidence that the presenter of a VC is, in fact, one of the subjects of that VC.

brentzundel:
Please see https://github.com/w3c/vc-data-model/issues/1467#issuecomment-2045437624

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-04-10-vcwg#section2-6) on 2024-04-10

- no resolutions were taken

#### 2.6. Add issuee definition (issue vc-data-model#1467)

_See github issue [vc-data-model#1467](https://github.com/w3c/vc-data-model/issues/1467)._




_See github pull request [vc-data-model#1468](https://github.com/w3c/vc-data-model/pull/1468)._




**Brent Zundel:** ok, we're at time, thanks everyone.  
… before we close, I do want to note issue 1467, I made a chair statement on that, I'm not seeing any new info that justifies the conversation taking further time.  
… that said, if in the comments of the issue and the resulting PR, if people can come to consensus on the way forward, I won't step in the way.  
… I'm just not seeing consensus form.  

---


</details>

msporny:
7 days marked as pending close with no objections, closing.

PR #1470, opened 2024-04-02
Update ecosystem.svg (fix Holder's verb tense)

TallTed:


decentralgabe (approved):

iherman:
> Direct SVG updates will be lost when the source .drawio file is updated. See https://github.com/w3c/vc-data-model/blob/main/diagrams/README.md. In this case perhaps you can get away with updating the string direct in the .drawio file too.

In general, that is true. In this case, the best is to change _both_ the drawio and the svg files. (In theory, a script is run against the generated SVG to remove unnecessary size attributes in the SVG element and format the file; see also the readme. For a single letter change it should be all right to hack both files.)

iherman (approved):

TallTed:
Thanks, @davidlehn. I've patched the `.drawio` to match the patched SVG.

dlongley (approved):

brentzundel (approved):

msporny (approved):

msporny:

Issue #1285, opened 2023-09-17
Pre-CR review from Jeffrey Yasskin


msporny:
This is a review submitted by @jyasskin on the VCDM v2.0:

---------

[kristina] Chairs are in the process of breaking up this issue into smaller issues. we are putting [kristina]/[brent] to the action items we plan to do/already took. these notes will be removed from this issue comment and be opened as a separate issue summarizing all the changes, but for now this is easier to track progress - thanks for understanding.

---------

Hi all,

Brent asked me to look over the VC spec before it gets to CR, so I used a chunk of my flight to do that. I found one maybe-serious issue and a big pile of minor issues and nitpicks. Apologies if I left an issue in here that I figured out later in my read of the document.

General

[Important] Where is the algorithm that a verifier is supposed to run in order to check that the claims in a VC have been vouched by an issuer it trusts?

-> [kristina] will elaborate why one algorithm cannot be defined.

Are "attribute" and "property" synonyms? Probably align on just one of the terms, or clarify the difference.

-> [kristina] same. need post-CR to unify to one.

There are lots of definitions of properties, but very few statements that say which properties may or must appear on Verifiable Credentials. Did I miss a section for that? 4.11 Presentations does it for Verifiable Presentations.

-> [kristina] each section already say what is required or not. can change the language to make it clearer, but not sure it is necessary..

1.2 Ecosystem Overview

It's unfortunate that the different efforts in this space use different terminology. E.g. the verifier would probably be called the "relying party" in an OAuth or FedCM context. An issuer might be an "attester" in a privacy pass context. Would it be possible to mention those translations in this section next to the synonyms that are used in this document? I see that they're defined in 2 Terminology, but I think this might all be clearer if the definitions were in the sections where they naturally appear instead of being duplicated into a single Terminology section.

-> [kristina] Issue opened

1.3 Use Cases and Requirements

As this section is non-normative, it's best to avoid using the normative keywords like "Issuers revoking verifiable credentials should distinguish…" even if those keywords are kept lowercase so that RFC8174 says they don't have their normative meanings. Alternatively, you could move these statements to a section of conformance requirements for issuers (or their other subjects).

-> [kristina] Issue opened

1.4 Conformance

[...]

PR #1484, opened 2024-05-05
Rework content integrity section based on @jyasskin's feedback.

msporny:
This PR is an attempt to partially address issue #1348 by reworking the content integrity section in the specification based on feedback from @jyasskin. 

NOTE: It contains content that was objected to by @selfissued in PR #1464 and is, thus, not mergeable in its current state. This PR will be used to drive the discussion and come to consensus on what can be merged by consensus of the group.

Specifically:

#### 5.4 Integrity of Related Resources

- [x] Editorial clean-up and simplification of normative statements to match other sections. Fixed in b3ac23cafe92eb92980b8c2d4612d0cdd9908b96 and e0c09880ba284d9da44d23b5c151fdf4bf113f7b and re-applied in 692947e2febafad5b5e9ae55bf181d5ab02e245d.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1484.html" title="Last updated on May 27, 2024, 6:57 PM UTC (cc8b0e3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1484/b11eec7...cc8b0e3.html" title="Last updated on May 27, 2024, 6:57 PM UTC (cc8b0e3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3180,16 +3180,10 @@ IANA Media Types</a> registry.
 One or more cryptographic digests, as defined in [[[SRI]]].
                   </td>
                 </tr>
-                <tr>
-                  <td>`digestMultibase`</td>
-                  <td>
-One or more cryptographic digests, as defined in [[[VC-DATA-INTEGRITY]]].
-                  </td>
-                </tr>
               </tbody>
             </table>
 Each object associated with `relatedResource` MUST contain at least a
-`digestSRI` or `digestMultibase` value.
+`digestSRI` value.
           </dd>
         </dl>
 
@@ -3217,7 +3211,7 @@ media type.
         <p>
 Any object in the [=verifiable credential=] that contains an `id`
 property MAY be annotated with integrity information by adding the
-`digestSRI` or `digestMultibase` and `mediaType` properties.
+`digestSRI` and `mediaType` properties.
         </p>
 
         <p>
@@ -3273,18 +3267,6 @@ An example of an object in a `credentialSubject` that is refering to an
 integrity protected image.
         </p>
 
-        <pre class="example nohighlight"
-          title="An integrity-protected image that is associated with a credentialSubject">
-"credentialSubject": {
-  "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
-  "image": {
-    "id": "https://university.example.org/images/58473",
-    "mediaType": "application/svg+xml",
-    "digestMultibase": "zQmdfTbBqBPQ7VNxZEYEj14VmRuZBkqFbiwReogJgS1zR1n"
-  },
-  ...
-}
-        </pre>
       </section>
 
       <section>
--- a/index.html
+++ b/index.html
@@ -3193,7 +3193,7 @@ Each object associated with `relatedResource` MUST contain at least a
           </dd>
         </dl>
 
-        <p class="issue" title="Unification of cryptographic hash expression formats are under discussion">
+        <p class="issue" data-number="1489">
 The Working Group is currently attempting to determine whether cryptographic hash
 expression formats can be unified across all of the VCWG core specifications.
 Candidates for this mechanism include `digestSRI` and `digestMultibase`. There
--- a/index.html
+++ b/index.html
@@ -3265,11 +3265,6 @@ An example of a related resource integrity object referencing JSON-LD contexts.
 }]
         </pre>
 
-        <p>
-An example of an object in a `credentialSubject` that is refering to an
-integrity protected image.
-        </p>
-
       </section>
 
       <section>
--- a/index.html
+++ b/index.html
@@ -3127,16 +3127,16 @@ interoperability.
         <h2>Integrity of Related Resources</h2>
         <p>
 When including a link to an external resource in a [=verifiable credential=], it
-is desirable to know whether the resource has been modified after the
+is desirable to know whether the resource has been modified since the
 [=verifiable credential=] was issued. This applies to cases where there is an
 external resource that is remotely retrieved, as well as to cases where the
-[=issuer=] and/or [=verifier=] might have local cached copies of a resource. It
-is also desirable to know that the contents of the JSON-LD context(s) used in
-the [=verifiable credential=] are the same when used by both the [=issuer=] and
-[=verifier=].
+[=issuer=] and/or [=verifier=] might have locally cached copies of a resource.
+It is also desirable to know that the contents of the JSON-LD context(s) used
+in the [=verifiable credential=] are the same when used by the [=verifier=] as
+they were when used by the [=issuer=].
         </p>
 
-        <p class="issue" title="Mandatory listing of contexts in relatedResouce are under debate.">
+        <p class="issue" title="Mandatory listing of contexts in `relatedResource` is under debate.">
 The requirement that contexts be listed in `relatedResource` is currently being
 debated in the VCWG. This requirement might be removed in future iterations of
 the specification.
@@ -3244,7 +3244,11 @@ implementers.
         <p class="issue">
 The working group is discussing if we will adopt more aspects of subresource
 integrity as defined in [[SRI]] is adopted into the [[JSON-LD11]] specification as
-noted in that specifications <a href="https://www.w3.org/TR/json-ld11/#security">
+noted in that specifications <a href="https://www.w3.org/TR/json-ld11/#security"> 
+current security considerations</a> of that specification, the
+approach described in this section can serve as an additional check towards
+ensuring that a cached context used when issuing
+a [=verifiable credential=] matches the remote resource.
 current security considerations</a> of that specification, the
 approach described in this section can serve as an additional check towards
 ensuring that a cached context used when issuing
--- a/index.html
+++ b/index.html
@@ -3202,7 +3202,7 @@ are arguments for and against unification that the WG is currently debating.
 
         <p>
 If a `mediaType` is listed, implementations that retrieve the resource
-using [[[?RFC9110]]] SHOULD:
+identified by the `id` property using [[[?RFC9110]]] SHOULD:
         </p>
         <ul>
           <li>
@@ -3244,7 +3244,7 @@ implementers.
         <p class="issue">
 The working group is discussing if we will adopt more aspects of subresource
 integrity as defined in [[SRI]] is adopted into the [[JSON-LD11]] specification as
-noted in that specifications <a href="https://www.w3.org/TR/json-ld11/#security"> 
+noted in that specifications <a href="https://www.w3.org/TR/json-ld11/#security">
 current security considerations</a> of that specification, the
 approach described in this section can serve as an additional check towards
 ensuring that a cached context used when issuing
--- a/index.html
+++ b/index.html
@@ -3150,8 +3150,8 @@ To extend integrity protection to a related resource, an [=issuer=] of a
         <dl>
           <dt id="defn-relatedResource">relatedResource</dt>
           <dd>
-The value of the `relatedResource` property MUST be associated with one or
-more objects of the following form:
+The value of the `relatedResource` property MUST be one or more objects of the
+following form:
             <table class="simple">
               <thead>
                 <th>Property</th>
--- a/index.html
+++ b/index.html
@@ -3177,7 +3177,10 @@ IANA Media Types</a> registry.
                 <tr>
                   <td>`digestSRI`</td>
                   <td>
-One or more cryptographic digests, as defined in [[[SRI]]].
+One or more cryptographic digests, as defined by the `hash-expression` ABNF
+grammar defined in the [[[SRI]]] specification,
+<a data-cite="SRI#the-integrity-attribute">Section 3.5: The `integrity`
+attribute</a>.
                   </td>
                 </tr>
               </tbody>
--- a/index.html
+++ b/index.html
@@ -3209,7 +3209,8 @@ identified by the `id` property using [[[?RFC9110]]] SHOULD:
 use the media type in the `Accept` HTTP Header, and
           </li>
           <li>
-use the media type in the `Content-Type` HTTP Header.
+reject the response if it includes a `Content-Type` HTTP Header with a different
+media type.
           </li>
         </ul>
 
--- a/index.html
+++ b/index.html
@@ -3177,13 +3177,13 @@ IANA Media Types</a> registry.
                 <tr>
                   <td>`digestSRI`</td>
                   <td>
-A cryptographic digest, as defined in [[[SRI]]].
+One or more cryptographic digests, as defined in [[[SRI]]].
                   </td>
                 </tr>
                 <tr>
                   <td>`digestMultibase`</td>
                   <td>
-A cryptographic digest, as defined in [[[VC-DATA-INTEGRITY]]].
+One or more cryptographic digests, as defined in [[[VC-DATA-INTEGRITY]]].
                   </td>
                 </tr>
               </tbody>
--- a/index.html
+++ b/index.html
@@ -3250,10 +3250,6 @@ current security considerations</a> of that specification, the
 approach described in this section can serve as an additional check towards
 ensuring that a cached context used when issuing
 a [=verifiable credential=] matches the remote resource.
-current security considerations</a> of that specification, the
-approach described in this section can serve as an additional check towards
-ensuring that a cached context used when issuing
-a [=verifiable credential=] matches the remote resource.
         </p>
         <p>
 An example of a related resource integrity object referencing JSON-LD contexts.
--- a/index.html
+++ b/index.html
@@ -3216,8 +3216,8 @@ media type.
 
         <p>
 Any object in the [=verifiable credential=] that contains an `id`
-property MAY be annotated with integrity information as specified in this
-section.
+property MAY be annotated with integrity information by adding the
+`digestSRI` or `digestMultibase` and `mediaType` properties.
         </p>
 
         <p>
--- a/index.html
+++ b/index.html
@@ -3126,85 +3126,110 @@ interoperability.
       <section>
         <h2>Integrity of Related Resources</h2>
         <p>
-When including a link to an external resource in a [=verifiable credential=],
-it is desirable to know whether the resource that is pointed to is the same at
-signing time as it is at verification time. This applies to cases where there is
-an external resource that is remotely retrieved as well as to cases where the
-[=issuer=] and/or [=verifier=] may have local cached copies of a resource.
+When including a link to an external resource in a [=verifiable credential=], it
+is desirable to know whether the resource has been modified after the
+[=verifiable credential=] was issued. This applies to cases where there is an
+external resource that is remotely retrieved, as well as to cases where the
+[=issuer=] and/or [=verifier=] might have local cached copies of a resource. It
+is also desirable to know that the contents of the JSON-LD context(s) used in
+the [=verifiable credential=] are the same when used by both the [=issuer=] and
+[=verifier=].
         </p>
-        <p>
-It is also desirable to know that the contents of the JSON-LD context(s) used in
-the [=verifiable credential=] are the same when used by both the
-[=issuer=] and [=verifier=].
-        </p>
-        <p>
-To validate that a resource referenced by a [=verifiable credential=] is the
-same at verification time as it is at issuing time, an implementer MAY include a
-property named <code id="defn-relatedResource">relatedResource</code> that
-stores an array of objects that describe additional integrity metadata about
-each resource referenced by the [=verifiable credential=]. If
-`relatedResource` is present, there MUST be an object in the array
-for each remote resource for each context used in the verifiable credential.
-        </p>
+
         <p class="issue" title="Mandatory listing of contexts in relatedResouce are under debate.">
 The requirement that contexts be listed in `relatedResource` is currently being
 debated in the VCWG. This requirement might be removed in future iterations of
 the specification.
         </p>
+
         <p>
-Each object in the `relatedResource` array MUST contain the
-following: the [[URL]] to the resource named `id` and the
-<code id="defn-digestSRI">digestSRI</code> information for the resource
-constructed using the method specified in
-<a href="https://www.w3.org/TR/SRI/#integrity-metadata">Subresource Integrity</a>.
+To extend integrity protection to a related resource, an [=issuer=] of a
+[=verifiable credential=] MAY include the `relatedResource` property:
         </p>
+
+        <dl>
+          <dt id="defn-relatedResource">relatedResource</dt>
+          <dd>
+The value of the `relatedResource` property MUST be associated with one or
+more objects of the following form:
+            <table class="simple">
+              <thead>
+                <th>Property</th>
+                <th>Description</th>
+              </thead>
+              <tbody>
+                <tr>
+                  <td>`id`</td>
+                  <td>
+The identifier for the resource is REQUIRED and conforms to the format defined
+in Section [[[#identifiers]]]. The value MUST be unique among the list of
+related resource objects.
+                  </td>
+                </tr>
+                <tr>
+                  <td>`mediaType`</td>
+                  <td>
+An OPTIONAL valid media type as listed in the
+<a href="https://www.iana.org/assignments/media-types/media-types.xhtml">
+IANA Media Types</a> registry.
+                  </td>
+                </tr>
+                <tr>
+                  <td>`digestSRI`</td>
+                  <td>
+A cryptographic digest, as defined in [[[SRI]]].
+                  </td>
+                </tr>
+                <tr>
+                  <td>`digestMultibase`</td>
+                  <td>
+A cryptographic digest, as defined in [[[VC-DATA-INTEGRITY]]].
+                  </td>
+                </tr>
+              </tbody>
+            </table>
+Each object associated with `relatedResource` MUST contain at least a
+`digestSRI` or `digestMultibase` value.
+          </dd>
+        </dl>
+
         <p class="issue" title="Unification of cryptographic hash expression formats are under discussion">
 The Working Group is currently attempting to determine whether cryptographic hash
 expression formats can be unified across all of the VCWG core specifications.
 Candidates for this mechanism include `digestSRI` and `digestMultibase`. There
 are arguments for and against unification that the WG is currently debating.
         </p>
+
         <p>
-There MUST NOT be more than one object in the `relatedResource` per
-`id`.
-        </p>
-        <p>
-An object in the `relatedResource` array MAY contain a property named
-`mediaType` that indicates the expected media type for the indicated
-`resource`. If a `mediaType` is included, its value
-SHOULD:
+If a `mediaType` is listed, implementations that retrieve the resource
+using [[[?RFC9110]]] SHOULD:
         </p>
         <ul>
           <li>
-be a valid media type as listed in the
-<a href="https://www.iana.org/assignments/media-types/media-types.xhtml">
-IANA Media Types</a> registry
+use the media type in the `Accept` HTTP Header, and
           </li>
           <li>
-be used when retrieving the content, such as via the `Accept` HTTP Header
-          </li>
-          <li>
-match the retrieved content media type, such as via the `Content-Type` HTTP
-Header.
+use the media type in the `Content-Type` HTTP Header.
           </li>
         </ul>
 
         <p>
-Any object in the [=verifiable credential=] that contains an `id` [[URL]]
+Any object in the [=verifiable credential=] that contains an `id`
 property MAY be annotated with integrity information as specified in this
-section by inclusion of `digestSRI`
-in the object.
+section.
         </p>
+
         <p>
-Any objects for which selective disclosure is desired SHOULD NOT be included as
-an object in the `relatedResource` array.
+Any objects for which selective disclosure or unlinkable disclosure is desired
+SHOULD NOT be included as an object in the `relatedResource` array.
         </p>
+
         <p>
 Specification authors that write algorithms that fetch a resource based on the
 `id` of an object inside a [=conforming document=] need to consider whether
 that resource's content is vital to the validity of that document. If it is, the
-specification MUST produce a validation error unless the resource has the
-expected media type and its bytes hash to the expected digest.
+specification MUST produce a validation error unless the resource matches the
+expected media type and cryptographic digest.
         </p>
         <p>
 Implementers are urged to consult appropriate sources, such as the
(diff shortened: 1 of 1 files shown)


dlongley (approved):

index.html
```diff
@@ -3040,110 +3040,136 @@ <h4>Semantic Interoperability</h4>
...
+                  <td>
+A cryptographic digest, as defined in [[[SRI]]].
+                  </td>
+                </tr>
+                <tr>
+                  <td>`digestMultibase`</td>
+                  <td>
+A cryptographic digest, as defined in [[[VC-DATA-INTEGRITY]]].
+                  </td>
+                </tr>
```
selfissued, line 3102: This pulls the use of digestMultibase into the VCDM specification, which I object to, per previous discussions.  This is particularly unnecessary, when we already have a perfectly good W3C digest standard to use - digestSRI.  Choosing one digest standard would increase interoperability.

Please delete the lines adding the use of digestMultibase to the specification.

selfissued (changes requested):
Please delete the lines adding the use of digestMultibase to the specification.

index.html
jyasskin: While I don't have an opinion on whether to include multibase in the core VC data model, I agree that it's more than an editorial change and not necessary to fix the issue I commented on in [#1285](https://github.com/w3c/vc-data-model/issues/1285#:~:text=5.4%20Integrity%20of%20Related%20Resources).

index.html
```diff
@@ -3040,110 +3040,136 @@ <h4>Semantic Interoperability</h4>
...
-<a href="https://www.w3.org/TR/SRI/#integrity-metadata">Subresource Integrity</a>.
+To extend integrity protection to a related resource, an [=issuer=] of a
+[=verifiable credential=] MAY include the `relatedResource` property:
         </p>
+
+        <dl>
+          <dt id="defn-relatedResource">relatedResource</dt>
+          <dd>
+The value of the `relatedResource` property MUST be associated with one or
+more objects of the following form:
```
jyasskin, line 3068: I think I've gotten lost in the RDF/JSON-LD terminology here. As I understand it, the `relatedResource` property appears in a [claim](https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1484.html#claims) that has a subject and a value. In this document, most uses of "X is associated with Y" appear to mean that there's a claim in which Y is the property and X is the value, although I think I see [some](https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1484.html#types) where Y is the subject instead.

If we interpret "associated with" to just mean "appears in the same claim as", then this sentence means that there's one claim (X, `relatedResource`, Y), and then there must be other claims of the form (Y, P, Z) where Z is an object with `id`, `mediaType`, etc. fields. Do you actually just mean that Y must be an object with those properties?

We might have another editorial issue that "associated with" should generally be replaced with better-defined terms.

index.html
```diff
@@ -3040,110 +3040,136 @@ <h4>Semantic Interoperability</h4>
...
+                <th>Property</th>
+                <th>Description</th>
+              </thead>
+              <tbody>
+                <tr>
+                  <td>`id`</td>
+                  <td>
+The identifier for the resource is REQUIRED and conforms to the format defined
+in Section [[[#identifiers]]]. The value MUST be unique among the list of
+related resource objects.
```
jyasskin, line 3080: Doesn't this uniqueness just fall out of how JSON-LD works? Any two objects in a JSON-LD serialization with the same ID are just the same RDF object, right?

index.html
```diff
@@ -3040,110 +3040,136 @@ <h4>Semantic Interoperability</h4>
...
-There MUST NOT be more than one object in the `relatedResource` per
-`id`.
-        </p>
-        <p>
-An object in the `relatedResource` array MAY contain a property named
-`mediaType` that indicates the expected media type for the indicated
-`resource`. If a `mediaType` is included, its value
-SHOULD:
+If a `mediaType` is listed, implementations that retrieve the resource
+using [[[?RFC9110]]] SHOULD:
```
jyasskin, line 3119: What resource? I think it's the ID of Y in the (X, `relatedResource`, Y) claim, but it could also be X's ID.

index.html
```diff
@@ -3040,110 +3040,136 @@ <h4>Semantic Interoperability</h4>
...
+is desirable to know whether the resource has been modified after the
+[=verifiable credential=] was issued. This applies to cases where there is an
+external resource that is remotely retrieved, as well as to cases where the
+[=issuer=] and/or [=verifier=] might have local cached copies of a resource. It
+is also desirable to know that the contents of the JSON-LD context(s) used in
+the [=verifiable credential=] are the same when used by both the [=issuer=] and
+[=verifier=].
         </p>
+
         <p class="issue" title="Mandatory listing of contexts in relatedResouce are under debate.">
```
jyasskin, line 3053: Spelling:

```suggestion
        <p class="issue" title="Mandatory listing of contexts in relatedResource are under debate.">
```

index.html
```diff
@@ -3040,110 +3040,136 @@ <h4>Semantic Interoperability</h4>
...
-IANA Media Types</a> registry
+use the media type in the `Accept` HTTP Header, and
           </li>
           <li>
-be used when retrieving the content, such as via the `Accept` HTTP Header
-          </li>
-          <li>
-match the retrieved content media type, such as via the `Content-Type` HTTP
-Header.
+use the media type in the `Content-Type` HTTP Header.
```
jyasskin, line 3126: Since the requester doesn't set the `Content-Type` header, do you mean "reject the response if the response includes a `Content-Type` HTTP header with a different media type"?

index.html
```diff
@@ -3040,110 +3040,136 @@ <h4>Semantic Interoperability</h4>
...
         </p>
+
         <p>
-Each object in the `relatedResource` array MUST contain the
-following: the [[URL]] to the resource named `id` and the
-<code id="defn-digestSRI">digestSRI</code> information for the resource
-constructed using the method specified in
-<a href="https://www.w3.org/TR/SRI/#integrity-metadata">Subresource Integrity</a>.
+To extend integrity protection to a related resource, an [=issuer=] of a
+[=verifiable credential=] MAY include the `relatedResource` property:
```
jyasskin, line 3061: Are there any constraints on the allowable subject of the `relatedResource` property? If the constraints are based on the `id` of the value of that property, I think the current text requires implementations to recursively scan everything the issuer could be said to have "included" in order to find all uses of the property. That'd be easy for an RDF processor but probably not a JSON processor.

index.html
```diff
@@ -3040,110 +3040,136 @@ <h4>Semantic Interoperability</h4>
...
           </li>
         </ul>
 
         <p>
-Any object in the [=verifiable credential=] that contains an `id` [[URL]]
+Any object in the [=verifiable credential=] that contains an `id`
 property MAY be annotated with integrity information as specified in this
-section by inclusion of `digestSRI`
-in the object.
+section.
```
jyasskin, line 3133: This section doesn't directly say how to annotate an object that contains an `id` property. In particular, do you make it the value of a `relatedResource` property or the subject? Normally I'd "annotate" something by making it the subject of a claim, but I think `relatedResource` is the opposite.

index.html
```diff
@@ -3040,110 +3040,136 @@ <h4>Semantic Interoperability</h4>
...
+                <tr>
+                  <td>`digestSRI`</td>
+                  <td>
+A cryptographic digest, as defined in [[[SRI]]].
+                  </td>
+                </tr>
+                <tr>
+                  <td>`digestMultibase`</td>
+                  <td>
+A cryptographic digest, as defined in [[[VC-DATA-INTEGRITY]]].
```
jyasskin, line 3100: Can an object have more than one of each kind of digest specified? If you write "a", I bet JSON-based processors will forget to look for arrays.

index.html
```diff
@@ -3040,110 +3040,136 @@ <h4>Semantic Interoperability</h4>
...
         </p>
+
         <p>
 Specification authors that write algorithms that fetch a resource based on the
 `id` of an object inside a [=conforming document=] need to consider whether
 that resource's content is vital to the validity of that document. If it is, the
-specification MUST produce a validation error unless the resource has the
-expected media type and its bytes hash to the expected digest.
+specification MUST produce a validation error unless the resource matches the
+expected media type and cryptographic digest.
```
jyasskin, line 3146: Is this constraint only for objects that are the value of a `relatedResource` claim, or any subject of the `mediaType` or `digestSRI` claims?

index.html
```diff
@@ -3040,110 +3040,136 @@ <h4>Semantic Interoperability</h4>
...
-        <p>
-To validate that a resource referenced by a [=verifiable credential=] is the
-same at verification time as it is at issuing time, an implementer MAY include a
-property named <code id="defn-relatedResource">relatedResource</code> that
-stores an array of objects that describe additional integrity metadata about
-each resource referenced by the [=verifiable credential=]. If
-`relatedResource` is present, there MUST be an object in the array
-for each remote resource for each context used in the verifiable credential.
+When including a link to an external resource in a [=verifiable credential=], it
+is desirable to know whether the resource has been modified after the
```
TallTed, line 3044:
```suggestion
is desirable to know whether the resource has been modified since the
```

index.html
```diff
@@ -3040,110 +3040,136 @@ <h4>Semantic Interoperability</h4>
...
+[=verifiable credential=] was issued. This applies to cases where there is an
+external resource that is remotely retrieved, as well as to cases where the
+[=issuer=] and/or [=verifier=] might have local cached copies of a resource. It
+is also desirable to know that the contents of the JSON-LD context(s) used in
+the [=verifiable credential=] are the same when used by both the [=issuer=] and
+[=verifier=].
```
TallTed, line 3050:
```suggestion
[=issuer=] and/or [=verifier=] might have locally cached copies of a resource.
It is also desirable to know that the contents of the JSON-LD context(s) used
in the [=verifiable credential=] are the same when used by the [=verifier=] as
they were when used by the [=issuer=].
```

index.html
```diff
@@ -3040,110 +3040,136 @@ <h4>Semantic Interoperability</h4>
...
+Each object associated with `relatedResource` MUST contain at least a
+`digestSRI` or `digestMultibase` value.
+          </dd>
+        </dl>
+
         <p class="issue" title="Unification of cryptographic hash expression formats are under discussion">
 The Working Group is currently attempting to determine whether cryptographic hash
 expression formats can be unified across all of the VCWG core specifications.
 Candidates for this mechanism include `digestSRI` and `digestMultibase`. There
 are arguments for and against unification that the WG is currently debating.
```
TallTed, line 3114: It seems like there should be a numbered issue associated with this "Unification of cryptographic hash expression formats" discussion, where the "arguments for and against unification" (and description of what is meant by "unification", here) are (or can be) clearly enumerated, including those for and against inclusion of `digestMultibase` and/or `digestSRI`, also including possibly removing both (since including either one alone can be read as picking a winner where we perhaps should not), and thus debated more clearly....

index.html
```diff
@@ -3040,110 +3040,136 @@ <h4>Semantic Interoperability</h4>
...
 Commercial National Security Algorithm Suite 2.0</a> to ensure that they are
-chosing a current and reliable hash algorithm. At the time of this writing
+choosing a current and reliable hash algorithm. At the time of this writing
 `sha384` SHOULD be considered the minimum strength hash algorithm for use by
 implementers.
         </p>
         <p class="issue">
 The working group is discussing if we will adopt more aspects of subresource
 integrity as defined in [[SRI]] is adopted into the [[JSON-LD11]] specification as
 noted in that specifications <a href="https://www.w3.org/TR/json-ld11/#security">
```
TallTed, line 3161: I cannot decipher this sentence (which forms this entire paragraph, which I cannot select in its entirety, since half of it's being changed by this PR) sufficiently to rephrase it more clearly.

TallTed (changes requested):

David-Chadwick (commented):
I think the related resource additions are not pure editorial changes (as requested) but are technical additions so they should be in a separate PR

brianorwhatever (commented):
This requires implementers that want to verify the integrity of related resources to handle both SRI and multibase. There is already enough optionality in VCs. If the WG can't come to consensus on SRI vs multibase it should decide on one hash algorithm and use it (ie sha256).

`"digest": "c2ac819c487c5fb6380a9c5d97c33ed28d540e881ecf067d0f692ebbc356326e"`

index.html
TallTed: fix plural/singular disagreement (and spelling, as @jyasskin caught)
```suggestion
        <p class="issue" title="Mandatory listing of contexts in `relatedResource` is under debate.">
```

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-05-08-vcwg#section2-1) on 2024-05-08

- no resolutions were taken

#### 2.1. Rework content integrity section based on @jyasskin's feedback. (pr vc-data-model#1484)

_See github pull request [vc-data-model#1484](https://github.com/w3c/vc-data-model/pull/1484)._




**Manu Sporny:** The content integrity section has a number of objections on it. I presume we'll talk about those at some point. The rest of the specs are doing their normal thing, that's it from me.  

> *Dmitri Zagidulin:* do you have links to the content integrity objections?

_See github pull request [vc-data-model#1484](https://github.com/w3c/vc-data-model/pull/1484)._




**Brent Zundel:** We have some time to look at the content integrity PR.  
… we have some time to look at that PR.  

</details>

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-05-08-vcwg#section2-1) on 2024-05-08

- no resolutions were taken

#### 2.1. Rework content integrity section based on @jyasskin's feedback. (pr vc-data-model#1484)

_See github pull request [vc-data-model#1484](https://github.com/w3c/vc-data-model/pull/1484)._




**Manu Sporny:** The content integrity section has a number of objections on it. I presume we'll talk about those at some point. The rest of the specs are doing their normal thing, that's it from me.  

> *Dmitri Zagidulin:* do you have links to the content integrity objections?

_See github pull request [vc-data-model#1484](https://github.com/w3c/vc-data-model/pull/1484)._




**Brent Zundel:** We have some time to look at the content integrity PR.  
… we have some time to look at that PR.  

</details>

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-05-08-vcwg#section2-3) on 2024-05-08

- no resolutions were taken

#### 2.3. Rework content integrity section based on @jyasskin's feedback. (pr vc-data-model#1484)

_See github pull request [vc-data-model#1484](https://github.com/w3c/vc-data-model/pull/1484)._




**Brent Zundel:** This PR has had a number of comments on it, number of requests for changes, Manu could you talk us through it?  
… Folks that would like to see changes, please do so.  

**Manu Sporny:** This PR started out as a request from Jeffrey Yaskin to rewrite the content integrity section so that it matched the pattern of the other sections. This section was originally written by Mike Prorock, it was the first attempt at it and it went through a lot of churn and needed clean up.  
… There were a number of things agreed to -- as far as "please raise an issue noting X and Y" in the content integrity section. There was not agreement on the digest format we would use, but there was agreement that it would be between `digestSRI` and `digestMultibase` or both and leave it to implementers to decide on which ones to implement, if not both of them.  
… This section was largely rewritten editorially, but there are objections to the way it was rewritten because we now mention both `digestSRI` and `digestMultibase` in the spec; one argument is that was always true, there was an issue marker in the spec that said we would rely on implementers to pick.  
… The counterargument to that is that we never agreed to that but that's challenged as well. Largely this is an editorial rewrite of the section and the thing that's under contention is that `digestMultibase` is being specified more clearly based on the issue marker.  
… I would be interested in hearing Mike's concerns here and Jeffrey Yaskin has also jumped in with good questions we should answer and while we're reworking the section we should address those. 80% are questions with answers, 20% we might want to discuss.  

**Michael Jones:** Obviously, I'm the one who had made the point that we shouldn't be normatively adding multibase to the VCDM spec since we have been careful to not include it before. I understand that there is an issue comment, but I'm not willing to elevate multibase beyond that issue comment.  

**David Chadwick:** This new text should have removed the note, suggest removing that bit, rework editorial, and discuss later.  

**Manu Sporny:** Yes, I think that the new text, actually, if you look at it carefully, the changes have been made, it's clearly addressing Jeffrey's comments. I think the new text on multibase isn't replacing anything. We should have struck out the note. Then we can put that into a new PR and just work on that PR alone.  

**Brent Zundel:** ok, so smaller more controversial bit to its own PR.  

**Dmitri Zagidulin:** I'm curious, with regards to `digestMultibase`, since Mike's objection seems to be primarily to the base encoding part, perhaps not the multihash aspect of it, I'm wondering if there's still time or opportunity to change direction and require the base to be base64url and have it be a digest multihash, which provides significant size benefits over the digest SRI approach.  
… Would Manu and other proponents consider pivoting and leaving the multihash aspect.  

**Manu Sporny:** Sure. I think there's a misunderstanding about what's in the spec today. Today we normatively reference DI and VC-JOSE-COSE -- because this group decided that because we're publishing those we will normatively reference those from VCDM. As a result, we can pull in those normative things from those specs. If the concern is not defining them in VCDM, that's not happening.  
… We're using it from the other specs. The more confusing thing is that we're taking the definitions from DI -- and moved them into the controller document. The thing with the normative definition for multibase is in the controller document. Now we'd have to ref the controller document from the VC data model. In the same way we ref SRI.  
… We can't just ref SRI, we have to specify a field and the format of that field and we have language around that. We have to define it somewhere. We have to figure out where to define it -- we could define it in controller document, we could define it in VC-JOSE-COSE, we could define it in DI. I don't have a strong feeling where it goes.  
… But to use that feature we have to talk about it in the VCDM. I don't quite understand what Mike's objection -- are you objecting to where it's defined (it must be defined somewhere, but we have to normatively reference it to use it). I don't really care where it's defined but it has to be somewhere, the easiest place is in the VCDM, but we can put it in DI or somewhere else.  
… We could throw it all in the controller document.  

**Brent Zundel:** Just to summarize, currently the VCDM normatively references `digestSRI` and defines a profile of how we're using it?  

**Manu Sporny:** No, it references SRI, which is a web browser thing. We use a tiny piece of SRI but we can't just do that. We have to define what the value is in the `@context`, its range, how it's encoded, etc. SRI doesn't do that, it's targeted at browsers.  
… We chose to do that work in VCDM, it's a couple of sentences, it's simple. But we have to do that for `digestMultibase` as well, but it's already defined in DI and we can just ref it.  

**Brent Zundel:** That was helpful.  

**Michael Jones:** As Dmitri points out, the problem with digestMultibase is its dependent upon multibase, where inexplicably it defines something like 26 encodings for the same data, which is a travesty for interop.  
… my objections are not about where this is defined, it's to using it at all, because it hurts interop. I'm on public record about that.  
… The multibase encoding is a failure to make a choice, we should steer clear of it.  

> *Dmitri Zagidulin:* @dlongley - but if we're gonna use 'u' prefix only, why don't we just specify that we're using u, and save the extra character.

**Dave Longley:** 2nd Dmitri's proposal to resolve this, to say that we only support the base64url prefix, which is the multibase prefix for base64url, nopad encoded, if we do that, then that gets us past the objection, then we get data savings, and so on.  

**Brent Zundel:** proposal on the table is we specify which base and do that in this section of VCDM. Mike, would that address your objections?  

**Dmitri Zagidulin:** I like Dave's suggestion, but if we're fixing that we're just fixing base64url, and drop the 'u' and save ourselves an extra character.  

**Michael Jones:** Well, Dmitri's suggestion is rational, threw me off. In response to brent's suggestion, I'm unwilling to define any form of digestMultibase in our core spec, I understand the irony, which defines this in controller document, and I do appreciate we got to compromise to DI where we don't normatively ref 26 choices, but we decide w/ what Dmitri said, what choices we make, I do not know whether a form of digestMultibase is defined in controller document, if we're going to define that w/ base64url encoding I'd do it there to keep everything together. The other question, I do support defining the fields necessary to use digestSRI in our spec in the Content Integrity section. Finally, I haven't read Jeffrey's comments, read them w/ eye, is he asking for Multibase to be specified. or is he asking unrelated editorial things?  

**Dave Longley:** One of the advantages of using multibase is it introduces layers and a separation of concerns, so you can compress any multibase data regardless of whichever data is done, if you want to compress using CBOR-LD, it can do that sort of compression, or use tech that is not a VC, different choices on encoding, here in our group we can say base64url, no one has to go rewrite new tech, type is multibase, no new tech needs to be written to that particular field, that other tech continues to work, we get benefit that there is interop around that prefix, we get best of both worlds if we make that choice.  

> *Dmitri Zagidulin:* @dlongley - since that's driven off of an RDF data type, can't we define a base64url data type, and have processors just work off that/.

**Dmitri Zagidulin:** Dave's point about CBOR-LD processors makes a lot of sense, since signal processors are working off of, RDF type multibase, why dont' we define RDF type for base64url, same logic, same compression can be invoked?  

> *Dmitri Zagidulin:* but there's like only 3-4 encodings realistically (base64url, base32lower, base32upper... like what else are you gonna use).

**Dave Longley:** What we're suggesting now is a new RDF type for every encoding type, or we can re-use work out there, already implementations using that... if the world had decided to take a different approach, and communities were using RDF, then that might have worked, but there are separations of concerns, trying to re-use technologies, if we re-invent yet another way to do it, we're less interoperable.  
… simplest thing to do is take implementations already out there, will work w/ other tooling, we should invent fewer things, we should profile it down, it addresses Mike's concern, they don't have to use it either and will just work w/ VCs w/ one prefix.  

**Michael Jones:** I guess I disagree with the characterization, it's not inventing something, it's something that's been around for a decade and a half. If we need context and vocabulary term, let's create it.  

> *Michael Jones:* Multiformats Considered Harmful [https://self-issued.info/?p=2408](https://self-issued.info/?p=2408).

**Michael Jones:** To Dave's point, that's what people have already used, that people chose to not make a choice doesn't mean we should promulgate that or promote that.  
… I did a blog post on the issue, I think multibase is an interop mistake and we should get rid of it, if we chose a prefix, we're half getting rid of it, we'll see how this works out.  

**Dave Longley:** First to clarify, I didn't mean to imply that base64url encoding , the alphabet is what's being invented. We'd be inventing a new RDF type for a single base-encoding. Regarding failure to make a choice, depends on whether or not people feel things are equivalent. Community that built Multibase had their reasons. JWK allows different keys to be parameterized instead of saying one key format. We tried to re-use that work, and we've successfully been interoping with that technology. I would prefer us to not invent a new thing and cut it there because it seems like picking a prefix of U and putting it on there solves all the problems doesn't solve... we're building on other communities to greatest extend that we can profile those down, we should do that, instead of inventing something new.  

**Manu Sporny:** There are a number of misconceptions around what multibase is including in that post that Mike wrote.  
… I think we're talking about reinventing the wheel here, we're going to invent a single RDF type for a specific encoding of base64url-no-pad.  
… That is hyper-specializing the encoding format when we don't need to do that when we can just put a `u` on the front and say VCs use that and that addresses the "not making a choice" thing that Mike is concerned about.  
… Why would people choose different base encodings? There are legitimate reasons for doing so. Base32, for example, when encoding in a QR format, compresses way better than base64url. There are technical reasons for it.  
… You can't think that all encodings work for all use cases. When you get into embedded and compression use cases, those encodings matter.  
… There are totally valid reasons why people would want to use different base encodings and multibase understands that and that not every base encoding works everywhere. Base64 doesn't work in URLs. That's what Base64Url was created.  
… There are many ways to send things and the ecosystem matters, it's not a failure to decide. It's understanding that the world is bigger than just us in this group and utilizing a way that lets you signal what base encoding is used so we can just use the multibase format across all the layers instead of having to reinvent the stuff at every layer.  
… I appreciate you wrote that article but I think it's off base and it's not the reason multibase exists, it misses the point.  

**Brent Zundel:** It sounds like, maybe overly optimistic, it sounds like we're dancing around a middle ground that not everyone likes but we can live with, use base64url encoding, that's the choice we're making for multibase option for content integrity section.  
… feels like we're close there.  

> *Dave Longley:* the effect is larger than the invention of a tag itself.

**Michael Jones:** Mainly going to respond to Dave's comment, defining base64url RDF type would be invention, perhaps, but it's not a very big invention, it's just tagging what this field uses. Creating the tag is perhaps an invention, but that's a good thing, it let's people know what choice was made.  

> *Ted Thibodeau Jr.:* I think a better title for that blog post would be "I Consider Multiformats Harmful" as it doesn't seem to incorporate anyone else's opinion.

**Brent Zundel:** please chime in in the PR, let's explore that middle ground.  

---


</details>

index.html
msporny: > Are there any constraints on the allowable subject of the relatedResource property?

No, there are no constraints. Any URL would do.

> That'd be easy for an RDF processor but probably not a JSON processor.

Hmm, I don't understand why? Why wouldn't a "JSON processor" just look for "id" values?

index.html
msporny: Hrm, "associated with" in this case is meant to mean: "The value of the `relatedResource` property MUST be one or more objects of the following form". IOW, I think we can just delete the word "associated with" here and be fine.

index.html
msporny: > Doesn't this uniqueness just fall out of how JSON-LD works? 

Yes.

> Any two objects in a JSON-LD serialization with the same ID are just the same RDF object, right?

Correct.

However, there are people in the WG that don't want to do "full JSON-LD processing" and this language helps them avoid writing code that merges duplicates in the list. 

We advise against going "full RDF" and creating multiple nodes with the same identifier in VCs; people that do that shouldn't expect to interoperate as well as if they would just avoid doing that. 

I agree that the benefit of this language is debatable.

index.html
```diff
@@ -3126,110 +3126,140 @@ <h4>Semantic Interoperability</h4>
...
+                  <td>
+An OPTIONAL valid media type as listed in the
+<a href="https://www.iana.org/assignments/media-types/media-types.xhtml">
+IANA Media Types</a> registry.
+                  </td>
+                </tr>
+                <tr>
+                  <td>`digestSRI`</td>
+                  <td>
+A cryptographic digest, as defined in [[[SRI]]].
```
msporny, line 3180:
```suggestion
One or more cryptographic digests, as defined in [[[SRI]]].
```

index.html
```diff
@@ -3126,110 +3126,140 @@ <h4>Semantic Interoperability</h4>
...
+                  </td>
+                </tr>
+                <tr>
+                  <td>`digestMultibase`</td>
+                  <td>
+A cryptographic digest, as defined in [[[VC-DATA-INTEGRITY]]].
```
msporny, line 3186:
```suggestion
One or more cryptographic digests, as defined in [[[VC-DATA-INTEGRITY]]].
```

index.html
msporny: Good point. Fixed in https://github.com/w3c/vc-data-model/pull/1484/commits/eed1251f3b2f17d93ac8448a074dd282a352a92a.

index.html
msporny: Clarified in b0896f222209d57814f2cec9814a46d8bcdd0d78.

index.html
msporny: Good catch, fixed in ac98e170076b7b4ebcf42b26ce03092d5c79b242.

index.html
msporny: Fixed in f72d137c6facbb267b5d3b3fdaf9dc50b12f8d65.

index.html
msporny: Its for any subject of the `digestSRI` or `digestMultibase` claims. The previous sentence specifies that expectation.

index.html
```diff
@@ -3126,110 +3126,141 @@ <h4>Semantic Interoperability</h4>
...
+                <tr>
+                  <td>`digestSRI`</td>
+                  <td>
+One or more cryptographic digests, as defined in [[[SRI]]].
+                  </td>
+                </tr>
+                <tr>
+                  <td>`digestMultibase`</td>
+                  <td>
+One or more cryptographic digests, as defined in [[[VC-DATA-INTEGRITY]]].
```
iherman, line 3186: We should use a direct reference to the multibase specification in the DI spec, that is §2.4:
```suggestion
One or more cryptographic digests, as defined in <a data-cite="VC-DATA-INTEGRITY#multibase-0">VC Data Integrity</a> specification.
```

index.html
```diff
@@ -3126,110 +3126,141 @@ <h4>Semantic Interoperability</h4>
...
+                  <td>
+An OPTIONAL valid media type as listed in the
+<a href="https://www.iana.org/assignments/media-types/media-types.xhtml">
+IANA Media Types</a> registry.
+                  </td>
+                </tr>
+                <tr>
+                  <td>`digestSRI`</td>
+                  <td>
+One or more cryptographic digests, as defined in [[[SRI]]].
```
iherman, line 3180: The reference to the SRI specification may be a bit problematic; that specification describes a general framework for HTML, and we use only one part of it, namely the [integrity attribute](https://www.w3.org/TR/SRI/#the-integrity-attribute). To be even more precise, we use the BNF definition therein. 

Note that we have, in our spec, a precise definition for an [SRI string definition](https://www.w3.org/TR/SRI/#the-integrity-attribute) which does that precise reference. Wouldn't it be better to refer directly to that section?

iherman (approved):

index.html
msporny: Removed in 869ebeb69796c39a341dc78a41674b48e966f9c9 with issue #1489 raised to track the disagreement in this PR (and previous PRs related to digest mechanism).

msporny:
@selfissued wrote:
> Please delete the lines adding the use of digestMultibase to the specification.

Removed in 869ebeb with issue #1489 raised to track the disagreement in this PR (and previous PRs related to digest mechanism).

index.html
msporny: Issue #1489 has been raised to track the disagreement in this PR, and previous PRs related to digest mechanism, and that issue number is now cited from the specification.

index.html
msporny: Fixed in 4e7d4b366e564d9d158f082a1f58e8f8e7ed4a3a.

index.html
msporny: I've removed the `digestMultibase` entry due to @selfissued's objection. We'll pick that item up in another PR.

msporny:
Editorial, multiple reviews, changes requested and made, objections moved to issue #1489, no remaining objections for this PR, merging.

index.html
jyasskin: What I think you have here is serialized as

```json
{ "id": "vc://id",
  "type": "VerifiableCredential",
  "relatedResource": {
    "id": "the://resource",
    "mediaType": "media/type",
   "digestSRI": "sha384-hash"
  }
}
```

In RDF, that's
```n3
<the://resource> <https://schema.org/encodingFormat> "media/type" .
<the://resource> <https://www.w3.org/2018/credentials#digestSRI> "sha384-hash" .
<vc://id> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <vc://VerifiableCredential> .
<vc://id> <https://www.w3.org/ns/credentials/issuer-dependent#relatedResource> <the://resource> .
```

But I don't see any constraint in the text that the "relatedResource" property has to map from the VC to the constrained resource. The issuer could as easily "include the relatedResource property" as:

```json
{ "id": "vc://id",
  "type": "VerifiableCredential",
  "foo": { "bar": { "baz": { "quux": {
    "relatedResource": {
     "id": "the://resource",
     "mediaType": "media/type",
     "digestSRI": "sha384-hash"
    }
  }}}}
}
```

A JSON processor then has to recursively scan all the objects to look for `relatedResource` properties. RDF APIs seem to have native queries for "find all the triples with a property of X", but that's not a natural operation on JSON.

PR #1491, opened 2024-05-31
update json schemas

### LINKED ISSUE #1486: Inconsistency between spec and schema
As per schema https://github.com/w3c/vc-data-model/blob/0e6d623d15fb6a3ffadf1249580b5ba55d7894e4/schema/verifiable-credential/verifiable-credential-schema.json#L210C6-L225, `credentialSchema` does not allow an array. It differs from other fields   such as `refreshService` or `termsOfUse` where both an array or object are allowed.

> ```
>     "credentialSchema": {
>       "type": "object",
>       "properties": {
>         "id": {
>           "type": "string"
>         },
>         "type": {
>           "$ref": "#/$defs/type"
>         }
>       },
>       "required": [
>         "id",
>         "type"
>       ],
>       "additionalProperties": true
>     },
> ```

However as per the spec in [example 22](https://www.w3.org/TR/vc-data-model-2.0/#example-usage-of-the-credentialschema-property-to-perform-json-schema-validation) and https://www.w3.org/TR/vc-data-model-2.0/#data-schemas it states

> The value of the credentialSchema [property](https://www.w3.org/TR/vc-data-model-2.0/#dfn-property) MUST be one or more data schemas...

Please clarify which is correct, the `schema.json` or the written document.

decentralgabe:
Fix #1486 

- support arrays for credential status, credential schema, evidence
- fix bug in representing type (can be a single value or an array)
- added regex for xml date time


schema/verifiable-credential/verifiable-credential-schema.json
```diff
@@ -172,7 +212,12 @@
       ]
     },
     "validFrom": {
-      "type": "string"
+      "type": "string",
+      "pattern": ".*(Z|(\\+|-)[0-9][0-9]:[0-9][0-9])"
+    },
+    "validUntil": {
+      "type": "string",
+      "pattern": ".*(Z|(\\+|-)[0-9][0-9]:[0-9][0-9])"
```
iherman, line 220: Are we sure about this regex? Valid from/until are supposed to be dates, and these only seem to restrict the time part. Why not using a regex for the full xsd datatype?

schema/verifiable-presentation/verifiable-presentation-schema.json
```diff
@@ -56,7 +53,8 @@
           ]
         },
         "created": {
-          "type": "string"
+          "type": "string",
+          "pattern": ".*(Z|(\\+|-)[0-9][0-9]:[0-9][0-9])"
```
iherman, line 57: Just marking with the same comment as above

schema/verifiable-credential/verifiable-credential-schema.json
msporny: IIUC, these regexes are incorrect. The correct regex can be found here:

https://www.w3.org/TR/vc-data-model-2.0/#example-regular-expression-to-detect-a-valid-xml-schema-1-1-part-2-datetimestamp

schema/verifiable-presentation/verifiable-presentation-schema.json
msporny: Same issue, use the regex here instead:

https://www.w3.org/TR/vc-data-model-2.0/#example-regular-expression-to-detect-a-valid-xml-schema-1-1-part-2-datetimestamp

msporny (changes requested):
Regexes for XMLDatetimeStamp are incorrect. Otherwise, LGTM.

dlongley (approved):
Approving assuming the regexes will be updated as suggested.

decentralgabe:
thanks @iherman @msporny updated the regex!

iherman (approved):

msporny (approved):

TallTed (approved):

decentralgabe:
changes addressed; opened over a week; editorial approval; merging!

Issue #1437, opened 2024-02-14
Remove at risk issue markers for property extension points.


msporny:
The WG discussed the process for removing "AT RISK" issue markers for properties and property sections during the call today. The following was the agreed-upon process (see transcript below). In order for an "AT RISK" marker to be removed, and the property to remain in the specification, the section MUST:

1. Update the example to use an extension specification that has been implemented, and values that utilize the extension point.
2. Have the `type` field specified, using a type specified by the extension specification.
3. A test in the VCDM v2.0 test suite that tests the example by issuing a VC that contains the example.
4. Result in at least two implementations issuing a conforming VC that contains the example.

The following property extension points are currently at risk:

* confidenceMethod
* evidence
* refreshService
* renderMethod
* termsOfUse

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-02-14-vcwg#section2-1) on 2024-02-14

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 2.1. Remove the at risk issue marker for `Evidence` (issue vc-data-model#1303)

_See github issue [vc-data-model#1303](https://github.com/w3c/vc-data-model/issues/1303)._




**Michael Jones:** looking at 1303 first Remove the at risk issue marker for Evidence.  

> *Manu Sporny:* See [section in the spec](https://w3c.github.io/vc-specs-dir/#evidence).

**Manu Sporny:** question for the group: Jefferey Yaskin asked for a mini-registry and to be done with this. we could do that. w.r.t tests what did we say we would do? I have two recollections.  
… have to demonstrate there is a spec using the property. there are multiple impls. 1EdTech have an evidence property. thought the tests we were writing were just testing the 'type' for evidence. need to ask what are we testing for these extension points.  
… normative guidance we give is : it can't be empty, have to specify it's type, id should be a URL, ... 1EdTech has multiple impls. at what point do we remove the at risk marker? when we create tests?  
… is that the bar we're trying to hit?  

**Ivan Herman:** from a practical point of view, the only obligation we have is to remove these markers and feature itself when we go to PR. at this point there is no rush. at some point we'll have to look at the whole test suite report and then risk markers.  
… issue was raised before CR. why bother at this point? went to CR with marker in.  

**Brent Zundel:** less a post-CR and then a pre-PR?  

**Ivan Herman:** yes.  

[...]

PR #1497, opened 2024-06-01
Remove at risk marker for `refreshService` property.

msporny:
This PR is an attempt to partially address issue #1437 by removing the at risk issue marker for the `refreshService` property.

There are [two registered specifications](https://w3c.github.io/vc-specs-dir/#refresh-service) in the VC Specifications Directory that use the extension point. The 1EdTech VC Refresh v1.0 specification uses the extension point. The Conexxus Age Verification v1.0 standard uses the extension point, which has been implemented by (at least) Spruce (California DMV), TruAge, and Digital Bazaar.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1497.html" title="Last updated on Jul 17, 2024, 8:24 PM UTC (45f87c3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1497/b29c107...45f87c3.html" title="Last updated on Jul 17, 2024, 8:24 PM UTC (45f87c3)">Diff</a>

PR #1495, opened 2024-06-01
Reserve `confidenceMethod` and `renderMethod`.

msporny:
This PR is an attempt to partially address issue #1437 by reserving the `confidenceMethod` and `renderMethod` properties. The VCWG is unlikely to define, implement, and test either feature in the next several months and is reserving both properties in order to advance them during a future Working Group.

PR #1498, opened 2024-06-01
Removes the `termsOfUse` section from the specification and reserves the property.

msporny:
This PR is an attempt to partially address issue #1437 by removing the `termsOfUse` section from the specification while continuing to reserve the property for potential future re-definition.

There are [two registered specifications](https://w3c.github.io/vc-specs-dir/#terms-of-use) in the VC Specifications Directory that use the extension point. However, the first registration does not seem to have been implemented and the second registration does not seem to conform to VCDM v2.0 nor does it have any available demonstrations of implementation or continued usage.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1498.html" title="Last updated on Jul 17, 2024, 8:29 PM UTC (6b53cbb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1498/97f7b90...6b53cbb.html" title="Last updated on Jul 17, 2024, 8:29 PM UTC (6b53cbb)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1491,16 +1491,6 @@ A valid [=credential=] status [=type=]. For example,<br>
 
             <tr>
               <td>
-<a href="#terms-of-use">termsOfUse</a>&nbsp;object
-              </td>
-              <td>
-A valid terms of use [=type=]. For example,<br>
-`"type": "ExampleTermsPolicy"`
-              </td>
-            </tr>
-
-            <tr>
-              <td>
 <a href="#evidence">evidence</a>&nbsp;object
               </td>
               <td>
@@ -3152,7 +3142,7 @@ documents with implementations, or enabling aggressive caching of contexts.
 Implementers are advised to pay close attention to the extension points in this
 specification, such as in Sections <a href="#proofs-signatures"></a>,
 <a href="#status"></a>, <a href="#data-schemas"></a>,<a href="#refreshing"></a>,
-<a href="#terms-of-use"></a>, and <a href="#evidence"></a>. While this
+and <a href="#evidence"></a>. While this
 specification does not define concrete implementations for those extension
 points, the Verifiable Credential Specifications Directory [[?VC-SPECS]]
 provides an unofficial, curated list of extensions that developers can use from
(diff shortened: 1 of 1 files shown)


iherman (commented):
I am not against the removal of the "at risk" notes per se, but I would prefer to wait in doing so. There is no urgency in doing this; on the other hand, if those features are not listed as "at risk", it invalidates the argumentation in the new charter proposal (see https://github.com/w3c/vc-wg-charter/pull/119).

The plan is to move ahead with the chartering next week by the W3C management. We should be in a better state to do these at around TPAC.

msporny:
@iherman wrote:
> There is no urgency in doing this

We need to ship VCDM v2.0 and the rest of our specifications, we don't want to drag this out more than necessary.

> I am not against the removal of the "at risk" notes per se, but I would prefer to wait in doing so. 

That's fine, we can wait until after the Charter vote to avoid confusing those voting on the charter. I don't think it's necessary given that the charter specifically calls out the features that could be developed further in v2.1.

dlongley (approved):

decentralgabe (approved):

brentzundel (approved):

TallTed (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-06-05-vcwg#section5-3) on 2024-06-05

- no resolutions were taken

#### 5.3. Removing at risk issue markers.

_See github pull request [vc-data-model#1495](https://github.com/w3c/vc-data-model/pull/1495)._




_See github pull request [vc-data-model#1496](https://github.com/w3c/vc-data-model/pull/1496)._




_See github pull request [vc-data-model#1497](https://github.com/w3c/vc-data-model/pull/1497)._




_See github pull request [vc-data-model#1498](https://github.com/w3c/vc-data-model/pull/1498)._




**Manu Sporny:** 1495, 1496, 197, 1498, each one of these either reserves terms (confidenceMethod, renderMethod), don't have time to finish them but will reserve terms.  
… remove terms of use.  
… had it in the spec before, not enough implementations, keep term reserved but remove section from spec.  
… for all others, enough implementations based on criteria previously agreed on to keep them in spec.  
… refresh service and evidence kept in spec, reserve confidenceMethod and renderMethod.  

**Brent Zundel:** the course this group agreed to at the beginning on extension points, agreed that if implementers exist using these extensions we will keep them in the spec.  
… these PRs are an expression of that intent, at this point it would be inappropriate to challenge the intent.  
… if you have qualms about the PRs it would be inappropriate to be about their course.  
… all are welcome to change their mind but hopefully people stay the course here.  

**Ivan Herman:** my issue is timing not intent, as I said at the beginning of this call, the charter/proposal that goes out says that there is an exception for terms that are at risk but in the spec.  
… if I go out to the AC now and there is no at risk feature in the spec we have a problem.  
… propose we agree with the PRs but do not merge them before the vote is out at the AC.  

**Manu Sporny:** +1, I think we can wait, there will just need to be some maintenance on the PRs.  

**Brent Zundel:** any other PRs we need to look at.  

</details>

David-Chadwick (changes requested):
We spent a lot of time on modifying the terms of use property to ensure that the example in the VCDM conformed to the way that it is being used in practise. At the time there were a number of members who said the ToU was being used by EBSI and so the first example in the spec used the EBSI terms of use (modified to conform to VCDMv2 as opposed to its original VCDMv1). We also agreed that the test suite would include tests for the ToU to ensure that implementation conformed to this specification e.g. by ignoring ToUs that they did not understand. So it is detrimental to remove it now. 
However, I can agree to the ToU property being modified to remove any mention of the holder inserting ToUs into a VP, because I do not believe that we have anyone actually doing this right now (but I might be wrong).

David-Chadwick:
@msporny said 

> There are [two registered specifications](https://w3c.github.io/vc-specs-dir/#terms-of-use) in the VC Specifications Directory that use the extension point. However, the first registration does not seem to have been implemented and the second registration does not seem to conform to VCDM v2.0 nor does it have any available demonstrations of implementation or continued usage.

I have no objection to the extensions directory removing the ToU examples. The second example, authored by myself, conforms to VCDMv1 and was implemented and demonstrated in an EU project. But it has not been enhanced to VCDMv2. This brings up an important point about the extensions directory. Given that it was started before VCDMv2 was finalised, then it quite likely that a number of the extensions will be VCDMv1 conformant rather than VCDMv2 conformant. So should there be two different extension directories, or should all extensions that are not VCDMv2 conformant be deleted?

Notwithstanding discussions about the extensions directory, this should not have any bearing on the removal of the ToU property from the VCDMv2 specification, since the ToU property uses examples from the EBSI implementation and not from the extensions directory.

brentzundel:
The working group agreed to remove and reserve extension points that don't
yet have two independent implemented specifications.

@David-Chadwick are you asserting that there are two different
specifications that use termsOfUse and that both specifications have been
implemented?

David-Chadwick:
I am asserting that there is one specification of ToU by EBSI and multiple independent implementations of it

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-06-12-vcwg#section2-1) on 2024-06-12

- no resolutions were taken

#### 2.1. Removes the `termsOfUse` section from the specification and reserves the property. (pr vc-data-model#1498)

_See github pull request [vc-data-model#1498](https://github.com/w3c/vc-data-model/pull/1498)._




**Brent Zundel:** looking at PR 1498. Manu raised this to help clean up the "this might be removed" type of extension points. if we can show there are specs/impls using this extension we can keep it in the spec. if we cannot claim others are using them, we will keep them as reserved terms but they won't formally be a part of the spec.  
… all of those PRs have approvals except 1498. David would like to keep it in the spec. please state your position.  

**David Chadwick:** there is terms of use in the VC and terms of use in the VP. in the latter case don't believe anyone has documented it--happy to remove. for issuing, several groups have implemented it, like EBSI. they have multiple independent implementations of it.  
… the v1 termsOfUse had 'dummy' VCs/examples. I spent time researching what EBSI had done, and documented that. Note I am not a member of EBSI. I believe it should stay, since there are at least 2+ implementers of the feature. for VPs, I believe it can be removed.  

**Brent Zundel:** David makes a good case - there are a couple implementations. what do folks think?  

**Benjamin Young:** when we mention that it is in use, do those features need to be analyzed in test suites, or can we point to them in the wild, in docs, etc?  

**Brent Zundel:** I have my own opinion on this but would love to hear from folks in the group. my take is the WG said we would demonstrate multiple implementations of each feature, and do that by the test suite. so yes it would be useful to have tests for this.  

**Ivan Herman:** for the W3C requirement. we have had cases for terms in a vocab..if there are no actionable semantics on the term - just a vocab requirement - there should be 2 implementations. I don't know whether the usage has additional semantics around it.  

**Ted Thibodeau Jr.:** usually evidence of implementation comes from the implementer. not usually a 3rd party reportage of that usage. part of the reason for that is unless the implementer submits it, there's no guarantee it will survive the next week, let alone the spec. I question whether it is the right thing to allow 3rd party usage. they could submit it.  
… unless David is participating with another group, I do not think it carries the same weight.  

**David Chadwick:** I believe every impl that is using the test suite will conform. If you don't recognize the identifier 'termsOfUse' you can ignore it. It is a good test ... more preferable that the EBSI folks come and say this themselves.  
… in the original PR, someone pointed me to the EBSI people. there have been folks here more familiar than I was. only because I was editor of the PR that I went and documented it.  

**Benjamin Young:** I linked to one of the tests we have now. there is a test that only Digital Bazaar is passing now. question around testing rigor. if just vocab terms, then these can be there. the test will be the same. even if the extension is in the future. looking for feedback from the group - is what we're testing now sufficient? would it change depending on how termsOfUse is defined?  

**Brent Zundel:** in the past there have been concerns raised about the end-to-end interoperable nature of the VCDM. with all extension points, including, at the time, the proof extension point. v1 eeked by in spite of folks voicing those concerns. going into this version those concerns were re-raised.  
… led to the intention we've been following all along. for these extension points to stick around there need to be actual implementations. testing it simply as a vocabulary item is not sufficient - my understanding and the group's.  
… for this particular item, the path we're on, if folks that have implemented this come to the group and show their implementations, then the group would be OK keeping the feature. seeing it exists and pointing to it is not sufficient.  

**Benjamin Young:** want to clarify, the automated tests focus on the 'must' language in the spec. since we're testing json docs and not human brains, we are just saying 'the thing doesn't choke' - if there is more wanted EBSI/whomever would need to come and demonstrate usage.  

**Brent Zundel:** here's how I expect the conversation to go: we have a test suite, which tests acceptance of the data model. for extension points, which in the past were tested this way, we see the data model and its limitations, but where are the impls? the specs we can follow? can it truly be end-to-end interoperable.  
… we can say it has vc-json-schema, multiple ways to secure the spec, etc. we can point to specs in addition to the report generated by the test suite.  
… last time we had the test suite but folks came and said 'yeah well but!' and we need to surpass that bar this time.  
… so what do we do about termsOfUse? do we remove it? or do we keep it in the spec and hope that the folks from EBSI come and demonstrate its usage.  

**David Chadwick:** I would like to suggest another way: we leave terms of use for issuance, remove for holder/presentation. we try to make contact with EBSI to make their case.  

> *Dmitri Zagidulin:* just to be clear - will termsOfUse still remain in the `@context`?

**Brent Zundel:** how do folks feel about that?  

**Gabe Cohen:** we are planning on implementing at TBD. how can I show that?  

**Brent Zundel:** are you planning on implementing this as EBSI are?  

**Gabe Cohen:** I will take a closer look at the PR.  

**Brent Zundel:** let's continue with David's suggestion. with that we have 1 more PR to look at quickly, then issues.  

</details>

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-06-19-vcwg#section2-2) on 2024-06-19

- no resolutions were taken

#### 2.2. Removes the `termsOfUse` section from the specification and reserves the property. (issue vc-data-model#1498)

_See github issue [vc-data-model#1498](https://github.com/w3c/vc-data-model/issues/1498)._




**Brent Zundel:** Next is 1498.  
… PRs are waiting to be merged. If the new charter gets approved, then we can take care of all of this. The only one we should have a conversation about is 1498.  

**David Chadwick:** This week, I have managed to contact EBSI, they will send someone to next week's meeting. They are using the terms of use. Also contacted Lal Chandran, and they confirmed it was being used as well.  

**Brent Zundel:** Reminder to the group. After our last conversation, if we get response from EBSI that they have implemented the spec, then we will be in a good place for anyone to say "Terms of use" is bad and should go away.  

**David Chadwick:** Not sure if they have agreed to participate in the test suite.  
… They have implemented VCDM v 1. They may not have yet upgraded so impractical for them to participate in test suite.  
… Question for Manu: To the best of my knowledge, nobody is using VPs sent by the holder, so OK to remove that part of the specification. Example in there is spurious.  

**Manu Sporny:** David, why don't you do the removal.  

**David Chadwick:** If I do an edit that removes the VP stuff, then we can accept one of Manu's PRs or mine, and reject the other.  

**Manu Sporny:** Doesn't think the fact that they're using the v1 context matters much, all that matters is that they have written software around it. We will use one of their examples. May run into a problem with their context.  
… We will issue a VC that includes the feature and if we get two implementations we will keep the feature.  
… Let's see how many implementations they have and if they're committed to using the feature we will keep it.  
… There are multiple implementations using v1 already integrated into the test suite so we can test the extension point mechanism.  

**David Chadwick:** All implementations should be able to handle all examples in the spec. Upgraded it from v1 to v2 by changing the context, every implementation should see the terms of use and ignore it if it doesn't recognize the type.  

</details>

msporny:
Per the call today, the Chair noted that we [no longer need to hold off on merging this PR](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-07-17-vcwg#section3-2) based on waiting for the Charter to go out for review. That said, we are still waiting to hear back from EBSI to report on their use of this feature.

KMORGAN008 (approved):

msporny:
EBSI has demonstrated that the `termsOfUse` feature is in use and will provide updated examples in PR #1533. 

This PR has been superseded by PR #1533. Closing.

Issue #1503, opened 2024-06-18
SD-JWT fields in the v2 context should use `"@type": "@json"`


dlongley:
For example, see: https://github.com/w3c/vc-data-model/blob/main/contexts/credentials/v2#L59-L67

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-06-19-vcwg#section2-5) on 2024-06-19

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 2.5. SD-JWT fields in the v2 context should use `"@type": "@json"` (issue vc-data-model#1503)

_See github issue [vc-data-model#1503](https://github.com/w3c/vc-data-model/issues/1503)._




**Brent Zundel:** Next is 1503. Issues like this are a good sign. Means people are finding little nits like this.  
… Dave Longley is not on the call. There need to be a couple of modifications to the context because of this issue.  
… Manu assigned 1502 and 1503.  

</details>

msporny:
PR #1505 has been raised to address this issue. This issue will be closed once PR #1505 has been merged.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-06-26-vcwg#section2-4) on 2024-06-26

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 2.4. SD-JWT fields in the v2 context should use `"@type": "@json"` (issue vc-data-model#1503)

_See github issue [vc-data-model#1503](https://github.com/w3c/vc-data-model/issues/1503)._




**Brent Zundel:** if I remember correctly this issue should be addressed.  
… PR#1505 addresses this.  

_See github pull request [vc-data-model#1505](https://github.com/w3c/vc-data-model/pull/1505)._




</details>

msporny:
PR #1505 has been merged, closing.

PR #1505, opened 2024-06-19
Added `@json` type to `_sd` property in v2 context.

msporny:
This PR is an attempt to address issue #1503 by ensuring that the `_sd` property's JSON values remain as JSON.

I did check the `_sd_alg` and `...` properties and both are always string values (so didn't use the `@json` type for those).

### DIFF
--- a/contexts/credentials/v2
+++ b/contexts/credentials/v2
@@ -60,7 +60,8 @@
       "@id": "https://www.iana.org/assignments/jwt#_sd_alg"
     },
     "_sd": {
-      "@id": "https://www.iana.org/assignments/jwt#_sd"
+      "@id": "https://www.iana.org/assignments/jwt#_sd",
+      "@type": "@json"
     },
     "...": {
       "@id": "https://www.iana.org/assignments/jwt#..."


dlongley (approved):
+1, but... I would recommend that the `_sd_alg` and `...` values still be marked as `"@type": "@json"` because it's not clear to me that they won't change to support multiple algorithms and/or digests, necessitating the need for ordered arrays of values in the future (that spec still isn't finished).

bc-pi:
> +1, but... I would recommend that the `_sd_alg` and `...` values still be marked as `"@type": "@json"` because it's not clear to me that they won't change to support multiple algorithms and/or digests, necessitating the need for ordered arrays of values in the future (that spec still isn't finished).

true it's not finished but the `_sd_alg` and `...` properties will almost certainly not change from being string values

bc-pi (approved):

msporny:
@bc-pi wrote:
> true it's not finished but the `_sd_alg` and `...` properties will almost certainly not change from being string values

Thanks for the confirmation, @bc-pi -- that was very helpful to give us confidence that we're typing the data correctly. 

@dlongley based on that, I'm leaning towards keeping those two properties as untyped (strings). Any concerns with doing that. I do admit that using `@json` would give us 100% certainty if these values were to change for whatever reason before SD-JWT was standardized.

dlongley:
@msporny, given the above, I'm ok with leaving them how they are.

decentralgabe (approved):

OR13 (approved):

brentzundel (approved):

David-Chadwick (approved):

longpd (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-06-26-vcwg#section2-4) on 2024-06-26

- no resolutions were taken

#### 2.4. SD-JWT fields in the v2 context should use `"@type": "@json"` (issue vc-data-model#1503)

_See github issue [vc-data-model#1503](https://github.com/w3c/vc-data-model/issues/1503)._




**Brent Zundel:** if I remember correctly this issue should be addressed.  
… PR#1505 addresses this.  

_See github pull request [vc-data-model#1505](https://github.com/w3c/vc-data-model/pull/1505)._




</details>

msporny:

Issue #1512, opened 2024-06-28
Imprecise definition of what should be secured in a VP


iherman:
In the section ["Securing Mechanisms Specifications"](https://www.w3.org/TR/vc-data-model-2.0/#securing-mechanism-specifications) the text says (defining what exactly is secured in the case of a VP):

> In a [verifiable presentation](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verifiable-presentation), the property MUST secure the [default graph](https://www.w3.org/TR/vc-data-model-2.0/#dfn-default-graph) of the [presentation](https://www.w3.org/TR/vc-data-model-2.0/#dfn-presentation) as well as every [proof graph](https://www.w3.org/TR/vc-data-model-2.0/#dfn-proof-graph) related to each [verifiable credential](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verifiable-credential) in the [presentation](https://www.w3.org/TR/vc-data-model-2.0/#dfn-presentation).

But this is not entirely precise. In my view, it should say:

> In a [verifiable presentation](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verifiable-presentation), the property MUST secure the [default graph](https://www.w3.org/TR/vc-data-model-2.0/#dfn-default-graph) of the [presentation](https://www.w3.org/TR/vc-data-model-2.0/#dfn-presentation) as well as each [verifiable credential](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verifiable-credential) with the related [proof graph](https://www.w3.org/TR/vc-data-model-2.0/#dfn-proof-graph) in the [presentation](https://www.w3.org/TR/vc-data-model-2.0/#dfn-presentation).

Why? In the case of a VP, each credential is a separate named graph, which is then secured by a proof graph (see [Figure 9](https://www.w3.org/TR/vc-data-model-2.0/#info-graph-vp)). The original text may be read as if only the proof graphs should be secured, forgetting about the credentials themselves.

(I am happy to make a PR of course, if there is agreement.)

cc @msporny @dlongley

msporny:
@iherman I'm fine with your language, though we might ALSO want to point to @dlongley's new text in DI on the matter, which is more generalized:

https://w3c.github.io/vc-data-integrity/#proof-graphs

We can fine tune the language once you've been able to raise a PR, please raise a PR.

iherman:
[...]

PR #1515, opened 2024-07-01
Change on what exactly must be secured in a presentation

iherman:
This is to answer to the issue raised in #1512

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1515.html" title="Last updated on Jul 2, 2024, 9:25 AM UTC (501eb24)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1515/5dd94d3...501eb24.html" title="Last updated on Jul 2, 2024, 9:25 AM UTC (501eb24)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4266,10 +4266,9 @@ In a [=verifiable credential=], the property MUST secure the
 [=default graph=].
           </li>
           <li>
-In a [=verifiable presentation=], the property MUST secure the 
-[=default graph=] of the [=presentation=], and each 
-[=verifiable credential=], as well as its related [=proof graph=], in the 
-[=presentation=].
+In a [=verifiable presentation=], the property MUST secure all
+graphs in the [=presentation=] not referred to by the property itself,
+including each [=verifiable credential=] and any of its related [=proof graphs=].
           </li>
           <li>
 The `proof` property as defined in [[VC-DATA-INTEGRITY]] MAY be used by the
--- a/index.html
+++ b/index.html
@@ -4266,9 +4266,10 @@ In a [=verifiable credential=], the property MUST secure the
 [=default graph=].
           </li>
           <li>
-In a [=verifiable presentation=], the property MUST secure the
-[=default graph=] of the [=presentation=] as well as every [=proof graph=]
-related to each [=verifiable credential=] in the [=presentation=].
+In a [=verifiable presentation=], the property MUST secure the 
+[=default graph=] of the [=presentation=], and each 
+[=verifiable credential=], as well as its related [=proof graph=], in the 
+[=presentation=].
           </li>
           <li>
 The `proof` property as defined in [[VC-DATA-INTEGRITY]] MAY be used by the


index.html
```diff
@@ -4266,9 +4266,10 @@ <h3>Securing Mechanism Specifications</h3>
 [=default graph=].
```
dlongley, line 4266: I can't make a suggestion here, but I believe this should not be a statement about a "property" but about a securing mechanism, and it should be simplified to "The securing mechanism MUST secure all graphs in a [=verifiable credential=]." There's a minor technicality in that, a proof value itself (in a proof graph itself) cannot be "self-protected", but maybe that's obvious.

Similarly, the above "The property MUST define" seems more like it should say "The securing mechanism MUST define". This list should perhaps be about the requirements for the property *and* the new securing mechanism.

index.html
```diff
@@ -4266,9 +4266,10 @@ <h3>Securing Mechanism Specifications</h3>
...
           </li>
           <li>
-In a [=verifiable presentation=], the property MUST secure the
-[=default graph=] of the [=presentation=] as well as every [=proof graph=]
-related to each [=verifiable credential=] in the [=presentation=].
+In a [=verifiable presentation=], the property MUST secure the 
+[=default graph=] of the [=presentation=], and each 
+[=verifiable credential=], as well as its related [=proof graph=], in the 
+[=presentation=].
```
dlongley, line 4272: Hmm, something more like this?

```suggestion
In a [=verifiable presentation=], the property MUST secure all
graphs in the [=presentation=], including each 
[=verifiable credential=] and any of its related [=proof graphs=].
```

decentralgabe (approved):

index.html
TallTed: I wonder whether —
```
Securing mechanism specifications that create new types of [=embedded proofs=] 
MUST specify a [=property=] for securing
```

— should become the following (which implies further changes to the bullet list, and possibly more) — 
```
Securing mechanism specifications that create new types of [=embedded proofs=]
MUST specify an object for securing both [=verifiable credentials=] and
[=verifiable presentations=].
```
— (should `object` be `[=object=]`?) with appropriate bullet adjustments, such as — 
```
The object MUST provide values for all [=properties] used by the [=proof graph=].
```

index.html
TallTed: Along the lines of my above, including @dlongley's suggestion --
```suggestion
In a [=verifiable presentation=], the object MUST secure all graphs in
the [=presentation=] not referred to by the object itself, including
each [=verifiable credential=] and any of its related [=proof graphs=].
```

longpd (approved):
With dlongley's amendments made by TallTed I support PR #1515

index.html
dlongley: @TallTed, I think this should probably say: "the object MUST secure all graphs in
the [=presentation=] not referred to by the property,", right?

index.html
iherman: I not against making this change, but that is more complex change than what this PR proposes to do. I would prefer to finalize this PR first and, if needed, do that in a separate PR.

index.html
iherman: See my comment above. I'd prefer to close this PR and look at the overall change in a separate PR.

iherman:
I have merged the proposal of @dlongley but I am not absolutely sure this is 1005 correct. Let us take the hypothetical situation as follows.

- I use a securing mechanism using `proof` from DI with one of the cryptosuites.
- I ***also*** use, in the same VP (or VC) a securing mechanism that defines the property `signature`.

Question: does `proof` have to also secure the proof graph defined by `signature`? In my original text, which explicitly lists what has to be secured, the answer is 'no'. In the text proposed by @dlongley the answer is, actually, 'yes'? I believe the intention is 'no', because the goal is to secure the core information in the VC/VP, but that requires an explicit listing of _what_ is secured...

index.html
iherman: B.t.w., @TallTed, the (RDF) "object" term is not used in the core VCDM terminology, only "value".

msporny (approved):

iherman:
We should close this PR without merge, and concentrate on teh more comprehensive #1519 instead.

iherman:
The discussion has clearly moved to #1519, closing this PR without further action, to avoid misunderstandings.

Issue #1517, opened 2024-07-01
Pin down the input type of verification algorithms


jyasskin:
https://w3c.github.io/vc-data-model/#securing-mechanism-specifications says

> A securing mechanism's verification algorithm MUST provide an interface that receives inputs of a media type ([string](https://infra.spec.whatwg.org/#string) inputMediaType) paired with either a sequence of bytes ([byte sequence](https://infra.spec.whatwg.org/#byte-sequence) inputBytes) or a document ([map](https://infra.spec.whatwg.org/#ordered-map) inputDocument)

Does this require that the securing mechanisms accept whichever of byte-sequence or map their caller wants to provide, or does it require the caller to send whichever the securing mechanism prefers? If the latter, how does the caller figure out which one the mechanism prefers, since that's not included in the [associated registry](https://w3c.github.io/vc-specs-dir/#securing-mechanisms)?

msporny:
It's the latter, and the presumption here is that human readable documentation for the cryptosuite library clearly defines what input types it supports. I could add a comment to that effect (and try to figure out how to align the input options so they're used in the algorithms), but at this point, I'm questioning if we should just say that it's up to the library author to pick one or both and understanding what the library supports is implementation specific.

jyasskin:
If you handle #1518 by parsing to an Infra map early, then I think it's easiest to have the securing specs take that map. Otherwise, you could have the call sites in this spec include something like "if the securing mechanism expects bytes, pass _theBytes_, otherwise pass the result of [=parsing JSON bytes to an Infra value=] given _theBytes_".

msporny:
> If you handle #1518 by parsing to an Infra map early, then I think it's easiest to have the securing specs take that map.

I don't think we can do this since we have some securing mechanisms (JOSE and COSE) that take opaque bytes, which are not JSON bytes, as input.

> Otherwise, you could have the call sites in this spec include something like "if the securing mechanism expects bytes, pass theBytes, otherwise pass the result of [=parsing JSON bytes to an Infra value=] given theBytes".

Yes, good, this is what I was hoping to do. Thank you for reaffirming that doing so will work for you. PR incoming.

msporny:
PR #1522 has been raised to address this issue. This issue will be closed once PR #1522 has been merged.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-07-17-vcwg#section3-4) on 2024-07-17

[...]

PR #1519, opened 2024-07-03
Reformulating the definition of securing mechanism

iherman:
This is an alternative to #1515, along the lines of @dlongley's comment in https://github.com/w3c/vc-data-model/pull/1515#pullrequestreview-2151735713, but also my comment in https://github.com/w3c/vc-data-model/pull/1515#issuecomment-2202687385. 

I would propose to close #1515 without merge and concentrate on this PR instead. Merging this would also take care of #1512.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1519.html" title="Last updated on Jul 9, 2024, 1:53 PM UTC (1eaa93e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1519/8e9ae4e...1eaa93e.html" title="Last updated on Jul 9, 2024, 1:53 PM UTC (1eaa93e)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4248,7 +4248,7 @@ can achieve this protection are discussed in Section
 Securing mechanism specifications that create new types of [=embedded proofs=]
 MUST specify a [=property=] that relates the [=verifiable credential=] or [=verifiable
 presentation=] to a [=proof graph=]. 
-The requirement on the securing mechanism are as follows.
+The requirements on the securing mechanism are as follow:
         </p>
         <ul>
           <li>
@@ -4265,7 +4265,7 @@ or the [=verifiable presentation=] itself.
         </ul>
 
       <p class="note">
-The last requirement means that the securing mechanism must secure the [=default graph=] and, 
+The last requirement means that the securing mechanism secures the [=default graph=] and, 
 for [=verifiable presentations=], each [=verifiable credential=] of the presentation, together with 
 their respective [=proof graphs=].
 Also, if the verifiable presentation uses several proofs, the respective proof graphs of these proofs
--- a/index.html
+++ b/index.html
@@ -4266,10 +4266,9 @@ In a [=verifiable credential=], the property MUST secure the
 [=default graph=].
           </li>
           <li>
-In a [=verifiable presentation=], the property MUST secure the 
-[=default graph=] of the [=presentation=], and each 
-[=verifiable credential=], as well as its related [=proof graph=], in the 
-[=presentation=].
+In a [=verifiable presentation=], the property MUST secure all
+graphs in the [=presentation=] not referred to by the property itself,
+including each [=verifiable credential=] and any of its related [=proof graphs=].
           </li>
           <li>
 The `proof` property as defined in [[VC-DATA-INTEGRITY]] MAY be used by the
--- a/index.html
+++ b/index.html
@@ -4266,9 +4266,10 @@ In a [=verifiable credential=], the property MUST secure the
 [=default graph=].
           </li>
           <li>
-In a [=verifiable presentation=], the property MUST secure the
-[=default graph=] of the [=presentation=] as well as every [=proof graph=]
-related to each [=verifiable credential=] in the [=presentation=].
+In a [=verifiable presentation=], the property MUST secure the 
+[=default graph=] of the [=presentation=], and each 
+[=verifiable credential=], as well as its related [=proof graph=], in the 
+[=presentation=].
           </li>
           <li>
 The `proof` property as defined in [[VC-DATA-INTEGRITY]] MAY be used by the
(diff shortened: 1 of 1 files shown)


dlongley (approved):

brentzundel (approved):

index.html
```diff
@@ -4246,37 +4246,39 @@ <h3>Securing Mechanism Specifications</h3>
...
+The securing mechanism MUST secure all graphs in the [=verifiable credential=] or the [=verifiable
+presentation=], except for any [=proof graphs=] securing the [=verifiable credential=] 
+or the [=verifiable presentation=] itself.
           </li>
-          <li>
+
+        </ul>
+
+      <p class="note">
+The last requirement means that the securing mechanism must secure the [=default graph=] and,
```
msporny, line 4268:
```suggestion
The last requirement means that the securing mechanism secures the [=default graph=] and, 
```

index.html
```diff
@@ -4246,37 +4246,39 @@ <h3>Securing Mechanism Specifications</h3>
...
-          <li>
+
+        </ul>
+
+      <p class="note">
+The last requirement means that the securing mechanism must secure the [=default graph=] and, 
+for [=verifiable presentations=], each [=verifiable credential=] of the presentation, together with 
+their respective [=proof graphs=].
+Also, if the verifiable presentation uses several proofs, the respective proof graphs of these proofs
+are <em>not</em> mutually secured.
```
msporny, line 4272: I don't think this is true if chained proofs are used, where a 2nd proof has a dependency on the first proof in the verifiable presentation. 

```suggestion
```

msporny (changes requested):
LGTM, except for the comment about multiple proofs on a presentation (and the editorial nit).

index.html
```diff
@@ -4246,37 +4246,39 @@ <h3>Securing Mechanism Specifications</h3>
 
         <p>
 Securing mechanism specifications that create new types of [=embedded proofs=]
-MUST specify a [=property=] for securing both [=verifiable credentials=] and
-[=verifiable presentations=]. The requirements for the property used by the
-embedded securing mechanism are as follows:
+MUST specify a [=property=] that relates the [=verifiable credential=] or [=verifiable
+presentation=] to a [=proof graph=]. 
+The requirement on the securing mechanism are as follows.
```
TallTed, line 4251:
```suggestion
The requirements on the securing mechanism are as follow:
```

TallTed (changes requested):
Small grammar tweaks, in addition to the prior suggestions.

index.html
iherman: Well... we may have a problem that goes beyond this PR. 

1. looking at [Create proof section](https://www.w3.org/TR/vc-di-eddsa/#create-proof-eddsa-rdfc-2022) in the EdDSA spec, point (4) says that you take the hash of _transformedData_ and _proofConfig_. 
2. _proofConfig_  refers to the properties of the current proof in the making. We can put that aside for the moment. 
3. _transformedData_ refers back to the DI spec via step (3), through the concept of _unsecuredDocument_ .
4. _unsecuredDocument_ is defined in [§ 4 Algorithms](https://www.w3.org/TR/vc-data-integrity/#dfn-unsecured-data-document) as follows:

> An unsecured data document is a [map](https://infra.spec.whatwg.org/#ordered-map) with no proof [key](https://infra.spec.whatwg.org/#map-key).

This means two things:

1. I claim (but you may prove me wrong) that my interpretation is correct, including the case when proofs are chained: the mechanism only secures the original credential data, not the previous proof.
2. We do have a problem with the definition of unsecured data document, because that **_removes the proofs for  the embedded credential in the case of verifiable presentations_**. This is not what we want.

Maybe we do this in a wrong order. The spec I propose here essentially **_defines_** the concept of unsecured data, and maybe this should be the official definition and in the aforementioned section. And, if we think that the chained proofs _MUST_ be included in the unsecured data, then we have to include it in this definition explicitly (which would then be automatically valid through the reference chain).

B.t.w., do we have tests for the case of chained proofs? (cc @BigBlueHat)

iherman: Looking at it a bit further, mathematically speaking my second statement on presentations may not be correct: in a VP, _on the top level_ there is no `proof` key in the original data, because the `proof` appear only "behind" the `verifiableCredential` key, so to say. But the definition of unsecured document is still easy to misunderstand; it does not say that it should exclude `proof` on the top level only...

msporny: @iherman yes, the "unsecure document" language doesn't work for documents that are getting a second chained signature on them. We'll need to fix that.

TallTed: > [@msporny] @iherman ... We'll need to fix that.

New issue to track?

iherman: I have

- Raised a separate issue (#1527)
- Made a commit removing that remark on multiple proofs.

I believe that cleanest would be to merge this PR now, and then work on solving #1517 which may force us to come back to this section, too. Leaving this PR open may lead to further complications...

msporny (approved):

msporny:

PR #1522, opened 2024-07-06
Clarify input type of verification algorithms.

msporny:
This PR is an attempt to address issue #1517 by clarifying the language around input types to verification algorithms.

/cc @jyasskin @PatStLouis @filip @silverpill @flechtner

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1522.html" title="Last updated on Jul 21, 2024, 7:55 PM UTC (c90b313)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1522/ccbe460...c90b313.html" title="Last updated on Jul 21, 2024, 7:55 PM UTC (c90b313)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -4122,16 +4122,16 @@ algorithms MAY be general in nature and MAY be used to secure data other than
 
         <p>
 Securing mechanism specifications MUST provide a verification algorithm that
-returns only the information in the [=conforming document=] that has been
-secured, without any securing mechanism information included, such as `proof` or
+returns the information in the [=conforming document=] that has been
+secured in isolation, without any securing mechanism information included, such as `proof` or
 JOSE/COSE header parameters and signatures. Verification algorithms MAY return
 additional information that might be helpful (for example, during validation or
 for debugging purposes), such as securing mechanism details. A verification
 algorithm MUST provide an interface that receives a media type ([=string=]
-|inputMediaType|) and secured data ([=byte sequence=] or [=map=] |securedData|).
-If the verification algorithm expects a [=byte sequence=], provide |securedData|
+|inputMediaType|) and input data ([=byte sequence=] or [=map=] |inputData|).
+If the verification algorithm expects a [=byte sequence=], provide |inputData|
 to the algorithm. If the verification algorithm expects a [=map=], provide the
-result of [=parsing JSON bytes to an Infra value=] given |securedData|. Securing
+result of [=parsing JSON bytes to an Infra value=] given |inputData|. Securing
 mechanism specifications MAY provide additional algorithms and interfaces to the
 ones specified in this document. The verification algorithm returns a
 verification result with at least the following [=struct/items=]:
--- a/index.html
+++ b/index.html
@@ -4129,12 +4129,9 @@ additional information that might be helpful (for example, during validation or
 for debugging purposes), such as details of the securing mechanism. A verification
 algorithm MUST provide an interface that receives a media type ([=string=]
 |inputMediaType|) and input data ([=byte sequence=] or [=map=] |inputData|).
-If the verification algorithm expects a [=byte sequence=], provide |inputData|
-to the algorithm. If the verification algorithm expects a [=map=], provide the
-result of [=parsing JSON bytes to an Infra value=] given |inputData|. Securing
-mechanism specifications MAY provide algorithms and interfaces in addition to the
-ones specified in this document. The verification algorithm returns a
-verification result with at least the following [=struct/items=]:
+Securing mechanism specifications MAY provide algorithms and interfaces in
+addition to the ones specified in this document. The verification algorithm
+returns a verification result with at least the following [=struct/items=]:
         </p>
 
         <dl>
@@ -4736,11 +4733,10 @@ versions of this specification.
         <h3>Verification</h3>
 
         <p>
-This section contains an algorithm that [=conforming verifier
-implementations=] MUST run when verifying a [=verifiable credential=] or a
-[=verifiable presentation=]. This algorithm takes a media
-type ([=string=] |inputMediaType|) and secured data
-([=byte sequence=] or [=map=] |inputData|) and returns a [=map=] that
+This section contains an algorithm that [=conforming verifier implementations=]
+MUST run when verifying a [=verifiable credential=] or a [=verifiable
+presentation=]. This algorithm takes a media type ([=string=] |inputMediaType|)
+and secured data ([=byte sequence=] |inputData|) and returns a [=map=] that
 contains the following:
         </p>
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -4197,16 +4197,20 @@ <h3>Securing Mechanism Specifications</h3>
         </p>
 
         <p>
-Securing mechanism specifications MUST provide a verification mechanism that
+Securing mechanism specifications MUST provide a verification algorithm that
 returns only the information in the [=conforming document=] that has been
 secured, without any securing mechanism information included, such as `proof` or
```
dlongley, line 4202: I'm not sure if anyone else would get tripped up by this, but in reading this description it sounded to me like it would be a contradiction to return additional information below about the securing mechanism -- since this says "returns only... without any securing mechanism information included...". So a minor reshuffling here to indicate the secured information must be returned "in isolation" might help:

```suggestion
Securing mechanism specifications MUST provide a verification algorithm that
returns the information in the [=conforming document=] that has been
secured in isolation, without any securing mechanism information included, such as `proof` or
```

index.html
```diff
@@ -4197,16 +4197,20 @@ <h3>Securing Mechanism Specifications</h3>
...
-([=byte sequence=] |inputBytes|) or a document ([=map=] |inputDocument|), and
-returns a verification result with at least the following [=struct/items=]:
+JOSE/COSE header parameters and signatures. Verification algorithms MAY return
+additional information that might be helpful (for example, during validation or
+for debugging purposes), such as securing mechanism details. A verification
+algorithm MUST provide an interface that receives a media type ([=string=]
+|inputMediaType|) and secured data ([=byte sequence=] or [=map=] |securedData|).
+If the verification algorithm expects a [=byte sequence=], provide |securedData|
+to the algorithm. If the verification algorithm expects a [=map=], provide the
+result of [=parsing JSON bytes to an Infra value=] given |securedData|. Securing
```
dlongley, line 4210: Maybe "input data" instead of "securedData" since this algorithm makes the "securing determination"? I can live with it either way (or with some other name), just drawing attention here.

```suggestion
|inputMediaType|) and input data ([=byte sequence=] or [=map=] |inputData|).
If the verification algorithm expects a [=byte sequence=], provide |inputData|
to the algorithm. If the verification algorithm expects a [=map=], provide the
result of [=parsing JSON bytes to an Infra value=] given |inputData|. Securing
```

dlongley (approved):

longpd (approved):

longpd (approved):

decentralgabe (approved):

index.html
```diff
@@ -4197,16 +4197,20 @@ <h3>Securing Mechanism Specifications</h3>
...
-JOSE/COSE header parameters and signatures. Specifications MAY provide additional mechanisms to convey
-other information that might be helpful (for example, during validation or for
-debugging purposes), such as securing mechanism data. A securing mechanism's
-verification algorithm MUST provide an interface that receives inputs of a media
-type ([=string=] |inputMediaType|) paired with either a sequence of bytes
-([=byte sequence=] |inputBytes|) or a document ([=map=] |inputDocument|), and
-returns a verification result with at least the following [=struct/items=]:
+Securing mechanism specifications MUST provide a verification algorithm that
+returns the information in the [=conforming document=] that has been
+secured in isolation, without any securing mechanism information included, such as `proof` or
```
TallTed, line 4202:
```suggestion
returns the information in the [=conforming document=] that has been secured, in
isolation, without including any securing mechanism information, such as `proof` or
```

index.html
```diff
@@ -4197,16 +4197,20 @@ <h3>Securing Mechanism Specifications</h3>
...
+JOSE/COSE header parameters and signatures. Verification algorithms MAY return
+additional information that might be helpful (for example, during validation or
+for debugging purposes), such as securing mechanism details. A verification
```
TallTed, line 4205:
```suggestion
for debugging purposes), such as details of the securing mechanism. A verification
```

index.html
```diff
@@ -4197,16 +4197,20 @@ <h3>Securing Mechanism Specifications</h3>
...
+algorithm MUST provide an interface that receives a media type ([=string=]
+|inputMediaType|) and input data ([=byte sequence=] or [=map=] |inputData|).
+If the verification algorithm expects a [=byte sequence=], provide |inputData|
+to the algorithm. If the verification algorithm expects a [=map=], provide the
+result of [=parsing JSON bytes to an Infra value=] given |inputData|. Securing
+mechanism specifications MAY provide additional algorithms and interfaces to the
```
TallTed, line 4211:
```suggestion
mechanism specifications MAY provide algorithms and interfaces in addition to the
```

TallTed (changes requested):

PatStLouis:
@msporny in accordance with this we should remove the `returnCredential`/`returnPresentation` options from the verify [credential](https://w3c-ccg.github.io/vc-api/#verify-credential)/[presentation](https://w3c-ccg.github.io/vc-api/#verify-presentation) endpoints of the vc-api and consider it an implicit MUST all cases.

In the case of a presentation, should the `verfiableCredential` array be returned with credentials stripped of their proofs as well? Or do we want this to only apply to the presentation layer?

PatStLouis (approved):

msporny:
> @msporny in accordance with this we should remove the `returnCredential`/`returnPresentation` options from the verify [credential](https://w3c-ccg.github.io/vc-api/#verify-credential)/[presentation](https://w3c-ccg.github.io/vc-api/#verify-presentation) endpoints of the vc-api and consider it an implicit MUST all cases.

I'm not sure we should do that yet. There might be use cases where you just want to know that the signature is valid, but want to do "other processing" in your application for efficiency reasons. For example, if the VCs you're dealing with are 1MB or 2MB in size... and you're running a high volume site, maybe it's more efficient for you to just get a "yes/no" answer from the API and then strip the `proof` off yourself at the application layer. I realize the use case is a bit contrived, but I'm concerned that there are others like this.

> In the case of a presentation, should the `verfiableCredential` array be returned with credentials stripped of their proofs as well? Or do we want this to only apply to the presentation layer?

Yes, I believe so. There is a complexity around what to do with chained proofs vs. set proofs and the verification endpoint in VC API... but that's something we'll handle at that layer of the stack, which is higher up than this layer.

TallTed (approved):

index.html
```diff
@@ -4197,16 +4197,20 @@ <h3>Securing Mechanism Specifications</h3>
...
+returns the information in the [=conforming document=] that has been secured, in
+isolation, without including any securing mechanism information, such as `proof` or
+JOSE/COSE header parameters and signatures. Verification algorithms MAY return
+additional information that might be helpful (for example, during validation or
+for debugging purposes), such as details of the securing mechanism. A verification
+algorithm MUST provide an interface that receives a media type ([=string=]
+|inputMediaType|) and input data ([=byte sequence=] or [=map=] |inputData|).
+If the verification algorithm expects a [=byte sequence=], provide |inputData|
+to the algorithm. If the verification algorithm expects a [=map=], provide the
+result of [=parsing JSON bytes to an Infra value=] given |inputData|. Securing
```
jyasskin, line 4210: I think these instructions are out of place here, because there's no live |inputData| variable in this paragraph. I think this belongs in step 1.2 of https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1522.html#verification, where you are translating an |inputData| into the parameter type of this interface. Then there's one awkward case: if Verification's caller passed a map, but the securing mechanism expects a byte sequence, we don't have instructions for what to do.

jyasskin (changes requested):
Thanks for working on fixing this!

index.html
msporny: Fixed in c90b313fe9d516361088a3cd9b76bd274fafbf8f.

msporny:

PR #1524, opened 2024-07-06
Add warnings related to `@vocab` usage

msporny:
This PR is a partial attempt to address issue #1514 by warning about the usage of `@vocab`, including guidance on when it might be acceptable to use it.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1524.html" title="Last updated on Jul 15, 2024, 2:29 PM UTC (c4d3da3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1524/8e9ae4e...c4d3da3.html" title="Last updated on Jul 15, 2024, 2:29 PM UTC (c4d3da3)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1172,7 +1172,7 @@ document.
 
         <p>
 This specification is designed to ease the prototyping of new types of
-[=verifiable credentials=]. Developers can copy the template below and paste it
+[=verifiable credential=]. Developers can copy the template below and paste it
 into common [=verifiable credential=] tooling to start issuing, holding, and
 verifying prototype credentials.
         </p>
@@ -1183,7 +1183,7 @@ they would like to create. Since [=verifiable credentials=] talk about subjects,
 each property-value pair in the `credentialSubject` object expresses a
 particular property of the credential subject. Once a developer has added a
 number of these property-value combinations, the modified object can be sent to
-the [=verifiable credential=] issuer software and a [=verifiable credential=]
+the [=verifiable credential=] issuer software, and a [=verifiable credential=]
 will be created for the developer. From a prototyping standpoint, that is all a
 developer needs to do.
         </p>
@@ -1206,9 +1206,9 @@ developer needs to do.
 Once a developer has prototyped their credential to a point where they believe
 all of the credential properties are stable, it is advised that they generate
 vocabulary and context files for their application and publish them at stable
-URLs so that other developers can use the same vocabulary and context to achieve
+URLs, so that other developers can use the same vocabulary and context to achieve
 interoperability. The `https://www.w3.org/ns/credentials/examples/v2` URL above
-will then be replaced with the URL to an application-specific context. This
+would then be replaced with the URL of an application-specific context. This
 process is covered in Section <a href="#extensibility"></a>. Alternatively,
 developers can reuse existing vocabulary and context files that happen to fit
 their use case. They can explore the [[[VC-SPECS]]] [[VC-SPECS]] for reusable
@@ -4406,10 +4406,10 @@ might not perform generalized JSON-LD processing. Authors of [=conforming
 documents=] are advised that interoperability might be reduced if JSON-LD
 keywords in the `@context` value are used to globally affect values in a
 [=verifiable credential=] or [=verifiable presentation=], such as by
-setting the `@base` or `@vocab` keyword. For example, setting these values
-might trigger a failure in a mis-implemented JSON Schema check on the `@context`
+setting either or both of the `@base` or `@vocab` keywords. For example, setting these values
+might trigger a failure in a mis-implemented JSON Schema test of the `@context`
 value in an implementation that is performing [=type-specific credential
-processing=] and not expecting the `@base` value or `@vocab` value to be
+processing=] and not expecting the `@base` and/or `@vocab` value to be
 expressed in the `@context` value.
         </p>
 
@@ -4422,14 +4422,14 @@ urged to not use JSON-LD features that are not easily detected when performing
         <ul>
           <li>
 In-line declaration of JSON-LD keywords in the `@context` value that globally
-modify document term and value processing, such as setting `@base` or `@vocab`,
+modify document term and value processing, such as setting `@base` or `@vocab`
           </li>
           <li>
 Use of JSON-LD contexts that override declarations in previous contexts, such as
-resetting `@vocab`,
+resetting `@vocab`
           </li>
           <li>
-In-line declaration of JSON-LD contexts in the `@context` property, and
+In-line declaration of JSON-LD contexts in the `@context` property
           </li>
           <li>
 Use of full URLs for JSON-LD terms and types (e.g.,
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -4415,22 +4421,39 @@ <h3>JSON-LD</h3>
...
+legitimate usages of the feature, such as to ease experimentation, development,
+and localized deployment. If an application developer desires to use `@vocab` in
+production, which is strongly advised against, they are urged to understand that
+any usage of `@vocab` will disable undefined term error reporting, and
+subsequent usage will override previous `@vocab` declarations. Subsequent usage
+of `@vocab` changes the semantics of the information contained in the document
+and so it is important to understand how these changes will affect the
+application being developed. Application developers MUST understand every
+JSON-LD context used by their application to the extent that it affects
+the semantics of the terms that are used by their application.
```
TallTed, line 4455:
```suggestion
While this specification cautions against the use of `@vocab`, there are
legitimate uses of the feature, such as to ease experimentation, development,
and localized deployment. If an application developer wants to use `@vocab` in
production — even though strongly advised against — they are urged to understand that
any use of `@vocab` will disable reporting of "undefined term" errors, and
later use(s) will override any previous `@vocab` declaration(s). Different values
of `@vocab` can change the semantics of the information contained in the document,
so it is important to understand whether and how these changes will affect the
application being developed. Application developers MUST understand every
JSON-LD context used by their application, at least to the extent that it affects
the semantics of the terms that are used by their application.
```

TallTed (changes requested):

index.html
```diff
@@ -4415,22 +4421,39 @@ <h3>JSON-LD</h3>
...
+`https://www.w3.org/ns/credentials/v2`).
           </li>
-       </ul>
+        </ul>
+
+        <p>
+While this specification warns against the usage of `@vocab`, there are
+legitimate usages of the feature, such as to ease experimentation, development,
+and localized deployment. If an application developer desires to use `@vocab` in
+production, which is strongly advised against, they are urged to understand that
```
decentralgabe, line 4448:
```suggestion
production, which is advised against, they are urged to understand that
```

perhaps something like 

```suggestion
production, which is advised against to reduce term collisions and leverage the benefits of semantic interoperability, they are urged to understand that
```

msporny: When I pulled in @TallTed's change suggestions, it invalidated this one, can you please re-raise it on the updated text so you get credit for the change?

PatStLouis (approved):

index.html
```diff
@@ -1201,12 +1205,14 @@ <h3>Getting Started</h3>
...
-vocabulary and context files for their application and publish them at stable URLs
-so that other developers can use the same vocabulary and context to achieve
-interoperability. This process is covered in Section
-<a href="#extensibility"></a>. Alternatively, developers can reuse existing vocabulary
-and context files that happen to fit their use case. They can explore the
-[[[VC-SPECS]]] [[VC-SPECS]] for reusable resources.
+vocabulary and context files for their application and publish them at stable
+URLs, so that other developers can use the same vocabulary and context to achieve
+interoperability. The `https://www.w3.org/ns/credentials/examples/v2` URL above
+would then be replaced with the URL of an application-specific context. This
```
dlongley, line 1211: I can live with "application-specific" but I think "use-case-specific" is better:

```suggestion
would then be replaced with the URL of an use-case-specific context. This
```

index.html
```diff
@@ -4415,22 +4421,39 @@ <h3>JSON-LD</h3>
...
+legitimate uses of the feature, such as to ease experimentation, development,
+and localized deployment. If an application developer wants to use `@vocab` in
+production — even though strongly advised against — they are urged to understand that
+any use of `@vocab` will disable reporting of "undefined term" errors, and
+later use(s) will override any previous `@vocab` declaration(s). Different values
+of `@vocab` can change the semantics of the information contained in the document,
+so it is important to understand whether and how these changes will affect the
+application being developed. Application developers MUST understand every
+JSON-LD context used by their application, at least to the extent that it affects
+the semantics of the terms that are used by their application.
```
dlongley, line 4455:
```suggestion
the semantics of the terms that are used by their application. Applications MAY
use JSON-LD <a data-cite="JSON-LD11-API#compaction-algorithms">compaction
algorithms</a> to transform a document that uses an unknown JSON-LD context
to one that does not, so the new document's terms will match expectations.
```

dlongley (approved):

index.html
TallTed: Because, what would a change request be without a nitpick? `a` instead of `an`, here.
```suggestion
would then be replaced with the URL of a use-case-specific context. This
```

index.html
```diff
@@ -4415,22 +4421,42 @@ <h3>JSON-LD</h3>
...
+any use of `@vocab` will disable reporting of "undefined term" errors, and
+later use(s) will override any previous `@vocab` declaration(s). Different values
+of `@vocab` can change the semantics of the information contained in the document,
+so it is important to understand whether and how these changes will affect the
+application being developed. Application developers MUST understand every
+JSON-LD context used by their application, at least to the extent that it affects
+the semantics of the terms that are used by their application. Applications MAY
+use JSON-LD <a data-cite="JSON-LD11-API#compaction-algorithms">compaction
+algorithms</a> to transform a document that uses an unknown JSON-LD context
+to one that does not, so the new document's terms will match expectations.
```
TallTed, line 4458: Some additional detail might be needed, with the addition of that sentence. I think this, because I don't know (and I haven't been able to figure it out, after more than a few minutes reading the linked page and others. I had *thought* I had a good grasp of JSON-LD mechanics, but this makes me wonder!) how the suggested transformation would work. Maybe add some more-descriptive, psuedo-algorithmic language here, with fewer low-level details than the linked page?

longpd (approved):

longpd (approved):

index.html
msporny: I'm thinking that this statement is going to require it's own section. Both in the Data Integrity spec, and possibly in the VCDM spec. People seem to be requesting it over here (see item 3): https://github.com/w3c/vc-data-integrity/issues/272#issuecomment-2212258255

If we get consensus to do item 3 (above), then I expect we'll dedicate a section to that sentence. So, we can take it out for now, or leave it in and expand upon it later. I'm fine either way.

index.html
```diff
@@ -4415,22 +4421,42 @@ <h3>JSON-LD</h3>
...
           </li>
-       </ul>
+        </ul>
+
+        <p>
+While this specification cautions against the use of `@vocab`, there are
+legitimate uses of the feature, such as to ease experimentation, development,
+and localized deployment. If an application developer wants to use `@vocab` in
+production — even though strongly advised against — they are urged to understand that
+any use of `@vocab` will disable reporting of "undefined term" errors, and
```
decentralgabe, line 4449:
```suggestion
production, which is advised against to reduce term collisions and leverage the benefits
of semantic interoperability, they are urged to understand that any use of `@vocab` will
disable reporting of "undefined term" errors, and
```

tried to maintain line spacing but may need another reformat

index.html
TallTed: Either will work for me, for now, if a tracking issue is opened before/upon merge of this PR.

msporny: This is now being tracked in https://github.com/w3c/vc-data-model/issues/1529

msporny:
Editorial, multiple reviews, changes requested and made, @TallTed's concern tracked in a new issue, no objections, merging.

Issue #1529, opened 2024-07-15
Enhance Context Validation


msporny:
From https://github.com/w3c/vc-data-integrity/issues/272#issuecomment-2212258255, @decentralgabe wrote:
> Add normative language to the DI specification requiring implementers to rigorously validate `@context` values. This should include:
a) Checking against a list of trusted contexts
b) Verifying the integrity of context contents using cryptographic hashes
c) Rejecting or transforming documents with untrusted contexts before processing

From https://github.com/w3c/vc-data-model/pull/1524/files#r1669382600, @TallTed wrote:
>> Applications MAY use JSON-LD <a data-cite="JSON-LD11-API#compaction-algorithms">compaction algorithms</a> to transform a document that uses an unknown JSON-LD context to one that does not, so the new document's terms will match expectations.
>
> Some additional detail might be needed, with the addition of that sentence. I think this, because I don't know (and I haven't been able to figure it out, after more than a few minutes reading the linked page and others. I had thought I had a good grasp of JSON-LD mechanics, but this makes me wonder!) how the suggested transformation would work. Maybe add some more-descriptive, psuedo-algorithmic language here, with fewer low-level details than the linked page?

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2024-07-17-vcwg#section3-6) on 2024-07-17

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 3.6. Enhance Context Validation (issue vc-data-model#1529)

_See github issue [vc-data-model#1529](https://github.com/w3c/vc-data-model/issues/1529)._




**Brent Zundel:** enhanced context validation.  

**Manu Sporny:** Gave made a proposal that we add normative language to data integrity spec, but it might be good to put in VCDM.  
… some sort of normative language to say you are checking issuers.  
… Ted also suggested some detail on how you can use compaction algorithms to get rid of untrusted contexts.  
… This PR is creating new normative language to doing that.  
… We would be testing for that at the application layer, which we haven't done before.  
… But it seems that the group is willing to go there.  
… action is to raise a PR that does that.  

**Brent Zundel:** any concerns, speak up now.  

**Manu Sporny:** I do have a question to the group. The specs we are creating have architectural layers, e.g., the securing layer is lower on the stack and validation is higher.  
… I'm trying to figure out if it is worth using this language in the data integrity specification.  
… Note that data integrity can work on things without @context.  
[...]

PR #1535, opened 2024-07-21
Add section on Context Validation.

msporny:
This PR attempts to address issue #1529 by adding normative text requiring verifiers to implement context validation.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1535.html" title="Last updated on Jul 27, 2024, 7:25 PM UTC (71e3446)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1535/ccbe460...71e3446.html" title="Last updated on Jul 27, 2024, 7:25 PM UTC (71e3446)">Diff</a>


index.html
```diff
@@ -4820,6 +4820,113 @@ <h3>Verification</h3>
...
+          <li>
+Set |result|.|status| to `false`.
+          </li>
+          <li>
+For each |contextValue| expressed in every `@context` property, ensure that
+either:
+            <ol class="algorithm">
+              <li>
+The |contextValue| exists in |approvedContexts| and matches a cryptographic
+hash associated with the matching approved context, or
```
dlongley, line 4863:
```suggestion
The |contextValue| exists in |approvedContexts| and matches an exact value
or a cryptographic hash associated with the matching approved context, or
```

index.html
```diff
@@ -4820,6 +4820,113 @@ <h3>Verification</h3>
...
+              </li>
+              <li>
+optionally, for the context values that do not exist in |approvedContexts|,
+remove the |contextValue| from the `@context` property values. Set
+|result|.|document| to the result of running the
+<a data-cite="JSON-LD11-API#compaction-algorithm">
+JSON-LD Compaction Algorithm</a> with the modified |inputDocument| as input.
```
dlongley, line 4870: Hmm, this isn't quite right.

```suggestion
optionally, if any of the context values do not exist in |approvedContexts|,
clone the `@context` property and remove each context value that was
not in |approvedContext| from the clone. Set |result|.|document| to
the result of running the <a data-cite="JSON-LD11-API#compaction-algorithm">
JSON-LD Compaction Algorithm</a> with the |inputDocument| and
the cloned and modified context as input.
```

index.html
```diff
@@ -4820,6 +4820,113 @@ <h3>Verification</h3>
...
+Another approach a [=verifier=] could use would be to use a list of well known
+context URLs and their associated approved cryptographic hashes, without storing
+every context file locally, which would allow contexts to be safely loaded and
+cached from the network without compromising the security expectations of the
+[=verifier=]. This approach would also be effectively equivalent to the
+algorithm provided in this section.
+        </p>
+
+        <p>
+Yet another valid approach would be for the [=holder=] to
```
dlongley, line 4914: Listing multiple ways to perform validation is good, but it then seems that the above MUST on the specific algorithm provided there should be a MAY instead -- or some other text should make it clear that either the algorithm can be run or one of these other equivalent approaches can be taken.

index.html
```diff
@@ -4820,6 +4820,113 @@ <h3>Verification</h3>
 
       </section>
 
+      <section class="normative">
+        <h3>Context Validation</h3>
+
+        <p>
+This section contains an algorithm that [=verifiers=] MUST run after running the
+algorithm in Section [[[#verification]]], when validating a [=verifiable
+credential=] or a [=verifiable presentation=]. This algorithm takes a document
```
iherman, line 4829: See my comment in https://github.com/w3c/vc-data-integrity/pull/291#discussion_r1687883171. Any reason why this algorithm shouldn't be run before even touching verification?

dlongley: It seems to me that it could be up to an implementer whether it is run before or after verification. This an application-level check, and the best time to run it can be application / use-case dependent.

iherman: Than the text should say so! It currently has a numbered list, which suggests mandatory ordering...

TallTed:
There's a lot of repeated text, here and in https://github.com/w3c/vc-data-integrity/pull/291, which needs similar if not identical changes (which suggestions from myself and others do not currently reflect), possibly being adjusted to be *said* in one place, and *referenced* in the other.

index.html
msporny: Fixed in 8dbad0b15b3b4d2ad46f6ba381951b6d73ca8990.

index.html
msporny: Fixed in 8dbad0b15b3b4d2ad46f6ba381951b6d73ca8990.

msporny:
@TallTed wrote:
> There's a lot of repeated text, here and in [w3c/vc-data-integrity#291](https://github.com/w3c/vc-data-integrity/pull/291), which needs similar if not identical changes (which suggestions from myself and others do not currently reflect), possibly being adjusted to be _said_ in one place, and _referenced_ in the other.

Yep, I had considered these issues:

1. We duplicate this text in DI and VC JOSE COSE (because the VC JOSE COSE folks have historically not wanted to refer to DI and also because it would be weird for DI or VC JOSE COSE, which don't reference each other for anything, to cross-reference either way on this particular point.
2. We have DI and VC JOSE COSE point to the language in this PR, which would be a bit of a strange layering violation since they're both at a lower architectural level than VCDM (it would be weird to ref "up" to VCDM from DI).
3. DI stands on it's own... you don't have to use it with VCDM, but this guidance is generally applicable to any JSON-LD document secured using DI, so it should probably exist over there even if it didn't exist in VCDM.

So, the solution seems to be to put it in both places, even though most of the text is duplicated, OR, if we can get past the "VC JOSE COSE needs to be at the same level as DI", we could just ref DI from VCDM (at least). I don't think we will get to consensus on that approach, so duplicating the text in VCDM and DI is the best I think we can achieve.

That said, happy to not duplicate the text, though the only reasonable path I see for that is to put this in DI (or maybe the JSON-LD spec, which you've raised as a concern @TallTed... but that doesn't happen before we get the VCWG specs to REC, IMHO).

Anyone else see a better way through this?

longpd:
> There's a lot of repeated text, here and in [w3c/vc-data-integrity#291](https://github.com/w3c/vc-data-integrity/pull/291), which needs similar if not identical changes (which suggestions from myself and others do not currently reflect), possibly being adjusted to be _said_ in one place, and _referenced_ in the other.

So where @TallTed should the core changes be made, e.g., in (https://github.com/w3c/vc-data-integrity/pull/291) or (https://github.com/w3c/vc-data-model/pull/1535) with the pointer from the other?

longpd:
> So, the solution seems to be to put it in both places, even though most of the text is duplicated, OR, if we can get past the "VC JOSE COSE needs to be at the same level as DI", we could just ref DI from VCDM (at least). I don't think we will get to consensus on that approach, so duplicating the text in VCDM and DI is the best I think we can achieve.

That looks like the most pragmatic thing to do, that is, see if we can get consensus on referencing DI from VCDM, and if that's not possible then duplicating the text in VCDM and DI makes sense. We'll just have to insure (as best we can) that any changes to text VCDM  are propagated to DI (and vice-versa) to they stay in sync.

msporny:
@longpd wrote:
> We'll just have to insure (as best we can) that any changes to text VCDM are propagated to DI (and vice-versa) to they stay in sync.

Update... the text across both PRs are now wildly out of sync (which only underscores @TallTed's concern)! Let's try to focus getting the language right on the DI PR and I'll update the VCDM PR (this one) once the DI text settles. It's largely the algorithm that changed (and is improved in the DI PR).

iherman (approved):

msporny:
@longpd wrote:
> See if we can get consensus on referencing DI from VCDM, and if that's not possible then duplicating the text in VCDM and DI makes sense. 

Alright, that is what this PR now does. The spec text asserting that context validation MUST happen was already in the spec. This PR now just points to the DI spec, whose Context Validation section is quite thorough now with an explanation and a normative algorithm, to say how to do it. This doesn't change the normative reference to DI (we already had that). IOW, this is now an editorial PR and I'm merging it as such.

If folks object to that, we can duplicate the text in the DI spec into the VCDM spec.

msporny:

PR #1566, opened 2024-09-24
Update "refreshService" example to VCDM2.0

peacekeeper:
This example currently seems to be based on VCDM 1.1.

***
<a href="https://pr-preview.s3.amazonaws.com/peacekeeper/vc-data-model/pull/1566.html" title="Last updated on Sep 24, 2024, 3:06 PM UTC (7d241fa)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1566/fde26dd...peacekeeper:7d241fa.html" title="Last updated on Sep 24, 2024, 3:06 PM UTC (7d241fa)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -3280,14 +3280,13 @@ definition.
           title="Usage of the refreshService property by an issuer">
 {
   "@context": [
-    "https://www.w3.org/2018/credentials/v1",
-    "https://w3id.org/age/v1",
-    "https://w3id.org/security/suites/ed25519-2020/v1"
+    "https://www.w3.org/ns/credentials/v2",
+    "https://w3id.org/age/v1"
   ],
   "type": ["VerifiableCredential", "AgeVerificationCredential"],
   "issuer": "did:key:z6MksFxi8wnHkNq4zgEskSZF45SuWQ4HndWSAVYRRGe9qDks",
-  "issuanceDate": "2024-04-03T00:00:00.000Z",
-  "expirationDate": "2024-12-15T00:00:00.000Z",
+  "validFrom": "2024-04-03T00:00:00.000Z",
+  "validUntil": "2024-12-15T00:00:00.000Z",
   "name": "Age Verification Credential",
   "credentialSubject": {
     "overAge": 21


msporny (changes requested):
While it is true that the example is a v1 example, the group also decided to use real world / representative examples. The credential refresh example (which is what the TruAge example uses) only uses v1, not v2 and upgrading it to v2 would result in an example that isn't grounded in the real world. 

I suggest we keep this as-is for now until we can get confirmation from the US convenience retail industry on their plans to upgrade to v2.0 for their credential refresh mechanism. AFAIK, they have no plans to start the upgrade process (which will take 2-3 years once started) until v2.0 is out.

iherman:
> 
> I suggest we keep this as-is for now until we can get confirmation from the US convenience retail industry on their plans to upgrade to v2.0 for their credential refresh mechanism. AFAIK, they have no plans to start the upgrade process (which will take 2-3 years once started) until v2.0 is out.

I do not think I agree. The example in the document must be along the line of this specification, and having a v1 example can become extremely confusing. It may not be a good idea for a new CR2 version, which is meant to be the Proposed Rec...

One possibility is to add a note making what you describe explicit, ie, that the example does not represent current deployment 100% because the US convenience retail industry has not yet upgraded to V2. (I guess that would require to use an imaginary `https://w3id.org/age/v2` as a context file as well.)

A suboptimal solution is to leave things as is but add a note which makes it explicitly that the example is incorrect V2 and should be updated to V2 when the time comes. But that still is confusing if people try to learn from examples and do not really read the spec... Hence suboptimal.

chaals:
I agree with Ivan's thinking here. People do copy examples and work from them, instead of carefully reading the spec, and it is important that the spec itself at least encourages that to work out right.

Explaining why an example isn't practising what the spec preaches helps, but is suboptimal, and leaving an example that doesn't match without very clear explanation of that is IMHO a Bad Idea™.

One approach might be to show the example from reality, noting the situation, and accompany it with an example of how it would look when updated.

(I do think it's good to have examples from reality, even though they represent a snapshot from the past by the time you publish them. But making that a strict test for anything labeled example is IMHO not the most helpful way to apply the principle.)

msporny:
Another option that ***we should definitely not do right now*** is to define refresh services in the WG... but that can wait for v2.1. :)

@peacekeeper, would you mind modifying this PR to add a warning about the example as @iherman and @chaals have suggested (even though it's suboptimal)?

Lp2125:
IMG-20241019-WA0024

brentzundel:
I would rather not have an example than have an example that doesn't conform to the specification, regardless of how realistic the example is.
I would rather have a completely invented example that conforms to the specification than a real-world example that doesn't.

Keeping the refreshService example as is would be a disservice to implementers, even with a warning.

selfissued:
I agree with @brentzundel that either the example should match THIS spec or it should be deleted.

iherman:
>  I would rather have a completely invented example that conforms to the specification than a real-world example that doesn't.

Yes.

msporny:
@peacekeeper can you make modifications to this PR given all the WG input above? It looks like the options are:

* Remove all examples of `refreshService`.
* Add a fake (but conforming) example  of `refreshService` (like an ExampleRefreshService).

I'd be ok with the second option, and it seems like others would as well. Any chance you can modify this PR? We need to find a way to merging this PR or closing it and trying again.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2025-01-15-vcwg#section1-4) on 2025-01-15

- no resolutions were taken

#### 1.4. Update "refreshService" example to VCDM2.0 (pr vc-data-model#1566)

_See github pull request [vc-data-model#1566](https://github.com/w3c/vc-data-model/pull/1566)._




**Brent Zundel:** This PR update refreshService.  

**Manu Sporny:** this was initially to address using real examples in the specification.  
… This was done mostly accept for refreshService.  
… We dont have a v2.0 example.  
… I think there is pushback to using an 2.0 example.  

</details>

msporny (approved):
Based on the discussion this past week, I'm withdrawing my objection to updating the example. 

I have ensured that the new example is a valid JSON-LD example (it is).

msporny:
Editorial, multiple reviews, changes requested (but not made), no objections, merging.

PR #1586, opened 2025-01-20
First item of VC context is a URL with the value https://www.w3.org/ns/credentials/v2

absoludity:
While working on a downstream use-case, I noticed that the existing schema checks only that the context *contains* "https://www.w3.org/ns/credentials/v2", while the [v2 spec explicitly says the following regarding context](https://www.w3.org/TR/vc-data-model-2.0/#contexts):

> The value of the @context [property](https://www.w3.org/TR/vc-data-model-2.0/#dfn-property) MUST be an [ordered set](https://infra.spec.whatwg.org/#ordered-set) where the first item is a [URL](https://www.w3.org/TR/vc-data-model-2.0/#dfn-url) with the value https://www.w3.org/ns/credentials/v2.

With that in mind, I think the use of `prefixItems` to ensure that the first item is the expected URL is more accurate.

While there, I also noted that:

> An ordered set is a [list](https://infra.spec.whatwg.org/#list) with the additional semantic that it must not contain the same [item](https://infra.spec.whatwg.org/#list-item) twice.

hence the "uniqueItems" addition.

Tests still pass with those changes (I was unsure if I should add examples that should fail and update the tests to ensure they fail? Let me know), although I do get the warning:

> strict mode: "prefixItems" is 1-tuple, but minItems or maxItems/items are not specified or different at path "#/properties/%40context"

which [appears to be a known ajv issue](https://github.com/ajv-validator/ajv/issues/2241).

I was also keen to address the last part of the spec for context:

> Subsequent items in the [ordered set](https://infra.spec.whatwg.org/#ordered-set) MUST be composed of any combination of [URLs](https://www.w3.org/TR/vc-data-model-2.0/#dfn-url) and objects,

but will do that in another PR if it's wanted.

Finally, where do I sign or agree to the w3c patent policy? (apparently the company I'm working for - gosource.com.au - isn't an affiliate member, but I've created an account at https://www.w3.org/account/request/ and listed gosource as my organisation link, which it allowed, and now linked my GH account)

Thanks

PR #1588, opened 2025-01-21
A context array item is a string or a non-empty object.

absoludity:
This follows on from #1586 where I mentioned:

> I was also keen to address the last part of the spec for context:
>
> >   Subsequent items in the [ordered set](https://infra.spec.whatwg.org/#ordered-set) MUST be composed of any combination of [URLs](https://www.w3.org/TR/vc-data-model-2.0/#dfn-url) and objects,
>
> but will do that in another PR if it's wanted.

and so this PR just ensures that the items in a `@context` array can only be either a non-empty string or a non-empty object.

I did play around with adding a `format` requiring an `iri` for the (non-object) strings, but it requires using [ajv-formats-draft2019](https://github.com/luzlab/ajv-formats-draft2019) in the test and I'm not sure that that's wanted here (let me know, happy to update it if so, and remove the `__fixture__/ajv` which appears to just be tech-debt - I could get tests running fine without it, just using the npm installed version of `ajv`).

If this PR is wanted, I'll probably need to rebase this once #1586 lands.


msporny (approved):

dlongley (approved):

msporny:

absoludity:
Thanks for the review and merge. Just wondering, is there interest in requiring the strings to be `"format": "iri"`? As mentioned above:

> I did play around with adding a `format` requiring an `iri` for the (non-object) strings, but it requires using [ajv-formats-draft2019](https://github.com/luzlab/ajv-formats-draft2019) in the test and I'm not sure that that's wanted here (let me know, happy to update it if so, and remove the `__fixture__/ajv` which appears to just be tech-debt - I could get tests running fine without it, just using the npm installed version of `ajv`).

Fine either way.

PR #1606, opened 2025-08-19
Delete .github/workflows directory

chaneloco:


iherman:
I will close, without further ado, all the PR-s that this user has submitted, considered as spam

PR #1630, opened 2026-05-11
Deprecate digest sri datatype vocab item

iherman:
And handled a bug in the template, which did not manifest itself yet.


iherman (commented):
@msporny, I did not want to commit directly into your branch, hence this one...

msporny (approved):

PR #1638, opened 2026-08-03
Add basic threat model.

msporny:
Attempt to address PR #1637 by adding a basic threat model for the VC Data model based on the W3C Threat Modeling guide. 

This PR is currently a work in progress.

Issue #1637, opened 2026-08-03
Add new W3C Threat Model to specification


msporny:
In order to complete our work in our current charter, we need to create a number of threat models for the VC Data Model, the VC API for Lifecycle Management protocol, Recognized Entities, and a variety of other specifications. Some specifications need to refer to the VC Data Model threat model. We need to create a base VC Data Model threat model that can be referred to from other specifications.

msporny:
PR #1638 and #1639 have been raised to address this issue. This issue will be closed once those PRs have been merged.

msporny:
PR #1638 and #1639 have been merged; closing.

PR #1639, opened 2026-08-09
Migrate Security/Privacy Considerations to Threat Model

msporny:
Attempt to address PR #1637 by refactoring the security and privacy considerations sections in the VCDM into the threat model. While the language had to be reworked a bit (some are threats, some are responses) to fit nicely into the way to do threat models, all the basic content should be the same (nothing really new, just re-stating what we've stated for many years now in the language of threat models).

This PR builds on top of PR #1638, which established the base threat model and data flow diagram. This PR purposefully does not delete any content from the VCDM. We will do that if this PR gets merged.

### DIFF
--- a/threat-model/threats/t11-bearer-credential-reuse.yaml
+++ b/threat-model/threats/t11-bearer-credential-reuse.yaml
@@ -1,0 +1,36 @@
+id: T11
+name: Bearer Credential Reuse and Correlation
+elements: [E1, E2, E3, P2, F3]
+tags: [privacy]
+taxonomyName: STRIDE
+taxonomyClass: Information Disclosure
+description: >-
+  A [=bearer credential=] entitles whoever holds it to a resource without
+  requiring the [=E2|holder=] to identify themselves, by omitting the
+  [=subject=] identifier normally expressed with the <code>id</code>
+  [=property=]. This privacy-enhancing design can still enable correlation.
+  Repeated use of the same [=bearer credential=] across multiple sites lets
+  those sites collude to track the [=holder=], and information that appears
+  non-identifying on its own, such as a birthdate and a postal code carried in
+  the same bearer credential, can statistically identify an individual. Whether
+  this threat is present depends on how an [=E1|issuer=] designs the bearer
+  credential and how the [=E2|holder=] uses it.
+response:
+  - id: R1
+    name: Single-Use, Non-Identifying Bearer Credentials
+    type: Reduce
+    description: >-
+      [=E1|Issuers=] of [=bearer credentials=] design them to be single-use
+      where possible, to avoid carrying personally identifying information, and
+      to avoid being unduly correlatable. [=E3|Verifiers=] avoid requesting
+      [=bearer credentials=] known to carry information that can be used to
+      illicitly correlate the [=holder=].
+  - id: R2
+    name: Warn Holders of Correlation Risk
+    type: Reduce
+    description: >-
+      Software for [=E2|holders=] warns the holder when it detects that a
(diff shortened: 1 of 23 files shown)


dlongley (approved):

dlongley:
I wonder if we want a statement somewhere about the fact that each threat in the model is treated in isolation -- so that solution providers know they really need to consider all relevant threats in aggregate.

Something like: "Each threat in this model is examined in isolation. However, good, scalable solutions always require balance, and almost any decision comes with trade offs. Readers should therefore take all relevant threats, advice, and mitigations into consideration to balance them appropriately to optimize for their solution's widest target audience."

threat-model/threats/t11-bearer-credential-reuse.yaml
```diff
@@ -0,0 +1,36 @@
...
+name: Bearer Credential Reuse and Correlation
+elements: [E1, E2, E3, P2, F3]
+tags: [privacy]
+taxonomyName: STRIDE
+taxonomyClass: Information Disclosure
+description: >-
+  A [=bearer credential=] entitles whoever holds it to a resource without
+  requiring the [=E2|holder=] to identify themselves, by omitting the
+  [=subject=] identifier normally expressed with the <code>id</code>
+  [=property=]. This privacy-enhancing design can still enable correlation.
```
iherman, line 11: I have a strange feeling about this. The text does not describe the threat; it rather describes the way the specification reacts to a threat, namely by making a bearer credential possible.  This sounds like the wrong order to me...

threat-model/threats/t12-correlation-via-status-lookup.yaml
```diff
@@ -0,0 +1,43 @@
...
+  population of [=subjects=] the list covers. Whether this threat is present
+  depends on the status mechanism an [=E1|issuer=] deploys and how often it is
+  updated.
+response:
+  - id: R1
+    name: Privacy-Preserving Status Mechanisms
+    type: Reduce
+    description: >-
+      [=E1|Issuers=] avoid status mechanisms, such as per-credential revocation
+      lists, that reveal which [=credential=] is being checked during
```
iherman, line 30: Shouldn't we put a reference to our very own status list spec? It is there between the lines, but not explicit...

iherman (approved):

threat-model/threats/t11-bearer-credential-reuse.yaml
msporny: This PR is meant to move what we have in the VCDM into the threat model, not re-write the content. The threat described is that a bearer credential, if re-used, can be used for correlation. That is what the text is describing, IMHO. 

If we want to revise the text, we'll have to do so in a separate PR, because you're both asking for a change in what we say in VCDM. I've opened an issue to track this concern:

https://github.com/w3c/vc-data-model/issues/1640

threat-model/threats/t12-correlation-via-status-lookup.yaml
msporny: Yes, and I'll do that in a separate PR. This PR is only meant to move all threats from the VCDM into the threat model, not add new content or concepts.

msporny:
@dlongley wrote:
> I wonder if we want a statement somewhere about the fact that each threat in the model is treated in isolation -- so that solution providers know they really need to consider all relevant threats in aggregate.

Fixed in 8192074d93a1146a4fee9c4b31deab7f1f095eac.

msporny:
Editorial, multiple reviews, changes requested and made, some requests migrated to issues, no objections, merging.

Issue #1641, opened 2026-08-19
Add a "localizeable value" term to address i18n concerns


msporny:
The specification is currently not clear that when we say "string" in some places, we really mean "a string or an object that expresses a string value and that has language and direction information attached/associated. This came up in an i18n review https://github.com/w3c/vc-recognized-entities/issues/111#issuecomment-5338398788 of the Recognized Entities specification.

The suggested fix was to do the following:

* Define an exported term that is along the lines of "localizable value" that states that the value is either a string or an object that follows the JSON-LD `@value/@language/@dir` pattern.
* Refer to that exported term from any VCWG specification that supports a localizable value.

See also: https://w3c.github.io/json-ld-syntax/#value-objects

iherman:
This was discussed during the [vcwg meeting on 19 August 2026](https://w3c.github.io/vc-wg/minutes/2026-08-19.html#8c31).

<details><summary><i>View the transcript</i></summary>

<h4><a href="https://github.com/w3c/vc-data-model/issues/1641" rel="noopener noreferrer">w3c/<wbr>vc-data-model#1641</a></h4>
<p><cite>manu:</cite> Re: 1641, the internationalization group gave feedback during the horizontal review, that we were being inconsistent in our use of string values , and we need to be specific about when strings are localizable.</p>
<p><cite>manu:</cite> the current proposal is to add a new term for a localizable value in the verifiable credential data model.</p>
<p><cite>manu:</cite> I can raise the pull request unless there is an objection, or if someone has a better way to address this issue.</p>
<p><cite>Ivan:</cite> I have no objection, there is separate vocab specification.  The tool has been upgraded to address the features Manu expressed, but the work needs to be done</p>
<p><cite>JoeAndrieu:</cite> +1 to management proposals, coming up with a term for what is localizable. How do we deal with images, because it is not covered and there may be no solution.</p>
<p><cite>manu:</cite> +1 to that Joe, I think this one is specific to string values, hyper focused, trying to solve the hyper focused thing, we're vague when we say string, need to fix that</p>
<p><cite>manu:</cite> other can of worms problem, absolutely struggling with that in Render Methods spec, tag render template with thing  - images, audio, everything covered, person viewing has said they have arabic as localization, I'll go to arabic template . . .</p>
<p><cite>&lt;dlongley&gt;</cite> +1 that "within the template" using the HTML render method might solve a lot</p>
[...]

PR #1646, opened 2026-09-05
Export definition of language value object; revise prose.

msporny:
Attempts to address issue #1641 by exporting the existing definition of `language value object` so that we can use it in other VC specifications.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1646.html" title="Last updated on Sep 13, 2026, 11:00 PM UTC (c326d70)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1646/28545b3...c326d70.html" title="Last updated on Sep 13, 2026, 11:00 PM UTC (c326d70)">Diff</a>


dlongley (approved):

index.html
```diff
@@ -6290,17 +6290,17 @@ <h3>Language and Base Direction</h3>
...
+possible across multiple expression syntaxes, such as [[[JSON-LD11]]],
+[[[JSON]]], and [[[?RFC7049]]].
         </p>
 
         <p>
-The general design pattern is to use the following markup template when
-expressing a text string that is tagged with a language and, optionally, a
-specific base direction.
+A <dfn class="export">language value object</dfn> is a human-readable text
+string, its corresponding language, and, optionally, its text direction. The
```
aphillips, line 6301: The language tag is also optional, defaulting to the language assigned to the credential as a whole, although it's strongly recommended ("SHOULD" in the text below). Not sure if you want to indicate that here.

aphillips (approved):
One question about indicating optionality of the the language tag. Otherwise LGTM. Thanks!!

index.html
msporny: I agonized back and forth on that when writing this sentence. I ended up landing on "Well, we say it's SHOULD later on (and that's the normative statement that matters), but we really want people to use the tag, so while it's technically optional, we really don't want to suggest that we don't have an opinion on whether or not you should use it vs. direction, which is less regularly necessary." IOW, you really should be using a language tag if you're going to use a language value object.

I also thought that maybe we should make it a MUST (go the other direction), but there are valid reasons to not have it (if you have a language for the document, which we discourage due to it's blanket application to all text fields -- which might be incorrect in some places).

All that said, I could go with any strong-ish argument in either direction. I think I like the text as-is, but wondering what others, including you, think.

iherman: I am ok with the text as is (including with the _SHOULD_ in the normative text some paragraphs down). This is an introductory text, kind of a non-normative paragraph, a bit of ambiguity is fine if it makes the text more readable imho.

I would definitely vote against a MUST; the spec does refer to the possibility of setting a default value (whether we like it or not, it is there).

iherman (approved):

iherman (approved):

index.html
filip26: To add another perspective, from the EU’s perspective (24 official languages), there is no default language; everything must be explicitly language-tagged. Enforcing this would mean that there would be no plain string representing textual information, but only language-value objects where the language is mandatory. This is very strict and perhaps overkill for many other use cases.

The proposed language-value object alongside plain strings, with no language explicitly set, seems like a good compromise, allowing a range of use cases, from strict language-tagging requirements to VCs where all text might default.

aphillips: Many documents are mono-lingual, in which case using `@language` at the document context level is fine. 

It is also sometimes helpful to have a default language, for cases in which the document's provided languages do not match the consumer's language preferences. For example, if the credential has the EU 24, but the recipient is Chinese. If the issuer has some bias or preference, indicating which language is the default can be useful for cases like this.

Note that we have [Developing Localizable Manifests](https://www.w3.org/TR/localizable-manifests/) that discusses some of the issues and use cases.

index.html
```diff
@@ -6312,17 +6312,17 @@ <h3>Language and Base Direction</h3>
...
-When the language value object is used in place of a string value, the object
-MUST contain a `@value` property whose value is a string, and SHOULD contain a
-`@language` property whose value is a string containing a well-formed
-`Language-Tag` as defined by [[BCP47]], and MAY contain a `@direction` property
-whose value is a [=base direction=] string defined by the `@direction`
-property in [[JSON-LD11]]. The language value object MUST NOT include any other
-keys beyond `@value`, `@language`, and `@direction`.
+When the [=language value object=] is used in place of a string value, the
+object MUST contain a `@value` property whose value is a string, and SHOULD
+contain a `@language` property whose value is a string containing a well-formed
```
aphillips, line 6317: It might be useful to link "well-formed", which has a specific meaning in BCP47. I18N Glossary has the term exported. https://www.w3.org/TR/i18n-glossary/#dfn-well-formed

Note that there is a dual recommendation for language tags. We recommend that specifications only _require_ well-formed tags, so that implementations are not required to validate the language tags every time they interact with them. But we suggest that specifications require language tags be "valid" (that is, authors should check the contents of the tags using the IETF subtag registry and perform the necessary canonicalizations when writing the values). You can find discussion in our best practices around [here](https://www.w3.org/TR/international-specs/#lang_values_specificity)

To implement, I would probably break up this long sentence. Approximately:

> When the `language value object` is used in place of a string value, the object MUST contain a `@value` property, SHOULD contain a `@language` property, and MAY also contain a `@direction` property.
> The `@value` property consists of a string.
> The `@language` property consists of a string that is a _well-formed_ language tag as defined by \[BCP47\]. 
> Any language tags used SHOULD be _valid_.
> The `@direction` property consists of a string whose value is a _base direction_ string as defined by the `@direction` property in \[JSON-LD11\]

dmitrizagidulin (approved):

shigeya (approved):

index.html
msporny: Fixed. Used your text in 81d0a4c63b535f0740066ee90c3ae28d9aacff7a.

msporny:

Issue #1643, opened 2026-08-23
Suggestion: Show mixed localized/non-localized arrays in i18n examples


filip26:
While there are existing examples of i18n, it might be helpful to explicitly note that non-localized values can be mixed with language-tagged localized objects in the same array. Adding a mixed example like the following could clarify this valid JSON-LD pattern for implementers:

```JSON
"description": [
  "Popis kryptograficky ověřitelného dokladu",
  "Alternativní popis v češtině",
  {
    "@value": "Credential Description in English",
    "@language": "en"
  }
]
```

Note, this example is just syntactic sugar over using value objects with just the `@value` property.

msporny:
PR #1647 has been raised to address this issue. This issue will be closed once PR #1647 has been merged.

msporny:
PR #1647 has been merged, closing.

PR #1647, opened 2026-09-05
Add an example of a mixed array of strings and language value objects

msporny:
This PR is an attempt at addressing issue #1643 by adding an example to the internationalization section that demonstrates how you can have a mixed array of strings and language value objects associated with a single property.

This PR also removes an example that demonstrates how one would use a language index, which the WG suggests not doing in the very next section.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/pull/1647.html" title="Last updated on Sep 13, 2026, 11:05 PM UTC (bd628d2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-data-model/1647/b73b6d1...bd628d2.html" title="Last updated on Sep 13, 2026, 11:05 PM UTC (bd628d2)">Diff</a>


dlongley (approved):

index.html
```diff
@@ -6372,39 +6372,21 @@ <h3>Language and Base Direction</h3>
...
-          "en": "Successfully installed the Example application."
-        }
-      }
-    }
+        <pre class="example nohighlight" title="Text strings and language value objects can be mixed in a single array">
+"title": [
+  "<span class="highlight">Příklad názvu knihy</span>",
+  {<span class="highlight">
+    "@value": "An Example of a Book Title",
+    "@language": "en"</span>
```
aphillips: I like this example (because it doesn't privilege the English!!), although having the Czech without a language value object (and thus, presumably, in the default language of the credential) might be confusing. 

Note that the `<span>` in your spec markup here should have `lang="cs"` (as a best practice) to facilitate rendering and especially assistive technologies working with the spec.

index.html
```diff
@@ -6372,39 +6372,21 @@ <h3>Language and Base Direction</h3>
...
-      "criteria": {
-        "narrative": {
-          "es": "Instaló exitosamente de la aplicación Example.",
-          "en": "Successfully installed the Example application."
-        }
-      }
-    }
+        <pre class="example nohighlight" title="Text strings and language value objects can be mixed in a single array">
+"title": [
+  "<span class="highlight">Příklad názvu knihy</span>",
```
msporny:
```suggestion
  "<span class="highlight" lang="cs">Příklad názvu knihy</span>",
```

index.html
msporny: Done in 0536661.

index.html
```diff
@@ -6372,39 +6372,21 @@ <h3>Language and Base Direction</h3>
...
-      "criteria": {
-        "narrative": {
-          "es": "Instaló exitosamente de la aplicación Example.",
-          "en": "Successfully installed the Example application."
-        }
-      }
-    }
+        <pre class="example nohighlight" title="Text strings and language value objects can be mixed in a single array">
+"title": [
+  "<span class="highlight" lang="cs">Příklad názvu knihy</span>",
```
iherman, line 6382: In #1646 we are arguing that the text _SHOULD_ set the language. Do we really want to add an example that does not follow that? There are pros and cons, but if we decide to keep this, then we may want to add a note after the example saying that one should not really do that, but use the structure with `@language`.

Many people use the example as a learning material, without looking at the details of the spec…

iherman (approved):

index.html
msporny: @filip26 thoughts on the above. While it's possible, it's probably not a good practice because a machine could only determine the language through a LLM of some kind (or any program that depends heavily on heuristics). IOW, do we want to provide an example that is not a best practice?

filip26: The issue I raised was to show that you can easily mix untagged strings with tagged strings in one array. It is basically just syntactic sugar over value objects that contain only the value.

Whether it is good practice is another question; clearly, there are pros and cons. 

Given the state of adoption, I favor this approach instead of a strict rule requiring a plain string (the question of what the default language is remains) or an array of language-tagged objects only. What happens when one value in the array has no `@language` defined? Should it make the whole VC invalid? etc.

filip26: Looking at the #1646 discussion, if we say that the language tag is optional, then it implies that this is conceptually correct. Otherwise, we should say that it is either a single string in the default language or an array of language-value objects, where the language tag must be present for every element.

msporny: > What happens when one value in the array has no @language defined? Should it make the whole VC invalid? etc.

There is an answer to this question here:

https://www.w3.org/TR/vc-data-model-2.1/#providing-default-language-and-direction

> Natural language string values that do not have a language associated with them SHOULD be treated as if the language value is undefined (language tag "und"). Natural language string values that do not have a base direction associated with them SHOULD be treated as if the direction value is "auto".

... so it doesn't make the whole VC invalid.

I think what you're saying is that we shouldn't forbid mixed arrays, which I agree with... and it looks like the language in name and description doesn't provide for that allowance:

https://www.w3.org/TR/vc-data-model-2.1/#names-and-descriptions

... so, we should probably change the specs to allow all variations because they are all legal, but perhaps we should prefer some expressions over others.

I don't have a strong opinion on any of this as long as i18n is happy and we are consistent across all of the specifications.

dmitrizagidulin (approved):

index.html
msporny: Fixed in bd628d2405bd60f8f3a11de067cdd700005ae7ef.

msporny:

